PR #425, opened 2020-08-24
[CIS-200] New channel observation

mantergo:
After creating new channel it's not observed cause it's not linked to any ChannelListQuery. 
So we need some `Worker` to do the job.

**In this PR:** 

- Introduce `NewChannelQueryUpdater`

### DIFF
--- a/Sources_v3/ChatClient.swift
+++ b/Sources_v3/ChatClient.swift
@@ -158,7 +158,8 @@ public class Client<ExtraData: ExtraDataTypes> {
     public convenience init(config: ChatClientConfig) {
         // All production workers
         let workerBuilders: [WorkerBuilder] = [
-            MessageSender<ExtraData>.init
+            MessageSender<ExtraData>.init,
+            NewChannelQueryUpdater<ExtraData>.init
         ]
         
         self.init(config: config,
--- a/Sources_v3/ChatClient_Tests.swift
+++ b/Sources_v3/ChatClient_Tests.swift
@@ -34,6 +34,11 @@ class ChatClient_Tests: StressTestCase {
         super.tearDown()
     }
     
+    var workerBuilders: [WorkerBuilder] = [
+        MessageSender<DefaultDataTypes>.init,
+        NewChannelQueryUpdater<DefaultDataTypes>.init
+    ]
+    
     // MARK: - Database stack tests
     
     func test_clientDatabaseStackInitialization_whenLocalStorageEnabled_respectsConfigValues() {
@@ -119,7 +124,7 @@ class ChatClient_Tests: StressTestCase {
         // Use in-memory store
         // Create a new chat client
         let client = ChatClient(config: inMemoryStorageConfig,
-                                workerBuilders: [MessageSender<DefaultDataTypes>.init],
+                                workerBuilders: workerBuilders,
                                 environment: testEnv.environment)
         
         // Assert the init parameters are correct
@@ -140,7 +145,7 @@ class ChatClient_Tests: StressTestCase {
     func test_clientProvidesConnectionId() throws {
         // Create a new chat client
         let client = ChatClient(config: inMemoryStorageConfig,
-                                workerBuilders: [MessageSender<DefaultDataTypes>.init],
+                                workerBuilders: workerBuilders,
                                 environment: testEnv.environment)
         
         // Set a connection Id waiter and assert it's `nil`
@@ -180,7 +185,7 @@ class ChatClient_Tests: StressTestCase {
     func test_clientProvidesToken() throws {
         // Create a new chat client
         let client = ChatClient(config: inMemoryStorageConfig,
-                                workerBuilders: [MessageSender<DefaultDataTypes>.init],
+                                workerBuilders: workerBuilders,
                                 environment: testEnv.environment)
         
         // Assert the token of anonymous user is `nil`
@@ -206,7 +211,7 @@ class ChatClient_Tests: StressTestCase {
         
         // Create a new chat client
         let client = ChatClient(config: config,
-                                workerBuilders: [MessageSender<DefaultDataTypes>.init],
+                                workerBuilders: workerBuilders,
                                 environment: testEnv.environment)
         
         // Assert the token of anonymous user is `nil`
@@ -223,7 +228,7 @@ class ChatClient_Tests: StressTestCase {
     func test_apiClientIsInitialized() throws {
         // Create a new chat client
         _ = ChatClient(config: inMemoryStorageConfig,
-                       workerBuilders: [MessageSender<DefaultDataTypes>.init],
+                       workerBuilders: workerBuilders,
                        environment: testEnv.environment)
         
         assertMandatoryHeaderFields(testEnv.apiClient?.init_sessionConfiguration)
@@ -240,6 +245,7 @@ class ChatClient_Tests: StressTestCase {
         
         // Check all the mandatory background workers are initialized
         XCTAssert(client.backgroundWorkers.contains { $0 is MessageSender<DefaultDataTypes> })
+        XCTAssert(client.backgroundWorkers.contains { $0 is NewChannelQueryUpdater<DefaultDataTypes> })
     }
     
     func test_backgroundWorkersAreInitialized() {
@@ -284,7 +290,7 @@ class ChatClient_Tests: StressTestCase {
     
     func test_settingAnonymousUser() {
         let client = Client(config: inMemoryStorageConfig,
-                            workerBuilders: [MessageSender<DefaultDataTypes>.init],
+                            workerBuilders: workerBuilders,
                             environment: testEnv.environment)
         assert(testEnv.webSocketClient!.connect_calledCounter == 0)
         
@@ -323,7 +329,7 @@ class ChatClient_Tests: StressTestCase {
     
     func test_settingUser() {
         let client = Client(config: inMemoryStorageConfig,
-                            workerBuilders: [MessageSender<DefaultDataTypes>.init],
+                            workerBuilders: workerBuilders,
                             environment: testEnv.environment)
         
         assert(testEnv.webSocketClient!.connect_calledCounter == 0)
@@ -382,7 +388,7 @@ class ChatClient_Tests: StressTestCase {
     
     func test_settingGuestUser() {
         let client = Client(config: inMemoryStorageConfig,
-                            workerBuilders: [MessageSender<DefaultDataTypes>.init],
+                            workerBuilders: workerBuilders,
                             environment: testEnv.environment)
         
         assert(testEnv.webSocketClient!.connect_calledCounter == 0)
@@ -461,7 +467,7 @@ class ChatClient_Tests: StressTestCase {
     func test_disconnectAndConnect() {
         // Set up a new anonymous user
         let client = Client(config: inMemoryStorageConfig,
-                            workerBuilders: [MessageSender<DefaultDataTypes>.init],
+                            workerBuilders: workerBuilders,
                             environment: testEnv.environment)
         
         // Set up a new anonymous user and wait for completion
--- a/Sources_v3/Controllers/ListDatabaseObserver.swift
+++ b/Sources_v3/Controllers/ListDatabaseObserver.swift
@@ -187,3 +187,16 @@ class ListChangeAggregator<DTO: NSManagedObject, Item>: NSObject, NSFetchedResul
         onChange?(currentChanges)
     }
 }
+
+extension ListDatabaseObserver where DTO == Item {
+    convenience init(
+        context: NSManagedObjectContext,
+        fetchRequest: NSFetchRequest<DTO>
+    ) {
+        self.init(
+            context: context,
+            fetchRequest: fetchRequest,
+            itemCreator: { $0 }
+        )
+    }
+}
--- a/Sources_v3/Database/DTOs/ChannelDTO.swift
+++ b/Sources_v3/Database/DTOs/ChannelDTO.swift
@@ -117,13 +117,20 @@ extension NSManagedObjectContext {
 
 extension ChannelDTO {
     static func channelListFetchRequest(query: ChannelListQuery) -> NSFetchRequest<ChannelDTO> {
-        let request = NSFetchRequest<ChannelDTO>(entityName: "ChannelDTO")
+        let request = NSFetchRequest<ChannelDTO>(entityName: ChannelDTO.entityName)
         
         // Fetch results controller requires at least one sorting descriptor.
         let sortDescriptors = query.sort.compactMap { $0.key.sortDescriptor(isAscending: $0.isAscending) }
         request.sortDescriptors = sortDescriptors.isEmpty ? [ChannelListSortingKey.defaultSortDescriptor] : sortDescriptors
         
         request.predicate = NSPredicate(format: "ANY queries.filterHash == %@", query.filter.filterHash)
+        return request
+    }
+
+    static var channelWithoutQueryFetchRequest: NSFetchRequest<ChannelDTO> {
+        let request = NSFetchRequest<ChannelDTO>(entityName: ChannelDTO.entityName)
+        request.sortDescriptors = [ChannelListSortingKey.defaultSortDescriptor]
+        request.predicate = NSPredicate(format: "queries.@count == 0")
         return request
     }
 }
--- a/Sources_v3/Database/DTOs/ChannelListQueryDTO.swift
+++ b/Sources_v3/Database/DTOs/ChannelListQueryDTO.swift
@@ -23,21 +23,17 @@ class ChannelListQueryDTO: NSManagedObject {
         request.predicate = NSPredicate(format: "filterHash == %@", filterHash)
         return try? context.fetch(request).first
     }
-    
-    static func loadOrCreate(filterHash: String, context: NSManagedObjectContext) -> ChannelListQueryDTO {
-        if let existing = Self.load(filterHash: filterHash, context: context) {
-            return existing
-        }
-        
-        let new = NSEntityDescription.insertNewObject(forEntityName: Self.entityName, into: context) as! ChannelListQueryDTO
-        new.filterHash = filterHash
-        return new
-    }
 }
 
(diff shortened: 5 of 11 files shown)


Sources_v3/Controllers/ChannelListController.swift
```diff
@@ -71,6 +71,9 @@ public class ChannelListControllerGeneric<ExtraData: ExtraDataTypes>: Controller
         return observer
     }()
     
+    /// Used for observing new channels creation and updating required `ChannelListQueries`.
+    private lazy var newChannelObserver = NewChannelObserver<ExtraData>(channelListQueryUpdated: worker)
+
```
VojtaStavik, line 76: Should be a background worker in `Client`

Sources_v3/Database/DTOs/ChannelDTO.swift
```diff
@@ -126,6 +126,12 @@ extension ChannelDTO {
         request.predicate = NSPredicate(format: "ANY queries.filterHash == %@", query.filter.filterHash)
         return request
     }
+
+    static var channelListFetchRequestAll: NSFetchRequest<ChannelDTO> {
+        let request = NSFetchRequest<ChannelDTO>(entityName: ChannelDTO.entityName)
+        request.sortDescriptors = [ChannelListSortingKey.defaultSortDescriptor]
+        return request
+    }
```
VojtaStavik, line 134: Or this can be channels without any filterHash

Sources_v3/Database/DTOs/ChannelListQueryDTO.swift
```diff
@@ -12,7 +12,7 @@ class ChannelListQueryDTO: NSManagedObject {
     @NSManaged var filterHash: String
     
     /// Serialized `Filter` JSON which can be used in cases the query needs to be repeated, i.e. for newly created channels.
-    @NSManaged var filterJSONData: Data
+    @NSManaged var filterJSONData: Data?
```
VojtaStavik, line 15: This can't be optional. We must be able to always store the serialized filter, no?

Sources_v3/Query/Filter.swift
```diff
@@ -62,6 +62,11 @@ public enum Filter: Encodable, CustomStringConvertible {
     /// Filter without any of filters (like `not or`).
     indirect case nor([Filter])
     
+    // MARK: - Technical
+
+    // Add comment!
+    indirect case explicitFilterHash(Filter, String)
+
```
VojtaStavik, line 69: Do we still need this? Maybe we can just change `ChannelUpdater.update` method to accept optional explicit filter hash?

Sources_v3/Workers/Background/NewChannelObserver.swift
```diff
@@ -0,0 +1,74 @@
...
+import CoreData
+
+final class NewChannelObserver<ExtraData: ExtraDataTypes>: Worker, NSFetchedResultsControllerDelegate {
+    private let channelListQueryUpdated: ChannelListQueryUpdater<ExtraData>
+    private let fetchRequest = ChannelDTO.channelListFetchRequestAll
+    
+    private(set) lazy var frc: NSFetchedResultsController<ChannelDTO> = .init(fetchRequest: self.fetchRequest,
+                                                                              managedObjectContext: self.database.writableContext,
+                                                                              sectionNameKeyPath: nil,
+                                                                              cacheName: nil)
```
VojtaStavik, line 14: You can use `ListDatabaseObserver`. See `MessageSender` for reference

Sources_v3/Database/DTOs/ChannelDTO.swift
mantergo: Nice!

Sources_v3/Database/DTOs/ChannelListQueryDTO.swift
mantergo: It should have filterJSONData, right.
I just don't want this `Data` to be overwritten. 
As far as I understand we can save ChannelListQueryDTO only through function that is inside `extension NSManagedObjectContext` and we are trying to keep this rule regarding all DTOs. In this case I believe it's safe enough.

Sources_v3/Query/Filter.swift
mantergo: It was first idea, but in that case we will have to modify 3 functions. 
It seems like this technical enum solves this issue nicer.

Sources_v3/Workers/Background/NewChannelObserver.swift
mantergo: I started with `ListDatabaseObserver` but I don't need `ItemCreator` which is mandatory, so I don't think I should use some class that has functionality which is not needed and doesn't bring much simplicity in my case.

Sources_v3/Database/DTOs/ChannelListQueryDTO.swift
VojtaStavik: I see. In that case, I'd suggest doing just:
```swift
    func saveQuery(query: ChannelListQuery) -> ChannelListQueryDTO {
        if let existingDTO = ChannelListQueryDTO.load(filterHash: query.filter.filterHash, context: self) {
            return existingDTO
        }
        ...
```
to prevent orerriding the query, if the entity already exists.

Sources_v3/Query/Filter.swift
VojtaStavik: All right then 👍

Sources_v3/Workers/Background/NewChannelObserver.swift
VojtaStavik: I see, however, I'd suggest using the `ListDatabaseObserver` anyway. It will enable us to do some further optimizations, like sharing the underlying FRCs for multiple observers, etc., without having to touch the "public" api.

Maybe, we could introduce some convenience init for the observers without the creator if `Item == DTO`.

Sources_v3/Controllers/ChannelListController.swift
mantergo: What do u think about `NewChannelQueryUpdater`?

Sources_v3/Controllers/ListDatabaseObserver.swift
```diff
@@ -187,3 +187,16 @@ class ListChangeAggregator<DTO: NSManagedObject, Item>: NSObject, NSFetchedResul
         onChange?(currentChanges)
     }
 }
+
+extension ListDatabaseObserver where DTO == Item {
+    convenience init(
+        context: NSManagedObjectContext,
+        fetchRequest: NSFetchRequest<DTO>
+    ) {
```
VojtaStavik, line 195: ;-)

Sources_v3/ChatClient.swift
```diff
@@ -158,7 +158,8 @@ public class Client<ExtraData: ExtraDataTypes> {
     public convenience init(config: ChatClientConfig) {
         // All production workers
         let workerBuilders: [WorkerBuilder] = [
-            MessageSender<ExtraData>.init
+            MessageSender<ExtraData>.init,
+            NewChannelQueryUpdater<ExtraData>.init
```
VojtaStavik, line 162: I like the name 👍

Sources_v3/Database/DTOs/ChannelDTO.swift
```diff
@@ -126,6 +126,13 @@ extension ChannelDTO {
         request.predicate = NSPredicate(format: "ANY queries.filterHash == %@", query.filter.filterHash)
         return request
     }
+
+    static var newChannelsFetchRequest: NSFetchRequest<ChannelDTO> {
```
VojtaStavik, line 130: I'd suggest renaming this to `channelWithoutQueryFetchRequest`. It's possible to have channels without queries. It doesn't mean they're always new.

Sources_v3/Workers/Background/NewChannelQueryUpdater.swift
```diff
@@ -0,0 +1,113 @@
...
+    init(database: DatabaseContainer, webSocketClient: WebSocketClient, apiClient: APIClient, env: Environment) {
+        environment = env
+        super.init(database: database, webSocketClient: webSocketClient, apiClient: apiClient)
+        
+        startObserving()
+    }
+    
+    override convenience init(database: DatabaseContainer, webSocketClient: WebSocketClient, apiClient: APIClient) {
+        self.init(database: database, webSocketClient: webSocketClient, apiClient: apiClient, env: .init())
+    }
```
VojtaStavik, line 48: Ugh, I see there's a problem with initializers from the superclass. 🤔 Any idea how to solve this?

Sources_v3/Workers/Background/NewChannelQueryUpdater.swift
```diff
@@ -0,0 +1,113 @@
...
+    typealias DatabaseObserver = ListDatabaseObserver<ChannelDTO, ChannelDTO>
+    
+    private let environment: Environment
+        
+    private lazy var channelListQueryUpdater: ChannelListQueryUpdater<ExtraData> = self.environment
+        .createChannelListQueryUpdater(database,
+                                       webSocketClient,
+                                       apiClient)
+    
+    private lazy var channelsObserver: ListDatabaseObserver = .init(context: self.database.writableContext,
```
VojtaStavik, line 25: The observer should be attached to `backgroundReadOnlyContext`

Sources_v3/Workers/Background/NewChannelQueryUpdater.swift
```diff
@@ -0,0 +1,113 @@
...
+    
+    override convenience init(database: DatabaseContainer, webSocketClient: WebSocketClient, apiClient: APIClient) {
+        self.init(database: database, webSocketClient: webSocketClient, apiClient: apiClient, env: .init())
+    }
+    
+    private func startObserving() {
+        // The observer can be set up on the background queue
+        DispatchQueue.global().async {
+            do {
+                self.channelsObserver.onChange = { [unowned self] changes in
```
VojtaStavik, line 54: Be careful, even the `self` in `self.channelsObserver` can cause issues. It must be `async { [weak self] in ... }`. I had similar issues and random crashes with `MessageSender`.

Sources_v3/Workers/Background/NewChannelQueryUpdater.swift
```diff
@@ -0,0 +1,113 @@
...
+                    self.updateChannelListQuery(changes: changes)
+                }
+                try self.channelsObserver.startObserving()
+                
+            } catch {
+                log.error("Error starting MessageSender observer: \(error)")
```
VojtaStavik, line 60: `MessageSender` :)

Sources_v3/Workers/Background/NewChannelQueryUpdater.swift
```diff
@@ -0,0 +1,113 @@
...
+    }
+    
+    private func startObserving() {
+        // The observer can be set up on the background queue
+        DispatchQueue.global().async {
+            do {
+                self.channelsObserver.onChange = { [unowned self] changes in
+                    self.updateChannelListQuery(changes: changes)
+                }
+                try self.channelsObserver.startObserving()
```
VojtaStavik, line 57: You also need to account for already existing "new" channels in the database.

Sources_v3/ChatClient.swift
VojtaStavik: Please also add a test for this, since it's a mandatory worker.

Sources_v3/Workers/Background/NewChannelQueryUpdater.swift
```diff
@@ -0,0 +1,113 @@
...
+
+/// After creating new channel it's not observed cause it's not linked to any ChannelListQuery.
+/// The only job of `NewChannelQueryUpdater` is to find whether new channel belongs to any of the exsisting queries
+/// and link it to the channel if so.
+///     1. This worker observers DB for the insertations of the new `ChannelDTO`s without any linked queries.
+///     2. When new channel is found, all exsisting queries are fetched from DB and we modify exsiting queries filters so
+///     in response for `update(channelListQuery` request new channel will be returned if it is part of the original query filter.
+///     3. After sending `update(channelListQuery` for all queries `ChannelListQueryUpdater` does the job of linking
+///     corresponding queries to the channel.
+final class NewChannelQueryUpdater<ExtraData: ExtraDataTypes>: Worker, NSFetchedResultsControllerDelegate {
```
VojtaStavik, line 15: Doesn't have to be `NSFetchedResultsControllerDelegate`

Sources_v3/Workers/Background/NewChannelQueryUpdater.swift
```diff
@@ -0,0 +1,113 @@
...
+    typealias DatabaseObserver = ListDatabaseObserver<ChannelDTO, ChannelDTO>
```
VojtaStavik, line 16: This doesn't seem to be used?

Sources_v3/Workers/Background/NewChannelQueryUpdater.swift
```diff
@@ -0,0 +1,113 @@
...
+    private lazy var channelsObserver: ListDatabaseObserver = .init(context: self.database.writableContext,
+                                                                    fetchRequest: ChannelDTO.newChannelsFetchRequest)
+    
+    private var queries: [ChannelListQueryDTO] {
+        do {
+            let queries = try database.backgroundReadOnlyContext
+                .fetch(NSFetchRequest<ChannelListQueryDTO>(entityName: ChannelListQueryDTO.entityName))
+            return queries
+        } catch {
+            log.error("Internal error: Failed to fetch [ChannelListQueryDTO].")
```
VojtaStavik, line 34: Please also include the error to the message we can easily debug.

Sources_v3/Workers/Background/NewChannelQueryUpdater.swift
```diff
@@ -0,0 +1,113 @@
...
+                var updatedQueries: [ChannelListQuery] = []
+                
+                do {
+                    updatedQueries = try queries.map {
+                        // Modify original query filter
+                        try $0.asChannelListQueryWithUpdatedFilter(filterToAdd: .equal("cid", to: channelDTO.cid))
+                    }
+                    
+                } catch {
+                    log.error("Internal error. Failed to update ChannelListQueries for the new channel.")
```
VojtaStavik, line 79: Same here

Sources_v3/Workers/Background/NewChannelQueryUpdater.swift
```diff
@@ -0,0 +1,113 @@
...
+                }
+                
+                // Send `update(channelListQuery:` requests so corresponding queries will be linked to the channel
+                updatedQueries.forEach {
+                    channelListQueryUpdater.update(channelListQuery: $0) { error in
+                        log
+                            .error("Internal error. Failed to update ChannelListQueries for the new channel: \(String(describing: error))")
```
VojtaStavik, line 86: If I'm not mistaken, this closure is called always, not just when error.

Sources_v3/Workers/Background/NewChannelQueryUpdater_Tests.swift
```diff
@@ -0,0 +1,105 @@
...
+        let filter1: Filter = .contains(.unique, String.unique)
+        let filter2: Filter = .notEqual(.unique, to: 1)
+        
+        try database.createChannelListQuery(filter: filter1)
+        try database.createChannelListQuery(filter: filter2)
+                
+        try database.createChannel()
+        
+        AssertAsync.willBeEqual(env!.channelQueryUpdater?.update_calls_counter, 2)
+    }
```
VojtaStavik, line 71: It's funny how `NewChannelQueryUpdater` is not even mentioned in this test :)

Sources_v3/Workers/Background/NewChannelQueryUpdater_Tests.swift
```diff
@@ -0,0 +1,105 @@
...
+    func test_filter_is_Modified() throws {
+        let cid: ChannelId = .unique
+        let filter: Filter = .notEqual(.unique, to: 1)
+        
+        try database.createChannelListQuery(filter: filter)
+        try database.createChannel(cid: cid)
+        
+        let expectedFilter: Filter = .and([filter, .equal("cid", to: cid)])
+        
+        AssertAsync.willBeEqual(env!.channelQueryUpdater?.update_query?.filter.description, expectedFilter.description)
```
VojtaStavik, line 82: Can we also assert that even though the filter was modified, the hash stays the same?

VojtaStavik (changes requested):
Great work, thanks! See, if it wasn't the Filter decoding, it would have been a simple task as I promised :))

Let's try to figure out the `init` situation of `Worker`

Sources_v3/Workers/Background/NewChannelQueryUpdater_Tests.swift
mantergo: It's due to special behavior of our `technical` enum case.
It returns filter hash from original `Filter`, but description taken from the modified filter that is inside this enum.

VojtaStavik: But the hash should be also the same, no? I think we should verify that:
`env!.channelQueryUpdater?.update_query?.filter.hash == expectedFilter.hash`

mantergo: We keep the original` filterHash` of DB query so later we can fetch it to link to our new channel.

mantergo: And our technical enum returns this original `filterHash`, not the hash of underlying filter.

Sources_v3/ChatClient_Tests.swift
```diff
@@ -34,6 +34,11 @@ class ChatClient_Tests: StressTestCase {
         super.tearDown()
     }
     
+    var workerBuilders: [WorkerBuilder] = [
+        MessageSender<DefaultDataTypes>.init,
+        NewChannelQueryUpdater<DefaultDataTypes>.init
+    ]
```
VojtaStavik, line 40: 👍

PR #426, opened 2020-08-25
Add composer to v3 simple chat sample

cardoso:
Replaced the alert with a simple composer similar to iMessage

Screen Shot 2020-08-24 at 22 28 33

#no_changelog


VojtaStavik (approved):
Looks like you've done this before :)) 

One tip: I think you can the `/Sample_v3` folder to `pre-commit.sh`, too. It will make sure we have the same code formatting both in the SDK and in the Sample app.

You may have to run `bootstrap.sh` script again, too.

cardoso:
Good tip! I'll add it in the next PR

PR #427, opened 2020-08-25
[CIS-289] Extract event-posting logic

evsaev:
**This PR:**
- Introduces `EventNotificationCenter` that pre-process incoming events before being published
- Uses `EventNotificationCenter` in `WebSocketClient`

### DIFF
--- a/Sources_v3/ChatClient.swift
+++ b/Sources_v3/ChatClient.swift
@@ -54,7 +54,7 @@ public class Client<ExtraData: ExtraDataTypes> {
     private(set) var backgroundWorkers: [Worker]!
             
     /// The notification center used to send and receive notifications about incoming events.
-    private(set) lazy var notificationCenter = environment.notificationCenterBuilder([
+    private(set) lazy var eventNotificationCenter = environment.notificationCenterBuilder([
         EventDataProcessorMiddleware<ExtraData>(database: databaseContainer)
     ])
     
@@ -84,7 +84,7 @@ public class Client<ExtraData: ExtraDataTypes> {
                                                                  urlSessionConfiguration,
                                                                  encoder,
                                                                  EventDecoder<ExtraData>(),
-                                                                 notificationCenter)
+                                                                 eventNotificationCenter)
         
         webSocketClient.connectionStateDelegate = self
         
@@ -410,7 +410,7 @@ extension Client {
                             sessionConfiguration: $1,
                             requestEncoder: $2,
                             eventDecoder: $3,
-                            notificationCenter: $4)
+                            eventNotificationCenter: $4)
         }
         
         var databaseContainerBuilder: (_ kind: DatabaseContainer.Kind) throws -> DatabaseContainer = {
--- a/Sources_v3/ChatClient.swift
+++ b/Sources_v3/ChatClient.swift
@@ -52,6 +52,11 @@ public class Client<ExtraData: ExtraDataTypes> {
     /// observing the current state of the system and perform work if needed (i.e. when a new message pending sent appears in the
     /// database, a worker tries to send it.)
     private(set) var backgroundWorkers: [Worker]!
+            
+    /// The notification center used to send and receive notifications about incoming events.
+    private(set) lazy var notificationCenter = environment.notificationCenterBuilder([
+        EventDataProcessorMiddleware<ExtraData>(database: databaseContainer)
+    ])
     
     /// The `APIClient` instance `Client` uses to communicate with Stream REST API.
     lazy var apiClient: APIClient = {
@@ -74,17 +79,12 @@ public class Client<ExtraData: ExtraDataTypes> {
         
         let connectEndpoint = webSocketConnectEndpoint(userId: currentUserId)
         
-        // Create middlewares.
-        let eventMiddlewares: [EventMiddleware] = [
-            EventDataProcessorMiddleware<ExtraData>(database: databaseContainer)
-        ]
-        
         // Create a WebSocketClient.
         let webSocketClient = environment.webSocketClientBuilder(webSocketEndpoint,
                                                                  urlSessionConfiguration,
                                                                  encoder,
                                                                  EventDecoder<ExtraData>(),
-                                                                 eventMiddlewares)
+                                                                 notificationCenter)
         
         webSocketClient.connectionStateDelegate = self
         
@@ -404,13 +404,13 @@ extension Client {
             _ sessionConfiguration: URLSessionConfiguration,
             _ requestEncoder: RequestEncoder,
             _ eventDecoder: AnyEventDecoder,
-            _ eventMiddlewares: [EventMiddleware]
+            _ notificationCenter: EventNotificationCenter
         ) -> WebSocketClient = {
             WebSocketClient(connectEndpoint: $0,
                             sessionConfiguration: $1,
                             requestEncoder: $2,
                             eventDecoder: $3,
-                            eventMiddlewares: $4)
+                            notificationCenter: $4)
         }
         
         var databaseContainerBuilder: (_ kind: DatabaseContainer.Kind) throws -> DatabaseContainer = {
@@ -421,6 +421,8 @@ extension Client {
         var requestDecoderBuilder: () -> RequestDecoder = DefaultRequestDecoder.init
         
         var eventDecoderBuilder: () -> EventDecoder<ExtraData> = EventDecoder<ExtraData>.init
+        
+        var notificationCenterBuilder: ([EventMiddleware]) -> EventNotificationCenter = EventNotificationCenter.init
     }
 }
 
--- a/Sources_v3/WebSocketClient/HealthCheckMiddleware_Tests.swift
+++ b/Sources_v3/WebSocketClient/HealthCheckMiddleware_Tests.swift
@@ -1,0 +1,79 @@
+//
+// Copyright © 2020 Stream.io Inc. All rights reserved.
+//
+
+@testable import StreamChatClient
+import XCTest
+
+final class HealthCheckMiddleware_Tests: XCTestCase {
+    var middleware: HealthCheckMiddleware!
+    var webSocketClient: WebSocketClientMock!
+    
+    // MARK: - Setup
+    
+    override func setUp() {
+        super.setUp()
+        
+        webSocketClient = WebSocketClientMock()
+        middleware = HealthCheckMiddleware(webSocketClient: webSocketClient)
+    }
+    
+    override func tearDown() {
+        AssertAsync.canBeReleased(&webSocketClient)
+        
+        super.tearDown()
+    }
+    
+    // MARK: - Tests
+    
+    func test_middleware_forwardsNonHealthCheckEvents() throws {
+        let event = TestEvent()
+        
+        // Simulate incoming public event
+        let forwardedEvent = try await {
+            middleware.handle(event: event, completion: $0)
+        }
+        
+        // Assert event is forwared as it is
+        XCTAssertEqual(forwardedEvent as? TestEvent, event)
+    }
+    
+    func test_middleware_filtersHealthCheckEvents_ifClientIsDeallocated() throws {
+        let event = HealthCheckEvent(connectionId: .unique)
+        
+        // Deallocate the client
+        AssertAsync.canBeReleased(&webSocketClient)
+        
+        // Simulate `HealthCheckEvent`
+        var forwardedEvent: Event?
+        middleware.handle(event: event) {
+            forwardedEvent = $0
+        }
+        
+        // Assert event is not forwared
+        AssertAsync.staysTrue(forwardedEvent == nil)
+    }
+    
+    func test_middleware_handlesHealthCheckEvents() throws {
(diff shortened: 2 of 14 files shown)


Sources_v3/ChatClient.swift
```diff
@@ -52,6 +52,19 @@ public class Client<ExtraData: ExtraDataTypes> {
...
+            
+    /// The notification center used to send and receive notifications about incoming events.
+    private(set) lazy var eventNotificationCenter: EventNotificationCenter = {
+        // Create middlewares.
+        let eventMiddlewares: [EventMiddleware] = [
+            EventDataProcessorMiddleware<ExtraData>(database: databaseContainer)
+        ]
+        
+        // Create notification center
+        let notificationCenter = environment.createNotificationCenter()
```
VojtaStavik, line 64: This should probably be inside the `EventNotificationCenter` and its own environment

Sources_v3/Workers/EventNotificationCenter.swift
```diff
@@ -0,0 +1,37 @@
...
+class EventNotificationCenter {
+    let notificationCenter: NotificationCenter
+    let middlewares: [EventMiddleware]
+    
+    init(notificationCenter: NotificationCenter, middlewares: [EventMiddleware]) {
+        self.notificationCenter = notificationCenter
+        self.middlewares = middlewares
+    }
+    
+    func publish(_ event: Event, completion: ((Error?) -> Void)? = nil) {
```
VojtaStavik, line 16: I'd suggest renaming this to `process`. `publish` can be misleading since not every event must be published, if some of the middlewares decide not to do it.

Sources_v3/Workers/EventNotificationCenter.swift
```diff
@@ -0,0 +1,37 @@
+//
+// Copyright © 2020 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+class EventNotificationCenter {
```
VojtaStavik, line 7: I wonder, if it made sense to actually make this a subclass of `NotificationCenter`? When using it, you will do `eventNotificationCenter.notificationCenter` all the time, anyway.

Sources_v3/Workers/EventNotificationCenter.swift
```diff
@@ -0,0 +1,37 @@
...
+    func publish(_ event: Event, completion: ((Error?) -> Void)? = nil) {
+        middlewares.process(event: event) { [weak self] in
+            guard let self = self else { return }
+            
+            if let processedEvent = $0 {
+                let notification = Notification(newEventReceived: processedEvent, sender: self)
+                self.notificationCenter.post(notification)
+                completion?(nil)
+            } else {
+                completion?(ClientError.EventProcessingFailed(event: event))
```
VojtaStavik, line 25: It doesn't mean failure, it can just mean, the event is private, and shouldn't be published. I'm not sure if we even need this completion to be honest.

Sources_v3/WebSocketClient/WebSocketClient.swift
```diff
@@ -246,21 +258,7 @@ extension WebSocketClient: WebSocketEngineDelegate {
...
-                
-                if let event = event as? HealthCheckEvent {
-                    self.pingController.pongRecieved()
-                    
-                    if self.connectionState.isConnected == false {
-                        self.connectionState = .connected(connectionId: event.connectionId)
-                    }
-                } else {
-                    self.notificationCenter.post(Notification(newEventReceived: event, sender: self))
-                }
```
VojtaStavik, line 261: `HealthCheck` is a special kind of event. It's only WebSocketRelated. I'd suggest leaving the special handle logic there as it was and only forward non-health-check events.

Sources_v3/Workers/EventObservers/HealthCheckEventObserver.swift
```diff
@@ -0,0 +1,16 @@
...
+final class HealthCheckEventObserver: EventObserver {
+    init(
+        notificationCenter: NotificationCenter,
+        callback: @escaping (HealthCheckEvent) -> Void
+    ) {
+        super.init(notificationCenter: notificationCenter,
+                   tranform: { $0 as? HealthCheckEvent },
+                   callback: callback)
+    }
+}
```
VojtaStavik, line 16: This is not needed if we left the special case processing inside WS client.

Sources_v3/Workers/EventNotificationCenter.swift
evsaev: I was thinking about the case when we wanna publish multiple events (that we get from `/sync` endpoint, for example)

Won't it be better to wait until the current event is processed before starting with the next one?

VojtaStavik: I see. This hopefully won't be needed, as the order of the events doesn't really matter and the whole middleware pre-processing is built the way it can (should) be possible to run in parallel.

VojtaStavik: For example `database.write { }` happens on one serial queue. So if you give it a hundred events it should process them one by one in the order you provide them.

evsaev: understood. The completion is not there anymore

Sources_v3/WebSocketClient/WebSocketClient.swift
```diff
@@ -88,6 +88,9 @@ class WebSocketClient {
         return pingController
     }()
     
+    /// The observer used to listen to `HealthCheckEvent`
+    private var healthCheckEventObserver: HealthCheckEventObserver?
```
evsaev, line 92: Finally figured out that we need either **health-check event observer** OR a **completion** in `EventNotificationCenter.process`.

The reason is **HealthCheckEvent** should also be processed via middlewares (especially by `EventDataProcessorMiddleware`) as it contains current-user payload that should be stored in a DB.

Decided to go with the observer as we don't need a completion anywhere else.

Sources_v3/ChatClient.swift
evsaev: made a subclass in 9ccf3e7d5bc8ffbd0021704e53abe9deed9320bb

Sources_v3/Workers/EventNotificationCenter.swift
evsaev: done in 9ccf3e7d5bc8ffbd0021704e53abe9deed9320bb

Sources_v3/Workers/EventNotificationCenter.swift
evsaev: done in 9ccf3e7d5bc8ffbd0021704e53abe9deed9320bb

Sources_v3/Workers/EventNotificationCenter.swift
evsaev: done in 9ccf3e7d5bc8ffbd0021704e53abe9deed9320bb

Sources_v3/WebSocketClient/WebSocketClient.swift
evsaev: Please check https://github.com/GetStream/stream-chat-swift/pull/427#discussion_r477208120

Sources_v3/WebSocketClient/WebSocketClient_Tests.swift
```diff
@@ -530,8 +536,12 @@ extension WebSocketEngineError: Equatable {
 }
 
 /// A test middleware that can be initiated with a closure/
-private struct ClosureBasedMiddleware: EventMiddleware {
-    let closure: (_ event: Event, _ completion: @escaping (Event?) -> Void) -> Void
+private class ClosureBasedMiddleware: EventMiddleware {
```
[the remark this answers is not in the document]
evsaev: We save the middleware in a property but when we pass it to the notification center it is copied -> when we assign another closure to forward a custom event the copy notification center owns doesn't get updated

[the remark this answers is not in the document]
VojtaStavik: Understood 👍

Sources_v3/WebSocketClient/WebSocketClient_Tests.swift
```diff
@@ -526,8 +526,12 @@ extension WebSocketEngineError: Equatable {
 }
 
 /// A test middleware that can be initiated with a closure/
-private struct ClosureBasedMiddleware: EventMiddleware {
-    let closure: (_ event: Event, _ completion: @escaping (Event?) -> Void) -> Void
+final class EventMiddlewareMock: EventMiddleware {
```
VojtaStavik: Love the new name! 😍

Sources_v3/WebSocketClient/HealthCheckMiddleware_Tests.swift
```diff
@@ -0,0 +1,83 @@
...
+    
+    override func setUp() {
+        super.setUp()
+        
+        webSocketClient = WebSocketClientMock()
+        middleware = HealthCheckMiddleware(webSocketClient: webSocketClient)
+    }
+    
+    override func tearDown() {
+        AssertAsync.canBeReleased(&webSocketClient)
```
VojtaStavik, line 22: 👍

Sources_v3/WebSocketClient/HealthCheckMiddleware_Tests.swift
```diff
@@ -0,0 +1,83 @@
...
+        
+        // Assert event is forwared as it is
+        XCTAssertEqual(forwardedEvent as? TestEvent, event)
+    }
+    
+    func test_middleware_filtersHealthCheckEvents_ifClientIsDeallocated() throws {
+        let event = HealthCheckEvent(connectionId: .unique)
+        
+        // Deallocate the client
+        webSocketClient = nil
```
VojtaStavik, line 45: You can actually use `canBeReleased()` here, too. It's even better because it will fail if there are retain cycles in the WSClient. So you then don't need to check if the client is deallocated later.

Sources_v3/WebSocketClient/WebSocketClient.swift
```diff
@@ -353,3 +353,25 @@ protocol BackgroundTaskScheduler {
...
+        }
+        
+        if let webSocketClient = webSocketClient {
+            webSocketClient.pingController.pongRecieved()
+            if webSocketClient.connectionState.isConnected == false {
+                webSocketClient.connectionState = .connected(connectionId: healthCheckEvent.connectionId)
+            }
+        }
+        
+        // Stop the middleware chain
```
VojtaStavik, line 356: nit: It actually doesn't stop the chain. Maybe something like `// Don't forward the event` is more on spot.

Sources_v3/WebSocketClient/WebSocketClient.swift
```diff
@@ -13,7 +13,7 @@ class WebSocketClient {
     }
     
     /// The notification center `WebSocketClient` uses to send notifications about incoming events.
-    private(set) lazy var notificationCenter: NotificationCenter = environment.createNotificationCenter()
+    let notificationCenter: EventNotificationCenter
```
VojtaStavik, line 16: nit: Can we rename this to `eventNotificationCenter` so it's clear it's not just a random notification center?

Sources_v3/WebSocketClient/WebSocketClient.swift
```diff
@@ -107,17 +104,18 @@ class WebSocketClient {
...
     ) {
         self.environment = environment
         self.connectEndpoint = connectEndpoint
         self.requestEncoder = requestEncoder
         self.sessionConfiguration = sessionConfiguration
-        middlewares = eventMiddlewares
         self.reconnectionStrategy = reconnectionStrategy
         self.eventDecoder = eventDecoder
+        self.notificationCenter = notificationCenter
+        self.notificationCenter.add(middleware: HealthCheckMiddleware(webSocketClient: self))
```
VojtaStavik, line 118: It'd be great to add a test that will verify we add the middleware and the delegate is set.

VojtaStavik (approved):
Very nice! Thanks. Just a couple of small comments and then good to merge 🚢

Sources_v3/WebSocketClient/HealthCheckMiddleware_Tests.swift
evsaev: fixed in c7b85be94bb9bec60b9efe17f7f2c4921730f77b

Sources_v3/WebSocketClient/WebSocketClient.swift
evsaev: fixed in c7b85be94bb9bec60b9efe17f7f2c4921730f77b

Sources_v3/WebSocketClient/WebSocketClient.swift
evsaev: renamed in b043edad1abd90e8ee4e462d7a0bbec9dc176168

Sources_v3/WebSocketClient/WebSocketClient.swift
evsaev: added in d274230163815e0e3c136eaff77d254452d7f2e3

PR #410, opened 2020-08-17
[CIS-266] Update SwiftLint and SwiftFormatter

buh:
- updated code in `v2` by linting
- updated `.swiftformat` file
- added SwiftLint for `v3`
- updated code for `v3` by linting and formatting
- updated `bootstrap.sh` to install a specific version of `swiftformat`

#skip_changelog

PR #430, opened 2020-08-25
Add v3 sample folder to pre-commit formatter

cardoso:


Sample_v3/Samples/SimpleChat/Views/ComposerView.swift
```diff
@@ -31,38 +27,36 @@ class ComposerView: UIView {
...
-        NotificationCenter.default.addObserver(
-            self,
-            selector: #selector(handle(keyboardShowNotification:)),
-            name: UIResponder.keyboardWillShowNotification,
-            object: nil
-        )
+        NotificationCenter.default.addObserver(self,
+                                               selector: #selector(handle(keyboardShowNotification:)),
+                                               name: UIResponder.keyboardWillShowNotification,
+                                               object: nil)
```
VojtaStavik, line 50: @buh , why is this happening? I thought you fixed it in #410 ?

buh: This is a different formatter parameter.

buh: We can fix it here if you want.

buh: I left it like with for arguments, because we use it mostly in this way, but sometimes of course it's not so good. But I think I had to make it optional to use both styles.

VojtaStavik: I see, so we allow this format for function declarations but not for the situation where we call it? I'd suggest using the style as it was originally here as the preferred one.

VojtaStavik:
@cardoso let's wait for #431 to be merged so you can apply the latests changes and then merge.

cardoso:
Assuming this is ready

PR #432, opened 2020-08-26
[CIS-247] Make it possible to reset ephemeral values in the DB

VojtaStavik:
### In this PR

CIS-247 is about implementing user presence (`isOnline`). However, it has already been implemented and properly tested. No work needed there. Instead, I at least introduced a way to reset it on the start, because it's just an ephemeral value.

- We now have the `EphemeralValuesContainer` protocol which DTOs can conform to, in case they need to reset some of their values on the start.
- When `DatabaseContainer` is initialized, it asynchronously iterates over existing entities and if they conform to `EphemeralValuesContainer`, calls `resetEphemeralValues()` on them.

_Note_: Because the reset happens asynchronously (we don't want to delay `DatabaseContainer` init because of that, it possible that the original values will still be there for a small amount of time after the start. However, they should be reset to their default state shortly after.


Sources_v3/Database/DatabaseContainer.swift
```diff
@@ -157,10 +159,27 @@ class DatabaseContainer: NSPersistentContainer {
...
+    /// Iterates over all items and if the DTO conforms to `EphemeralValueContainers` calls `resetEphemeralValues()` on
+    /// every object.
+    func resetEphemeralValues() {
+        write({ session in
+            let session = session as! NSManagedObjectContext
+            
+            try self.managedObjectModel.entities.forEach { entityDescription in
+                guard let entityName = entityDescription.name else { return }
+                let fetchRequest: NSFetchRequest<NSFetchRequestResult> = NSFetchRequest(entityName: entityName)
+                fetchRequest.returnsObjectsAsFaults = true
```
mantergo, line 172: It seems like it's `true` by default.

evsaev (approved):

mantergo (approved):

Sources_v3/Database/DatabaseContainer.swift
VojtaStavik: fixed in 3ff2891

PR #434, opened 2020-08-26
[CIS-250] Message pagination in `ChannelController`

mantergo:
**In this PR:**

- Implement message pagination in `ChannelController`
- Make use of message pagination in `SampleApp`

### DIFF
--- a/Sample_v3/Samples/SimpleChat/DetailViewController.swift
+++ b/Sample_v3/Samples/SimpleChat/DetailViewController.swift
@@ -19,7 +19,7 @@ class DetailViewController: UIViewController {
     
     var channelId: ChannelId! {
         didSet {
-             controller = chatClient.channelController(for: channelId)
+            controller = chatClient.channelController(for: .init(cid: channelId, messagesPagination: [.limit(25)], options: .all))
         }
     }
     
@@ -42,6 +42,7 @@ class DetailViewController: UIViewController {
         super.viewDidLoad()
         
         tableView.dataSource = self
+        tableView.delegate = self
         tableView.translatesAutoresizingMaskIntoConstraints = false
         view.addSubview(tableView)
         NSLayoutConstraint.activate([
@@ -140,7 +141,7 @@ extension DetailViewController: ChannelControllerDelegate {
     }
 }
 
-extension DetailViewController: UITableViewDataSource {
+extension DetailViewController: UITableViewDataSource, UITableViewDelegate {
     func tableView(_ tableView: UITableView, numberOfRowsInSection section: Int) -> Int {
         messages.count
     }
@@ -161,6 +162,13 @@ extension DetailViewController: UITableViewDataSource {
         cell.backgroundColor = message.localState == nil ? .white : .lightGray
         
         return cell
+    }
+    
+    func tableView(_ tableView: UITableView, willDisplay cell: UITableViewCell, forRowAt indexPath: IndexPath) {
+        if indexPath.section == tableView.numberOfSections - 1 &&
+            indexPath.row == tableView.numberOfRows(inSection: indexPath.section) - 1 {
+            controller?.loadNextMessages()
+        }
     }
 }
 
--- a/Sources_v3/Controllers/ChannelController.swift
+++ b/Sources_v3/Controllers/ChannelController.swift
@@ -381,6 +381,64 @@ public extension ChannelControllerGeneric {
         }
     }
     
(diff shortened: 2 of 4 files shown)


Sources_v3/Controllers/ChannelController.swift
```diff
@@ -381,6 +381,66 @@ public extension ChannelControllerGeneric {
...
+            log.error(ClientError.ChannelEmptyMessages().localizedDescription)
+            callback { completion?(ClientError.ChannelEmptyMessages()) }
+            return
+        }
+        
+        channelQuery.messagesPagination = [.limit(limit), .lessThan(messageId)]
+    
+        worker.update(channelQuery: channelQuery) { [weak self] error in
+            guard let self = self else { return }
+            self.callback { completion?(error) }
```
VojtaStavik, line 410: You can just do `self?.callback { completion?(error) }`

Sources_v3/Controllers/ChannelController.swift
```diff
@@ -381,6 +381,66 @@ public extension ChannelControllerGeneric {
...
+        guard isChannelAlreadyCreated else {
+            channelModificationFailed(completion)
+            return
+        }
+        
+        guard let messageId = messageId ?? messages.last?.id else {
+            log.error(ClientError.ChannelEmptyMessages().localizedDescription)
+            callback { completion?(ClientError.ChannelEmptyMessages()) }
+            return
+        }
```
VojtaStavik, line 404: Hmm, this is a question. Maybe we should just retry the original query and only log the error? Something like:
```
log.error("Failed to load next messages because the channel has no messages locally. Trying to reload all messages from the server.")
``` 
The intention of the user is they want more messages, we should try to get them to her/him. WDYT?

Sources_v3/Controllers/ChannelController.swift
```diff
@@ -381,6 +381,66 @@ public extension ChannelControllerGeneric {
...
+        guard isChannelAlreadyCreated else {
+            channelModificationFailed(completion)
+            return
+        }
+        
+        guard let messageId = messageId ?? messages.first?.id else {
+            log.error(ClientError.ChannelEmptyMessages().localizedDescription)
+            callback { completion?(ClientError.ChannelEmptyMessages()) }
+            return
+        }
```
VojtaStavik, line 434: same here

Sources_v3/Controllers/ChannelController_Tests.swift
```diff
@@ -836,6 +836,163 @@ class ChannelController_Tests: StressTestCase {
...
+        XCTAssertFalse(completionCalled)
+        
+        // Simulate successfull udpate
+        env.channelUpdater?.update_completion?(nil)
+        
+        // Completion should be called
+        AssertAsync.willBeTrue(completionCalled)
+        
+        // Assert correct `Pagination` is created
+        XCTAssertEqual(env!.channelUpdater?.update_channelQuery?.messagesPagination, [.limit(25), .lessThan(messageId ?? "")])
```
VojtaStavik, line 882: Just do `messageId!` of `try XCTUnwrap(messageId)`

Sample_v3/Samples/SimpleChat/DetailViewController.swift
```diff
@@ -162,5 +163,12 @@ extension DetailViewController: UITableViewDataSource {
         
         return cell
     }
+    
+    func tableView(_ tableView: UITableView, willDisplay cell: UITableViewCell, forRowAt indexPath: IndexPath) {
+        if indexPath.section == tableView.numberOfSections - 1 &&
+            indexPath.row == tableView.numberOfRows(inSection: indexPath.section) - 1 {
+            controller?.loadNextMessages()
```
VojtaStavik, line 170: @cardoso FYI: You can use the completion of this function and hook it up to some activity indicator.

VojtaStavik (changes requested):
Looks great, thanks!

Sources_v3/Controllers/ChannelController.swift
mantergo: So you are talking about the case when there is actually some messages on backend but fetch failed previously?
I'm not sure cause in this case I believe it's not the job of this function and this situation will be handled during `startUpdating` call that will throw some error.

VojtaStavik: OK, let's leave it as is for now and we'll see.

VojtaStavik (approved):

PR #438, opened 2020-08-26
[CIS-259] Channels pagination

mantergo:
**In this PR:**

- Implement channels pagination in `ChannelListController`
- Make use of pagination in `SampleApp`

_**Q**: I’m not doing any checks for empty `channels` in `loadNextChannels` function cause it’s not an issue to have 0 offset. But there is a possibility that this function will be called successfully before `startUpdating` call. (`channels` has empty array as fallback value) If we want to restrict this we can check for controller `state` but I’m not sure if we should do that._

### DIFF
--- a/Sources_v3/Controllers/ChannelListController_Tests.swift
+++ b/Sources_v3/Controllers/ChannelListController_Tests.swift
@@ -14,6 +14,9 @@ class ChannelListController_Tests: StressTestCase {
     var query: ChannelListQuery!
     
     var controller: ChannelListController!
+    var controllerCallbackQueueID: UUID!
+    /// Workaround for uwrapping **controllerCallbackQueueID!** in each closure that captures it
+    private var callbackQueueID: UUID { controllerCallbackQueueID }
     
     override func setUp() {
         super.setUp()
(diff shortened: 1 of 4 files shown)


Sources_v3/Controllers/ChannelListController_Tests.swift
```diff
@@ -231,6 +236,46 @@ class ChannelListController_Tests: StressTestCase {
...
+    // MARK: - Channels pagination
+    
+    func test_loadNextChannels_callsChannelListQueryUpdater() {
+        var completionCalled = false
+        let limit = 42
+        controller.loadNextChannels(limit: limit) { [callbackQueueID] error in
+            AssertTestQueue(withId: callbackQueueID)
+            XCTAssertNil(error)
+            completionCalled = true
+        }
```
VojtaStavik, line 249: Smells like we should create some abstraction for this boilerplate, too. It's everywhere.

VojtaStavik (approved):
Nice and easy 👍 Thanks

Sources_v3/Controllers/ChannelListController_Tests.swift
mantergo: Probably yes. Will think about it when will touch this again :)

PR #439, opened 2020-08-27
[CIS-304] Bump swift version to 5.2

### LINKED ISSUE #414: Channel Extra Data Codable - map with key path issue
## What did you do?
try to use 2.2.9 and won't compile as there is an issue with a map `extension:`. error of `Type of expression is ambiguous without more context`

`                // Save generated name to nameAndImageForCurrentUser since there isn't any
                nameAndImageForCurrentUser.name = namingStrategy.name(for: User.current,
                                                                      members: members.map(\.user))
`

## GetStream Environment
**GetStream Chat version:**
2.2.9

**iOS version:**
13.2.2

**Swift version:**
5.1.2

**Xcode version:**
11.2.1

gz#5427

b-onc:
Fixes #414

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -72,3 +72,4 @@ vendor/bundle/
 .swiftpm
 Example/Carthage/.env
 Example/Carthage/fastlane/report.xml
+Sample/Cocoapods/Podfile.lock
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### ✅ Added
 - `ChannelsPresenter.delete` & `ChannelsPresenter.rx.delete` methods to delete a channel and remove it from items list [#394](https://github.com/GetStream/stream-chat-swift/pull/394)
 
+### 🔄 Changed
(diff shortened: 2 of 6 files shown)


.gitignore
```diff
@@ -72,3 +72,4 @@ vendor/bundle/
 .swiftpm
 Example/Carthage/.env
 Example/Carthage/fastlane/report.xml
+Sample/Cocoapods/Podfile.lock
```
VojtaStavik, line 75: Why this?
Also, we should probably update the `Example/` paths above to match the new structure, too.

VojtaStavik (approved):
The SPM failure looks like something we should care about:

Screen Shot 2020-08-27 at 14 05 46

.gitignore
b-onc: I've tried to make sure Cocoapods example always builds the podspec swift version, but seems like Xcode treats all swift 5 versions equally and cannot build with old versions. So in my little experiment, `pod install` created the podfile, but IMHO the podfile.lock should never be checked in, since it'd mess with CI and stuff

VojtaStavik: OK, for samples app, we always want the latest versions.

buh (approved):

PR #441, opened 2020-08-27
Add SwiftUI sample stub

cardoso:
No Stream Chat logic in it yet. Just some basic UI elements to start.

#no_changelog

### DIFF
--- a/Sample_v3/Samples/SwiftUISimpleChat/ChatView.swift
+++ b/Sample_v3/Samples/SwiftUISimpleChat/ChatView.swift
@@ -7,9 +7,9 @@ import SwiftUI
 
 @available(iOS 13, *)
 struct ChatView: View {
-    @State  var channelId: String
-    @State  var text: String = ""
-    @State  var messages: [String] = []
+    @State var channelId: String
+    @State var text: String = ""
+    @State var messages: [String] = []
     
     var body: some View {
         VStack {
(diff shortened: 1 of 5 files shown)


Sample_v3/Samples/SwiftUISimpleChat/ChatView.swift
```diff
@@ -0,0 +1,49 @@
...
+import StreamChatClient
+
+@available(iOS 13, *)
+struct ChatView: View {
+    @State
+    var channelId: String
+    @State
+    var text: String = ""
+    @State
+    var messages: [String] = []
```
VojtaStavik, line 17: lol the formatter :D Should be fixed now, right @buh ?

VojtaStavik (approved):
Thank you for such a quick PR!

Sample_v3/Samples/SwiftUISimpleChat/ChatView.swift
buh: This is I explicitly force to write on the same line. I no idea why it's here.
But yes it should be fixed.

cardoso: I ran the formatter manually and pushed a fix. Thought it would do it on pre-commit

VojtaStavik:
@cardoso please squash when merging ;-)

PR #442, opened 2020-08-27
[CIS-305] Network call duration logging

buh:


Sources/Client/Client/ClientLogger.swift
```diff
@@ -227,6 +229,32 @@ public final class ClientLogger {
...
     }
     
+    public func logTaskDuration(_ description: String, task: () -> Void) {
+        let id = logTaskStarted(description)
+        task()
+        logTaskFinished(taskId: id)
+    }
+    
+    @discardableResult
+    public func logTaskStarted(_ description: String, taskId: UUID = .init()) -> UUID {
```
VojtaStavik, line 239: missing documentation

Sources/Client/Client/ClientLogger.swift
```diff
@@ -227,6 +229,32 @@ public final class ClientLogger {
...
+        measuredTasksStart.update {
+            var tasks = $0
+            tasks[taskId] = (start: Date(), description: description)
+            return tasks
+        }
+        return taskId
+    }
+
+    public func logTaskFinished(taskId id: UUID) {
```
VojtaStavik, line 248: missing documentation

Sources/Client/Client/ClientLogger.swift
```diff
@@ -227,6 +229,32 @@ public final class ClientLogger {
...
+        guard let task = measuredTasksStart[id] else {
+            log("Trying to finish a logger measuring task with \(id) but the task wasn't started.", level: .error)
+            return
+        }
+        
+        let duration = Date().timeIntervalSince1970 - task.start.timeIntervalSince1970
+        log("⏱ \(task.description) finished in \(duration) seconds.", level: .debug)
```
VojtaStavik: Should we introduce a special option for this? Something like `.performance`?

Sources/Client/Client/ClientLogger.swift
```diff
@@ -227,6 +229,32 @@ public final class ClientLogger {
         lastTime = startTime
     }
     
+    public func logTaskDuration(_ description: String, task: () -> Void) {
```
VojtaStavik, line 232: This api seems to be dangerous, because you can do `logger?.logTaskDuration( task )` because the task won't be executed if the logger is nil. I'd suggest removing the function completely.

Sources/Client/Model/Channel/Channel.swift
```diff
@@ -75,6 +77,10 @@ public final class Channel: Codable {
     var membership: Member?
     /// A list of channel members.
     public internal(set) var members = Set<Member>()
+    
+    // TODO: Implement updates
+    public let memberCount: Int
```
VojtaStavik, line 82: We need to update this value as the events come

Sources/Client/Model/Channel/Channel.swift
```diff
@@ -165,6 +171,8 @@ public final class Channel: Codable {
         self.team = team
         self.namingStrategy = namingStrategy
         self.config = config
+        // TODO: properly?
+        self.memberCount = members.count
```
VojtaStavik, line 175: this probably should come from the initializer

Sources/Client/Model/Channel/Channel.swift
```diff
@@ -176,6 +184,10 @@ public final class Channel: Codable {
         cid = try container.decode(ChannelId.self, forKey: .cid)
         let members = try container.decodeIfPresent([Member].self, forKey: .members) ?? []
         self.members = Set<Member>(members)
+        
+        // TODO: properly?
+        self.memberCount = try container.decodeIfPresent(Int.self, forKey: .memberCount) ?? 0
+
```
VojtaStavik, line 190: Should we fail the decoding if member count is missing?

Sources/Client/Model/Channel/ChannelsQuery.swift
```diff
@@ -45,6 +51,7 @@ public struct ChannelsQuery: Encodable {
                 sort: [Sorting] = [],
                 pagination: Pagination = [.channelsPageSize],
                 messagesLimit: Pagination = [.messagesPageSize],
+                memberLimit: Int = 100,
```
VojtaStavik: needs to be added to the documentation + exposed up to the chain (rx extension etc.)

Sources/Client/Client/ClientLogger.swift
```diff
@@ -216,7 +216,7 @@ public final class ClientLogger {
     private var startTime: CFTimeInterval
     private let level: Level
     
-    private var measuredTasksStart: Atomic<[UUID: (start: Date, description: String)]> = .init([:])
+    private var measuredTasksStart: Atomic<[UUID: (start: CFTimeInterval, description: String)]> = .init([:])
```
VojtaStavik: why this change?

Sources/Client/Model/Channel/ChannelsQuery.swift
```diff
@@ -19,7 +19,7 @@ public struct ChannelsQuery: Encodable {
         case presence
         case pagination
         case messagesLimit = "message_limit"
-        case memberLimit = "member_limit"
+        case membersLimit = "member_limit"
```
VojtaStavik: why this change when the field is named `"member_limit"`?

VojtaStavik (commented):
Put my comments there for things I think should be done/answered before we release this. + We need to update the changelog.

Sources/Client/Client/ClientLogger.swift
buh: I don't think we should do that. It's not so much noise and actually in early logger version v1 had measurements, but didn't have elegant implementation like this.
I think developers could check JSON output to understand numbers.

Sources/Client/Model/Channel/Channel.swift
buh: It will failed decoding only if it will be presented as like this `member_count: "hmm"`. But I agree to set default value as `members.count` instead of `0`.

Sources/Client/Client/ClientLogger.swift
buh: It's more precisely and easy to manage. Always in seconds.

Sources/Client/Model/Channel/ChannelsQuery.swift
buh: To be consistent with other properties.

VojtaStavik (approved):
LGTM. Please squash when merging

PR #443, opened 2020-08-27
[CIS-307] Fix Device object failing to decode due to backend issue

### LINKED ISSUE #266: Device objects in User.devices have different fields on iOS and Android SDKs
## What did you do?
On iOS app tried to load channels one of which in `members` field have a user created on Android app

Channels presented correctly

## What happened instead?
Decoding error

## GetStream Environment
**GetStream Chat version: 2.2.0**
**iOS version: 13.4**
**Swift version: Swift 5**
**Xcode version: 11.4.1**
**Device: iPhone 11 Pro Simulator**

## Additional context
`Device` struct in StreamChatClient has required fields `id` and `created_at`, where `Device` class in 
stream-chat-android-client has `id` and `push_provider`, which leads to decoding error on iOS

b-onc:
Fixes #266


buh:
This backend issue doesn't relate to devices only. It's in many places.

VojtaStavik (approved):
🚢 🇮🇹

PR #444, opened 2020-08-27
[CIS-246] Implement observer for typing events

mantergo:
**In this PR:** 

- Introduce `TypingEventObserver`
- Make prefix renaming `UserTyping` -> `Typing`

### DIFF
--- a/Sources_v3/ChatClient.swift
+++ b/Sources_v3/ChatClient.swift
@@ -55,7 +55,10 @@ public class Client<ExtraData: ExtraDataTypes> {
             
     /// The notification center used to send and receive notifications about incoming events.
     private(set) lazy var eventNotificationCenter = environment.notificationCenterBuilder([
-        EventDataProcessorMiddleware<ExtraData>(database: databaseContainer)
+        EventDataProcessorMiddleware<ExtraData>(database: databaseContainer),
+        TypingStartCleanupMiddleware<ExtraData>(
+            excludedUserIds: { [weak self] in Set([self?.currentUserId].compactMap { $0 }) }
+        )
     ])
     
     /// The `APIClient` instance `Client` uses to communicate with Stream REST API.
@@ -374,7 +377,7 @@ public class Client<ExtraData: ExtraDataTypes> {
         // Set up a new user id
         currentUserId = userId
         
-        // Set a new WebSocketClient connect endpoint and refresh the middlewares.
+        // Set a new WebSocketClient connect endpoint
         webSocketClient.connectEndpoint = webSocketConnectEndpoint(userId: userId, role: role, extraData: extraData)
         
         // Reset all existing data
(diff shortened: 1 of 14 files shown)


Sources_v3/WebSocketClient/EventMiddlewares/TypingStartCleanupMiddleware.swift
```diff
@@ -5,7 +5,7 @@
 import Foundation
 
 /// Automatically sends a `TypingStop` event if it hasn't come in a specified time after `TypingStart`.
-class UserTypingStartCleanupMiddleware<ExtraData: ExtraDataTypes>: EventMiddleware {
+class TypingStartCleanupMiddleware<ExtraData: ExtraDataTypes>: EventMiddleware {
```
VojtaStavik, line 8: When on it, can you please add this middleware to the middlewares of `EventNotificationCenter`? It seems we have the implementation and tests, but we don't actually use it anywhere :) Be aware that `excludedUserIds` must reflect the current logged in user, which can change.

Sources_v3/WebSocketClient/Events/TypingEvent.swift
```diff
@@ -19,7 +18,7 @@ public struct UserTypingEvent<ExtraData: ExtraDataTypes>: EventWithUserPayload,
         self.payload = payload
     }
     
-    init(from response: EventPayload<ExtraData>) throws {
+    init<ExtraData: ExtraDataTypes>(from response: EventPayload<ExtraData>) throws {
```
VojtaStavik, line 21: Thanks 👍

Sources_v3/Workers/EventObservers/TypingEventObserver.swift
```diff
@@ -0,0 +1,36 @@
...
+}
+
+extension TypingEventObserver {
+    convenience init(notificationCenter: NotificationCenter, callback: @escaping (TypingEvent) -> Void) {
+        self.init(
+            notificationCenter: notificationCenter,
+            filter: { _ in true },
+            callback: callback
+        )
+    }
```
VojtaStavik, line 27: I don't think we will even need this init ;)

VojtaStavik (approved):
Looks good, thanks! Can you please all add an example of the usage to the sample app? Just print the currently typing user to the console or something...

Sources_v3/ChatClient.swift
```diff
@@ -54,9 +54,7 @@ public class Client<ExtraData: ExtraDataTypes> {
     private(set) var backgroundWorkers: [Worker]!
             
     /// The notification center used to send and receive notifications about incoming events.
-    private(set) lazy var eventNotificationCenter = environment.notificationCenterBuilder([
-        EventDataProcessorMiddleware<ExtraData>(database: databaseContainer)
-    ])
+    @Cached private var eventNotificationCenter: EventNotificationCenter
```
VojtaStavik, line 57: We can't do this. The instance must stay the same. Other workers and elements in the system can use it.

VojtaStavik: I think you can make the `excludedUserIds` variable computed or something like that, to make this work without having to recreate the center.

VojtaStavik:
@mantergo Merge please ;-)

PR #446, opened 2020-08-28
[CIS-264] Delegate type erasers auto-generation

evsaev:
This PR adds `Sourcery` tool to the project that automatically generates type-erasers for all protocols with `/// sourcery: TypeErase` annotation. The code-generation is added as a build-phase that goes right before the `Compile Sources` phase.

**TBD**
- do we need generated files to be formated by `swiftformat` (currently they are)
- didn't find the way to freeze `Sourcery` package version
- what else we'd like to have generated automatically

### DIFF
--- a/Sources_v3/.swiftlint.yml
+++ b/Sources_v3/.swiftlint.yml
@@ -60,3 +60,6 @@ cyclomatic_complexity:
   ignores_case_statements: true
   warning: 25
   error: 30
+
+excluded:
+  - Generated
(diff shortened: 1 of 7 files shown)


VojtaStavik:
I don't understand much what's happening but it sure looks like something we really need :) Thanks for exploring this. Let's put a proper task for this to the next sprint 👍

VojtaStavik:
I'm gonna close this for now. We'll re-open next sprint.

PR #356, opened 2020-07-17
[CIS-202] Create Combine wrapper for `ChannelListController`

VojtaStavik:
### In this PR:

This is just exploration work and a proof of concept. However, I think it's worth it to review and merge this, so we have some base for further improvements.

*Changes:*

- `Controller`s have a `state` variable that allows tracking, whether the controller is in the active or inactive state.

- `ChannelListController.Publishers` object wraps `ChannelListController` and exposes its properties via publishers. The main trick here is that the `Publishers` object sets itself as the controller's delegate. It contains a couple of "backing" subjects, that acts as private parent publishers for the publicly exposed ones.

PR #447, opened 2020-08-28
[CIS-291] ObservableObject for controllers (part 1: `ChannelListController`)

VojtaStavik:
This PR adds and `ObservableObject` wrapper for `ChannelListController` to be used in SwiftUI.

#### Changes:

- We use `MulticastDelegate` to make it possible to attach additional delegates for the wrappers, while keeping the underlying controller fully functional. This is a much better experience than originally designed in #356 , where the wrapped controller wasn't accessible. 
_Note_: Refactoring this was a piece of cake thanks to our existing test coverage 😎 

- The `ObservableObject` wrapper is just a simple object that aggregates the changes reported by the delegate.

- Notice the simplicity of the tests. We don't need to test the actual functionality. We just test that the wrapper listens to the delegate callbacks.

### DIFF
--- a/Sources_v3/Controllers/ChannelListController+SwiftUI.swift
+++ b/Sources_v3/Controllers/ChannelListController+SwiftUI.swift
@@ -1,0 +1,55 @@
+//
+// Copyright © 2020 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+import SwiftUI
+
+@available(iOS 13, *)
+extension ChannelListControllerGeneric {
+    /// A wrapper object that exposes the controller variables in the form of `ObservableObject` to be used in SwiftUI.
+    public var observableObject: ObservableObject { .init(controller: self) }
+    
+    /// A wrapper object for `ChannelListController` type which makes it possible to use the controller comfortably in SwiftUI.
+    public class ObservableObject: SwiftUI.ObservableObject {
+        /// The underlying controller. You can still access it and call methods on it.
+        public let controller: ChannelListControllerGeneric
+        
+        /// The channels matching the query.
+        @Published public private(set) var channels: [ChannelModel<ExtraData>] = []
+        
+        /// The current state of the Controller.
+        @Published public private(set) var state: Controller.State
+        
+        /// Creates a new `ObservableObject` wrapper with the provided controller instance.
+        init(controller: ChannelListControllerGeneric<ExtraData>) {
+            self.controller = controller
+            state = controller.state
+            
+            controller.multicastDelegate.additionalDelegates.append(AnyChannelListControllerDelegate(self))
+            
+            if controller.state == .inactive {
+                // Start updating and load the current data
+                controller.startUpdating()
+            }
+            
+            channels = controller.channels
+        }
+    }
+}
+
+@available(iOS 13, *)
+extension ChannelListControllerGeneric.ObservableObject: ChannelListControllerDelegateGeneric {
+    public func controller(
+        _ controller: ChannelListControllerGeneric<ExtraData>,
+        didChangeChannels changes: [ListChange<ChannelModel<ExtraData>>]
+    ) {
+        // We don't care about detailed changes. We just need to update the `channels` property and keep SwiftUI
+        // deal with the rest.
+        channels = controller.channels
+    }
+    
+    public func controller(_ controller: Controller, didChangeState state: Controller.State) {
+        self.state = state
+    }
+}
--- a/Sources_v3/Controllers/ChannelListController.swift
+++ b/Sources_v3/Controllers/ChannelListController.swift
@@ -75,10 +75,6 @@ public class ChannelListControllerGeneric<ExtraData: ExtraDataTypes>: Controller
         return observer
     }()
     
-    /// An wrapper object that exposes the controller variables in the form of `ObservableObject` to be used in SwiftUI.
-    @available(iOS 13, *)
-    public private(set) lazy var observableObject: ObservableObject = ObservableObject(controller: self)
-    
     private let environment: Environment
     
     /// Creates a new `ChannelListController`.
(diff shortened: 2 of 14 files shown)


VojtaStavik:
Sorry for the messier commit history, the pre-commit hook doesn't work properly and it commits unwanted things :/

Sources_v3/Controllers/ChannelListController+SwiftUI.swift
```diff
@@ -0,0 +1,52 @@
...
+
+import Foundation
+import SwiftUI
+
+@available(iOS 13, *)
+extension ChannelListControllerGeneric {
+    /// A wrapper object for `ChannelListController` type which makes it possible to use the controller comfortably in SwiftUI.
+    public class ObservableObject: SwiftUI.ObservableObject {
+        /// The underlying controller. You can still access it and call methods on it.
+        public let controller: ChannelListControllerGeneric
```
evsaev, line 13: We have a retain-cycle here, don't we?

VojtaStavik: 🤔 , we have, however, here it's correct. The observable object have to keep the controller alive. The problem is here https://github.com/GetStream/stream-chat-swift/pull/447/files/c3dc0f4af62feda2e3dedbfb4611fca00135ceb9#diff-0581a70d6a01bca9f6e926b164ff74bfR80 . There it should be `weak` but that would be strange. Uff. I need to think about this.

VojtaStavik: fixed in 33cea36

Sources_v3/Controllers/ChannelListController.swift
```diff
@@ -179,8 +195,8 @@ extension ChannelListControllerGeneric where ExtraData == DefaultDataTypes {
     /// limits of Swift and the way it handles protocols with associated types, it's required to use `setDelegate` method
     /// instead to set the delegate, if you're using custom extra data types.
     public weak var delegate: ChannelListControllerDelegate? {
-        set { anyDelegate = AnyChannelListControllerDelegate(newValue) }
-        get { anyDelegate?.wrappedDelegate as? ChannelListControllerDelegate }
+        set { multicastDelegate.mainDelegate = AnyChannelListControllerDelegate(newValue) }
+        get { multicastDelegate.mainDelegate?.wrappedDelegate as? ChannelListControllerDelegate }
```
buh, line 183: I think you can keep `delegate` as is and use `didSet`. The benefit of that you can remove `mainDelegate` from `MulticastDelegate`.

Sources_v3/Utils/MulticastDelegate.swift
```diff
@@ -0,0 +1,27 @@
...
+// Copyright © 2020 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+/// A wrapper object that allows attaching additional delegates and call the callbacks on all of them.
+///
+/// - Warning: ⚠️ Because `MulticastDelegate` keeps strong references to the delegates, it's strongly recommended to use
+/// an additional wrapper which will make sure the delegate instance is referenced only weakly.
+struct MulticastDelegate<WrappedDelegate> {
```
buh, line 11: You can use internal wrapper for weak references:

```swift
final class WeakRef<T: AnyObject> {
    weak var value: T?
    
    init(_ value: T) {
        self.value = value
    }
}
```

and skip empty values.

evsaev (approved):

Sources_v3/Controllers/ChannelListController.swift
```diff
@@ -49,13 +49,13 @@ public class ChannelListControllerGeneric<ExtraData: ExtraDataTypes>: Controller
         )
 
     /// A type-erased delegate.
-    private(set) var anyDelegate: AnyChannelListControllerDelegate<ExtraData>? {
+    var multicastDelegate: MulticastDelegate<AnyChannelListControllerDelegate<ExtraData>> = .init() {
         didSet {
-            // Set `Controller` `ControllerStateDelegate` delegate
-            stateDelegate = anyDelegate
+            stateMulticastDelegate.mainDelegate = multicastDelegate.mainDelegate
+            stateMulticastDelegate.additionalDelegates = multicastDelegate.additionalDelegates
```
buh, line 55: I'm not sure that it works. So on did set it has empty values for:
`MulticastDelegate<AnyChannelListControllerDelegate<ExtraData>> = .init()`
Probably you have to update it additionally on `setDelegate` and `delegate { set {...} }`.
The same for `ChannelController` and `CurrentUserController`.

buh (approved):

Sources_v3/Controllers/ChannelListController.swift
VojtaStavik: > I think you can keep delegate as is and use didSet. 

What `didSet`? There's no did set in this variable.

> The benefit of that you can remove mainDelegate from MulticastDelegate.

I don't want to remove `mainDelegate`. It's the public-facing delegate. I want to be sure the following is valid:
```swift
controller.delegate = self
assert(controller.delegate === self)
```

Sources_v3/Utils/MulticastDelegate.swift
VojtaStavik: This doesn't work for protocols.
```
protocol DelegateProtocol {}
let weak: WeakRef<DelegateProtocol> // -> compile error, T must be a concrete type
```

Sources_v3/Controllers/ChannelListController.swift
VojtaStavik: Not possible.
`stateMulticastDelegate` is MulticastDelegate<ControllerStateDelegate>`
`multicastDelegate` is `MulticastDelegate<AnyChannelListControllerDelegate<ExtraData>>`
Polymorphysm doesn't work for type generics.

Sources_v3/Controllers/ChannelListController.swift
buh: I thought about this:
```swift
public weak var delegate: ChannelListControllerDelegate? {
    didSet {
       multicastDelegate.addDelegate(delegate)
    }
}
```

VojtaStavik: The `delegate` variable lives in the extension and is accessible only when `ExtraData == DefaultExtraData`. It can't be a stored property.

PR #448, opened 2020-08-31
[CIS-287] Run v3 tests on old ios versions

b-onc:
Investigation complete: caching iOS11 runtime may be possible but it's too time consuming, so having iOS11 tests nightly will have to do it


VojtaStavik (commented):
Should we merge this even though we know the tests are failing?

b-onc:
@VojtaStavik The new iOS12 job is not required for merging, I'd say we merge this and fix the failing tests in another PR.

VojtaStavik (approved):
All right then, 🚢 🇮🇹

PR #450, opened 2020-08-31
Change `User` from struct to class

b-onc:
This seems like the only solution to prevent the Atomic crashes, proved by our customer's reports


VojtaStavik (approved):
I thought we already merged this 🤔

buh (approved):

PR #451, opened 2020-08-31
[CIS-202] Combine wrapper for `ChannelListController`

VojtaStavik:
This PR follows the same pattern we use for the **SwiftUI** wrapper.

Two interesting bits:
- We use the internal `BasePublishers` object which is shared for all exposed publishers. It simplifies the implementation a lot because we don't need to implement custom `Publisher` types, and we just expose them as `AnyPublisher`.

- We have a `keepAlive(_: AnyObject)` helper on the `Publisher` type. It allows "attaching" any object's lifetime to the lifetime of the publisher chain. Without this helper, the following construct won't work, because the result of the `client.channelListController()`call - the actual controller - would get deallocated and wouldn't listen for database updates:

```swift
client.channelListController(query: ...)
   .channelsChangesPublisher
   .map { ... } 
   .assign(to: ...)
   .store(in: &cancellables)
```

### DIFF
--- a/Sample_v3/Samples/SimpleChat/MasterViewController.swift
+++ b/Sample_v3/Samples/SimpleChat/MasterViewController.swift
@@ -2,10 +2,14 @@
 // Copyright © 2020 Stream.io Inc. All rights reserved.
 //
 
+import Combine
 import StreamChatClient
 import UIKit
 
 class MasterViewController: UITableViewController {
+    @available(iOS 13, *)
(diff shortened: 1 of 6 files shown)


VojtaStavik:
@cardoso This will probably be a separate example in the sample app - "CombineSimpleChat"? I added an example to the UIKit version, please move it to the proper location, when you have it :)

evsaev (approved):

PR #452, opened 2020-08-31
[CIS-238] Message actions

evsaev:
**This PR:**
- introduces `MessageController` that can be used to take **edit/delete** actions a specific message
- introduces `MessageEditor` - background worker that listens for messages in a `pendingSync` state and syncs them with the backend
- introduces `MessageUpdater` that acts as an API wrapper for message related endpoints
- adds new `pendingSync/syncing/syncingFailed` local message states that allows `MessageEditor` to do it's job

### DIFF
--- a/Sample_v3/Samples/SimpleChat/SimpleChatViewController.swift
+++ b/Sample_v3/Samples/SimpleChat/SimpleChatViewController.swift
@@ -161,6 +161,58 @@ final class SimpleChatViewController: UITableViewController, ChannelControllerDe
         }
     }
     
+    ///
+    /// # canEditRowAt
+    ///
+    /// The method below returns a bool indicating whether the message can be edited.
+    /// The editing is allowed for non-deleted messages
+    ///
+    override func tableView(_ tableView: UITableView, canEditRowAt indexPath: IndexPath) -> Bool {
+        channelController.messages[indexPath.row].deletedAt == nil
+    }
+    
+    ///
+    /// # willBeginEditingRowAt
+    ///
+    /// In this method we get a swipe container and apply the same transform as the tableView has.
+    /// It fixes the bug when the list is in `listOrdering == .bottomToTop` but the swipe actions are
+    /// upside-down
+    ///
+    override func tableView(_ tableView: UITableView, willBeginEditingRowAt indexPath: IndexPath) {
+        tableView.cellForRow(at: indexPath)?.swipeActionsContainer?.transform = tableView.transform
+    }
+    
+    ///
+    /// # trailingSwipeActionsConfigurationForRowAt
+    ///
+    /// The method below returns a list of swipe actions available on the message
+    ///
+    override func tableView(
+        _ tableView: UITableView,
+        trailingSwipeActionsConfigurationForRowAt indexPath: IndexPath
+    ) -> UISwipeActionsConfiguration? {
+        let message = channelController.messages[indexPath.row]
+        
+        let messageController = chatClient.messageController(
+            cid: channelController.channelQuery.cid,
+            messageId: message.id
+        )
+        
+        let deleteAction = UIContextualAction(style: .destructive, title: "Delete") { _, _, _ in
+            messageController.deleteMessage()
+        }
+        
+        let editAction = UIContextualAction(style: .normal, title: "Edit") { [weak self] _, _, _ in
+            self?.showTextEditingAlert(for: message.text) {
+                messageController.editMessage(text: $0)
+            }
+        }
+        
+        let configuration = UISwipeActionsConfiguration(actions: [deleteAction, editAction])
+        configuration.performsFirstActionWithFullSwipe = true
+        return configuration
+    }
+    
     // MARK: - Button Actions
 
     ///
@@ -241,6 +293,20 @@ final class SimpleChatViewController: UITableViewController, ChannelControllerDe
         composerView.layoutMargins = view.layoutMargins
         composerView.directionalLayoutMargins = systemMinimumLayoutMargins
         return composerView
+    }
+    
+    private func showTextEditingAlert(for text: String, completion: @escaping (_ editedText: String) -> Void) {
+        let alert = UIAlertController(title: "Edit message text", message: nil, preferredStyle: .alert)
+
+        alert.addTextField { textField in
+            textField.text = text
+        }
+
+        alert.addAction(UIAlertAction(title: "Save", style: .default, handler: { [weak alert] _ in
+            completion(alert?.textFields?.first?.text ?? "")
+        }))
+
+        present(alert, animated: true)
     }
 }
 
@@ -332,5 +398,20 @@ extension SimpleChatViewController {
         tableView.contentInsetAdjustmentBehavior = .never
         tableView.separatorColor = .clear
         tableView.transform = CGAffineTransform(scaleX: 1, y: -1)
+    }
+}
+
+// MARK: - Private
+
+private extension UITableViewCell {
+    var swipeActionsContainer: UIView? {
+        guard
+            let superview = superview,
+            let swipeContainerViewType = NSClassFromString("_UITableViewCellSwipeContainerView"),
+            let swipeViewType = NSClassFromString("UISwipeActionPullView"),
+            superview.isKind(of: swipeContainerViewType)
+        else { return nil }
+        
+        return superview.subviews.first { $0.isKind(of: swipeViewType) }
     }
 }
--- a/Sources_v3/APIClient/Endpoints/MessageEndpoints.swift
+++ b/Sources_v3/APIClient/Endpoints/MessageEndpoints.swift
@@ -1,0 +1,44 @@
+//
+// Copyright © 2020 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+extension Endpoint {
+    static func getMessage<ExtraData: ExtraDataTypes>(messageId: MessageId) -> Endpoint<MessagePayload<ExtraData>> {
+        .init(
+            path: messageId.path,
+            method: .get,
+            queryItems: nil,
+            requiresConnectionId: false,
+            body: nil
+        )
+    }
+    
+    static func deleteMessage(messageId: MessageId) -> Endpoint<EmptyResponse> {
+        .init(
+            path: messageId.path,
+            method: .delete,
+            queryItems: nil,
+            requiresConnectionId: false,
+            body: nil
+        )
+    }
+    
+    static func editMessage<ExtraData: ExtraDataTypes>(payload: MessageRequestBody<ExtraData>)
+        -> Endpoint<EmptyResponse> {
+        .init(
+            path: payload.id.path,
+            method: .post,
+            queryItems: nil,
+            requiresConnectionId: false,
+            body: ["message": payload]
+        )
+    }
+}
+
+private extension MessageId {
+    var path: String {
+        "messages/\(self)"
+    }
+}
--- a/Sources_v3/ChatClient.swift
+++ b/Sources_v3/ChatClient.swift
@@ -174,7 +174,8 @@ public class Client<ExtraData: ExtraDataTypes> {
         let workerBuilders: [WorkerBuilder] = [
             MessageSender<ExtraData>.init,
             NewChannelQueryUpdater<ExtraData>.init,
-            ChannelWatchStateUpdater<ExtraData>.init
+            ChannelWatchStateUpdater<ExtraData>.init,
+            MessageEditor<ExtraData>.init
         ]
         
         self.init(
--- a/Sources_v3/Database/DTOs/MessageDTO.swift
+++ b/Sources_v3/Database/DTOs/MessageDTO.swift
@@ -171,6 +171,10 @@ extension NSManagedObjectContext: MessageDatabaseSession {
     }
     
     func message(id: MessageId) -> MessageDTO? { .load(id: id, context: self) }
+    
+    func delete(message: MessageDTO) {
+        delete(message)
+    }
 }
 
 extension MessageDTO {
--- a/Sources_v3/Database/DTOs/MessageDTO.swift
+++ b/Sources_v3/Database/DTOs/MessageDTO.swift
@@ -58,11 +58,27 @@ class MessageDTO: NSManagedObject {
         return request
     }
     
+    /// Returns a fetch request for messages pending sync.
+    static func messagesPendingSyncFetchRequest() -> NSFetchRequest<MessageDTO> {
+        let request = NSFetchRequest<MessageDTO>(entityName: MessageDTO.entityName)
+        request.sortDescriptors = [NSSortDescriptor(keyPath: \MessageDTO.locallyCreatedAt, ascending: true)]
+        request.predicate = NSPredicate(format: "localMessageStateRaw == %@", LocalMessageState.pendingSync.rawValue)
+        return request
+    }
+    
     /// Returns a fetch request for messages from the channel with the provided `cid`.
     static func messagesFetchRequest(for cid: ChannelId, sortAscending: Bool = false) -> NSFetchRequest<MessageDTO> {
         let request = NSFetchRequest<MessageDTO>(entityName: MessageDTO.entityName)
         request.sortDescriptors = [NSSortDescriptor(keyPath: \MessageDTO.defaultSortingKey, ascending: sortAscending)]
         request.predicate = NSPredicate(format: "channel.cid == %@", cid.rawValue)
+        return request
+    }
+    
+    /// Returns a fetch request for the dto with a specific `messageId`.
+    static func message(withID messageId: MessageId) -> NSFetchRequest<MessageDTO> {
+        let request = NSFetchRequest<MessageDTO>(entityName: MessageDTO.entityName)
+        request.sortDescriptors = [NSSortDescriptor(keyPath: \MessageDTO.defaultSortingKey, ascending: false)]
+        request.predicate = NSPredicate(format: "id == %@", messageId)
         return request
     }
     
--- a/Sources_v3/Database/DatabaseSession.swift
+++ b/Sources_v3/Database/DatabaseSession.swift
@@ -53,6 +53,10 @@ protocol MessageDatabaseSession {
     
     /// Fetchtes `MessageDTO` with the given `id` from the DB. Returns `nil` if no `MessageDTO` matching the `id` exists.
     func message(id: MessageId) -> MessageDTO?
+    
+    /// Deletes the provided dto from a database
+    /// - Parameter message: The DTO to be deleted
+    func delete(message: MessageDTO)
 }
 
 extension MessageDatabaseSession {
--- a/Sources_v3/Models/Message.swift
+++ b/Sources_v3/Models/Message.swift
@@ -33,6 +33,13 @@ public enum MessageType: String, Codable {
 
 /// A possible additional local state of the message. Applies only for the messages of the current user.
 public enum LocalMessageState: String {
+    /// The message is waiting to be synced.
+    case pendingSync
+    /// The message is currently being synced
+    case syncing
+    /// Syncing of the message failed after multiple of tries. The system is not trying to sync this message anymore.
+    case syncingFailed
+    
     /// The message is waiting to be sent.
     case pendingSend
     /// The message is currently being sent to the servers.
--- a/Sources_v3/Workers/Background/MessageEditor.swift
+++ b/Sources_v3/Workers/Background/MessageEditor.swift
@@ -1,0 +1,122 @@
+//
+// Copyright © 2020 Stream.io Inc. All rights reserved.
+//
+
+import CoreData
+import Foundation
+
+/// Observers the storage for messages in a `pendingSync` state and updates them on the backend.
+///
+/// Sending of the message has the following phases:
+///     1. When a message with `pendingSync` state local state appears in the db, the editor enques it in the sending queue.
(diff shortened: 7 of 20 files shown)


Sources_v3/APIClient/Endpoints/MessageEndpoints.swift
```diff
@@ -0,0 +1,27 @@
...
+    
+    static func editMessage(userId: UserId, messageId: MessageId, text: String) -> Endpoint<EmptyResponse> {
+        .init(
+            path: "messages/\(messageId)",
+            method: .post,
+            queryItems: nil,
+            requiresConnectionId: true,
+            body: EditMessageRequestBody(clientId: userId, text: text)
+        )
+    }
```
VojtaStavik, line 26: Do we have to use this endpoint then? Did you try it with the message-send endpoint?

Sources_v3/APIClient/Endpoints/Requests/EditMessageRequestBody.swift
```diff
@@ -0,0 +1,22 @@
...
+    let clientId: UserId
+    let text: String
+    
+    func encode(to encoder: Encoder) throws {
+        var container = encoder.container(keyedBy: CodingKeys.self)
+
+        try container.encode(clientId, forKey: .clientId)
+        try container.encode(["text": text], forKey: .message)
+    }
+}
```
VojtaStavik, line 22: Do we need this? Maybe we could be able to use `MessageRequestBody`, the structure looks similar.

Sources_v3/Controllers/MessageControllerGeneric.swift
```diff
@@ -0,0 +1,201 @@
+//
+// Copyright © 2020 Stream.io Inc. All rights reserved.
+//
+
+import CoreData
+import Foundation
+
+public extension Client {
+    func messageController(messageId: MessageId) -> MessageControllerGeneric<ExtraData> {
+        .init(client: self, messageId: messageId, environment: .init())
```
VojtaStavik, line 10: nit: Environment should always have a default value. We change it only in tests, ideally we don't touch it in production code at all.

Sources_v3/Controllers/MessageControllerGeneric.swift
```diff
@@ -0,0 +1,201 @@
...
+        self.client = client
+        self.messageId = messageId
+        self.environment = environment
+    }
+    
+    public func startUpdating(_ completion: ((Error?) -> Void)? = nil) {
+        do {
+            try messageObserver.startObserving()
+        } catch {
+            callback { completion?(ClientError.FetchFailed()) }
```
VojtaStavik, line 54: Please put the error we catch here as the underlying error of `ClientError.FetchFailed`

Sources_v3/Controllers/MessageControllerGeneric.swift
```diff
@@ -0,0 +1,201 @@
...
+            return
+        }
+        
+        state = .localDataFetched
+
```
VojtaStavik, line 59: We need to make an API call here to fetch the message. It's possible the message doesn't exist in the database yet.

Sources_v3/Controllers/MessageControllerGeneric.swift
```diff
@@ -0,0 +1,201 @@
...
+// MARK: - Actions
+
+public extension MessageControllerGeneric {
+    func editMessage(text: String, completion: ((Error?) -> Void)? = nil) {
+        messageUpdater.editMessage(messageId: messageId, text: text) { [weak self] error in
+            self?.callback {
+                completion?(error)
+            }
+        }
+    }
```
VojtaStavik, line 73: This will need to a little bit smarter. What if you edit a message that hasn't been sent yet? It should just change the text locally and let the message be sent.

This logic should probably happen in the updater, not in the controller, of course.

Sources_v3/Controllers/MessageControllerGeneric.swift
```diff
@@ -0,0 +1,201 @@
...
+    
+    func delete(completion: ((Error?) -> Void)? = nil) {
+        messageUpdater.deleteMessage(messageId: messageId) { [weak self] error in
+            self?.callback {
+                completion?(error)
+            }
+        }
+    }
```
VojtaStavik, line 81: Same here, what if you delete a message that hasn't been sent yet? We should probably just delete it from the DB for good.

Sources_v3/Models/Message.swift
```diff
@@ -33,6 +33,8 @@ public enum MessageType: String, Codable {
 
 /// A possible additional local state of the message. Applies only for the messages of the current user.
 public enum LocalMessageState: String {
+    /// The message is waiting to be edited.
+    case pendingEdit
     /// The message is waiting to be sent.
     case pendingSend
     /// The message is currently being sent to the servers.
```
VojtaStavik, line 40: Do we really need this? Can this be just `pendingSync` for both cases?

VojtaStavik (commented):
Yes, I think the direction is correct. Let's talk more about possible simplifications.

Sources_v3/APIClient/Endpoints/MessageEndpoints.swift
evsaev: Yes, that was the very first thing I tried. When we use the send-message endpoint for editing, we get the error saying that the message with the provided `id` already exist

Sources_v3/Controllers/MessageControllerGeneric.swift
```diff
@@ -0,0 +1,206 @@
+//
+// Copyright © 2020 Stream.io Inc. All rights reserved.
+//
+
+import CoreData
+import Foundation
+
+public extension Client {
+    func messageController(channelId: ChannelId, messageId: MessageId) -> MessageControllerGeneric<ExtraData> {
```
VojtaStavik, line 9: nit: `channelId` -> `cid`

Sources_v3/APIClient/Endpoints/MessageEndpoints.swift
```diff
@@ -0,0 +1,44 @@
...
+            body: ["message": payload]
+        )
+    }
+}
+
+private extension MessageId {
+    var path: String {
+        "messages/\(self)"
+    }
+}
```
VojtaStavik, line 44: Yeah, we need to figure out how to do compose the paths in some reasonable way...

Sources_v3/ChatClient.swift
```diff
@@ -166,6 +166,7 @@ public class Client<ExtraData: ExtraDataTypes> {
         // All production workers
         let workerBuilders: [WorkerBuilder] = [
             MessageSender<ExtraData>.init,
+            MessageEditor<ExtraData>.init,
```
VojtaStavik, line 169: I don't know, the name doesn't work for me, but I don't have any other idea than `MessageChangesUpdater` and I don't like it either.

Sources_v3/Controllers/MessageControllerGeneric.swift
```diff
@@ -0,0 +1,206 @@
...
+    
+    public var message: MessageModel<ExtraData>? {
+        state != .inactive ? messageObserver.item : nil
+    }
+    
+    private let environment: Environment
+    
+    private(set) var anyDelegate: AnyMessageControllerDelegate<ExtraData>? {
+        didSet { stateDelegate = anyDelegate }
+    }
```
VojtaStavik, line 29: Rebase, this has changed recently a lot

Sources_v3/Controllers/MessageControllerGeneric.swift
```diff
@@ -0,0 +1,206 @@
...
+            self.state = error == nil ? .remoteDataFetched : .remoteDataFetchFailed(ClientError(with: error))
+            self.callback { completion?(error) }
+        }
+    }
+}
+
+// MARK: - Actions
+
+public extension MessageControllerGeneric {
+    func editMessage(text: String, completion: ((Error?) -> Void)? = nil) {
```
VojtaStavik, line 72: It either should be `edit(` or you should change `delete(` to `deleteMessage(` ... I'll leave that up to you, I don't have preferences.

Sources_v3/Database/DTOs/MessageDTO.swift
```diff
@@ -58,6 +58,14 @@ class MessageDTO: NSManagedObject {
         return request
     }
     
+    /// Returns a fetch request for messages pending edit.
+    static func messagesPendingEditFetchRequest() -> NSFetchRequest<MessageDTO> {
```
VojtaStavik, line 62: I'm not sure about `pendingEdit` wording. It's not the message "waiting to be edited" as the name suggests. It's messages that have been edited and they are waiting to be synced. So maybe, `pendingSync`? Same for the local state.

Sources_v3/Database/DatabaseSession.swift
```diff
@@ -95,6 +95,8 @@ protocol DatabaseSession: UserDatabaseSession, CurrentUserDatabaseSession, Messa
     ) throws -> ChannelDTO
     
     func loadChannel<ExtraData: ExtraDataTypes>(cid: ChannelId) -> ChannelModel<ExtraData>?
+    
+    func delete(_ dto: NSManagedObject)
```
VojtaStavik, line 99: Can we please make this a DTO specific. For example: `delete(_ message: MessageDTO)` and move it to `MessageDatabaseSession`? The goal is to have `DatabaseSession` empty and have it acting only as an umbrella protocol for the concrete session protocols.

Sources_v3/Models/Message.swift
```diff
@@ -33,6 +33,13 @@ public enum MessageType: String, Codable {
 
 /// A possible additional local state of the message. Applies only for the messages of the current user.
 public enum LocalMessageState: String {
+    /// The message is waiting to be edit.
+    case pendingEdit
+    /// The message is currently being sent to the servers for an edit.
+    case editing
+    /// Editing of the message failed after multiple of tries. The system is not trying to edit this message anymore.
+    case editingFailed
+
```
VojtaStavik, line 42: I'd suggest:
```
pendingSync, syncing, syncingFailed
```

Sources_v3/Workers/Background/MessageEditor.swift
```diff
@@ -0,0 +1,108 @@
...
+        
+        super.init(database: database, webSocketClient: webSocketClient, apiClient: apiClient)
+        
+        startObserving()
+    }
+}
+
+// MARK: - Private
+
+private extension MessageEditor {
```
VojtaStavik, line 33: Why have this in an extension?

Sources_v3/Workers/Background/MessageEditor.swift
```diff
@@ -0,0 +1,108 @@
...
+            }
+        }
+    }
+    
+    func removeMessageIDAndContinue(_ messageId: MessageId) {
+        pendingMessageIDs.remove(messageId)
+        sendingQueue.async { self.processNextMessage() }
+    }
+    
+    func markMessage(withID id: MessageId, as state: LocalMessageState?, completion: @escaping () -> Void) {
```
VojtaStavik, line 88: nice 👍

Sources_v3/Workers/Background/MessageEditor.swift
```diff
@@ -0,0 +1,108 @@
...
+        database.write({
+            $0.message(id: id)?.localMessageState = state
+        }, completion: { _ in
+            completion()
+        })
```
VojtaStavik, line 93: please log the error

Sources_v3/Workers/MessageUpdater.swift
```diff
@@ -0,0 +1,109 @@
...
+        }
+    }
+    
+    func deleteMessage(messageId: MessageId, completion: ((Error?) -> Void)? = nil) {
+        var shouldDeleteOnBackend = true
+        
+        database.write({ session in
+            guard
+                let dto = session.message(id: messageId), [.pendingSend, .sendingFailed].contains(dto.localMessageState)
+            else { return }
```
VojtaStavik, line 31: This is a very critical piece of code. Can we make it a tiny bit more expressive? If I understand it correctly, what is happening here is: 
```
if <message exists only locally> {
  delete it and set shouldDeleteOnBackend to false
}
```
I'd like the actual code to read the same as the pseudo-code above :)

Sources_v3/Workers/MessageUpdater.swift
```diff
@@ -0,0 +1,109 @@
...
+            session.delete(dto)
+            shouldDeleteOnBackend = false
+        }, completion: { error in
+            guard shouldDeleteOnBackend, error == nil else {
+                completion?(error)
+                return
+            }
+            
+            self.database.write({ session in
+                session.message(id: messageId)?.localMessageState = .deleting
```
VojtaStavik, line 42: I think you can do this already in the session above, you know the state will change to `deleting` if the message exists on the backend.

Sources_v3/Workers/MessageUpdater.swift
```diff
@@ -0,0 +1,109 @@
...
+            }
+        }
+    }
+    
+    func deleteMessage(messageId: MessageId, completion: ((Error?) -> Void)? = nil) {
+        var shouldDeleteOnBackend = true
+        
+        database.write({ session in
+            guard
+                let dto = session.message(id: messageId), [.pendingSend, .sendingFailed].contains(dto.localMessageState)
```
VojtaStavik, line 30: Should we also throw `ClientError.MessageDoesNotExist` when the message doesn't exist?

VojtaStavik (commented):
Happy with the progress

Sources_v3/Controllers/MessageControllerGeneric.swift
evsaev: Now it says `deleteMessage`

Sources_v3/Database/DatabaseSession.swift
evsaev: Changed to `delete(message: MessageDTO)`. The parameter name is there to let us call `delete(_ object: NSManagedObject)` from the inside of this method

Sources_v3/Workers/Background/MessageEditor.swift
evsaev: When we declare methods in `private extensions` we don't need to mark each one as `private`

Sources_v3/Workers/MessageUpdater.swift
evsaev: I don't think so. We can have a message that exists on the backend but hasn't been fetched yet (we discussed that we don't require `startUpdating` to be called before accessing `deleteMessage`)

Sources_v3/Workers/MessageUpdater_Tests.swift
```diff
@@ -0,0 +1,559 @@
...
+        var completionCalled = false
+        messageUpdater.editMessage(messageId: createdMessageId, text: updatedText) {
+            XCTAssertNil($0)
+            completionCalled = true
+        }
+        
+        // Load created message
+        var message: MessageDTO? {
+            database.viewContext.message(id: createdMessageId)
+        }
```
VojtaStavik, line 110: I realized this is an unnecessary pattern when working with DTOs. We used to have it like this for the models. For DTOs you can simply do:
```swift
let message = database.viewContext.message(id: createdMessageId)
```

only for models it's required to do the following, since models are immutable:
```swift
var messageModel: Message? { database.viewContext.message(id: createdMessageId)?.asModel() }
```

Sources_v3/Workers/MessageUpdater_Tests.swift
```diff
@@ -0,0 +1,559 @@
...
+        
+        AssertAsync {
+            // Assert completion is called
+            Assert.willBeTrue(completionCalled)
+            // Assert message still has `pendingSync` state
+            Assert.willBeEqual(message?.localMessageState, .pendingSync)
+            // Assert message text is updated correctly
+            Assert.willBeEqual(message?.text, updatedText)
+        }
```
VojtaStavik, line 150: I think this can be simplified to:
```swift
        let editError = try await {
            messageUpdater.editMessage(messageId: createdMessageId, text: updatedText, completion: $0)
        }
        XCTAssertNil(editError)

        let message = database.viewContext.message(id: createdMessageId)
        
        XCTAssertEqual(message?.localMessageState, .pendingSync)
        XCTAssertEqual(message?.text, updatedText)
```

... same of the rest of the tests

Sources_v3/Workers/MessageUpdater_Tests.swift
```diff
@@ -0,0 +1,559 @@
...
+        try assert_editMessage_throwsTheError_and_doesNotChangeTheMessage(localState: .sendingFailed)
+    }
+    
+    func test_editMessage_propogatesError_ifLocalStateIsSyncing() throws {
+        try assert_editMessage_throwsTheError_and_doesNotChangeTheMessage(localState: .syncing)
+    }
+    
+    func test_editMessage_propogatesError_ifLocalStateIsSyncingFailed() throws {
+        try assert_editMessage_throwsTheError_and_doesNotChangeTheMessage(localState: .syncingFailed)
+    }
```
VojtaStavik, line 206: Do you think we need this? Can just make one test like `test_editMessage_returnsError_forUnsupportedLocalStates` and just loop over the unsupported states there? Having this seems like an unnecesary boilerplate.

Sources_v3/Workers/MessageUpdater_Tests.swift
```diff
@@ -0,0 +1,559 @@
...
+            // Assert completion is called
+            Assert.willBeTrue(completionCalled)
+            // Assert message still has `pendingSync` state
+            Assert.willBeEqual(message?.localMessageState, .pendingSync)
+            // Assert message text is updated correctly
+            Assert.willBeEqual(message?.text, updatedText)
+        }
+    }
+    
+    func test_editMessage_updatesMessageLocalStateCorrectly_ifLocalStateIsPendingSend() throws {
```
VojtaStavik, line 153: `test_editMessage_updatesMessageLocalStateCorrectly_ifLocalStateIsPendingSend`
`test_editMessage_updatesLocalMessageCorrectly_ifLocalStateIsPendingSync `
`test_editMessage_updatesLocalMessageCorrectly_ifThereIsNoLocalState`

Are almost identical tests, the only thing that changes is the local state condition. Can we make a single test for it and just loop over various local states?

Sources_v3/Workers/Background/MessageEditor.swift
```diff
@@ -0,0 +1,122 @@
...
+            removeMessageIDAndContinue(messageId)
+            return
+        }
+        
+        guard dto.localMessageState == .pendingSync else {
+            removeMessageIDAndContinue(messageId)
+            return
+        }
+        
+        let requestBody = dto.asRequestBody() as MessageRequestBody<ExtraData>
```
VojtaStavik, line 77: I think it safer to create the request body _after_ you mark the message as `syncing`. You always want to be sure you have the latest version of the object.

VojtaStavik (commented):
Looks like you had a lot of fun :)

Sources_v3/Workers/MessageUpdater_Tests.swift
evsaev: I can imagine one case where this pattern can be useful: when we are waiting for a DTO to be created and check it asynchronously in tests. But that's not the case here 😄

Sources_v3/Workers/MessageUpdater.swift
```diff
@@ -0,0 +1,153 @@
...
+}
+
+extension ClientError {
+    class MessageDoesNotExist: ClientError {
+        init(messageId: MessageId) {
+            super.init("There is no `MessageDTO` instance in the DB matching id: \(messageId).")
+        }
+    }
+    
+    class MessageCanNotBeUpdatedByCurrentUser: ClientError {
```
VojtaStavik, line 136: nit: `Cannot` is one word :)

Sample_v3/Samples/SimpleChat/SimpleChatViewController.swift
```diff
@@ -334,3 +400,18 @@ extension SimpleChatViewController {
...
+        guard
+            let superview = superview,
+            let swipeContainerViewType = NSClassFromString("_UITableViewCellSwipeContainerView"),
+            let swipeViewType = NSClassFromString("UISwipeActionPullView"),
+            superview.isKind(of: swipeContainerViewType)
+        else { return nil }
+        
+        return superview.subviews.first { $0.isKind(of: swipeViewType) }
+    }
+}
```
VojtaStavik, line 417: What kind of sorcery is this? :D

VojtaStavik (approved):
The amount of code here is very difficult to review properly but it looks great at first glance! Great work, thank you Pavel 👍

Sample_v3/Samples/SimpleChat/SimpleChatViewController.swift
evsaev: Ahaha 😄 That's the only solution I found (with some creativity to make it look nicer) that deals with the following issue:
 
Simulator Screen Shot - iPhone 8 - 2020-09-07 at 12 16 54

evsaev: Hmm.. 🤔  I might be wrong, but seems it's not enough to change the message ordering

If we don't mirror the table view the `load more` logic has be adjusted +without flipping the table view we also have to auto-scroll the table to the last cell so that users can see the latest messages first

VojtaStavik: I'm not saying it's the only required change, but it's the change that allows us to make this without any hacks.

PR #453, opened 2020-09-01
Improve v3 UIKit sample chat UI

cardoso:
This some improvements to the UIKit sample

- Inverts the UITableView
- Adds keyboard avoidance behavior
- IRC styled message cell
- Start on Channels Screen

Preview:
Screen Shot 2020-08-31 at 21 54 19

### DIFF
--- a/Sample_v3/Samples/SimpleChat/SimpleChatViewController.swift
+++ b/Sample_v3/Samples/SimpleChat/SimpleChatViewController.swift
@@ -5,9 +5,7 @@
 import StreamChatClient
 import UIKit
 
-class DetailViewController: UIViewController {
-    private let tableView = UITableView()
-    
+class SimpleChatViewController: UITableViewController {
     private var controller: ChannelController?
     
     var messages: [Message] { controller?.messages ?? [] }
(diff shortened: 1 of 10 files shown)


Sample_v3/Samples/SimpleChat/SimpleChatViewController.swift
```diff
@@ -140,33 +144,70 @@ extension DetailViewController: ChannelControllerDelegate {
...
     }
 }
+
+// MARK: - TableView
+
+extension SimpleChatViewController {
+    func setupTableView() {
+        tableView.contentInsetAdjustmentBehavior = .never
+        tableView.separatorColor = .clear
+        tableView.transform = CGAffineTransform(scaleX: 1, y: -1)
```
VojtaStavik, line 211: Is this the preferred way of doing this? Maybe we should rather make the message ordering configurable, no?

cardoso: This is a common way to do it. And the way the ChannelController outputs the updates, I'd have to transform the indices everywhere they're used: (messages.count - index - 1 & indexPath = .init(row: messages.count - indexPath.row - 1 ) and so on.

I think either the ChannelController should have a reversed option or we do it this way to avoid doing index transformation everywhere.

Wdyt?

VojtaStavik: Yeah, we need to expose the ordering option. Let us know next time about if there are strange workarounds you have to do. It's a clear sign we haven't thought about something.

VojtaStavik: I created CIS-314 covering this. We can remove this workaround once that one is implemented.

VojtaStavik (approved):

PR #454, opened 2020-09-02
[CIS-314] Make it possible to change message ordering in `ChannelController`

VojtaStavik:
This is a local-only problem and has no effect on the API calls.

- We have a new `MessageSorting` enum
- `ChannelController` respects this value when fetching messages

---

@cardoso , this should allow you to do
```swfit 
controller.messageSorting = .bottomToTop
```
in the sample app, and remove the "transform" workarounds.


Sources_v3/Controllers/ChannelController.swift
```diff
@@ -78,6 +78,15 @@ extension Client {
 /// A convenience typealias for `ChannelControllerGeneric` with `DefaultDataTypes`
 public typealias ChannelController = ChannelControllerGeneric<DefaultDataTypes>
 
+/// Describes the flow of the messages in the channel.
+public enum MessageSorting {
+    /// New messages appear on the top of the list.
+    case topToBottom
```
b-onc, line 84: Why not a simple `ascending` and `descending`, and we can use the enum everywhere:
```swift
public enum Sorting: Int {
  case ascending = -1, descending = 1
}
```
I have the feeling `topToBottom` and `bottomToTop` is encouraging laziness 😛

Sources_v3/Database/DatabaseContainer_Mock.swift
```diff
@@ -60,9 +60,13 @@ extension DatabaseContainer {
     }
     
     /// Synchrnously creates a new ChannelDTO in the DB with the given cid.
-    func createChannel(cid: ChannelId = .unique) throws {
+    func createChannel(cid: ChannelId = .unique, includeMessages: Bool = true) throws {
```
b-onc, line 63: `createChannel(includeMessage: true)` removes the channel's messages? A bit confusing, I suggest renaming param to `withMessages` so `if !withMessages { removeAllMessages }`

b-onc (approved):
2 minor confusions, otherwise LGTM

Sources_v3/Controllers/ChannelController.swift
VojtaStavik: Laziness? Why? I didn't use `ascending`/`descending` because it can be misleading. Ascending can be understood as "from the oldest message to the latest one" which is not true. We always show latests messages.

I think the naming works good because you don't have to think about it :)

Sources_v3/Database/DatabaseContainer_Mock.swift
VojtaStavik: `createChannel(includeMessage: true)` keeps the messages, it's the default. The condition is `if !includeMessages { }`. Maybe you didn't notice the `!`?

Sources_v3/Controllers/ChannelController.swift
b-onc: There's a certain brain exercise involved in thinking "messages sorted by their date, in ascending order... does that put latest messages on bottom or top?", that's what I meant by "encouraging laziness" :D
I don't think it'd be misleading or vague, I think the confusion you meant is still present: "topToBottom means I'll get older messages?" so a general solution (ascending/descending) would be more useful.

Sources_v3/Database/DatabaseContainer_Mock.swift
b-onc: I see.. `createChannel(includeMessages: true)` as default will create a mock channel **with messages**, and `includeMessages: false` will return a channel without messages. My suggestion still stands, `includeMessages` is a bit confusing, where it's not clear _which_ messages will be included (to me)

Sources_v3/Controllers/ChannelController.swift
VojtaStavik: OK, what if we renamed `MessageSorting` -> `MessageFlow` ?

VojtaStavik: I want to avoid calling it `sorting` because it's not about sorting.

Sources_v3/Database/DatabaseContainer_Mock.swift
VojtaStavik: Renamed to `withMessages` in 5fa87ff

PR #455, opened 2020-09-02
[CIS-317] Give users access to any object by its id

VojtaStavik:
We should allow users to load any arbitrary object from the DB based on its id.

I solved this by introducing a `DataStoreProvider` protocol. Every object conforming to this protocol automatically exposes a `dataStore` variable, that acts like a proxy to the db.

Currently, only controllers expose `dataStore`, but any other object having access to `Client` (like the client itself) can adopt this protocol, too.

---

@cardoso This should fix the issue you have with not having access to the user model from typing events.

### DIFF
--- a/Sources_v3/Database/DatabaseContainer_Mock.swift
+++ b/Sources_v3/Database/DatabaseContainer_Mock.swift
@@ -47,8 +47,15 @@ extension DatabaseContainer {
             throw error
         }
     }
-    
-    /// Synchrnously creates a new CurrentUserDTO in the DB with the given id.
+
+    /// Synchronously creates a new UserDTO in the DB with the given id.
+    func createUser(id: UserId = .unique) throws {
+        try writeSynchronously { session in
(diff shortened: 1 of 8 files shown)


Sources_v3/Database/DatabaseContainer_Mock.swift
```diff
@@ -47,8 +47,15 @@ extension DatabaseContainer {
             throw error
         }
     }
-    
-    /// Synchrnously creates a new CurrentUserDTO in the DB with the given id.
+
+    /// Synchronously creates a new CurrentUserDTO in the DB with the given id.
```
b-onc, line 51:
```suggestion
    /// Synchronously creates a new UserDTO in the DB with the given id.
```

b-onc (approved):

Sources_v3/Database/DatabaseContainer_Mock.swift
VojtaStavik: fixed in b1c3d7d

PR #456, opened 2020-09-02
[CIS-316] Connection events

VojtaStavik:
This is a blocker for CIS-249 and CIS-315.

- I introduced a new enum `ChatClientConnectionStatus`. It's a simplified version of the web socket's `ConnectionState`. We use it to hide implementation details and also to minimize the public API surface area.

- I added `TESTS` compiler flag to `Debug` and `Release Tests` configurations. This allows us to write better mocks in the tests. See `WebSocketClient` test helper.

- `WebSocketClient` publishes connection events when the connection status changes.

- `Client` now exposes `connectionStatus` publicly.


Sources_v3/WebSocketClient/ChatClientConnectionStatus.swift
```diff
@@ -4,12 +4,49 @@
 
 import Foundation
 
-public typealias ConnectionId = String
+// `ChatClientConnectionStatus` is just a simplified and friendlier wrapper around `WebSocketConnectionState`.
+
+/// Describes the possible states of the client connection to the servers.
+public enum ChatClientConnectionStatus: Equatable {
```
buh, line 10: Maybe better to name it as `ClientConnectionStatus`, because `ChatClient` is just a type alias for `Client`?

Sources_v3/WebSocketClient/ChatClientConnectionStatus.swift
```diff
@@ -4,12 +4,49 @@
...
+    /// The client is disconnected. This is an initial state. Optionally contains an error, if the connection was disconnected
+    /// due to an error.
+    case disconnected(error: ClientError? = nil)
```
buh, line 13: I liked how it was in `ConnectionState` as `notConnected`, but if you decide to change it, I think it should be changed in `ConnectionState` too to avoid confusing.

Sources_v3/WebSocketClient/Events/ConnectionEvents.swift
```diff
@@ -32,3 +32,15 @@ public struct HealthCheckEvent: ConnectionEvent, EventWithPayload {
         )
     }
 }
+
+/// Emitted when `Client` changes it's connection status. You can listen to this event and indicate the different connection
+/// states in the UI (banners like "Offline", "Reconnecting"", etc.).
+public struct ConnectionStatusUpdated: Event {
+    /// The current connection status of `Client`
+    public var connectionStatus: ChatClientConnectionStatus {
```
buh, line 40: I think could be `lazy var` to avoid recreating of `ChatClientConnectionStatus`.

buh (approved):
Please check a couple suggestions.

Sources_v3/WebSocketClient/Events/ConnectionEvents.swift
VojtaStavik: This has to be a computed variable, otherwise it'd just cashed the first "state" value and then never change again.

Sources_v3/WebSocketClient/ChatClientConnectionStatus.swift
VojtaStavik: I renamed it to `ConnectionStatus` in `e75bc67`. It's clear enough since it's the only public type like that we expose.

Sources_v3/WebSocketClient/ChatClientConnectionStatus.swift
VojtaStavik: Renamed to `disconnected` in a219d59

evsaev (approved):

PR #457, opened 2020-09-02
[CIS-185] Internet connection for WebSocketClient

buh:
In the PR:
- fixed InternetConnection crash and refactor the status notification userInfo key
- added: `ConnectionState.DisconnectionSource.internetConnectionUnavailable`
- added `InternetConnection` as a dependency for `WebSocketClient` and `DefaultReconnectionStrategy`
- added subscription for `InternetConnection` notifications in `WebSocketClient`
- fixed: removed `WebSocketClient` as an observer for App State notifications on `deinit()`
- changed: `WebSocketClient.disconnect()` will check if `connectionState.isActive` before changing the state to `.disconnecting`
- changed: when app did enter background and `WebSocketClient` doesn't allowed to keep connected, then if `connectionState.isActive` it do `disconnect()`
- added: `DefaultWebSocketReconnectionStrategy` will skip reconnecting if Internet connection unavailable
- added: `Client` will start `InternetConnection` monitoring on `connect()` and will stop on `disconnect()`
- added tests for `WebSocketClient` when Internet connection available and unavailable

The flow:
image

⚠️ Note: I didn't test it on a device because we don't have yet configured signing certificates for Sample v3 app.

### DIFF
--- a/Sources_v3/Utils/Internet Connection/InternetConnection.swift
+++ b/Sources_v3/Utils/Internet Connection/InternetConnection.swift
@@ -9,8 +9,12 @@ extension Notification.Name {
     static let internetConnectionStatusDidChange = Notification.Name("io.getstream.StreamChatClient.internetConnectionStatus")
 }
 
-extension InternetConnection {
-    static let statusUserInfoKey = "statusUserInfoKey"
+extension Notification {
+    static let internetConnectionStatusUserInfoKey = "internetConnectionStatus"
+    
+    var internetConnectionStatus: InternetConnection.Status? {
+        userInfo?[Self.internetConnectionStatusUserInfoKey] as? InternetConnection.Status
+    }
 }
 
 /// An Internet Connection monitor.
@@ -45,6 +49,10 @@ class InternetConnection {
         self.monitor.delegate = self
     }
     
+    deinit {
+        stop()
+    }
+    
     func start() {
         monitor.start()
     }
@@ -56,10 +64,12 @@ class InternetConnection {
 
 extension InternetConnection: InternetConnectionDelegate {
     func internetConnectionStatusDidChange(status: Status) {
+        log.info("Internet Connection: \(status)")
+        
         notificationCenter.post(
             name: .internetConnectionStatusDidChange,
             object: self,
-            userInfo: [InternetConnection.statusUserInfoKey: status]
+            userInfo: [Notification.internetConnectionStatusUserInfoKey: status]
         )
     }
 }
@@ -146,12 +156,11 @@ private extension InternetConnection {
         func stop() {
             monitor?.cancel()
             monitor = nil
-            delegate?.internetConnectionStatusDidChange(status: .unknown)
         }
         
         private func createMonitor() -> NWPathMonitor {
             let monitor = NWPathMonitor()
-            monitor.pathUpdateHandler = { [unowned self] in self.updateStatus(with: $0) }
+            monitor.pathUpdateHandler = { [weak self] in self?.updateStatus(with: $0) }
             return monitor
         }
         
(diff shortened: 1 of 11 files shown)


Sources_v3/Utils/Internet Connection/InternetConnection.swift
```diff
@@ -45,6 +49,10 @@ class InternetConnection {
         self.monitor.delegate = self
     }
     
+    deinit {
+        stop()
+    }
+    
     func start() {
         monitor.start()
     }
```
VojtaStavik, line 58: Why we don't start listening automatically when the object is created? Seems to me like an unnecessary complication. `start()` and `stop()` imo should even be public methods 🤷‍♂️

Sources_v3/Utils/Internet Connection/InternetConnection.swift
```diff
@@ -146,12 +156,11 @@ private extension InternetConnection {
         func stop() {
             monitor?.cancel()
             monitor = nil
-            delegate?.internetConnectionStatusDidChange(status: .unknown)
         }
         
         private func createMonitor() -> NWPathMonitor {
             let monitor = NWPathMonitor()
-            monitor.pathUpdateHandler = { [unowned self] in self.updateStatus(with: $0) }
+            monitor.pathUpdateHandler = { [weak self] in self?.updateStatus(with: $0) }
```
VojtaStavik, line 163: Why? The original version seems to be more correct.

Sources_v3/WebSocketClient/WebSocketReconnectionStrategy.swift
```diff
@@ -19,23 +19,31 @@ class DefaultReconnectionStrategy: WebSocketClientReconnectionStrategy {
     static let maximumReconnectionDelay: TimeInterval = 25
     
     private var consecutiveFailures = 0
+    private let inernetConnection: InternetConnection
+    
+    init(inernetConnection: InternetConnection) {
+        self.inernetConnection = inernetConnection
+    }
```
VojtaStavik, line 26: 🤷‍♂️

VojtaStavik (commented):
This seems overly complex and I need some time to wrap my head around this. I'll review later.

Sources_v3/Utils/Internet Connection/InternetConnection.swift
buh: IMHO
1. `WebSocketClient` should handle Internet connection when it's ready, user and token should be setupped.
2. On Client disconnect (basically logout) we shouldn't do anything related to WebSocket connect or anything else.
3. It's internal work/responsibility to handle connection until user do logout.
4. I'm not sure about public API, probably it's good to share it for UI, but `Client.internetConnection` should be internal var.

Sources_v3/Utils/Internet Connection/InternetConnection.swift
buh: It crashed on this point. Easy to reproduce, but difficult to say why.

Sources_v3/WebSocketClient/WebSocketReconnectionStrategy.swift
buh: We discussed that. `DefaultReconnectionStrategy` shouldn't retry reconnect while Internet is unavailable.

VojtaStavik: I thought you'd something like this https://stackoverflow.com/a/31652395/3789847

VojtaStavik:
@buh Let me finish this :)

buh:
@VojtaStavik go ahead! I think you're right. We need to use `NSError` for `DefaultReconnectionStrategy`

PR #462, opened 2020-09-03
[CIS-315] Re-watch existing channels after the WSConnection is established

mantergo:
**In this PR:** 

- Introduce `ChannelWatchStateUpdater `

### DIFF
--- a/Sources_v3/ChatClient.swift
+++ b/Sources_v3/ChatClient.swift
@@ -173,7 +173,8 @@ public class Client<ExtraData: ExtraDataTypes> {
         // All production workers
         let workerBuilders: [WorkerBuilder] = [
             MessageSender<ExtraData>.init,
-            NewChannelQueryUpdater<ExtraData>.init
+            NewChannelQueryUpdater<ExtraData>.init,
+            ChannelWatchStateUpdater<ExtraData>.init
         ]
         
         self.init(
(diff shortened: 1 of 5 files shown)


Sources_v3/Workers/Background/WebSocketConnectedProcessor.swift
```diff
@@ -0,0 +1,65 @@
+//
+// Copyright © 2020 Stream.io Inc. All rights reserved.
+//
+
+import CoreData
+
+/// After WebSocket changes it's state to `connected` we need to start watching existing channels so we can receive updated on them.
+/// This background worker observes the WebSocket state and on `connected` we are fetching all channels from DB and sending
```
VojtaStavik, line 8: nit: ~observes the WebSocket state~ -> listents to `ConnectionStatusUpdated` event ?

Sources_v3/Workers/Background/WebSocketConnectedProcessor.swift
```diff
@@ -0,0 +1,65 @@
...
+/// ChannelListQuery to backend to start `watching` channels.
+final class WebSocketConnectedProcessor<ExtraData: ExtraDataTypes>: Worker {
```
VojtaStavik, line 10: I think this is more about the channels than the web socket. My suggestion: `ChannelWatcher` :)

VojtaStavik (approved):
Nice and easy :) 🚢 🇮🇹

VojtaStavik:
>  but I think it should not be related to channels cause probably in future we will need some other stuff on connected event.

The idea is that we will have many small workers doing small things. So it's likely, that we will create another object for that; possibly sharing some code with this one, of course.

PR #464, opened 2020-09-03
[CIS-323][CIS-324] ChannelList fixes

VojtaStavik:
### In this PR

- `ChannelListController` doesn't show deleted channels
- `ListDatabaseObserver` doesn't emit an explicit `update` case when moving an item. It assumes the update is implicit, same way UIKit does it.

---

@cardoso , please try with the sample app. This should fix the crashes.


Sources_v3/Controllers/ListDatabaseObserver.swift
```diff
@@ -9,8 +9,7 @@ public enum ListChange<Item> {
     /// A new item was inserted on the given index path.
     case insert(_ item: Item, index: IndexPath)
     
-    /// An item was moved from `fromIndex` to `toIndex`. Moving an item also automatically generates
-    /// an `update` change.
+    /// An item was moved from `fromIndex` to `toIndex`. Moving an item also automatically mean you should reload its UI.
     case move(_ item: Item, fromIndex: IndexPath, toIndex: IndexPath)
     
     /// An item was updated at the given `index`. An `update` change is also automatically generated by moving an item.
```
mantergo, line 15: ~~An `update` change is also automatically generated by moving an item.~~

mantergo (approved):

PR #465, opened 2020-09-04
Failing `ChannelWatchStateUpdater` test

mantergo:
**In this PR:**

- Fix failing `ChannelWatchStateUpdater` test

_The test fails because we are comparing endpoints and in `==` method of `AnyEndpoint` 2 arrays with different elements order are treated as not equal. It's hard to make any sorting cause we are dealing with String(describing: `Encodable`) comparing. I decided that test modification doesn't affect test goal so it's the best way to fix error rather than modifying client code for test._


VojtaStavik (approved):
OK, fair enough 👍  Thanks for the explanation

PR #468, opened 2020-09-04
[CIS-326] Fix carthage building v3

b-onc:
..by removing unncessary shared schemes.
Carthage, by default, uses the shared schemes in the current folder. We had v3 schemes shared, and since v2 and v3 had both the same name (StreamChatClient) it was causing an overlap.
I've tested by creating a fresh project, and fetching carthage from local folder.


buh (approved):

NicolasLourenco:
Tested, it's ok for me 👌

PR #471, opened 2020-09-07
[CIS-185] Connect the web socket when connection restored

VojtaStavik:
This is an alternative approach to #457. I tried to come up with a different and more straightforward solution that relies more on the existing infrastructure.

#### A typical scenario looks like:
- The engine returns "no internet" error
- `ReconnectionStrategy` returns `nil` and no reconnection is scheduled
- An `InternetConnection` waiter is scheduled
- Once the connection is back online, WS tries to reconnect

It seems we don't need any special handling for the foreground/background app state. As far as I understand, the existing implementation covers that situation properly.

---

@buh Please take a look and let me know if I missed something. It seems to me this is all we need to do.

### DIFF
--- a/Sample_v3/LogStore.swift
+++ b/Sample_v3/LogStore.swift
@@ -6,7 +6,7 @@ import Foundation
 import StreamChatClient
 
 class LogStore: BaseLogDestination {
-    var logs = ""
+    @Atomic var logs = ""
     
     static let shared = LogStore()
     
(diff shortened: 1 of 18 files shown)


buh (approved):
Looks great!
I had to try at the beginning to avoid `start`/`stop` monitors. It increases unneeded complexity.
Great clean up for WebSocket engines! 👍

PR #472, opened 2020-09-07
Mark v3 schemes as shared... again...

b-onc:
...so that fastlane scan can discover them. The only way we can have 2 StreamChatClient schemes is if we can have v2 project under a workspace. That's the first thing carthage will search.

I've tested a fresh carthage project from local repo and it works for me, @buh can you also please test this one too? Now both carthage should work, and CI should pass

xctestplan file is edited since I run test_v3 fastlane lane locally and testplan file had changes, since it was only whitespace changes I think they should be committed

### DIFF
--- a/StreamChat.xcworkspace/contents.xcworkspacedata
+++ b/StreamChat.xcworkspace/contents.xcworkspacedata
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Workspace
+   version = "1.0">
+   <FileRef
+      location = "group:StreamChat.xcodeproj">
+   </FileRef>
+</Workspace>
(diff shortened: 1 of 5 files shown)


Tests_v3/StreamChatClientTestPlan.xctestplan
```diff
@@ -1,32 +1,37 @@
...
+        "containerPath": "container:StreamChat_v3.xcodeproj",
+        "identifier": "799C9450247D59B1001F1104",
+        "name": "StreamChatClientTests"
       }
     }
   ],
-  "version" : 1
-}
+  "version": 1
+}
```
VojtaStavik, line 37: This shouldn't be here, it's just noise from running CI script locally.

b-onc: Yeah exactly I wrote in the description :D IMHO if this noise will be generated everytime test_v3 lane is run, why don't we commit it? Running a lane locally should not generate changes in repo

VojtaStavik: Two things:
- when you edit the testplan in Xcode, it will revert all whitespace changes back
- you can see, it adds the `CI` flag there. You don't want to run it locally with the CI flag

buh (approved):
It works

VojtaStavik:
@b-onc Can we please merge this so we can have CI support again?

Tests_v3/StreamChatClientTestPlan.xctestplan
VojtaStavik: I think it's on our list in the Fastfile improvements JIRA ticket :)

PR #476, opened 2020-09-08
[CIS-330] UIKit+Combine example

mantergo:
**In this PR:** 

- Add Combine example for `SimpleChatChannelsViewController`
- Add Combine example for `SimpleChatViewController`

### DIFF
--- a/Sample_v3/LoginViewController.swift
+++ b/Sample_v3/LoginViewController.swift
@@ -111,6 +111,24 @@ extension LoginViewController {
                     )
                 })
             }
+        case .combineUIKitSimpleChatIndexPath:
+            if #available(iOS 13, *) {
+                let storyboard = UIStoryboard(name: "CombineSimpleChat", bundle: nil)
+                
+                guard
+                    let initial = storyboard.instantiateInitialViewController() as? SplitViewController,
+                    let navigation = initial.viewControllers.first as? UINavigationController,
+                    let channels = navigation.children.first as? CombineSimpleChannelsViewController
+                else {
+                    return
+                }
+                
+                channels.channelListController = channelListController
+
+                UIView.transition(with: view.window!, duration: 0.5, options: .transitionFlipFromLeft, animations: {
+                    self.view.window?.rootViewController = initial
+                })
+            }
         default:
             return
         }
@@ -123,6 +141,7 @@ private extension IndexPath {
     static let jwtIndexPath = IndexPath(row: 4, section: 0)
     static let simpleChatIndexPath = IndexPath(row: 0, section: 1)
     static let swiftUISimpleChatIndexPath = IndexPath(row: 1, section: 1)
+    static let combineUIKitSimpleChatIndexPath = IndexPath(row: 2, section: 1)
 }
 
 // MARK: - Inputs
--- a/Sample_v3/Samples/CombineSimpleChat/CombineSimpleChannelsViewController.swift
+++ b/Sample_v3/Samples/CombineSimpleChat/CombineSimpleChannelsViewController.swift
@@ -1,0 +1,228 @@
+//
+// Copyright © 2020 Stream.io Inc. All rights reserved.
+//
+
+import Combine
+import StreamChatClient
+import UIKit
+
+@available(iOS 13, *)
+class CombineSimpleChannelsViewController: UITableViewController {
+    private lazy var longPressRecognizer = UILongPressGestureRecognizer(
+        target: self,
+        action: #selector(handleLongPress)
+    )
+
+    var channelListController: ChannelListController!
+    
+    var chatClient: ChatClient {
+        channelListController.client
+    }
+    
+    var detailViewController: CombineSimpleChatViewController?
+    
+    private lazy var cancellables: Set<AnyCancellable> = []
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        
+        subscribeToCombinePublishers()
+        
+        // Do any additional setup after loading the view.
+        navigationItem.leftBarButtonItem = UIBarButtonItem(
+            title: "Settings",
+            style: .plain,
+            target: self,
+            action: #selector(handleSettingsButton)
+        )
+
+        let addButton = UIBarButtonItem(barButtonSystemItem: .add, target: self, action: #selector(insertNewObject(_:)))
+        navigationItem.rightBarButtonItem = addButton
+        if let split = splitViewController {
+            let controllers = split.viewControllers
+            detailViewController = (controllers[controllers.count - 1] as! UINavigationController)
+                .topViewController as? CombineSimpleChatViewController
+        }
+
+        tableView.addGestureRecognizer(longPressRecognizer)
+    }
+    
+    private func subscribeToCombinePublishers() {
+        channelListController
+            .statePublisher
+            .sink { (state) in
+                print("State changed: \(state)")
+            }
+            .store(in: &cancellables)
+            
+        channelListController
+            .channelsChangesPublisher
+            .receive(on: RunLoop.main)
+            .sink { [weak self] in self?.tableView.applyListChanges(changes: $0) }
+            .store(in: &cancellables)
+    }
+
+    override func viewWillAppear(_ animated: Bool) {
+        clearsSelectionOnViewWillAppear = splitViewController!.isCollapsed
+        super.viewWillAppear(animated)
+    }
+    
+    @objc
+    func handleSettingsButton(_ sender: Any) {
+        guard let settingsViewController = UIStoryboard.settings.instantiateInitialViewController() else {
+            return
+        }
+        
+        present(settingsViewController, animated: true)
+    }
+
+    @objc
+    func insertNewObject(_ sender: Any) {
+        let id = UUID().uuidString
+        let controller = chatClient.channelController(
+            createChannelWithId: .init(type: .messaging, id: id),
+            members: [chatClient.currentUserId],
+            extraData: .init(name: "Channel" + id.prefix(4), imageURL: nil)
+        )
+        controller.startUpdating()
+    }
+
+    // MARK: - Segues
+
+    override func prepare(for segue: UIStoryboardSegue, sender: Any?) {
+        if segue.identifier == "showDetail" {
+            if let indexPath = tableView.indexPathForSelectedRow {
+                let channel = channelListController.channels[indexPath.row]
+                let controller = (segue.destination as! UINavigationController)
+                    .topViewController as! CombineSimpleChatViewController
+                controller.channelController = chatClient.channelController(for: channel.cid)
+                controller.navigationItem.leftBarButtonItem = splitViewController?.displayModeButtonItem
+                controller.navigationItem.leftItemsSupplementBackButton = true
+                detailViewController = controller
+            }
+        }
+    }
+
+    // MARK: - Table View
+
+    override func tableView(_ tableView: UITableView, numberOfRowsInSection section: Int) -> Int {
+        channelListController.channels.count
+    }
+
+    override func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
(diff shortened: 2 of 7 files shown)


VojtaStavik (changes requested):
Pavel, I'm not sure this works well as sample code. It's too clever and hard to understand for someone who has never seen this before.

Can you please just copy&paste the classes? Maybe, we can introduce some common superclass with all the UI code shared for both examples + the things not relevant to the usage of the SDK??

mantergo:
> Pavel, I'm not sure this works well as sample code. It's too clever and hard to understand for someone who has never seen this before.
> 
> Can you please just copy&paste the classes? Maybe, we can introduce some common superclass with all the UI code shared for both examples + the things not relevant to the usage of the SDK??

That is something I started with, but copy-pasting is not an option cause it will be a mess with supporting both examples.
Also I believe inheritance will be bring even more complexity and will make it harder to navigate through example source code and files.
I suggest just to add some heavy comments on that or ask other guys opinions.

Sample_v3/Samples/CombineSimpleChat/CombineSimpleChannelsViewController.swift
```diff
@@ -0,0 +1,216 @@
...
+    lazy var channelListController: ChannelListController = chatClient
+        .channelListController(query: ChannelListQuery(
+            filter: .in("members", [chatClient.currentUserId]),
+            pagination: [.limit(25)],
+            options: [.watch]
+        ))
+    
+    var detailViewController: CombineSimpleChatViewController?
+    
+    @available(iOS 13, *)
```
VojtaStavik, line 24: I think this whole class should be `@available(iOS 13, *)`

Sample_v3/Samples/CombineSimpleChat/CombineSimpleChannelsViewController.swift
```diff
@@ -0,0 +1,216 @@
...
+        if let split = splitViewController {
+            let controllers = split.viewControllers
+            detailViewController = (controllers[controllers.count - 1] as! UINavigationController)
+                .topViewController as? CombineSimpleChatViewController
+        }
+
+        tableView.addGestureRecognizer(longPressRecognizer)
+    }
+    
+    private func bind() {
```
VojtaStavik, line 51: Please don't forget this is sample code for our customers. Can we use a more descriptive name for this? Like `setUpCombineSubscribers` or similar?

Sample_v3/Samples/CombineSimpleChat/CombineSimpleChatViewController.swift
```diff
@@ -0,0 +1,434 @@
...
+            if let channel = channelController.channel {
+                didUpdate(channel: .update(channel))
+            }
+        }
+    }
+    
+    // MARK: - Combine
+    
+    @available(iOS 13, *)
+    private lazy var cancellables: Set<AnyCancellable> = []
```
VojtaStavik, line 36: Same thing here. The whole class should ideally be `@available(iOS 13, *)`

Sources_v3/Controllers/ListDatabaseObserver.swift
```diff
@@ -200,3 +201,24 @@ extension ListDatabaseObserver where DTO == Item {
...
+            case let .update(_, index: index):
+                reloadRows(at: [index], with: .automatic)
+            case let .remove(_, index: index):
+                deleteRows(at: [index], with: .automatic)
+            }
+        }
+        
+        endUpdates()
+    }
+}
```
VojtaStavik, line 224: I don't think we're ready to have this in the SDK yet. We'd need to make this properly - have the animations parametrized etc.. I think this can live in the sample app though :)

VojtaStavik (commented):
It's a lot of duplicated code but that's exactly what we need for the sample app. Can you spend a little bit more time on it, and make the whole controller more Combine-like? For example, it's very common to use `assign` instead of sink, when you just update the value. We could probably do something like:
```swift
// pseudo code
channelController
    .channelChangePublisher
    .map { $0.name }
    .assign(to: \.title, on: self)
    .store(in: &cancellables)
```

Please think a little bit about these things, and try to find was how to use `Combine` to its full potential. Maybe there are no use cases like this, and that's fine. We just need to be sure.

Sample_v3/Samples/CombineSimpleChat/CombineSimpleChatViewController.swift
```diff
@@ -0,0 +1,438 @@
...
+            .store(in: &cancellables)
+        
+        /// The subscription  below receives a `TypingEvent` and updates the view controller's `navigationItem.prompt` to show that an user is currently typing.
+        channelController
+            .typingEventPublisher
+            .map { [weak self] in (self?.channelController.dataStore.user(id: $0.userId), $0) }
+            .filter { $0.0 != nil }
+            .map { $0.1.isTyping ? "\($0.0?.name ?? $0.1.userId) is typing..." : "" }
+            .assign(to: \.navigationItem.prompt, on: self)
+            .store(in: &cancellables)
```
VojtaStavik, line 87: It'd be great to add some comments to these operations ;)

Sample_v3/Samples/CombineSimpleChat/CombineSimpleChannelsViewController.swift
```diff
@@ -0,0 +1,213 @@
...
+    lazy var channelListController: ChannelListController = chatClient
+        .channelListController(query: ChannelListQuery(
+            filter: .in("members", [chatClient.currentUserId]),
+            pagination: [.limit(25)],
+            options: [.watch]
+        ))
+    
+    var detailViewController: CombineSimpleChatViewController?
+    
+    @available(iOS 13, *)
```
VojtaStavik, line 25: Is this `available` still needed?

Sample_v3/Samples/CombineSimpleChat/CombineSimpleChatViewController.swift
```diff
@@ -0,0 +1,437 @@
...
+    ///
+    /// # subscribeToCombinePublishers
+    ///
+    /// Here we bind `channelControllers` publishers so we can observe the changes.
+    private func subscribeToCombinePublishers() {
+        /// `ChannelController` will not trigger the `channelChangePublisher` on the initial channel set so
+        /// we can` prepend` our `channelChangePublisher` sequence with the initial channel manually.
+        let initialChannel = Just(channelController.channel)
+            .compactMap { $0 }
+            .map { EntityChange<Channel>.update($0) }
```
VojtaStavik, line 44: I'm not sure this is a good idea actually. It's desired that the locally available channels are loaded without animation, only the changes are animated. Seem to me that the preferred solution is just to call `reloadData()` in `channelController { didSet { } }`?

Sample_v3/Samples/CombineSimpleChat/CombineSimpleChatViewController.swift
```diff
@@ -0,0 +1,437 @@
...
+                switch change {
+                case .create:
+                    return nil
+                case let .update(channel):
+                    return channel
+                case .remove:
+                    self?.dismiss(animated: true)
+                    return nil
+                }
+            }
```
VojtaStavik, line 64: Nice 👍

Sample_v3/Samples/CombineSimpleChat/CombineSimpleChatViewController.swift
```diff
@@ -0,0 +1,437 @@
...
+            .share()
+        
+        updatedChannel
+            .map { $0.extraData.name ?? $0.cid.description }
+            .receive(on: RunLoop.main)
```
VojtaStavik, line 69: Maybe `updatedChannel` can be received on main loop so you don't have to write it every time?

Sample_v3/Samples/CombineSimpleChat/CombineSimpleChatViewController.swift
```diff
@@ -0,0 +1,437 @@
...
+            .sink { [weak self] in self?.tableView.applyListChanges(changes: $0) }
+            .store(in: &cancellables)
+        
+        /// The subscription  below receives a `TypingEvent` and updates the view controller's `navigationItem.prompt` to show that an user is currently typing.
+        channelController
+            .typingEventPublisher
+            /// Map user with the typing event.
+            .map { [weak self] in (self?.channelController.dataStore.user(id: $0.userId), $0) }
+            /// Skip if user was not fetched succesfully from DB.
+            .filter { $0.0 != nil }
```
VojtaStavik, line 93: Aren't these two lines equivalent to `compactMap { }` or you split in like this on purpose?

VojtaStavik (approved):
Cool, great work! Thanks 👍

Sample_v3/Samples/CombineSimpleChat/CombineSimpleChatViewController.swift
mantergo: Seems like we mixed up here.
It's channel in "Chat" VC with messages.

Sample_v3/Samples/CombineSimpleChat/CombineSimpleChatViewController.swift
mantergo: I can't see `compactMap` 🧐
`typingEvent` from sequence is never `nil` as well as tuple itself.

Sample_v3/LoginViewController.swift
```diff
@@ -95,6 +95,16 @@ extension LoginViewController {
                     )
                 })
             }
+        case .combineUIKitSimpleChatIndexPath:
+            if #available(iOS 13, *) {
+                logIn(apiKey: apiKey, userId: userId, userName: userName, token: token)
```
cardoso, line 100: This line should be removed

Sample_v3/Samples/SimpleChat/SimpleChatViewController.swift
```diff
@@ -45,22 +45,7 @@ final class SimpleChatViewController: UITableViewController, ChannelControllerDe
...
-                tableView.moveRow(at: fromIndex, to: toIndex)
-            case let .update(_, index: index):
-                tableView.reloadRows(at: [index], with: .automatic)
-            case let .remove(_, index: index):
-                tableView.deleteRows(at: [index], with: .automatic)
-            }
-        }
-        
-        tableView.endUpdates()
+        tableView.applyListChanges(changes: changes)
```
cardoso, line 48: Feel bad to ask this, but I'd leave this explicit for now for educational purposes (this extension is not shipped with the sdk yet)

Sample_v3/Samples/CombineSimpleChat/CombineSimpleChannelsViewController.swift
```diff
@@ -0,0 +1,212 @@
...
+                detailViewController = controller
+            }
+        }
+    }
+
+    // MARK: - Table View
+
+    override func numberOfSections(in tableView: UITableView) -> Int {
+        1
+    }
```
cardoso, line 111: This can be left out I think.

Sample_v3/Samples/CombineSimpleChat/CombineSimpleChatViewController.swift
```diff
@@ -0,0 +1,436 @@
...
+        composerView.textView.text = ""
+    }
+    
+    ///
+    /// # showChannelActionsAlert
+    ///
+    /// The method below displays a `UIAlertController` with many actions that can be taken on the `channelController` such as `addMembers`, `removeMembers`,
+    /// and `deleteChannel`.
+    ///
+    @objc func showChannelActionsAlert() {
```
cardoso, line 261: Can you copy latest version of this from `SimpleChatViewController.swift` ? It allows specifying member id for the actions.

cardoso (approved):
Nice work @mantergo 👍

Sample_v3/Samples/SimpleChat/SimpleChatViewController.swift
mantergo: Ok, moved and used it only in combine example 🙂

PR #406, opened 2020-08-14
Preparation for 1-1 channel creation

mantergo:
**In this PR:**

- Add possibility to reset DB observers on receiving channel id from backend

**Q:** If we will stick to this changes, should we also perform them to `ChannelListController` for unification even though they are not needed from technical perspective at the moment?

PR #477, opened 2020-09-08
Fixes ChannelController.startUpdating completion called multiple times

b-onc:
As it is, ChannelController.startUpdating passes its completion param to multiple places:
- to setupDatabaseObservers
- to channelCreatedCallback
so, if an error occurs in multiple places, it'll be reported in multiple places. If this is what we want, then it's correct.
Another issue is that `startDatabaseObservers` should not be called before channel is actually created (so `isChannelAlreadyCreated` should be true) but it's not checked. I've also refactored it so `startDatabaseObservers` and `setupEventObservers` are called once, in the correct place: if channel exists, called in `startUpdating` otherwise called in `channelCreatedCallback` (hence in set(channelQuery:))

### DIFF
--- a/Sources_v3/Controllers/ChannelController/ChannelController.swift
+++ b/Sources_v3/Controllers/ChannelController/ChannelController.swift
@@ -168,11 +168,11 @@ public class ChannelControllerGeneric<ExtraData: ExtraDataTypes>: Controller, De
     // This callback is called after channel is created on backend but before channel is saved to DB. When channel is created
     // we receive backend generated cid and setting up current `ChannelController` to observe this channel DB changes.
     // Completion will be called if DB fetch will fail after setting new `ChannelQuery`.
-    private func channelCreated(completion: ((_ error: Error?) -> Void)?) -> ((ChannelId) -> Void) {
+    private func channelCreated(forwardErrorTo completion: ((_ error: Error?) -> Void)?) -> ((ChannelId) -> Void) {
         return { [weak self] cid in
             guard let self = self else { return }
             self.isChannelAlreadyCreated = true
-            self.set(channelQuery: ChannelQuery(cid: cid, channelQuery: self.channelQuery), completion: completion)
+            completion?(self.set(cid: cid))
         }
     }
     
@@ -250,13 +250,16 @@ public class ChannelControllerGeneric<ExtraData: ExtraDataTypes>: Controller, De
     /// - Parameter completion: Called when the controller has finished fetching remote data.
     ///                         If the data fetching fails, the `error` variable contains more details about the problem.
     public func startUpdating(_ completion: ((_ error: Error?) -> Void)? = nil) {
-        eventObservers.removeAll()
-        startDatabaseObservers(completion)
-
-        // Update observing state
-        state = .localDataFetched
-
-        let channelCreatedCallback = isChannelAlreadyCreated ? nil : channelCreated(completion: completion)
+        let setStateBasedOnError: ((_ error: Error?) -> Void) = { [weak self] error in
+            // Update observing state
+            self?.state = error == nil ? .localDataFetched : .localDataFetchFailed(ClientError(with: error))
+        }
+        
+        if isChannelAlreadyCreated {
+            setStateBasedOnError(startDatabaseObservers())
+        }
+        
+        let channelCreatedCallback = isChannelAlreadyCreated ? nil : channelCreated(forwardErrorTo: setStateBasedOnError)
         updater.update(
             channelQuery: channelQuery,
             channelCreatedCallback: channelCreatedCallback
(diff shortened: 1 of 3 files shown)


Sources_v3/Controllers/ChannelController/ChannelController_Tests.swift
```diff
@@ -310,6 +310,9 @@ class ChannelController_Tests: StressTestCase {
 
         // Simulate `startUpdating` call
         controller.startUpdating()
+        
+        // Simulate updater's channelCreatedCallback call
+        env.channelUpdater!.update_channelCreatedCallback!(channelId)
```
b-onc, line 315: This test was actually wrong, and when I fixed the ChannelUpdater this test failed.
ChannelController shouldn't have the channel object before `channelCreatedCallback` is called from its worker.

Sources_v3/Controllers/ChannelController/ChannelController.swift
```diff
@@ -266,15 +269,19 @@ public class ChannelControllerGeneric<ExtraData: ExtraDataTypes>: Controller, De
...
     // After setting new channelQuery we need to reset current DB observers to observe new channel.
     // Completion will be called if observers will fail to fetch data for the new channel.
     private func set(channelQuery: ChannelQuery<ExtraData>, completion: ((_ error: Error?) -> Void)?) {
-        guard self.channelQuery.cid != channelQuery.cid else { return }
-        self.channelQuery = channelQuery
+        if self.channelQuery.cid != channelQuery.cid {
+            self.channelQuery = channelQuery
+        }
         startDatabaseObservers(completion)
+        setupEventObservers()
```
VojtaStavik, line 284: The functionality changed here a bit. Before, we guarded for `cid != channelQuery.cid` and did nothing if it was the same. Now we always start database observers and event observers. Is this intentional?

Sources_v3/Controllers/ChannelController/ChannelController.swift
```diff
@@ -284,14 +291,15 @@ public class ChannelControllerGeneric<ExtraData: ExtraDataTypes>: Controller, De
         do {
             try channelObserver.startObserving()
             try messagesObserver.startObserving()
+            completion?(nil)
```
VojtaStavik, line 294: Why we even use the completion here? Why not just throw an error? We already talked about it here https://github.com/GetStream/stream-chat-swift/pull/406#discussion_r472901467 and it seems to me, that with the recent changes, it makes much more sense. It will simplify the double-completion brain breaker in `startUpdating()`

VojtaStavik (changes requested):
This piece of logic is quite complex and I have a feeling it should be possible to make this simpler. Would you mind spending a little bit more time exploring ways to simplify this?

Sources_v3/Controllers/ChannelController/ChannelController.swift
b-onc: Yes, definitely. I've thought about this after opening the PR. It doesn't need to have a completion, just a return value, or throws would be enough

Sources_v3/Controllers/ChannelController/ChannelController.swift
b-onc: Yes, this is intentional.
If `isChannelAlreadyCreated` is false, we don't setup observers in `startUpdating`. In that case, we'll start observers in `set(query:)` but in this case, if the created channel has the same cid (so if it isn't a DM channel), we won't be setting it here. So, we should still setup observers even if the cid matches.

Sources_v3/Controllers/ChannelController/ChannelController.swift
```diff
@@ -266,32 +269,42 @@ public class ChannelControllerGeneric<ExtraData: ExtraDataTypes>: Controller, De
...
-        self.channelQuery = channelQuery
-        startDatabaseObservers(completion)
+    /// Sets new cid of the query if necessary, and resets event and database observers.
+    ///
+    /// This should only be called when the controller is initialized with a new channel (which doesn't exsit on backend), and after that channel is created on backend.
+    /// If the newly created channel has a different cid than initially thought (such is the case for direct messages - backend generates custom cid), this function will set the new cid and reset observers.
+    /// If the cid is still the same, this function will only reset the observers - since we don't need to set a new query in that case.
+    /// - Parameter cid: New cid for the channel
+    /// - Returns: Erorr if it occurs while setting up database observers.
+    private func set(cid: ChannelId) -> Error? {
```
VojtaStavik, line 284: Returning `Error?` is the same thing as marking the method `throws` only less elegant, don't you think? 🤷‍♂️

b-onc: Dependens, IMHO. `throws` requires the caller to use `do/catch` and it's not elegant in this case, propagating error via return values is more elegant.
If I were to use `throws` in this func, I'd have to catch it in `channelCreated(forwardErrorTo:)` return value and call `completion` with its return value, so I saw no value ¯\_(ツ)_/¯

VojtaStavik: OK, up to you.

VojtaStavik (approved):

PR #478, opened 2020-09-08
[CIS-341] [CIS-342] V3 Sample: allow specifying id in member actions & show alert on login error

cardoso:
Decided to do these two tasks in one PR because both could use the alert extension. The alert for the login error is pretty much never called since, for now, the setUser method keeps retrying forever.

Also fixes:
- a glitch when the member actions alert shows up the tableview offset can go wrong
- chatClient not receiving the API key specified in the login screen


VojtaStavik (approved):
LGTM 👍

PR #463, opened 2020-09-03
[CIS-310] Send keystroke events

buh:
- added channel event endpoint
- added `currentTime` for our `Timer` type
- added `EventSender` worker
- integrated `EventSender` into `ChannelController` as `eventWorker` and renamed `worker` as `updaterWorker`
- cleaned up a little bit `TypingStartCleanupMiddleware`

PR #480, opened 2020-09-10
[CIS-339] V3 Sample: Send keystroke events

cardoso:
List of changes:
- SimpleChatViewController.swift: Send keystroke events
- Multiple files: Remove global reference to chat client (it's now passed down via the controllers)
- SimpleChannelsViewController.swift: Comment & refactor to focus on `ChannelListController` usage.

### DIFF
--- a/Sources_v3/APIClient/Endpoints/ChannelEndpoints.swift
+++ b/Sources_v3/APIClient/Endpoints/ChannelEndpoints.swift
@@ -128,9 +128,9 @@ extension Endpoint {
         )
     }
     
-    static func sendEvent<ExtraData: ExtraDataTypes>(cid: ChannelId, eventType: EventType) -> Endpoint<EventPayload<ExtraData>> {
+    static func sendEvent(cid: ChannelId, eventType: EventType) -> Endpoint<EmptyResponse> {
         .init(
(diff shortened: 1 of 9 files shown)


Sources_v3/APIClient/Endpoints/ChannelEndpoints.swift
```diff
@@ -128,9 +128,9 @@ extension Endpoint {
         )
     }
     
-    static func sendEvent<ExtraData: ExtraDataTypes>(cid: ChannelId, eventType: EventType) -> Endpoint<EventPayload<ExtraData>> {
+    static func sendEvent(cid: ChannelId, eventType: EventType) -> Endpoint<EmptyResponse> {
         .init(
-            path: "channels/\(cid.type)/\(cid.id)",
+            path: "channels/\(cid.type)/\(cid.id)/event",
```
VojtaStavik, line 133: @buh I was surprised that you changed this but you didn't have to change the test. Then I realized then you didn't implement it in the first place, not event after the review :) https://github.com/GetStream/stream-chat-swift/pull/463#discussion_r484265095

buh: My fault. I'll add.

PR #481, opened 2020-09-10
[CIS-336] Move current-user logic to controller

evsaev:
**This PR**:
- merges `ConnectionController` to `CurrentUserController` and gets rid of `ConnectionController`;
- moves `connect/disconnect/setUser/setGuestUser/setAnonymousUser` from `ChatClient` to `CurrentUserController`;
- updates mock `ChatClient` environment to create dependencies with passed params and exposes mocked-objects in tests;


Sources_v3/ChatClient.swift
```diff
@@ -164,7 +164,7 @@ public class Client<ExtraData: ExtraDataTypes> {
     @Atomic private var connectionIdWaiters: [(String?) -> Void] = []
     
     /// The token of the current user. If the current user is anonymous, the token is `nil`.
-    @Atomic private var currentToken: Token?
+    @Atomic var currentToken: Token?
```
evsaev, line 167: I don't like making the `currentToken: Token` even internal as far as the only one who needs the access is **CurrentUserControllerGeneric**. 

I'd use a closure-based approach here but wanna discuss if it's not an overhead:
```swift

struct Getter<T> {
    private let closure: () -> T
    
    init(closure: @escaping () -> T) {
        self.closure = closure
    }
    
    var value: T {
        closure()
    }
}

struct Setter<T> {
    private let closure: (T) -> Void
    
    init(closure: @escaping (T) -> Void) {
        self.closure = closure
    }
    
    func set(_ value: T) {
        closure(value)
    }
}

struct GetterAndSetter<T> {
    private let getter: Getter<T>
    private let setter: Setter<T>

    init(getter: Getter<T>, setter: Setter<T>) {
        self.getter = getter
        self.setter = setter
    }
    
    var value: T {
        get { getter.value }
        set { setter.set(newValue) }
    }
}
```

Sources_v3/Controllers/CurrentUserController/CurrentUserController.swift
```diff
@@ -100,6 +100,121 @@ public class CurrentUserControllerGeneric<ExtraData: ExtraDataTypes>: Controller
...
+    ///                 If setting up the new user fails the completion will be called with an error.
+    func setGuestUser(userId: UserId, extraData: ExtraData.User, completion: ((Error?) -> Void)? = nil) {
+        client.disconnect()
+        client.prepareEnvironmentForNewUser(userId: userId, role: .guest, extraData: extraData) { error in
+            guard error == nil else {
+                completion?(error)
+                return
+            }
+            
+            self.client.apiClient.request(endpoint: .guestUserToken(userId: userId, extraData: extraData)) {
```
evsaev, line 144: Previously we did `[weak self]` but I think it's not the best decision:
1. once we set a new user I'd expect it to be finished regardless of a **CurrentUserControllerGeneric** lifecycle
2. as a developer I want be able to do `client.currentUserController.setUser(...)` and don't care about maintaining a strong ref to **CurrentUserControllerGeneric**

Sources_v3/ChatClient.swift
```diff
@@ -363,7 +249,7 @@ public class Client<ExtraData: ExtraDataTypes> {
     }
     
     // TODO: Not used & tested yet -> CIS-224
-    private func refreshToken(completion: @escaping (Error?) -> Void) {
+    func refreshToken(completion: @escaping (Error?) -> Void) {
```
evsaev, line 252: `refreshToken` and `prepareEnvironmentForNewUser` are the candidates for moving to **CurrentUserControllerGeneric**

At the same time, it's also IMHO OK to leave those methods in a **ChatClient** as they are mostly about modifying the data owned by a **ChatClient**

Sources_v3/Controllers/CurrentUserController/CurrentUserController.swift
VojtaStavik: Yep, makes total sense 👍

Sources_v3/ChatClient.swift
VojtaStavik: I think `prepareEnvironmentForNewUser` can/should live in `CurerentUserController`. `refreshToken` must IMO live in the `Client` since it must be accessible all the time and it would be strange if `Client` has to create a controller to do something itself. `Client` shouldn't even know something like controllers exist.

Sources_v3/ChatClient.swift
VojtaStavik: I'm not sure I understand how this would be used + what are the benefits over having the `token` internal

VojtaStavik (commented):
Looks good so far 👍

Sources_v3/ChatClient.swift
evsaev: The benefit is that we can keep `token` private and send `GetterAndSetter<Token?>` that wraps `client.token` to current-user controller.

VojtaStavik: I see where you're going with this but if we make the setter accessible internally, what difference does it make from having the `token` accessible internally?

evsaev:
The missing tests for `set***User` method family will be added shortly. Decided to push without them in place not to block the review process

Sources_v3/APIClient/Endpoints/WebSocketConnectEndpoint.swift
```diff
@@ -0,0 +1,21 @@
...
+        extraData: UserData? = nil
+    ) -> Endpoint<EmptyResponse> {
+        .init(
+            path: "connect",
+            method: .get,
+            queryItems: nil,
+            requiresConnectionId: false,
+            body: ["json": WebSocketConnectPayload(userId: userId, userRole: role, extraData: extraData)]
+        )
+    }
```
VojtaStavik, line 20: Thanks 👍

VojtaStavik (approved):
LGTM 👍 Thanks!

Sources_v3/WebSocketClient/EventMiddlewares/ChannelReadUpdaterMiddleware.swift
```diff
@@ -11,15 +11,9 @@ struct ChannelReadUpdaterMiddleware<ExtraData: ExtraDataTypes>: EventMiddleware
     func handle(event: Event, completion: @escaping (Event?) -> Void) {
         if let event = event as? MessageReadEvent<ExtraData> {
```
evsaev, line 12: After the rebase to `master` some tests related to exposing connection events were failing. 

The root cause was in `ChannelReadUpdaterMiddleware` that was blocking events that didn't enter any if statement.

FYI @b-onc @VojtaStavik

PR #487, opened 2020-09-11
[CIS-346] Background workers ownership

evsaev:
**This PR:**

- updates each background worker **not to retain** itself via database/api-client completion;
- adds tests that check workers can be released if completions haven't called yet;

### DIFF
--- a/Sources_v3/Workers/Background/ChannelWatchStateUpdater.swift
+++ b/Sources_v3/Workers/Background/ChannelWatchStateUpdater.swift
@@ -30,7 +30,7 @@ final class ChannelWatchStateUpdater<ExtraData: ExtraDataTypes>: Worker {
         webSocketConnectedObserver = WebSocketConnectedObserver(
             notificationCenter: webSocketClient.eventNotificationCenter,
             filter: { $0.connectionStatus == .connected },
-            callback: { [weak self] in self?.watchChannels() }
+            callback: { [unowned self] in self.watchChannels() }
         )
     }
     
--- a/Sources_v3/Workers/Background/MessageEditor.swift
+++ b/Sources_v3/Workers/Background/MessageEditor.swift
@@ -21,18 +21,15 @@ import Foundation
 class MessageEditor<ExtraData: ExtraDataTypes>: Worker {
     @Atomic private var pendingMessageIDs: Set<MessageId> = []
     private let observer: ListDatabaseObserver<MessageDTO, MessageDTO>
-    private let sendingQueue: DispatchQueue
+    private let backgroundContext: NSManagedObjectContext
 
     override init(database: DatabaseContainer, webSocketClient: WebSocketClient, apiClient: APIClient) {
+        backgroundContext = database.backgroundReadOnlyContext
+        
         observer = .init(
-            context: database.backgroundReadOnlyContext,
+            context: backgroundContext,
             fetchRequest: MessageDTO.messagesPendingSyncFetchRequest(),
             itemCreator: { $0 }
-        )
-        
-        sendingQueue = .init(
-            label: "co.getStream.ChatClient.MessageEditorQueue",
-            qos: .userInitiated
         )
         
         super.init(database: database, webSocketClient: webSocketClient, apiClient: apiClient)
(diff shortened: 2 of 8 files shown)


Sources_v3/Workers/Background/ChannelWatchStateUpdater.swift
```diff
@@ -30,7 +30,7 @@ final class ChannelWatchStateUpdater<ExtraData: ExtraDataTypes>: Worker {
         webSocketConnectedObserver = WebSocketConnectedObserver(
             notificationCenter: webSocketClient.eventNotificationCenter,
             filter: { $0.connectionStatus == .connected },
-            callback: { [weak self] in self?.watchChannels() }
+            callback: { [unowned self] in self.watchChannels() }
```
evsaev, line 33: That's not required but I'd like to see the crash if the lifecycle of the internal observer doesn't follow the lifecycle of a worker

Sources_v3/Workers/Background/MessageEditor.swift
```diff
@@ -21,20 +21,17 @@ import Foundation
 class MessageEditor<ExtraData: ExtraDataTypes>: Worker {
```
evsaev, line 21: In git-blame I see the fix for the crash in `MessageEditor` on iOS 12. 

However, the bug was solved partially: the `processNextMessage()` was scheduled on bg-context queue only from `handleChanges()` method but not from `removeMessageIDAndContinue()`. 

While resolving this issue I came up with the refactoring you see here. With such bug-fix + refactoring in place it was super simple to implement the **[weak self]** stuff

Sources_v3/Workers/Background/MessageEditor.swift
```diff
@@ -57,58 +54,43 @@ class MessageEditor<ExtraData: ExtraDataTypes>: Worker {
...
-                    }
-                case .failure:
-                    self.markMessage(withID: messageId, as: .syncingFailed) {
-                        self.removeMessageIDAndContinue(messageId)
-                    }
+            
+            self?.markMessage(withID: messageId, as: .syncing)
+            self?.apiClient.request(endpoint: .editMessage(payload: requestBody)) { result in
+                self?.backgroundContext.perform {
+                    self?.markMessage(withID: messageId, as: result.error == nil ? nil : .syncingFailed)
```
b-onc, line 75: `markMessage` calls `database.write` (so switches queues) and `removeMessageIDAndContinue` calls `processNextMessage` which in turn calls `backgroundContext.perform`, so do these 2 should be called in another `backgroundContext.perform`?

Sources_v3/Workers/Background/MessageEditor.swift
```diff
@@ -57,58 +54,43 @@ class MessageEditor<ExtraData: ExtraDataTypes>: Worker {
...
-                case .success:
-                    self.markMessage(withID: messageId, as: nil) {
-                        self.removeMessageIDAndContinue(messageId)
-                    }
-                case .failure:
-                    self.markMessage(withID: messageId, as: .syncingFailed) {
-                        self.removeMessageIDAndContinue(messageId)
-                    }
+            
+            self?.markMessage(withID: messageId, as: .syncing)
```
evsaev, line 72: There are plenty of `self?` statements but that's what we expect, I think. 

If we say `guard let self = self else { return }` in the request completion, the worker's lifetime can exceed the current user's session and some unexpected database write operations can happen before it finally get's deallocated

Sources_v3/Workers/Background/MessageEditor.swift
```diff
@@ -57,58 +54,43 @@ class MessageEditor<ExtraData: ExtraDataTypes>: Worker {
...
-        }, completion: {
-            if let error = $0 {
-                log.error("Error changing localMessageState for message with id \(id) to `\(String(describing: state))`: \(error)")
-            }
-            self.database.backgroundReadOnlyContext.perform {
-                completion()
-            }
-        })
+    private func markMessage(withID id: MessageId, as state: LocalMessageState?) {
+        backgroundContext.message(id: id)?.localMessageState = state
```
b-onc, line 88: We're using `backgroundReadOnlyContext` to mutate an object.. shouldn't we use `writableContext`?

Sources_v3/Workers/Background/MessageEditor.swift
b-onc: In the same light, if we don't hold a strong ref to self here (and always use `self?`), worker can be deallocated during the middle of processing, so after marking a message as syncing (line 72) but before being able to mark it as synced?

Sources_v3/Workers/Background/MessageSender.swift
```diff
@@ -111,7 +117,9 @@ private struct MessageSendingQueue<ExtraData: ExtraDataTypes> {
         let wasEmpty = self.requests.isEmpty
         requests.forEach { self.requests.insert($0) }
         if wasEmpty {
-            sendNextMessage()
+            dispatchQueue.async { [weak self] in
+                self?.sendNextMessage()
```
b-onc, line 121: Why don't we wrap `sendNextMessage` in `dispatchQueue.sync` so it's safe to call from everywhere?

b-onc (commented):
Loving how easy it is to add tests for memory leaks!

Sources_v3/Workers/Background/MessageEditor.swift
evsaev: If we call `setUser` with the same user-id as the current one, we don't need to flush the database nor recreate background workers -> workers will not be deallocated and messages will be updated.

If we change the user, the database will be flushed -> there will be no messages from the previous user -> we don't care about message local states

Sources_v3/Workers/Background/MessageEditor.swift
evsaev: `markMessage` doesn't access `database.write` but work with a `backgroundContext` directly so we don't switch the queues.

Regarding the safety of nested `perform()` calls, I think it's safe as it should just schedule the block execution on the context's private queue (but it's not mentioned directly in the docs as it is mentioned for `performAndWait`: **May safely be called reentrantly**) 😄

Sources_v3/Workers/Background/MessageEditor.swift
evsaev: that's a very good point 🤔 But this breaks the linear flow 😄 will  update the implementation to use `writableContext`

Sources_v3/Workers/Background/MessageSender.swift
evsaev: yep, that's a good point. The same was done in **MessageEditor** but with `perform`.
But I'd use `dispatchQueue.async` instead of `sync` not to block the calling thread

Sources_v3/Workers/Background/MessageEditor.swift
evsaev: fixed in ded903840ae45a33977d9b77006517eb03188a57

Sources_v3/Workers/Background/MessageSender.swift
evsaev: fixed in f606817667ea1950595a786d1ff86ef4c5e4d197

Sources_v3/Workers/Background/MessageSender.swift
```diff
@@ -117,70 +117,70 @@ private class MessageSendingQueue<ExtraData: ExtraDataTypes> {
...
-                self?.markMessageAsFailedToSend(id: request.messageId) {
-                    self?.removeRequestAndContinue(request)
-                }
-                
+            // Check the message with the given id is still in the DB.
+            guard let dto = self?.database.backgroundReadOnlyContext.message(id: request.messageId) else {
+                log.error("Trying to send a message with id \(request.messageId) but the message was deleted.")
+                self?.removeRequestAndContinue(request)
                 return
             }
```
VojtaStavik, line 135: The DTO must be obtained inside the `backgroundReadOnlyContext.peform { }` otherwise you can't safely retain it and work with it lately. It can be modified in the meantime from another queue.

VojtaStavik (changes requested):
Thanks for doing this, many good fixes!

Sources_v3/Workers/Background/MessageSender.swift
evsaev: yes, but this is in the scope of `CIS-333`. The motivation for fixing it in `MessageEditor` is here https://github.com/GetStream/stream-chat-swift/pull/487#discussion_r486959623

VojtaStavik: Right, sorry for the noise :)

VojtaStavik (approved):

b-onc (approved):

PR #482, opened 2020-09-10
[CIS-288] Calculate channel unread count based on latest channel read

b-onc:
The missing piece is, parsing `messageRead` events and updating DB's `ChannelRead` objects accordingly. I believe an event observer will be used for this purpose. Should it be done in this PR?

PR #490, opened 2020-09-14
[CIS-333] Fix accessing DTOs without `backgroundReadOnlyContext.perform`

b-onc:
..Also adds the middleware to be used in the client, which I forgot in #482 

Let's see how the CI reacts

### DIFF
--- a/Sources_v3/ChatClient.swift
+++ b/Sources_v3/ChatClient.swift
@@ -66,7 +66,8 @@ public class Client<ExtraData: ExtraDataTypes> {
         EventDataProcessorMiddleware<ExtraData>(database: databaseContainer),
         TypingStartCleanupMiddleware<ExtraData>(
             excludedUserIds: { [weak self] in Set([self?.currentUserId].compactMap { $0 }) }
-        )
+        ),
+        ChannelReadUpdaterMiddleware<ExtraData>(database: databaseContainer)
     ])
     
     /// The `APIClient` instance `Client` uses to communicate with Stream REST API.
--- a/Sources_v3/ChatClient_Tests.swift
+++ b/Sources_v3/ChatClient_Tests.swift
@@ -143,6 +143,26 @@ class ChatClient_Tests: StressTestCase {
         XCTAssert(webSocket?.init_requestEncoder.connectionDetailsProviderDelegate === client)
     }
     
+    func test_webSocketClient_hasAllMandatoryMiddlewares() throws {
+        // Use in-memory store
+        // Create a new chat client, which in turn creates a webSocketClient
+        _ = ChatClient(
+            config: inMemoryStorageConfig,
+            workerBuilders: workerBuilders,
+            environment: testEnv.environment
+        )
+        
+        // Assert that mandatory middlewares exists
+        let middlewares = try XCTUnwrap(testEnv.webSocketClient?.init_eventNotificationCenter.middlewares)
+        
+        // Assert `EventDataProcessorMiddleware` exists
+        XCTAssert(middlewares.contains(where: { $0 is EventDataProcessorMiddleware<DefaultDataTypes> }))
+        // Assert `TypingStartCleanupMiddleware` exists
+        XCTAssert(middlewares.contains(where: { $0 is TypingStartCleanupMiddleware<DefaultDataTypes> }))
+        // Assert `ChannelReadUpdaterMiddleware` exists
+        XCTAssert(middlewares.contains(where: { $0 is ChannelReadUpdaterMiddleware<DefaultDataTypes> }))
+    }
+    
     func test_connectionStatus_isExposed() {
         // Create a new chat client
         let client = ChatClient(
--- a/Sources_v3/Database/DatabaseContainer.swift
+++ b/Sources_v3/Database/DatabaseContainer.swift
@@ -16,6 +16,9 @@ class DatabaseContainer: NSPersistentContainer {
         case onDisk(databaseFileURL: URL)
     }
     
+    /// We use `writableContext` for having just one place to save changes
+    /// so it’s not possible to have conflicts when saving payloads from various sources.
+    /// All writes are happening serially using this context and its `write { }` methods.
     lazy var writableContext: NSManagedObjectContext = {
         let context = newBackgroundContext()
         context.automaticallyMergesChangesFromParent = true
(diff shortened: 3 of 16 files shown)


Sources_v3/ChatClient.swift
```diff
@@ -66,7 +66,8 @@ public class Client<ExtraData: ExtraDataTypes> {
         EventDataProcessorMiddleware<ExtraData>(database: databaseContainer),
         TypingStartCleanupMiddleware<ExtraData>(
             excludedUserIds: { [weak self] in Set([self?.currentUserId].compactMap { $0 }) }
-        )
+        ),
+        ChannelReadUpdaterMiddleware<ExtraData>(database: databaseContainer)
```
VojtaStavik, line 70: We have tests for the mandatory middlewares, too

VojtaStavik (approved):

Sources_v3/Database/DatabaseContainer.swift
```diff
@@ -16,13 +16,24 @@ class DatabaseContainer: NSPersistentContainer {
         case onDisk(databaseFileURL: URL)
     }
     
+    /// We use `writableContext` for having just one place to save changes
+    /// so it’s not possible to have conflicts when saving payloads from various sources.
+    /// All writes are happening serially using this context and its `write { }` methods.
```
VojtaStavik, line 21: ... off topic: when writing this, I felt "seriously" wasn't the right word. "Serially" was the word :D

Sources_v3/ChatClient_Tests.swift
```diff
@@ -143,6 +143,32 @@ class ChatClient_Tests: StressTestCase {
...
+            config: inMemoryStorageConfig,
+            workerBuilders: workerBuilders,
+            environment: testEnv.environment
+        )
+        
+        // Assert that mandatory middlewares exists
+        let middlewares = try XCTUnwrap(testEnv.webSocketClient?.init_eventNotificationCenter.middlewares)
+        // We currently have 4 mandatory middlewares, update this if you add more
+        // in `Client.eventNotificationCenter` lazy variable
+        // (3 of them added in `eventNotificationCenter` init, 1 added in `WebSocketClient` init)
```
VojtaStavik, line 159: I wouldn't assert on the count in this case. The WS-init middleware is already tested. This should ideally cover only the middlewares set up by Client

Sources_v3/ChatClient_Tests.swift
```diff
@@ -143,6 +143,32 @@ class ChatClient_Tests: StressTestCase {
...
+        XCTAssertEqual(middlewares.count, 4)
+        
+        // Assert `HealthCheckMiddleware` exists
+        XCTAssert(middlewares.contains(where: { $0 is HealthCheckMiddleware }))
+        // Assert `EventDataProcessorMiddleware` exists
+        XCTAssert(middlewares.contains(where: { $0 is EventDataProcessorMiddleware<DefaultDataTypes> }))
+        // Assert `TypingStartCleanupMiddleware` exists
+        XCTAssert(middlewares.contains(where: { $0 is TypingStartCleanupMiddleware<DefaultDataTypes> }))
+        // Assert `ChannelReadUpdaterMiddleware` exists
+        XCTAssert(middlewares.contains(where: { $0 is ChannelReadUpdaterMiddleware<DefaultDataTypes> }))
```
VojtaStavik, line 169: This is enough for the test imo

Sources_v3/Workers/Background/MessageSender.swift
```diff
@@ -111,70 +111,74 @@ private class MessageSendingQueue<ExtraData: ExtraDataTypes> {
     /// multiple times without having to worry about that.
     @Atomic private(set) var requests: Set<SendRequest> = []
     
-    /// Schedules sending of the message. All already scheduled messages with `createdLocallyAt` older that this one will
+    /// Schedules sending of the message. All already scheduled messages with `createdLocallyAt` older than these ones will
```
VojtaStavik, line 114: 👍

VojtaStavik (commented):
LGTM

Sources_v3/ChatClient_Tests.swift
b-onc: But if we don't assert the count, newly added middlewares won't be caught (just as I've added ReadUpdater middleware but forgot to add it to the list) ?

Sources_v3/Database/DatabaseContainer.swift
b-onc: I imagined some angry middle aged worker, named "Context", writing entities into a big excel file with a very serious attitude on his face

Sources_v3/ChatClient_Tests.swift
VojtaStavik: I see your point. OK then :) I'd adjust the comment to exactly what you just said to me + I'd remove the `HealthCheckMiddleware` check here

b-onc: @VojtaStavik on a second thought, you're right. Asserting count is not needed, since if someone forgot to add the middleware, count will still hold and test will not warn them. Only way would be to check conformers of `EventMiddleware` protocol and make sure each conformer is added to the list of middlewares, which might be a bit overkill

Sources_v3/Workers/Background/MissingEventsPublisher.swift
```diff
@@ -67,7 +67,7 @@ class MissingEventsPublisher<ExtraData: ExtraDataTypes>: Worker {
                 cids: watchedChannelIDs
             )
             
-            self?.apiClient.request(endpoint: endpoint) { [weak self] in
+            self?.apiClient.request(endpoint: endpoint) {
```
VojtaStavik: You removed this because `self` is already `weak` or was there another reason?

Sources_v3/Workers/Background/MissingEventsPublisher_Tests.swift
```diff
@@ -100,6 +101,9 @@ final class MissingEventsPublisher_Tests: StressTestCase {
             currentUser.lastReceivedEventDate = Date()
         }
         
+        // Simulate error response when request comes
+        apiClient.test_simulateResponseWhenRequestComes(Result<MissingEventsPayload<ExtraData>, Error>.failure(TestError()))
+
```
VojtaStavik, line 106: Can you explain this change?

Sources_v3/Workers/Background/MissingEventsPublisher.swift
b-onc: Exactly that reason

Sources_v3/Workers/Background/MissingEventsPublisher_Tests.swift
b-onc: Since everything basically happens inside the `database.backgroundReadOnlyContext.perform` block async, the request does not arrive sync, so `apiClient.test_simulateResponse` fails to simulate the response since there's no request at that point. We needed a way to be able to provide the response when the request arrives, so I added this func. It's in the commit 0e50686d7ef5a6de1cd36016da99137c296b9d36

VojtaStavik: Another option would be to add something like:
```
// Assert a network request is created
AssertAsync.willBeEqual(apiClient.request_allRecordedCalls.count, 1)
```
I'm not against your solution; just want to offer an alternative one within the existing infrasturcture.

b-onc: @VojtaStavik I like that one better so I've refactored to that in 13db1038f142aa0d2913b634f9e1fe964bd055bb

PR #491, opened 2020-09-14
Sample v3: fix settings screen crash

cardoso:
Settings screen was crashing on SwiftUI and combine samples.


VojtaStavik (approved):
What was the reason for the crash? It's not clear from the changes nor the description of the PR.

cardoso:
@VojtaStavik oops. Sorry. The crashes were because the currentUserController wasn't being passed before displaying the settings view controller after we changed from global ChatClient reference to dependency injection of the controllers.

PR #494, opened 2020-09-17
[CIS-194] Logger improvements

mantergo:
**In this PR:**

- Add `assertationFailure` to `Logger`
- Revise log leves in `Logger`
- Pretty print JSON responses
- Move logging to custom queue
- Introduce `PrettyPrintable` protocol

`struct EventPayload<ExtraData: ExtraDataTypes>: Decodable {`: 
before

`struct EventPayload<ExtraData: ExtraDataTypes>: Decodable, PrettyPrintable {`: 
after

### DIFF
--- a/Sample_v3/AppDelegate.swift
+++ b/Sample_v3/AppDelegate.swift
@@ -22,7 +22,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         LogConfig.showDate = false
         LogConfig.showFunctionName = false
         
-        LogConfig.level = .info
+        LogConfig.level = .debug
         
         LogStore.registerShared()
         return true
--- a/Sources_v3/APIClient/RequestDecoder.swift
+++ b/Sources_v3/APIClient/RequestDecoder.swift
@@ -43,8 +43,7 @@ struct DefaultRequestDecoder: RequestDecoder {
             throw ClientError.ResponseBodyEmpty()
         }
         
-        // TODO: improve
-        log.info("URL request response: \(httpResponse), data: \(String(data: data, encoding: .utf8) ?? data.description))")
+        log.info("URL request response: \(httpResponse), data: \(data))")
         
         guard httpResponse.statusCode < 400 else {
             guard let serverError = try? JSONDecoder.default.decode(ErrorPayload.self, from: data) else {
--- a/Sources_v3/ChatClient.swift
+++ b/Sources_v3/ChatClient.swift
@@ -128,7 +128,7 @@ public class Client<ExtraData: ExtraDataTypes> {
             }
             
         } catch is ClientError.MissingLocalStorageURL {
-            log.assert(false, "The URL provided in ChatClientConfig can't be `nil`. Falling back to the in-memory option.")
+            log.assertationFailure("The URL provided in ChatClientConfig can't be `nil`. Falling back to the in-memory option.")
             
         } catch {
             log.error("Failed to initalized the local storage with error: \(error). Falling back to the in-memory option.")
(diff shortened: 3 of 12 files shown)


b-onc (commented):
Can we make it so `Data` is pretty-printed on the logger level, without needing custom protocols or referring to them as `.prettyPrinted`? Basically, pasing any valid data/json to the logger should support pretty-printing

mantergo:
> Can we make it so `Data` is pretty-printed on the logger level, without needing custom protocols or referring to them as `.prettyPrinted`? Basically, pasing any valid data/json to the logger should support pretty-printing

Yep, that would be ideal.
We are passing message to `Logger` as `Any` and in case of passing `Data` or `Dictionary` it's possible to do some down casting and pretty print it. But we passing it like `log.info("URL request response: \(httpResponse), data: \(data.prettyPrinted))")` so I'm afraid it's not possible.

VojtaStavik (approved):
LGTM! Why the change with the queue?

b-onc:
@mantergo If we add these custom string interpolations, leveraging swift's power, we're able to do it.
```swift
extension String.StringInterpolation {
    mutating func appendInterpolation(_ value: Data) {
        guard
            let object = try? JSONSerialization.jsonObject(with: value, options: []),
            let data = try? JSONSerialization.data(withJSONObject: object, options: [.prettyPrinted]),
            let prettyPrintedString = String(data: data, encoding: .utf8)
        else {
            appendLiteral(String(describing: value))
            return
        }
        appendLiteral(prettyPrintedString)
    }
}

extension String.StringInterpolation {
    mutating func appendInterpolation<T>(_ value: EventPayload<T>) {
        var description = "-----\(type(of: self))-----\n"
        let mirror = Mirror(reflecting: self)
        for child in mirror.children {
            if let propertyName = child.label {
                description += "    -\(propertyName): \(child.value)\n"
            }
        }
        appendLiteral(description)
    }
}
```

We can not-omit the parameter name too, if we want:
```swift
log.info("URL request response: \(httpResponse), data: \(prettyPrinted: data))")
```

Beauty of this is, we can add as many custom interpolations as we want, without needing to change the original code (such as adding protocol conformance)

mantergo:
> @mantergo If we add these custom string interpolations, leveraging swift's power, we're able to do it.
> 
> ```swift
> extension String.StringInterpolation {
>     mutating func appendInterpolation(_ value: Data) {
>         guard
>             let object = try? JSONSerialization.jsonObject(with: value, options: []),
>             let data = try? JSONSerialization.data(withJSONObject: object, options: [.prettyPrinted]),
>             let prettyPrintedString = String(data: data, encoding: .utf8)
>         else {
>             appendLiteral(String(describing: value))
>             return
>         }
>         appendLiteral(prettyPrintedString)
>     }
> }
> 
> // Haven't tested this one, not sure if works with generics
> extension String.StringInterpolation {
>     mutating func appendInterpolation<T>(_ value: EventPayload<T>) {
>         var description = "-----\(type(of: self))-----\n"
>         let mirror = Mirror(reflecting: self)
>         for child in mirror.children {
>             if let propertyName = child.label {
>                 description += "    -\(propertyName): \(child.value)\n"
>             }
>         }
>         appendLiteral(description)
>     }
> }
> ```
> 
> We can not-omit the parameter name too, if we want:
> 
> ```swift
> log.info("URL request response: \(httpResponse), data: \(prettyPrinted: data))")
> ```
> 
> Beauty of this is, we can add as many custom interpolations as we want, without needing to change the original code (such as adding protocol conformance)

Nice! I like it! 
In this case we just need to pay attention that we always using string interpolation, but seems like we doing it always anyway.

b-onc (approved):

PR #497, opened 2020-09-18
[CIS-300] Fix deletion events delivery

evsaev:


Sources_v3/Database/DatabaseContainer.swift
```diff
@@ -156,25 +156,13 @@ class DatabaseContainer: NSPersistentContainer {
...
-                if persistentStoreDescriptions.contains(where: { $0.type == NSInMemoryStoreType }) {
-                    // If we use `NSInMemoryStoreType` we can't use `NSBatchDeleteRequest` and we have to delete
-                    // the objects one by one.
-                    let objects = try session.fetch(fetchRequest) as? [NSManagedObject]
-                    objects?.forEach {
-                        session.delete($0)
-                    }
-                    
-                } else {
-                    let deleteRequest = NSBatchDeleteRequest(fetchRequest: fetchRequest)
```
evsaev, line 175: When using `NSBatchDeleteRequest` FRC doesn't notify the delegate about deletion events. So in both cases, we iterate over all the objects -> the check for `store type` is no longer needed

Sources_v3/Controllers/ListDatabaseObserver.swift
```diff
@@ -186,6 +199,50 @@ class ListChangeAggregator<DTO: NSManagedObject, Item>: NSObject, NSFetchedResul
...
+        deletedItems.removeAll()
+        (deletedDTOs + missmatchingDTOs).forEach { dtoToDelete in
+            guard let trackedDTO = fetchedDTOs.first(where: { $0.objectID == dtoToDelete.objectID }) else { return }
+            
+            // Create item from DTO in advance before the actual deletion happen.
+            // This is required because after the deletion, entity relations become invalid
+            // and lossless `DTO -> Item` converion become impossible
+            deletedItems[trackedDTO.objectID] = itemCreator(trackedDTO)
+        }
+    }
```
evsaev, line 245: The problem I'm solving here is that the entity relationships become invalid after the `DTO -> Item` transformation is not possible.

I don't like neither `No action` delete rule nor `N-step-database-flush` so come up with performing `DTO -> Item` transformation **before** the actual deletion happen.

VojtaStavik:
Implemented in https://github.com/GetStream/stream-chat-swift/pull/530

PR #502, opened 2020-09-21
Sample v3: Upgrade SwiftUI sample to iOS 14

cardoso:
Additional changes:
- Chat and composer will now adjust for keyboard (iOS 14 does this by default)
- Composer is now multiline and autoexpands (using TextEditor + Text ZStack workaround)
- Can now specify user id to add / remove in a channel (wrapping the existing UIKit alert extension)


VojtaStavik (approved):
👍

PR #509, opened 2020-09-23
[CIS-361] Fix memberCount and watcherCount for ChatChannel

mantergo:
**In this PR:**

- Fix `memberCount` and `watcherCount` for `ChatChannel`


VojtaStavik (approved):
LGTM 👍 Can you please also add tests for these fields? `Payload` -> `DTO` -> `Model`

PR #511, opened 2020-09-24
[CIS-209] Bump `Startscream` dependency

evsaev:
**This PR:** 

- bumps `Startscream` dependency from `~> 3.1` to `~> 4.0` for both **v2** and **v3** SDK versions for all dependency managers (Carthage/CocoaPods/SPM)
- updates **v2** and **v3** SDK versions to meet the new API

### DIFF
--- a/Sources/Client/WebSocket/Provider/StarscreamWebSocketProvider.swift
+++ b/Sources/Client/WebSocket/Provider/StarscreamWebSocketProvider.swift
@@ -20,7 +20,7 @@ final class StarscreamWebSocketProvider: WebSocketProvider {
         }
     }
     
-    var isConnected: Bool { webSocket.isConnected }
+    private(set) var isConnected = false
     var callbackQueue: DispatchQueue { webSocket.callbackQueue }
     weak var delegate: WebSocketProviderDelegate?
     
@@ -35,7 +35,7 @@ final class StarscreamWebSocketProvider: WebSocketProvider {
     }
     
     func disconnect() {
-        webSocket.disconnect(forceTimeout: 0)
+        webSocket.disconnect()
     }
     
     func sendPing() {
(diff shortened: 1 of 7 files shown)


Sources/Client/WebSocket/Provider/StarscreamWebSocketProvider.swift
```diff
@@ -20,7 +20,7 @@ final class StarscreamWebSocketProvider: WebSocketProvider {
         }
     }
     
-    var isConnected: Bool { webSocket.isConnected }
```
evsaev, line 23: Starting from `4.0.0` Starscream no longer has such API. Implemented it in the same way it is done in `URLSessionWebSocketProvider`

Sources/Client/WebSocket/Provider/StarscreamWebSocketProvider.swift
```diff
@@ -35,7 +35,7 @@ final class StarscreamWebSocketProvider: WebSocketProvider {
     }
     
     func disconnect() {
-        webSocket.disconnect(forceTimeout: 0)
+        webSocket.disconnect()
```
evsaev, line 38: The function we use here is `disconnect(closeCode: UInt16 = CloseCode.normal.rawValue)` but as you see it takes a `closeCode`. Do we need to update `WebSocketProvider.disconnect()` to require this value as an input? (maybe wrapped into some enum)

Sources_v3/WebSocketClient/Engine/StarscreamWebSocketEngine.swift
```diff
@@ -46,19 +46,28 @@ import Foundation
...
+                let error = WebSocketEngineError(reason: reason, code: Int(code), engineError: nil)
+                delegate?.webSocketDidDisconnect(error: error)
+            case .cancelled:
+                let error = WebSocketEngineError(
+                    reason: "Cancelled",
+                    code: WebSocketEngineError.stopErrorCode,
+                    engineError: nil
+                )
+                delegate?.webSocketDidDisconnect(error: error)
+            default:
```
evsaev, line 67: Here is the enum so you can see which cases trap into `default`:
```swift
public enum WebSocketEvent {
    case connected([String: String])
    case disconnected(String, UInt16)
    case text(String)
    case binary(Data)
    case pong(Data?)
    case ping(Data?)
    case error(Error?)
    case viabilityChanged(Bool)
    case reconnectSuggested(Bool)
    case cancelled
}
```
There is a `.reconnectSuggested` event, maybe we want to use it somehow instead/together with `WebSocketClientReconnectionStrategy`?

Sources_v3/WebSocketClient/Engine/StarscreamWebSocketEngine.swift
```diff
@@ -46,19 +46,28 @@ import Foundation
...
+                delegate?.webSocketDidReceiveMessage(text)
+            case let .error(error):
+                delegate?.webSocketDidDisconnect(error: error.map(WebSocketEngineError.init))
+            case let .disconnected(reason, code):
+                let error = WebSocketEngineError(reason: reason, code: Int(code), engineError: nil)
+                delegate?.webSocketDidDisconnect(error: error)
+            case .cancelled:
+                let error = WebSocketEngineError(
+                    reason: "Cancelled",
+                    code: WebSocketEngineError.stopErrorCode,
```
evsaev, line 63: I'm not sure `WebSocketEngineError.stopErrorCode` suites here 🤔  We might want some other error code and handle it a different way

VojtaStavik: Hmm, I wouldn't use stop code here 🤔 It's usually used by the backend signalizing that reconnecting is not suitable.
Just put `-1` there probably :)

Sources_v3/WebSocketClient/Engine/StarscreamWebSocketEngine.swift
VojtaStavik: Interesting. Did you test for example the `pong` case? Is it fired with the `text` case, too? In other words - is `WebSocketClient` still receiving pong events even via `text`?

Sources/Client/WebSocket/Provider/StarscreamWebSocketProvider.swift
VojtaStavik: I think it's fine as is.

VojtaStavik (commented):
Looks good 👍 Can you please verify both the v2 is still working correctly? Especially if pong events are received.

Sources_v3/WebSocketClient/Engine/StarscreamWebSocketEngine.swift
evsaev: Checked it. The `.pong` event doesn't come with `.text`

Sources_v3/WebSocketClient/Engine/StarscreamWebSocketEngine.swift
evsaev: fixed for both v2 and v3

Sources_v3/WebSocketClient/Engine/StarscreamWebSocketEngine.swift
evsaev: What we treat as **pong** is a `HealthCheckEvent` event and we receive it with `.text` event. So the `WebSocketClient` does receive **pong** events.

Do you think we should introduce `WebSocketEngineDelegate.webSocketDidReceivePong()`, invoke it when `.pong` event comes and create `HealthCheckEvent` manually with existed connection id?

VojtaStavik: OK good, I understood it differently from your message. The important thing is `WebSocketClient` receives the health check events via `.text` 👍

VojtaStavik:
@evsaev Please wait with merging this after we release the path today. We need more time to properly test this.

Sources_v3/WebSocketClient/Engine/WebSocketEngine.swift
```diff
@@ -24,7 +24,8 @@ protocol WebSocketEngineDelegate: AnyObject {
 
 struct WebSocketEngineError: Error {
     static let stopErrorCode = 1000
-    
+    static let cancelErrorCode = -1
+
```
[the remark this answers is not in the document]
evsaev: fixed 😄

VojtaStavik:
OK, let's merge this and see what breaks :)

VojtaStavik (approved):

PR #513, opened 2020-09-25
Sample v3: Improve login experience

cardoso:
- Added configuration section with storage and region options
- Improve layout (separate low-level samples from high-level samples placeholders)
- Fix login being called when pressing cells other than samples
- Fix crash on login error

Also small change in SDK:
- Don't do anything when flush is enabled but local storage is not (flushing without local storage causes crash)

Screen Shot 2020-09-25 at 11 12 42

#no_changelog

### DIFF
--- a/Sources_v3/Database/DatabaseContainer.swift
+++ b/Sources_v3/Database/DatabaseContainer.swift
@@ -83,7 +83,7 @@ class DatabaseContainer: NSPersistentContainer {
         
         persistentStoreDescriptions = [description]
                 
-        if shouldFlushOnStart {
+        if shouldFlushOnStart, kind != .inMemory {
             try recreatePersistentStore()
         } else {
             try setupPersistentStore()
--- a/Sources_v3/Database/DatabaseContainer.swift
+++ b/Sources_v3/Database/DatabaseContainer.swift
@@ -202,7 +202,7 @@ class DatabaseContainer: NSPersistentContainer {
     }
     
     /// Recreate rersistent store
-    private func recreatePersistentStore() throws {
+    func recreatePersistentStore() throws {
         guard let url = persistentStoreDescriptions.first!.url
         else { throw ClientError("Internal Error. Wrong database file url.") }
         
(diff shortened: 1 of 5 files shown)


Sources_v3/Database/DatabaseContainer.swift
```diff
@@ -82,7 +82,7 @@ class DatabaseContainer: NSPersistentContainer {
         
         persistentStoreDescriptions = [description]
                 
-        if shouldFlushOnStart {
+        if shouldFlushOnStart, kind != .inMemory {
```
VojtaStavik, line 85: This needs to be tested. Can you take care of that @mantergo and push a commit with the missing test, please?

VojtaStavik (commented):
Looks good 👍 I like the changes to the login screen. Just a couple of thoughts and questions:

  - We should decide, whether the sample app should be a test app for the developers to use and test, or a canonical example of how to use our SDK. If it's the latter, we probably don't need to expose the settings since they are not meant to be exposed?

 - Low Level Client in an internal name for the SDK and we probably shouldn't  mention it anywhere publicly. The official names are `StreamChat` and `StreamChat UI`.

 - I'm a little bit afraid that the examples themselves are getting lost in all the settings. Maybe we should change it? Having all the settings on a different screen (or at least at the bottom of this screen?) and have the examples listed as the main thing we want to sell?

Sources_v3/Database/DatabaseContainer.swift
mantergo: Thanks for pinging, added test 🙂

VojtaStavik:
@cardoso We just merged a PR that adds more locations, can you extend the list with Singapore and Sydney?

cardoso:
@VojtaStavik all done.

About whether this is a test or canonical sample: I think it should be both, it's just too useful to have a configurable app to test all kinds of settings and reproduce certain use cases quickly. It already isn't 100% canonical because it has multiple sample inside.

But I can explore hiding some of the test stuff in other PRs. Probably move the credentials and configuration stuff to a new modal screen.

VojtaStavik (approved):
Perfect, thank you 👍 Can you please remove the (Low-Lever Client) mention from the sample names, too? Feel free to merge afterwards.

cardoso:
@VojtaStavik oops. Missed those. Done 👍

PR #518, opened 2020-09-28
Add custom environment variable usage in SwiftUI sample

mantergo:
The goal was to use `@Environment` in our sample app to make it more `SwiftUI`-ish.

The problem there is that we need to initialize our `.ObservableObject` using `chatClient`.
It seems like right now the only way to achieve that is to introduce another level of `View` abstraction. (nil, lazy etc. didn't work)
As soon as it is Sample app I'm not sure we should do that.

Another issue is that we can't introduce some valid `defaultValue`. 
Possible solution was to throw `fatalError()` on attempt of accessing `defaultValue` but that is also not possible cause `defaultValue` is accessed each time we initialize some view even if we provide this value with `.environment` `ViewModifier`.

### DIFF
--- a/Sample_v3/LoginViewController.swift
+++ b/Sample_v3/LoginViewController.swift
@@ -184,7 +184,8 @@ extension LoginViewController {
                     self.view.window?.rootViewController = UIHostingController(
                         rootView:
                         NavigationView {
-                            ChannelListView(channelList: channelListController.observableObject)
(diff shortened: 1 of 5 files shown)


VojtaStavik:
Interesting findings! I'd suggest closing this for now and work on it again once we have feedback from users on _how_ they want to use it. Feel free to close this PR for now @mantergo

PR #519, opened 2020-09-30
[CIS-370] Call `setUser` completion with error when WS connection fails

VojtaStavik:
The original implementation tried to reconnect the web socket when the connection failed with a 400...499 status code.

The new implementation doesn't try to reconnect in those cases, and the completion of `set(Guest/Anonymous)User` is called with an error instead.

---

@cardoso It seems the sample app doesn't handle this case properly in `setUserCompletion`. Would you mind taking a look and fixing it, please? Feel free to push more commits to this PR directly.

### DIFF
--- a/Sources_v3/ChatClient_Tests.swift
+++ b/Sources_v3/ChatClient_Tests.swift
@@ -235,6 +235,36 @@ class ChatClient_Tests: StressTestCase {
         AssertAsync.staysTrue(providedConnectionId == nil)
     }
     
+    func test_client_failsConnectionIdWaiters_whenWebSocketIsDisconnected() {
+        // Create a new chat client
+        let client = ChatClient(
+            config: inMemoryStorageConfig,
+            workerBuilders: workerBuilders,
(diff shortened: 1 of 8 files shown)


Sources_v3/ChatClient_Tests.swift
```diff
@@ -235,6 +235,38 @@ class ChatClient_Tests: StressTestCase {
...
+        // Simulate WebSocketConnection change to "disconnected"
+        let error = ClientError(with: TestError())
+        testEnv.webSocketClient?.connectionStateDelegate?
+            .webSocketClient(testEnv.webSocketClient!, didUpdateConectionState: .disconnected(error: error))
+        
+        // Assert the provided connection id is `nil`
+        XCTAssertNil(providedConnectionId)
+        
+        // Verify the callback is not called multiple times
+
```
mantergo, line 262: nit: something lost here

mantergo (approved):

b-onc (approved):

PR #520, opened 2020-09-30
[CIS-364] Use `defaultValue` when extra decoding fails

VojtaStavik:
This PR changes the extra decoding logic to be more robust and nicely recover by using `defaultValue` if extra data decoding fails.

### DIFF
--- a/Sources_v3/Database/DTOs/UserDTO_Tests.swift
+++ b/Sources_v3/Database/DTOs/UserDTO_Tests.swift
@@ -82,6 +82,33 @@ class UserDTO_Tests: XCTestCase {
         }
     }
     
+    func test_defaultExtraDataIsUsed_whenExtraDataDecodingFails() throws {
+        let userId: UserId = .unique
+        
+        let payload: UserPayload<DefaultExtraData.User> = .init(
+            id: userId,
+            role: .admin,
+            createdAt: .unique,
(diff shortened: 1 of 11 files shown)


Sources_v3/Database/DTOs/UserDTO_Tests.swift
```diff
@@ -82,6 +82,33 @@ class UserDTO_Tests: XCTestCase {
...
+        )
+        
+        try database.writeSynchronously { session in
+            // Save the user
+            let userDTO = try! session.saveUser(payload: payload)
+            // Make the extra data JSON invalid
+            userDTO.extraData = #"{"invalid": json}"# .data(using: .utf8)!
+        }
+        
+        let loadedMessage: ChatUser? = database.viewContext.user(id: userId)?.asModel()
```
evsaev, line 108:
```suggestion
        let loadedUser: ChatUser? = database.viewContext.user(id: userId)?.asModel()
```

evsaev (approved):
👍

Sources_v3/Database/DTOs/UserDTO_Tests.swift
VojtaStavik: Fixed in 0ae98e4ceba2bd6e34439ba1f8b42d91c1b1fde4

PR #521, opened 2020-09-30
[CIS-178] Add queryMembers API

b-onc:
Docs also updated


Sources/Client/Client/Client+Channel.swift
```diff
@@ -533,6 +533,64 @@ public extension Client {
         return request(endpoint: .inviteAnswer(answer), completion)
     }
     
+    /// Query a channel's members.
```
VojtaStavik, line 536: nit: remove `a`?

Sources/Client/Model/Channel/Channel+Requests.swift
```diff
@@ -284,6 +284,37 @@ public extension Channel {
...
+    ///   - offset: Offset of pagination. Defaults to 0.
+    ///   - completion: Completion block with `MembersQueryResponse`
+    @discardableResult
+    func queryMembers(filter: Filter,
+                      sorting: [Sorting] = [],
+                      limit: Int = 100,
+                      offset: Int = 0,
+                      _ completion: @escaping Client.Completion<MembersQueryResponse>) -> Cancellable {
+        let query: MembersQuery
+        if !id.isEmpty {
```
VojtaStavik, line 301: nit: why do you use `if !isEmpty { } else { }` when you can use `if isEmpty { } esle { }`? :)

Sources/Core/Client/Channel/Channel+RxRequests.swift
```diff
@@ -225,6 +225,34 @@ public extension Reactive where Base == Channel {
...
+    ///   - filter: Filter conditions for query
+    ///   - sorting: Sorting conditions for query
+    ///   - limit: Limit for number of members to return. Defaults to 100.
+    ///   - offset: Offset of pagination. Defaults to 0.
+    func queryMembers(filter: Filter,
+                      sorting: [Sorting] = [],
+                      limit: Int = 100,
+                      offset: Int = 0) -> Observable<MembersQueryResponse> {
+        let query: MembersQuery
+        if base.id.isEmpty {
```
VojtaStavik, line 239: ... like you do it here :)

sample-bootstrap.sh
```diff
@@ -15,4 +15,4 @@ echo 'EXCLUDED_ARCHS__EFFECTIVE_PLATFORM_SUFFIX_simulator__NATIVE_ARCH_64_BIT_x8
 echo 'EXCLUDED_ARCHS = $(inherited) $(EXCLUDED_ARCHS__EFFECTIVE_PLATFORM_SUFFIX_$(EFFECTIVE_PLATFORM_SUFFIX)__NATIVE_ARCH_64_BIT_$(NATIVE_ARCH_64_BIT)__XCODE_$(XCODE_VERSION_MAJOR))' >> $xcconfig
 
 export XCODE_XCCONFIG_FILE="$xcconfig"
-carthage bootstrap
+carthage bootstrap --platform iOS
```
VojtaStavik, line 18: https://getstream.slack.com/archives/G019728A97W/p1601387865020100?thread_ts=1601376554.016900&cid=G019728A97W

VojtaStavik (approved):
Looks good 👍

sample-bootstrap.sh
b-onc: Yep I added it after the comment :D

Sources/Core/Client/Channel/Channel+RxRequests.swift
b-onc: That's actually a bug! It should also be `!base.id.isEmpty` but I'll reverse it

sample-bootstrap.sh
VojtaStavik: ?? But it shouldn't be there, the dependencies are used by both v2 and v3 samples

PR #523, opened 2020-09-30
[CIS-359] Change execution order for stress tests

evsaev:
This PR changes tests execution order Random -> Alphabetical for stress plan. The goal is to apply this change and see if it helps because debugging stress tests locally takes much time.


VojtaStavik (approved):
OK, but please change the name of the PR. This doesn't actually fix the problem :)

PR #525, opened 2020-10-01
[CIS-359] Ensure delegates always have expected-queue-id set in tests

evsaev:
This PR is another iteration of eliminating the failing stress tests issue. It updates `QueueAwareDelegate` to receive `expectedQueueId` via init other than having it as a mutable optional property. This ensures the `expectedQueueId` to always exist when any of the delegate method is called.


VojtaStavik (approved):
LGTM 👍

PR #526, opened 2020-10-01
[CIS-261] User muting

evsaev:
**This PR:**
- introduces `ChatUserController` that provided the API for performing user-related actions;
- updates `CurrentUserDatabaseSession` to take `mutedUsers` into account;
- introduces `UserUpdater` that sends API requests representing user-related actions;

### DIFF
--- a/Sources_v3/Query/UserListQuery.swift
+++ b/Sources_v3/Query/UserListQuery.swift
@@ -51,3 +51,12 @@ public struct UserListQuery: Encodable {
         try options.encode(to: encoder)
     }
 }
+
+extension UserListQuery {
+    /// Builds `UserListQuery` for a user with the provided `userId`
+    /// - Parameter userId: The user identifier
+    /// - Returns: `UserListQuery` for a specific user
+    static func user(withID userId: UserId) -> Self {
+        .init(filter: .equal("id", to: userId))
+    }
+}
--- a/Sources_v3/Workers/UserUpdater.swift
+++ b/Sources_v3/Workers/UserUpdater.swift
@@ -1,0 +1,75 @@
+//
+// Copyright © 2020 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+/// Makes user-related calls to the backend and updates the local storage with the results.
+class UserUpdater<ExtraData: ExtraDataTypes>: Worker {
+    /// Mutes the user with the provided `userId`.
+    /// - Parameters:
+    ///   - userId: The user identifier.
+    ///   - completion: Called when the API call is finished. Called with `Error` if the remote update fails.
+    ///
+    func muteUser(_ userId: UserId, completion: ((Error?) -> Void)? = nil) {
+        apiClient.request(endpoint: .muteUser(userId)) {
+            completion?($0.error)
+        }
+    }
+    
+    /// Unmutes the user with the provided `userId`.
+    /// - Parameters:
+    ///   - userId: The user identifier.
+    ///   - completion: Called when the API call is finished. Called with `Error` if the remote update fails.
+    ///
+    func unmuteUser(_ userId: UserId, completion: ((Error?) -> Void)? = nil) {
+        apiClient.request(endpoint: .unmuteUser(userId)) {
+            completion?($0.error)
+        }
+    }
+    
+    /// Makes a single user query call to the backend and updates the local storage with the results.
+    ///
+    /// - Parameters:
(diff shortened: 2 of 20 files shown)


VojtaStavik (commented):
Seems to me to be the right direction 👍

Tests_v3/Dummy data/MutedUser.swift
```diff
@@ -0,0 +1,31 @@
+//
+// Copyright © 2020 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+@testable import StreamChatClient
+
+extension MutedUser {
```
VojtaStavik, line 8: `MutedUser` should be probably renamed to `MutedUserPayload`

Sources_v3/Query/UserListQuery.swift
```diff
@@ -51,3 +51,12 @@ public struct UserListQuery: Encodable {
...
     }
 }
+
+extension UserListQuery {
+    /// Builds `UserListQuery` for a user with the provided `userId`
+    /// - Parameter userId: The user identifier
+    /// - Returns: `UserListQuery` for a specific user
+    static func user(withID userId: UserId) -> Self {
+        .init(filter: .equal("id", to: userId))
+    }
```
VojtaStavik, line 61: 👀 We don't have a classic REST `users/<id>` GET call?

Sources_v3/Workers/UserUpdater.swift
```diff
@@ -0,0 +1,73 @@
...
+    ///
+    func loadUser(_ userId: UserId, completion: ((Error?) -> Void)? = nil) {
+        apiClient
+            .request(endpoint: .users(query: .user(withID: userId))) { (result: Result<UserListPayload<ExtraData.User>, Error>) in
+                switch result {
+                case let .success(payload):
+                    log.assert(
+                        payload.users.count <= 1,
+                        "UserUpdater.loadUser must fetch exactly 0 or 1 user. Fetched: \(payload.users)"
+                    )
```
VojtaStavik, line 45: You should have asked the backed team to implement the missing endpoint. These workarounds shouldn't be needed.

Sources_v3/Workers/UserUpdater_Tests.swift
```diff
@@ -0,0 +1,217 @@
...
+        // Load the user
+        var user: UserDTO? {
+            database.viewContext.user(id: userPayload.id)
+        }
+        
+        // Assert the completion is called
+        AssertAsync.willBeTrue(completionIsCalled)
+        
+        // Assert the user is saved to the database
+        AssertAsync.willBeEqual(user?.id, userPayload.id)
```
VojtaStavik, line 215: nit: You can merge these into one `AssertAsync { }`, it will execute faster

VojtaStavik (approved):
Very nice 👍 It's great to see how the patters get more established and it's more and more clear how we should do things.

I don't like the single-user query workaround. Let's discuss in slack.

Sources_v3/Workers/UserUpdater.swift
```diff
@@ -0,0 +1,73 @@
...
+    /// - Parameters:
+    ///   - userId: The user identifier
+    ///   - completion: Called when the API call is finished. Called with `Error` if the remote update fails.
+    ///
+    func loadUser(_ userId: UserId, completion: ((Error?) -> Void)? = nil) {
+        apiClient
+            .request(endpoint: .users(query: .user(withID: userId))) { (result: Result<UserListPayload<ExtraData.User>, Error>) in
+                switch result {
+                case let .success(payload):
+                    log.assert(
```
VojtaStavik, line 42: Let's not use assert because this value comes from the server. A simple `log.error` should be enough here. We don't want the app to start crashing when something goes wrong on the server :)

Sources_v3/Query/UserListQuery.swift
evsaev: Yep, there is no such

Tests_v3/Dummy data/MutedUser.swift
evsaev: fixed in 1a1e81ac0ce4ffde69a610f60713cd5291b62159

Sources_v3/Workers/UserUpdater.swift
evsaev: fixed in f703749d3bae7c64fff00849a4725c7ff49ce13c

PR #529, opened 2020-10-05
[CIS-375] NewUserQueryUpdater

mantergo:
**In this PR:** 

- Introduce `NewUserQueryUpdater`

_I’ve tried unifying it by creating generic NewDTOQueryUpdater, but it will lead to creating a lot of protocols, and methods of modifying filters will still be different so I decided we don’t need it at this point._

### DIFF
--- a/Sources_v3/Workers/Background/NewUserQueryUpdater_Tests.swift
+++ b/Sources_v3/Workers/Background/NewUserQueryUpdater_Tests.swift
@@ -1,0 +1,125 @@
+//
+// Copyright © 2020 Stream.io Inc. All rights reserved.
+//
+
+@testable import StreamChatClient
+import XCTest
+
+class NewUserQueryUpdater_Tests: StressTestCase {
+    typealias ExtraData = DefaultExtraData.User
+    
+    private var env: TestEnvironment!
+    
+    var database: DatabaseContainerMock!
+    var webSocketClient: WebSocketClientMock!
+    var apiClient: APIClientMock!
+    
+    var newUserQueryUpdater: NewUserQueryUpdater<ExtraData>?
+    
+    override func setUp() {
+        super.setUp()
+        env = TestEnvironment()
+        
+        database = try! DatabaseContainerMock(kind: .inMemory)
+        webSocketClient = WebSocketClientMock()
+        apiClient = APIClientMock()
+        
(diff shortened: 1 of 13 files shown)


Sources_v3/Workers/Background/NewUserQueryUpdater_Tests.swift
```diff
@@ -0,0 +1,125 @@
...
+            Assert.canBeReleased(&database)
+            Assert.canBeReleased(&webSocketClient)
+            Assert.canBeReleased(&apiClient)
+            Assert.canBeReleased(&env)
+        }
+        
+        super.tearDown()
+    }
+    
+    func test_update_called_for_each_query() throws {
```
VojtaStavik, line 49: nit: The method name should have `_` between all words. The pattern usually is: `test_something_shouldDoSomething_underCertainConditions`

Sources_v3/Workers/Background/NewUserQueryUpdater_Tests.swift
```diff
@@ -0,0 +1,125 @@
...
+        let filter1: Filter = .contains(.unique, String.unique)
+        let filter2: Filter = .notEqual(.unique, to: 1)
+        
+        try database.createUserListQuery(filter: filter1)
+        try database.createUserListQuery(filter: filter2)
+                
+        try database.createUser()
+        
+        // Assert `update(userListQuery` called for each query in DB
+        AssertAsync.willBeEqual(env!.userQueryUpdater?.update_calls_counter, 2)
```
VojtaStavik, line 59: What if it was called twice for the `filter1`? 🤷‍♂️ :trollface:

VojtaStavik (approved):
LGTM, thanks 👍

Sources_v3/Workers/Background/NewUserQueryUpdater_Tests.swift
mantergo: Yep, thank you. I have some issues with test naming 😅
Fixed in 95ec895f

Sources_v3/Workers/Background/NewUserQueryUpdater_Tests.swift
mantergo: Fixed in 95ec895f 🤓

PR #530, opened 2020-10-05
[CIS-300] Change `removeAllData` implementation to emit callback

VojtaStavik:
This PR fixed the problem specified in CIS-300.

**Solution:**
- `DatabaseContainer.removeAllData` is changed such that it sends `WillRemoveAllData` and `DidRemoveAllData` notifications for every context is owns. It also uses a different (and simplified) way of deleting all data by destroying the underlying persistent store.
- Because we never use `NSFetchedResults` controller directly (this turned out to be a very wise decision) we can tweak the `ListDatabaseObserver` and `EntityDatabaseObsever` implementation. The observers now listen for the notifications above posted by the context they observe. If the notification is received, the observers simulate a callback from the underlying FRC like when the data were deleted. The observers also call `performFetch()` on their FRC which forces them to reload the data, and get the data from the persistent store, which is empty -> no data loaded.

---

This PR is based on the approach suggested in #497, thanks @evsaev!

### DIFF
--- a/Sources_v3/Controllers/EntityDatabaseObserver.swift
+++ b/Sources_v3/Controllers/EntityDatabaseObserver.swift
@@ -111,6 +111,9 @@ class EntityDatabaseObserver<Item, DTO: NSManagedObject> {
     let request: NSFetchRequest<DTO>
     let context: NSManagedObjectContext
     
+    /// When called, release the nofication observers
+    var releaseNotificationObservers: (() -> Void)?
+    
     /// Creates a new `ListObserver`.
     ///
     /// Please note that no updates are reported until you call `startUpdating`.
@@ -144,6 +147,12 @@ class EntityDatabaseObserver<Item, DTO: NSManagedObject> {
             
             return fetchedObjects.first.flatMap(itemCreator)
         }
+        
+        listenForRemoveAllDataNotifications()
+    }
+    
+    deinit {
+        releaseNotificationObservers?()
     }
     
     /// Starts observing the changes in the database. The current items in the list are synchronously available in the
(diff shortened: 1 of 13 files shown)


Sources_v3/Database/DatabaseContainer.swift
```diff
@@ -201,14 +223,30 @@ class DatabaseContainer: NSPersistentContainer {
...
+
+        // Remove all loaded persistent stores first
+        try persistentStoreCoordinator.persistentStores.forEach { store in
+            try persistentStoreCoordinator.remove(store)
+        }
         
+        // If the store was SQLite store, remove the actual DB file
+        if storeDescription.type == NSSQLiteStoreType,
+            let storeURL = storeDescription.url,
+            storeURL.absoluteString.hasSuffix("/dev/null") == false {
```
mantergo, line 245: Seems like now we can remove `kind != .inMemory` check from `init` as we are safe inside `recreatePersistentStore()`.

mantergo (approved):
Liked this PR!

Sources_v3/Database/DatabaseContainer.swift
VojtaStavik: We can but it's not necessary. The check is technically correct. There's nothing to recreate if we use in-memory store :)

VojtaStavik: Fixed in 92cbe57af5ba29cb17434a9926d0d481d50d3b51

evsaev (commented):
Happy to have this handled in such an elegant way 👍

b-onc (approved):
🚢 🇮🇹

Sources_v3/Controllers/EntityDatabaseObserver.swift
```diff
@@ -157,6 +162,48 @@ class EntityDatabaseObserver<Item, DTO: NSManagedObject> {
...
     }
+    
+    /// Listens for `Will/DidRemoveAllData` notifications from the context and simulates the callback when the notifications
+    /// are received.
+    private func listenForRemoveAllDataNotifications() {
+        // When `WillRemoveAllDataNotification` is received, we need to simulate the callback from change observer, like all
+        // existing entities are being removed. At this point, these entities still existing in the context, and it's safe to
+        // access and serialize them.
+        let willRemoveAllDataNotificationObserver = NotificationCenter.default
+            .addObserver(forName: DatabaseContainer.WillRemoveAllDataNotification, object: context, queue: .main) { [weak self] _ in
```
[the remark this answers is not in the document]
VojtaStavik: True! Fixed in 5072196792012f40fd51fd7d453b420f2b25d220 and 51e463d6e36511bf94c3c530d2f51bd5b0227617

Sources_v3/Controllers/EntityDatabaseObserver.swift
```diff
@@ -111,6 +111,9 @@ class EntityDatabaseObserver<Item, DTO: NSManagedObject> {
     let request: NSFetchRequest<DTO>
     let context: NSManagedObjectContext
     
+    /// The observers for `DatabaseContaner.Will/DidRemoveAllDataNotificationName`.
+    var removeAllDataNotificationObservers: [NSObjectProtocol] = []
```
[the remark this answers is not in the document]
VojtaStavik: Like it 👍 Fixed in 5072196792012f40fd51fd7d453b420f2b25d220 and 51e463d6e36511bf94c3c530d2f51bd5b0227617

PR #533, opened 2020-10-06
[CIS-383] Remove .none Filter option

mantergo:
**In this PR:**

- Remove `.none` `Filter` option
- Adjust `NewUserQueryUpdater` to handle `nil` filter
- Adjust `Sample` app after introducing `nil` `Filter`

### DIFF
--- a/Sample_v3/Samples/CombineSimpleChat/CombineSimpleChannelsViewController.swift
+++ b/Sample_v3/Samples/CombineSimpleChat/CombineSimpleChannelsViewController.swift
@@ -236,8 +236,7 @@ class CombineSimpleChannelsViewController: UITableViewController {
         else { return }
         
         usersViewController.userListController = chatClient
-            // TODO: Change filter to all users after Filter adjustments
-            .userListController(query: .init(filter: .autocomplete("name", with: "a")))
+            .userListController(query: .init())
         navigationController?.pushViewController(usersViewController, animated: true)
     }
     
--- a/Sources_v3/Database/DTOs/UserDTO.swift
+++ b/Sources_v3/Database/DTOs/UserDTO.swift
@@ -92,6 +92,16 @@ extension NSManagedObjectContext: UserDatabaseSession {
         
         return dto
     }
+    
+    func updateQuery(
+        for userId: UserId,
+        queryFilterHash: String
+    ) {
+        if let userDTO = user(id: userId),
+            let queryDTO = userListQuery(filterHash: queryFilterHash) {
+            queryDTO.users.insert(userDTO)
+        }
+    }
 }
 
 extension UserDTO {
@@ -122,7 +132,7 @@ extension UserDTO {
         let sortDescriptors = query.sort.compactMap { $0.key.sortDescriptor(isAscending: $0.isAscending) }
         request.sortDescriptors = sortDescriptors.isEmpty ? [UserListSortingKey.defaultSortDescriptor] : sortDescriptors
                 
-        request.predicate = NSPredicate(format: "ANY queries.filterHash == %@", query.filter.filterHash)
+        request.predicate = NSPredicate(format: "ANY queries.filterHash == %@", query.filter?.filterHash ?? Filter.nilFilterHash)
         return request
     }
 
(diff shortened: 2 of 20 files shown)


Sources_v3/Query/UserListQuery.swift
```diff
@@ -25,24 +25,29 @@ public struct UserListQuery: Encodable {
...
     public func encode(to encoder: Encoder) throws {
         var container = encoder.container(keyedBy: CodingKeys.self)
-        try container.encode(filter, forKey: .filter)
-        
+        switch filter {
+        case let .some(value): try container.encode(value, forKey: .filter)
+        case .none:
+            // Backend expects empty object for "filter_conditions" in case no filter specified.
+            struct EmptyObject: Encodable {}
+            try container.encode(EmptyObject(), forKey: .filter)
```
mantergo, line 48: I don't like this part 🙂
Do you know more elegant way to simulate empty object?
Backend refuses to accept `null` or empty string.

VojtaStavik: Yeah, this is unfortunately needed. You can do something like:
```swift
try container.encode([String](), forKey: .filter)
```
But to honest, I find you solution easier to understand and I don't mind having it in the codebase.

Sources_v3/Database/DTOs/UserDTO.swift
```diff
@@ -92,6 +92,16 @@ extension NSManagedObjectContext: UserDatabaseSession {
         
         return dto
     }
+    
+    func updateQuery(
+        for userId: UserId,
+        queryFilterHash: String
+    ) throws {
```
VojtaStavik, line 99: Either is the `throws` here redundant or you forgot to throw some errors :)

Sources_v3/Workers/Background/NewUserQueryUpdater_Tests.swift
```diff
@@ -57,11 +57,29 @@ class NewUserQueryUpdater_Tests: StressTestCase {
...
+        try database.createUser()
+        
+        // Assert `update(userListQuery` is not called for .none filter query
+        AssertAsync.willBeTrue(env!.userQueryUpdater?.update_queries.isEmpty)
+    }
+    
+    func test_newUser_updated_withNilFilterQuery() throws {
+        let id: UserId = .unique
+        
+        try database.createUserListQuery(filter: .none)
```
VojtaStavik, line 76: This `.none` is `Optional.none` not `Filter.none` 🤯 It took me a while to understand. However, I'd rather use `nil` :D :D

Sample_v3/Samples/CombineSimpleChat/CombineSimpleChannelsViewController.swift
```diff
@@ -236,8 +236,7 @@ class CombineSimpleChannelsViewController: UITableViewController {
         else { return }
         
         usersViewController.userListController = chatClient
-            // TODO: Change filter to all users after Filter adjustments
-            .userListController(query: .init(filter: .autocomplete("name", with: "a")))
+            .userListController(query: .init())
```
VojtaStavik, line 239: Maybe `.init()` should be the default parameter of `. userListController()` then?

Sources_v3/Workers/Background/NewUserQueryUpdater_Tests.swift
mantergo: Haha, I changed it to `nil` in another test, but forgot here cause it doesn't throw compilation error 😅

Sample_v3/Samples/CombineSimpleChat/CombineSimpleChannelsViewController.swift
mantergo: I thought about it, but probably we want developers always use query explicitly, even if it default one.

Sources_v3/Query/UserListQuery.swift
```diff
@@ -25,24 +25,29 @@ public struct UserListQuery: Encodable {
...
         self.filter = filter
         self.sort = sort
         self.pagination = pagination
     }
     
     public func encode(to encoder: Encoder) throws {
         var container = encoder.container(keyedBy: CodingKeys.self)
-        try container.encode(filter, forKey: .filter)
-        
+        switch filter {
```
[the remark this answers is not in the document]
mantergo: Nope 👀

PR #534, opened 2020-10-06
[CIS-380] Implement member controller

evsaev:
**This PR**: 
- introduces `ChatMemberController` for actions on a channel member (currently, only **ban/unban** actions are in)
- introduces `ChannelMemberListUpdater ` and stuff for querying the channel members
- introduces `ChannelMemberUpdater`  and stuff for channel member related API calls
- introduces `ChannelMemberListQueryDTO` and database session for local representation of member queries

### DIFF
--- a/Sources_v3/Controllers/MemberController/MemberController.swift
+++ b/Sources_v3/Controllers/MemberController/MemberController.swift
@@ -1,0 +1,341 @@
+//
+// Copyright © 2020 Stream.io Inc. All rights reserved.
+//
+
+import CoreData
+import Foundation
+
+public extension _ChatClient {
+    /// Creates a new `_ChatChannelMemberController` for the user with the provided `userId` and `cid`.
+    /// - Parameters:
+    ///   - userId: The user identifier.
+    ///   - cid: The channel identifier.
+    /// - Returns: A new instance of `_ChatChannelMemberController`.
+    func memberController(userId: UserId, in cid: ChannelId) -> _ChatChannelMemberController<ExtraData> {
+        .init(userId: userId, cid: cid, client: self)
+    }
+}
+
+/// `ChatChannelMemberController` is a controller class which allows mutating and observing changes of a specific chat member.
+///
+/// `ChatChannelMemberController` objects are lightweight, and they can be used for both, continuous data change observations,
+/// and for quick user actions (like ban/unban).
+///
+/// - Note: `ChatChannelMemberController` is a typealias of `_ChatChannelMemberController` with default extra data.
+/// If you're using custom extra data, create your own typealias of `_ChatChannelMemberController`.
+///
+/// Learn more about using custom extra data in our [cheat sheet](https://github.com/GetStream/stream-chat-swift/wiki/StreamChat-SDK-Cheat-Sheet#working-with-extra-data).
+///
+public typealias ChatChannelMemberController = _ChatChannelMemberController<DefaultExtraData>
+
+/// `_ChatChannelMemberController` is a controller class which allows mutating and observing changes of a specific chat member.
+///
+/// `_ChatChannelMemberController` objects are lightweight, and they can be used for both, continuous data change observations,
+/// and for quick user actions (like mute/unmute).
+///
+/// - Note: `ChatChannelMemberController` is a typealias of `_ChatChannelMemberController` with default extra data.
+/// If you're using custom extra data, create your own typealias of `_ChatChannelMemberController`.
+///
+/// Learn more about using custom extra data in our [cheat sheet](https://github.com/GetStream/stream-chat-swift/wiki/StreamChat-SDK-Cheat-Sheet#working-with-extra-data).
+///
+public class _ChatChannelMemberController<ExtraData: ExtraDataTypes>: DataController, DelegateCallable, DataStoreProvider {
+    /// The identifier of the user this controller observes.
+    public let userId: UserId
+    
+    /// The identifier of the channel the user is member of.
+    public let cid: ChannelId
+    
+    /// The `ChatClient` instance this controller belongs to.
+    public let client: _ChatClient<ExtraData>
+    
+    /// The user the controller represents.
+    ///
+    /// To observe changes of the chat member, set your class as a delegate of this controller or use the provided
+    /// `Combine` publishers.
+    public var member: _ChatChannelMember<ExtraData.User>? {
+        startObservingIfNeeded()
+        return memberObserver.item
+    }
+    
+    /// A type-erased delegate.
+    var multicastDelegate: MulticastDelegate<AnyChatMemberControllerDelegate<ExtraData>> = .init() {
+        didSet {
+            stateMulticastDelegate.mainDelegate = multicastDelegate.mainDelegate
+            stateMulticastDelegate.additionalDelegates = multicastDelegate.additionalDelegates
+            startObservingIfNeeded()
+        }
+    }
+    
+    /// The worker used to update channel members.
+    private lazy var memberUpdater = createMemberUpdater()
+    
+    /// The worker used to fetch channel members.
+    private lazy var memberListUpdater = createMemberListUpdater()
+    
+    /// The observer used to track the user changes in the database.
+    private lazy var memberObserver = createMemberObserver()
+        .onChange { [unowned self] change in
+            self.delegateCallback {
+                $0.memberController(self, didUpdateMember: change)
+            }
+        }
+    
+    private let environment: Environment
+
+    /// Creates a new `_ChatChannelMemberController`
+    /// - Parameters:
+    ///   - userId: The user identifier.
+    ///   - cid: The channel identifier the user is member of.
+    ///   - client: The `Client` this controller belongs to.
+    ///   - environment: Environment for this controller.
+    init(
+        userId: UserId,
+        cid: ChannelId,
+        client: _ChatClient<ExtraData>,
+        environment: Environment = .init()
+    ) {
+        self.userId = userId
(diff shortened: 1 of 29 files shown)


Sources_v3/Query/Sorting.swift
```diff
@@ -156,3 +156,19 @@ extension UserListSortingKey {
         return nil
     }
 }
+
+// MARK: - Members Sorting
+
+/// `ChannelMemberListSortingKey` describes the keys by which you can get sorted channel members after query.
+public struct ChannelMemberListSortingKey: RawRepresentable, Equatable, SortingKey {
```
evsaev, line 163: Other types for sorting keys we manage via enums but I think the `RawRepresentable` fits better: we don't need to provide explicit `encode` implementation and jump around `.custom(String)` case.

Sources_v3/Database/DTOs/ChannelMemberListQueryDTO.swift
```diff
@@ -0,0 +1,23 @@
+//
+// Copyright © 2020 Stream.io Inc. All rights reserved.
+//
+
+import CoreData
+
+@objc(ChannelMemberListQueryDTO)
+final class ChannelMemberListQueryDTO: NSManagedObject {
```
evsaev, line 8: Was not sure if we want to have the relation to the `channel` here. 

If so, the logic in `ChannelMemberListUpdater` should to updated to fetch the channel from backend first and then perform the query call.

evsaev:
When working with queries the bg-worker that links new entities to existed local query is needed. However, at this point it's not requited since we work with the single entity and fetch it directly from backend in `controller.synchronize`. The bg-worker will be required when we start working on **MemberListController**.

Sources_v3/Controllers/MemberController/MemberController_Tests.swift
```diff
@@ -0,0 +1,471 @@
...
+        // Assert `userId` is correct.
+        XCTAssertEqual(controller.userId, userId)
+
+        // Assert `cid` is correct.
+        XCTAssertEqual(controller.cid, cid)
+    }
+
+    // MARK: - Synchronize tests
+
+    func test_synchronize_changesStateand_and_callsCompletionOnCallbackQueue() {
```
VojtaStavik, line 79: nit: The test name has some typos

VojtaStavik (approved):
Looks great! Thanks 👍 

---

*Q:* What functionality did you implement?
*A:* A controller that allows banning and unbanning of a member 😎 
*Q:* Sounds cool! How many lines of code did you need?
*A:* Almost 2000 😭 ... but half of that is comments!

Sources_v3/Database/DTOs/ChannelMemberListQueryDTO.swift
evsaev: I vote for the relationship with the `ChannelDTO`: 
- if one time we decide to have a model for this query, it will require the channel model
- `Filter` is already too complicated and it is for general purpose, not only channel related queries

VojtaStavik: 👍

Sources_v3/Controllers/MemberController/MemberController.swift
```diff
@@ -0,0 +1,339 @@
...
+///
+/// `ChatMemberController` objects are lightweight, and they can be used for both, continuous data change observations,
+/// and for quick user actions (like ban/unban).
+///
+/// - Note: `ChatMemberController` is a typealias of `_ChatMemberController` with default extra data. If you're using custom
+/// extra data, create your own typealias of `_ChatMemberController`.
+///
+/// Learn more about using custom extra data in our [cheat sheet](https://github.com/GetStream/stream-chat-swift/wiki/StreamChat-SDK-Cheat-Sheet#working-with-extra-data).
+///
+public typealias ChatMemberController = _ChatMemberController<DefaultExtraData>
```
VojtaStavik, line 29: Should this be called `ChatChannelMemberController` ? cc @b-onc

Sources_v3/Database/DTOs/ChannelMemberListQueryDTO.swift
evsaev: fixed in 3429b45af6c0a8f22bff5116ca2b745cb2719351 and 4c3c349d78702726a80fd1eec3f7710080edca1c

Sources_v3/Query/Sorting.swift
```diff
@@ -156,3 +156,19 @@ extension UserListSortingKey {
...
+    public let rawValue: String
+    
+    public init(rawValue: String) {
+        self.rawValue = rawValue
+    }
+}
+
+public extension ChannelMemberListSortingKey {
+    static let createdAt = Self(rawValue: "created_at")
+    static let updatedAt = Self(rawValue: "updated_at")
```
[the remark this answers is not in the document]
evsaev: fixed in 4ff526dc8b1b0cd1d3ce21c552737bd1707c4cdf

Sources_v3/Database/DatabaseSession.swift
```diff
@@ -120,12 +120,22 @@ protocol MemberDatabaseSession {
     func member(userId: UserId, cid: ChannelId) -> MemberDTO?
 }
 
+protocol ChannelMemberListQueryDatabaseSession {
```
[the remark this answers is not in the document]
evsaev: fixed in 4c3c349d78702726a80fd1eec3f7710080edca1c

Sources_v3/Controllers/MemberController/MemberController_Tests.swift
evsaev: fixed in 973619a02a94368cdd60541b66007f8e1ec5a3fe

Sources_v3/Controllers/MemberController/MemberController.swift
evsaev: fixed in 973619a02a94368cdd60541b66007f8e1ec5a3fe and also renamed `BanUserRequestPayload -> ChannelMemberBanRequestPayload`

Sources_v3/Query/ChannelMemberListQuery.swift
```diff
@@ -0,0 +1,69 @@
...
+        try container.encode(cid.type, forKey: .channelType)
+        try container.encode(filter, forKey: .filter)
+        try pagination.encode(to: encoder)
+        if !sort.isEmpty { try container.encode(sort, forKey: .sort) }
+    }
+}
+
+extension ChannelMemberListQuery {
+    var hash: String {
+        filter.filterHash.appending(cid.rawValue)
```
evsaev, line 56: For `ChannelMemberListQuery` we cannot just go with **filter.filterHash** because we can have queries with the same filter but for different channels. So we should make the **cid** be a part of DTO identifier. 

Should we also use **sort** property when calculating the hash?

VojtaStavik: Understood 👍 

I think `sort` should also be included because it affects pagination and it can cause issues.

Sources_v3/Query/ChannelMemberListQuery.swift
```diff
@@ -0,0 +1,69 @@
...
+        try container.encode(cid.id, forKey: .channelId)
+        try container.encode(cid.type, forKey: .channelType)
+        try container.encode(filter, forKey: .filter)
+        try pagination.encode(to: encoder)
+        if !sort.isEmpty { try container.encode(sort, forKey: .sort) }
+    }
+}
+
+extension ChannelMemberListQuery {
+    var hash: String {
```
VojtaStavik, line 55: Can we rename this to `queryHash` or something like that? So it doesn't like like `Hashable` hash.

Sources_v3/Workers/ChannelMemberListUpdater.swift
```diff
@@ -0,0 +1,63 @@
...
+
+/// Makes a channel members query call to the backend and updates the local storage with the results.
+class ChannelMemberListUpdater<ExtraData: ExtraDataTypes>: Worker {
+    /// Makes a channel members query call to the backend and updates the local storage with the results.
+    /// - Parameters:
+    ///   - query: The query used in the request.
+    ///   - completion: Called when the API call is finished. Called with `Error` if the remote update fails.
+    func load(_ query: ChannelMemberListQuery, completion: ((Error?) -> Void)? = nil) {
+        let channelEndpoint: Endpoint<ChannelPayload<ExtraData>> = .channel(query: .init(cid: query.cid))
+        apiClient.request(endpoint: channelEndpoint) { channelResult in
```
VojtaStavik, line 15: This API call is needed only when the channel doesn't exist locally yet, which will be very rare.

Sources_v3/Query/ChannelMemberListQuery.swift
evsaev: ok, will do 👍 The same applies for other queries, doesn't it?

VojtaStavik: Yes 👍

Sources_v3/Workers/ChannelMemberListUpdater.swift
evsaev: fixed in 24ae4cbb7982f78158fb534cc23319c533ddd44c

Sources_v3/Query/ChannelMemberListQuery.swift
evsaev: fixed in 8726a15382587d0082e984502c7fdf8e5eff1219

PR #536, opened 2020-10-08
Sample v3: Separate login configurations into a new screen + persistence

cardoso:
This PR separates the credentials and other configuration from the first screen into a new modal screen. It also persists these parameters in UserDefaults.

Screen Shot 2020-10-07 at 21 03 27

#no_changelog

### DIFF
--- a/Sample_v3/Configuration.swift
+++ b/Sample_v3/Configuration.swift
@@ -6,51 +6,74 @@ import Foundation
 import StreamChatClient
 
 struct Configuration {
+    struct PersistenceKeys {
+        static let apiKey = "apiKey"
+        static let userId = "userId"
+        static let userName = "userName"
+        static let baseURL = "baseURL"
+        static let token = "token"
+        static let isLocalStorageEnabled = "isLocalStorageEnabled"
+        static let shouldFlushLocalStorageOnStart = "shouldFlushLocalStorageOnStart"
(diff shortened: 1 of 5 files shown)


Sample_v3/Configuration.swift
```diff
@@ -0,0 +1,56 @@
...
+    static var isLocalStorageEnabled: Bool {
+        get { UserDefaults.standard.value(forKey: "isLocalStorageEnabled") as? Bool ?? true }
+        set { UserDefaults.standard.setValue(newValue, forKey: "isLocalStorageEnabled") }
+    }
+    
+    static var shouldFlushLocalStorageOnStart: Bool {
+        get { UserDefaults.standard.value(forKey: "shouldFlushLocalStorageOnStart") as? Bool ?? false }
+        set { UserDefaults.standard.setValue(newValue, forKey: "shouldFlushLocalStorageOnStart") }
+    }
+}
```
VojtaStavik, line 56: Can we clean this up a bit? Concretely, by putting all the default values in one place & extracting the keys into variables?

VojtaStavik (approved):
Looks good, thank you 👍

Sample_v3/Configuration.swift
cardoso: Done :)

PR #537, opened 2020-10-08
[CIS-381] Member controller wrappers

evsaev:
**This PR:**
- adds `SwiftUI` wrapper for `ChatChannelMemberController`
- adds `Combine` wrapper for `ChatChannelMemberController`
- fixes the delegate types naming

### DIFF
--- a/Sources_v3/Controllers/MemberController/MemberController+SwiftUI.swift
+++ b/Sources_v3/Controllers/MemberController/MemberController+SwiftUI.swift
@@ -1,0 +1,48 @@
+//
+// Copyright © 2020 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+import SwiftUI
+
+@available(iOS 13, *)
+extension _ChatChannelMemberController {
+    /// A wrapper object that exposes the controller variables in the form of `ObservableObject` to be used in SwiftUI.
+    public var observableObject: ObservableObject { .init(controller: self) }
+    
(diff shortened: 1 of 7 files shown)


Sources_v3/Controllers/MemberController/MemberController+SwiftUI.swift
```diff
@@ -0,0 +1,48 @@
...
+    /// A wrapper object that exposes the controller variables in the form of `ObservableObject` to be used in SwiftUI.
+    public var observableObject: ObservableObject { .init(controller: self) }
+    
+    /// A wrapper object for `_ChatChannelMemberController` type which makes it possible to use the controller comfortably in SwiftUI.
+    public class ObservableObject: SwiftUI.ObservableObject {
+        /// The underlying controller. You can still access it and call methods on it.
+        public let controller: _ChatChannelMemberController
+        
+        /// The channel member.
+        @Published public private(set) var member: _ChatChannelMember<ExtraData.User>?
```
evsaev, line 19: Did it in the same way as in other wrappers but wondering why we expose `Item` but not the `Change<Item>`? 🤔

VojtaStavik: Because SwiftUI figures out the changes itself. We do it the same way for array.

VojtaStavik (approved):
Looks good 👍

PR #539, opened 2020-10-08
[CIS-386] Member list controller

evsaev:
**This PR**:
- introduces `ChatChannelMemberListController` that manages channel member query
- introduces mock `ListDatabaseObserver`
- adds sort descriptors to `ChannelMemberListSortingKey`

### DIFF
--- a/Sources_v3/Controllers/MemberListController/MemberListController.swift
+++ b/Sources_v3/Controllers/MemberListController/MemberListController.swift
@@ -1,0 +1,305 @@
+//
+// Copyright © 2020 Stream.io Inc. All rights reserved.
+//
+
+import CoreData
+import Foundation
+
+extension _ChatClient {
+    /// Creates a new `_ChatChannelMemberListController` with the provided query.
+    /// - Parameter query: The query specify the filter and sorting options for members the controller should fetch.
+    /// - Returns: A new instance of `_ChatChannelMemberListController`.
+    public func memberListController(query: ChannelMemberListQuery) -> _ChatChannelMemberListController<ExtraData> {
+        .init(query: query, client: self)
+    }
+}
+
+/// `_ChatChannelMemberListController` is a controller class which allows observing a list of
+/// channel members based on the provided query.
+///
+/// Learn more about `_ChatChannelMemberListController` and its usage in our [cheat sheet](https://github.com/GetStream/stream-chat-swift/wiki/StreamChat-SDK-Cheat-Sheet#user-list).
+///
+/// - Note: `ChatChannelMemberListController` is a typealias of `_ChatChannelMemberListController` with default extra data.
+/// If you're using custom extra data, create your own typealias of `_ChatChannelMemberListController`.
+///
+/// Learn more about using custom extra data in our [cheat sheet](https://github.com/GetStream/stream-chat-swift/wiki/StreamChat-SDK-Cheat-Sheet#working-with-extra-data).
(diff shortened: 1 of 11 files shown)


Sources_v3/Controllers/MemberListController/MemberListController.swift
```diff
@@ -0,0 +1,305 @@
...
+    ///   - completion: The completion. Will be called on a **callbackQueue** when the network request is finished.
+    ///                 If request fails, the completion will be called with an error.
+    func loadNextMembers(
+        limit: Int = 25,
+        completion: ((Error?) -> Void)? = nil
+    ) {
+        var updatedQuery = query
+        updatedQuery.pagination = [.limit(limit), .offset(members.count)]
+        memberListUpdater.load(updatedQuery) { [weak self] error in
+            self?.query = updatedQuery
```
evsaev, line 163: We don't update queries in other `***ListController` types. Seems like we do need to update to keep it relevant.

Tests_v3/Dummy data/MemberPayload.swift
```diff
@@ -7,12 +7,17 @@ import Foundation
...
+    ) -> MemberPayload {
         .init(
             user: .dummy(userId: userId),
             role: role,
-            createdAt: .unique,
-            updatedAt: .unique
+            createdAt: createdAt,
+            updatedAt: updatedAt
         )
     }
```
VojtaStavik, line 22: Do you think we will eventually end up with the same number of parameters as the `init` has and this helper will become pointless? :trollface:

Sources_v3/Controllers/MemberListController/MemberListController.swift
VojtaStavik: Hmmm, why do you think we need to update it? The only thing that changes is the pagination options and we don't need to persist it.

Sources_v3/Controllers/MemberListController/MemberListController_Tests.swift
```diff
@@ -0,0 +1,525 @@
...
+            return .insert(item[keyPath: path], index: at)
+        case let .move(item, from, to):
+            return .move(item[keyPath: path], fromIndex: from, toIndex: to)
+        case let .remove(item, at):
+            return .remove(item[keyPath: path], index: at)
+        case let .update(item, at):
+            return .update(item[keyPath: path], index: at)
+        }
+    }
+}
```
VojtaStavik, line 525: Why this?

VojtaStavik (approved):
Look good 👍 Left a couple of questions though

Sources_v3/Controllers/MemberListController/MemberListController.swift
evsaev: Decided to change the `controller.query` because `Pagination.limit/offset` are public and I assumed that end users might use these field -> would be nice to have them up-to-date

Sources_v3/Controllers/MemberListController/MemberListController_Tests.swift
evsaev: Good question! 

It's quite tricky to validate `[ListChange<Item>]` received by the delegate so I found those helpers useful but was too lazy to implement it in SDK target and write the tests. 

Do you think we should move this stuff outside test target?

evsaev: fixed in cc985467a6707399d7bbfa1d6b31f70134e6e0a1 and 3116d4f491d97e78646229d47352f7a98dbb217f

PR #540, opened 2020-10-08
[CIS-378] Showcase 1-1 channels and UserController in Sample app

mantergo:
**In this PR:** 

- Showcase `1-1 channels` and `UserController` in Sample app

**Notes:** 
- Kept both `Users` and `add channel` buttons for now cause probably what we need is `TabBar`.
- In `SwiftUI` sample after tapping the user direct chat will be created but not opened. (navigation limitation, decided not to dive into it at this point)
- We don't have `mutedUsers` available initially on `currentUser`. They will appear only after muting someone and receiving corresponding event. Are we aware of this?

### DIFF
--- a/Sources_v3/APIClient/Endpoints/ChannelEndpoints.swift
+++ b/Sources_v3/APIClient/Endpoints/ChannelEndpoints.swift
@@ -18,7 +18,7 @@ extension Endpoint {
     
     static func channel<ExtraData: ExtraDataTypes>(query: ChannelQuery<ExtraData>) -> Endpoint<ChannelPayload<ExtraData>> {
         .init(
-            path: "channels/\(query.cid.type.rawValue)/\(query.cid.id)/query",
+            path: "channels/\(query.cid.type.rawValue)\(query.cid.id == "*" ? "" : "/\(query.cid.id)")/query",
             method: .post,
             queryItems: nil,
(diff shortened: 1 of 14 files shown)


Sources_v3/APIClient/Endpoints/ChannelEndpoints.swift
```diff
@@ -18,7 +18,7 @@ extension Endpoint {
     
     static func channel<ExtraData: ExtraDataTypes>(query: ChannelQuery<ExtraData>) -> Endpoint<ChannelPayload<ExtraData>> {
         .init(
-            path: "channels/\(query.cid.type.rawValue)/\(query.cid.id)/query",
+            path: "channels/\(query.cid.type.rawValue)\(query.cid.id == "*" ? "" : "/\(query.cid.id)")/query",
```
mantergo, line 21: Started slack discussion on that 🙂

VojtaStavik (approved):
Looks good 👍 

 PLEASE, start adding screenshots to the PR description if you change something in the sample app. Thank you!

VojtaStavik:
> We don't have mutedUsers available initially on currentUser. They will appear only after muting someone and receiving corresponding event. Are we aware of this?

Once the connection is established, the muted users should be loaded. They are part of the first health check event. If they are not loaded, it's a bug -> add it to JIRA, or fix it 😉

PR #543, opened 2020-10-09
[CIS-389] Fix Sorting encoding

mantergo:
**In this PR:** 

- Fix `Sorting` encoding

_Note: Backend was not returning any errors but sorting was not applied._


VojtaStavik (approved):
Would you mind adding a simple test for this so we lock the correct behavior?

PR #545, opened 2020-10-09
[CIS-391] Fix compilation error in Sample

evsaev:
**This PR**:
- fixes the compilation error in Sample app


b-onc (commented):
Isn't `change.item` the current snapshot of `CurrentChatUser`? It should be the same with `controller.currentUser`, so `updateUserCell(with: change.item)` should be equivalent to `updateUser(with: controller.currentUser)`

VojtaStavik:
@b-onc exactly, that's why we're removing `change.item` from public API ;-)

PR #546, opened 2020-10-12
[CIS-394] ChannelId adjustment (missing id case)

mantergo:
**In this PR:** 

- Adjust `ChannelId` to always have valid value
- Separate `ChannelId` into `id` and `type` when id part of `ChannelId` is missing
- Make invalid `cid` property nil on `ChannelController` and `ChannelQuery`
- Remove parsing of invalid `ChannelId` in `EventPayload`

**Bonus:** 
- Fix typos in `ChannelController_Tests`


Sources_v3/Models/ChannelType.swift
```diff
@@ -47,11 +43,6 @@ public enum ChannelType: Codable, Hashable {
     ///
     /// - Parameter rawValue: a string raw value of a channel type.
     init(rawValue: String) {
-        if rawValue.isEmpty || rawValue.contains(" ") || rawValue == "unknown" {
-            self = .unknown
-            return
-        }
-
```
VojtaStavik, line 54: ?

Sources_v3/Query/ChannelQuery.swift
```diff
@@ -25,6 +27,18 @@ public struct ChannelQuery<ExtraData: ExtraDataTypes>: Encodable {
     public let options: QueryOptions
     /// ChannelCreatePayload that is needed only when creating channel
     let channelPayload: ChannelEditDetailPayload<ExtraData>?
+    
+    /// ChannelId this query handles.
+    var cid: ChannelId {
+        guard let id = id else { return .init(type: type, id: "") }
```
VojtaStavik, line 33: Doesn't this produce an invalid cid? Isn't that exactly what we wanted to avoid having? IMO this should be public and optional - or maybe I just don't understand what you meant by this 🤷‍♂️

Sources_v3/APIClient/Endpoints/Payloads/ChannelEditDetailPayload_Tests.swift
```diff
@@ -38,4 +38,26 @@ class ChannelEditDetailPayload_Tests: XCTestCase {
...
+        let cid: ChannelId = .unique
+        let payload2: ChannelEditDetailPayload<DefaultExtraData> = .init(
+            cid: cid,
+            team: nil,
+            members: [],
+            invites: [],
+            extraData: .defaultValue
+        )
+        
+        XCTAssertEqual(payload2.pathParameters, "\(payload2.type)/\(payload2.id!)")
```
VojtaStavik, line 61: Please add comments of what you're testing here

Sources_v3/Query/ChannelQuery_Tests.swift
```diff
@@ -38,4 +38,20 @@ class ChannelQuery_Tests: XCTestCase {
...
+            members: [.unique],
+            invites: [],
+            extraData: .defaultValue
+        ))
+        XCTAssertEqual(query1.pathParameters, "\(query1.type)")
+        
+        let cid: ChannelId = .unique
+        let query2: ChannelQuery<DefaultExtraData> = .init(cid: cid)
+        
+        XCTAssertEqual(query2.pathParameters, "\(query2.type)/\(query2.id!)")
```
VojtaStavik, line 55: Same here

Sources_v3/Controllers/ChannelController/ChannelController.swift
```diff
@@ -116,7 +116,7 @@ public class _ChatChannelController<ExtraData: ExtraDataTypes>: DataController,
     @Atomic public private(set) var channelQuery: ChannelQuery<ExtraData>
 
     /// The identifier of a channel this controller observes.
-    private var channelId: ChannelId { channelQuery.cid }
+    public var cid: ChannelId { channelQuery.cid }
```
VojtaStavik, line 119: Why when the value is already available by `controller.channelQuery.cid`?

Sources_v3/WebSocketClient/Events/EventPayload.swift
```diff
@@ -101,7 +101,7 @@ struct EventPayload<ExtraData: ExtraDataTypes>: Decodable {
         let container = try decoder.container(keyedBy: CodingKeys.self)
         eventType = try container.decode(EventType.self, forKey: .eventType)
         connectionId = try container.decodeIfPresent(String.self, forKey: .connectionId)
-        cid = try container.decodeIfPresent(ChannelId.self, forKey: .cid)
+        cid = try? container.decodeIfPresent(ChannelId.self, forKey: .cid)
```
VojtaStavik, line 104: Please add a comment why is this needed

VojtaStavik (commented):
I'm not sure I fully understand some of your decisions.

Sources_v3/Models/ChannelId.swift
```diff
@@ -9,60 +9,40 @@ import Foundation
...
-            let channelPair = cid.split(separator: ChannelId.separator)
-            let type = ChannelType(rawValue: String(channelPair[0]))
-            let id = String(channelPair[1])
-            self.init(type: type, id: id)
-        } else {
+        guard !cid.isEmpty, cid.split(separator: ChannelId.separator).count == 2 else {
             throw ClientError.InvalidChannelId("The channel id has invalid format and can't be decoded: \(cid)")
         }
+        
+        rawValue = cid
```
b-onc, line 31: Why don't we store it as separated, so we don't have to `split` it every time for `type` and `id`?

VojtaStavik: @b-onc Because then you will have to create the raw value every time you need it? I think it doesn't matter, to be honest :) It's either this or that.

Sources_v3/Query/ChannelQuery.swift
mantergo: From my perspective idea was to remove "invalid" cid from the `ChannelId` implementation. 
We can't control if id is valid from the point of` id` string inside `ChannelId`.
Here we have 2 options, right: optional or some dummy id. Making it optional will bring another complexity to already complicated `ChannelController` implementation, ton of new `ifs/guards` so I decided to go with dummy id.
We can make it optional but I'm not sure it's critical there. 
WDYT?

Sources_v3/Controllers/ChannelController/ChannelController.swift
mantergo: I don't have strong points on that. 
Just decided that it will be more friendly API if we will expose it on controller.

Sources_v3/Query/ChannelQuery.swift
VojtaStavik: I think it is critical tbh. Imagine this:
```
let controller1 = client.channelController(createDirectMessageChannelWith: members)
let controller2 = client.channelController(for: controller1.cid)
```
It's a completely valid thing to do and the consumer of the api has no ida, the cid is not a valid cid.

VojtaStavik: I think we can get rid of `isChannelAlreadyCreated` internally and replace it with `guard let cid`?

Sources_v3/APIClient/Endpoints/Payloads/ChannelEditDetailPayload_Tests.swift
mantergo: fixed in 000802cf

Sources_v3/Query/ChannelQuery_Tests.swift
mantergo: fixed in 000802cf

Sources_v3/Query/ChannelQuery.swift
mantergo: I see, right.
So I made it optional on `ChannelController` and now we have points to make `cid` public.

Sources_v3/WebSocketClient/Events/EventPayload.swift
mantergo: fixed in a6938696

Sources_v3/Models/ChannelType.swift
mantergo: I don't want to make `ChannelType` optional on `ChannelId` or throwable `init` cause we don't have any fallback for this check.
I made `ChannelId` init more safe instead. Check 13ffaab7.

Sources_v3/Models/ChannelId.swift
```diff
@@ -9,60 +9,46 @@ import Foundation
...
-            let type = ChannelType(rawValue: String(channelPair[0]))
-            let id = String(channelPair[1])
-            self.init(type: type, id: id)
-        } else {
+        let channelPair = cid.split(separator: ChannelId.separator)
+    
+        guard
+            channelPair.count == 2,
+            !channelPair[0].replacingOccurrences(of: " ", with: "").isEmpty,
+            !channelPair[1].replacingOccurrences(of: " ", with: "").isEmpty
```
VojtaStavik, line 32: Do I understand this correctly to be a fancy-schmancy way to check if the string isn't `"<space><space><space>"`? :)

Sources_v3/APIClient/Endpoints/Payloads/ChannelEditDetailPayload.swift
```diff
@@ -43,4 +60,11 @@ struct ChannelEditDetailPayload<ExtraData: ExtraDataTypes>: Encodable {
 
         try extraData.encode(to: encoder)
     }
+    
+    var pathParameters: String {
+        guard let id = id else {
+            return "\(type)"
+        }
+        return "\(type)/\(id)"
+    }
```
VojtaStavik, line 69: nit: When we have time, it would be nice to introduce some kind of `APIPathConvertible` protocol these objects conform to. User will return id, channel cid, etc.

Sources_v3/Query/ChannelQuery.swift
VojtaStavik: @mantergo But `ChannelQuery.cid` should be optional, too.

Sample_v3/Samples/CombineSimpleChat/CombineSimpleChatViewController.swift
```diff
@@ -257,7 +257,7 @@ final class CombineSimpleChatViewController: UITableViewController, UITextViewDe
         let message = channelController.messages[indexPath.row]
         
         let messageController = channelController.client.messageController(
-            cid: channelController.channelQuery.cid,
+            cid: channelController.cid!,
             messageId: message.id
         )
```
VojtaStavik, line 262: Please, let's use `if let` in the sample code ;-) If it's not possible for `cid` to be `nil` like in this case (?), we can make it a `fatalError`.

Sources_v3/Controllers/ChannelController/ChannelController.swift
```diff
@@ -114,9 +114,27 @@ public typealias ChatChannelController = _ChatChannelController<DefaultExtraData
...
     /// The identifier of a channel this controller observes.
-    private var channelId: ChannelId { channelQuery.cid }
+    /// To perform any modifications on channel it needs to be created on backend side and have a valid cid.
+    /// If channel is not yet created on backend it will be `nil` here even though it's available in `ChannelQuery`.
+    /// Using `cid` value from here will guarantee that channel is created on backend and it's safe to create another
+    /// controllers where `cid` is required with this value.
+    public var cid: ChannelId? {
+        guard isChannelAlreadyCreated else { return nil }
+        return channelQuery.cid
+    }
```
VojtaStavik, line 137: I'm not sure about this assumption. The fact, that the channel hasn't been created yet, doesn't mean, its `cid` is invalid.

Let's summarize this:
- if you know the channel exists, you use `channelController(for cid: ChannelId)`, `controller.cid` will always return a valid cid
- for 1-1 channels, you use `channelController(createDirectMessageChannelWith members: Set<UserId>)`, before the channel is created, `controller.cid` will return `nil`.
- when you create a channel with `channelController(createChannelWithId cid: ChannelId)`, I don't see a reason why `controller.cid` should return `nil`, because the `cid` is known at the time. If you use the `cid` before the channel is created (you will need to check for `controller.state == .remoteDataFetched`), it's a programmer error 🤷‍♂️ 

There are two things mixed up together here. Exposing an invalid `cid` and a channel that doesn't exist yet.  The first version of the PR returned an invalid cid for some cases and the user had no chance to learn about it.  Having `cid` optional fixes this problem. IMO it shouldn't be related to the problem of mutating non-existing channel. It's already handled in the action methods.

Sources_v3/Query/ChannelQuery.swift
```diff
@@ -28,9 +28,10 @@ public struct ChannelQuery<ExtraData: ExtraDataTypes>: Encodable {
...
     let channelPayload: ChannelEditDetailPayload<ExtraData>?
     
-    /// ChannelId this query handles.
-    var cid: ChannelId {
-        guard let id = id else { return .init(type: type, id: "") }
+    /// `ChannelId` this query handles.
+    /// If `id` part is missing then it's impossible to create valid `ChannelId`.
+    public var cid: ChannelId? {
+        guard let id = id else { return nil }
         return ChannelId(type: type, id: id)
```
VojtaStavik: nit: You can do something like `return id.map { ChannelId(type: type, id: $0) }`

Sources_v3/Query/ChannelQuery.swift
```diff
@@ -28,9 +28,10 @@ public struct ChannelQuery<ExtraData: ExtraDataTypes>: Encodable {
     /// ChannelCreatePayload that is needed only when creating channel
     let channelPayload: ChannelEditDetailPayload<ExtraData>?
     
-    /// ChannelId this query handles.
-    var cid: ChannelId {
-        guard let id = id else { return .init(type: type, id: "") }
+    /// `ChannelId` this query handles.
+    /// If `id` part is missing then it's impossible to create valid `ChannelId`.
+    public var cid: ChannelId? {
```
VojtaStavik, line 33: nit: Here you're changing code you introduced in 000802cfcd718fbced3850c9abb7e9756d8feca5. It makes the review more difficult if each line is not changed exactly once.

Sources_v3/WebSocketClient/Events/EventPayload.swift
```diff
@@ -101,7 +101,9 @@ struct EventPayload<ExtraData: ExtraDataTypes>: Decodable {
         let container = try decoder.container(keyedBy: CodingKeys.self)
         eventType = try container.decode(EventType.self, forKey: .eventType)
         connectionId = try container.decodeIfPresent(String.self, forKey: .connectionId)
-        cid = try container.decodeIfPresent(ChannelId.self, forKey: .cid)
+        // In healthCheck event we can receive invalid id containing "*".
+        // We don't need to throw error in that case and can treat it like missing cid.
```
VojtaStavik, line 105: 👍 Thank you

VojtaStavik (commented):
Looks much better I think. The only important comment I have is about the logic of when to expose `cid` as `nil`.

Sources_v3/Models/ChannelId.swift
mantergo: Yes :)
Will add a test for that

Sources_v3/Controllers/ChannelController/ChannelController.swift
mantergo: But if we will look into case similar to the one you provided previously:
```
let controller1 = client.channelController(createChannelWithId: cid)
let controller2 = client.channelController(for: controller1.cid)
```
It will be possible to perform modifications on `controller2`.
I can agree that it's programmers error, but we are helping to avoid that.
And I see it logically correct: we have cid in `ChannelQuery` cause it's valid but `ChannelController` doesn't have one yet cause it's not observing any channel atm.

Sources_v3/Query/ChannelQuery.swift
mantergo: Yep, sorry for that.
Mixed up with `fixups`.

Sources_v3/Controllers/ChannelController/ChannelController.swift
VojtaStavik: The case I originally pointed out with the following:
```swift
let controller1 = client.channelController(createDirectMessageChannelWith: members)
let controller2 = client.channelController(for: controller1.cid)
```
The main difference is that in this case, `controller1` _looks like_ it has a valid `cid` but it doesn't.

The case you present is basically this:
```swift
let newCid = <some new cid>
let controller1 = client.channelController(createChannelWithId: newCid)
let controller2 = client.channelController(for: newCid)
```
If someone does this, then it's a clear misusage of the SDK.

VojtaStavik (approved):
🚢 🇮🇹 🎉

PR #549, opened 2020-10-14
[CIS-363] Sample app improvements

b-onc:
Question: should we also build sample app in iOS12?
Currently, Github Actions has Xcode11.7 as default. [Default Xcode will be changed to Xcode 12.0 on October, 20](https://github.com/actions/virtual-environments/issues/1712), but we should still be fine.

### DIFF
--- a/Sample_v3/ConfigurationViewController.swift
+++ b/Sample_v3/ConfigurationViewController.swift
@@ -19,13 +19,38 @@ class ConfigurationViewController: UITableViewController {
     @IBOutlet var localStorageEnabledSwitch: UISwitch!
     @IBOutlet var flushLocalStorageSwitch: UISwitch!
     
+    private let jwtCellIndexPath = IndexPath(row: 4, section: 0)
+    
     override func viewDidLoad() {
         super.viewDidLoad()
         tokenTypeSegmentedControl.addTarget(self, action: #selector(tokenTypeSegmentedControlDidChangeValue), for: .valueChanged)
     }
     
-    @objc func tokenTypeSegmentedControlDidChangeValue() {
+    @objc func tokenTypeSegmentedControlDidChangeValue(_ control: UISegmentedControl) {
         tableView.beginUpdates()
+        
+        switch control.selectedSegmentIndex {
+        case 1:
+            tokenTypeSegmentedControl.selectedSegmentIndex = 1
+            jwtTextField.isEnabled = false
+            if token != .development {
+                tableView.deleteRows(at: [jwtCellIndexPath], with: .top)
+            }
+            jwtTextField.text = nil
+        case 2:
+            tokenTypeSegmentedControl.selectedSegmentIndex = 2
+            jwtTextField.isEnabled = false
+            if token != "" {
+                tableView.deleteRows(at: [jwtCellIndexPath], with: .top)
+            }
+            jwtTextField.text = .development
+        default:
+            tokenTypeSegmentedControl.selectedSegmentIndex = 0
+            jwtTextField.isEnabled = true
+            jwtTextField.text = Configuration.TestUser.defaults.first(where: { $0.id == userId })?.token ?? ""
+            tableView.insertRows(at: [jwtCellIndexPath], with: .top)
+        }
+        
         tableView.endUpdates()
     }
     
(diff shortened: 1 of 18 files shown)


Sample_v3/ConfigurationViewController.swift
```diff
@@ -52,28 +52,10 @@ class ConfigurationViewController: UITableViewController {
...
-            ),
-            (
-                name: "Steep Moon",
-                id: "steep-moon-9",
-                jwt: "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1c2VyX2lkIjoic3RlZXAtbW9vbi05In0.xwGjOwnTy3r4o2owevNTyzZLWMsMh_bK7e5s1OQ2zXU"
-            )
-        ]
-        
-        if let user = users.randomElement() {
+        if let user = Configuration.TestUser.defaults.shuffled().first(where: { $0.id != userId }) {
```
VojtaStavik, line 80: Nice trick 👍  I'd do `filter { $0.id != userId}.randomElement()` but yours is probably more efficient :)

Sample_v3/Samples/SwiftUISimpleChat/ChatView.swift
```diff
@@ -2,178 +2,185 @@
...
-    @StateObject var channel: ChatChannelController.ObservableObject
-    /// The `text` written in the message composer.
-    @State var text: String = ""
-    /// Binding for message actions and channel actions ActionSheet.
-    @State var actionSheetTrigger: Bool = false
-    /// Message being edited.
-    @State var editingMessage: ChatMessage?
+    import Combine
+    import StreamChatClient
+    import SwiftUI
```
VojtaStavik, line 9: nit: I don't think we need to intend code wrapped inside `#if endif`

Sample_v3/Samples/SwiftUISimpleChat/UserListView.swift
```diff
@@ -2,88 +2,92 @@
...
+            guard encounteredIndex == userList.users.count - 1 else { return }
+            userList.controller.loadNextUsers()
+        }
     }
-}
 
-/// Protocol conformance needed for ActionSheet presenting.
-extension ChatUser: Identifiable {}
+    /// Protocol conformance needed for ActionSheet presenting.
+    extension ChatUser: Identifiable {}
```
VojtaStavik, line 91: I already commented somewhere that this should live inside the SDK. Someone must have ignored it 🙃

VojtaStavik (approved):
Looks good, thanks! What was the fix in `Sample app: fix logged assertion warnings` about? I didn't get it from the commit.

Sample_v3/Samples/SwiftUISimpleChat/ChatView.swift
b-onc: hmm, seems like formatter indented it, not me.  I agree with you, let me check if I can disable that

b-onc:
@VojtaStavik 
> What was the fix in Sample app: fix logged assertion warnings about? I didn't get it from the commit.

We were hiding the token cell in `heightForRowAt` by returning 0 as height, but to check if the cell was token cell, we were doing:
```swift
override func tableView(_ tableView: UITableView, heightForRowAt indexPath: IndexPath) -> CGFloat {
        switch tableView.cellForRow(at: indexPath) {
        case jwtCell:
            return heightForJwtCell()
        default:
            return super.tableView(tableView, heightForRowAt: indexPath)
        }
    }
```
calling `cellForRow(at:)` inside `heightForRowAt` is illegal, so I refactored the logic for hiding the token cell. Now the animation is nicer too :)

Sample_v3/Samples/SwiftUISimpleChat/ChatView.swift
b-onc: > Question: should we also build sample app in iOS12?
> I'd say yes just because we have so many `#if` checks in there :)

Fixed in 5e88f6bb42f75923f4910d6c27fd271c82fa0f08

b-onc: Fixed in f2b08921dda0f6e3d7f7481d91bfd4db3aeb88b3

Sample_v3/Samples/SwiftUISimpleChat/UserListView.swift
b-onc: Fixed in 63e1bca7835606a1113024bab02705576844fd34

Sources_v3/Models/User.swift
```diff
@@ -124,6 +124,9 @@ public extension ChatUser {
     }
 }
 
+/// Protocol conformance needed for ActionSheet presenting.
+extension ChatUser: Identifiable {}
+
```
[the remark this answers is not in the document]
b-onc: Fixed in 04964ffe5b86daa454d624a9ff0d67506f4cab8f
Doesn't need to be public since the original typealias is public

PR #551, opened 2020-10-15
[CIS-376] Make query types more user-friendly

mantergo:
**In this PR:**
- Make `QueryOptions` internal
- Simplify `Pagination` usage

### DIFF
--- a/Sources_v3/APIClient/Endpoints/ChannelEndpoints_Tests.swift
+++ b/Sources_v3/APIClient/Endpoints/ChannelEndpoints_Tests.swift
@@ -41,14 +41,20 @@ final class ChannelEndpoints_Tests: XCTestCase {
     }
     
     func test_channel_buildsCorrectly() {
-        let channelID = ChannelId(type: .livestream, id: "qwerty")
+        let cid = ChannelId(type: .livestream, id: "qwerty")
+        
+        func channelQuery(options: QueryOptions) -> ChannelQuery<DefaultExtraData> {
+            var query: ChannelQuery<DefaultExtraData> = .init(cid: cid)
+            query.options = options
+            return query
+        }
         
         let testCases: [(ChannelQuery<DefaultExtraData>, Bool)] = [
-            (.init(cid: channelID, options: .state), true),
-            (.init(cid: channelID, options: .presence), true),
-            (.init(cid: channelID, options: .watch), true),
-            (.init(cid: channelID, options: .all), true),
-            (.init(cid: channelID, options: []), false)
+            (channelQuery(options: .state), true),
+            (channelQuery(options: .presence), true),
+            (channelQuery(options: .watch), true),
+            (channelQuery(options: .all), true),
+            (channelQuery(options: []), false)
         ]
         
         for (query, requiresConnectionId) in testCases {
--- a/Sources_v3/APIClient/Endpoints/ChannelEndpoints_Tests.swift
+++ b/Sources_v3/APIClient/Endpoints/ChannelEndpoints_Tests.swift
@@ -9,12 +9,18 @@ final class ChannelEndpoints_Tests: XCTestCase {
     func test_channels_buildsCorrectly() {
         let filter: Filter<ChannelListFilterScope<NameAndImageExtraData>> = .containMembers(userIds: [.unique])
         
+        func channelListQuery(options: QueryOptions) -> ChannelListQuery<NameAndImageExtraData> {
+            var query: ChannelListQuery<NameAndImageExtraData> = .init(filter: filter)
+            query.options = options
+            return query
+        }
+        
         let testCases: [(ChannelListQuery<NameAndImageExtraData>, Bool)] = [
-            (.init(filter: filter, options: .state), true),
-            (.init(filter: filter, options: .presence), true),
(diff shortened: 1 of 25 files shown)


Sources_v3/Query/Pagination.swift
```diff
@@ -4,67 +4,60 @@
 
 import Foundation
 
-public typealias Pagination = Set<PaginationOption>
+public extension Int {
```
mantergo, line 7: What should we do with this? 🙂

Sources_v3/Query/ChannelListQuery.swift
```diff
@@ -87,27 +87,24 @@ public struct ChannelListQuery<ExtraData: ChannelExtraData>: Encodable {
     /// A number of messages inside each channel.
     public let messagesLimit: Pagination
     /// Query options.
-    public let options: QueryOptions
+    let options: QueryOptions = [.watch]
```
VojtaStavik, line 90: Let's make it `var` if it's not public? I think it will be easier to work with like that.

Sources_v3/APIClient/Endpoints/ChannelEndpoints_Tests.swift
```diff
@@ -27,32 +27,24 @@ final class ChannelEndpoints_Tests: XCTestCase {
...
+                queryItems: nil,
+                requiresConnectionId: true,
+                body: query
+            )
+        
+        // Build endpoint
+        let endpoint: Endpoint<ChannelPayload<DefaultExtraData>> = .channel(query: query)
+        
+        // Assert endpoint is built correctly
+        XCTAssertEqual(AnyEndpoint(expectedEndpoint), AnyEndpoint(endpoint))
```
VojtaStavik, line 47: This test should stay. It tests that `requiresConnectionId` is set properly based on the options.

Sources_v3/Query/ChannelQuery.swift
```diff
@@ -24,7 +24,7 @@ public struct ChannelQuery<ExtraData: ExtraDataTypes>: Encodable {
     /// A pagination for watchers (see `Pagination`). You can use `.limit` and `.offset`.
     public let watchersPagination: Pagination
     /// A query options.
-    public let options: QueryOptions
+    let options: QueryOptions = .all
```
VojtaStavik, line 27: `let` -> `var`

Sources_v3/Query/UserListQuery.swift
```diff
@@ -76,7 +76,7 @@ public struct UserListQuery<ExtraData: UserExtraData>: Encodable {
     public var pagination: Pagination
     
     /// Query options. By default the query options contain `presence`.
-    var options: QueryOptions = [.presence]
+    let options: QueryOptions = [.presence]
```
VojtaStavik, line 79: 🤷‍♂️ 😄

Sources_v3/Query/Pagination.swift
VojtaStavik: I think we should think about the value if they make sense but, in general, I'd keep the constants around

Sources_v3/Query/Pagination.swift
```diff
@@ -4,67 +4,60 @@
...
+    /// A default users page size.
+    static let usersPageSize = 30
+    /// A default channel members page size.
+    static let channelMembersPageSize = 30
+}
 
-public extension Pagination {
-    var limit: Int? {
-        first(where: { $0.limit != nil })?.limit
+public struct Pagination: Encodable, Equatable {
```
VojtaStavik, line 22: I think we have two "versions" of pagination and we should even distinguish them type-wise (?):
- one should have `pageSize` value and optional `offset`
- the other has optional page size and accepts a set of `greaterThan` `greaterThanOrEqual` `lessThan` `lessThanOrEqual`

Some queries support one and some support the other.

VojtaStavik (changes requested):
I think this the right direction. I'd suggest making the pagination logic less error-prone. @b-onc please chime in

Sources_v3/Query/Pagination.swift
mantergo: Ok, I removed `nextPage***` constants cause we haven't even used them and kept others.

Sources_v3/Query/UserListQuery.swift
mantergo: 😅

Sources_v3/Query/Pagination.swift
```diff
@@ -4,71 +4,75 @@
...
+        try container.encode(pageSize, forKey: .pageSize)
+        try offset.map { try container.encode($0, forKey: .offset) }
     }
 }
 
-public extension KeyedEncodingContainer {
-    mutating func encode(_ value: Pagination, forKey key: Self.Key) throws {
-        let encoder = superEncoder(forKey: key)
-        try value.forEach { try $0.encode(to: encoder) }
+public struct MessagesPagination: Encodable, Equatable {
```
b-onc, line 43: Don't we need `limit` for this too?

mantergo: That is not mentioned in docs and I didn't get confirmation for that 😔

b-onc (approved):

PR #554, opened 2020-10-16
[CIS-388] Expose channel-member-list in SampleApp

evsaev:
**This PR**:
- adds channel-member-list screen to `UIKit + Delegate` sample
- adds channel-member-list screen to `UIKit + Combine` sample
- adds channel-member-list screen to `SwiftUI` sample

Simulator Screen Shot - iPhone 11 Pro Max - 2020-10-16 at 17 44 19

### DIFF
--- a/Sample_v3/Extensions/ChannelMember+Extensions.swift
+++ b/Sample_v3/Extensions/ChannelMember+Extensions.swift
@@ -1,0 +1,38 @@
+//
+// Copyright © 2020 Stream.io Inc. All rights reserved.
+//
+
+import StreamChat
+
+/// Creates formatted string for channel member online status.
+///
+/// Example result: `online` or `4 min ago`
+func createMemberOnlineStatusInfoString(for member: ChatChannelMember) -> String? {
+    if member.isOnline {
+        return "online"
+    } else {
+        return member.lastActiveAt?.timeAgo
+    }
+}
+
+/// Creates formatted string for channel member name and ban status.
+///
+/// Example result: `Luke Skywalker`, or `Luke Skywalker 🚫`, or `Luke Skywalker (You)` for the current user.
+func createMemberNameAndStatusInfoString(for member: ChatChannelMember, isCurrentUser: Bool) -> String {
+    let name = member.name ?? "No name"
+    let banIcon = member.isBanned ? "🚫" : ""
+    
+    let nameInfo: String
+    if isCurrentUser {
+        nameInfo = "\(name) (You)"
+    } else {
+        nameInfo = name
+    }
+    
+    return [nameInfo, banIcon].joined(separator: " ")
+}
+
+/// Creates member role string for channel member if it differs from `.member`.
+func createMemberRoleString(for member: ChatChannelMember) -> String? {
+    member.memberRole != .member ? member.memberRole.rawValue : nil
+}
--- a/Sample_v3/Extensions/Date+Extensions.swift
+++ b/Sample_v3/Extensions/Date+Extensions.swift
@@ -1,0 +1,42 @@
+//
+// Copyright © 2020 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+extension Date {
+    /// The code snippet is taken from [stackoverflow](https://stackoverflow.com/a/44087489)
+    var timeAgo: String? {
+        let now = Date()
+        if #available(iOS 13.0, *) {
+            let formatter = RelativeDateTimeFormatter()
+            formatter.unitsStyle = .full
+            return formatter.localizedString(for: self, relativeTo: now)
+        } else {
+            let calendar = Calendar.current
+            guard
+                let minuteAgo = calendar.date(byAdding: .minute, value: -1, to: now),
+                let hourAgo = calendar.date(byAdding: .hour, value: -1, to: now),
+                let dayAgo = calendar.date(byAdding: .day, value: -1, to: now),
+                let weekAgo = calendar.date(byAdding: .day, value: -7, to: now)
+            else { return nil }
+            
+            if minuteAgo < self {
+                let diff = calendar.dateComponents([.second], from: self, to: now).second ?? 0
+                return "\(diff) sec ago"
+            } else if hourAgo < self {
+                let diff = calendar.dateComponents([.minute], from: self, to: now).minute ?? 0
+                return "\(diff) min ago"
+            } else if dayAgo < self {
+                let diff = calendar.dateComponents([.hour], from: self, to: now).hour ?? 0
+                return "\(diff) hrs ago"
+            } else if weekAgo < self {
+                let diff = calendar.dateComponents([.day], from: self, to: now).day ?? 0
+                return "\(diff) days ago"
+            } else {
+                let diff = calendar.dateComponents([.weekOfYear], from: self, to: now).weekOfYear ?? 0
+                return "\(diff) weeks ago"
+            }
+        }
+    }
+}
--- a/Sample_v3/Extensions/UITableViewController+Cells.swift
+++ b/Sample_v3/Extensions/UITableViewController+Cells.swift
@@ -2,6 +2,7 @@
 // Copyright © 2020 Stream.io Inc. All rights reserved.
 //
 
+import StreamChat
 import UIKit
 
 extension UITableViewController {
@@ -63,6 +64,32 @@ extension UITableViewController {
         } else {
             cell?.textLabel?.text = messageText
         }
+        
+        return cell
+    }
+    
+    func memberCell(_ member: ChatChannelMember, isCurrentUser: Bool) -> UITableViewCell {
+        let cell = tableView
+            .dequeueReusableCell(withIdentifier: "MemberCell") ?? .init(style: .default, reuseIdentifier: "MemberCell")
+        cell.textLabel?.text = createMemberNameAndStatusInfoString(for: member, isCurrentUser: isCurrentUser)
+        cell.textLabel?.textColor = member.name.flatMap(UIColor.forUsername) ?? .black
+        cell.textLabel?.font = UIFontMetrics(forTextStyle: .body).scaledFont(for: .boldSystemFont(ofSize: UIFont.systemFontSize))
+        cell.textLabel?.adjustsFontSizeToFitWidth = true
+
+        cell.detailTextLabel?.text = createMemberOnlineStatusInfoString(for: member)
+        cell.detailTextLabel?.textColor = member.isOnline ? .blue : .lightGray
+        cell.detailTextLabel?.font = UIFontMetrics(forTextStyle: .footnote)
+            .scaledFont(for: .systemFont(ofSize: UIFont.smallSystemFontSize))
+        cell.detailTextLabel?.adjustsFontSizeToFitWidth = true
+
+        let banStatusLabel = UILabel(frame: .init(x: 0, y: 0, width: 100, height: 40))
+        banStatusLabel.text = createMemberRoleString(for: member)
+        banStatusLabel.textAlignment = .right
+        banStatusLabel.textColor = .darkGray
+        banStatusLabel.font = UIFontMetrics(forTextStyle: .footnote)
+            .scaledFont(for: .boldSystemFont(ofSize: UIFont.smallSystemFontSize))
+        banStatusLabel.adjustsFontForContentSizeCategory = true
+        cell.accessoryView = banStatusLabel
         
         return cell
     }
--- a/Sample_v3/Samples/SimpleChat/SimpleChannelMembersViewController.swift
+++ b/Sample_v3/Samples/SimpleChat/SimpleChannelMembersViewController.swift
@@ -182,7 +182,7 @@ class SimpleChannelMembersViewController: UITableViewController, ChatChannelMemb
             query: .init(sort: [.init(key: .lastActivityAt)])
         )
 
-        usersViewController.openDirectMessagesChat = { [weak self] userId in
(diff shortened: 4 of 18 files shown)


Sample_v3/Extensions/Date+Extensions.swift
```diff
@@ -0,0 +1,41 @@
...
+                let diff = calendar.dateComponents([.hour], from: self, to: now).hour ?? 0
+                return "\(diff) hrs ago"
+            } else if weekAgo < self {
+                let diff = calendar.dateComponents([.day], from: self, to: now).day ?? 0
+                return "\(diff) days ago"
+            } else {
+                let diff = calendar.dateComponents([.weekOfYear], from: self, to: now).weekOfYear ?? 0
+                return "\(diff) weeks ago"
+            }
+        }
```
VojtaStavik, line 39: What's the source of this snippet? Please add a link to the source and copyright if needed.

Sample_v3/Extensions/ChannelMember+Extensions.swift
```diff
@@ -0,0 +1,38 @@
...
+    if member.isOnline {
+        return "online"
+    } else {
+        return member.lastActiveAt?.timeAgo
+    }
+}
+
+/// Creates formatted string for channel member name and ban status.
+///
+/// Example result: `Luke Skywalker` or `Luke Skywalker 🚫`
```
VojtaStavik, line 20: ... or `Luke Skywalker (You)` for the current user.

Sample_v3/Extensions/UITableViewController+Cells.swift
```diff
@@ -66,4 +67,24 @@ extension UITableViewController {
...
+        let cell = tableView
+            .dequeueReusableCell(withIdentifier: "MemberCell") ?? .init(style: .default, reuseIdentifier: "MemberCell")
+        cell.textLabel?.text = createMemberNameAndStatusInfoString(for: member, isCurrentUser: isCurrentUser)
+        cell.textLabel?.textColor = member.name.flatMap(UIColor.forUsername) ?? .black
+        cell.textLabel?.font = .boldSystemFont(ofSize: cell.textLabel?.font.pointSize ?? UIFont.systemFontSize)
+        
+        cell.detailTextLabel?.text = createMemberOnlineStatusInfoString(for: member)
+        cell.detailTextLabel?.textColor = member.isOnline ? .blue : .lightGray
+
+        let banStateLabel = UILabel(frame: .init(x: 0, y: 0, width: 100, height: 40))
```
VojtaStavik, line 81: nit: `banStatusLabel`, `state` means something different

Sample_v3/Extensions/UITableViewController+Cells.swift
```diff
@@ -66,4 +67,24 @@ extension UITableViewController {
...
+        banStateLabel.font = .boldSystemFont(ofSize: 12)
```
VojtaStavik, line 82: Does this work properly with the dynamic font settings?

Sample_v3/Samples/SimpleChat/SimpleChannelMembersViewController.swift
```diff
@@ -0,0 +1,201 @@
...
+        guard
+            let me = memberListController.members.first(where: { $0.id == memberListController.client.currentUserId }),
+            me.memberRole != .member,
+            me.id != member.id
+        else { return nil }
+
+        let removeAction = UIContextualAction(style: .destructive, title: "Remove") { [weak self] _, _, completion in
+            guard let self = self else { return }
+            
+            let channelController = self.memberListController.client.channelController(for: cid)
```
VojtaStavik, line 105: nit: Can you create a local variable like this `let client = memberListController.client` in the scope of the function? It will make the code a tiny bit clearer.

Sample_v3/Samples/SimpleChat/SimpleChannelMembersViewController.swift
```diff
@@ -0,0 +1,201 @@
...
+        )
+
+        usersViewController.openDirectMessagesChat = { [weak self] userId in
+            self?.dismiss(animated: true) {
+                guard let self = self else { return }
+                
+                let channelController = self.memberListController.client.channelController(for: self.memberListController.query.cid)
+                channelController.addMembers(userIds: [userId]) { [channelController] _ in
+                    _ = channelController
+                }
```
VojtaStavik, line 195: I don't think this is correct. The direct message channel controller must be created by `channelController(createDirectMessageChannelWith members: Set<UserId>)`.

Sample_v3/Samples/SwiftUISimpleChat/ChannelListView.swift
```diff
@@ -133,7 +133,14 @@ struct ChannelListView: View {
     /// `UserLisView` for users matching the query.
     var userListView: some View {
         let controller = channelList.controller.client.userListController()
-        return UserListView(userList: controller.observableObject)
+        return UserListView(
+            userList: controller.observableObject,
+            didSelectUser: { userId in
+                createDirectMessageChannel(with: userId) {
+                    activeSheet = nil
+                }
```
VojtaStavik, line 141: I don't understand this. We create the controller but we don't use it?

Sample_v3/Samples/SwiftUISimpleChat/ChatView.swift
```diff
@@ -18,23 +18,17 @@ struct ChatView: View {
...
-    @State var userAction: ((String) -> Void)?
-    
+    /// Member list trigger.
+    @State private var isShowingMemberList = false
+
     var body: some View {
         VStack {
             self.messageList().layoutPriority(1)
             self.composerView()
+            NavigationLink(destination: memberList, isActive: $isShowingMemberList) { EmptyView() }
```
VojtaStavik, line 28: I'm that familiar with SwiftUI but the `EmptyView()` looks like a leftover here :) All good if it's intentional.

VojtaStavik (changes requested):
I'm confused by the way we work with DM channels, to be honest 🙃

VojtaStavik:
@evsaev Thank you for the screenshot btw! 🙏

Sample_v3/Extensions/Date+Extensions.swift
evsaev: That's from stackoverflow: https://stackoverflow.com/a/44087489
Do we need to put a link in this case?

VojtaStavik: Yes please :)

Sample_v3/Samples/SimpleChat/SimpleChannelMembersViewController.swift
evsaev: We don't open direct messages here but add the selected user as a channel member.

You are confused because I forgot to rename `openDirectMessagesChat` -> `didSelectUser` in **SimpleUsersViewController**. 

Will do the rename, thanks for careful review!

Sample_v3/Samples/SwiftUISimpleChat/ChatView.swift
evsaev: Me neither but seems it is required. We are using `public init(destination: Destination, isActive: Binding<Bool>, @ViewBuilder label: () -> Label)` and the `label` closure is required.

Do we want to extend the `NavigationLink` to have  `public init(destination: Destination, isActive: Binding<Bool>)` that will hide this label details?

Sample_v3/Samples/SwiftUISimpleChat/ChannelListView.swift
evsaev: `createDirectMessageChannel` creates a new DM channel and calls the completion when it is done. In completion we just hide the action sheet. The DM `controller` is created and synchronized inside `createDirectMessageChannel`

Sample_v3/Samples/SimpleChat/SimpleChannelMembersViewController.swift
VojtaStavik: That was the reason :) Thanks

Sample_v3/Extensions/Date+Extensions.swift
evsaev: fixed in 219be0ce76f13e87d12c0724f8c6b05d98eb8096

Sample_v3/Extensions/ChannelMember+Extensions.swift
evsaev: fixed in e0456857327a8ac49840441c24f7d2216cccb305

Sample_v3/Extensions/UITableViewController+Cells.swift
evsaev: fixed in 9a22e8c091f2816e348792f97aed982af3afd5fe

Sample_v3/Extensions/UITableViewController+Cells.swift
evsaev: fixed in 9a22e8c091f2816e348792f97aed982af3afd5fe for all labels

Sample_v3/Samples/SimpleChat/SimpleChannelMembersViewController.swift
evsaev: fixed in d8708db39a3c56bc1f670ad843c1a3991cf0ec48 and 4f7f9bee45bcaa346ad40a2a13beda87bcce5c36 for both `SimpleXXX` and `CombineSimpleXXX` controllers

Sample_v3/Samples/SwiftUISimpleChat/ChatView.swift
VojtaStavik: All good, please ignore my comment.

Sample_v3/Samples/SimpleChat/SimpleChannelMembersViewController.swift
```diff
@@ -0,0 +1,198 @@
...
+            query: .init(sort: [.init(key: .lastActivityAt)])
+        )
+
+        usersViewController.didSelectUser = { [weak self] userId in
+            self?.dismiss(animated: true) {
+                guard let self = self else { return }
+                
+                let channelController = self.memberListController.client.channelController(for: self.memberListController.query.cid)
+                channelController.addMembers(userIds: [userId]) { [channelController] _ in
+                    _ = channelController
```
VojtaStavik, line 191: The thing you're doing here with `[channelController]` is to keep it alive? This should work out of the box, if not, our API is wrong.

Sample_v3/Samples/SwiftUISimpleChat/ChannelListView.swift
VojtaStavik: Sorry, still not understand. Why do we create the channel if we don't use it? There's no need to create the channel upfront. The channel controller should create it once it's presented.

Sample_v3/Samples/SwiftUISimpleChat/ChannelListView.swift
```diff
@@ -203,6 +210,20 @@ struct ChannelListView: View {
...
+        
+        let newChannelController = try! client.channelController(
+            createDirectMessageChannelWith: [client.currentUserId, userId],
+            extraData: .init()
+        )
+        
+        newChannelController.synchronize { [newChannelController] _ in
+            _ = newChannelController
+            completion()
+        }
```
VojtaStavik, line 225: I don't understand this. Why we create a controller just to create a channel and then throw it away?

Sample_v3/Samples/SwiftUISimpleChat/UserListView.swift
```diff
@@ -31,16 +31,7 @@ struct UserListView: View {
...
-                        .onTapGesture {
-                            let client = userList.controller.client
-                            let directChatUsers = [client.currentUserId, userList.users[index].id]
-                            let channelController = try! client.channelController(
-                                createDirectMessageChannelWith: Set(directChatUsers),
-                                extraData: .init()
-                            )
-                            channelController.synchronize()
-                            presentationMode.wrappedValue.dismiss()
-                        }
```
evsaev, line 43: @VojtaStavik `createDirectMessageChannel` does the same thing that was happening in `UserListView` when a user is selected. The thing is `UserListView` can be opened not only from channel-list, but also from channel-member-list. In the former case we don't want to create 1-1 channel but add selected user to the channel instead.

Answering your question - yes, we create a new channel and throw the controller away. It should open the created 1-1 channel but it was not implemented so I just preserved the current behavior.

Sample_v3/Samples/SimpleChat/SimpleChannelMembersViewController.swift
evsaev: Yes, that was the reason. Currently, `synchronize` doesn't care about keeping a controller alive

VojtaStavik: But you don't call `synchronize` here. You call `addMembers` and it should be possible to successfully do that without keeping the controller alive (like we do it for all other controllers and their mutating actions).

Sample_v3/Samples/SwiftUISimpleChat/UserListView.swift
VojtaStavik: @evsaev If you could fix it, it would be great. The sample code in this form doesn't make much sense and it brings more questions than answers.

VojtaStavik (approved):
I think it looks good 👍 Thank you. We just need to clarify the part with creating a controller without using it.

PR #555, opened 2020-10-19
Fix composerView retaining messages

b-onc:
Steps to reproduce:
1. Enable Network Link Conditioner to 3G speed
2. To any channel, write a message
3. Tap send
4. Immediately after 3, write something else

Expected:
- New text will be input on a empty field

Actual:
- New text is input after the last message (until message is actually sent)

Example:
1. Write "Hello"
2. Tap send
3. Write "Again"

Expected to see only "Again", but seeing "HelloAgain" in the text field

Unfortunately, we can't support optimistic updates for now, so messages appearing in the channel immediately after tapping send is not possible.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Upcoming
 
-### 🔄 Changed
+### 🐞 Fixed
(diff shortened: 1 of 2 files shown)


VojtaStavik:
@b-onc Anything blocking merging this?

b-onc:
@VojtaStavik I'm testing it still, will merge it soon 👍

PR #559, opened 2020-10-21
For #550: Fix replies request pagination not encoded correctly

### LINKED ISSUE #550: Replies pagination not working.
## What did you do?
trying to paginate replies from parent message following [Thread pagination](https://getstream.io/chat/docs/threads/?language=swift&q=thread#thread-pagination)

Replies fetched following pagination parameter.

## What happened instead?
Fetched only the latest replies.

## GetStream Environment
StreamChatClient/StreamChatCore 2.3.3

## Context
To add more context in order to find this bug, I turned on the debug logger and it seems that pagination is not serialized when request is performed.

b-onc:
Since we've changed the type of Pagination, replies endpoint did not encode pagination correctly. This was caused from Pagination being an array, and we don't support encoding arrays as query items.

Fixes #550

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### 🐞 Fixed
 - `ComposerView` disables send button but retains the message after user taps send. Now, composerView resets its state after user taps send, so user can send multiple messages [#555](https://github.com/GetStream/stream-chat-swift/issues/555)
+- Fix pagination for message replies not encoded correctly [#550](https://github.com/GetStream/stream-chat-swift/issues/550)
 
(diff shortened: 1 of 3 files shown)


VojtaStavik (approved):
So the fix is that we encode the pagination options as `Dictionary` instead of `Array`?

b-onc:
@VojtaStavik They're not even encoded when we pass them as array, since we use
In `Client+Request.swift:L191`:
```swift
if let endpointQueryItem = endpoint.queryItem {
  if let data = try? JSONEncoder.default.encode(AnyEncodable(endpointQueryItem)),
     let json = try? JSONSerialization.jsonObject(with: data) as? [String: Any] { // <-- this returns nil
                // actual encoding done here
....
```

VojtaStavik:
@b-onc Gotcha, thanks for the context 👍

PR #560, opened 2020-10-21
[CIS-405] User flagging

evsaev:
**This PR**:
- exposes `flag/unflag` API by `UserController`
- adds `CurrentUserDTO <->> UserDTO` relation for flagged users and updates the models
- adds `flagUser` to `UserUpdater`
- introduces `FlagUserPayload`
- updates mock `UserUpdater` to have a `cleanUp` function

### DIFF
--- a/Sources_v3/APIClient/Endpoints/Payloads/FlagUserPayload_Tests.swift
+++ b/Sources_v3/APIClient/Endpoints/Payloads/FlagUserPayload_Tests.swift
@@ -1,0 +1,71 @@
+//
+// Copyright © 2020 Stream.io Inc. All rights reserved.
+//
+
+@testable import StreamChat
+import XCTest
+
+final class FlagUserPayload_Tests: XCTestCase {
+    func test_json_isDeserialized_withDefaultExtraData() throws {
+        let json = XCTestCase.mockData(fromFile: "FlagUserPayload+DefaultExtraData", extension: "json")
+        let payload = try JSONDecoder.default.decode(FlagUserPayload<DefaultExtraData.User>.self, from: json)
+        
+        // Assert current user payload is deserialized correctly.
+        let currentUser = payload.currentUser
+        XCTAssertEqual(currentUser.id, "broken-waterfall-5")
+        XCTAssertEqual(
+            currentUser.extraData,
+            .init(
+                name: "Broken Waterfall",
+                imageURL: URL(string: "https://s3.amazonaws.com/eventmobi-test-assets/eventsbyids/8024/people/100no-pic.png")
+            )
+        )
+        
+        // Assert flagged user payload is deserialized correctly.
+        let flaggedUser = payload.flaggedUser
+        XCTAssertEqual(flaggedUser.id, "steep-moon-9")
+        XCTAssertEqual(
+            flaggedUser.extraData,
+            .init(
+                name: "Steep Moon",
+                imageURL: URL(string: "https://i.imgur.com/EgEPqWZ.jpg")
+            )
+        )
+    }
+    
+    func test_json_isDeserialized_withNoExtraData() throws {
+        let json = XCTestCase.mockData(fromFile: "FlagUserPayload+NoExtraData", extension: "json")
+        let payload = try JSONDecoder.default.decode(FlagUserPayload<NoExtraDataTypes.User>.self, from: json)
+        
+        // Assert current user payload is deserialized correctly.
+        XCTAssertEqual(payload.currentUser.id, "broken-waterfall-5")
+        
+        // Assert flagged user payload is deserialized correctly.
+        XCTAssertEqual(payload.flaggedUser.id, "steep-moon-9")
+    }
+
+    func test_json_isDeserialized_withCustomData() throws {
+        let json = XCTestCase.mockData(fromFile: "FlagUserPayload+CustomExtraData", extension: "json")
+        let payload = try JSONDecoder.default.decode(FlagUserPayload<TestExtraData>.self, from: json)
+            
+        // Assert current user payload is deserialized correctly.
+        let currentUser = payload.currentUser
+        XCTAssertEqual(currentUser.id, "broken-waterfall-5")
+        XCTAssertEqual(currentUser.extraData.secretNote, "broken-waterfall-5 is Vader ;-)")
+        
+        // Assert flagged user payload is deserialized correctly.
+        let flaggedUser = payload.flaggedUser
+        XCTAssertEqual(flaggedUser.id, "steep-moon-9")
+        XCTAssertEqual(flaggedUser.extraData.secretNote, "Anakin is Vader ;-)")
+    }
+}
+
+private struct TestExtraData: UserExtraData {
+    enum CodingKeys: String, CodingKey {
+        case secretNote = "secret_note"
+    }
+    
+    static let defaultValue = TestExtraData(secretNote: "no secrets")
+    
+    let secretNote: String
+}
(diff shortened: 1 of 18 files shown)


Sources_v3/APIClient/Endpoints/Payloads/FlagUserPayload_Tests.swift
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    func test_json_isDeserialized_withCustomData() throws {
+        let json = XCTestCase.mockData(fromFile: "FlagUserPayload+CustomExtraData", extension: "json")
+        let payload = try JSONDecoder.default.decode(FlagUserPayload<TestExtraData>.self, from: json)
+            
+        // Assert current user payload is deserialized correctly.
+        let currentUser = payload.currentUser
+        XCTAssertEqual(currentUser.id, "broken-waterfall-5")
+        XCTAssertEqual(currentUser.extraData.secretNote, "broken-waterfall-5 is Vader ;-)")
```
VojtaStavik, line 54: Nice 😁

Sources_v3/Models/User.swift
```diff
@@ -34,6 +34,9 @@ public class _ChatUser<ExtraData: UserExtraData> {
     /// An indicator whether the user is banned.
     public let isBanned: Bool
     
+    /// An indicator whether the user is flagged by the current user.
+    public let isFlaggedByCurrentUser: Bool
+
```
VojtaStavik, line 39: I like this API! Let's maybe add one more thing to the documentation comment:
```
/// - Note: Please be aware that the value of this field is not persisted on the server, and is valid only locally for the current session.
```

Sources_v3/Workers/UserUpdater.swift
```diff
@@ -64,6 +64,38 @@ class UserUpdater<ExtraData: ExtraDataTypes>: Worker {
                 }
             }
     }
+    
+    /// Flags or unflags the user with the provided `userId` depending on `flag` value.
+    /// - Parameters:
+    ///   - flag: The indicator saying whether the user should be flagged or unflagged.
+    ///   - userId: The user identifier.
```
VojtaStavik, line 71: _what_ user identifier? The one who flags or the one who is flagged?

Sources_v3/Controllers/UserController/UserController_Tests.swift
```diff
@@ -28,6 +28,8 @@ final class UserController_Tests: StressTestCase {
     }
     
     override func tearDown() {
+        env.userUpdater?.cleanUp()
+
```
VojtaStavik, line 32: Why do we need this? The whole `env` is destroyed on L39

Sources_v3/Controllers/UserController/UserController_Tests.swift
```diff
@@ -393,6 +393,160 @@ final class UserController_Tests: StressTestCase {
...
+        // Simulate `flag` call.
+        controller.flag()
+        
+        // Assert updater is called with correct `flag`
+        XCTAssertEqual(env.userUpdater!.flagUser_flag, true)
+        // Assert updater is called with correct `userId`
+        XCTAssertEqual(env.userUpdater!.flagUser_userId, controller.userId)
+    }
+    
+    func test_flagUser_keepsControllerAlive() {
```
VojtaStavik, line 443: 😊

Sources_v3/Controllers/UserController/UserController_Tests.swift
```diff
@@ -393,6 +393,160 @@ final class UserController_Tests: StressTestCase {
...
+        AssertAsync.staysTrue(weakController != nil)
+        
+        // Simulate successful network response.
+        env.userUpdater!.flagUser_completion!(nil)
+        
+        // Assert completion is called.
+        AssertAsync.willBeTrue(completionIsCalled)
+        
+        // Clean up the mock updater that keeps a callback closure with the controller.
+        env.userUpdater!.cleanUp()
```
VojtaStavik, line 468: I see what you're doing here but I find it a bit confusing. What about doing it in a more straightforward way:
```swift
        // Simulate a successful network response.
         env.userUpdater!.flagUser_completion!(nil)
         // The recorded completion block has a strong reference to controller, we need to reset it:
         env.userUpdater!.flagUser_completion = nil
```

A little bit nicer solution would be to introduce API like:
```swift
extension UserUpdaterMock {
    /// Simulates API response for all recorded completion blocks and resets them.
    func simulateAPICallResponse(_ error: Error?) {
        muteUser_completion?(error)
        muteUser_completion = nil
        
        unmuteUser_completion?(error)
        unmuteUser_completion = nil
        
        // etc ...
    }
}
```

VojtaStavik (approved):
Nice work 👍 I have some comments, nothing serious. Thanks!

Sources_v3/Controllers/UserController/UserController_Tests.swift
evsaev: We need this because controller owns the updater and updater holds a strong reference to the controller by a completion closure. Even though we release the env, we have a retain cycle so controller and updater cannot be released.

Exactly the same we have for all `XXXUpdater` types. We do `apiClient.cleanUp` as the first step in `tearDown`

Sources_v3/Controllers/UserController/UserController_Tests.swift
VojtaStavik: @evsaev I forgot one thing - do you think we could make this test simpler? Remove all unnecessary asserts and things we already test elsewhere? Keep just code related to testing it keeps it alive.

Sources_v3/Controllers/UserController/UserController_Tests.swift
evsaev: Regarding the snippet you shared, I'm not sure about it to be honest: 
1. We don't reset other values related to some method -> there will be a mix of nil and non-nil values.
2. When we do `env.userUpdater!.flagUser_completion!(nil)` we get a crash if the completion we are invoking is nil. It's useful to quickly understand that something is wrong.
3. This method doesn't work if we want to call some completion > 1 time if we performed 2 API calls.
4. we still need `env.userUpdater.cleanUp` in `tearDown` because there are some tests where we check that updater is called with correct values but we don't simulate a response (`test_flagUser_callsUserUpdater_withCorrectValues/test_unflagUser_callsUserUpdater_withCorrectValues`)

VojtaStavik: My main point here is that we keep the completion block just to call it. Once we call it, it's no longer needed. This is IMO implementation detail of the mock and we shouldn't expose it. Ideally, the mock has the same behavior as the real implementation - and the _real_ updater doesn't keep the closure alive after the call is finished.

I hear your points about my snippet and they make sense. Happy to see your alternatives to the issue I mentioned above.

VojtaStavik: If you think it's not worth spending time on time feel free to keep it. The PR is approved :)

Sources_v3/Models/User.swift
evsaev: fixed in 40e3bc6e8e720ed9c958e2bc39e3a4f705990696

Sources_v3/Workers/UserUpdater.swift
evsaev: fixed in 63b93cd2dacb4ad249e2ec74858fc46725feec3d

Sources_v3/Controllers/UserController/UserController_Tests.swift
evsaev: fixed in a700a03db59c86ecd1fe241925124e009f3035c3

Sources_v3/Controllers/UserController/UserController_Tests.swift
evsaev: Yep, I suggest to go with the `cleanUp` we already familiar with. If we decide we are not OK with it, then we can have a ticket and fix it thought all the tests at once

PR #562, opened 2020-10-22
[CIS-406] Message flagging

evsaev:
**This PR:**
- exposes `flagMessage/unflagMessage` by `MessageController`
- expose `flagMessage/unflagMessage` by `MessageUpdater`
- adds `CurrentUserDTO <->> MessageDTO` relation for flagged messages and updates the models
- introduces `FlagMessagePayload` and `flagMessage/unflagMessage` endpoints

### DIFF
--- a/Sources_v3/Models/CurrentUser.swift
+++ b/Sources_v3/Models/CurrentUser.swift
@@ -47,7 +47,16 @@ public class _CurrentChatUser<ExtraData: UserExtraData>: _ChatUser<ExtraData> {
     public let mutedUsers: Set<_ChatUser<ExtraData>>
     
     /// A set of users flagged by the user.
+    ///
+    /// - Note: Please be aware that the value of this field is not persisted on the server,
+    /// and is valid only locally for the current session.
     public let flaggedUsers: Set<_ChatUser<ExtraData>>
+    
+    /// A set of message ids flagged by the user.
+    ///
+    /// - Note: Please be aware that the value of this field is not persisted on the server,
+    /// and is valid only locally for the current session.
+    public let flaggedMessageIDs: Set<MessageId>
     
     /// The unread counts for the current user.
     public let unreadCount: UnreadCount
@@ -65,12 +74,14 @@ public class _CurrentChatUser<ExtraData: UserExtraData>: _ChatUser<ExtraData> {
         currentDevice: Device? = nil,
         mutedUsers: Set<_ChatUser<ExtraData>> = [],
         flaggedUsers: Set<_ChatUser<ExtraData>> = [],
+        flaggedMessageIDs: Set<MessageId> = [],
         unreadCount: UnreadCount = .noUnread
     ) {
         self.devices = devices
         self.currentDevice = currentDevice
         self.mutedUsers = mutedUsers
         self.flaggedUsers = flaggedUsers
+        self.flaggedMessageIDs = flaggedMessageIDs
         self.unreadCount = unreadCount
         
         super.init(
--- a/Sources_v3/Workers/MessageUpdater.swift
+++ b/Sources_v3/Workers/MessageUpdater.swift
@@ -197,6 +197,68 @@ class MessageUpdater<ExtraData: ExtraDataTypes>: Worker {
             }
         }
     }
+    
+    /// Flags or unflags the message with the provided `messageId` depending on `flag` value.
+    /// If the message doesn't exist locally it will be fetched and saved locally first first.
+    ///
+    /// - Parameters:
+    ///   - flag: The indicator saying whether the messageId should be flagged or unflagged.
+    ///   - messageId: The identifier of a messageId that should be flagged or unflagged.
+    ///   - completion: Called when the API call is finished. Called with `Error` if the remote update fails.
+    ///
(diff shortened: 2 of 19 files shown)


Sources_v3/Models/CurrentUser.swift
```diff
@@ -49,6 +49,9 @@ public class _CurrentChatUser<ExtraData: UserExtraData>: _ChatUser<ExtraData> {
     /// A set of users flagged by the user.
     public let flaggedUsers: Set<_ChatUser<ExtraData>>
     
+    /// A set of message ids flagged by the user.
+    public let flaggedMessageIDs: Set<MessageId>
+
```
VojtaStavik, line 54: I just noticed: we have flagged users but flagged message ids. It should probably be `flaggedUserIds`, too. Also, muted user ids.

VojtaStavik: We should mention here that these values are not synced too

Sources_v3/Workers/MessageUpdater.swift
```diff
@@ -197,6 +197,68 @@ class MessageUpdater<ExtraData: ExtraDataTypes>: Worker {
...
+    /// Flags or unflags the message with the provided `messageId` depending on `flag` value.
+    /// If the message doesn't exist locally it will be fetched and saved locally first first.
+    ///
+    /// - Parameters:
+    ///   - flag: The indicator saying whether the messageId should be flagged or unflagged.
+    ///   - messageId: The identifier of a messageId that should be flagged or unflagged.
+    ///   - completion: Called when the API call is finished. Called with `Error` if the remote update fails.
+    ///
+    func flagMessage(_ flag: Bool, with messageId: MessageId, in cid: ChannelId, completion: ((Error?) -> Void)? = nil) {
+        fetchAndSaveMessageIfNeeded(messageId, cid: cid) { error in
```
VojtaStavik, line 210: I have no idea why we do this here 🤔

Sources_v3/Models/CurrentUser.swift
```diff
@@ -49,6 +49,9 @@ public class _CurrentChatUser<ExtraData: UserExtraData>: _ChatUser<ExtraData> {
     /// A set of users flagged by the user.
     public let flaggedUsers: Set<_ChatUser<ExtraData>>
     
+    /// A set of message ids flagged by the user.
+    public let flaggedMessageIDs: Set<MessageId>
```
evsaev, line 53: We cannot have `flaggedMessage: Set<_ChatMessage>` because the current-user's model has **ExtraData: UserExtraData** while the message has **ExtraData: ExtraDataTypes**. 

I suggest to go with `flaggedMessageIDs` and change it to `flaggedMessage` when we update the `CurrentChatUser`:
1. to be a `struct` as other models are
2. to have have `user: _ChatUser` via the composition + @dynamicMemberLookup for the underlaying values
3. to change **ExtraData: UserExtraData** to **ExtraData: ExtraDataTypes**

That's mostly done in c4ad5b7ef383ac126d658bd11d3c41dde0a53f9b

VojtaStavik (commented):
Sorry, I can't review it because I don't understand everything.

Sources_v3/Models/CurrentUser.swift
VojtaStavik: This is a very nice comment that should've been in the PR description before you asked someone for a review 😅

VojtaStavik: BTW - I think the approach with ids is better and exposing the model objects directly was a mistake that slipped through the review.

evsaev: Gotcha 👍 Will also create a ticket for the `flaggedUsers` and `mutedUsers` fix

Sources_v3/Workers/MessageUpdater.swift
evsaev: It's here because it might happen the message does not exist locally -> if there is no message in the database, there is nothing to **add/remove** from `currentUser.flaggedMessages`. 

We have to make sure the message is loaded before we flag/unflag it. I thought it's the right place, because end users can invoke `controller.flag/unflag()` without invoking `controller.synchronize`.

We have nearly the same thing in `ChannelMemberListUpdater` where we fetch and save a channel if needed before queuing the members.

Sources_v3/Models/CurrentUser.swift
evsaev: fixed in 9ac76c106a4f3d6dac95b55c2ae1209e9ace4e20 for both `flaggedUsers` and `flaggedMessageIDs`

Sources_v3/Models/CurrentUser.swift
evsaev: Here it is [CIS-408](https://stream-io.atlassian.net/browse/CIS-408)

VojtaStavik: Thanks 👍

Sources_v3/Workers/MessageUpdater.swift
VojtaStavik: -> Slack

VojtaStavik (approved):
LGTM! Thanks 👍

PR #565, opened 2020-10-23
[CIS-240] Message reactions

evsaev:
**This PR**:
- exposes `addReaction/deleteReaction` by `MessageController`
- updates `MessageDTO` and `ChatMessage` to have reactions info
- introduces DTO and model for `MessageReaction`
- introduces generic `MessageReactionPayload` and makes use of it
- introduce `MessageReactionsMiddleware` that listens for reaction events and updated the database accordingly

### DIFF
--- a/Sources_v3/StreamChat/APIClient/Endpoints/Payloads/MessagePayloads.swift
+++ b/Sources_v3/StreamChat/APIClient/Endpoints/Payloads/MessagePayloads.swift
@@ -50,7 +50,7 @@ struct MessagePayload<ExtraData: ExtraDataTypes>: Decodable {
     
     let latestReactions: [ReactionPayload] = []
     let ownReactions: [ReactionPayload] = []
-    let reactionScores: [String: Int]
+    let reactionScores: [MessageReactionType: Int]
     let isSilent: Bool
     
     init(from decoder: Decoder) throws {
@@ -70,7 +70,9 @@ struct MessagePayload<ExtraData: ExtraDataTypes>: Decodable {
         mentionedUsers = try container.decode([UserPayload<ExtraData.User>].self, forKey: .mentionedUsers)
         replyCount = try container.decode(Int.self, forKey: .replyCount)
         
-        reactionScores = try container.decodeIfPresent([String: Int].self, forKey: .reactionScores) ?? [:]
+        reactionScores = try container
+            .decodeIfPresent([String: Int].self, forKey: .reactionScores)?
+            .mapKeys { MessageReactionType(rawValue: $0) } ?? [:]
         extraData = try ExtraData.Message(from: decoder)
     }
     
@@ -89,7 +91,7 @@ struct MessagePayload<ExtraData: ExtraDataTypes>: Decodable {
         mentionedUsers: [UserPayload<ExtraData.User>],
         replyCount: Int,
         extraData: ExtraData.Message,
-        reactionScores: [String: Int],
+        reactionScores: [MessageReactionType: Int],
         isSilent: Bool
     ) {
         self.id = id
--- a/Sources_v3/StreamChat/APIClient/Endpoints/Payloads/MessagePayloads.swift
+++ b/Sources_v3/StreamChat/APIClient/Endpoints/Payloads/MessagePayloads.swift
@@ -20,8 +20,8 @@ enum MessagePayloadsCodingKeys: String, CodingKey {
     case showReplyInChannel = "show_in_channel"
     case mentionedUsers = "mentioned_users"
     case replyCount = "reply_count"
-    //        case latestReactions = "latest_reactions"
-    //        case ownReactions = "own_reactions"
+    case latestReactions = "latest_reactions"
+    case ownReactions = "own_reactions"
     case reactionScores = "reaction_scores"
     case isSilent = "silent"
     //        case i18n
@@ -48,8 +48,8 @@ struct MessagePayload<ExtraData: ExtraDataTypes>: Decodable {
     // TODO: Attachments
     // TODO: Translations
     
-    let latestReactions: [ReactionPayload] = []
-    let ownReactions: [ReactionPayload] = []
+    let latestReactions: [MessageReactionPayload<ExtraData>]
+    let ownReactions: [MessageReactionPayload<ExtraData>]
     let reactionScores: [MessageReactionType: Int]
     let isSilent: Bool
     
@@ -70,6 +70,8 @@ struct MessagePayload<ExtraData: ExtraDataTypes>: Decodable {
         mentionedUsers = try container.decode([UserPayload<ExtraData.User>].self, forKey: .mentionedUsers)
         replyCount = try container.decode(Int.self, forKey: .replyCount)
         
+        latestReactions = try container.decode([MessageReactionPayload<ExtraData>].self, forKey: .latestReactions)
+        ownReactions = try container.decode([MessageReactionPayload<ExtraData>].self, forKey: .ownReactions)
         reactionScores = try container
             .decodeIfPresent([String: Int].self, forKey: .reactionScores)?
             .mapKeys { MessageReactionType(rawValue: $0) } ?? [:]
@@ -91,6 +93,8 @@ struct MessagePayload<ExtraData: ExtraDataTypes>: Decodable {
         mentionedUsers: [UserPayload<ExtraData.User>],
         replyCount: Int,
         extraData: ExtraData.Message,
+        latestReactions: [MessageReactionPayload<ExtraData>] = [],
+        ownReactions: [MessageReactionPayload<ExtraData>] = [],
         reactionScores: [MessageReactionType: Int],
         isSilent: Bool
     ) {
@@ -108,6 +112,8 @@ struct MessagePayload<ExtraData: ExtraDataTypes>: Decodable {
         self.mentionedUsers = mentionedUsers
         self.replyCount = replyCount
         self.extraData = extraData
+        self.latestReactions = latestReactions
+        self.ownReactions = ownReactions
         self.reactionScores = reactionScores
         self.isSilent = isSilent
     }
--- a/Sources_v3/StreamChat/Database/DTOs/MessageDTO.swift
+++ b/Sources_v3/StreamChat/Database/DTOs/MessageDTO.swift
@@ -29,6 +29,7 @@ class MessageDTO: NSManagedObject {
     @NSManaged var channel: ChannelDTO
     @NSManaged var replies: Set<MessageDTO>
     @NSManaged var flaggedBy: CurrentUserDTO?
+    @NSManaged var reactions: Set<MessageReactionDTO>
     
     // The timestamp the message was created locally. Applies only for the messages of the current user.
     @NSManaged var locallyCreatedAt: Date?
@@ -211,11 +212,14 @@ extension NSManagedObjectContext: MessageDatabaseSession {
         dto.replyCount = Int32(payload.replyCount)
         dto.extraData = try JSONEncoder.default.encode(payload.extraData)
         dto.isSilent = payload.isSilent
-        dto.reactionScores = payload.reactionScores.mapKeys { $0.rawValue }
         dto.channel = ChannelDTO.loadOrCreate(cid: cid, context: self)
         
         let user = try saveUser(payload: payload.user)
         dto.user = user
+        
+        dto.reactionScores = payload.reactionScores.mapKeys { $0.rawValue }
+        let reactions = payload.latestReactions + payload.ownReactions
+        try reactions.forEach { try saveReaction(payload: $0) }
         
         try payload.mentionedUsers.forEach { userPayload in
             let user = try saveUser(payload: userPayload)
@@ -268,6 +272,8 @@ extension MessageDTO {
 
 extension _ChatMessage {
     fileprivate init(fromDTO dto: MessageDTO) {
+        let context = dto.managedObjectContext!
+        
         id = dto.id
         text = dto.text
         type = MessageType(rawValue: dto.type) ?? .regular
@@ -296,12 +302,28 @@ extension _ChatMessage {
         mentionedUsers = Set(dto.mentionedUsers.map { $0.asModel() })
 
         latestReplies = MessageDTO
-            .loadReplies(for: dto.id, limit: 25, context: dto.managedObjectContext!)
+            .loadReplies(for: dto.id, limit: 25, context: context)
             .map(_ChatMessage.init)
         
         localState = dto.localMessageState
         
         isFlaggedByCurrentUser = dto.flaggedBy != nil
+        
+        latestReactions = Set(
+            MessageReactionDTO
+                .loadLatestReactions(for: dto.id, limit: 10, context: context)
+                .map { $0.asModel() }
+        )
+        
+        if let currentUser = context.currentUser() {
+            currentUserReactions = Set(
+                MessageReactionDTO
+                    .loadReactions(for: dto.id, authoredBy: currentUser.user.id, context: context)
+                    .map { $0.asModel() }
+            )
+        } else {
+            currentUserReactions = []
+        }
     }
 }
 
--- a/Sources_v3/StreamChat/Database/DTOs/MessageDTO.swift
+++ b/Sources_v3/StreamChat/Database/DTOs/MessageDTO.swift
@@ -211,7 +211,7 @@ extension NSManagedObjectContext: MessageDatabaseSession {
         dto.replyCount = Int32(payload.replyCount)
         dto.extraData = try JSONEncoder.default.encode(payload.extraData)
         dto.isSilent = payload.isSilent
-        dto.reactionScores = payload.reactionScores
+        dto.reactionScores = payload.reactionScores.mapKeys { $0.rawValue }
         dto.channel = ChannelDTO.loadOrCreate(cid: cid, context: self)
         
         let user = try saveUser(payload: payload.user)
@@ -281,7 +281,7 @@ extension _ChatMessage {
         showReplyInChannel = dto.showReplyInChannel
         replyCount = Int(dto.replyCount)
         isSilent = dto.isSilent
-        reactionScores = dto.reactionScores
+        reactionScores = dto.reactionScores.mapKeys { MessageReactionType(rawValue: $0) }
         
         let extraData: ExtraData.Message
         do {
--- a/Sources_v3/StreamChat/Database/DTOs/MessageDTO_Tests.swift
+++ b/Sources_v3/StreamChat/Database/DTOs/MessageDTO_Tests.swift
@@ -2,6 +2,7 @@
 // Copyright © 2020 Stream.io Inc. All rights reserved.
 //
 
+import CoreData
 @testable import StreamChat
 import XCTest
 
@@ -20,7 +21,16 @@ class MessageDTO_Tests: XCTestCase {
         
         let channelPayload: ChannelPayload<DefaultExtraData> = dummyPayload(with: channelId)
         
-        let messagePayload: MessagePayload<DefaultExtraData> = .dummy(messageId: messageId, authorUserId: userId)
+        let messagePayload: MessagePayload<DefaultExtraData> = .dummy(
+            messageId: messageId,
+            authorUserId: userId,
+            latestReactions: [
+                .dummy(messageId: messageId, user: UserPayload.dummy(userId: .unique))
+            ],
+            ownReactions: [
+                .dummy(messageId: messageId, user: UserPayload.dummy(userId: userId))
+            ]
+        )
         
         // Asynchronously save the payload to the db
         database.write { session in
@@ -34,6 +44,13 @@ class MessageDTO_Tests: XCTestCase {
         // Load the message from the db and check the fields are correct
         var loadedMessage: MessageDTO? {
             database.viewContext.message(id: messageId)
+        }
+        
+        // Load the message reactions from the db
+        var loadedReactions: Set<MessageReactionDTO> {
+            let request = NSFetchRequest<MessageReactionDTO>(entityName: MessageReactionDTO.entityName)
+            request.predicate = .init(format: "message.id == %@", messageId)
+            return Set(try! database.viewContext.fetch(request))
         }
         
         AssertAsync {
@@ -57,6 +74,7 @@ class MessageDTO_Tests: XCTestCase {
                 try? JSONDecoder.default.decode(NoExtraData.self, from: $0.extraData)
             })
             Assert.willBeEqual(messagePayload.reactionScores, loadedMessage?.reactionScores.mapKeys(MessageReactionType.init))
+            Assert.willBeEqual(loadedMessage?.reactions, loadedReactions)
             Assert.willBeEqual(messagePayload.isSilent, loadedMessage?.isSilent)
         }
     }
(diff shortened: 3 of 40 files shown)


Sources_v3/StreamChat/APIClient/Endpoints/Payloads/MessagePayloads.swift
```diff
@@ -70,7 +70,11 @@ struct MessagePayload<ExtraData: ExtraDataTypes>: Decodable {
         mentionedUsers = try container.decode([UserPayload<ExtraData.User>].self, forKey: .mentionedUsers)
         replyCount = try container.decode(Int.self, forKey: .replyCount)
         
-        reactionScores = try container.decodeIfPresent([String: Int].self, forKey: .reactionScores) ?? [:]
+        latestReactions = try container.decode([MessageReactionPayload<ExtraData>].self, forKey: .latestReactions)
+        ownReactions = try container.decode([MessageReactionPayload<ExtraData>].self, forKey: .ownReactions)
+        reactionScores = try container
+            .decodeIfPresent([String: Int].self, forKey: .reactionScores)?
+            .reduce(into: [:]) { $0[.init(rawValue: $1.key)] = $1.value } ?? [:]
```
VojtaStavik, line 77: `[ReactionType: Int].self` doesn't work?

Sources_v3/StreamChat/Database/DTOs/UserDTO.swift
```diff
@@ -18,7 +18,9 @@ class UserDTO: NSManagedObject {
     @NSManaged var userUpdatedAt: Date
     
     @NSManaged var flaggedBy: CurrentUserDTO?
-    
+    @NSManaged var currentUser: CurrentUserDTO?
+    var isCurrentUser: Bool { currentUser != nil }
```
b-onc, line 22: I'm very skeptical of this. In places where we needed this, we always fetched the instance of `CurrentUserDTO` and used it. Doesn't that approach work now?

Sources_v3/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -318,3 +343,7 @@ extension ClientError {
         }
     }
 }
+
+extension Int {
+    static let maxNumberOfLatestReactions = 10
+}
```
b-onc, line 349: IMHO we should limit extending standard types to a minimum, and avoid where unnecessary

Sources_v3/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -302,6 +312,21 @@ extension _ChatMessage {
         localState = dto.localMessageState
         
         isFlaggedByCurrentUser = dto.flaggedBy != nil
+        
+        latestReactions = Set(
+            dto
+                .reactions
```
b-onc, line 318: Won't this still fetch all the reactions? Maybe `.lazy` will save it?

Sources_v3/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -211,12 +213,18 @@ extension NSManagedObjectContext: MessageDatabaseSession {
...
         
         let user = try saveUser(payload: payload.user)
         dto.user = user
         
+        try (payload.latestReactions + payload.ownReactions).forEach {
+            try saveReaction(payload: $0)
+        }
+        dto.reactionScores = payload.reactionScores.reduce(into: [:]) { result, reaction in
+            result[reaction.key.rawValue] = reaction.value
+        }
```
VojtaStavik, line 226: This is here just to convert `String` -> `ReactionType`? If so, it'd be great to add a comment explaining that.

Sources_v3/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -281,7 +289,9 @@ extension _ChatMessage {
         showReplyInChannel = dto.showReplyInChannel
         replyCount = Int(dto.replyCount)
         isSilent = dto.isSilent
-        reactionScores = dto.reactionScores
+        reactionScores = dto.reactionScores.reduce(into: [:]) {
+            $0[.init(rawValue: $1.key)] = $1.value
+        }
```
b-onc, line 294: This is repeated 3 times, we might need a helper. Also I didn't totally understand why we need to `reduce` it

Sources_v3/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -302,6 +312,21 @@ extension _ChatMessage {
...
         
         isFlaggedByCurrentUser = dto.flaggedBy != nil
+        
+        latestReactions = Set(
+            dto
+                .reactions
+                .sorted(by: { $0.updatedAt > $1.updatedAt })
+                .prefix(.maxNumberOfLatestReactions)
+                .map { $0.asModel() }
+        )
```
VojtaStavik, line 322: I think it's better to use a fetch request for this (like we do it for replies). If there are thousands of reactions it makes no sense to load them all to memory, sort them, and then use 10 latest ones.

Sources_v3/StreamChat/Database/DTOs/MessageDTO.swift
VojtaStavik, line 349: This should ideally live in the config. Since we don't do that yet for these limits, just use the number directly in the code. This feels like a done solution and it's not.

Sources_v3/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -302,6 +312,21 @@ extension _ChatMessage {
...
+        
+        currentUserReactions = Set(
+            dto
+                .reactions
+                .filter { $0.user.isCurrentUser }
+                .map { $0.asModel() }
+        )
```
VojtaStavik, line 329: Same here. Better to use fetch request + `CurrentUserDTO.user`

Sources_v3/StreamChat/Database/DTOs/UserDTO.swift
```diff
@@ -18,7 +18,9 @@ class UserDTO: NSManagedObject {
...
+
```
VojtaStavik, line 23: This is not needed. You can get `CurrentUserDTO` from any context by calling `session.currentUser()`

Sources_v3/StreamChat/Database/DatabaseSession.swift
```diff
@@ -149,13 +149,23 @@ protocol MemberListQueryDatabaseSession {
...
 
+protocol MessageReactionDatabaseSesssion {
+    func reaction(messageId: MessageId, userId: UserId, type: ReactionType) throws -> MessageReactionDTO?
+    
+    @discardableResult
+    func saveReaction<ExtraData: ExtraDataTypes>(payload: MessageReactionPayload<ExtraData>) throws -> MessageReactionDTO
+    
+    func delete(reaction: MessageReactionDTO)
+}
+
```
VojtaStavik, line 160: Why do you think we need an extra protocol for this? Isn't this part of `MessageDatabaseSession`?

Sources_v3/StreamChat/WebSocketClient/EventMiddlewares/MessageReactionsMiddleware.swift
```diff
@@ -0,0 +1,38 @@
...
+                try session.saveReaction(payload: event.reaction)
+            } else if let event = reactionEvent as? ReactionUpdatedEvent<ExtraData> {
+                try session.saveReaction(payload: event.reaction)
+            } else if let event = reactionEvent as? ReactionDeletedEvent<ExtraData> {
+                let dto = try session.reaction(
+                    messageId: event.messageId,
+                    userId: event.userId,
+                    type: event.reactionType
+                )
+                dto.flatMap(session.delete)
```
VojtaStavik, line 27: ??

Sources_v3/StreamChat/Database/DTOs/MessageDTO.swift
VojtaStavik: Yep, we need something like `mapValues` but for keys -> `mapKey` :)

Sources_v3/StreamChat/Database/DTOs/UserDTO.swift
evsaev: In this case, we access `UserDTO.isCurrentUser` not from db-session method (where `self` is NSMangedObjectContext) but from `ChatMessage.init(dto: MessageDTO)`.

We can jump around optional `dto.managedObjectContext` or use the `isCurrentUser` helper, but I'm not sure it's better keeping in mind that from performance perspective there is not difference because core-data is lazy and `currentUser` is fetched only when we access it.

Sources_v3/StreamChat/Database/DTOs/MessageDTO.swift
evsaev: Ok, I see your point. Let's have the constant in **MemberDTO** then: 
```swift
private static let maxNumberOfLatestReactions = 10
```

Sources_v3/StreamChat/APIClient/Endpoints/Payloads/MessagePayloads.swift
evsaev: Yep, doesn't work

Sources_v3/StreamChat/Database/DTOs/UserDTO.swift
evsaev: @VojtaStavik WDYT on that?

VojtaStavik: I second to @b-onc here though my main concern is architecture-related. I just don't think it's a good idea to have a back reference from `User` to `CurrentUser`.

b-onc: We're using `dto.managedObjectContext!` in places already:
```swift
latestReplies = MessageDTO
            .loadReplies(for: dto.id, limit: 25, context: dto.managedObjectContext!)
            .map(_ChatMessage.init)
```
and
```swift
        let context = dto.managedObjectContext!
        
        let latestMessages: [_ChatMessage<ExtraData>] = MessageDTO
            .load(for: dto.cid, limit: 25, context: context)
            .map { $0.asModel() 
```
I believe we should use the same approach here

Sources_v3/StreamChat/Database/DTOs/MessageDTO.swift
evsaev: @b-onc already mentioned it'd nice to have a helper. I'm thinking about the extension for `Dictionary` type with `mapKeys` method.

Sources_v3/StreamChat/Database/DTOs/UserDTO.swift
VojtaStavik: -> also please see my comments regarding using fetch requests, i don't think we need anything like this

Sources_v3/StreamChat/Database/DatabaseSession.swift
evsaev: haha, you are right! 😄

Sources_v3/StreamChat/WebSocketClient/EventMiddlewares/MessageReactionsMiddleware.swift
evsaev: Ok 😄 Let me replace it with boring:
```swift
guard let dto = ... else { return }
session.delete(dto)
```

VojtaStavik: Your future yourself will thank me!

VojtaStavik: btw, why do we even have `try` there? Under which circumstances the function will throw?

Sources_v3/StreamChat/Database/DTOs/MessageDTO.swift
VojtaStavik: `mapKeys` sounds great 👍

Sources_v3/StreamChat/APIClient/Endpoints/Payloads/MessagePayloads.swift
VojtaStavik: How come? It should work IMO. Maybe `ReactionType` is not encoded properly using `singleValueContainer`?

VojtaStavik: @evsaev Probably the generated `Codable` conformance encodes it as `{ "rawValue": "😱" }` instead of `"😱"`

Sources_v3/StreamChat/WebSocketClient/EventMiddlewares/MessageReactionsMiddleware.swift
evsaev: It throws the error if there's no `MessageDTO` in the database

Sources_v3/StreamChat/Database/DTOs/UserDTO.swift
evsaev: fixed in 8ac7bb42aac85e9e420e5cfa0450d6536bfbda46

Sources_v3/StreamChat/Database/DTOs/MessageDTO.swift
evsaev: resolved in 8ac7bb42aac85e9e420e5cfa0450d6536bfbda46

Sources_v3/StreamChat/Database/DTOs/MessageDTO.swift
evsaev: `mapKeys` is implemented in 24cc3daf34d39da8aede5d8a21d13a2cf5e120a9

Sources_v3/StreamChat/Database/DTOs/MessageDTO.swift
evsaev: see https://github.com/GetStream/stream-chat-swift/pull/565#discussion_r512732433

Sources_v3/StreamChat/Database/DTOs/MessageDTO.swift
evsaev: fixed in 8ac7bb42aac85e9e420e5cfa0450d6536bfbda46

Sources_v3/StreamChat/Database/DTOs/MessageDTO.swift
evsaev: fixed in 8ac7bb42aac85e9e420e5cfa0450d6536bfbda46

Sources_v3/StreamChat/Database/DTOs/UserDTO.swift
evsaev: replaced with a fetch-request in 8ac7bb42aac85e9e420e5cfa0450d6536bfbda46

Sources_v3/StreamChat/WebSocketClient/EventMiddlewares/MessageReactionsMiddleware.swift
evsaev: updated in a4ea692b67bf9ebbfe08525503d4459ef18b14ac

Sources_v3/StreamChat/Database/DTOs/MessageReactionDTO.swift
```diff
@@ -40,6 +40,32 @@ extension MessageReactionDTO {
...
     }
     
+    static func loadCurrentUserReactions(
+        for messageId: MessageId,
+        context: NSManagedObjectContext
+    ) -> [MessageReactionDTO] {
+        let request = NSFetchRequest<MessageReactionDTO>(entityName: MessageReactionDTO.entityName)
+        request.predicate = NSCompoundPredicate(andPredicateWithSubpredicates: [
+            NSPredicate(format: "message.id == %@", messageId),
+            NSPredicate(format: "user.currentUser != nil")
```
VojtaStavik, line 50: Where does the `user.currentUser` come from? I couldn't find it anywhere in the PR 🤔

Sources_v3/StreamChat/WebSocketClient/Events/ReactionEvents.swift
```diff
@@ -12,15 +12,17 @@ public struct ReactionNewEvent<ExtraData: ExtraDataTypes>: EventWithReactionPayl
...
         cid = try response.value(at: \.cid)
-        messageId = try response.value(at: \.message?.id)
+        messageId = try response.value(at: \.reaction?.messageId)
         reactionType = try response.value(at: \.reaction?.type)
         reactionScore = try response.value(at: \.reaction?.score)
         createdAt = try response.value(at: \.reaction?.createdAt)
+        reaction = try response.value(at: \.reaction)
         payload = response
     }
 }
```
VojtaStavik, line 28: This is not really what want, take a look at `MemberEvent` and `TypingEvent` for example. The fact, that events are generic is wrong and leftover from a wrong implementation. Only the initializer should be generic. 

The payload itself is stored in the type-erased form in `payload`.

Sources_v3/StreamChat/WebSocketClient/EventMiddlewares/MessageReactionsMiddleware_Tests.swift
```diff
@@ -0,0 +1,209 @@
...
+        let event = TestEvent()
+        
+        // Handle non-reaction event
+        let forwardedEvent = try await {
+            self.middleware.handle(event: event, completion: $0)
+        }
+        
+        // Assert event is forwarded as it is
+        XCTAssertEqual(forwardedEvent as! TestEvent, event)
+    }
```
VojtaStavik, line 40: Nice 👍

VojtaStavik (changes requested):
Amazing! Almost 2k lines of changes and it was so easy to review this PR. Great work @evsaev! 👍

Sources_v3/StreamChat/Models/MessageReactionType.swift
```diff
@@ -0,0 +1,18 @@
+//
+// Copyright © 2020 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+/// The type that describes a message reaction type.
+public struct MessageReactionType: RawRepresentable, Codable, Hashable, ExpressibleByStringLiteral {
```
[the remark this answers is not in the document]
evsaev: That's the bug, have a look at https://bugs.swift.org/browse/SR-7788. The ticket describes the issue with encoding but 
the same happens during the decoding process.

One of the suggested workarounds is to have a property wrapper just not sure it's worth it.

Sources_v3/StreamChat/Database/DTOs/MessageDTO_Tests.swift
```diff
@@ -168,6 +209,24 @@ class MessageDTO_Tests: XCTestCase {
...
+        }
+        
+        // Load message reactions left by the current user.
+        var currentUserReactions: Set<_ChatMessageReaction<DefaultExtraData.MessageReaction>> {
+            Set(
+                MessageReactionDTO
+                    .loadCurrentUserReactions(for: messageId, context: database.viewContext)
+                    .map { $0.asModel() }
+            )
+        }
```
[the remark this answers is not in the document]
evsaev: It's mine but it was accepted by the formatter: nothing has changed when the pre-commit hook was triggered 😄

Sources_v3/StreamChat/Database/DTOs/MessageReactionDTO.swift
evsaev: `user.currentUser` is the reverse relation for `currentUser.user`. We don't expose this relation via property but it exists in the db model.

Sources_v3/StreamChat/Models/MessageReactionType.swift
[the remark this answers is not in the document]
evsaev: haha, agree with you here. Will the following work?

> /// The type that describes a message reaction type.
> ///
> /// The reaction has underlaying type `String` what gives the flexibility to choose the way how the reaction
> /// will be displayed in the application.
> ///
> /// Common examples are: "like", "love", "smile", etc.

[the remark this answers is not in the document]
VojtaStavik: @evsaev Good 👍

Sources_v3/StreamChat/Database/DTOs/MessageReactionDTO.swift
VojtaStavik: You can't use a nasty trick like this and not even mention it in a comment! 😅 I find this very dangerous, to be honest. I think a more _boring_ approach with getting the current user DTO first and use it's userId would be better here. @b-onc ?

b-onc: I wasn't aware this could be done 🤯 Agreed though, let's keep it in the back of our heads (that this is also a viable solution) but my vote is for the boring approach.

Sources_v3/StreamChat/Models/MessageReactionType.swift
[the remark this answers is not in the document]
evsaev: fixed in bb301d3f5c9ecb755b3c95550a9660f0c668dd69

Sources_v3/StreamChat/Models/MessageReaction.swift
```diff
@@ -0,0 +1,58 @@
...
+    public let createdAt: Date
+    
+    /// The date the reaction was last updated.
+    public let updatedAt: Date
+    
+    /// The reaction's extra data.
+    public let extraData: ExtraData
+    
+    /// The reaction's author identifier.
+    public let authorId: UserId
```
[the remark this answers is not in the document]
evsaev: fixed in 9f2cf54ba88600dc0f97f39cd678eb2c997c98d7

Sources_v3/StreamChat/Database/DTOs/MessageReactionDTO.swift
evsaev: fixed in bf46e6ec6d59ac8a40deb6a06142e0ff179fe5e2

Sources_v3/StreamChat/WebSocketClient/Events/ReactionEvents.swift
evsaev: fixed in a788b191cbddc1055753012de3cc2b8945d34c9f

Sources_v3/StreamChat/Models/MessageReactionType.swift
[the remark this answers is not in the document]
VojtaStavik: As mentioned in my original comment, can you please add tests that verify that `MessageReactionType(rawValue: "love")` is encoded as `"love"` end vice-versa with decoding?

VojtaStavik (approved):

Sources_v3/StreamChat/Models/MessageReactionType.swift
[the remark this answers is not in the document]
evsaev: fixed in 71e05093ab0308b60e0ad5157d875a1b86509ec7

PR #567, opened 2020-10-26
Sample v3: Use non-inverted LazyVStack in SwiftUI sample's ChatView

cardoso:
This PR reworks the approach in the SwiftUI sample's chat view to use the .bottomToTop list ordering in ChatController and use a LazyVStack with a custom ScrollView wrapper instead of an inverted List which was causing visual glitches in the context menu.

#no_changelog

### DIFF
--- a/Sample_v3/Samples/SwiftUISimpleChat/ChannelListView.swift
+++ b/Sample_v3/Samples/SwiftUISimpleChat/ChannelListView.swift
@@ -203,11 +203,13 @@ struct ChannelListView: View {
     }
     
     private func chatView(for index: Int) -> ChatView {
-        ChatView(
-            channel: channelList.controller.client.channelController(
-                for: channelList.channels[index].cid
-            ).observableObject
+        let channelController = channelList.controller.client.channelController(
+            for: channelList.channels[index].cid
         )
+        
+        channelController.listOrdering = .bottomToTop
+        
+        return ChatView(channel: channelController.observableObject)
     }
     
     private func openDirectMessages(with userId: UserId) {
--- a/Sample_v3/Samples/SwiftUISimpleChat/ChatScrollView.swift
+++ b/Sample_v3/Samples/SwiftUISimpleChat/ChatScrollView.swift
@@ -4,7 +4,8 @@
 
 import SwiftUI
 
-@available(iOS 13, *)
+/// `ChatScrollView` implements the bottom-to-top scrolling behavior common to chats.
+@available(iOS 14, *)
 struct ChatScrollView<Content>: View where Content: View {
     var content: () -> Content
 
@@ -90,7 +91,7 @@ extension ViewHeightKey: ViewModifier {
     }
 }
 
-@available(iOS 13, *)
+@available(iOS 14, *)
 extension View {
     func onKeyboardAppear(_ callback: @escaping () -> Void) -> some View {
         onAppear {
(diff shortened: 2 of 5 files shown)


Sample_v3/Samples/SwiftUISimpleChat/ChannelListView.swift
```diff
@@ -203,11 +203,13 @@ struct ChannelListView: View {
...
     private func chatView(for index: Int) -> ChatView {
-        ChatView(
-            channel: channelList.controller.client.channelController(
-                for: channelList.channels[index].cid
-            ).observableObject
+        let channelController = channelList.controller.client.channelController(
+            for: channelList.channels[index].cid
         )
+        
+        channelController.listOrdering = .bottomToTop
```
VojtaStavik, line 210: IMO because `ChatView` expects the ordering to be in a particular form, it should be `ChatView` who should set the ordering.

Sample_v3/Samples/SwiftUISimpleChat/ChatScrollView.swift
```diff
@@ -0,0 +1,103 @@
+//
+// Copyright © 2020 Stream.io Inc. All rights reserved.
+//
+
+import SwiftUI
+
+@available(iOS 13, *)
```
VojtaStavik, line 7: 13 or 14? 🤔

Sample_v3/Samples/SwiftUISimpleChat/ChatScrollView.swift
```diff
@@ -0,0 +1,103 @@
...
+struct ChatScrollView<Content>: View where Content: View {
```
VojtaStavik, line 8: Please add a comment why we have/need this.

Sample_v3/Samples/SwiftUISimpleChat/ChatScrollView.swift
```diff
@@ -0,0 +1,103 @@
...
+            content()
+        }
+        .modifier(ViewHeightKey())
+        .onPreferenceChange(ViewHeightKey.self) {
+            self.updateHeight(with: $0, outerHeight: geometry.size.height)
+        }
+        .frame(height: geometry.size.height, alignment: .bottom)
+        .offset(y: contentOffset + scrollOffset)
+        .animation(.easeInOut)
+        .background(Color.white)
```
VojtaStavik, line 33: Does this work with dark/light theme?

Sample_v3/Samples/SwiftUISimpleChat/ChatView.swift
```diff
@@ -48,12 +48,17 @@ struct ChatView: View {
...
-        List(channel.messages, id: \.self) { message in
-            self.messageView(for: message)
+        ChatScrollView {
+            ScrollViewReader { scrollView in
+                LazyVStack(alignment: .leading) {
+                    ForEach(channel.messages, id: \.self) { message in
+                        self.messageView(for: message)
+                    }
+                }.onKeyboardAppear {
+                    scrollView.scrollTo(channel.messages[0])
```
VojtaStavik, line 58: Please add a comment why

VojtaStavik (approved):
I don't have Xcode12 on my laptop so I couldn't test the actual feeling of this to see how close the behavior is to the native SrollView. Code-wise this looks good 👍

Sample_v3/Samples/SwiftUISimpleChat/ChannelListView.swift
cardoso: Tried setting it in onAppear of ChatView before synchronize, but it causes the list to visibly change the ordering

PR #572, opened 2020-10-28
Fix plus signs in User fields becoming spaces in the set user request

cardoso:
The default encoding when building the query parameter with `URLQueryItem` will not encode the + symbols, thus making them become spaces. I fixed it by appending the `json` query parameter manually with the proper percent encoding to include the plus sign.

This can be specially annoying if a field is a base64 string, like an encoded public key, but it's also preventing creating a user with a plus in the name which is valid in other clients such as the JS client.

Example:
mhPcgwhH8L8FqFni20qLZ/bIHbwleWvaqm+zcByNp4FWQj0ufl8H9cHV/pbAYo74gCEUdH/MxXdtX+SmW2o3Rg==
becomes
mhPcgwhH8L8FqFni20qLZ/bIHbwleWvaqm zcByNp4FWQj0ufl8H9cHV/pbAYo74gCEUdH/MxXdtX SmW2o3Rg==
which is invalid base64

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Upcoming
 
-### 🔄 Changed
+### 🐞 Fixed
+- Fix plus signs in User fields becoming spaces in the set user request, preventing names with + signs, base64 custom fields, etc. [#572](https://github.com/GetStream/stream-chat-swift/issues/572)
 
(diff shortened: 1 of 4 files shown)


VojtaStavik:
@cardoso I added a test for a similar case for v3. It seems to be working fine. Can you confirm the change in a4124f4 is the problem you observed?

b-onc (approved):

cardoso:
@VojtaStavik the bug also existed in v3. Your test was passing because you weren't checking the final absolute string to see if the plus was encoded. Should be fine now.

PR #575, opened 2020-10-29
Change the workaround solution for encoding `+` in URLs

VojtaStavik:
This is a different solution for the current v3 workaround for the incorrect¹ encoding of the `+` character in the URL query. The description of the problem and other solutions can be found here https://stackoverflow.com/questions/43052657/encode-using-urlcomponents-in-swift .

@cardoso, thank you for fixing this! Great work. I just took your solution and tests and made it fit a tiny bit better into the rest of the SDK.

---
¹ _Technically, the encoding is actually correct, `+` is an allowed character in the URL query, it's just not understood as `+`_.


b-onc (approved):

cardoso (approved):
This looks good 👍

PR #581, opened 2020-11-05
Sample v3: Ask for channel name when creating channel

cardoso:
Show an alert with textfield after pressing add channel button

#no_changelog


mantergo (approved):
👍

PR #587, opened 2020-11-13
[CIS-414] Implement layout helpers

b-onc:
I wanted to implement something like SnapKit, while not having SnapKit and not building our own DSL from scratch... `NSLayoutAnchor` is not really expandable.

I could only come up with this small change, which allows us to:
```swift
let box = UIView()
let miniBox = UIView()

var widthConstraint: NSLayoutConstraint? = box.widthAnchor.constraint(equalToConstant: 50)

box.activateConstraints {
    widthConstraint!
    box.leftAnchor.constraint(equalTo: miniBox.leftAnchor)
    box.rightAnchor.constraint(equalTo: miniBox.rightAnchor)
}
```

We can't do `box.leftAnchor.rightAnchor.constraint(equalToConstant: 10)` unfortunately.

So I've opened this PR as a discussion starter: What helpers are we seeing, other than what we currently have (`embed` and `embedUsingSystemSpacing`) that we can have?

### DIFF
--- a/Sources_v3/StreamChatUI/Utils/UIView+Extensions.swift
+++ b/Sources_v3/StreamChatUI/Utils/UIView+Extensions.swift
@@ -5,6 +5,8 @@
 import UIKit
 
 extension UIView {
+    // MARK: - `embed` family of helpers
+    
     func embed(_ subview: UIView, insets: NSDirectionalEdgeInsets = .zero) {
         addSubview(subview)
         subview.translatesAutoresizingMaskIntoConstraints = false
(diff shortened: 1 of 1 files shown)


VojtaStavik:
Hmm, to be honest, I was just hoping for something like:
```
label.pin(to: view.layoutMarginsGuide, edges: [.top, .left]) // pin top and left edges of label to view margin guides
label.pin(to: view) // pin edges of label to edges of view
```

VojtaStavik:
... and maybe even this
```swift
label.pin(width: 150) // set width to 150
label.pin(widthTo: view.widthAnchor) // set width same as view width
```

b-onc:
@VojtaStavik good pointers, thanks 👍

b-onc:
@VojtaStavik I've updated the PR with new helpers. The issues I see are:
1. It's possible to do nonsense: `label.pin(anchors: [.left], to: 250)`, it won't take affect though.
2. Not possible to pass in constants to constraints between 2 views.
3. Not possible to pass in multipliers.

Biggest issue I've faced is, we can't just return `NSLayoutAnchor` from a function. So we require multiple `makeConstraint` funcs in the `LayoutAnchorName` enum, rather than a single `anchor(for view: UIView) -> NSLayoutAnchor` func.

I guess we'll figure stuff out as we move forward?

VojtaStavik (approved):
I would suggest removing the `_functionBuilder` code because this is production code and it shouldn't be using officially unreleased things.
The `pin` helpers look OK. The possible edge cases are fine for now given that this is just our internal helper. Let's merge and move forward 🚢 🇮🇹 Thank you!!

PR #588, opened 2020-11-13
[CIS-421] Layout for channel-list

evsaev:
**This PR**:
- makes `ChatChannelListCollectionViewCell` self-sizing
- updates `ChatChannelListCollectionViewLayout` to incapsulate spacing and estimated item size
- adds dynamic font and layout margins support to channel list
- updates the way `ChatChannelListVC` loads next channels and updates the list
- fixes layout helpers

### DIFF
--- a/Sources_v3/StreamChatUI/ChatChannelList/ChatChannelListCollectionView.swift
+++ b/Sources_v3/StreamChatUI/ChatChannelList/ChatChannelListCollectionView.swift
@@ -6,7 +6,7 @@ import Foundation
 import UIKit
 
 open class ChatChannelListCollectionView: UICollectionView {
-    public required init(layout: ChatChannelListCollectionViewLayout) {
+    public required init(layout: UICollectionViewLayout) {
         super.init(frame: .zero, collectionViewLayout: layout)
     }
     
--- a/Sources_v3/StreamChatUI/ChatChannelList/ChatChannelListCollectionViewCell.swift
+++ b/Sources_v3/StreamChatUI/ChatChannelList/ChatChannelListCollectionViewCell.swift
@@ -5,4 +5,36 @@
 import Foundation
 import UIKit
 
-open class ChatChannelListCollectionViewCell: UICollectionViewCell {}
+open class ChatChannelListCollectionViewCell<ExtraData: UIExtraDataTypes>: UICollectionViewCell {
+    
+    // MARK: - Properties
+    
+    var uiConfig: UIConfig<ExtraData> = .default
+    
+    public private(set) lazy var channelView: ChatChannelView<ExtraData> = {
+        let view = uiConfig.channelList.channelView.init(uiConfig: uiConfig)
+        contentView.embed(view)
+        return view
+    }()
+    
+    // MARK: - Layout
+    
+    override open func preferredLayoutAttributesFitting(
+        _ layoutAttributes: UICollectionViewLayoutAttributes
+    ) -> UICollectionViewLayoutAttributes {
+        let preferredAttributes = super.preferredLayoutAttributesFitting(layoutAttributes)
+        
+        let targetSize = CGSize(
+            width: layoutAttributes.frame.width,
+            height: UIView.layoutFittingCompressedSize.height
+        )
+        
+        preferredAttributes.frame.size = contentView.systemLayoutSizeFitting(
+            targetSize,
+            withHorizontalFittingPriority: .required,
+            verticalFittingPriority: .fittingSizeLevel
+        )
+        
+        return preferredAttributes
+    }
+}
--- a/Sources_v3/StreamChatUI/ChatChannelList/ChatChannelListCollectionViewLayout.swift
+++ b/Sources_v3/StreamChatUI/ChatChannelList/ChatChannelListCollectionViewLayout.swift
@@ -6,11 +6,34 @@ import Foundation
 import UIKit
 
 open class ChatChannelListCollectionViewLayout: UICollectionViewFlowLayout {
-    override public required init() {
+    
+    // MARK: - Init & Deinit
+    
+    override public init() {
         super.init()
+        commonInit()
     }
     
     public required init?(coder: NSCoder) {
         super.init(coder: coder)
+        commonInit()
+    }
+    
+    // MARK: - Overrides
+    
+    override open func prepare() {
+        super.prepare()
+        
+        estimatedItemSize = .init(
+            width: collectionView?.bounds.width ?? 0,
+            height: 60
+        )
+    }
+    
+    // MARK: - Private
+    
+    private func commonInit() {
+        minimumInteritemSpacing = 0
+        minimumLineSpacing = 0
     }
 }
--- a/Sources_v3/StreamChatUI/ChatChannelList/ChatChannelListCollectionViewLayout.swift
+++ b/Sources_v3/StreamChatUI/ChatChannelList/ChatChannelListCollectionViewLayout.swift
@@ -6,6 +6,7 @@ import Foundation
 import UIKit
 
 open class ChatChannelListCollectionViewLayout: UICollectionViewFlowLayout {
+    private var stopObserving: (() -> Void)?
     
     // MARK: - Init & Deinit
     
@@ -17,6 +18,10 @@ open class ChatChannelListCollectionViewLayout: UICollectionViewFlowLayout {
     public required init?(coder: NSCoder) {
         super.init(coder: coder)
         commonInit()
+    }
+    
+    deinit {
+        stopObserving?()
     }
     
     // MARK: - Overrides
@@ -35,5 +40,26 @@ open class ChatChannelListCollectionViewLayout: UICollectionViewFlowLayout {
     private func commonInit() {
         minimumInteritemSpacing = 0
         minimumLineSpacing = 0
+        
+        subscribeToNotifications()
+    }
+    
+    private func subscribeToNotifications() {
+        let center = NotificationCenter.default
+        
+        center.addObserver(
+            self,
+            selector: #selector(didChangeContentSizeCategory),
+            name: UIContentSizeCategory.didChangeNotification,
+            object: nil
+        )
+        
+        stopObserving = { [unowned self] in
+            center.removeObserver(self)
+        }
+    }
+    
+    @objc private func didChangeContentSizeCategory(_ notification: Notification) {
+        invalidateLayout()
     }
 }
--- a/Sources_v3/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
+++ b/Sources_v3/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
@@ -62,12 +62,11 @@ open class ChatChannelListVC<ExtraData: UIExtraDataTypes>: UIViewController,
         didSelectChannel(channel)
     }
     
-    public func collectionView(
-        _ collectionView: UICollectionView,
-        willDisplay cell: UICollectionViewCell,
-        forItemAt indexPath: IndexPath
-    ) {
-        guard indexPath.row == controller.channels.count - 1 else { return }
+    // MARK: - UIScrollViewDelegate
+    
+    public func scrollViewDidEndDecelerating(_ scrollView: UIScrollView) {
+        let bottomEdge = scrollView.contentOffset.y + scrollView.bounds.height
+        guard bottomEdge >= scrollView.contentSize.height else { return }
         controller.loadNextChannels()
     }
 }
--- a/Sources_v3/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
+++ b/Sources_v3/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
@@ -34,15 +34,6 @@ open class ChatChannelListVC<ExtraData: UIExtraDataTypes>: UIViewController,
         controller.synchronize()
     }
     
-    override open func viewWillLayoutSubviews() {
-        super.viewWillLayoutSubviews()
-        
-        (collectionView.collectionViewLayout as! UICollectionViewFlowLayout).itemSize = .init(
-            width: view.bounds.width,
-            height: 70
-        )
-    }
-    
     // MARK: - UICollectionViewDataSource
     
     public func collectionView(_ collectionView: UICollectionView, numberOfItemsInSection section: Int) -> Int {
--- a/Sources_v3/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
+++ b/Sources_v3/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
@@ -78,6 +78,19 @@ extension ChatChannelListVC: _ChatChannelListControllerDelegate {
         _ controller: _ChatChannelListController<ExtraData>,
         didChangeChannels changes: [ListChange<_ChatChannel<ExtraData>>]
     ) {
-        collectionView.reloadData()
+        collectionView.performBatchUpdates({
+            for change in changes {
+                switch change {
+                case let .insert(_, index):
+                    collectionView.insertItems(at: [index])
+                case let .move(_, fromIndex, toIndex):
+                    collectionView.moveItem(at: fromIndex, to: toIndex)
+                case let .remove(_, index):
+                    collectionView.deleteItems(at: [index])
+                case let .update(_, index):
+                    collectionView.reloadItems(at: [index])
+                }
+            }
+        }, completion: nil)
     }
 }
--- a/Sources_v3/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
+++ b/Sources_v3/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
@@ -44,16 +44,13 @@ open class ChatChannelListVC<ExtraData: UIExtraDataTypes>: UIViewController,
         _ collectionView: UICollectionView,
         cellForItemAt indexPath: IndexPath
     ) -> UICollectionViewCell {
-        let channel = controller.channels[indexPath.row]
-        
-        let cell = collectionView.dequeueReusableCell(withReuseIdentifier: "Cell", for: indexPath)
-        
-        if let channelView = cell.contentView.subviews.first as? ChatChannelView<ExtraData> {
-            channelView.channel = channel
-        } else {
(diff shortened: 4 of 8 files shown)


Sources_v3/StreamChatUI/ChatChannelList/ChatChannelListCollectionViewCell.swift
```diff
@@ -5,4 +5,62 @@
...
+            
+            attributes.frame.size = contentView.systemLayoutSizeFitting(
+                targetSize,
+                withHorizontalFittingPriority: .required,
+                verticalFittingPriority: .defaultLow
+            )
+        }
+        
+        return attributes
+    }
```
evsaev, line 65: As we discussed the cells should be **self-sizing** in order to grow depending on the content and/or accessibility changes. To have self-sizing cells we need all power of **auto-layout**. 

At the same time we want to have a collection view layout being overridable by end users. Custom collection view layouts usually depend on **manual cell attributes calculation** which doesn't work for us since we need auto-layout.

One of the ways to meet all the requirements is to have a cell that calculates its attributes by itself. We have nice separation of concerns then:
1. `ChannelView` is responsible for the content
2. `ChannelViewCell` is responsible for how the `ChannelView` is laid out
3. `ChanneListLayout` is responsible for relations in-between multiple `ChannelViewCell` (like spacing)

Sources_v3/StreamChatUI/ChatChannelList/ChatChannelListCollectionViewCell.swift
```diff
@@ -5,4 +5,62 @@
 import Foundation
 import UIKit
 
-open class ChatChannelListCollectionViewCell: UICollectionViewCell {}
+open class ChatChannelListCollectionViewCell<ExtraData: UIExtraDataTypes>: UICollectionViewCell {
+    // MARK: - Properties
+    
+    public weak var collectionView: UICollectionView?
```
evsaev, line 11: We have a **collectionView: UICollectionView?** instead of **collectionViewSize: CGSize?** to make the `layout` accessible in overridden `preferredLayoutAttributesFitting` in cell subclasses.

Sources_v3/StreamChatUI/ChatChannelList/ChatChannelView.swift
```diff
@@ -90,13 +89,13 @@ open class ChatChannelView<ExtraData: UIExtraDataTypes>: UIView {
...
     open func setupAppearance() {
-        channelNameLabel.numberOfLines = 1
+        channelNameLabel.numberOfLines = 0
         channelNameLabel.font = UIFontMetrics(forTextStyle: .subheadline)
             .scaledFont(for: .boldSystemFont(ofSize: UIFont.systemFontSize))
     }
     
     open func setupLayout() {
-        embedUsingSystemSpacing(container)
+        embed(container, insets: directionalLayoutMargins)
```
evsaev, line 98: Used `embed(container, insets: directionalLayoutMargins)` because `embedUsingSystemSpacing(container)` with negative multipliers doesn't seem to work. Looks like the multipliers get clamped to the `0...1` range.

Sources_v3/StreamChatUI/Utils/UIView+Extensions.swift
```diff
@@ -10,18 +10,18 @@ extension UIView {
...
     }
     
     func embedUsingSystemSpacing(
         _ subview: UIView,
         topInsetMultiplier: CGFloat = 1,
-        bottomInsetMultiplier: CGFloat = 1,
+        bottomInsetMultiplier: CGFloat = -1,
         leadingInsetMultiplier: CGFloat = 1,
-        trailingInsetMultiplier: CGFloat = 1
+        trailingInsetMultiplier: CGFloat = -1
```
evsaev, line 24: This helper doesn't work properly (take a look at https://github.com/GetStream/stream-chat-swift/pull/588/files#r523258367).

Sources_v3/StreamChatUI/ChatChannelList/ChatChannelListCollectionViewLayout.swift
```diff
@@ -6,11 +6,19 @@ import Foundation
...
+        commonInit()
     }
     
     public required init?(coder: NSCoder) {
         super.init(coder: coder)
+        commonInit()
+    }
+    
+    private func commonInit() {
+        estimatedItemSize = UICollectionViewFlowLayout.automaticSize
```
VojtaStavik, line 20: This is not entirely true. We know that the width will always be the full width, only the heigh is automatic.

Sources_v3/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
```diff
@@ -61,15 +46,19 @@ open class ChatChannelListVC<ExtraData: UIExtraDataTypes>: UIViewController,
...
+            for: indexPath
+        ) as! ChatChannelListCollectionViewCell<ExtraData>
         
-        if let channelView = cell.contentView.subviews.first as? ChatChannelView<ExtraData> {
+        if let channelView = cell.channelView {
             channelView.channel = channel
         } else {
-            let channelView = uiConfig.channelList.channelView.init(channel: channel, uiConfig: uiConfig)
-            cell.contentView.embed(channelView)
+            cell.channelView = uiConfig.channelList.channelView.init(channel: channel, uiConfig: uiConfig)
```
VojtaStavik, line 57: Yeah, this shouldn't be needeed. The channel view should always be there. What is the reason for having the cell without its content?

Sources_v3/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
```diff
@@ -61,15 +46,19 @@ open class ChatChannelListVC<ExtraData: UIExtraDataTypes>: UIViewController,
...
         }
         
+        cell.collectionView = collectionView
+
```
VojtaStavik, line 61: I hope we will not need to do this tbh.

Sources_v3/StreamChatUI/Utils/UIView+Extensions.swift
VojtaStavik: I have no idea what's happening here, to be honest :) Why do we need negative multipliers? You just need to swap the order in which way you set up the constraints, don't you?
aka:
A.contraint(B, -1) == B.constraint(A, 1)

Maybe I'm wrong but I think that's how it should work :)

Sources_v3/StreamChatUI/ChatChannelList/ChatChannelView.swift
VojtaStavik: I think the container view should be added to the hierarchy outside of `setupLayout()`. If you override this without calling super, the container view should still be there in the correct position imo. Ideally. you'd just change the way the subviews are stacked there.

VojtaStavik: It seems to me that the main stack view should be pinned directly to the edges and we should use something like (not tested):
```
container.preservesSuperviewLayoutMargins = true
container.isLayoutMarginsRelativeArrangement = true
```

This way, the views inside the container will respect the layout margins of the cell. If users don't want to respect this, they can easily turn it off and layout the views edge-to-edge

VojtaStavik (changes requested):
Looks good 👍 I have just a couple of suggestions, mainly to make it more flexible.

Sources_v3/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
evsaev: I've taken the cell from git stash as it was.

Pavel L. took the task about container cell and working on it right now but it looks like we have to stay synchronized on that topic.

Sources_v3/StreamChatUI/ChatChannelList/ChatChannelListCollectionViewLayout.swift
evsaev: Good point! I think we can set `estimatedItemSize` in `prepare` to ensure we get the width from collection view.

Sources_v3/StreamChatUI/ChatChannelList/ChatChannelListCollectionViewCell.swift
```diff
@@ -5,4 +5,62 @@
...
+        commonInit()
+    }
+
+    private func commonInit() {
+        contentView.translatesAutoresizingMaskIntoConstraints = false
+        NSLayoutConstraint.activate([
+            contentView.leadingAnchor.constraint(equalTo: leadingAnchor),
+            contentView.trailingAnchor.constraint(equalTo: trailingAnchor),
+            contentView.topAnchor.constraint(equalTo: topAnchor),
+            contentView.bottomAnchor.constraint(equalTo: bottomAnchor)
```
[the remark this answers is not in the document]
evsaev: Hah, that's interesting!

`contentView` uses autoresizing-mask to have the same size with the cell. What I tried to do here is to get rid of the mask approach (by `contentView.translatesAutoresizingMaskIntoConstraints = false`) and setup the constraints manually. But it makes no sense since by default `content view translates its autoresizing-mask into constraints` 😄

Sources_v3/StreamChatUI/ChatChannelList/ChatChannelListCollectionViewCell.swift
```diff
@@ -5,4 +5,62 @@
...
+    public weak var collectionView: UICollectionView?
+    
+    public var channelView: ChatChannelView<ExtraData>? {
+        didSet {
+            oldValue?.removeFromSuperview()
+            if let new = channelView {
+                contentView.embed(new)
+            }
+        }
+    }
```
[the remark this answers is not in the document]
evsaev: responded in https://github.com/GetStream/stream-chat-swift/pull/588#discussion_r524060689

Sources_v3/StreamChatUI/ChatChannelList/ChatChannelListCollectionView.swift
```diff
@@ -6,7 +6,7 @@ import Foundation
 import UIKit
 
 open class ChatChannelListCollectionView: UICollectionView {
-    public required init(layout: ChatChannelListCollectionViewLayout) {
+    public required init(layout: UICollectionViewLayout) {
```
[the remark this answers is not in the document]
evsaev: Hmm, didn't actually get what you mean. We were working with `ChatChannelListCollectionViewLayout` but now we accept any `UICollectionViewLayout` 🙂

Sources_v3/StreamChatUI/Utils/UIView+Extensions.swift
evsaev: The docs says: 
> receiver [= | ≥ | ≤] 'anchor' + 'multiplier' * system space

When we work with `trailing/bottom` anchors we specify the negative offset to make it behave as we expect. I realized we can provide `multiplier == -1` and have the correct inset but seems we can't.

Sources_v3/StreamChatUI/ChatChannelList/ChatChannelListCollectionViewCell.swift
evsaev: updated to set `estimatedItemSize` in layout is being prepared and calculate cell size based on the estimated size

Sources_v3/StreamChatUI/ChatChannelList/ChatChannelListCollectionViewCell.swift
evsaev: fixed

Sources_v3/StreamChatUI/ChatChannelList/ChatChannelView.swift
evsaev: updated

Sources_v3/StreamChatUI/Utils/UIView+Extensions.swift
evsaev: Removed this helper. We can use the `preservesSuperviewLayoutMargins` and subview the superviews anchors

Sources_v3/StreamChatUI/ChatChannelList/ChatChannelListCollectionViewCell.swift
```diff
@@ -5,4 +5,62 @@
...
+    
+    // MARK: - Layout
+    
+    override open func preferredLayoutAttributesFitting(
+        _ layoutAttributes: UICollectionViewLayoutAttributes
+    ) -> UICollectionViewLayoutAttributes {
+        let attributes = super.preferredLayoutAttributesFitting(layoutAttributes)
+        
+        if let collectionViewSize = collectionView?.bounds.size {
```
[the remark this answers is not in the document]
evsaev: fixed

Sources_v3/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
evsaev: not needed, really. Removed 👍

Sources_v3/StreamChatUI/ChatChannelList/ChatChannelListCollectionViewCell.swift
```diff
@@ -5,4 +5,62 @@
...
+            let targetSize = CGSize(
+                width: collectionViewSize.width,
+                height: UIView.layoutFittingCompressedSize.height
+            )
```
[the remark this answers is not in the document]
evsaev: The **estimatedItemSize** is propagated to the cell so the `attributes.frame.size.width/height == UICollectionViewFlowLayout.automaticSize.width/height` check can be done.

However if in `prepare` we set just 1 dimension to be dynamic:
```swift
estimatedItemSize = .init(
  width: collectionView?.bounds.width ?? 0,
  height: UICollectionViewFlowLayout.automaticSize.height
) 
```
the cell’s `preferredLayoutAttributesFitting` is not called and the collection view looks empty.

Sources_v3/StreamChatUI/ChatChannelList/ChatChannelListCollectionView.swift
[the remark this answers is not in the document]
VojtaStavik: Aaaah, sorry, I misread the change. I though it was `UICollectionViewLayout` -> `ChatChannelListCollectionViewLayout`. The change you made was what I was suggesting. All good :)

Sources_v3/StreamChatUI/ChatChannelList/ChatChannelListCollectionViewCell.swift
```diff
@@ -5,4 +5,36 @@
 import Foundation
 import UIKit
 
-open class ChatChannelListCollectionViewCell: UICollectionViewCell {}
+open class ChatChannelListCollectionViewCell<ExtraData: UIExtraDataTypes>: UICollectionViewCell {
+    
+    // MARK: - Properties
+    
+    var uiConfig: UIConfig<ExtraData> = .default
```
VojtaStavik, line 12: This is not correct. The config must be passed from the super view.

Sources_v3/StreamChatUI/ChatChannelList/ChatChannelListCollectionViewCell.swift
```diff
@@ -5,4 +5,36 @@
...
+    
+    public private(set) lazy var channelView: ChatChannelView<ExtraData> = {
+        let view = uiConfig.channelList.channelView.init()
```
VojtaStavik, line 15: we should pass the config down to the channel view

Sources_v3/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
```diff
@@ -44,16 +44,13 @@ open class ChatChannelListVC<ExtraData: UIExtraDataTypes>: UIViewController,
...
-        } else {
-            let channelView = uiConfig.channelList.channelView.init(channel: channel, uiConfig: uiConfig)
-            cell.contentView.embed(channelView)
-        }
+        let cell = collectionView.dequeueReusableCell(
+            withReuseIdentifier: "Cell",
+            for: indexPath
+        ) as! ChatChannelListCollectionViewCell<ExtraData>
+    
+        cell.uiConfig = uiConfig
```
VojtaStavik, line 52: This needs to be done before. Should we do it in  `dequeueReusableCell` in our custom collection view subclass?

VojtaStavik (approved):
Look good, thank you 👍 Just a couple of minor comments.

Sources_v3/StreamChatUI/ChatChannelList/ChatChannelListCollectionViewCell.swift
evsaev: The collection view instantiates the registered cells by itself and returns non-optinal cell from `dequeueReusableCell`. I didn't find the way how to invoke our custom `init(config: UIConfig)` to create a cell instance. 

We have task in which scope the further investigation makes sense: [CIS-423](https://stream-io.atlassian.net/secure/RapidBoard.jspa?rapidView=61&projectKey=CIS&modal=detail&selectedIssue=CIS-423)

Sources_v3/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
evsaev: Yes, we can wrap `cell.uiConfig = uiConfig` by overriding `dequeueReusableCell` in our custom collection-view but the issue with calling `init(config: UIConfig)` will still be opened.

Suggest to work on it in a dedicated task. 

It might happen we will occasionally come up with smth similar to SwiftUI's environment variables: we will be iterating up by view hierarchy checking if a superview provides a `config: UIConfig<ExtraData>` and use `.default` if it wasn't found.

VojtaStavik: OK 👍

Sources_v3/StreamChatUI/ChatChannelList/ChatChannelListCollectionViewCell.swift
evsaev: fixed in 105094bf5e1ad5b8196f48f8e76edb87adc31b65

PR #593, opened 2020-11-17
[CIS-424] Channel list navigation bar

evsaev:
**This PR**:
- introduces `ChatNavigationBar` and makes of it in **SampleApp**
- introduces `CreateNewChannelButton` and makes of it in **SampleApp** on channel-list
- introduces `CurrentChatUserAvatarView` and makes of it in **SampleApp** on channel-list
- introduces `ChatRouter/ChatChannelListRouter ` and makes of it in **SampleApp** on channel-list
- updates `ChatChannelListVC` to implement `AppearanceSetting`
- adds asset catalog to `StreamChatUI`

Simulator Screen Shot - iPhone 11 - 2020-11-18 at 18 54 29


Sources_v3/StreamChatUI/Common Views/AvatarView.swift
```diff
@@ -14,6 +13,10 @@ open class AvatarView: UIView {
         return imageView
     }()
     
+    override open var intrinsicContentSize: CGSize {
+        .init(width: 40, height: 40)
+    }
+
```
VojtaStavik, line 19: This should be configurable using the default appearance mechanism. Where exactly does this size come from?

Sources_v3/StreamChatUI/Common Views/CreateNewChannelButton.swift
```diff
@@ -0,0 +1,32 @@
...
+
+import StreamChat
+import UIKit
+
+open class CreateNewChannelButton: UIButton, AppearanceSetting {
+    // MARK: - Overrides
+    
+    override open var intrinsicContentSize: CGSize {
+        .init(width: 44, height: 44)
+    }
```
VojtaStavik, line 13: Same here.... btw - why is this 44x44 and the avatar view is 40x40 ? :)

Sources_v3/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
```diff
@@ -13,6 +13,23 @@ open class ChatChannelListVC<ExtraData: UIExtraDataTypes>: UIViewController,
     public var controller: _ChatChannelListController<ExtraData>!
     public var uiConfig: UIConfig<ExtraData> = .default
     public var didSelectChannel: (_ChatChannel<ExtraData>) -> Void = { _ in }
+    public var didSelectCurrentUser: () -> Void = {}
+    public var didSelectCreateNewChannel: () -> Void = {}
+
+    private lazy var userAvatarView: CurrentChatUserAvatarView<ExtraData> = {
+        let avatar = uiConfig.channelList.currentUserAvatarView.init()
+        avatar.translatesAutoresizingMaskIntoConstraints = false
+        avatar.addGestureRecognizer(UITapGestureRecognizer(target: self, action: #selector(didTapOnCurrentUserAvatar)))
```
VojtaStavik, line 22: This is not ideal because we lose the pressed state. Maybe we should rather wrap this whole thing in a button?

Sources_v3/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
```diff
@@ -69,6 +90,16 @@ open class ChatChannelListVC<ExtraData: UIExtraDataTypes>: UIViewController,
...
+    
+    // MARK: Actions
+    
+    @objc private func didTapOnCurrentUserAvatar(_ sender: UITapGestureRecognizer) {
+        didSelectCurrentUser()
+    }
+    
+    @objc private func didTapCreateNewChannel(_ sender: UIBarButtonItem) {
+        didSelectCreateNewChannel()
+    }
```
VojtaStavik, line 102: Instead of `private` these methods should rather be `public/open` :)

Sources_v3/StreamChatUI/Common Views/CreateNewChannelButton.swift
```diff
@@ -0,0 +1,32 @@
...
+    
+    public required init?(coder: NSCoder) {
+        super.init(coder: coder)
+        applyDefaultAppearance()
+    }
+    
+    // MARK: - AppearanceSetting
+
+    public static func initialAppearanceSetup(_ button: CreateNewChannelButton) {
+        button.setImage(UIImage(named: "icn_new_chat", in: Bundle(for: Self.self), compatibleWith: nil), for: .normal)
```
VojtaStavik, line 30: Two things here:
- ideally, we'd have these images defined as constants like `UIImage.stream.iconNewChat`
- are we sure loading the images like this is safe when the SDK is imported via SPM?

Sources_v3/StreamChatUI/Common Views/CurrentChatUserAvatarView.swift
```diff
@@ -0,0 +1,40 @@
...
+            imageView.setImage(from: imageURL)
+        } else {
+            imageView.image = nil
+        }
+    }
+    
+    // MARK: - Overrides
+    
+    override open var intrinsicContentSize: CGSize {
+        .init(width: 28, height: 28)
```
VojtaStavik, line 26: Same comments here - this should be configurable + why 28x28? :)

Sources_v3/StreamChatUI/Common Views/AvatarView.swift
evsaev: 1. That's the size of channel avatar taken from the design of a channel list 
2. The `intrinsicContentSize` has only **{ get }** accessor so we cannot set it like we do for colors, etc. 

Possible workaround is:
```swift
open var defaultIntrinsicContentSize = CGSize(wight: UIView.noIntrincisMetrics, height: UIView.noIntrincisMetrics)

override open var intrinsicContentSize: CGSize {
  defaultIntrinsicContentSize
}

public static func initialAppearanceSetup(_ view: AvatarView) {
  view.defaultIntrinsicContentSize = .init(wight: 40, height: 40)
}
```

VojtaStavik: @evsaev Yes, that's exactly what I meant 👍

VojtaStavik: Ideally, `defaultIntrinsicContentSize` should be `CGSize?` and when it's not set `super.intrinsicContentSize` is used.

VojtaStavik:
Regarding the open questions:

> Do we need to listen to current-user updates in channel-list to keep the avatar view up-to-date?

That a very good question! It almost feels like there should be some kind of `CurrentUserAvatarViewController` that would take care of this and keep the avatar view up to date. You'd be able to use it as a component everywhere.

> When the new channel button is clicked, should we create the controller for a new channel directly in ChannelListVC? 

Feels like there should be some kind of coordinator object here responsible for the VC hierarchy, right?

> The navigation bar is shared across all the controllers in the navigation stack, not only channel-list. Does it makes sense to rename ChatChannelListNavigationBar -> ChatNavigationBar? 

Sounds good to me 👍 

> Are there any other things/appearance/behavior we want to customize via custom ChatNavigationBar?

We'll see :) I think your approach with letting `UINavigationItem` decide everything is correct, so it's possible `ChatNavigationBar` will be a very simple and "dumb" subclass.

Sources_v3/StreamChatUI/Common Views/CreateNewChannelButton.swift
evsaev: 1. The avatar view is `40x40` in the designs
2. The `44` is the navigation bar height. As the only place the `CreateNewChannelButton` is shown is the navigation bar I did it the same size.

Sources_v3/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
evsaev: hm 🤔 You suggest to keep `CurrentUserAvatar: AvatarView` and wrap it into the button OR have `CurrentUserAvatar: UIButton`?

VojtaStavik: Wrapping it

Sources_v3/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
evsaev: Hm.. We use closures to handle the actions -> end-users can provide any handlers. Do you think we want those methods to be `public/open`?

VojtaStavik: I'd say so. Let's try to use the public-by-default approach. Unless there's a serious reason why not to make it private, let's make it public.

Sources_v3/StreamChatUI/Common Views/CreateNewChannelButton.swift
evsaev: The way resources are exposed from `StreamChatUI` feels like big separate issue and requires some time. Using some tool that generates the way images will be accessed (`UIImage.stream.iconNewChat`) also feels more reasonable than handling it manually.

Couple of opened PRs require `.xcassets` in `StreamChatUI` so this can also be part of the `resource backbone` PR.

b-onc: Agreed with @evsaev , we need a resources backbone PR. Using resource bundles (not asset catalogs) is supported by all dependency managers

Sources_v3/StreamChatUI/Common Views/CurrentChatUserAvatarView.swift
evsaev: The first idea was to go with the **40x40** size defined by `AvatarView` but it in the UI it looks wild: the current user avatar doesn't fit the bar and has different left inset with channel avatars. 

That's was mainly the motivation to make the `CurrentChatUserAvatarView` smaller 🙂 

Simulator Screen Shot - iPhone 11 - 2020-11-18 at 13 32 25

VojtaStavik: Oh yeah, the designs have Android-height of 56pt, so we have to scale it down proportionally.

nit: Are you sure the bar margins are correct? I'd expect they should be aligned with the margins of the cells 🤔

Sources_v3/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
```diff
@@ -13,6 +13,23 @@ open class ChatChannelListVC<ExtraData: UIExtraDataTypes>: UIViewController,
     public var controller: _ChatChannelListController<ExtraData>!
     public var uiConfig: UIConfig<ExtraData> = .default
     public var didSelectChannel: (_ChatChannel<ExtraData>) -> Void = { _ in }
+    public var didSelectCurrentUser: () -> Void = {}
+    public var didSelectCreateNewChannel: () -> Void = {}
```
b-onc, line 17: I hope we'll move to delegates as the project matures, to be both inline with LLC and UIKit

Sources_v3/StreamChatUI/Common Views/CreateNewChannelButton.swift
evsaev: @VojtaStavik @b-onc added as a subtask to [Backbone for UI SDK](https://stream-io.atlassian.net/secure/RapidBoard.jspa?rapidView=61&projectKey=CIS&modal=detail&selectedIssue=CIS-409)

Sources_v3/StreamChatUI/Common Views/CurrentChatUserAvatarView.swift
```diff
@@ -0,0 +1,40 @@
...
+    }
+    
+    override open func setupAppearance() {
+        applyDefaultAppearance()
+        
+        super.setupAppearance()
```
b-onc, line 32: Isn't this called on `super.init()`?

Sources_v3/StreamChatUI/UIConfig.swift
```diff
@@ -46,20 +46,26 @@ public extension UIConfig {
...
+
         public init(
             channelCollectionView: ChatChannelListCollectionView.Type = ChatChannelListCollectionView.self,
             channelCollectionLayout: UICollectionViewLayout.Type = ChatChannelListCollectionViewLayout.self,
             channelView: ChatChannelView<ExtraData>.Type = ChatChannelView<ExtraData>.self,
             channelViewCell: ChatChannelListCollectionViewCell<ExtraData>.Type = ChatChannelListCollectionViewCell<ExtraData>.self,
-            avatarView: AvatarView.Type = AvatarView.self
+            channelAvatarView: AvatarView.Type = AvatarView.self,
+            currentUserAvatarView: CurrentChatUserAvatarView<ExtraData>.Type = CurrentChatUserAvatarView<ExtraData>.self,
+            newChannelButton: CreateNewChannelButton.Type = CreateNewChannelButton.self
```
b-onc, line 60: The `init` continues to grow....

Sources_v3/StreamChatUI/Common Views/CurrentChatUserAvatarView.swift
VojtaStavik: @evsaev See how the navbar margins match the cell margins in LLC sample app? 

Screen Shot 2020-11-18 at 13 21 25

Sources_v3/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
```diff
@@ -12,7 +12,22 @@ open class ChatChannelListVC<ExtraData: UIExtraDataTypes>: UIViewController,
     
     public var controller: _ChatChannelListController<ExtraData>!
     public var uiConfig: UIConfig<ExtraData> = .default
-    public var didSelectChannel: (_ChatChannel<ExtraData>) -> Void = { _ in }
+    public lazy var router: ChatChannelListRouter<ExtraData> = .init(rootViewController: self)
```
evsaev, line 15: Is it a good idea to have a `NavigationConfig` with router types?

Sources_v3/StreamChatUI/Utils/ChatRouter.swift
```diff
@@ -0,0 +1,17 @@
+//
+// Copyright © 2020 Stream.io Inc. All rights reserved.
+//
+
+import UIKit
+
+open class ChatRouter<Controller: UIViewController> {
```
evsaev, line 7: Decided to go with `ChatRouter` instead of `ChatCoordinator` because the latter usually has 2 responsibilities:
1. builds and shows the screen
2. handles the navigation from the screen

IMO better to follow the **SRP**

Another point is that `Coordinator` approach usually introduces parent-child relation between coordinators but currently it just complicates things. It's OK that the whole module graph is kept alive by the view-controller that is shown to the user (or lays somewhere in the navigation stack).

Sources_v3/StreamChatUI/Common Views/AvatarView.swift
evsaev: fixed

Sources_v3/StreamChatUI/Common Views/CreateNewChannelButton.swift
evsaev: fixed

Sources_v3/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
evsaev: resolved by having a `UIControl` subclass

Sources_v3/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
evsaev: made it `open`

Sources_v3/StreamChatUI/Common Views/CurrentChatUserAvatarView.swift
evsaev: hm. Checking what's is different in the sample app

Sources_v3/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
evsaev: The thing is that are not plain callbacks but the navigation. I tried to incapsulate it in a single object of `ChatChannelListRouter` type. Will be happy to listen to your thoughts on it 🙂

Sources_v3/StreamChatUI/Common Views/CurrentChatUserAvatarView.swift
evsaev: That's the first limitation we face with `AppearanceSetting`: if we subclass a view which implements `AppearanceSetting`, the base view method is invoked instead of a one implemented in subclass

Sources_v3/StreamChatUI/ChatChannelList/ChatChannelListRouter.swift
```diff
@@ -0,0 +1,20 @@
+//
+// Copyright © 2020 Stream.io Inc. All rights reserved.
+//
+
+import StreamChat
+import UIKit
+
+open class ChatChannelListRouter<ExtraData: UIExtraDataTypes>: ChatRouter<ChatChannelListVC<ExtraData>> {
+    open func openCurrentUserProfile(for currentUser: _CurrentChatUser<ExtraData.User>) {
+        debugPrint(currentUser)
```
VojtaStavik, line 10: first time I see `debugPrint` to be honest

Sample_v3/LoginViewController.swift
```diff
@@ -136,15 +136,11 @@ extension LoginViewController {
                 $0.channelNameLabel.backgroundColor = .yellow
             }
             
-            let navigation = UINavigationController(rootViewController: channelList)
+            // An example of using the custom router
+            channelList.router = MyChatChannelListRouter(rootViewController: channelList)
```
VojtaStavik, line 140: I think this should be configured via `UIConfig`

Sources_v3/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
VojtaStavik: Just 1 config is enough :) It should contain all classes we need but you can create a `Navigation` struct within it.

Sources_v3/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
```diff
@@ -28,6 +43,10 @@ open class ChatChannelListVC<ExtraData: UIExtraDataTypes>: UIViewController,
     override open func viewDidLoad() {
         super.viewDidLoad()
         
+        title = "Stream Chat"
+        navigationItem.leftBarButtonItem = UIBarButtonItem(customView: userAvatarView)
+        navigationItem.rightBarButtonItem = UIBarButtonItem(customView: createNewChannelButton)
+
```
VojtaStavik, line 49: Should we extend `ChatChannelListVC` with `AppearanceSetting` and do this there?

Sources_v3/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
VojtaStavik: I don't think the router should be owned by the VC but I guess it's OK for now.

Sources_v3/StreamChatUI/Common Views/CurrentChatUserAvatarView.swift
```diff
@@ -0,0 +1,112 @@
...
+    
+    override open var isSelected: Bool {
+        get { super.isSelected }
+        set { super.isSelected = newValue; updateContent() }
+    }
+    
+    override open var isEnabled: Bool {
+        get { super.isEnabled }
+        set { super.isEnabled = newValue; updateContent() }
+    }
```
VojtaStavik, line 48: Can't we just use `addTarget(self, action: #selector(updateContent), for: .allEvents)` and update the content there?

Sources_v3/StreamChatUI/UIConfig.swift
```diff
@@ -48,18 +51,39 @@ public extension UIConfig {
...
+        public var newChannelButton: CreateNewChannelButton.Type
+
         public init(
             channelCollectionView: ChatChannelListCollectionView.Type = ChatChannelListCollectionView.self,
             channelCollectionLayout: UICollectionViewLayout.Type = ChatChannelListCollectionViewLayout.self,
             channelView: ChatChannelView<ExtraData>.Type = ChatChannelView<ExtraData>.self,
             channelViewCell: ChatChannelListCollectionViewCell<ExtraData>.Type = ChatChannelListCollectionViewCell<ExtraData>.self,
-            avatarView: AvatarView.Type = AvatarView.self
+            avatarView: AvatarView.Type = AvatarView.self,
+            newChannelButton: CreateNewChannelButton.Type = CreateNewChannelButton.self
```
VojtaStavik, line 62: Yeah, I think it's time to forget about this idea of having everything in the init tbh

Sources_v3/StreamChatUI/Utils/ChatRouter.swift
VojtaStavik: Understood. We will have to think this through properly, ideally in a separate task. Can you create it, please?

VojtaStavik (commented):
Good! A lot of new things and patterns here, let's see what it will be like to use them. I think this is good to be finalized and merged.

Sources_v3/StreamChatUI/Utils/ChatRouter.swift
evsaev: Here we are https://stream-io.atlassian.net/browse/CIS-429

Sample_v3/LoginViewController.swift
evsaev: router is taken from config in ecd8af0b4ca1b2819df422cb35162cb072eca5ea

Sources_v3/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
evsaev: done in d4cc33bf143609304b77a3d9b43e8c6ad2086893

Sources_v3/StreamChatUI/Common Views/CurrentChatUserAvatarView.swift
evsaev: fixed in dff94e4bd93179075c8351a22fcc7bba2fa70e72

Sources_v3/StreamChatUI/Common Views/CurrentChatUserAvatarView.swift
VojtaStavik: @evsaev Did you find the issue?

evsaev: The navigation bar layout margins are the same on all the screens, doesn't matter what's in navigation item so that's probably the cell issue. Decided to get back to this later as we are a bit out of time now.

Sample_v3/LoginViewController.swift
```diff
@@ -158,3 +150,13 @@ extension LoginViewController {
...
+
+final class MyChatChannelListRouter: ChatChannelListRouter<DefaultExtraData> {
+    override func openChat(for channel: _ChatChannel<DefaultExtraData>) {
+        let chatStoryboard = UIStoryboard(name: "SimpleChat", bundle: nil)
+        let chatScreen = chatStoryboard
+            .instantiateViewController(withIdentifier: "SimpleChatViewController") as! SimpleChatViewController
+        chatScreen.channelController = rootViewController?.controller.client.channelController(for: channel.cid)
+        navigationController?.pushViewController(chatScreen, animated: true)
+    }
+}
```
b-onc, line 162: Why don't we introduce delegates for VCs instead of Routers?

evsaev: I'm actually thinking about the router as a `navigation delegate`. 

There's no much difference between `router.openCurrentUserProfile` and `delegate?.controller(self, didSelectCurrentUser:)`

Sources_v3/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
```diff
@@ -23,11 +23,21 @@ open class ChatChannelListVC<ExtraData: UIExtraDataTypes>: UIViewController,
...
+        button.addTarget(self, action: #selector(didTapCreateNewChannel), for: .touchUpInside)
+        return button
+    }()
+    
     // MARK: - Life Cycle
     
     override open func viewDidLoad() {
         super.viewDidLoad()
         
+        title = "Stream Chat"
```
VojtaStavik: We should probably make this configurable using `DefaultAppearance` too 🤔

Sources_v3/StreamChatUI/Common Views/AvatarView.swift
```diff
@@ -47,7 +54,14 @@ open class AvatarView: UIView {
...
+        imageView.setContentCompressionResistancePriority(.defaultLow, for: .vertical)
         embed(imageView)
-        widthAnchor.constraint(equalTo: heightAnchor, multiplier: 1).isActive = true
+    }
+    
+    // MARK: - AppearanceSetting
+
+    public static func initialAppearanceSetup(_ view: AvatarView) {
+        view.defaultIntrinsicContentSize = .init(width: 40, height: 40)
     }
```
VojtaStavik, line 66: What do you think about starting a habit of putting these into a standalone extension?

Sources_v3/StreamChatUI/UIConfig.swift
```diff
@@ -83,3 +86,20 @@ public extension UIConfig {
         }
     }
 }
+
+// MARK: - CurrentUser
+
+public extension UIConfig {
+    struct CurrentUserUI {
+        public var currentUserView: CurrentChatUserAvatarView<ExtraData>.Type
+        public var currentUserAvatarView: AvatarView.Type
```
VojtaStavik, line 98: I'm confused by the naming here. What is what?

Sample_v3/LoginViewController.swift
VojtaStavik: This is an interesting question we don't need to answer now. Let's stick with what we have here.

Sources_v3/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
```diff
@@ -126,3 +124,13 @@ extension ChatChannelListVC: _ChatChannelListControllerDelegate {
         }, completion: nil)
     }
 }
+
+// MARK: - AppearanceSetting
+
+extension ChatChannelListVC: AppearanceSetting {
+    public static func initialAppearanceSetup(_ controller: ChatChannelListVC<ExtraData>) {
+        controller.title = "Stream Chat"
```
VojtaStavik, line 132: Oh, you did it here :)

VojtaStavik (approved):
It still has many rough edges but that's expected. Many interesting new concepts, let's see what it will be like to use them. Thank you!

Sources_v3/StreamChatUI/Common Views/AvatarView.swift
evsaev: put to extension in c74de15cc35e6f3778c97729e52b844870e4813b, af50f91a41cb2cb784f4d76a0f5802a520564c56, and ffc43d6fd9e14ca0d6984b00311c2ce5e4c851f3

Sources_v3/StreamChatUI/UIConfig.swift
evsaev: fixed in 979c9e0bc42e75728c9641a26f7c971d766da7bf

Sources_v3/StreamChatUI/Common Views/CurrentChatUserAvatarView.swift
evsaev: Rolled back in af50f91a41cb2cb784f4d76a0f5802a520564c56: registering the selector (doesn't matter for **.allEvent** or **.touchEvents**) keeps view highlighted after `.touchUpInside` event

PR #594, opened 2020-11-17
[CIS-392] Safe sorting option for queries

mantergo:
**In this PR:**

- Add `cid` sorting key to `ChannelListSortingKey`
- Add safe sorting option to `ChannelListQuery`
- Add safe sorting option to `UserListQuery`

_`ChannelMemberListQuery` is left out cause there is no safe sorting option available._

### DIFF
--- a/Sources_v3/StreamChat/Query/Sorting/ChannelListSortingKey.swift
+++ b/Sources_v3/StreamChat/Query/Sorting/ChannelListSortingKey.swift
@@ -47,3 +47,15 @@ extension ChannelListSortingKey {
         .init(key: rawValue, ascending: isAscending)
     }
 }
+
+/// In case elements have same or `nil` target sorting values it's not possible to guarantee that order of elements will be the same
+/// all the time. So we need to additionally provide safe sorting option.
+extension Array where Element == Sorting<ChannelListSortingKey> {
+    func appendingCidSortingKey() -> [Sorting<ChannelListSortingKey>] {
+        guard !contains(where: { $0.key == .cid }), !isEmpty else {
+            return self
+        }
+        
+        return self + [.init(key: .cid)]
+    }
+}
--- a/Sources_v3/StreamChat/Query/Sorting/ChannelListSortingKey.swift
+++ b/Sources_v3/StreamChat/Query/Sorting/ChannelListSortingKey.swift
@@ -16,6 +16,9 @@ public enum ChannelListSortingKey: String, SortingKey {
     case lastMessageAt
     /// Sort channels by number of members.
     case memberCount
+    /// Sort channels by `cid`.
+    /// **Note**: This sorting option can extend your response waiting time if used as primary one.
+    case cid
     
     public func encode(to encoder: Encoder) throws {
         var container = encoder.singleValueContainer()
@@ -27,6 +30,7 @@ public enum ChannelListSortingKey: String, SortingKey {
         case .updatedAt: value = "updated_at"
         case .lastMessageAt: value = "last_message_at"
         case .memberCount: value = "member_count"
+        case .cid: value = "cid"
         }
         
         try container.encode(value)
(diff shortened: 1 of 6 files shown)


Sources_v3/StreamChat/Query/Sorting/ChannelListSortingKey.swift
```diff
@@ -16,6 +16,9 @@ public enum ChannelListSortingKey: String, SortingKey {
     case lastMessageAt
     /// Sort channels by number of members.
     case memberCount
+    /// Sort channels by `cid`.
+    /// **Note**: Avoid using this one cause it can be expensive from backend side.
```
VojtaStavik, line 20: This sound quite harsh :) If we don't want users to use it (there's probably no reason why they even should), let's not make it public.

Sources_v3/StreamChat/Query/Sorting/ChannelListSortingKey.swift
```diff
@@ -47,3 +47,15 @@ extension ChannelListSortingKey {
         .init(key: rawValue, ascending: isAscending)
     }
 }
+
+/// In case elements have same or `nil` target sorting values it's not possible to guarantee that order of elements will be the same
+/// all the time. So we need to additionally provide safe sorting option.
+extension Array where Element == Sorting<ChannelListSortingKey> {
+    func appendingSafeSortingIfNeeded() -> [Sorting<ChannelListSortingKey>] {
```
VojtaStavik, line 54: Why not call it with a more descriptive name, for example `appendingCidSortingKey`? There's nothing like "safe sorting" :) Basically, what we're doing here is to make sure the sorting is always stable by appending a sort descriptor that's guaranteed to be unique.

VojtaStavik (approved):
Looks good, thank you 👍 Just some comments regarding naming.

Sources_v3/StreamChat/Query/Sorting/ChannelListSortingKey.swift
```diff
@@ -43,3 +47,15 @@ extension ChannelListSortingKey {
...
+        guard !contains(where: { $0.key == .cid }), !isEmpty else {
+            return self
+        }
```
evsaev, line 57: Do we really need `isEmpty == false` check? Shouldn't the `cid` sorting key be added if the array is empty?

Sources_v3/StreamChat/Query/Sorting/UserListSortingKey.swift
```diff
@@ -40,3 +40,15 @@ extension UserListSortingKey {
...
     }
 }
+
+/// In case elements have same or `nil` target sorting values it's not possible to guarantee that order of elements will be the same
+/// all the time. So we need to additionally provide safe sorting option.
+extension Array where Element == Sorting<UserListSortingKey> {
+    func appendingSafeSortingIfNeeded() -> [Sorting<UserListSortingKey>] {
+        guard !contains(where: { $0.key == .id }), !isEmpty else {
+            return self
+        }
```
evsaev, line 50: The same as https://github.com/GetStream/stream-chat-swift/pull/594/files#r525915508

Sources_v3/StreamChat/Query/Sorting/ChannelListSortingKey.swift
b-onc: AFAIK backend will sort them "safely" if client sends an empty sorting

mantergo: Yep, @b-onc is right.
Backend will apply default sorting with secondary by `cid` in this case.

evsaev (approved):
👍

Sources_v3/StreamChat/Query/Sorting/ChannelListSortingKey.swift
mantergo: After discussions decided to go with updated comment.
See 34bee8a.

PR #595, opened 2020-11-17
[CIS-355] Fix bootstrap script

ManWithBear:
Since September, homebrew does not allow installs from GitHub.

Instead, you required to create your own tap. More info:
https://discourse.brew.sh/t/has-brew-install-force-formula-raw-path-been-taken-out/8793

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -6,9 +6,10 @@
 
 # Install swiftformat 0.45.6
 brew unlink swiftformat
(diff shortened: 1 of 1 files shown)


mantergo (dismissed):

VojtaStavik (approved):
I'm not exactly sure what's happening here but it seems to be working 🤷‍♂️ :) Thank you and congratulations on your first PR for Stream 🎉

PR #596, opened 2020-11-19
[CIS-423] Container cell

evsaev:
### Problem description

We would like to instantiate `ChatChannelView<ExtraData>` and add to cell view hierarchy when the cell is created. To instantiate `ChatChannelView<ExtraData>` of the correct type, the type should be taken from `UIConfig<ExtraData>` (as it can be overridden). We cannot create custom `init(uiConfig: UIConfig<ExtraData>)` for the cell and invoke it since the cells are created inside the `UICollectionView` via `init(frame: CGRect)/init(coder: )`.

### Possible solutions:
1. have a generic cell + stored `var uiConfig<ExtraData> = .default`
2. have a generic cell + computed `uiConfig<ExtraData>` and iterate up by view hierarchy searching for ancestor that provides `UIConfig<ExtraData>`
3. have a non-generic cell with `content: UIView?` and let the controller do the rest

### Solutions issues:
1. when we have stored config we have to make sure `cell.uiConfig = controller.uiConfig` happens before the  `ChatChannelView<ExtraData>` is created. Otherwise we risk to create another `ChatChannelView<ExtraData>` from default config other than from controller's config
2. I pin hopes on this approach since it is similar to environment variables in `SwiftUI`, can be applied for all the views and helps to avoid storing `let config + init(config)`. The problem is that we create subviews from the inside of view's `init` and at this point the view itself hasn't yet had a superview -> there is no view hierarchy to search the `uiConfig<ExtraData>` for.
3. The type of content view is not fixed to any specific one. On the other hand, it allows the cell to be reused in other collections (not only channel-list)

### DIFF
--- a/Sources_v3/StreamChatUI/Utils/UIView+UIConfig.swift
+++ b/Sources_v3/StreamChatUI/Utils/UIView+UIConfig.swift
@@ -1,0 +1,34 @@
+//
+// Copyright © 2020 Stream.io Inc. All rights reserved.
+//
+
+import UIKit
+
+// MARK: - Backing type-erased config
+
+private extension UIView {
+    static var anyUIConfigKey: UInt8 = 0
+    
+    var anyUIConfig: Any? {
+        get { objc_getAssociatedObject(self, &Self.anyUIConfigKey) as? String }
+        set { objc_setAssociatedObject(self, &Self.anyUIConfigKey, newValue, .OBJC_ASSOCIATION_RETAIN) }
+    }
+}
+
+// MARK: - Public config API
+
+public extension UIView {
+    func register<T: UIExtraDataTypes>(config: UIConfig<T>) {
+        anyUIConfig = config
+    }
+    
+    func uiConfig<T: UIExtraDataTypes>(_ type: T.Type = T.self) -> UIConfig<T> {
+        if let config = anyUIConfig as? UIConfig<T> {
+            return config
+        } else if let superview = superview {
+            return superview.uiConfig(type)
+        } else {
+            return .default
+        }
+    }
+}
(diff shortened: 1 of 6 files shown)


Sources_v3/StreamChatUI/Utils/UIView+UIConfig.swift
```diff
@@ -0,0 +1,34 @@
...
+        set { objc_setAssociatedObject(self, &Self.anyUIConfigKey, newValue, .OBJC_ASSOCIATION_RETAIN) }
+    }
+}
+
+// MARK: - Public config API
+
+public extension UIView {
+    func register<T: UIExtraDataTypes>(config: UIConfig<T>) {
+        anyUIConfig = config
+    }
```
evsaev, line 23: That's how we can override the `UIConfig`. The new config will be applied for all the views down the hierarchy.

Sources_v3/StreamChatUI/Utils/UIView+UIConfig.swift
```diff
@@ -0,0 +1,34 @@
...
+    
+    func uiConfig<T: UIExtraDataTypes>(_ type: T.Type = T.self) -> UIConfig<T> {
+        if let config = anyUIConfig as? UIConfig<T> {
+            return config
+        } else if let superview = superview {
+            return superview.uiConfig(type)
+        } else {
+            return .default
+        }
+    }
```
evsaev, line 33: We check if there is a config registered in the current view. 
If not - we go up the hierarchy looking for the config. 
If none found, the `.default` is taken

b-onc:
Work was moved to #602 
Closing

PR #601, opened 2020-11-23
Improve bootstrapping the environment

VojtaStavik:
### In this PR

- We use SwiftLint and SwiftFormat via Mint
- SwiftLint version used by Danger should be the same as the one used locally
- The bootstrap script is updated and it provides nicer feedback
- We run SwiftLint for the v3 Sample app, we don't run it for the v2 codebase anymore


bootstrap.sh
```diff
@@ -1,27 +1,45 @@
...
+    echo "Homebrew not installed. Please install."
+    exit 1
+fi
+
+echo
+echo "👉 Generate mintfile"
+echo "# Warning - this file is generated using bootstrap.sh, don't edit it directly." > mintfile
+echo "nicklockwood/SwiftFormat@$SWIFTFORMAT_VERSION" >> mintfile
+echo "realm/SwiftLint@$SWIFTLINT_VERSION" >> mintfile
+cat mintfile
```
b-onc, line 21: Why don't we just check-in the mintfile to repo?

VojtaStavik: I'm just trying something out, will probably end up having it checked out to the repo directly...

VojtaStavik: Unfortunately, Danger usesSWIFTLINT_VERSION env variable for configuration and we have to use it in Mint file, too. I wanted to have a single source of truth, so generating Mint file like this seems easier than parsing it and extracting the value from it.

VojtaStavik:
This PR introduces a lot of warnings in the Sample app. We forgot to run SwiftLint for it, too 🤦‍♂️

bootstrap.sh
```diff
@@ -1,27 +1,46 @@
...
+
+echo
+echo -e "👉 Install Mint if needed"
+# List installed Mint versions, if fails, install Mint
+brew list mint || brew install mint
```
DominikBucher12, line 26: IMHO This seems too complicated :D 

Let me explain. Right now, we are doing: 

1. Install one package manager (🍺).
1. This 🍺 doesn't really install anything unless step 3.
1. With 🍺 we do install now 🌱 .
1. with 🌱 we finally do our nice thing and install `Swiftlint` and `Swiftformat`

Personally I would get rid of the dependencies and install this stuff just in the scope of the repo, also I secretely hate 🍺 because it can ruin your environment if misused or misunderstood which happens quite often.

1. Swiftlint can be installed via `CocoaPods`. Or Just 🍺. Both are already used before this PR.
2. SwiftFormat can be stand-alone command line tool (without any use of any package manager...) Or it can be installed via brew, downsideof the stand-alone thing is we would need to update it everytime it gets updated... 

My idea is to either:

- Use 🍺 for both Swiftlint and Swiftformat
- Use `cocoaPods` for `Swiftlint` (nicely integrates into the project), Use `standalone binary` of `swiftformat`
- (Your idea)

bootstrap.sh
ManWithBear:
```sh
cat Mintfile | grep SwiftLint | awk -F@ '{print $2}'
```

bootstrap.sh
ManWithBear: `Homebrew` is bad, because:
1. You cannot lock version, it always install most recent one
2. You cannot install old one except via dirty tricks (look at previous installation of SwiftFormat)
3. Install in global environment

We don't want to use `CocoaPods`, because:
1. We are not using them right now
2. super heavy and invasive into your project

We don't want use global installations:
1. I have multiple projects with different versions of SwiftLint, with global installation no way for me to handle it
2. We need to be able sync tools versions through team (writing in slack is not a way)

bootstrap.sh
```diff
@@ -1,27 +1,46 @@
...
 set -euo pipefail
 
+echo
+echo -e "👉 Bootstrap Mint dependencies"
+mint bootstrap
+
 # Symlink hooks folder to .git/hooks folder
+echo
+echo -e "👉 Create symlink for pre-commit hooks"
 ln -sf ../../hooks/pre-commit.sh .git/hooks/pre-commit
```
ManWithBear, line 38: Am I missing something or why `../../hooks` and not `./hooks`?

bootstrap.sh
ManWithBear: IMHO Mintfile generation sounds fishy.
I would expect it to be clear declaration

ManWithBear (changes requested):

bootstrap.sh
VojtaStavik: OK, I'll get it from the mintfile. Fixed in 08a0e2d

bootstrap.sh
VojtaStavik: @ManWithBear is right. I know it's not ideal to have 3 tools to install our dependencies but Homebrew doesn't provide us all we need. We use RubyGems and Bundler to run Fastlane, Mint to install all Swift-based dependencies, and Homebrew to install Mint :) Since almost everyone has Homebrew and it's also preinstalled on the CI, I think it's a setup we can live with.

bootstrap.sh
VojtaStavik: I have no idea 🤷‍♂️  It seems everything is working correctly with `./hooks`. We even use it on the following line too. Fixed in 08a0e2d

ManWithBear (approved):
👍🏾

PR #577, opened 2020-10-29
[CIS-410] Backbone UI

evsaev:
**This PR**:
- introduces `UIConfig` concept for overriding UI types
- introduces UI components generic over extra-data for channel list 
- introduces `ContainerStackView` for flexible layout
- showcases `StreamChatUI` usage in sample app

PR #602, opened 2020-11-23
[CIS-416] Explore lazy setup for views

b-onc:
### Main issue
Views are doing layout/setup operations in `commonInit`, which is called on `init`. These operations can be expensive, and cause performance issues [we don't have numbers yet]. We're searching for a way to "lazily" do these operations.
Comment which started the discussion: https://github.com/GetStream/stream-chat-swift/pull/577#discussion_r522078499

### Secondary issue
As #596 explores, passing `config` from views to subviews & initiating correct views from config:
> I pin hopes on this approach since it is similar to environment variables in SwiftUI, can be applied for all the views and helps to avoid storing let config + init(config). **The problem is that we create subviews from the inside of view's init and at this point the view itself hasn't yet had a superview -> there is no view hierarchy to search the uiConfig<ExtraData> for.**

So we need a way to lazily do the `setupAppearance`/`setupLayout` operations, in which subviews will be created. No subviews will be touched (hence, created) during init, or before having a subview.

### Solutions explored
- `layoutSubviews`
This is part of render loop, which will be called 120 times per second. Apple generally suggest not overriding this, because low performance code in this function can cause rendering issues. If we were to use this, we'd have to do:
```swift
var didLayoutSubviews = false
override func layoutSubviews() {
  super.layoutSubviews()
  guard !didLayoutSubviews else { return }
  // our code here, to be run 1 time only
  setupLayout()
  setupAppearance()
  didLayoutSubviews = true
}
```
which is considered a code smell.
Another big issue with this approach, since `setupLayout` and `setupAppearance` is `open`, users will be able to override these, and any costly operations users implement will cause performance issues.

- `updateConstraints`
The same findings as above. `updateConstraints` is also part of render loop, being called before `layoutSubviews`.

- `didMoveToSuperview`
This is the closest callback we have in UIKit to what we actually need. It's being called whenever view's `superview` property changes.

So I came up with the `OverridableView` base class that ensures `setupLayout` and `setupAppearance` are called in the correct order when the view has a valid superview, and subclasses only need to override these 2 functions to setup themselves.
Also has the added benefit of calling `applyDefaultAppearance` in the correct place.

### Open points
1. Since `CurrentUserAvatarView` is `UIControl` rather than `UIView`, we need another base class for it, `OverridableControl`. How can we avoid this? 2 base classes basically the same, yet we need 2 different classes.
1. Should `OverridableView` also conform to `AppearanceSetting`? I thought this was a nice addition - in the worst case where the view doesn't need Appearance configurability, it doesn't need to do anything - default implementation is provided.
1. Should `updateContent` also be part of base class?

### DIFF
--- a/Sources_v3/StreamChatUI/Common Views/AvatarView.swift
+++ b/Sources_v3/StreamChatUI/Common Views/AvatarView.swift
@@ -4,7 +4,7 @@
 
 import UIKit
 
-open class AvatarView: UIView {
+open class AvatarView: View {
     // MARK: - Subviews
     
     public let imageView: UIImageView = {
@@ -20,24 +20,6 @@ open class AvatarView: UIView {
         defaultIntrinsicContentSize ?? imageView.intrinsicContentSize
     }
     
-    // MARK: - Init
-    
-    override public init(frame: CGRect) {
-        super.init(frame: frame)
-        commonInit()
-    }
-    
-    public required init?(coder: NSCoder) {
-        super.init(coder: coder)
-        commonInit()
-    }
-    
-    private func commonInit() {
-        applyDefaultAppearance()
-        setupLayout()
-        setupAppearance()
-    }
-    
     // MARK: - Layout
     
     override open func layoutSubviews() {
@@ -47,23 +29,19 @@ open class AvatarView: UIView {
     }
     
     // MARK: - Public
+
+    public func defaultAppearance() {
+        defaultIntrinsicContentSize = .init(width: 40, height: 40)
+    }
     
-    open func setupAppearance() {
+    override open func setUpAppearance() {
         imageView.contentMode = .scaleAspectFit
         imageView.clipsToBounds = true
     }
     
-    open func setupLayout() {
+    override open func setUpLayout() {
         imageView.setContentCompressionResistancePriority(.defaultLow, for: .horizontal)
         imageView.setContentCompressionResistancePriority(.defaultLow, for: .vertical)
         embed(imageView)
-    }
-}
-
-// MARK: - AppearanceSetting
-
-extension AvatarView: AppearanceSetting {
-    public class func initialAppearanceSetup(_ view: AvatarView) {
-        view.defaultIntrinsicContentSize = .init(width: 40, height: 40)
     }
 }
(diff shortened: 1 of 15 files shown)


Sources_v3/StreamChatUI/Common Views/AvatarView.swift
```diff
@@ -62,7 +48,7 @@ open class AvatarView: UIView {
 
 // MARK: - AppearanceSetting
 
-extension AvatarView: AppearanceSetting {
+extension AvatarView {
     public class func initialAppearanceSetup(_ view: AvatarView) {
         view.defaultIntrinsicContentSize = .init(width: 40, height: 40)
     }
```
evsaev, line 54: Unfortunately, that won't work. The `OverridableView.initialAppearanceSetup` will be invoked in all the cases instead the method in the subclass.

To have `CustomView.initialAppearanceSetup` working it should be `override class` + some magic for getting the subclass type in run-time (via **type(of: self)**) + optional cast `OverridableView -> CustomView` in all the subclasses (since the method has to have the same signature to be overridden)

b-onc: Good point, thanks for the heads up! Do you think handling of `AppearanceSetting` should be in the base class at all?

The solution requires:
```swift
open class AvatarView: OverridableView {
    override public class func initialAppearanceSetup(_ view: OverridableView) {
        guard let view = view as? AvatarView else {
            return
        }
        view.defaultIntrinsicContentSize = .init(width: 40, height: 40)
    }
...
```
for every subclass we'll have, which is very repetitive.

b-onc: Addendum: casting is only required to access custom properties. If a subclass only styles default properties (from `UIView`) it'll work

evsaev: I'd move `AppearanceSetting` to subclasses. That requires calling `applyDefaultAppearance` in overridden `setupAppearance` but no optional casting in there + if end-users will subclass our custom view, better to have our custom view type in `class func initialAppearanceSetup` other than `OverridableView/OverridableControl`

VojtaStavik (approved):
image

PR #605, opened 2020-11-25
Delete most of Swiftlint warnings

DominikBucher12:
# What does this PR do
Clears most Swiftlint warnings across the project. I ran `swiftlint autocorrect` to fix some issues which Swiftlint could fix on it own. I also went through files with documentation and splitted the documentation on newLines so we don't get any `line_length` swiftlint rule (It was faster than figure out some automation thing)
# How to test this
 You can run the project go through it, run tests and see if anything is broken, but it really shouldn't be
# Where to start
 Some place worth a logic check are: 

- `ConfigurationViewController.swift` -> `isEmpty` condition
- `WebSocketProviderMock.swift` -> Possibly if the delegate is explicitly not meant to be weak (I don't see any reason) 
- `SimpleUsersViewController.swift`  -> `super.viewDidLoad`, was this intentional?
-  `CombineSimpleUsersViewController.swift` -> `super.viewDidLoad`, was this intentional?

The rest is just documentation, you can suggest if you want the rest of warnings (4) but I believe they are not important for now. At least now you don't see tons of warnings about `line_length` :)

### DIFF
--- a/Sample_v3/ConfigurationViewController.swift
+++ b/Sample_v3/ConfigurationViewController.swift
@@ -40,7 +40,8 @@ class ConfigurationViewController: UITableViewController {
         case 2:
             tokenTypeSegmentedControl.selectedSegmentIndex = 2
             jwtTextField.isEnabled = false
-            if token != "" {
+            if let token = token,
+                !token.isEmpty {
                 tableView.deleteRows(at: [jwtCellIndexPath], with: .top)
             }
             jwtTextField.text = .development
--- a/Sample_v3/Samples/CombineSimpleChat/CombineSimpleUsersViewController.swift
+++ b/Sample_v3/Samples/CombineSimpleChat/CombineSimpleUsersViewController.swift
@@ -9,7 +9,8 @@ import UIKit
 ///
 /// # CombineSimpleUsersViewController
 ///
-/// A `UITableViewController` subclass that displays and manages the list of users.  It uses the `ChatUserListController`  class to make calls to the Stream Chat API
+/// A `UITableViewController` subclass that displays and manages the list of users.
+/// It uses the `ChatUserListController`  class to make calls to the Stream Chat API
 /// and listens to events by conforming to `ChatUserListControllerDelegate`.
 ///
 @available(iOS 13, *)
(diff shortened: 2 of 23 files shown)


b-onc (approved):
You can include `v3` or `ui v3` labels to avoid danger's Changelog error

Sample_v3/Samples/CombineSimpleChat/CombineSimpleUsersViewController.swift
```diff
@@ -159,7 +166,7 @@ class CombineSimpleUsersViewController: UITableViewController, ChatUserListContr
         
         var cell = tableView.dequeueReusableCell(withIdentifier: "Cell")
         
-        if (!(cell != nil)) {
+        if !(cell != nil) {
```
VojtaStavik, line 169: `if cell == nil` ?

Sample_v3/Samples/SimpleChat/SimpleUsersViewController.swift
```diff
@@ -130,7 +138,7 @@ class SimpleUsersViewController: UITableViewController, ChatUserListControllerDe
         
         var cell = tableView.dequeueReusableCell(withIdentifier: "Cell")
         
-        if (!(cell != nil)) {
+        if !(cell != nil) {
```
VojtaStavik, line 141: same here

Sample_v3/ConfigurationViewController.swift
```diff
@@ -40,7 +40,7 @@ class ConfigurationViewController: UITableViewController {
         case 2:
             tokenTypeSegmentedControl.selectedSegmentIndex = 2
             jwtTextField.isEnabled = false
-            if token != "" {
+            if !(token?.isEmpty ?? true) {
```
VojtaStavik, line 43: Is this equal to `token?.isEmpty == false` or not?  🤔

VojtaStavik (approved):
Looks good. I must say that having the comments not formatted to a block but like this hurts my eyes but I guess we can improve it later. There's also @ManWithBear's PR that touches that.

Screen Shot 2020-11-26 at 10 31 29

VojtaStavik:
Hey @DominikBucher12, any problems with this PR?

Sample_v3/Samples/SimpleChat/SimpleUsersViewController.swift
DominikBucher12: +1 swiftlint didn't catch this somehow.

Sample_v3/Samples/CombineSimpleChat/CombineSimpleUsersViewController.swift
DominikBucher12: +1 swiftlint didn't catch this.

Sample_v3/ConfigurationViewController.swift
DominikBucher12: 🤔 

1. Given we pass this condition only when token is not `""` which is the same as `!token.isEmpty` 
==>>  `(token != "") == !(token.isEmpty)`
1. When the token is nil, it means (for 99% of usecases when we check for string value and want to use it somewhere) automatically it's empty. So we replace immediately the value `isEmpty` with true, because it makes sense.
1. When we negate the `true`, we get obviously false, so it won't pass when either the **token is nil** or the **token is empty**.

This question has some valid point since I was while explaining this I got confused, you got me there for a second.

PR #606, opened 2020-11-26
Adds Demo app target

b-onc:
Demo -> app as shown in figma designs. We're using a target instead of a separate project to make development easier


VojtaStavik (approved):
Can you please also add SwiftLint run phase and adjust the formatter script to format files in `Demo`? Btw - should we maybe call it `Demo-iOS` or `DemoApp-iOS`. or something like that? Just `Demo` sounds strange and too vague.

PR #607, opened 2020-11-26
Fix "typo" in default implementation of lowlevel controller delegates

ManWithBear:
We dropped a guard and copypaste demons find a way in our codebase.
With power granted to me by supreme leader @VojtaStavik I banish them for good!


VojtaStavik (approved):
Very nice 👍

PR #608, opened 2020-11-26
Add notificationChannelDeleted event

b-onc:
This event was missing even from backend, but now it's here

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Upcoming
 
+### ✅ Added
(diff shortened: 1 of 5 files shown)


Sample/Sources/LoginViewController.swift
```diff
@@ -138,6 +138,11 @@ final class LoginViewController: ViewController {
         }
         
         func showNext(_ result: Result<UserConnection, ClientError>) {
+            let channel = Client.shared.channel(type: .messaging, id: "new_channel_7970")
+            channel.delete { res in
+                print("###", res)
```
VojtaStavik, line 143: ?

VojtaStavik (approved):

Sample/Sources/LoginViewController.swift
b-onc: Shouldn't have checked this in :D

PR #616, opened 2020-12-02
[CIS-496] Preview URL for attachments

evsaev:
**This PR** adds `imagePreviewURL` to attachment payload, DTO, and model types


mantergo (approved):

VojtaStavik (approved):
👍 Do I understand it correctly that the value already exists in the payload we just didn't expose it?

evsaev:
> 👍 Do I understand it correctly that the value already exists in the payload we just didn't expose it?

@VojtaStavik yep, exactly

PR #618, opened 2020-12-02
[CIS-435] Switch from autolayout to frame based layout for message list collection view

ManWithBear:
For fast and smooth scrolling experience we decided to go for frame based layout for message cells. To achieve it we have:
- `ChatChannelCollectionViewLayout` - collection view layout that precalculate frames for all items in collection view
- `ChatChannelCollectionViewLayoutModel` - DTO containing desire heights for all items in collection view
- `ChatChannelCollectionViewLayoutDelegate` - delegate responsible for calculation of `ChatChannelCollectionViewLayoutModel` for current data, right now it's `ChatChannelVC`
- `ConfiguredLayoutProvider` - abstract class for layout calculators, provide access for `UIConfig`
- `View.LayoutProvider ` - brain of layout. Calculates size for cell + frames for cell subviews
- `View.Layout` - DTO that contains frames for views subviews. If frame is `nil` subview should be hidden

### DIFF
--- a/Sources_v3/StreamChatUI/ChatChannel/ChatChannelCollectionViewLayout.swift
+++ b/Sources_v3/StreamChatUI/ChatChannel/ChatChannelCollectionViewLayout.swift
@@ -5,63 +5,75 @@
 import Foundation
 import UIKit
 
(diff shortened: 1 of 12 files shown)


ManWithBear:
Postponed to Q1

PR #619, opened 2020-12-03
Replace UIInputAccessoryView with UIInputAccessoryViewController

DominikBucher12:
# What this PR Does
- Replaces implementation of `ChatChannelMessageComposerView` which was implemented just as `accessoryInputView` to implementation which now has `UIInputViewController` to holds the `inputView` since this is for us better way to implement complex logic with the `ChatChannelMessageComposerView`. Also changes `ChatChannelMessageComposerView` to subclass `UIInputView` instead.
# How to test this
- Start the v3 app, go to some random chat and expect the composerView to be there and behave the same way it did in the first implementation.

Note: This PR is for us to move on and implement the logic into this separate viewController.

### DIFF
--- a/Sources_v3/StreamChatUI/ChatChannel/ChatChannelMessageComposerView.swift
+++ b/Sources_v3/StreamChatUI/ChatChannel/ChatChannelMessageComposerView.swift
@@ -4,7 +4,7 @@
 
 import UIKit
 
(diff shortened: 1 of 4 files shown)


mantergo (approved):
Thanks! 👌

PR #621, opened 2020-12-03
[CIS-436] Message actions UI

evsaev:
**This PR**:
- adds and shows context menu for the message with available actions
- shows different message actions depending on message state (incoming/outgoing)
- handles `copy`, `delete`, `mute/unmute author` actions
- asks confirmation for `delete` action

**What's missing**:
1. Handling of `inline/thread reply` actions. Before working on UI part, the LLC should be updated with `repliesTo` field on the message + reply API improvements (currently we have confusing `createNewReply` on **_ChatMessageController** since it's not clear inline or thread reply will be created)
2. Handling of `edit` action since the message that is being edited should be placed into composer with all the attachments etc.
3. Deleted message cell UI (couldn't find it in the Figma, will clarify this)

Simulator Screen Shot - iPhone 11 - 2020-12-03 at 19 35 52

Simulator Screen Shot - iPhone 11 - 2020-12-03 at 19 36 13

Simulator Screen Shot - iPhone 11 - 2020-12-04 at 10 39 58


Sources_v3/StreamChatUI/ChatChannel/ChatChannelVC.swift
```diff
@@ -241,6 +232,100 @@ open class ChatChannelVC<ExtraData: UIExtraDataTypes>: ViewController,
...
+                        }
+                    })
+                }
+            }
+
+            return actions
+        }
+
+        return .init(
+            view: uiConfig.messageList.messageActionsView.init().withoutAutoresizingMaskConstraints,
```
ManWithBear, line 299: Why are you injecting view from outside?

Sources_v3/StreamChatUI/ChatChannel/ChatChannelVC.swift
```diff
@@ -241,6 +232,100 @@ open class ChatChannelVC<ExtraData: UIExtraDataTypes>: ViewController,
...
+
+        let popup = ChatMessagePopupViewController<ExtraData>(
+            message: messageData,
+            messageFrameToFit: cell.messageView.superview!.convert(cell.messageView.frame, to: nil),
+            messageActionsController: actionsController(
+                for: messageData,
+                messageController: messageController
+            ),
+            messageReactionsController: .init(
+                view: uiConfig.messageList.messageReactionsView.init().withoutAutoresizingMaskConstraints,
```
ManWithBear, line 320: Why are you injecting view from outside?

Sources_v3/StreamChatUI/ChatChannel/ChatChannelVC.swift
evsaev: `ChatMessageActionsViewController` doesn't inherit from `UIViewController` and doesn't serve as a `UIConfigProvider` so we cannot instantiate view of correct type without passing the `UIConfig` there. 

Decided to keep things simple and inject the views.

Sources_v3/StreamChatUI/ChatChannel/ChatChannelVC.swift
evsaev: responded in https://github.com/GetStream/stream-chat-swift/pull/621#discussion_r535892333

Sources_v3/StreamChatUI/MessageActionsPopup/MessageActionsView.swift
```diff
@@ -0,0 +1,69 @@
...
+    // MARK: - Init
+
+    public init(
+        view: MessageActionsView<ExtraData>,
+        availableActions: [ChatMessageActionItem]
+    ) {
+        self.view = view
+        self.view.actionItems = availableActions
+    }
+}
```
VojtaStavik, line 69: I get the idea of having `ViewController` in the sense of a "controller" of a "view" but I'd strongly recommend not using it for types that are not subclasses of `UIViewController`. It will be extremely confusing for our customers.

VojtaStavik: Any reason why you don't want to make it a `UIViewController` subclass?

Sources_v3/StreamChatUI/ChatChannel/ChatMessageReactionsView.swift
```diff
@@ -99,7 +99,7 @@ class MessageReactionView: UIView {
 
 // MARK: - Reactions
 
-class ChatMessageReactionsView: UIView {
+open class ChatMessageReactionsView: UIView {
```
VojtaStavik, line 102: This should inherit from `View`, then you won't need `commonInit()` and you can use the lifecycle methods

Sources_v3/StreamChatUI/MessageActionsPopup/ChatMessagePopupViewController.swift
```diff
@@ -0,0 +1,113 @@
...
+        messageActionsController: ChatMessageActionsViewController<ExtraData>,
+        messageReactionsController: ChatMessageReactionViewController<ExtraData>
+    ) {
+        self.message = message
+        messageViewFrame = messageFrameToFit
+        actionsController = messageActionsController
+        reactionsController = messageReactionsController
+        
+        super.init(nibName: nil, bundle: nil)
+    }
```
VojtaStavik, line 37: Interesting that you chose this pattern for instantiating the view controller? Why not having the VC to create the subviews from UIConfig?

Sources_v3/StreamChatUI/ChatChannel/ChatChannelVC.swift
VojtaStavik: -> Slack

Sources_v3/StreamChatUI/ChatChannel/ChatChannelRouter.swift
```diff
@@ -0,0 +1,35 @@
...
+//
+
+import StreamChat
+import UIKit
+
+open class ChatChannelRouter<ExtraData: UIExtraDataTypes>: ChatRouter<ChatChannelVC<ExtraData>> {
+    open func showMessageActionsPopUp(
+        messageContentFrame: CGRect,
+        messageData: _ChatMessageGroupPart<ExtraData>,
+        messageController: _ChatMessageController<ExtraData>,
```
VojtaStavik, line 12: First I was thinking that the routers should have access to `client` but this solution is much better 👍

Sources_v3/StreamChatUI/UIConfig.swift
```diff
@@ -41,6 +41,7 @@ public extension UIConfig {
     struct Navigation {
         public var navigationBar: ChatNavigationBar.Type = ChatNavigationBar.self
         public var channelListRouter: ChatChannelListRouter<ExtraData>.Type = ChatChannelListRouter<ExtraData>.self
+        public var messageListRouter: ChatChannelRouter<ExtraData>.Type = ChatChannelRouter<ExtraData>.self
```
VojtaStavik, line 44: nit: we don't really use "message list" as a term for chat detail. Maybe `channelDetailRouter`?

VojtaStavik (commented):
This is a great piece of work 👍 I started a couple of slack threads about topics from this PR that are not clear yet. Let's finish the discussion there.

Sources_v3/StreamChatUI/MessageActionsPopup/MessageActionsView+ActionButton.swift
```diff
@@ -0,0 +1,43 @@
+//
+// Copyright © 2020 Stream.io Inc. All rights reserved.
+//
+
+import UIKit
+
+extension MessageActionsView {
+    open class ActionButton: Button {
+        public var actionItem: ChatMessageActionItem? {
+            didSet { updateContent() }
```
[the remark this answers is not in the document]
evsaev: Good catch! Introducing `updateContentIfNeeded` is wonderful idea! 🎉 

It also feels like **it should be the only helper** for content update. Having `updateContent` without **superview != nil** check is not safe in terms of UIConfig usage.

Sources_v3/StreamChatUI/ChatChannel/ChatMessageReactionsView.swift
evsaev: Left it as it is because we have a separate user-story for message reactions

Sources_v3/StreamChatUI/MessageActionsPopup/ChatMessagePopupViewController.swift
evsaev: The subviews that can be overridden are owned by `messageActionsController/messageReactionsController` and instantiated from `UIConfig`

Or.. I didn't get your point 😄

Sources_v3/StreamChatUI/MessageActionsPopup/ChatMessageActionItem.swift
```diff
@@ -0,0 +1,107 @@
...
+        self.icon = icon
+        self.isDestructive = isDestructive
+        self.action = action
+    }
+}
+
+public extension ChatMessageActionItem {
+    static func inlineReply(action: @escaping () -> Void) -> Self {
+        .init(
+            title: "Reply",
```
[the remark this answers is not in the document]
evsaev: I'll add localization support right after the PR is merged in [CIS-501](https://stream-io.atlassian.net/secure/RapidBoard.jspa?rapidView=61&projectKey=CIS&modal=detail&selectedIssue=CIS-501).
Will also fix other hardcoded strings across the codebase.

Sources_v3/StreamChatUI/MessageActionsPopup/MessageActionsView+ActionButton.swift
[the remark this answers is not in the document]
evsaev: used `updateContentIfNeeded` in 498372f6ca35c4905742a6142443d28fba092b12

Sources_v3/StreamChatUI/MessageActionsPopup/MessageActionsView+ActionButton.swift
```diff
@@ -0,0 +1,43 @@
...
+        }
+
+        // MARK: Overrides
+
+        override open func defaultAppearance() {
+            backgroundColor = .chatBackground
+        }
+
+        override open func setUpAppearance() {
```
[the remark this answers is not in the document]
evsaev: fixed in 498372f6ca35c4905742a6142443d28fba092b12

Sources_v3/StreamChatUI/MessageActionsPopup/MessageActionsView.swift
evsaev: As discussed in Slack, controllers are updated to inherit from `UIViewController` in 7097ed8282ad0712fb5372836dba3c370639c30d and 498372f6ca35c4905742a6142443d28fba092b12

Sources_v3/StreamChatUI/MessageActionsPopup/ChatMessagePopupViewController.swift
VojtaStavik: Now when `ChatMessagePopupViewController` is a view config provider, it should initialize the subviews via `UIConfig`. We shouldn't override the initializer and set the dependencies (message and frame) the same way we set controllers for example.

Sources_v3/StreamChatUI/ChatChannel/ChatMessageReactionsView.swift
```diff
@@ -226,48 +226,62 @@ open class ChatMessageReactionsView: UIView {
...
 extension ChatMessageReactionViewController: _MessageControllerDelegate {
-    func messageController(
+    public func messageController(
         _ controller: _ChatMessageController<ExtraData>,
         didChangeMessage change: EntityChange<_ChatMessage<ExtraData>>
     ) {
         switch change {
         case .create, .remove: break
-        case .update: reloadView()
+        case .update: updateContent()
```
VojtaStavik, line 284: `updateContentIfNeeded()`

Sources_v3/StreamChatUI/MessageActionsPopup/ChatMessagePopupViewController.swift
```diff
@@ -0,0 +1,125 @@
...
+    override public func defaultAppearance() {
+        view.backgroundColor = UIColor.black.withAlphaComponent(0.2)
+    }
+
+    override open func setUpLayout() {
+        addChild(reactionsController)
+        reactionsController.willMove(toParent: self)
+        reactionsController.view.translatesAutoresizingMaskIntoConstraints = false
+        scrollContentView.addSubview(reactionsController.view)
+        reactionsController.didMove(toParent: self)
```
VojtaStavik, line 61: Why not use a helper for adding a child VC?

VojtaStavik (approved):
Looks good 👍 Left a couple of improvement suggestions

Sources_v3/StreamChatUI/MessageActionsPopup/ChatMessagePopupViewController.swift
evsaev: I **was** about introducing it but such helper usually does `view.addSubview(child.view)` inside but in ouк case the `child.view` should be added to `scrollContentView`

evsaev: The closure-based solution with generic parent controller doesn't work. To apply the helper for message pop-up it should take the view the child should be added to as input param which doesn't look nice:
 
```swift
extension UIViewController {
   //func add(_ child: UIViewController, addSubview: (Self, UIView) -> () = { $0.view.addSubview($1) }) {
    func add(_ child: UIViewController, to view: UIView? = nil) {
        addChild(child)
        child.willMove(toParent: self)
        (view ?? self.view)?.addSubview(child.view)
        child.didMove(toParent: self)
    }
}
```

Sources_v3/StreamChatUI/ChatChannel/ChatMessageReactionsView.swift
evsaev: We set the delegate in `setUp` so there is no way to receive a callback from the delegate before the view is loaded. But let's fix anyway 👍

Sources_v3/StreamChatUI/MessageActionsPopup/ChatMessagePopupViewController.swift
evsaev: added helpers in 6b291e9aa5ed08272d4ae5b268b63e72e1ef434a

Sources_v3/StreamChatUI/ChatChannel/ChatMessageReactionsView.swift
evsaev: fixed in 04bbe6a5652400bd339f7a661887ce0cbcc504c3

Sources_v3/StreamChatUI/MessageActionsPopup/ChatMessagePopupViewController.swift
evsaev: removed `init` and set the dependencies after the controller is created in 35afc5747712e70c838c2a6af65527ed985e46d6

PR #625, opened 2020-12-04
Improve view life cycle methods

VojtaStavik:
### In this PR:

- A new lifecycle method added - `setUp()`. It's meant to be used for the general setup of the view (setting delegates, adding subviews, etc.)
- A new helper method `updateContentIfNeeded()` which calls `updateContent()` only if it's safe.


b-onc (approved):

evsaev (approved):
🚢

PR #626, opened 2020-12-04
[CIS-502] Prevent user from overscrolling out of visible content

ManWithBear:
CollectionViewLayout flip comes with self-sizing cell issue: when you scroll to "estimated" cell bottom anchor, cell size calculated and you may jump right into middle of cell, because `collectionViewContentSize` have been changed, while `contentOffset` stay same.
To fight it we lock `collectionViewContentSize` to super big number simulating "infinite scroll", now newly calculated cell "expands" up into invisible yet area, removing jumps.

### DIFF
--- a/Sources_v3/StreamChatUI/ChatChannel/ChatChannelVC.swift
+++ b/Sources_v3/StreamChatUI/ChatChannel/ChatChannelVC.swift
@@ -18,10 +18,13 @@ open class ChatChannelVC<ExtraData: UIExtraDataTypes>: ViewController,
     public private(set) lazy var messageInputAccessoryViewController: MessageComposerInputAccessoryViewController<ExtraData> = {
         .init()
     }()
-        
+
+    public private(set) lazy var collectionViewLayout: ChatChannelCollectionViewLayout = uiConfig
+        .messageList
+        .collectionLayout
+        .init()
     public private(set) lazy var collectionView: UICollectionView = {
-        let layout = uiConfig.messageList.collectionLayout.init()
-        let collection = uiConfig.messageList.collectionView.init(layout: layout)
+        let collection = uiConfig.messageList.collectionView.init(layout: collectionViewLayout)
         collection.register(
             СhatIncomingMessageCollectionViewCell<ExtraData>.self,
             forCellWithReuseIdentifier: СhatIncomingMessageCollectionViewCell<ExtraData>.reuseId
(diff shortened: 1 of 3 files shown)


Sources_v3/StreamChatUI/UIConfig.swift
```diff
@@ -78,8 +78,10 @@ public extension UIConfig {
 public extension UIConfig {
     struct MessageListUI {
         public var collectionView: ChatChannelCollectionView.Type = ChatChannelCollectionView.self
-        public var collectionLayout: UICollectionViewLayout.Type = ChatChannelCollectionViewLayout.self
+        public var collectionLayout: ChatChannelCollectionViewLayout.Type = ChatChannelCollectionViewLayout.self
         public var minTimeInvteralBetweenMessagesInGroup: TimeInterval = 10
+        /// Point from top, after which we should start loading next batch on messages
+        public var nextMessageLoadingThreshold: CGFloat = 100
```
VojtaStavik, line 84: It took me a while to understand what exactly this is. I don't have a better name though :/

Sources_v3/StreamChatUI/ChatChannel/ChatChannelVC.swift
```diff
@@ -164,11 +172,22 @@ open class ChatChannelVC<ExtraData: UIExtraDataTypes>: ViewController,
...
+
+    public func scrollViewWillEndDragging(
+        _ scrollView: UIScrollView,
+        withVelocity velocity: CGPoint,
+        targetContentOffset: UnsafeMutablePointer<CGPoint>
+    ) {
+        if targetContentOffset.pointee.y < collectionViewLayout.zeroOffset.y {
+            targetContentOffset.pointee.y = collectionViewLayout.zeroOffset.y
+        }
+    }
```
VojtaStavik, line 184: Is there any chance how to do this from within the layout itself?

VojtaStavik (approved):
Looks good 👍 It would be amazing if we were able to encapsulate all the real-zero offset logic only within the layout class without having to pollute the VC with the layout specific things.

Sources_v3/StreamChatUI/UIConfig.swift
ManWithBear: What about:
```
/// Vertical contentOffset for message list, when next message batch should be requested
public var offsetToPreloadMoreMessages: CGFloat = 100
```

Sources_v3/StreamChatUI/ChatChannel/ChatChannelVC.swift
ManWithBear: Actually there is! 
Moved almost all logic inside.
Only thing we cannot prevent is to check offset for message preloading. But preloading can be done differently in general (via `willShowItemAtIndexPath`) so I let it be.

VojtaStavik: Very nice! I think we can live with having it done like this for now. Thank you!

PR #627, opened 2020-12-04
[TEMP] Stack View Gallery

VojtaStavik:
Just a PR so we can discuss things more easily

### DIFF
--- a/Sources_v3/StreamChatUI/ChatChannel/ChatMessageBubbleView.swift
+++ b/Sources_v3/StreamChatUI/ChatChannel/ChatMessageBubbleView.swift
@@ -7,7 +7,7 @@ import UIKit
 
 open class ChatMessageBubbleView<ExtraData: UIExtraDataTypes>: View, UIConfigProvider {
     public var message: _ChatMessageGroupPart<ExtraData>? {
-        didSet { updateContent() }
+        didSet { updateContentIfNeeded() }
     }
 
     public let showRepliedMessage: Bool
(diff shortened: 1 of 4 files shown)


Sources_v3/StreamChatUI/ChatChannel/ChatMessageBubbleView.swift
```diff
@@ -78,7 +87,14 @@ open class ChatMessageBubbleView<ExtraData: UIExtraDataTypes>: View, UIConfigPro
         if let repliedMessageView = repliedMessageView {
             stackView.addArrangedSubview(repliedMessageView)
         }
+        stackView.addArrangedSubview(imageGallery)
         stackView.addArrangedSubview(textView)
+
+        // Here we set the width constaint and expect that when imageGallery is hidden, the constaint is not active
+        imageGallery.widthAnchor.constraint(equalToConstant: UIScreen.main.bounds.width * 0.65).isActive = true
```
VojtaStavik, line 94: Why do we do this? The width of the gallery is defined by the distribution setting of the enclosing stack view, isn't it?

ManWithBear: If you have message with gallery only, it could be any size.
This make gallery size static and consistent between all messages

VojtaStavik: But the width of the message is specified by the layout, isn't it? If we want having square gallery views we could just add 1:1 ration constraint to it.

VojtaStavik: One thing that's for sure is that we shouldn't take the value from `UIScreen.main` :)

ManWithBear: Maximal width of message is limited by 0.75 of row.
Minimal width is not specified.
E.g. 2 messages with same image of size 10x10, first contains just image, second contains long text.
Without this constraint: first message will render as ≈10x10 cauz intrinsic size of gallery is image size, while second message will render as ≈250xLong, and image will be rendered as ≈250x250

PR #628, opened 2020-12-05
[CIS-501] Localization support

evsaev:
**This PR**:
- integrates `SwiftGen` via mint and uses `SwiftGen` for localized strings
- adds codegen build-phase to `StreamChatUI` target
- excludes generated files from `codecov` tool


Sources_v3/StreamChatUI/MessageActionsPopup/ChatMessageActionItem.swift
```diff
@@ -26,7 +26,7 @@ public struct ChatMessageActionItem {
 public extension ChatMessageActionItem {
     static func inlineReply(action: @escaping () -> Void) -> Self {
         .init(
-            title: "Reply",
+            title: "message.actions.inline-reply".localized(),
```
VojtaStavik, line 29: We need to generate these values from the strings file eventully. Ideally, we'd just do `Strings.message.actions.inlineReply` with autocompletion support :)

ManWithBear (changes requested):
Please add autogenerated `L10n` enum for localization constants.
e.g. [SwiftGen](https://github.com/SwiftGen/SwiftGen#strings)

VojtaStavik (approved):
Haha, I have the same comment as @ManWithBear. I don't think it's a blocker for merging this but we should add strings to the resources task https://stream-io.atlassian.net/browse/CIS-426?atlOrigin=eyJpIjoiYTI2MzVjOGQwNmRjNDRhM2IzMmZjMTZiZWMxYTMxOGYiLCJwIjoiaiJ9 . (I added it)

Sources_v3/StreamChatUI/MessageActionsPopup/ChatMessageActionItem.swift
ManWithBear: Only I would stick with standard [L10n](https://en.wiktionary.org/wiki/L10n) instead of `Strings`

VojtaStavik: So geeky...

ManWithBear (approved):
Awesome 👍🏾

Sources_v3/StreamChatUI/Generated/L10n.swift
```diff
@@ -0,0 +1,74 @@
+// swiftlint:disable all
```
VojtaStavik, line 1: Is this line part of the generated file or you added it? If the latter, it will get removed the next time the file gets regenerated.

VojtaStavik (approved):

Sources_v3/StreamChatUI/Generated/L10n.swift
evsaev: The generated one

PR #634, opened 2020-12-08
[CIS-503] Move `name` and `imageURL` to models

b-onc:
Resulting work of 2 days of effort


evsaev (commented):
Does adding `name` and `imageURL` to user/channel allow us to remove all references to `NameAndImageExtraData`?

Sources_v3/StreamChat/Models/Member.swift
```diff
@@ -12,7 +12,7 @@ import Foundation
 ///
 /// Learn more about using custom extra data in our [cheat sheet](https://github.com/GetStream/stream-chat-swift/wiki/StreamChat-SDK-Cheat-Sheet#working-with-extra-data).
 ///
-public typealias ChatChannelMember = _ChatChannelMember<NameAndImageExtraData>
+public typealias ChatChannelMember = _ChatChannelMember<DefaultExtraData.User>
```
VojtaStavik, line 15: This was actually wrong all the time, right? :)

VojtaStavik (approved):
Great 🎉 Let's see what breaks :)

Sources_v3/StreamChat/Models/Member.swift
b-onc: yep, in many places we've used `NameAndImageExtraData` where we should've used `DefaultExtraData.User`, I've refactored them all (but only for Channel and User, obv)

b-onc:
@evsaev yes!

VojtaStavik:
@b-onc This means we can get rid of `UIExtraData` too, right? 🎉

evsaev:
> @evsaev yes!

@b-onc hah, I somehow oversaw the commit with `Remove NameAndImageExtraData` message 🤦 Nice job!

b-onc:
@VojtaStavik yep, it'll come in a separate PR though

PR #635, opened 2020-12-08
[CIS-510] Add thread participants to message model

ManWithBear:
To help with thread preview, backend provides optional field `thread_participants`.
This PR adds support for field to our model


Sources_v3/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -240,14 +241,19 @@ extension NSManagedObjectContext: MessageDatabaseSession {
             let user = try saveUser(payload: userPayload)
             dto.mentionedUsers.insert(user)
         }
+
+        try payload.threadParticipants.forEach { userPayload in
+            let user = try saveUser(payload: userPayload)
+            dto.threadParticipants.insert(user)
+        }
```
evsaev, line 248: If user had left a thread reply and then deleted it will they stay in `threadParticipants` list? 🤔 

I'd suggest to reset the `dto.threadParticipants` to ensure we are consistent with backend.

```suggestion
        dto.threadParticipants = try Set(
          payload.threadParticipants.forEach { try saveUser(payload: $0) }
        )
```

evsaev (approved):
LGTM 👍

Sources_v3/StreamChat/Models/Message.swift
```diff
@@ -82,6 +82,9 @@ public struct _ChatMessage<ExtraData: ExtraDataTypes> {
     
     /// A list of users that are mentioned in this message.
     public let mentionedUsers: Set<_ChatUser<ExtraData.User>>
+
+    /// A list of users that participated in this message thread
+    public let threadParticipants: Set<_ChatUser<ExtraData.User>>
```
VojtaStavik, line 87: There's this rule of thumb we use when designing model APIs. If the field can have a non-specific number of elements inside, we don't expose it as model objects but only as ids. This is sill not done correctly for some cases (see https://stream-io.atlassian.net/browse/CIS-408?atlOrigin=eyJpIjoiYzg0MzRlMzQyMzNjNDVjOTgxYTNjOTY3MzliMzQ5ZWIiLCJwIjoiaiJ9) but `threadParticipants` is the case where it's better to expose just ids.

VojtaStavik: Will add this to Notion

VojtaStavik (approved):
Looks good, just one nit regarding ids. Thank you!

Sources_v3/StreamChat/Database/DTOs/MessageDTO_Tests.swift
```diff
@@ -170,7 +178,7 @@ class MessageDTO_Tests: XCTestCase {
             // Save the message
             let messageDTO = try! session.saveMessage(payload: messagePayload, for: channelId)
             // Make the extra data JSON invalid
-            messageDTO.extraData = #"{"invalid": json}"#.data(using: .utf8)!
+            messageDTO.extraData = #"{"invalid": json}"# .data(using: .utf8)!
```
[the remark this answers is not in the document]
ManWithBear: swiftformat

PR #629, opened 2020-12-06
[CIS-472] Improve ChatChannelMessageComposerView UI

mantergo:
In this PR various UI improvements were applied: 

- Add UI for `edit` and `reply` composer states
- Implement input shrink behaviour on typing 
- Implement correct `send` button behaviour (for editing and sending)
- Improve size calculation and layout
- Clean up `ChatChannelMessageComposerView` class

Some tricky UI stuff is left and will be done in separate ticket cause it takes too much time for the moment.

Screenshot 2020-12-06 at 18 14 33
Screenshot 2020-12-06 at 18 14 42
Screenshot 2020-12-06 at 18 15 14
Screenshot 2020-12-06 at 18 15 23

PR #636, opened 2020-12-08
[CIS-480] Add placeholder to `ChatChannelMessageInputTextView`

mantergo:
In this PR: 
- Add placeholder to `ChatChannelMessageInputTextView`

Should be merged after https://github.com/GetStream/stream-chat-swift/pull/629.

Screenshot 2020-12-08 at 22 54 56
Screenshot 2020-12-08 at 22 55 03
Screenshot 2020-12-08 at 22 55 11
Screenshot 2020-12-08 at 22 55 22

### DIFF
--- a/Sources_v3/StreamChatUI/ChatChannel/ChatChannelMessageInputTextView.swift
+++ b/Sources_v3/StreamChatUI/ChatChannel/ChatChannelMessageInputTextView.swift
@@ -1,0 +1,79 @@
+//
+// Copyright © 2020 Stream.io Inc. All rights reserved.
+//
+
+import UIKit
+
+open class ChatChannelMessageInputTextView<ExtraData: UIExtraDataTypes>: UITextView,
+    AppearanceSetting,
+    Customizable,
+    UIConfigProvider
+{
+    // MARK: - Subviews
+    
+    public lazy var placeholderLabel: UILabel = UILabel().withoutAutoresizingMaskConstraints
+    
+    // MARK: - Overrides
+    
+    override public var text: String! {
+        didSet {
+            textDidChange()
+        }
+    }
+    
+    override public var attributedText: NSAttributedString! {
+        didSet {
+            textDidChange()
+        }
+    }
+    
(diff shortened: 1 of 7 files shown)


Sources_v3/StreamChatUI/ChatChannel/ChatChannelMessageInputTextView.swift
```diff
@@ -0,0 +1,79 @@
...
+        setUpAppearance()
+        setUpLayout()
+        updateContent()
+    }
+    
+    // MARK: Public
+    
+    open func defaultAppearance() {
+        font = .preferredFont(forTextStyle: .callout)
+        textContainer.lineFragmentPadding = 10
```
VojtaStavik, line 45: Is it a good idea to introduce additional inner padding of the text? Wouldn't it be better to set this to `0` a position the view correctly within its superview?

Sources_v3/StreamChatUI/ChatChannel/ChatChannelMessageInputTextView.swift
```diff
@@ -0,0 +1,79 @@
...
+            object: nil
+        )
+    }
+    
+    open func setUpAppearance() {}
+    
+    open func setUpLayout() {
+        embed(placeholderLabel, insets: .init(
+            top: .zero,
+            leading: textContainer.lineFragmentPadding,
```
VojtaStavik, line 66: ... this workaround won't be neede then

VojtaStavik (approved):
Looks good 👍

DominikBucher12 (approved):
LGTM 👍🏼

Sources_v3/StreamChatUI/ChatChannel/ChatChannelMessageInputTextView.swift
mantergo: Actually we are not introducing it.
The default value was `5` and we need to adjust it just because of `cornerRadius` so it will look better.
IMO it's the right property to play with in this case.

PR #637, opened 2020-12-09
Fix UIConfig not being passed down for VCs

b-onc:
### Problem

If we use the VC's `view` property for storing/getting `uiConfig`, we'll assume by the time we need it, `view` will be initialised.

In the old impl, when we call:
```swift
chatList.uiConfig = uiConfig
```
we call `register`, which in turn does:
```
func register<T: ExtraDataTypes>(config: UIConfig<T>) {
        view.anyUIConfig = config
}
```
which accesses `view` so `loadView` and `viewDidLoad` gets called. This initializes all subviews of the VC _before actually assigning_ `uiConfig`

So we need to be able to assign `uiConfig` before initializing `view`

### DIFF
--- a/Sources_v3/StreamChatUI/Utils/UIConfigProvider.swift
+++ b/Sources_v3/StreamChatUI/Utils/UIConfigProvider.swift
@@ -7,11 +7,11 @@ import UIKit
 
 // MARK: - Stored property in UIView required to make this work
 
-private extension UIView {
+private extension UIResponder {
     static var anyUIConfigKey: UInt8 = 0
     
     var anyUIConfig: Any? {
-        get { objc_getAssociatedObject(self, &Self.anyUIConfigKey) as? String }
+        get { objc_getAssociatedObject(self, &Self.anyUIConfigKey) }
         set { objc_setAssociatedObject(self, &Self.anyUIConfigKey, newValue, .OBJC_ASSOCIATION_RETAIN) }
     }
 }
@@ -30,7 +30,7 @@ public protocol UIConfigProvider: GenericUIConfigProvider {
 
 // MARK: - Protocol extensions for UIView
 
-public extension GenericUIConfigProvider where Self: UIView {
+public extension GenericUIConfigProvider where Self: UIResponder {
     func register<T: ExtraDataTypes>(config: UIConfig<T>) {
         anyUIConfig = config
     }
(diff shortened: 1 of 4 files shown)


Sources_v3/StreamChatUI/Utils/UIConfigProvider.swift
```diff
@@ -11,7 +11,16 @@ private extension UIView {
     static var anyUIConfigKey: UInt8 = 0
     
     var anyUIConfig: Any? {
-        get { objc_getAssociatedObject(self, &Self.anyUIConfigKey) as? String }
+        get { objc_getAssociatedObject(self, &Self.anyUIConfigKey) }
+        set { objc_setAssociatedObject(self, &Self.anyUIConfigKey, newValue, .OBJC_ASSOCIATION_RETAIN) }
+    }
+}
+
+private extension UIViewController {
```
ManWithBear, line 19: Why not go with extension on `NSObject` instead?

Sources_v3/StreamChatUI/Utils/UIConfigProvider.swift
```diff
@@ -72,23 +81,23 @@ extension UIConfigProvider where Self: UIView {
 
 public extension GenericUIConfigProvider where Self: UIViewController {
     func register<T: ExtraDataTypes>(config: UIConfig<T>) {
-        view.anyUIConfig = config
+        anyUIConfig = config
```
ManWithBear, line 84: How view will get vc uiconfig now?

VojtaStavik (approved):
I think we're missing the part where we pass the config from the view controllers to their view 🤔 . Otherwise `viewController.uiConfig = configA` and `viewController.view.uiConfig = configB` would result in having two different hierarchies with two different configs

Sources_v3/StreamChatUI/Utils/UIConfigProvider.swift
ManWithBear: Actually if we go with `UIResponder` approach, then we can iterate over Responder chain via `UIResponder.next`

For `UIView` it will return either associated `UIViewController` if exist or `superview`
For `UIViewController`  it will return his view superview
https://developer.apple.com/documentation/uikit/uiresponder/1621099-next

b-onc: You're right, and the solution you've suggested would also solve passing vc's config to it's view.. I'll try it 👍 
cc @VojtaStavik

b-onc:
@VojtaStavik @ManWithBear PR updated with responder solution, I wrote some tests just to make sure it's working as expected, and it is 🎉

VojtaStavik (approved):
Very nice 👏

ManWithBear (approved):
Awesome! 👍🏾

PR #638, opened 2020-12-10
Add option to disable scroll to bottom when new messages are received

cardoso:
The table view always scrolls to the bottom when new messages are received. This can be annoying if you're looking at older messages and the channel is very active. I added an option `scrollOnNewData` that when set to `false` disables this behavior. If it's false, the table view will only scroll if it's already at the bottom.

Decided to rename `scrollEnabled` to `canScroll` so it reads better as a passive computed property.


b-onc (approved):
Please add a CHANGELOG :) Good work!

PR #640, opened 2020-12-10
[CIS-488] [Demo] Prototype Channel creation screens

b-onc:
This PR includes:
- DM Channel creation screen
- Group channel creation screen
- Group overview/naming screen

Which should've been 3 separate PRs.\

Prototypes will be polished in an upcoming PR

### DIFF
--- a/DemoApp/ChatPresenter.swift
+++ b/DemoApp/ChatPresenter.swift
@@ -9,6 +9,8 @@ import UIKit
 extension UIViewController {
     // TODO: Where to put this???
     func presentChat(userCredentials: UserCredentials) {
+        LogConfig.level = .debug
+        
         // Create client
         let config = ChatClientConfig(apiKey: .init(userCredentials.apiKey))
         let client = ChatClient(config: config)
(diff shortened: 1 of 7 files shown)


DemoApp/ChatPresenter.swift
```diff
@@ -33,3 +46,16 @@ extension UIViewController {
...
 }
+
+class DemoChatChannelListRouter: ChatChannelListRouter<DefaultExtraData> {
+    override func openCreateNewChannel() {
+        let client = rootViewController?.controller.client
+        let storyboard = UIStoryboard(name: "Main", bundle: .main)
+        
+        let chatViewController = storyboard.instantiateViewController(withIdentifier: "CreateChatViewController")
+            as! CreateChatViewController
+        chatViewController.client = client
```
VojtaStavik, line 57: nit: We usually pass in the controller instances, not the whole client.

DemoApp/CreateChatViewController.swift
```diff
@@ -0,0 +1,251 @@
...
+        userListController.delegate = self
+        userListController.synchronize { error in
+            if let error = error {
+                print("Error", error)
+            }
+        }
+        
+        delegate?.userSearchController(self, didFetchUsers: userListController.users)
+        
+        print("Searching", searchText)
```
VojtaStavik, line 57: The controller instance shouldn't change, only the query should -> Slack

DemoApp/CreateChatViewController.swift
```diff
@@ -0,0 +1,251 @@
...
+    }
+    
+    func input(text: String?) {
+        operation?.cancel()
+        operation = DispatchWorkItem { [weak self] in
+            self?.setUserListController(for: text)
+        }
```
VojtaStavik, line 64: 👍

DemoApp/CreateChatViewController.swift
```diff
@@ -0,0 +1,251 @@
...
+        
+        tableView.endUpdates()
+        
+        noMatchView.isHidden = !users.isEmpty
+    }
+}
+
+extension UIImage {
+    func resized(to targetSize: CGSize) -> UIImage {
```
VojtaStavik, line 233: Since we call this for every cell without cashing it, my guess would be that this is actually worse than just putting full images into the avatars :))

DemoApp/NameGroupViewController.swift
```diff
@@ -0,0 +1,157 @@
...
+            
+            removeButton.tintColor = .black
+            removeButton.setImage(UIImage(systemName: "xmark"), for: .normal)
+            removeButton.imageView?.contentMode = .scaleAspectFit
+            
+            NSLayoutConstraint.activate([
+                // AvatarView
+                avatarView.leftAnchor.constraint(equalTo: contentView.leftAnchor, constant: 0),
+                avatarView.topAnchor.constraint(equalTo: contentView.topAnchor, constant: 12),
+                avatarView.bottomAnchor.constraint(equalTo: contentView.bottomAnchor, constant: -12),
```
VojtaStavik, line 42: layout margins ;-)

VojtaStavik (commented):
Yep, I think something like this is what we need 👍

DemoApp/CreateChatViewController.swift
b-onc: We call this for search tokens (so only selected cells)

PR #641, opened 2020-12-10
[CIS-499] User Search Controller

b-onc:
### Idea:
SearchController creates a unique query and observes DB for updates to this query. Whenever the developer calls `search(text)` searchController will update this query (in DB) and will report results to the user. In `deinit`, searchController will remove this query from DB. 
The developer will always get `.insert(item` changes, never `.update`) (will be tested)
Additionally, this query will not be observed by `NewUserQueryUpdater` since it's an "ephemeral" query.

### DIFF
--- a/Sources_v3/StreamChat/Controllers/SearchControllers/UserSearchController.swift
+++ b/Sources_v3/StreamChat/Controllers/SearchControllers/UserSearchController.swift
@@ -1,0 +1,326 @@
+//
+// Copyright © 2020 Stream.io Inc. All rights reserved.
+//
+
+import CoreData
+import Foundation
+
+extension _ChatClient {
+    /// Creates a new `_ChatUserSearchController` with the provided user query.
+    ///
+    /// - Parameter query: The query specify the filter and sorting of the users the controller should fetch.
+    ///
+    /// - Returns: A new instance of `_ChatUserSearchController`.
+    ///
+    public func userSearchController() -> _ChatUserSearchController<ExtraData> {
+        .init(client: self)
+    }
+}
+
+/// `_ChatUserSearchController` is a controller class which allows observing a list of chat users based on the provided query.
+///
+/// Learn more about `_ChatUserSearchController` and its usage in our [cheat sheet](https://github.com/GetStream/stream-chat-swift/wiki/StreamChat-SDK-Cheat-Sheet#user-list).
+///
+/// - Note: `ChatUserSearchController` is a typealias of `_ChatUserSearchController` with default extra data. If you're using
+/// custom extra data, create your own typealias of `_ChatUserSearchController`.
+///
+/// Learn more about using custom extra data in our [cheat sheet](https://github.com/GetStream/stream-chat-swift/wiki/StreamChat-SDK-Cheat-Sheet#working-with-extra-data).
+///
+public typealias ChatUserSearchController = _ChatUserSearchController<DefaultExtraData>
+
+public class _ChatUserSearchController<ExtraData: ExtraDataTypes>: DataController, DelegateCallable, DataStoreProvider {
+    /// The `ChatClient` instance this controller belongs to.
+    public let client: _ChatClient<ExtraData>
+    
+    /// Filter hash this controller observes.
+    let explicitFilterHash = UUID().uuidString
+    
+    lazy var query: UserListQuery<ExtraData.User> = {
+        // Filter is just a mock, explicit hash will override it
+        var query = UserListQuery<ExtraData.User>(filter: .exists(.id))
+        // Setting `shouldBeObserved` to false prevents NewUserQueryUpdater to pick this query up
+        query.shouldBeUpdatedInBackground = false
+        // The initial DB fetch will return 0 users and this is expected
+        // In the future we'll implement DB search too
+        query.filter?.explicitHash = explicitFilterHash
+        
+        return query
+    }()
+    
+    /// Copy of last search query made, used for getting next page.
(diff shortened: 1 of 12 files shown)


VojtaStavik (commented):
Looks really good 👍

Sources_v3/StreamChat/Controllers/SearchControllers/UserSearchController.swift
```diff
@@ -136,18 +144,55 @@ public class _ChatUserSearchController<ExtraData: ExtraDataTypes>: DataControlle
     }
     
     public func search(term: String, completion: ((_ error: Error?) -> Void)? = nil) {
```
VojtaStavik, line 146: Please add documentation

Sources_v3/StreamChat/Controllers/SearchControllers/UserSearchController.swift
```diff
@@ -136,18 +144,55 @@ public class _ChatUserSearchController<ExtraData: ExtraDataTypes>: DataControlle
...
+            // with this filter hash (see `userListQueryUpdater` above)
+            
+            self.lastQuery = query
+            
+            self.userQueryUpdater.update(userListQuery: query) { [weak self] error in
+                guard let self = self else { return }
+                self.state = error == nil ? .remoteDataFetched : .remoteDataFetchFailed(ClientError(with: error))
+                self.callback { completion?(error) }
+            }
+        }
```
VojtaStavik: I think this is unnecessary complex and also not that elegant as it can be :) Also, it has this ugly intermediate step when the previous results disappear before the new results are visible. My suggestion:
- move all this logic to `userQueryUpdater`
- introduce: 
```swift
enum MergeStrategy {
    case merge, replace
}
update(userListQuery: query, resultsMergeStrategy: .replace) { }
```
The code here will be nicer and more expressive + it will allow us to do everything in a single session.

What do you think?

Sources_v3/StreamChat/Controllers/SearchControllers/UserSearchController.swift
```diff
@@ -136,18 +144,55 @@ public class _ChatUserSearchController<ExtraData: ExtraDataTypes>: DataControlle
...
+        completion: ((Error?) -> Void)? = nil
+    ) {
+        guard let lastQuery = lastQuery else {
+            completion?(ClientError("You should make a search before calling for next page."))
+            return
+        }
+        
+        var updatedQuery = lastQuery
+        updatedQuery.pagination = Pagination(pageSize: limit, offset: users.count)
+        userQueryUpdater.update(userListQuery: updatedQuery) { [weak self] error in
```
VojtaStavik, line 194: ... and this will be:
```swift
userQueryUpdater.update(userListQuery: updatedQuery, mergeStrategy: .merge) { ... }
```

VojtaStavik (changes requested):
Looks good 👍 I have a suggestion though :) 
Also, kinda sad you didn't like my suggestion here https://github.com/GetStream/stream-chat-swift/pull/641#discussion_r540283244

Sources_v3/StreamChat/Database/DTOs/UserListQueryDTO.swift
```diff
@@ -12,10 +12,19 @@ class UserListQueryDTO: NSManagedObject {
     /// Serialized `Filter` JSON which can be used in cases the query needs to be repeated.
     @NSManaged var filterJSONData: Data
     
+    /// Indicates if the query should be observed by background workers.
+    @NSManaged var shouldBeObserved: Bool
+
```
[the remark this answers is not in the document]
VojtaStavik: OK, one last try: `shouldBeUpdatedInBackgronud` `isEligibleForBackgroundUpdates` ... basically, something else than `shouldBeObserved` because it's too vague imo.

Sources_v3/StreamChat/Query/UserListQuery.swift
```diff
@@ -75,6 +75,9 @@ public struct UserListQuery<ExtraData: UserExtraData>: Encodable {
     /// Query options. By default the query options contain `presence`.
     var options: QueryOptions = [.presence]
     
+    /// Indicates if the query should be observed for new users.
+    var shouldBeObserved = true
+
```
VojtaStavik, line 80: The documentation here is not the best. What about: `If set to true, newly created users in the database are automatically included in the query if they fit the predicate.`

Sources_v3/StreamChat/Workers/UserListUpdater_Tests.swift
```diff
@@ -80,4 +80,142 @@ class UserListUpdater_Tests: StressTestCase {
...
+        try database.writeSynchronously { session in
+            do {
+                let dto = try session.saveQuery(query: query)
+                XCTAssertEqual(dto!.users.count, 2)
+                XCTAssertEqual(dto!.users.map(\.id), [user!, newUser!].map(\.id))
+            } catch {
+                XCTFail("Error trying to get query: \(error)")
+            }
+        }
+    }
```
VojtaStavik, line 220: This is a redundant test if you use `merge` as default on L103 :)

VojtaStavik (approved):
Great job, thank you! 🎉

Sources_v3/StreamChat/Workers/UserListUpdater_Tests.swift
b-onc: This is testing the default merge policy, no argument is being passed for this test case

PR #643, opened 2020-12-11
Fallback on inconsistent table view update (fix crash)

cardoso:
Discussed in internal channel


VojtaStavik (approved):
Looks good as a quick workaround to prevent crashes. Should we do the same check also for removed items?

cardoso:
@VojtaStavik could be, but I'm not sure how to test that case.

PR #644, opened 2020-12-12
[CIS-475] Replies UI in MessageComposer

mantergo:
In this PR:

- Improve `UITextView` text height calculation
- Introduce `ChatReplyBubbleView`
- Make use of `ChatReplyBubbleView` in `ChatChannelMessageComposerView`
- Adjust `MessageComposerInputAccessoryViewController.State` and add test reply functionality

Screenshot 2020-12-12 at 18 04 50
Screenshot 2020-12-12 at 18 05 09
Screenshot 2020-12-12 at 18 05 52

### DIFF
--- a/Sources_v3/StreamChatUI/ChatChannel/ChatReplyBubbleView.swift
+++ b/Sources_v3/StreamChatUI/ChatChannel/ChatReplyBubbleView.swift
@@ -1,0 +1,158 @@
+//
+// Copyright © 2020 Stream.io Inc. All rights reserved.
+//
+
+import StreamChat
+import UIKit
+
+open class ChatReplyBubbleView<ExtraData: ExtraDataTypes>: View, UIConfigProvider {
+    // MARK: - Properties
+    
+    public var avatarViewWidth: CGFloat = 24
+    public var attachmentPreviewWidth: CGFloat = 34
+    
+    public var message: _ChatMessage<ExtraData>? {
+        didSet {
+            updateContentIfNeeded()
+        }
+    }
+    
+    lazy var textViewHeightConstraint = textView.heightAnchor.constraint(greaterThanOrEqualToConstant: .zero)
+    
+    // MARK: - Subviews
+    
+    public private(set) lazy var container = ContainerStackView()
+        .withoutAutoresizingMaskConstraints
+    
+    public private(set) lazy var authorAvatarView = uiConfig
+        .messageComposer
+        .replyBubbleAvatarView.init()
+        .withoutAutoresizingMaskConstraints
+    
+    public private(set) lazy var attachmentPreview = UIImageView()
+        .withoutAutoresizingMaskConstraints
+
+    public private(set) lazy var textView = UITextView()
+        .withoutAutoresizingMaskConstraints
+    
+    // MARK: - Public
+    
+    override open func setUp() {
+        super.setUp()
+        
+        textView.isEditable = false
+        textView.dataDetectorTypes = .link
+        textView.isScrollEnabled = false
+        textView.adjustsFontForContentSizeCategory = true
+        textView.isUserInteractionEnabled = false
+    }
+    
+    override open func defaultAppearance() {
+        textView.textContainer.maximumNumberOfLines = 6
+        textView.textContainer.lineBreakMode = .byTruncatingTail
+        textView.textContainer.lineFragmentPadding = .zero
+        
+        textView.backgroundColor = .clear
+        textView.font = .preferredFont(forTextStyle: .footnote)
+        textView.textContainerInset = .zero
+
+        authorAvatarView.contentMode = .scaleAspectFit
+        
+        attachmentPreview.layer.cornerRadius = attachmentPreviewWidth / 4
+        attachmentPreview.layer.masksToBounds = true
+        
+        container.centerStackView.layer.cornerRadius = 16
+        container.centerStackView.layer.borderWidth = 1
+        container.centerStackView.layer.borderColor = uiConfig.colorPalette.messageComposerBorder.cgColor
+        container.centerStackView.layer.masksToBounds = true
+    }
+    
+    override open func setUpLayout() {
+        embed(container)
+        
+        preservesSuperviewLayoutMargins = true
+        
+        container.preservesSuperviewLayoutMargins = true
+        container.isLayoutMarginsRelativeArrangement = true
+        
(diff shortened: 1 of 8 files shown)


Sources_v3/StreamChatUI/Utils/UITextView+Extensions.swift
```diff
@@ -0,0 +1,31 @@
...
+//
+
+import UIKit
+
+extension UITextView {
+    ///
+    /// Reference:
+    /// https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/TextLayout/Tasks/StringHeight.html
+    ///
+    var calculatedTextHeight: CGFloat {
```
VojtaStavik, line 12: This should probably be a method, not a computer variable.

Sources_v3/StreamChatUI/Utils/UITextView+Extensions.swift
```diff
@@ -0,0 +1,31 @@
+//
+// Copyright © 2020 Stream.io Inc. All rights reserved.
+//
+
+import UIKit
+
+extension UITextView {
```
VojtaStavik, line 7: I'm not sure putting this into an extension is a good idea. It doesn't allow anyone to override it and change it.

Sources_v3/StreamChatUI/Utils/UITextView+Extensions.swift
```diff
@@ -0,0 +1,31 @@
...
+    ///
+    /// Reference:
+    /// https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/TextLayout/Tasks/StringHeight.html
+    ///
+    var calculatedTextHeight: CGFloat {
+        let textStorage = NSTextStorage(string: text)
+        let width = frame.width - textContainerInset.right - textContainerInset.left
+        let customTextContainer = NSTextContainer(size: .init(width: width, height: CGFloat.greatestFiniteMagnitude))
+        let layoutManager = NSLayoutManager()
```
VojtaStavik, line 16: Why do you create new instances of `NSTextContainer` and `NSLayoutManager`? Can't we just use to _actual_ container and manager of the text view?

Sources_v3/StreamChatUI/ChatChannel/ChatReplyBubbleView.swift
```diff
@@ -0,0 +1,159 @@
...
+//
+
+import StreamChat
+import UIKit
+
+open class ChatReplyBubbleView<ExtraData: ExtraDataTypes>: View, UIConfigProvider {
+    // MARK: - Properties
+    
+    public var avatarViewWidth: CGFloat = .zero
+    public var attachmentPreviewWidth: CGFloat = .zero
```
VojtaStavik, line 12: I find this pattern very confusing. Here it says `.zero` but in `defaultAppearance` it' set to 24 🤔

Sources_v3/StreamChatUI/ChatChannel/ChatReplyBubbleView.swift
```diff
@@ -0,0 +1,159 @@
...
+    }
+    
+    lazy var textViewHeightConstraint = textView.heightAnchor.constraint(greaterThanOrEqualToConstant: .zero)
+    
+    // MARK: - Subviews
+    
+    public private(set) lazy var container = ContainerStackView()
+        .withoutAutoresizingMaskConstraints
+    
+    public private(set) lazy var authorAvatarView = AvatarView()
```
VojtaStavik, line 27: from `UIConfig`

Sources_v3/StreamChatUI/ChatChannel/MessageComposerInputAccessoryViewController.swift
```diff
@@ -35,15 +33,12 @@ open class MessageComposerInputAccessoryViewController<ExtraData: ExtraDataTypes
...
         didSet {
-            state = .reply
-        }
-    }
-    
-    public var messageToEdit: _ChatMessage<ExtraData>? {
-        didSet {
-            state = .edit
+            composerView.attachmentButton.isHidden = !isEmpty
+            composerView.commandsButton.isHidden = !isEmpty
```
VojtaStavik, line 39: You can use isVisible 🤯

VojtaStavik (approved):
Looks good 👍 Just a couple of minor comments.

Sources_v3/StreamChatUI/Utils/UITextView+Extensions.swift
mantergo: But I don't expect this method to be overriden.
It can be changed in future if there will be some adjustments and they will be applied to all places it is used and that's how I want this to work.

Sources_v3/StreamChatUI/Utils/UITextView+Extensions.swift
mantergo: Cause we are creating artificial `NSTextContainer` here with our custom size.
We don't want to change anything about current `UITextView`.

Sources_v3/StreamChatUI/ChatChannel/ChatReplyBubbleView.swift
mantergo: Yeah, can agree.
Should we put 24 while declaring the variable or make it optional?

Sources_v3/StreamChatUI/ChatChannel/MessageComposerInputAccessoryViewController.swift
mantergo: TBH, I found `isVisible` API confusing a bit cause it means some of the properties we set will be `isHidden` and some will be `isVisible`. I feel like it will take more time for me to understand from the glance what is happening here. Or the idea is to use `isVisible` all the time?

Sources_v3/StreamChatUI/Utils/UITextView+Extensions.swift
VojtaStavik: Is there any specific reason you to explicitly prevent users from overriding this?

mantergo: Cause the purpose of this property is to return fitting height of `UITextView` and it should work the same way all the time for all `UITextViews`. There is no need to override it cause behaviour should be the same. If someone want to override it then someone needs new different property that is not related to this one.

PR #646, opened 2020-12-14
Fix markRead requests failing

b-onc:
They were failing since backend expects a valid body (not nil) for post requests


VojtaStavik (changes requested):
Shouldn't we rater solve it on the request encoder level? If `nil` and `post` encode empty body, not `nil`. No need to introduce this special case higher in the codebase.

b-onc:
@VojtaStavik Good catch, updated

VojtaStavik (approved):
Please add a test for that and we're good to 🚢

PR #647, opened 2020-12-14
Add CI v2 workflow and separate v2/v3 CIs

b-onc:
Also update the carthage key, in the hopes that this will resolve the caching issue

No longer our v2 jobs will run unnecessarily


VojtaStavik (approved):
Looks good 👍 Thanks, good idea.

PR #648, opened 2020-12-14
[CIS-463] Reuse Chat channel controller for threads

ManWithBear:
In this PR I:
- reduce amount of layout warnings within message view
- extract message list in separate view controller
- create thread view controller

### DIFF
--- a/Sources_v3/StreamChatUI/ChatChannel/ChatChannelVC.swift
+++ b/Sources_v3/StreamChatUI/ChatChannel/ChatChannelVC.swift
@@ -5,310 +5,96 @@
 import StreamChat
 import UIKit
 
-open class ChatChannelVC<ExtraData: ExtraDataTypes>: ViewController,
-    UICollectionViewDataSource,
-    UICollectionViewDelegate,
-    UIImagePickerControllerDelegate,
-    UINavigationControllerDelegate,
-    UIConfigProvider {
+open class ChatChannelVC<ExtraData: ExtraDataTypes>: ChatVC<ExtraData> {
     // MARK: - Properties
-    
-    public var controller: _ChatChannelController<ExtraData>!
-
-    public private(set) lazy var messageInputAccessoryViewController: MessageComposerInputAccessoryViewController<ExtraData> = {
-        let inputAccessoryVC = MessageComposerInputAccessoryViewController<ExtraData>()
-        
-        // `inputAccessoryViewController` is part of `_UIKeyboardWindowScene` so we need to manually pass
-        // tintColor down that `inputAccessoryViewController` view hierarchy.
-        inputAccessoryVC.view.tintColor = view.tintColor
-        
-        return inputAccessoryVC
-    }()
-
-    public private(set) lazy var collectionViewLayout: ChatChannelCollectionViewLayout = uiConfig
-        .messageList
-        .collectionLayout
-        .init()
-    public private(set) lazy var collectionView: UICollectionView = {
-        let collection = uiConfig.messageList.collectionView.init(layout: collectionViewLayout)
-        collection.register(
-            СhatIncomingMessageCollectionViewCell<ExtraData>.self,
-            forCellWithReuseIdentifier: СhatIncomingMessageCollectionViewCell<ExtraData>.reuseId
-        )
-        collection.register(
-            СhatOutgoingMessageCollectionViewCell<ExtraData>.self,
-            forCellWithReuseIdentifier: СhatOutgoingMessageCollectionViewCell<ExtraData>.reuseId
-        )
-        collection.showsHorizontalScrollIndicator = false
-        collection.dataSource = self
-        collection.delegate = self
-        
-        return collection
-    }()
-
-    private var navbarListener: ChatChannelNavigationBarListener<ExtraData>?
 
     public private(set) lazy var router = uiConfig.navigation.channelDetailRouter.init(rootViewController: self)
     
     // MARK: - Life Cycle
     
-    override open func viewDidLoad() {
-        super.viewDidLoad()
+    override open func setUp() {
+        super.setUp()
 
-        controller.setDelegate(self)
-        controller.synchronize()
-        navigationItem.largeTitleDisplayMode = .never
-
-        installLongPress()
-        messageInputAccessoryViewController.controller = controller
+        channelController.setDelegate(self)
+        channelController.synchronize()
     }
 
-    override open func viewDidDisappear(_ animated: Bool) {
-        super.viewDidDisappear(animated)
-        resignFirstResponder()
-    }
-    
-    override open func setUpLayout() {
-        super.setUpLayout()
-
-        view.addSubview(collectionView)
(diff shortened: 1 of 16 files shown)


Sources_v3/StreamChatUI/ChatChannel/ChatMessageContentView.swift
```diff
@@ -89,21 +89,21 @@ open class ChatMessageContentView<ExtraData: ExtraDataTypes>: View, UIConfigProv
...
             messageMetadataView.bottomAnchor.constraint(equalTo: bottomAnchor),
-
+            
             threadArrowView.widthAnchor.constraint(equalToConstant: 16),
             threadArrowView.topAnchor.constraint(equalTo: messageBubbleView.centerYAnchor),
             threadArrowView.bottomAnchor.constraint(equalTo: threadView.centerYAnchor),
-
+            
             threadView.topAnchor.constraint(equalToSystemSpacingBelow: messageBubbleView.bottomAnchor, multiplier: 1)
         ])
```
VojtaStavik, line 108: How come this is not fixed by the formatter 🤔

Sources_v3/StreamChatUI/ChatChannel/ChatMessageListVC.swift
```diff
@@ -0,0 +1,333 @@
...
+    public private(set) lazy var collectionView: UICollectionView = {
+        let collection = uiConfig.messageList.collectionView.init(layout: collectionViewLayout)
+        collection.register(
+            СhatIncomingMessageCollectionViewCell<ExtraData>.self,
+            forCellWithReuseIdentifier: СhatIncomingMessageCollectionViewCell<ExtraData>.reuseId
+        )
+        collection.register(
+            СhatOutgoingMessageCollectionViewCell<ExtraData>.self,
+            forCellWithReuseIdentifier: СhatOutgoingMessageCollectionViewCell<ExtraData>.reuseId
+        )
```
VojtaStavik, line 71: These things should also come from `UIConfig`

Sources_v3/StreamChatUI/ChatChannel/ChatMessageListVC.swift
```diff
@@ -0,0 +1,333 @@
...
+    }()
+
+    // MARK: - Life Cycle
+
+    override open func viewDidLoad() {
+        super.viewDidLoad()
+
+        let longPress = UILongPressGestureRecognizer(target: self, action: #selector(didLongPress(_:)))
+        collectionView.addGestureRecognizer(longPress)
+    }
```
VojtaStavik, line 86: Please use `setUp` methods for the general setup of the view controller.

Sources_v3/StreamChatUI/ChatChannel/ChatMessageListVC.swift
```diff
@@ -0,0 +1,333 @@
...
+                    collectionView.insertItems(at: [index])
+                case let .move(_, fromIndex, toIndex):
+                    collectionView.moveItem(at: fromIndex, to: toIndex)
+                case let .remove(_, index):
+                    collectionView.deleteItems(at: [index])
+                case let .update(_, index):
+                    collectionView.reloadItems(at: [index])
+                }
+            }
+        }, completion: completion)
```
VojtaStavik, line 126: nit: Should we introduce some kind of helper for this boilerplate? Also, I think the reloading should be probably done directly by changing the content of the cell if the cell is visible, not by reloading it completely since we don't change the cell type. This will remove the ugly fade animation.

Sources_v3/StreamChatUI/ChatChannel/ChatMessageListVC.swift
```diff
@@ -0,0 +1,333 @@
...
+            messageAtIndex: { _, _ in
+                fatalError("Method shouldn't be called on empty data source")
+            },
+            loadMoreMessages: { _ in },
+            replyMessageForMessageAtIndex: { _, _, _ in nil },
+            controllerForMessage: { _, _ in
+                fatalError("Method shouldn't be called on empty data source")
+            }
+        )
+    }
```
VojtaStavik, line 321: What was the reasoning for having `empty` instead of making the data source optional?

Sources_v3/StreamChatUI/ChatChannel/ChatChannelVC.swift
```diff
@@ -55,260 +15,86 @@ open class ChatChannelVC<ExtraData: ExtraDataTypes>: ViewController,
...
-
-        installLongPress()
-        messageInputAccessoryViewController.controller = controller
-    }
-
-    override open func viewDidDisappear(_ animated: Bool) {
-        super.viewDidDisappear(animated)
-        resignFirstResponder()
+        channelController.setDelegate(self)
+        channelController.synchronize()
```
VojtaStavik, line 19: -> `setUp`

VojtaStavik (commented):
I don't know what to think about this, to be honest. If you feel strongly about this being the correct direction, let's continue and see how it will work. 

I must say my gut feeling is this abstraction is unnecessary. Can't we achieve this in a more simple way by just reusing the types we have (collection view layout, cell classes, etc.) and some copy-paste boilerplate? You know, by following the rule: used twice -> copy&paste, used thrice -> generalize.

Sources_v3/StreamChatUI/ChatChannel/ChatMessageContentView.swift
ManWithBear: Actually I think it's my formatter adding whitespaces. At least 
```
mint run swiftformat Sources_v3/StreamChatUI/ChatChannel/ChatMessageContentView.swift
```
Do nothing ¯\_(ツ)_/¯

Sources_v3/StreamChatUI/ChatChannel/ChatMessageListVC.swift
ManWithBear: Whole controller doesn't make much sense without datasource. So datasource is required.
Since we cannot use custom init, it needs to be either:
- optional: need to unwrap it everywhere and fallback to something, makes it dirty and just pollute codebase
- implicitly unwrapped optional: better "fall early" approach
- empty: middle ground that plays safe, but solves optional pollution

Sources_v3/StreamChatUI/ChatChannel/ChatMessageListVC.swift
ManWithBear: I think it's definitely something we could work on later 👍🏾

ManWithBear:
> I must say my gut feeling is this abstraction is unnecessary. Can't we achieve this in a more simple way by just reusing the types we have (collection view layout, cell classes, etc.) and some copy-paste boilerplate? You know, by following the rule: used twice -> copy&paste, used thrice -> generalize.
I see where it coming from and agree with it in general.
In this case I think it will pay out, controller really differ only in data source + navbar title, everything else should be same

VojtaStavik (approved):

PR #649, opened 2020-12-14
[CIS-518] Restart attachment uploading API

evsaev:
**This PR:**
- exposes restart attachment uploading API by `ChatMessageController`
- exposes restart attachment uploading API by `MessageUpdater`
- introduces `AttachmentDoesNotExist` and `AttachmentEditing` error types

### DIFF
--- a/Sources_v3/StreamChat/Controllers/MessageController/MessageController.swift
+++ b/Sources_v3/StreamChat/Controllers/MessageController/MessageController.swift
@@ -336,6 +336,22 @@ public extension _ChatMessageController {
             }
         }
     }
+    
+    /// Updates local state of attachment with provided `id` to be enqueued by attachment uploader.
+    /// - Parameters:
+    ///   - id: The attachment identifier.
+    ///   - completion: The completion. Will be called on a **callbackQueue** when the database operation is finished.
+    ///                 If operation fails, the completion will be called with an error.
+    func restartFailedAttachmentUploading(
+        with id: AttachmentId,
+        completion: ((Error?) -> Void)? = nil
+    ) {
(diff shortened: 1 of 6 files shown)


Sources_v3/StreamChat/Controllers/ChannelController/ChannelController.swift
```diff
@@ -190,6 +190,12 @@ public class _ChatChannelController<ExtraData: ExtraDataTypes>: DataController,
         client.webSocketClient,
         client.apiClient
     )
+
+    private lazy var messageUpdater: MessageUpdater<ExtraData> = self.environment.messageUpdaterBuilder(
+        client.databaseContainer,
+        client.webSocketClient,
+        client.apiClient
+    )
```
VojtaStavik, line 198: Hmm, why did you decided to do this? It's not a pattern we usually use. Message-related methods belong to message controller 🤔

VojtaStavik (changes requested):
The functionality looks correct I just don't think we should expose this on channel controller. Message controller seems like a better option to me.

Sources_v3/StreamChat/Controllers/ChannelController/ChannelController.swift
evsaev: As far as we work with the message that hasn't been sent yet I realized we cannot have a message controller for it. But we return `messageId` from `createNewMessage` so that's possible 👍 

moved to `ChatMessageController` in 32f17a83201c3be8a9f27ced7f0f096cac368f24

VojtaStavik (approved):
Nice 👍 🚢 🇮🇹 once CI 🟢

PR #671, opened 2020-12-17
[CIS-519] Disable all interactions with message textview except links

ManWithBear:
Links were broken because to allow long press for message popup I disabled user interactions on text view.
If we just enable user interactions back, then on long press, user will select text inside text view instead of opening  message popup.

To disable press interception we have next options:
1. Use `UIGestureRecognaizerDelegate.gestureRecognizer(_:shouldRecognizeSimultaneouslyWith:)` and try to find which gesture recognizers in textview are responsible for links actions and for other actions.
This approach is unstable and rely on implementation details of text view
2. Override `hitTest` to skip all touches outside of link bounds.
Nice, simple, and elegant solution.
3. Intercept touches in some middle layer (e.g. view on top of text view) and pass only touches on links. 
Dirty, affect hierarchy, otherwise similar to 2
4. Disable `isSelectable`, add custom gesture recognizer on text view and check if is's inside link bounds.
Missing a lot of system features.

### DIFF
--- a/Sources_v3/StreamChatUI/Common Views/OnlyLinkTappableTextView.swift
+++ b/Sources_v3/StreamChatUI/Common Views/OnlyLinkTappableTextView.swift
@@ -1,0 +1,20 @@
+//
+// Copyright © 2020 Stream.io Inc. All rights reserved.
+//
+
+import UIKit
+
+/// Text View that ignore all user interactions except touches on links
+class OnlyLinkTappableTextView: UITextView {
+    override func hitTest(_ point: CGPoint, with event: UIEvent?) -> UIView? {
+        if let range = characterRange(at: point),
+            !range.isEmpty,
+            let position = closestPosition(to: point, within: range),
+            let styles = textStyling(at: position, in: .forward),
+            styles[.link] != nil {
+            return super.hitTest(point, with: event)
+        } else {
+            return nil
+        }
+    }
+}
--- a/Sources_v3/StreamChatUI/ChatChannel/ChatMessageBubbleView.swift
+++ b/Sources_v3/StreamChatUI/ChatChannel/ChatMessageBubbleView.swift
@@ -22,7 +22,7 @@ open class ChatMessageBubbleView<ExtraData: ExtraDataTypes>: View, UIConfigProvi
         .withoutAutoresizingMaskConstraints
 
     public private(set) lazy var textView: UITextView = {
-        let textView = UITextView()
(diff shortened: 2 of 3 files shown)


Sources_v3/StreamChatUI/Common Views/OnlyLinkTappableTextView.swift
```diff
@@ -0,0 +1,19 @@
...
+class OnlyLinkTappableTextView: UITextView {
+    override func hitTest(_ point: CGPoint, with event: UIEvent?) -> UIView? {
+        guard
+            let range = characterRange(at: point),
+            !range.isEmpty,
+            let position = closestPosition(to: point, within: range),
+            let styles = textStyling(at: position, in: .forward),
+            styles[.link] != nil
+        else { return nil }
+        return self
```
VojtaStavik, line 17: What's wrong with `if { } else { }`? :)

VojtaStavik (approved):
Looks OK, let's see what it breaks :) 

What do you mean exactly by `Missing a lot of system features.` in 4) ?

Sources_v3/StreamChatUI/Common Views/OnlyLinkTappableTextView.swift
VojtaStavik: Also, we should probably call `super` here, not just return `self`

ManWithBear: > What's wrong with if { } else { }? :)
All 3 calls return optional. So I either need to chain it with `Optional.flatMap`, use `if let` or `guard`.
Later is question of preferences.

ManWithBear: > What do you mean exactly by `Missing a lot of system features.` in 4) ?

Drag&drop of links, link preview, probably something else.

ManWithBear:
> What do you mean exactly by Missing a lot of system features. in 4) ?

Drag&drop of links, link preview, probably something else.

Sources_v3/StreamChatUI/Common Views/OnlyLinkTappableTextView.swift
VojtaStavik: I'd argue that it's a question of preference. 
`guard` means - it this condition is not met, the rest of the method doesn't matter
`if {} else {}` means - there are just two options here, either this or that

This case is the latter imo.

PR #672, opened 2020-12-17
Improve thread arrow aligment with message border

ManWithBear:
Designs in Figma are made in "points" resolution -> 1pt border should be 1pt border in our project.

Arrow was drawing outside it borders, which lead to small misalignment with message bubble

Screenshot 2020-12-17 at 17 54 13
Screenshot 2020-12-17 at 17 55 51


VojtaStavik (approved):
Can you please also add a before/after screenshot to the PR description for future reference? Thanks!

PR #674, opened 2020-12-18
[CIS-498] Layout feedback loop in composer

DominikBucher12:
# What is this PR
This PR Fixes layout feedback loop caused by mainly setting `composerView` instead of `inputView` inside `MessageComposerInputAccessoryViewController`. It does 2 things:
- Makes `composerView` subview to inputView, removing the `UIInputView` constraints as subclass. 
- Fixes compression resistance vertical priority for the `messageInputView` so `intrinsicSize` is always known.

# How to test this 
1. Run the application with `-UIViewLayoutFeedbackLoopDebuggingThreshold 100` argument.
2. Try to use the chat as usually.
3. The app should not crash on the argument or the layout shouldn't be somehow broken (except for resizing the textView on deletion the text) 

# Important note
As far as my progress goes, `UIInputViewController` is broken and should not be used so we should fallback to using `inputView`.

This error is thrown in the callstack, while the inputView is first responder and you present something over full screen and then it goes back, somehow the `ViewWillAppear` happens on `UIInputViewController` subclass while the view is actually disappearing this causes this strange behavior which also happens to make composer not showing from time to time and throws assert failure: 

`[Assert] UITextEffectsWindow should not become key. Please file a bug to Keyboard | iOS with this call stack`

In near future I will file a radar, which is closely tight with this one: https://openradar.appspot.com/30064691

### DIFF
--- a/Sources_v3/StreamChatUI/ChatChannel/MessageComposerInputAccessoryViewController.swift
+++ b/Sources_v3/StreamChatUI/ChatChannel/MessageComposerInputAccessoryViewController.swift
@@ -128,7 +128,9 @@ open class MessageComposerInputAccessoryViewController<ExtraData: ExtraDataTypes
     }
     
     func setupInputView() {
-        inputView = composerView
+        inputView?.embed(composerView)
+        inputView?.translatesAutoresizingMaskIntoConstraints = false
+        inputView?.allowsSelfSizing = true
         
         composerView.messageInputView.textView.delegate = self
         
--- a/Sources_v3/StreamChatUI/ChatChannel/ChatChannelMessageComposerView.swift
+++ b/Sources_v3/StreamChatUI/ChatChannel/ChatChannelMessageComposerView.swift
@@ -5,10 +5,7 @@
 import StreamChat
 import UIKit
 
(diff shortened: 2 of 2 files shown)


Sources_v3/StreamChatUI/ChatChannel/MessageComposerInputAccessoryViewController.swift
```diff
@@ -128,7 +128,9 @@ open class MessageComposerInputAccessoryViewController<ExtraData: ExtraDataTypes
     }
     
     func setupInputView() {
-        inputView = composerView
+        inputView?.embed(composerView)
+        inputView?.translatesAutoresizingMaskIntoConstraints = false
+        inputView?.allowsSelfSizing = true
```
VojtaStavik, line 133: Under which circumstances can `inputView` be `nil`? We should check for this and log an error. Otherwise, this will silently fail and we will have no idea.

VojtaStavik (approved):

DominikBucher12:
Closing this as this doesn't yet solve our issue, we will proceed with deleting `UIInputViewController` subclass as it doesn't fit our needs.

PR #675, opened 2020-12-18
[CIS-350][CIS-427][CIS-359] Fix failing LLC tests

VojtaStavik:
This PR fixes many different problems with the stability of our SDK:

- CoreData was complaining about us using an incorrect initializer for `AttachmentDTO`. See the commit description for more details.

- I fixed the incorrect usage of `@Atomic` with collections (CIS-427)

- Fixed a race condition of removing notifications in `deinit`

- Fixed DB observer's `startObserving` method not being thread safe

This should also fix out test suite, so I re-enabled it. Let's see :)

I put the new findings to our [Writing stable unit tests](https://github.com/GetStream/stream-chat-swift/wiki/Writing-stable-unit-tests-checklist) checklist.

---

This PR also closes CIS-427 and CIS-359.

PR #677, opened 2020-12-18
Fix initialization of DTOs

ManWithBear:
Catch up after #675


VojtaStavik (approved):
Great catch, thank you!

PR #681, opened 2020-12-21
[CIS-515] Thread replies checkmark

mantergo:
**In this PR:** 

- Introduce `MessageComposerCheckmarkControl`
- Introduce `MessageComposerCheckmarkView`
- Add `MessageComposerCheckmarkView` to `ChatChannelMessageComposerView`
- Implement thread reply logic

cc: @ManWithBear (you were asking about this one)

ezgif com-video-to-gif-2


Sources_v3/StreamChatUI/ChatChannel/MessageComposerCheckmarkControl.swift
```diff
@@ -0,0 +1,61 @@
...
+    }
+    
+    override public func defaultAppearance() {
+        imageView.layer.cornerRadius = 4
+        imageView.layer.borderWidth = 2
+        imageView.layer.masksToBounds = true
+    }
+    
+    override open func setUpLayout() {
+        let inset = (intrinsicContentSize.height - checkmarkHeight) / 2
```
VojtaStavik, line 44: Do we need this? Isn't placing the image view to the center enough?

Sources_v3/StreamChatUI/ChatChannel/MessageComposerInputAccessoryViewController.swift
```diff
@@ -32,6 +32,20 @@ open class MessageComposerInputAccessoryViewController<ExtraData: ExtraDataTypes
...
+        didSet {
+            guard let memberCount = controller?.channel?.memberCount else { return }
+            
+            composerView.setCheckmarkView(hidden: false)
+            
+            if memberCount > 2 {
+                composerView.checkmarkView.label.text = L10n.Composer.Checkmark.channelReply
+            } else {
+                composerView.checkmarkView.label.text = L10n.Composer.Checkmark.directMessageReply
+            }
```
VojtaStavik, line 45: Shouldn't this happen in `updateContent` and we should rather call `updateContentIfNeeded` here?

VojtaStavik (changes requested):
I think it looks good but I'm afraid the expectation is that the text label can also be used to toggle the switch. Check out the Slack iOS app as a reference.

Sources_v3/StreamChatUI/ChatChannel/ChatChannelMessageComposerView.swift
```diff
@@ -190,4 +198,8 @@ open class ChatChannelMessageComposerView<ExtraData: ExtraDataTypes>: UIInputVie
     }
     
     open func updateContent() {}
+    
+    open func setCheckmarkView(hidden: Bool) {
+        container.bottomStackView.isHidden = hidden
```
DominikBucher12, line 203: 👍🏼 nice

Sources_v3/StreamChatUI/ChatChannel/MessageComposerInputAccessoryViewController.swift
```diff
@@ -170,40 +184,52 @@ open class MessageComposerInputAccessoryViewController<ExtraData: ExtraDataTypes
...
-                cid: cid,
-                messageId: messageToReply.id
-            )
+            createNewMessage(text: text)
+        case .reply:
             // TODO:
             // 1. Attachments
             // 2. Should be inline reply after backend implementation.
-            messageController.createNewReply(text: textView.text, showReplyInChannel: true)
+            print("Inline reply sent.")
```
DominikBucher12, line 199: Didn't you forgot to delete this `print`? :)

Sources_v3/StreamChatUI/ChatChannel/MessageComposerInputAccessoryViewController.swift
```diff
@@ -170,40 +184,52 @@ open class MessageComposerInputAccessoryViewController<ExtraData: ExtraDataTypes
...
         case let .slashCommand(command):
-            controller.createNewMessage(text: "/\(command.name) " + text)
+            createNewMessage(text: "/\(command.name) " + text)
         }
         
         state = .initial
     }
     
+    func createNewMessage(text: String) {
+        guard let cid = controller?.cid else { return }
```
DominikBucher12, line 216: I see lot of those guards through the diff. Shouldn't we log somehow the issue when the values are nil when they shouldn't be? I mean this should never occur, so maybe use implicitly unwrapped optional in the declaration? It's a bit harder to debug while app doesn't at least `assert`-crash. What do you think, @VojtaStavik ?

Sources_v3/StreamChatUI/UIConfig.swift
```diff
@@ -72,6 +72,8 @@ public extension UIConfig {
         public var messageComposerButton: UIColor = .streamGray
         public var messageComposerStateIcon: UIColor = .streamGrayGainsboro
         public var messageComposerPlaceholder: UIColor = .streamGray
+        public var messageComposerCheckmarkBorder: UIColor = .streamGray
+        public var messageComposerCheckmark: UIColor = .white
```
ManWithBear, line 77: `.streamWhite`

Sources_v3/StreamChatUI/ChatChannel/MessageComposerInputAccessoryViewController.swift
mantergo: It was implicitly unwrapped before but there were some issue that it was accessed before setting. 
I can't remember what was that.

Sources_v3/StreamChatUI/ChatChannel/MessageComposerInputAccessoryViewController.swift
mantergo: Yeah, it's in `switch` so I can't live it empty.

Sources_v3/StreamChatUI/ChatChannel/ChatChannelMessageComposerView.swift
mantergo: Good catch! Tbh that's strange that we haven't faced any issues with that so far  🤔🤔

Sources_v3/StreamChatUI/ChatChannel/MessageComposerInputAccessoryViewController.swift
DominikBucher12: ahh okay so lets leave it like this :)

Sources_v3/StreamChatUI/ChatChannel/ChatChannelMessageComposerView.swift
DominikBucher12: I guess yeah, just flashback :) So its okay then :)

Sources_v3/StreamChatUI/ChatChannel/MessageComposerInputAccessoryViewController.swift
VojtaStavik: @DominikBucher12 has a good point here. Every silent failure like this is a potential hard-to-debug issue. Same below on L234 and L229. Someone will call `createNewMessage`, nothing will happen, and we will don't know there's a problem.

Let's add this to our guidelines - to avoid silent failures.

Sources_v3/StreamChatUI/Generated/L10n.swift
```diff
@@ -1,3 +1,7 @@
+//
+// Copyright © 2020 Stream.io Inc. All rights reserved.
+//
+
```
VojtaStavik, line 4: I know it sucks this file is generated and we still run formatter on it, but can you please commit only the changes that are relevant. Hopefully we will solve this soon. Thanks

VojtaStavik (approved):
Looks good 👍

Sources_v3/StreamChatUI/ChatChannel/MessageComposerInputAccessoryViewController.swift
mantergo: Do we have some pattern for this cases already?

DominikBucher12 (approved):

PR #688, opened 2020-12-23
[CIS-541] Make `StreamChat` usable from app extensions

VojtaStavik:
This PR is a part of the changes needed to enable notification extensions for `StreamChat`. The extension is needed to support our new backend service for push notifications.

### Changes
- `StreamChat` is marked as `APPLICATION_EXTENSION_API_ONLY` which means you can use it safely from app extensions.
- Accessing `UIApplication.shared` directly is no longer possible so I put it behind a check.
 
*However,* it can lead to issues if we run the full client from extensions. It would create a new web socket connection, mark the user as `online`, start background workers, etc. For these reasons, I made it not possible to actually create a client instance from extensions yet. Our notification extension will expose only safe API, not the full client, for the reasons above.

In general, we should make it possible to create a `ChatClient` instance in extensions eventually. However, this task will not be trivial.

### DIFF
--- a/Sources_v3/StreamChat/ChatClient.swift
+++ b/Sources_v3/StreamChat/ChatClient.swift
@@ -215,6 +215,7 @@ public class _ChatClient<ExtraData: ExtraDataTypes> {
     ///
     /// - Parameter config: The config object for the `Client`. See `ChatClientConfig` for all configuration options.
     ///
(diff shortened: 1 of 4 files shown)


b-onc (approved):
🚢 🇮🇹

PR #689, opened 2020-12-23
Make file preview layout less dynamic, to prevent layout loop

ManWithBear:
Looks like our layout was too dynamic and layout engine goes mad with it. Specifically:
- `ChatFileAttachmentListView.ItemView` 2 components that limit min height, positioned in the middle, nothing to limit max size
- `ChatMessageContentView.messageReactionsView` in general bad constraints. Will be reworked with reactions tail update

### DIFF
--- a/Sources_v3/StreamChatUI/ChatChannel/ChatMessage/ChatMessageContentView.swift
+++ b/Sources_v3/StreamChatUI/ChatChannel/ChatMessage/ChatMessageContentView.swift
@@ -205,9 +205,9 @@ open class ChatMessageContentView<ExtraData: ExtraDataTypes>: View, UIConfigProv
         }
 
         if message?.deletedAt == nil, !(message?.reactionScores.isEmpty ?? true) {
-            toActivate.append(bubbleToReactionsConstraint!)
+//            toActivate.append(bubbleToReactionsConstraint!)
         } else {
-            toDeactivate.append(bubbleToReactionsConstraint!)
+//            toDeactivate.append(bubbleToReactionsConstraint!)
         }
         
         if message?.isLastInGroup == true {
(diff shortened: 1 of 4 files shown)


Sources_v3/StreamChatUI/ChatChannel/ChatMessage/ChatMessageContentView.swift
```diff
@@ -205,9 +205,9 @@ open class ChatMessageContentView<ExtraData: ExtraDataTypes>: View, UIConfigProv
         }
 
         if message?.deletedAt == nil, !(message?.reactionScores.isEmpty ?? true) {
-            toActivate.append(bubbleToReactionsConstraint!)
+//            toActivate.append(bubbleToReactionsConstraint!)
         } else {
-            toDeactivate.append(bubbleToReactionsConstraint!)
+//            toDeactivate.append(bubbleToReactionsConstraint!)
```
ManWithBear, line 210: Reaction view will be back when reaction tail is finished

StreamChat_v3.xcodeproj/xcshareddata/xcschemes/DemoApp.xcscheme
```diff
@@ -50,6 +50,12 @@
             ReferencedContainer = "container:StreamChat_v3.xcodeproj">
          </BuildableReference>
       </BuildableProductRunnable>
+      <CommandLineArguments>
+         <CommandLineArgument
+            argument = "-UIViewLayoutFeedbackLoopDebuggingThreshold 100"
```
ManWithBear, line 55: I think we should keep this flag for future loop cases

VojtaStavik: 👍

VojtaStavik (approved):
Thanks for fixing this @ManWithBear 👍

PR #691, opened 2020-12-23
[CIS-445] Message reactions follow-up

evsaev:
**This PR** updates message reactions UI in **message list** and **message actions** screens to meet the design.

Alternative considered approaches for tail implementations:

1. Exporting message bubble icon from `Figma`, stretching the middle of the image while keeping the tail as it is and using as a **backgroundView** OR **mask**.

Doesn't work since the tail position is dynamic + at this point bubbles are build from multiple layers that can't be exported as a single image.

2. Have custom and drawing the tail in `drawRect`

This approach is similar to @ManWithBear implementation but instead of 2 image-views we have 2 custom views. The downside comparing to `image` approach is that the tail view doesn't have any content and preserving the width/height requires setting a width + height constraint explicitly which is not nice (providing `intrinsicContentSize` works only in combination with `setCompressionResitancePriority` in tail's superview)

Finally stick to current approach (generating the image) but incapsulated this logic in reactions bubble view and made it possible to provide different bubble implementation via `UIConfig`.

https://user-images.githubusercontent.com/12818985/103010016-a8fc4900-4548-11eb-8ca4-0a49ed0797a0.mov


Sources_v3/StreamChatUI/ChatChannel/Reactions/ChatMessageReactionsBubbleView.swift
```diff
@@ -0,0 +1,64 @@
...
+
+    public private(set) lazy var contentView = uiConfig
+        .messageList
+        .messageReactions
+        .reactionsView
+        .init()
+        .withoutAutoresizingMaskConstraints
+
+    open var tailLeadingAnchor: NSLayoutXAxisAnchor { contentView.centerXAnchor }
+    open var tailTrailingAnchor: NSLayoutXAxisAnchor { contentView.centerXAnchor }
```
evsaev, line 23: Don't like this part since `ChatMessageReactionsBubbleView` doesn't have a tail but knows about it. Those anchors are required to make the tail position dynamic.

Sources_v3/StreamChatUI/ChatChannel/Reactions/ChatMessageReactionsView.swift
```diff
@@ -0,0 +1,66 @@
+//
+// Copyright © 2020 Stream.io Inc. All rights reserved.
+//
+
+import StreamChat
+import UIKit
+
+open class ChatMessageReactionsView<ExtraData: ExtraDataTypes>: View, UIConfigProvider {
```
ManWithBear, line 8: Loud thoughts:

Aint it too much view classes for just small reaction bubble?
At this point I found myself on thought: "Wait, I have seen this one, haven't i?"

evsaev: Actually, we have the same count + 2 for bubbles (base and default one). It can feel we've got too many because previously all (an a VC) were located in the single file.

Sources_v3/StreamChatUI/ChatChannel/ChatMessage/ChatMessageContentView.swift
```diff
@@ -204,19 +238,19 @@ open class ChatMessageContentView<ExtraData: ExtraDataTypes>: View, UIConfigProv
             toDeactivate.append(contentsOf: outgoingMessageConstraints)
         }
 
-        if message?.deletedAt == nil, !(message?.reactionScores.isEmpty ?? true) {
-//            toActivate.append(bubbleToReactionsConstraint!)
+        if message.deletedAt == nil, !message.reactionScores.isEmpty {
+            toActivate.append(bubbleToReactionsConstraint!)
         } else {
 //            toDeactivate.append(bubbleToReactionsConstraint!)
```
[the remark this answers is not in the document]
evsaev: Good catch, thanks 👍 Fixed in 267ce397d591c538a2eb65b4016d1b03a78bbcba

Sources_v3/StreamChatUI/ChatChannel/Reactions/ChatMessageReactionsBubbleTail.swift
```diff
@@ -0,0 +1,112 @@
...
+                flipped: flipped
+            )
+        }
+
+        static func large(flipped: Bool) -> Self {
+            return .init(
+                smallRadius: 3,
+                bigRadius: 6,
+                border: 0,
+                outline: 1,
```
ManWithBear, line 30: Large tail does not have outline

Sources_v3/StreamChatUI/ChatChannel/Reactions/ChatMessageReactionsBubbleTail.swift
```diff
@@ -0,0 +1,112 @@
...
+        bigBorderFrame.insetBy(dx: -outline, dy: -outline)
+    }
+
+    var bigBorderFrame: CGRect {
+        bigFrame.insetBy(dx: -border, dy: -border)
+    }
+
+    var bigFrame: CGRect {
+        .init(
+            x: flipped ? imageSize.width - outline - border - 2*bigRadius : outline + border,
```
ManWithBear, line 89: nit:
I find ternary operator with not primitive variants (not `q ? x : y`) more readable when multilined.
```
x: flipped 
    ? imageSize.width - outline - border - 2 * bigRadius 
    : outline + border,
```

Sources_v3/StreamChatUI/ChatChannel/Reactions/ChatMessageReactionVC.swift
```diff
@@ -0,0 +1,73 @@
...
+
+        messageController.setDelegate(self)
+    }
+
+    override open func setUpLayout() {
+        view.embed(reactionsBubble)
+    }
+
+    override open func updateContent() {
+        reactionsBubble.content = messageController.message.flatMap { message in
```
ManWithBear, line 33: Why `flatMap` and not just `map`? I don't see any optionality in closure returning value

Sources_v3/StreamChatUI/ChatChannel/ChatMessage/ChatMessageContentView.swift
```diff
@@ -104,9 +111,9 @@ open class ChatMessageContentView<ExtraData: ExtraDataTypes>: View, UIConfigProv
             authorAvatarView.leadingAnchor.constraint(equalTo: leadingAnchor),
             authorAvatarView.bottomAnchor.constraint(equalTo: bottomAnchor),
             
-            messageReactionsView.topAnchor.constraint(equalTo: topAnchor),
+            reactionsBubble.topAnchor.constraint(equalTo: topAnchor),
             
-            messageBubbleView.trailingAnchor.constraint(equalTo: trailingAnchor).with(priority: .defaultHigh),
+            messageBubbleView.trailingAnchor.constraint(equalTo: trailingAnchor).with(priority: .required - 1),
```
ManWithBear, line 116: I found myself often in situation when I need `priority: 999`. Almost all our `with(Priority:)` needs to be 999
Maybe it's time to make:
```
extension UILayoutPriority {
    static var almostGuaranteed: UILayoutPriority { .init(rawValue: 999) }
}
```
Other possible names:
- `a999`: Will be used as `.with(priority: .a999)`. We all know what 999 means, so it's became domain knowledge. Leading `a` is required, because identifier cannot start with digit.
- `softRequired`
- your ideas?

Sources_v3/StreamChatUI/ChatChannel/Reactions/ChatMessageReactionsBubbleTail.swift
evsaev: The outline is needed here to share the same drawing logic as for small tails.  It's just it has `.clear` color and visually looks like a distance in-between small and big bubbles.

Sources_v3/StreamChatUI/ChatChannel/Reactions/ChatMessageReactionVC.swift
evsaev: **ChatMessageController** has `message?` that why the `flatMap` is here.

Sources_v3/StreamChatUI/ChatChannel/ChatMessage/ChatMessageContentView.swift
evsaev: Great idea, let's do that!

I'd rather we preserve the `required` terminology: smth like `almostRequired`. It also would be nice to have a helper on `NSLayoutContraint`:

```swift
extension UILayoutPriority {
    static var almostRequired: UILayoutPriority { .init(rawValue: 999) }
}

extension NSLayoutContraint {
    var almostRequired: NSLayoutContraint {
      self.with(priority: .almostRequired)
    }
}
```

Sources_v3/StreamChatUI/ChatChannel/Reactions/ChatMessageReactionsBubbleTail.swift
evsaev: fixed in 2e31bb0df942ff1a68e9603cc90d182880bb34c7

Sources_v3/StreamChatUI/ChatChannel/Reactions/ChatMessageReactionsView.swift
```diff
@@ -0,0 +1,66 @@
...
+        }
+
+        guard let content = content else { return }
+
+        content.reactions.forEach { reaction in
+            let itemView = uiConfig.messageList.messageReactions.reactionItemView.init()
+            itemView.content = .init(
+                useBigIcon: content.useBigIcons,
+                reaction: reaction,
+                onTap: { [weak self] in self?.content?.didTapOnReaction($0) }
```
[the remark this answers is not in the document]
evsaev: changed in ff1109b411130ff04ad42b4564aa52d7a1b5d737

ManWithBear:
Previously we had short conversation with @VojtaStavik about performance options on displaying image vs drawing.

In [Image and Graphics Best Practices WWDC2018](https://developer.apple.com/wwdc18/219) Apple states that drawing `UIImage` via `UIImageView` basically most performance efficient way to display something on screen (with couple points on how to efficiently load images).
And in most cases if you can, you should use `UIImageView` with either prepared images or images that were drawn in runtime once (via `UIGraphicsImageRenderer`) instead of drawing via other techniques.

Sources_v3/StreamChatUI/Utils/NSLayoutConstraint+Extensions.swift
```diff
@@ -4,9 +4,17 @@
...
+
 extension NSLayoutConstraint {
     func with(priority: UILayoutPriority) -> NSLayoutConstraint {
         self.priority = priority
         return self
     }
+
+    var almostRequired: NSLayoutConstraint {
+        with(priority: .almostRequired)
+    }
```
VojtaStavik, line 19: nit: when introducing helpers like this, please always add a comment explaining what's happening and how/where to use ti.

Sources_v3/StreamChatUI/ChatChannel/Reactions/ChatMessageReactionVC.swift
```diff
@@ -0,0 +1,73 @@
+//
+// Copyright © 2020 Stream.io Inc. All rights reserved.
+//
+
+import StreamChat
+import UIKit
+
+open class ChatMessageReactionVC<ExtraData: ExtraDataTypes>: ViewController, UIConfigProvider {
```
VojtaStavik, line 8: We use this just for the "big" UI case, right?

Sources_v3/StreamChatUI/UIConfig.swift
```diff
@@ -157,20 +158,56 @@ public extension UIConfig {
...
+
         public var availableReactions: [MessageReactionType] = [
             .init(rawValue: "like"),
             .init(rawValue: "haha"),
-            .init(rawValue: "facepalm"),
-            .init(rawValue: "roar")
+            .init(rawValue: "thumbsup"),
+            .init(rawValue: "thumbsdown"),
+            .init(rawValue: "wut")
         ]
```
VojtaStavik, line 184: Should we unify the way we specify reaction somewhere? Maybe even using a protocol such that users can add their own custom data?
```swift
protocol ReactionType { 
  let stringValue: String
  let smallIcon: UIImage
  let largeIcon: UIImage
}

UIConfig.default.availableReactions = [Reaction1, Reaction2, ...]

```

VojtaStavik (approved):
I still have feeling that the bubble logic can be simplified but it's possible that I just don't know enough about the problem. Let's move on and deal with it next month. Thanks @evsaev 👍

Sources_v3/StreamChatUI/ChatChannel/Reactions/ChatMessageReactionVC.swift
evsaev: Yep 👍 The controller comes into the game only when user activates actions on a message (the reactions view is styled as `big`)

Sources_v3/StreamChatUI/Utils/NSLayoutConstraint+Extensions.swift
evsaev: added in 3ab33bb2ab5e73b90518d92f655579839d03eabe

Sources_v3/StreamChatUI/UIConfig.swift
evsaev: Introduced `ReactionAppearanceType ` in 7d0da636ec9cdbc162fb1873d5428ee2a1d7e844:

```swift
/// The type describing message reaction appearance.
public protocol ReactionAppearanceType {
    var smallIcon: UIImage { get }
    var largeIcon: UIImage { get }
}
```

and updated `UIConfig` with `[MessageReactionType: ReactionAppearanceType]` in 7ab5481886074940bf325dd8b17328cc0157781d.

Used dictionary with reaction type as a key instead of array to make sure we don't have duplicates and avoid dealing with `Hashable`

PR #693, opened 2020-12-23
Fix UserListUpdater flaky test

b-onc:
`test_updateCompletion_calledAfterDBWriteCompletes` was flaky since listUpdater.update makes a database.write call, which calls writableContext.perform , and in the same perform block accessing viewContext crashes. but using same context (writableContext ) works flawless. So we need to call `viewContext` in main queue


VojtaStavik (approved):
Nice 👏 👏 👏

PR #698, opened 2020-12-28
Fix flaky userlistupdater test

b-onc:
This test was failing because `newUser { database.viewContext.user(...) }` was accessed inside a `database.writeSync` closure. `viewContext` should always be accessed in main queue.


VojtaStavik (changes requested):
I don't think the fix is the best. We should always check the view context to make sure the changes are propagated. So, a more _correct_ fix would be to just do the assertions on the view context after `writeSynchronously` finishes.

VojtaStavik (approved):

PR #701, opened 2020-12-28
V2: prevent crash on inconsistent row deletion in ChatViewController

cardoso:
Added the same fallback that is on row insertion


b-onc (approved):
CHANGELOG then LGTM

PR #702, opened 2020-12-28
[CIS-540] Save channel data from message payload

VojtaStavik:
Firstly - the way we decoded message endpoint responses were always wrong and never worked 🤷‍♂️ :))

Secondly - We found out that our backed _sometimes_ includes channel payload directly into message endpoint response. This is basically what solves issue described in CIS-320. This PR makes it possible to parse and save channel-related data to DB, too.

This is a preparation PR for adding push notification support to LLC. Thanks to this, we will have to make just one API call, instead of two.

---

This PR also closes CIS-320.


Sources_v3/StreamChat/APIClient/Endpoints/Payloads/MessagePayloads.swift
```diff
@@ -25,12 +25,16 @@ enum MessagePayloadsCodingKeys: String, CodingKey {
     case ownReactions = "own_reactions"
     case reactionScores = "reaction_scores"
     case isSilent = "silent"
+    case channel
     //        case i18n
 }
 
-/// An object describing the wrapped incoming message JSON payload.
-struct WrappedMessagePayload<ExtraData: ExtraDataTypes>: Decodable {
-    let message: MessagePayload<ExtraData>
```
b-onc, line 33: I liked `WrappedMessage` more than `MessagePayload.Box` :( Any particular reason to change it?

b-onc (approved):
LGTM

Sources_v3/StreamChat/APIClient/Endpoints/Payloads/MessagePayloads.swift
VojtaStavik: @b-onc I didn't like the name collision with optional wrapped type + I kinda like having this a nested type. Happy to hear your suggestions regarding the name.

ManWithBear: I'm a `Box` guy

Sources_v3/StreamChat/APIClient/Endpoints/Payloads/MessagePayloads.swift
```diff
@@ -59,6 +60,10 @@ struct MessagePayload<ExtraData: ExtraDataTypes>: Decodable {
     let reactionScores: [MessageReactionType: Int]
     let attachments: [AttachmentPayload<ExtraData.Attachment>]
     let isSilent: Bool
+
+    /// Only message payload from `getMessage` endpoint contains channel data. It's a convenience workaround for having to
+    /// mane an extra call do get channel details.
```
ManWithBear, line 65: nit: `make`?

ManWithBear (approved):

Sources_v3/StreamChat/APIClient/Endpoints/Payloads/MessagePayloads.swift
ManWithBear: On other hand from usage point of view I would expect `ed` suffix:
```
let endpoint: Endpoint<MessagePayload<DefaultExtraData>.Boxed>
let endpoint: Endpoint<MessagePayload<DefaultExtraData>.Wrapped>
```

Sources_v3/StreamChat/APIClient/Endpoints/Payloads/MessagePayloads.swift
VojtaStavik: fixed in f639bfb

Sources_v3/StreamChat/APIClient/Endpoints/Payloads/MessagePayloads.swift
VojtaStavik: I like the `Boxed` idea. Fixed in 2da2168

PR #703, opened 2020-12-29
[CIS-550] MessageComposer basic animation

mantergo:
**In this PR:**

- Update `textColor` with `ColorPalette`
- Fix sendting attachments without text
- Remove left and right extra insets for `MessageComposerView`
- Add basic animation to `MessageComposer`

_Note: haven't succeeded in animating intrinsicContentSize change on textView height expand._

https://user-images.githubusercontent.com/10098359/103308065-d9912680-4a11-11eb-9fe6-c75591f47073.mp4

### DIFF
--- a/Sources_v3/StreamChatUI/Utils/Animation.swift
+++ b/Sources_v3/StreamChatUI/Utils/Animation.swift
@@ -1,0 +1,23 @@
+//
+// Copyright © 2020 Stream.io Inc. All rights reserved.
+//
+
+import UIKit
+
+func Animate(_ actions: @escaping () -> Void, completion: ((Bool) -> Void)? = nil) {
+    guard !UIAccessibility.isReduceMotionEnabled else {
+        actions()
+        completion?(true)
+        return
+    }
+    
+    UIView.animate(
+        withDuration: 0.25,
+        delay: 0,
+        usingSpringWithDamping: 0.8,
+        initialSpringVelocity: 4,
+        options: .curveEaseInOut,
+        animations: actions,
+        completion: completion
+    )
+}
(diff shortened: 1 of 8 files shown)


VojtaStavik (changes requested):
Great start. I think it'd be nice if we had just once place where to perform animations so we can control is centrally.

Maybe something like: 
```swift
func WithAnimation(_ actions: () -> Void) {...}
```
Inside this function, we should also check for `UIAccessibilityIsReduceMotionEnabled` value and don't perform animations if it's set.

Finally, let's try to use a spring animation as default. The UI will feel much more snappier.

mantergo:
> Great start. I think it'd be nice if we had just once place where to perform animations so we can control is centrally.
> 
> Maybe something like:
> 
> ```swift
> func WithAnimation(_ actions: () -> Void) {...}
> ```
> 
> We'd use it as replacement for `UIView.animate { }`. Inside this function, we should also check for `UIAccessibilityIsReduceMotionEnabled` value and don't perform animations if it's set.
> 
> Finally, let's try to use a spring animation as default. The UI will feel much more snappier.

Yeah, good points! Thanks!

Sources_v3/StreamChatUI/Utils/Animation.swift
```diff
@@ -0,0 +1,22 @@
+//
+// Copyright © 2020 Stream.io Inc. All rights reserved.
+//
+
+import UIKit
+
+func WithAnimation(_ actions: @escaping () -> Void, completion: ((Bool) -> Void)? = nil) {
```
VojtaStavik, line 7: nit: Are you happy with the name? It seems I don't actually like it 🤦‍♂️ Maybe `Animate { }` would read better on the caller side?

Sources_v3/StreamChatUI/Utils/Animation.swift
```diff
@@ -0,0 +1,22 @@
...
+            usingSpringWithDamping: 0.5,
+            initialSpringVelocity: 4,
+            options: .curveEaseInOut,
+            animations: actions,
+            completion: completion
+        )
+    } else {
+        actions()
+        completion?(true)
+    }
```
VojtaStavik, line 21: nit: `guard` would work much better here

VojtaStavik (approved):

PR #705, opened 2020-12-30
Fix empty/incorrect Channel names on ChannelList

b-onc:
We didn't implement this logic and sample data had lots of empty names, looking bad on UI
A picture tells a thousand words: before and after
Screen Shot 2020-12-30 at 15 09 45

### DIFF
--- a/Sources_v3/StreamChatUI/ChatChannel/ChatChannelNavigationBarListener.swift
+++ b/Sources_v3/StreamChatUI/ChatChannel/ChatChannelNavigationBarListener.swift
@@ -11,17 +11,24 @@ class ChatChannelNavigationBarListener<ExtraData: ExtraDataTypes> {
     typealias NavbarData = (title: String?, subtitle: String?)
 
     let channelController: _ChatChannelController<ExtraData>
+    let namer: ChannelNamer
     var onDataChange: (NavbarData) -> Void = { _ in }
 
-    static func make(for channel: ChannelId, in client: _ChatClient<ExtraData>) -> ChatChannelNavigationBarListener {
+    static func make(
+        for channel: ChannelId,
+        in client: _ChatClient<ExtraData>,
+        using namer: ChannelNamer
+    ) -> ChatChannelNavigationBarListener {
         /// if we in channel room, channel will be here, but it always safe to fallback to group chat
         let isDirect = client.channelController(for: channel).channel?.isDirectMessageChannel ?? false
         return isDirect
-            ? DirectChatChannelNavigationBarListener(client: client, channel: channel)
-            : GroupChatChannelNavigationBarListener(client: client, channel: channel)
+            ? DirectChatChannelNavigationBarListener(client: client, channel: channel, namer: namer)
+            : GroupChatChannelNavigationBarListener(client: client, channel: channel, namer: namer)
     }
 
-    fileprivate init(client: _ChatClient<ExtraData>, channel: ChannelId) {
+    fileprivate init(client: _ChatClient<ExtraData>, channel: ChannelId, namer: ChannelNamer) {
+        self.namer = namer
+        
         channelController = client.channelController(for: channel)
         channelController.setDelegate(self)
 
@@ -50,7 +57,7 @@ extension ChatChannelNavigationBarListener: _ChatChannelControllerDelegate {
 private class GroupChatChannelNavigationBarListener<ExtraData: ExtraDataTypes>: ChatChannelNavigationBarListener<ExtraData> {
     override func makeNavbarData() -> NavbarData {
         guard let channel = channelController.channel else { return (nil, nil) }
-        let title = channel.name
+        let title = namer.name(for: channel, as: channelController.client.currentUserId)
         let subtitle = "\(channel.memberCount) members, \(channel.watcherCount) online"
         return (title, subtitle)
     }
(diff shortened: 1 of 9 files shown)


Sources_v3/StreamChatUI/Utils/ChannelNamer.swift
```diff
@@ -0,0 +1,36 @@
...
+import StreamChat
+
+open class ChannelNamer {
+    public required init() {}
+    
+    open func name<ExtraData: ExtraDataTypes>(for channel: _ChatChannel<ExtraData>, as currentUserId: UserId?) -> String {
+        if let channelName = channel.name, !channelName.isEmpty {
+            // If there's an assigned name and it's not empty, we use it
+            return channelName
+        } else if channel.cid.id.hasPrefix("!members") {
```
b-onc, line 15: There's `isDirectMessageChannel` flag on channel but it checks `member.count == 2`, this check is not required by backend, if we only pass user id's and not cid during channel creation, backend will generate a cid for us, and we assume this channel is "direct message", regardless of participant count

Sources_v3/StreamChatUI/Utils/ChannelNamer.swift
```diff
@@ -0,0 +1,36 @@
+//
+// Copyright © 2020 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+import StreamChat
+
+open class ChannelNamer {
```
ManWithBear, line 8: Why it needs to be class and not just function / closure?

b-onc: it's in-line with what we have in UIConfig (consistency)

Sources_v3/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
```diff
@@ -139,8 +139,12 @@ open class ChatChannelListItemView<ExtraData: ExtraDataTypes>: View, UIConfigPro
     
     override open func updateContent() {
         // Title
-        
-        titleLabel.text = channelAndUserId.channel?.name
+        if let channel = channelAndUserId.channel {
+            let namer = uiConfig.channelList.channelNamer.init()
+            titleLabel.text = namer.name(for: channel, as: channelAndUserId.currentUserId)
+        } else {
+            titleLabel.text = "NoChannel"
```
evsaev, line 146: nit: would be nice to have it localized

Sources_v3/StreamChatUI/Utils/ChannelNamer.swift
```diff
@@ -0,0 +1,36 @@
...
+        if let channelName = channel.name, !channelName.isEmpty {
+            // If there's an assigned name and it's not empty, we use it
+            return channelName
+        } else if channel.cid.id.hasPrefix("!members") {
+            // If this is a channel generated as DM
+            // we generate the name from users
+            var members = channel.cachedMembers.filter { $0.id != currentUserId }.compactMap(\.name)
+            let channelName: String
+            if members.isEmpty {
+                channelName = "OnlyCurrentUser"
```
evsaev, line 21: nit: localization

Sources_v3/StreamChatUI/ChatChannel/ChatChannelNavigationBarListener.swift
```diff
@@ -11,17 +11,24 @@ class ChatChannelNavigationBarListener<ExtraData: ExtraDataTypes> {
...
         let isDirect = client.channelController(for: channel).channel?.isDirectMessageChannel ?? false
         return isDirect
-            ? DirectChatChannelNavigationBarListener(client: client, channel: channel)
-            : GroupChatChannelNavigationBarListener(client: client, channel: channel)
+            ? DirectChatChannelNavigationBarListener(client: client, channel: channel, namer: namer)
+            : GroupChatChannelNavigationBarListener(client: client, channel: channel, namer: namer)
     }
 
-    fileprivate init(client: _ChatClient<ExtraData>, channel: ChannelId) {
+    fileprivate init(client: _ChatClient<ExtraData>, channel: ChannelId, namer: ChannelNamer) {
```
DominikBucher12, line 29: why not just `private`? IMHO `fileprivate` is almost unnecessary access control modifier. I don't see doing anything in any extension either. I would personally add some rule for not using `fileprivate`...

ManWithBear: Because subclasses needs to access this `init`

Sources_v3/StreamChatUI/Utils/ChannelNamer.swift
VojtaStavik: @b-onc I think we should fix `isDirectMessageChannel` to check only for the prefix. The `member.count == 2` is not necessary imo. Clients can easily do it if they need to.

StreamChat_v3.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -18,15 +18,6 @@
...
         }
-      },
-      {
-        "package": "SwiftyGif",
-        "repositoryURL": "https://github.com/kirualex/SwiftyGif.git",
-        "state": {
-          "branch": null,
-          "revision": "420d12cba8338b03993f73a72c1f6b70d5273cf2",
-          "version": "5.3.0"
-        }
```
ManWithBear, line 29: I think something went wrong on rebase / merge

ManWithBear (changes requested):

ManWithBear (approved):

PR #706, opened 2020-12-30
[CIS-524] Message actions UX

evsaev:
**This PR** improves message actions pop-up UX:
- fixes jumps of partially visible messages
- updates layout to respect safe-area insets
- autoscrolls to make all content + actions visible

https://user-images.githubusercontent.com/12818985/103361309-394b0880-4aca-11eb-9553-e8cc1209cb02.mov

### DIFF
--- a/Sources_v3/StreamChatUI/MessageActionsPopup/ChatMessagePopupViewController.swift
+++ b/Sources_v3/StreamChatUI/MessageActionsPopup/ChatMessagePopupViewController.swift
@@ -10,6 +10,8 @@ open class ChatMessagePopupViewController<ExtraData: ExtraDataTypes>: ViewContro
         .withoutAutoresizingMaskConstraints
     public private(set) lazy var scrollContentView = UIView()
         .withoutAutoresizingMaskConstraints
+    public private(set) lazy var contentView = UIView()
+        .withoutAutoresizingMaskConstraints
     public private(set) lazy var blurView = UIVisualEffectView(effect: UIBlurEffect(style: .light))
         .withoutAutoresizingMaskConstraints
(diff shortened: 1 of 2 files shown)


ManWithBear (approved):
Honestly I do not understand half of this constraint manipulations and math.
But result looks awesome 👍🏾

Sources_v3/StreamChatUI/MessageActionsPopup/ChatMessagePopupViewController.swift
```diff
@@ -97,6 +129,20 @@ open class ChatMessagePopupViewController<ExtraData: ExtraDataTypes>: ViewContro
...
+        let contentOffset = CGPoint(x: 0, y: max(0, -messageViewFrame.minY + spacing + reactionsViewHeight))
+        scrollView.setContentOffset(contentOffset, animated: false)
+    }
+
+    override open func viewDidAppear(_ animated: Bool) {
+        super.viewDidAppear(animated)
+
+        let contentRect = scrollContentView.convert(contentView.frame, to: scrollView)
+        scrollView.scrollRectToVisible(contentRect, animated: true)
+    }
```
VojtaStavik, line 144: Can we make this in `viewWillAppear` such that the animation happens at the same time as the presentation animation? It will look much nicer to have it in a single step.

VojtaStavik (approved):

Sources_v3/StreamChatUI/MessageActionsPopup/ChatMessagePopupViewController.swift
VojtaStavik: I see the autolayout is not done when `viewWillAppear` is called. Let's leave as is then and polish later.

PR #707, opened 2020-12-30
[CIS-465] Add reaction on tapping user mention

DominikBucher12:
# What this PR Does
- Implements tapping on user suggestions, so it fills the user id to the composer textView

https://user-images.githubusercontent.com/17381941/103372861-82538a80-4ad3-11eb-9a70-41c37da4f92d.mov

### DIFF
--- a/Sources_v3/StreamChatUI/ChatChannel/MessageComposer/MessageComposerViewController.swift
+++ b/Sources_v3/StreamChatUI/ChatChannel/MessageComposer/MessageComposerViewController.swift
@@ -436,19 +436,24 @@ open class MessageComposerViewController<ExtraData: ExtraDataTypes>: ViewControl
                 guard let self = self else { return }
 
                 let user = self.userSuggestionSearchController.users[userIndex]
-                let cursorPositon = textView.selectedRange
+                let cursorPosition = textView.selectedRange
 
                 let atRange = (textView.textStorage.string as NSString)
                     .rangeOfCharacter(
                         from: CharacterSet(charactersIn: "@"),
                         options: .backwards,
-                        range: NSRange(location: 0, length: cursorPositon.location)
+                        range: NSRange(location: 0, length: cursorPosition.location)
                     )
 
+                let oldPositionTillTheEnd = (textView.text as NSString).length - cursorPosition.location
+
                 textView.textStorage.replaceCharacters(
-                    in: NSRange(location: atRange.location, length: cursorPositon.location - atRange.location),
-                    with: "@\(user.id)"
+                    in: NSRange(location: atRange.location, length: cursorPosition.location - atRange.location),
+                    with: "@\(user.id) "
                 )
+
+                let newPosition = (textView.text as NSString).length - oldPositionTillTheEnd
+                textView.selectedRange = NSRange(location: newPosition, length: 0)
                 self.dismissSuggestionsViewController()
             }
         )
--- a/Sources_v3/StreamChatUI/ChatChannel/MessageComposer/MessageComposerViewController.swift
+++ b/Sources_v3/StreamChatUI/ChatChannel/MessageComposer/MessageComposerViewController.swift
@@ -27,6 +27,7 @@ open class MessageComposerViewController<ExtraData: ExtraDataTypes>: ViewControl
     // MARK: - Properties
 
     public var controller: _ChatChannelController<ExtraData>?
+    var shouldShowMentions = false
     
     public var state: State = .initial {
         didSet {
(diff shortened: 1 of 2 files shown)


Sources_v3/StreamChatUI/ChatChannel/MessageComposer/MessageComposerViewController.swift
```diff
@@ -384,15 +385,16 @@ open class MessageComposerViewController<ExtraData: ExtraDataTypes>: ViewControl
         composerView.invalidateIntrinsicContentSize()
     }
     
-    // MARK: UITextView
+    // MARK: Suggestions
 
-    @objc func promptSuggestionIfNeeded(for text: String) {
-        if text.trimmingCharacters(in: .whitespacesAndNewlines).contains("@") {
-            if let index = (text.range(of: "@")?.upperBound) {
+    private func promptSuggestionIfNeeded(for text: String) {
```
VojtaStavik, line 390: the rule is - unless there's a strong reason for not having something `open` it should be open. Tell me why `promptSuggestionIfNeeded(for text: String)` should not be `open` please :)

Sources_v3/StreamChatUI/ChatChannel/MessageComposer/MessageComposerViewController.swift
```diff
@@ -453,10 +467,26 @@ open class MessageComposerViewController<ExtraData: ExtraDataTypes>: ViewControl
...
         replaceTextWithSlashCommandViewIfNeeded()
-        promptSuggestionIfNeeded(for: textView.text)
+
+        updateMentionFlag(with: textView.text as NSString, till: textView.selectedRange.location)
+
+        promptSuggestionIfNeeded(for: textView.text!)
         composerView.invalidateIntrinsicContentSize()
     }
 
+    func updateMentionFlag(with text: NSString, till caret: Int) {
```
VojtaStavik, line 477: `NSString` ?? 🦕

VojtaStavik (commented):
I have questions :)

Sources_v3/StreamChatUI/ChatChannel/MessageComposer/MessageComposerViewController.swift
DominikBucher12: I guess you are right, maybe I didn't think too much high-level and started to complicate stuff, I thought that this function should be called only from the `textViewDidChange` so it makes kinda sense for the function to be private. Changed now.

Sources_v3/StreamChatUI/ChatChannel/MessageComposer/MessageComposerViewController.swift
DominikBucher12: `NSRange` and `Range<String.Index>` don't like each other and do not cooperate too well with different kind of strings. I feel this can be done better, in the same time I don't know how -> I spent already too much time figuring out how to do this with `String`, then @ManWithBear and prove me you can sometimes use something with `NS` prefix and it doesn't hurt...

VojtaStavik (approved):
Looks OK for now, let's merge and deal with details later. Thanks!

Sources_v3/StreamChatUI/ChatChannel/MessageComposer/MessageComposerViewController.swift
ManWithBear: In 2 words: 
- `String` and `NSString` have different unicode handling so you cannot interchange indexes between them
- `UITextView` uses `NSAttributedString` subclass for text storage (`NSTextStorage`), which operates on `NSString`

Sources_v3/StreamChatUI/ChatChannel/MessageComposer/MessageComposerViewController.swift
```diff
@@ -421,16 +423,28 @@ open class MessageComposerViewController<ExtraData: ExtraDataTypes>: ViewControl
...
+                    .rangeOfCharacter(
+                        from: CharacterSet(charactersIn: "@"),
+                        options: .backwards,
+                        range: NSRange(location: 0, length: cursorPositon.location)
+                    )
+
+                textView.textStorage.replaceCharacters(
+                    in: NSRange(location: atRange.location, length: cursorPositon.location - atRange.location),
+                    with: "@\(user.id)"
+                )
```
ManWithBear, line 446: After you replaced mention with id, move cursor to the end of `user.id` and add 1 space character.

ManWithBear (changes requested):
Here small thing that will be awesome to get done before merging.

Sources_v3/StreamChatUI/ChatChannel/MessageComposer/MessageComposerViewController.swift
DominikBucher12: Thanks! :) Done

ManWithBear (approved):

PR #708, opened 2020-12-31
Polish Chat creation screen

b-onc:
Adds state management and actual chat creation (with the help of composer!)

Missing: Displaying chat when the user(s) are selected

Checkout and test is recommended!

https://user-images.githubusercontent.com/770464/103410677-af07b080-4b6c-11eb-875a-386a1e957eab.mov


VojtaStavik (approved):
Of course, there's room for improvements, but I like how simple the solution is :) Thank you!

PR #709, opened 2020-12-31
Fix db crash on user change

b-onc:
### Issue description

We were removing `persistentStore` in `removeAllData` when we detected a new user login, but this causes crashes if there are ongoing `write` actions accessing the said store.

### Details

When we initialize the `DatabaseContainer` first, we reset ephemeral values with `resetEphemeralValues` async func:
```swift
/// Iterates over all items and if the DTO conforms to `EphemeralValueContainers` calls `resetEphemeralValues()` on
/// every object.
func resetEphemeralValues() {
    write({ session in
        let session = session as! NSManagedObjectContext
        
        try self.managedObjectModel.entities.forEach { entityDescription in
            guard let entityName = entityDescription.name else { return }
            let fetchRequest: NSFetchRequest<NSFetchRequestResult> = NSFetchRequest(entityName: entityName)
            let entities = try session.fetch(fetchRequest) as? [EphemeralValuesContainer]
            entities?.forEach { $0.resetEphemeralValues() }
        }
    }, completion: { error in
        if let error = error {
            log.error("Error resetting ephemeral values: \(error)")
        } else {
            log.debug("Ephemeral values reset.")
        }
    })
}
```
which doesn't complete immediately.

When there's a new user detected, we reset all db, in `prepareEnvironmentForNewUser`:
```swift
// Reset all existing data if the new user is not the same as the last logged-in one
if client.databaseContainer.viewContext.currentUser()?.user.id != userId {
    // Re-create backgroundWorker's so their ongoing requests won't affect database state

    do {
        // Reset all existing local data
        try client.databaseContainer.removeAllData(force: true)

    } catch {
        completion(error)
    }
}
```

When these 2 called back to back, we were removing the `persistentStore` while there's an ongoing write on it, so it caused the crash.

Moreover, `AttachmentUploader` and `MesageEditor` was calling `database.write` unnecessarily (while there are no pending items)

### Discussion

While testing the new login times, I didn't realize considerable delay/lag caused by `perform -> performAsync` change, though I don't have concrete numbers.

The changes in `ChannelUpdater` and `ChannelListUpdater` were trials on resolving the crashes, in the end it turned out they didn't have anything to do with it, but they are required for consistency among uploaders, so I've kept it.

### DIFF
--- a/Sources_v3/StreamChat/Database/DatabaseContainer.swift
+++ b/Sources_v3/StreamChat/Database/DatabaseContainer.swift
@@ -263,21 +263,19 @@ extension DatabaseContainer {
     /// Iterates over all items and if the DTO conforms to `EphemeralValueContainers` calls `resetEphemeralValues()` on
     /// every object.
     func resetEphemeralValues() {
-        write({ session in
-            let session = session as! NSManagedObjectContext
-            
-            try self.managedObjectModel.entities.forEach { entityDescription in
-                guard let entityName = entityDescription.name else { return }
-                let fetchRequest: NSFetchRequest<NSFetchRequestResult> = NSFetchRequest(entityName: entityName)
-                let entities = try session.fetch(fetchRequest) as? [EphemeralValuesContainer]
-                entities?.forEach { $0.resetEphemeralValues() }
-            }
-        }, completion: { error in
-            if let error = error {
-                log.error("Error resetting ephemeral values: \(error)")
-            } else {
+        writableContext.performAndWait {
+            do {
+                try self.managedObjectModel.entities.forEach { entityDescription in
+                    guard let entityName = entityDescription.name else { return }
+                    let fetchRequest: NSFetchRequest<NSFetchRequestResult> = NSFetchRequest(entityName: entityName)
+                    let entities = try writableContext.fetch(fetchRequest) as? [EphemeralValuesContainer]
+                    entities?.forEach { $0.resetEphemeralValues() }
+                }
+                try writableContext.save()
                 log.debug("Ephemeral values reset.")
+            } catch {
+                log.error("Error resetting ephemeral values: \(error)")
             }
-        })
+        }
     }
 }
--- a/Sources_v3/StreamChat/Workers/Background/AttachmentUploader.swift
+++ b/Sources_v3/StreamChat/Workers/Background/AttachmentUploader.swift
@@ -52,6 +52,8 @@ class AttachmentUploader<ExtraData: ExtraDataTypes>: Worker {
     }
 
     private func handleChanges(changes: [ListChange<AttachmentDTO>]) {
+        guard !changes.isEmpty else { return }
+        
         var wasEmpty: Bool!
         _pendingAttachmentIDs.mutate { pendingAttachmentIDs in
             wasEmpty = pendingAttachmentIDs.isEmpty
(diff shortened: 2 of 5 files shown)


Sources_v3/StreamChat/Workers/Background/AttachmentUploader.swift
```diff
@@ -57,7 +57,7 @@ class AttachmentUploader<ExtraData: ExtraDataTypes>: Worker {
             wasEmpty = pendingAttachmentIDs.isEmpty
             changes.pendingUploadAttachmentIDs.forEach { pendingAttachmentIDs.insert($0) }
         }
-        if wasEmpty {
+        if wasEmpty, !pendingAttachmentIDs.isEmpty {
```
ManWithBear, line 60: nit:
```
let notEmptyNow = !pendingAttachmentIDs.isEmpty
if wasEmpty && notEmptyNow {
```
I took me a while to understand why you don't have contradiction here (`wasEmpty = pendingAttachmentIDs.isEmpty`).
I think this small change will make intention more clear and readable

ManWithBear: Hm, it may be even better to just 
```
guard !changes.isEmpty else { return }
```
at beginning of function, so we would not lock context for no reason

Sources_v3/StreamChat/Workers/ChannelUpdater.swift
```diff
@@ -25,7 +25,12 @@ class ChannelUpdater<ExtraData: ExtraDataTypes>: Worker {
                 channelCreatedCallback?(payload.channel.cid)
                 self.database.write { (session) in
                     try session.saveChannel(payload: payload)
-                    completion?(nil)
+                } completion: { error in
+                    if let error = error {
+                        completion?(error)
```
ManWithBear, line 30: either you forgot `log` here, or this unwrap do nothing

Sources_v3/StreamChat/Database/DatabaseContainer.swift
```diff
@@ -263,21 +263,19 @@ extension DatabaseContainer {
...
-                guard let entityName = entityDescription.name else { return }
-                let fetchRequest: NSFetchRequest<NSFetchRequestResult> = NSFetchRequest(entityName: entityName)
-                let entities = try session.fetch(fetchRequest) as? [EphemeralValuesContainer]
-                entities?.forEach { $0.resetEphemeralValues() }
-            }
-        }, completion: { error in
-            if let error = error {
-                log.error("Error resetting ephemeral values: \(error)")
-            } else {
+        writableContext.performAndWait {
```
ManWithBear, line 266: I checked call stack for this method, and unless I missed something, it will be called on main thread.
So for with increasing storage, it will take more and more time, meaning locking main thread for longer and longer

b-onc: `performAndWait` documentation:
> Synchronously performs a given block on the *context’s queue*.
`writableContext` is a background context, so it shouldn't be executed on main thread, interesting that you've seen such behavior

ManWithBear: Yes, main point is: `Synchronously performs`

b-onc: Ooh you mean that point.. As I wrote in the discussion part of the description, I didn't see considerable delay, but that's just anectodal.. So what you've raised is correct, and I need suggestions on resolving this completely

VojtaStavik (approved):
I think this is a great fix for now. We will have to re-think the way we treat ephemeral values anyway, so this part will (hopefully) go away before it starts making troubles.

Sources_v3/StreamChat/Workers/ChannelUpdater.swift
b-onc: Fixed

Sources_v3/StreamChat/Workers/Background/AttachmentUploader.swift
b-onc: Fixed with `guard` solution

PR #713, opened 2021-01-01
Improve dark mode experience

ManWithBear:
Interesting points:
- `mode` naming could be better
- on traits changes we need to call `updateContent` because some UI change depends on data. But it should happen only twice a day, only for views on screen, so I would say it's fine

_Cauz of upload limits, video can be found in slack_

### DIFF
--- a/Sources_v3/StreamChatUI/ChatChannel/Attachments/ChatFileAttachmentListView+ItemView.swift
+++ b/Sources_v3/StreamChatUI/ChatChannel/Attachments/ChatFileAttachmentListView+ItemView.swift
@@ -1,5 +1,5 @@
 //
-// Copyright © 2020 Stream.io Inc. All rights reserved.
+// Copyright © 2021 Stream.io Inc. All rights reserved.
 //
 
 import StreamChat
@@ -18,7 +18,7 @@ extension ChatFileAttachmentListView {
         // MARK: - Overrides
 
         override public func defaultAppearance() {
-            backgroundColor = .white
+            backgroundColor = uiConfig.colorPalette.incomingMessageBubbleBackground
             layer.cornerRadius = 12
             layer.masksToBounds = true
             layer.borderWidth = 1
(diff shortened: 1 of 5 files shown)


Sources_v3/StreamChatUI/ChatChannel/Attachments/ChatFileAttachmentListView+ItemView.swift
```diff
@@ -1,5 +1,5 @@
 //
-// Copyright © 2020 Stream.io Inc. All rights reserved.
+// Copyright © 2021 Stream.io Inc. All rights reserved.
```
ManWithBear, line 2: This is stupid. Can we like prevent it somehow?

Sources_v3/StreamChatUI/UIConfig.swift
```diff
@@ -344,16 +344,33 @@ public extension UIConfig {
...
+    static let streamAccentBlue = mode(0x005fff, 0x005fff)
+    static let streamAccentRed = mode(0xff3742, 0xff3742)
+    static let streamAccentGreen = mode(0x20e070, 0x20e070)
+
+    static let streamBGGradientFrom = mode(0xf7f7f7, 0x101214)
+    static let streamBGGradientTo = mode(0xfcfcfc, 0x070a0d)
+    static let streamOverlay = mode(0x000000, lightAlpha: 0.2, 0x000000, darkAlpha: 0.4)
+    static let streamOverlayDark = mode(0x000000, lightAlpha: 0.6, 0xffffff, darkAlpha: 0.8)
+
+    static func mode(_ light: Int, lightAlpha: CGFloat = 1.0, _ dark: Int, darkAlpha: CGFloat = 1.0) -> UIColor {
```
DominikBucher12, line 365: Nice 👍

VojtaStavik (approved):
Looks great, thanks!

PR #715, opened 2021-01-04
Enable bouncing on small content message list

ManWithBear:
#### Instance Property
# alwaysBounceVertical
A Boolean value that determines whether bouncing always occurs when vertical scrolling reaches the end of the content.

### Declaration
```
var alwaysBounceVertical: Bool { get set }
```
### Discussion
If this property is set to `true` and bounces is `true`, vertical dragging is allowed even if the content is smaller than the bounds of the scroll view. The default value is `false`.

### DIFF
--- a/Sources_v3/StreamChatUI/ChatChannel/ChatMessageListVC.swift
+++ b/Sources_v3/StreamChatUI/ChatChannel/ChatMessageListVC.swift
@@ -1,5 +1,5 @@
 //
-// Copyright © 2020 Stream.io Inc. All rights reserved.
+// Copyright © 2021 Stream.io Inc. All rights reserved.
 //
 
 import StreamChat
--- a/Sources_v3/StreamChatUI/ChatChannel/ChatMessageListVC.swift
+++ b/Sources_v3/StreamChatUI/ChatChannel/ChatMessageListVC.swift
@@ -67,6 +67,7 @@ open class ChatMessageListVC<ExtraData: ExtraDataTypes>: ViewController,
         collection.register(incomingCell, forCellWithReuseIdentifier: incomingCell.reuseId)
         collection.register(outgoingCell, forCellWithReuseIdentifier: outgoingCell.reuseId)
         collection.showsHorizontalScrollIndicator = false
+        collection.alwaysBounceVertical = true
         collection.keyboardDismissMode = .onDrag
         collection.dataSource = self
         collection.delegate = self


VojtaStavik (commented):
Not sure if this is the best solution. Can you elaborate a bit, please?

ManWithBear:
Original task:
> The issue is that when there's just one message in the channel (thread), the message can't be scrolled and is frozen to the top.

We can extract 2 main points:
- when content of message list is smaller than frame of message list
- message list needs to bounce, but not bounce right now

So let's take a look on `bounce` property:
> A Boolean value that controls whether the scroll view bounces past the edge of content and back again.
> ### Declaration 
> ```
> var bounces: Bool { get set }
> ```
> ### Discussion
> If the value of this property is `true`, the scroll view bounces when it encounters a boundary of the content. Bouncing visually indicates that scrolling has reached an edge of the content. If the value is `false`, scrolling stops immediately at the content boundary without bouncing. The default value is `true`.

On first glance you think: "Yes, this is exactly what we need".
But second though says: "Hm, it's on by default, and content bounce, but not on small content".
So now can make resolution that this is not an issue.
Continue looking.

Without any ideas, we continue monitoring `bounds` documentation, until our sight catch "See also" section:
> ```
> var alwaysBounceVertical: Bool
> ```
> A Boolean value that determines whether bouncing always occurs when vertical scrolling reaches the end of the content.

"Hm, what a nonsense, it's exactly same as `bounces`!"
We decide to dig into it, because why here would be 2 properties for same thing? 
Somewhere in the middle of documentation we found our gem:
> If this property is set to `true` and bounces is `true`, vertical dragging **is allowed even if the content is smaller than the bounds of the scroll view**.

Attentive reader would shout at this point: "Wait! Does it mean `bounces` does not bounce when content is smaller than the bounds?!"
Yes, you right my friend. This is key to our problem.

And now we heart full of confidence we can commit and send PR that solves this horrible bug!

VojtaStavik:
Seems a bit overkill to me @ManWithBear. My gut feeling is there must be a simpler solution. What's your take on this @b-onc ?

ManWithBear:
Hm, I think @evsaev would agree with me, that sometime for ideal solution it's fine tog o over a line a little bit.

Sources_v3/StreamChatUI/ChatChannel/ChatMessageListVC.swift
```diff
@@ -1,5 +1,5 @@
 //
-// Copyright © 2020 Stream.io Inc. All rights reserved.
+// Copyright © 2021 Stream.io Inc. All rights reserved.
```
VojtaStavik, line 2: This should be in a separate commit.

ManWithBear: Fixed 👍🏾

VojtaStavik (approved):

PR #721, opened 2021-01-06
Minor SDK Fixes 

DominikBucher12:
# What does this PR do:
- Fix of Access Modifiers for `View` subclasses / `Customizable` implementations  across the SDK usage, i.e. now functions that should be public are public, functions that should be open (everything except `defaultAppearance`) are open 🎊 .
- Adds `SwipeableListItemView` to `UIConfig`.
- Adds 8 points inset from top to Message List UI.

Main thing to test in here is the inset (check if autoscroll to bottom works properly), other things should be clear. :)

### DIFF
--- a/Sources_v3/StreamChatUI/ChatChannel/ChatMessageListVC.swift
+++ b/Sources_v3/StreamChatUI/ChatChannel/ChatMessageListVC.swift
@@ -99,7 +99,7 @@ open class ChatMessageListVC<ExtraData: ExtraDataTypes>: ViewController,
 
         view.addSubview(collectionView)
         collectionView.pin(to: view.safeAreaLayoutGuide)
-        collectionView.contentInset = UIEdgeInsets(top: 0, left: 0, bottom: 8, right: 0)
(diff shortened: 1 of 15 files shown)


Sources_v3/StreamChatUI/ChatChannel/ChatMessageListVC.swift
```diff
@@ -99,7 +99,7 @@ open class ChatMessageListVC<ExtraData: ExtraDataTypes>: ViewController,
 
         view.addSubview(collectionView)
         collectionView.pin(to: view.safeAreaLayoutGuide)
-        collectionView.contentInset = UIEdgeInsets(top: 0, left: 0, bottom: 8, right: 0)
+        collectionView.contentInset = UIEdgeInsets(top: 8, left: 0, bottom: 8, right: 0)
```
VojtaStavik, line 102: Does this work correctly with our custom layout? cc @ManWithBear

ManWithBear: Looks like it's still working. Dunno why.

VojtaStavik (approved):
OK, let's merge and see what breaks :)

PR #723, opened 2021-01-06
[CIS-555] Make ChatClient available in app extensions

b-onc:
### ~Open~ Answered Questions
1. `isConnectionless` naming :)
> We decided to go with `isClientInActiveMode`, thinking that Client can be active vs passive 
1. Should we allow creation of connectionless clients for main app? Or full client in app extensions? (basically making `isConnectionless` a parameter)
> Why not? Good as long as we document it properly ¯\_(ツ)_/¯ 
1. For operations available to connectionless client but doesn't make sense (`connect`, `disconnect`, `setUser`) should we:
```swift
completion?(ClientError.ClientInConnectionlessMode())
```
or
```swift
log.warning("The client is in connectionless mode. Skipping the `connect` call.")
completion?(nil)
```
> Error returning approach is better

### DIFF
--- a/Sources_v3/StreamChat/ChatClient.swift
+++ b/Sources_v3/StreamChat/ChatClient.swift
@@ -120,7 +120,7 @@ public class _ChatClient<ExtraData: ExtraDataTypes> {
     }()
     
     /// The `WebSocketClient` instance `Client` uses to communicate with Stream WS servers.
-    lazy var webSocketClient: WebSocketClient = {
+    lazy var webSocketClient: WebSocketClient? = {
         // Create a connection request
         let webSocketEndpoint: Endpoint<EmptyResponse> = .webSocketConnect(
             userId: self.currentUserId,
@@ -142,7 +142,7 @@ public class _ChatClient<ExtraData: ExtraDataTypes> {
             internetConnection
         )
         
-        webSocketClient.connectionStateDelegate = self
+        webSocketClient?.connectionStateDelegate = self
         
         return webSocketClient
     }()
@@ -269,6 +269,7 @@ public class _ChatClient<ExtraData: ExtraDataTypes> {
         self.workerBuilders = workerBuilders
         
         createBackgroundWorkers()
+        createWebSocketClient()
     }
     
     deinit {
@@ -297,6 +298,13 @@ public class _ChatClient<ExtraData: ExtraDataTypes> {
             builder(self.databaseContainer, self.webSocketClient, self.apiClient)
         }
     }
+    
+    /// Before CIS-555, `webSocketClient` was initialized in `createBackgroundWorkers` in `init` since it was passed
+    /// to the background workers. Now, background workers do not access `webSocketClient` and it's not being
+    /// initialized in `init`, so we initialize it here explicitly.
+    private func createWebSocketClient() {
+        _ = webSocketClient
+    }
 }
 
 extension _ChatClient {
@@ -315,7 +323,7 @@ extension _ChatClient {
             _ eventDecoder: AnyEventDecoder,
             _ notificationCenter: EventNotificationCenter,
             _ internetConnection: InternetConnection
-        ) -> WebSocketClient = {
+        ) -> WebSocketClient? = {
             WebSocketClient(
                 connectEndpoint: $0,
                 sessionConfiguration: $1,
--- a/Sources_v3/StreamChat/ChatClient.swift
+++ b/Sources_v3/StreamChat/ChatClient.swift
@@ -96,6 +96,9 @@ public class _ChatClient<ExtraData: ExtraDataTypes> {
     
     /// Builder blocks used for creating `backgroundWorker`s when needed.
     private let workerBuilders: [WorkerBuilder]
+    
+    /// Builder blocks used for creating `backgroundWorker`s dealing with events when needed.
+    private let eventWorkerBuilders: [EventWorkerBuilder]
 
     /// The notification center used to send and receive notifications about incoming events.
     private(set) lazy var eventNotificationCenter = environment.notificationCenterBuilder([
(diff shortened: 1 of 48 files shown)


Sources_v3/StreamChat/Config/ChatClientConfig.swift
```diff
@@ -50,12 +50,19 @@ public struct ChatClientConfig {
     /// refreshed.
     public var tokenProvider: TokenProvider?
     
+    /// Flag for setting a ChatClient instance in connection-less mode.
+    /// A connection-less client is not able to connect to websocket and will not
+    /// receive websocket events. It can still observe and mutate database.
+    /// This flag is automatically set to `true` for app extensions
+    public let isConnectionless: Bool
+
```
VojtaStavik, line 58: maybe `isClientInActiveMode` + "This flag is automatically set to `false` for app extensions" ?

VojtaStavik: Or maybe we can create a `Mode` enum with `active` and `passive` values? 🤔

Sources_v3/StreamChat/Controllers/CurrentUserController/CurrentUserController.swift
```diff
@@ -133,14 +139,20 @@ public class _CurrentChatUserController<ExtraData: ExtraDataTypes>: Controller,
         extraData: ExtraData.User? = nil,
         completion: @escaping (Error?) -> Void
     ) {
+        // Setting a new user is not possible in connectionless mode
+        guard let webSocketClient = client.webSocketClient else {
+            completion(ClientError.ClientInConnectionlessMode())
+            return
+        }
+
```
VojtaStavik, line 147: 👍

Sources_v3/StreamChat/ChatClient.swift
```diff
@@ -406,6 +406,10 @@ extension _ChatClient: ConnectionDetailsProviderDelegate {
     func provideConnectionId(completion: @escaping (String?) -> Void) {
         if let connectionId = connectionId {
             completion(connectionId)
+        } else if config.isConnectionless {
+            // We're in connectionless mode
+            // We will never have connectionId
+            completion(nil)
```
VojtaStavik, line 412: 👍

VojtaStavik (commented):
Look very good to me 👍

VojtaStavik:
@b-onc I just realized, we shouldn't reset DB ephemeral values in passive clients. Only active clients should to that.

Sources_v3/StreamChat/ChatClient.swift
```diff
@@ -274,6 +274,7 @@ public class _ChatClient<ExtraData: ExtraDataTypes> {
         self.eventWorkerBuilders = eventWorkerBuilders
         
         createBackgroundWorkers()
+        createWebSocketClient()
```
VojtaStavik, line 277: What's wrong with having the WS client initialized lazily?

Sources_v3/StreamChat/ChatClient_Tests.swift
```diff
@@ -334,28 +344,26 @@ class ChatClient_Tests: StressTestCase {
...
         let client = _ChatClient(
             config: config,
             workerBuilders: [TestWorker.init],
+            eventWorkerBuilders: [],
             environment: testEnv.environment
         )
         
         let testWorker = client.backgroundWorkers.first as? TestWorker
         XCTAssert(testWorker?.init_database is DatabaseContainerMock)
-        XCTAssert(testWorker?.init_webSocketClient is WebSocketClientMock)
```
VojtaStavik, line 358: This should be probably replaced with a test for proper initialization of event workers?

Sources_v3/StreamChat/ChatClient.swift
b-onc: it is (was) initialized lazily, but that meant it was initialized in `createBackgroundWorkers` when it was passed to the workers. Tests assume this behavior (that wsclient is ready after `ChatClient.init`)
After I removed the dependency worker -> wsclient, tests failed because of that

Sources_v3/StreamChat/ChatClient.swift
```diff
@@ -304,6 +305,13 @@ public class _ChatClient<ExtraData: ExtraDataTypes> {
             builder(self.databaseContainer, self.eventNotificationCenter, self.apiClient)
         }
     }
+    
+    /// Before CIS-555, `webSocketClient` was initialized in `createBackgroundWorkers` in `init` since it was passed
+    /// to the background workers. Now, background workers do not access `webSocketClient` and it's not being
+    /// initialized in `init`, so we initialize it here explicitly.
+    private func createWebSocketClient() {
+        _ = webSocketClient
+    }
```
VojtaStavik, line 314: I don't get why we have the comment here. Can't we just do it the way such that it doesn't raise any questions and doesn't require explaining comments? :)

VojtaStavik (approved):
Looks good to me! 🚢 🇮🇹 Thanks @b-onc !

Sources_v3/StreamChat/ChatClient_Tests.swift
```diff
@@ -10,14 +10,21 @@ class ChatClient_Tests: StressTestCase {
...
         config.baseURL = BaseURL(urlString: .unique)!
         return config
     }
     
+    // Helper for providing config with in-memory DB option and passive (inactive) mode
+    lazy var inactiveInMemoryStorageConfig: ChatClientConfig = {
+        var config = inMemoryStorageConfig
+        config.isClientInActiveMode = false
+        return config
+    }()
```
VojtaStavik, line 26: nit: I don't think "passive" means the same thing as "inactive"

b-onc: Only in our context :D I opted for the boolean flag for simplicity, so anything inactive = passive (May Zeus protect me from grammar people)

PR #724, opened 2021-01-06
[CIS-571] Constraints priority

ManWithBear:
Main part of changes have been done via Xcode Find/Replace
```
Anchor.constraint(
```
Changed to
```
Anchor.pin(
```

I run app and it looks fine, all layout warnings are gone. 
But take a look as well, I could miss something important

Example of custom rule warning:
rule

### DIFF
--- a/Sources_v3/StreamChatUI/Utils/NSLayoutConstraint+Extensions.swift
+++ b/Sources_v3/StreamChatUI/Utils/NSLayoutConstraint+Extensions.swift
@@ -1,19 +1,14 @@
 //
-// Copyright © 2020 Stream.io Inc. All rights reserved.
+// Copyright © 2021 Stream.io Inc. All rights reserved.
 //
 
 import UIKit
 
 extension UILayoutPriority {
-    /// The `UILayoutPriority` that is less than required by 1.
-    ///
-    /// Such helper is especially handy when working with `UIStackView`:
-    /// when the arranged subview is hidden, the constraint for `height == 0`
-    /// with `.required` priority is activated.
-    ///
-    /// Having all height constraints inside the `UIStackView` with `almostRequired`
-    /// priority eliminates breaking constraints warnings in console.
-    static let almostRequired: Self = .required - 1
+    /// Having our default priority lower than `.required(1000)` allow user easily
+    /// override any default constraints and customize layout
+    static let streamRequire = UILayoutPriority(rawValue: 900)
+    static let streamAlmostRequire: UILayoutPriority = .streamRequire - 1
 }
 
 extension NSLayoutConstraint {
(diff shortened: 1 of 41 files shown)


Sources_v3/StreamChatUI/Utils/NSLayoutConstraint+Extensions.swift
```diff
@@ -1,19 +1,14 @@
...
-    /// when the arranged subview is hidden, the constraint for `height == 0`
-    /// with `.required` priority is activated.
-    ///
-    /// Having all height constraints inside the `UIStackView` with `almostRequired`
-    /// priority eliminates breaking constraints warnings in console.
-    static let almostRequired: Self = .required - 1
+    /// Having our default priority lower than `.required(1000)` allow user easily
+    /// override any default constraints and customize layout
+    static let streamRequire = UILayoutPriority(rawValue: 900)
+    static let streamAlmostRequire: UILayoutPriority = .streamRequire - 1
```
DominikBucher12, line 11: I'd add documentation also to this variable so people do not need to go to the code and see the concrete value (it's 899)  

Example:

```
/// Having our default priority lower than `.required(1000)` allow user easily
/// override any default constraints and customize layout
/// That's why this value equals `900` instead of `1000`.
static let streamRequire = UILayoutPriority(rawValue: 900) 

/// In our space, this is `streamRequire` - `1`, i.e. 899
static let streamAlmostRequire: UILayoutPriority = .streamRequire - 1
``` 

Or something like that to make it more clear :)

Sources_v3/StreamChatUI/Utils/NSLayoutConstraint+Extensions.swift
```diff
@@ -25,8 +20,13 @@ extension NSLayoutConstraint {
...
 
-    /// Returns updated `self` with `priority == .almostRequired`
+    /// Returns updated `self` with `priority == .streamAlmostRequire`
     var almostRequired: NSLayoutConstraint {
-        with(priority: .almostRequired)
+        with(priority: .streamAlmostRequire)
+    }
+
+    /// Default constraint priority
+    var stream: NSLayoutConstraint {
```
DominikBucher12, line 29: `NSLayoutConstraint().stream` feels strange. 

What about `NSLayoutConstraint().withStreamPriority` and `NSLayoutConstraint().withAlmostRequiredPriority`. IMO The Apple API for defining constraints `with(priority:) feels really good and those extensions might be not needed... because the apple function is explicit enough. But this is just a tought.

Sources_v3/StreamChatUI/Utils/NSLayoutConstraint+Extensions.swift
ManWithBear: IMHO final value is implementation detail and do not bring value to documentation.
If you need their concrete values for some reason, probably something wrong.

VojtaStavik (approved):
Looks great, thank you @ManWithBear !

PR #727, opened 2021-01-08
[V3] Fix crash when dismissing ChannelListVC presented modally

cardoso:
It looks like this stopObserving logic is not needed, or else `deinit` wouldn't be called anyway.


VojtaStavik (approved):
Thank you @cardoso !

PR #728, opened 2021-01-11
Add Slack bot on failed tests

b-onc:
_hurray_


VojtaStavik (approved):
Great! Just one detail - I think we only need to do it for tests run on master. No need to do it for PRs.

PR #730, opened 2021-01-11
[CIS-235] Update current user

nuno-vieira:
**In this PR:**
Adds a new method `updateCurrentUser()` to the LLC for updating the current user data. 
It uses the [partial update endpoint](https://getstream.io/chat/docs_rest/#partial-update-users). The endpoint supports resetting user data but is out of the scope of this PR.
Adds support to update the user data in the Sample app.

- [X] Updating partially new user data
- [x] Save most updated data to DB
- [x] Add support to change user data in the Sample App

https://user-images.githubusercontent.com/12814114/104375662-a4113080-551b-11eb-8f63-de2ef940e8db.mov


Sources_v3/StreamChat/APIClient/Endpoints/CurrentUserEndpoints_Tests.swift
```diff
@@ -0,0 +1,31 @@
...
+
+@testable import StreamChat
+import XCTest
+
+final class CurrentUserEndpoints_Tests: XCTestCase {
+    func test_updateCurrentUser_buildsCorrectly() {
+        let userId = "123"
+        let payload: CurrentUserUpdateRequestBody<DefaultExtraData.User> = .init(
+            id: userId,
+            set: .init(name: "Nuno", imageURL: URL(string: "Fake")!, extraData: .init()),
```
VojtaStavik, line 13: nit: Please use only StartWars related dummy data.

image

Sources_v3/StreamChat/APIClient/RequestEncoder.swift
```diff
@@ -151,6 +151,8 @@ struct DefaultRequestEncoder: RequestEncoder {
             try encodeGETorDELETERequestBody(request: &request, endpoint: endpoint)
         case .post:
             try encodePOSTRequestBody(request: &request, endpoint: endpoint)
+        case .patch:
+            try encodePATCHRequestBody(request: &request, endpoint: endpoint)
```
VojtaStavik, line 155: This will need to be tested, too

nuno-vieira:
Converted to draft, since it looks the base branch is not correct. This needs to be fixed.

PR #731, opened 2021-01-12
[CIS-574] Token based user setting

evsaev:
# Introduction

This PR simplifies user-setting API so we have the single flow for all the cases:
- refreshing a token for the current user
- using another token provider for the current user
- setting another user by setting a different token provider

# Making the current user ID optional

Having a non-optional `currentUserId` in **ChatClient**:
- is not consistent with `currentUser: CurrentUser?` in **CurrentUserController** which can be `nil` if there's no current user saved in the database
- is apparently invalid and can be confusing since we have `currentUserID` set even though there's no currently logged in user.

```swift
 public class _ChatClient<ExtraData: ExtraDataTypes> {
     /// The `UserId` of the currently logged in user.
-    @Atomic public var currentUserId: UserId = .anonymous
+    @Atomic public var currentUserId: UserId?
```

Previously the `currentUserId` property was initialized with random `.anonymous` value for the single purpose: make the access simple for end-users, avoiding unwrapping the optional. 

But it turns out if we add additional values to the models (`message.isSentByCurrentUser`, `reation.isCurrentUserReaction`) we can cover most of the cases where `currentUserID` previously had to be accessed and provide a friendly and convenient API for end-users.

_When **ChatClient** is instantiated, the `currentUserId` is taken from the database so it is now consistent with **CurrentUserController**._

# Moving reconnection logic to ChatClient

All stuff from **CurrentUserController** related to setting another user (`setUser/setGuestUser/setAnonymousUser`) is gone and replaces with the single `reconnectIfNeeded` (naming is not final) on a **ChatClient**. 

The decision to move  `reconnectIfNeeded` from **CurrentUserController ->  ChatClient** was made based on the following:
1. `reconnectIfNeeded` is more about establishing a connection other than updating/observing the current user
2. `reconnectIfNeeded` requires whipping the environment including **@Atomic** properties which should be done via acesssing `mutate` on the backing variables which cannot be done from the outside (previously we were doing it in the wrong non-thread-safe way)

# Token waiters

From the inside of **ChatClient.init** background-workers are instantiated. Workers start working immediately and can initiate API calls before the token for the current user is provided. 

_**The use-case:** SDK end-users decide to create a **ChatClient** on app launch but postpone `reconnectIfNeeded` until the chat is opened._

Such cases are handled in the same way we deal with API calls that require `connection-id`: 
- the **Endpoint** type is extended with `requiresToken: Bool`
- the **RequestEncoder** is updated to ask the delegate (the **ChatClient**) for the `token`
- the **ChatClient** is updated to have `tokenWaiters: [(Token?) -> Void]` property storing the closures to be called once the token is provided by the **TokenProvider** 

_If the token has already been provided it is given back to **RequestEncoder** synchronously_

### DIFF
--- a/DemoApp/ChatPresenter.swift
+++ b/DemoApp/ChatPresenter.swift
@@ -1,5 +1,5 @@
 //
-// Copyright © 2020 Stream.io Inc. All rights reserved.
+// Copyright © 2021 Stream.io Inc. All rights reserved.
 //
 
 import StreamChat
@@ -10,24 +10,14 @@ extension UIViewController {
     // TODO: Where to put this???
     func presentChat(userCredentials: UserCredentials) {
         LogConfig.level = .error
+
+        // Create a token
+        let token = try! Token(rawValue: userCredentials.token)
         
         // Create client
         let config = ChatClientConfig(apiKey: .init(userCredentials.apiKey))
-        let client = ChatClient(config: config)
-        
-        // Log in the current user
-        let currentUserController = client.currentUserController()
-        currentUserController.setUser(
-            userId: userCredentials.id,
-            name: userCredentials.name,
-            imageURL: userCredentials.avatarURL,
-            token: userCredentials.token
-        ) { error in
-            if let error = error {
-                print("User login failed: \(error)")
-            }
-        }
-        
+        let client = ChatClient(config: config, tokenProvider: .static(token))
+
         // Config
         var uiConfig = UIConfig<DefaultExtraData>()
         uiConfig.navigation.channelListRouter = DemoChatChannelListRouter.self
--- a/Sources_v3/StreamChat/APIClient/Endpoints/Endpoint.swift
+++ b/Sources_v3/StreamChat/APIClient/Endpoints/Endpoint.swift
@@ -9,7 +9,24 @@ struct Endpoint<ResponseType: Decodable> {
     let method: EndpointMethod
     let queryItems: Encodable?
     let requiresConnectionId: Bool
+    let requiresToken: Bool
     let body: Encodable?
+
+    init(
+        path: String,
+        method: EndpointMethod,
+        queryItems: Encodable? = nil,
+        requiresConnectionId: Bool = false,
+        requiresToken: Bool = true,
+        body: Encodable? = nil
+    ) {
+        self.path = path
+        self.method = method
+        self.queryItems = queryItems
+        self.requiresConnectionId = requiresConnectionId
+        self.requiresToken = requiresToken
+        self.body = body
+    }
 }
 
 enum EndpointMethod: String {
--- a/Sources_v3/StreamChat/APIClient/Endpoints/WebSocketConnectEndpoint.swift
+++ b/Sources_v3/StreamChat/APIClient/Endpoints/WebSocketConnectEndpoint.swift
@@ -1,29 +1,19 @@
 //
-// Copyright © 2020 Stream.io Inc. All rights reserved.
+// Copyright © 2021 Stream.io Inc. All rights reserved.
 //
 
 import Foundation
 
 extension Endpoint {
-    static func webSocketConnect<UserData: UserExtraData>(
-        userId: UserId,
-        name: String?,
-        imageURL: URL?,
-        role: UserRole = .user,
-        extraData: UserData? = nil
-    ) -> Endpoint<EmptyResponse> {
+    static func webSocketConnect(userId: UserId) -> Endpoint<EmptyResponse> {
         .init(
             path: "connect",
             method: .get,
             queryItems: nil,
             requiresConnectionId: false,
-            body: ["json": WebSocketConnectPayload(
-                userId: userId,
-                name: name,
-                imageURL: imageURL,
-                userRole: role,
-                extraData: extraData
-            )]
+            body: [
+                "json": WebSocketConnectPayload(userId: userId)
+            ]
         )
     }
 }
--- a/Sources_v3/StreamChat/APIClient/RequestEncoder.swift
+++ b/Sources_v3/StreamChat/APIClient/RequestEncoder.swift
@@ -81,46 +81,22 @@ struct DefaultRequestEncoder: RequestEncoder {
             }
             
             try encodeRequestBody(request: &request, endpoint: endpoint)
-            
-            // Add Stream auth headers
-            var headers = request.allHTTPHeaderFields ?? [:]
-            if let token = connectionDetailsProviderDelegate?.provideToken() {
-                headers["Stream-Auth-Type"] = "jwt"
-                headers["Authorization"] = token
-            } else {
-                headers["Stream-Auth-Type"] = "anonymous"
-            }
-            request.allHTTPHeaderFields = headers
-            
         } catch {
             completion(.failure(error))
             return
         }
-        
-        if endpoint.requiresConnectionId {
-            log.assert(
-                connectionDetailsProviderDelegate != nil,
-                "The endpoind requiers `connectionId` but `connectionDetailsProviderDelegate` is not set."
-            )
-            
-            connectionDetailsProviderDelegate?.provideConnectionId { (connectionId) in
-                guard let connectionId = connectionId else {
-                    completion(.failure(ClientError.MissingConnectionId("Failed to get `connectionId`, request can't be created.")))
-                    return
-                }
-                
-                do {
-                    request.url = try request.url?.appendingQueryItems(["connection_id": connectionId])
-                } catch {
-                    completion(.failure(error))
-                    return
-                }
-                
-                completion(.success(request))
+
+        addAuthorizationHeader(request: request, endpoint: endpoint) {
+            switch $0 {
+            case let .success(requestWithAuth):
+                self.addConnectionIdIfNeeded(
+                    request: requestWithAuth,
+                    endpoint: endpoint,
+                    completion: completion
+                )
+            case let .failure(error):
+                completion(.failure(error))
             }
-            
-        } else {
-            completion(.success(request))
         }
     }
     
@@ -130,6 +106,71 @@ struct DefaultRequestEncoder: RequestEncoder {
     }
     
     // MARK: - Private
+
+    private func addAuthorizationHeader<T: Decodable>(
+        request: URLRequest,
+        endpoint: Endpoint<T>,
+        completion: @escaping (Result<URLRequest, Error>) -> Void
+    ) {
+        guard endpoint.requiresToken else {
+            var updatedRequest = request
+            updatedRequest.setHTTPHeaders(.anonymousStreamAuth)
+            completion(.success(updatedRequest))
+            return
+        }
+
+        log.assert(
+            connectionDetailsProviderDelegate != nil,
+            "The endpoint requires `token` but `connectionDetailsProviderDelegate` is not set."
+        )
+
+        connectionDetailsProviderDelegate?.provideToken {
+            if let token = $0 {
+                var updatedRequest = request
+
+                if token.userId.isAnonymousUser {
+                    updatedRequest.setHTTPHeaders(.anonymousStreamAuth)
+                } else {
+                    updatedRequest.setHTTPHeaders(.jwtStreamAuth, .authorization(token.rawValue))
+                }
+
+                completion(.success(updatedRequest))
+            } else {
+                let error = ClientError.MissingToken("Failed to get `token`, request can't be created.")
+                completion(.failure(error))
+            }
+        }
+    }
+
+    private func addConnectionIdIfNeeded<T: Decodable>(
+        request: URLRequest,
+        endpoint: Endpoint<T>,
+        completion: @escaping (Result<URLRequest, Error>) -> Void
+    ) {
+        guard endpoint.requiresConnectionId else {
+            completion(.success(request))
+            return
+        }
+
+        log.assert(
+            connectionDetailsProviderDelegate != nil,
+            "The endpoint requires `connectionId` but `connectionDetailsProviderDelegate` is not set."
+        )
+
+        connectionDetailsProviderDelegate?.provideConnectionId {
+            do {
+                if let connectionId = $0 {
+                    var updatedRequest = request
+                    updatedRequest.url = try updatedRequest.url?.appendingQueryItems(["connection_id": connectionId])
+                    completion(.success(updatedRequest))
+                } else {
+                    throw ClientError.MissingConnectionId("Failed to get `connectionId`, request can't be created.")
+                }
+            } catch {
+                completion(.failure(error))
+            }
+        }
+    }
     
     private func encodeRequestURL<T: Decodable>(for endpoint: Endpoint<T>) throws -> URL {
         var urlComponents = URLComponents()
@@ -207,7 +248,7 @@ private extension URL {
 
 protocol ConnectionDetailsProviderDelegate: AnyObject {
     func provideConnectionId(completion: @escaping (_ connectionId: ConnectionId?) -> Void)
-    func provideToken() -> Token?
+    func provideToken(completion: @escaping (Token?) -> Void)
 }
 
 extension ClientError {
--- a/Sources_v3/StreamChat/ChatClient.swift
+++ b/Sources_v3/StreamChat/ChatClient.swift
@@ -283,30 +283,31 @@ public class _ChatClient<ExtraData: ExtraDataTypes> {
         self.environment = environment
         self.workerBuilders = workerBuilders
         self.eventWorkerBuilders = eventWorkerBuilders
+
+        currentUserId = fetchCurrentUserIdFromDatabase()
+
+        let updater = environment.clientUpdaterBuilder(self)
+        updater.reloadUserIfNeeded(completion: completion)
     }
     
     deinit {
         completeConnectionIdWaiters(connectionId: nil)
         completeTokenWaiters(token: nil)
     }
-    
-    // TODO: Not used & tested yet -> CIS-224
-    func refreshToken(completion: @escaping (Error?) -> Void) {
-        log.assert(config.tokenProvider != nil, "You can't call `refreshToken` when `Config.tokenProvider` is nil.")
 
-        guard let currentUserId = currentUserId else {
-            completion(ClientError.CurrentUserDoesNotExist())
-            return
-        }
-        
-        config.tokenProvider?(config.apiKey, currentUserId, { [weak self] (token) in
-            guard let token = token else {
-                completion(ClientError.MissingToken("Can't connect because `TokenProvider` didn't return a valid token."))
-                return
+    func fetchCurrentUserIdFromDatabase() -> UserId? {
+        var currentUserId: UserId?
+
+        let context = databaseContainer.viewContext
+        if Thread.isMainThread {
+            currentUserId = context.currentUser()?.user.id
+        } else {
+            context.performAndWait {
+                currentUserId = context.currentUser()?.user.id
             }
-            self?.currentToken = token
-            completion(nil)
-        })
+        }
+
+        return currentUserId
     }
     
     func createBackgroundWorkers() {
@@ -373,6 +374,8 @@ extension _ChatClient {
         var notificationCenterBuilder: ([EventMiddleware]) -> EventNotificationCenter = EventNotificationCenter.init
         
         var internetConnection: () -> InternetConnection = { InternetConnection() }
+
+        var clientUpdaterBuilder = ChatClientUpdater<ExtraData>.init
     }
 }
 
--- a/Sources_v3/StreamChat/ChatClient.swift
+++ b/Sources_v3/StreamChat/ChatClient.swift
@@ -213,12 +213,20 @@ public class _ChatClient<ExtraData: ExtraDataTypes> {
     
     /// The token of the current user. If the current user is anonymous, the token is `nil`.
     @Atomic var currentToken: Token?
+
+    public var tokenProvider: _TokenProvider<ExtraData>
     
     /// Creates a new instance of `ChatClient`.
-    ///
-    /// - Parameter config: The config object for the `Client`. See `ChatClientConfig` for all configuration options.
-    ///
-    public convenience init(config: ChatClientConfig) {
+    /// - Parameters:
+    ///   - config: The config object for the `Client`. See `ChatClientConfig` for all configuration options.
+    ///   - tokenProvider: The `_TokenProvider<ExtraData>` instance that incapsulates the logic of obtaining a JWT
+    ///   token used to communicate with REST-API.
+    ///   - completion: The completion that will be called once the **first** user session for the given token is setup.
+    public convenience init(
+        config: ChatClientConfig,
+        tokenProvider: _TokenProvider<ExtraData>,
+        completion: ((Error?) -> Void)? = nil
+    ) {
         let workerBuilders: [WorkerBuilder]
         let eventWorkerBuilders: [EventWorkerBuilder]
         var environment = Environment()
@@ -246,9 +254,11 @@ public class _ChatClient<ExtraData: ExtraDataTypes> {
         
         self.init(
             config: config,
+            tokenProvider: tokenProvider,
             workerBuilders: workerBuilders,
             eventWorkerBuilders: eventWorkerBuilders,
-            environment: environment
+            environment: environment,
+            completion: completion
         )
     }
     
@@ -262,11 +272,14 @@ public class _ChatClient<ExtraData: ExtraDataTypes> {
     ///
     init(
         config: ChatClientConfig,
+        tokenProvider: _TokenProvider<ExtraData>,
         workerBuilders: [WorkerBuilder],
         eventWorkerBuilders: [EventWorkerBuilder],
-        environment: Environment
+        environment: Environment,
+        completion: ((Error?) -> Void)? = nil
     ) {
         self.config = config
+        self.tokenProvider = tokenProvider
         self.environment = environment
         self.workerBuilders = workerBuilders
         self.eventWorkerBuilders = eventWorkerBuilders
--- a/Sources_v3/StreamChat/ChatClient.swift
+++ b/Sources_v3/StreamChat/ChatClient.swift
@@ -135,6 +135,10 @@ public class _ChatClient<ExtraData: ExtraDataTypes> {
             eventNotificationCenter,
             internetConnection
         )
+
+        if let currentUserId = currentUserId {
+            webSocketClient?.connectEndpoint = .webSocketConnect(userId: currentUserId)
+        }
         
         webSocketClient?.connectionStateDelegate = self
         
--- a/Sources_v3/StreamChat/ChatClient.swift
+++ b/Sources_v3/StreamChat/ChatClient.swift
@@ -76,12 +76,8 @@ public class _ChatClient<ExtraData: ExtraDataTypes> {
     /// To observe changes in the connection status, create an instance of `CurrentChatUserController`, and use it to receive
     /// callbacks when the connection status changes.
     ///
-    public var connectionStatus: ConnectionStatus {
-        ConnectionStatus(
-            webSocketConnectionState: webSocketClient?.connectionState ?? .disconnected(error: .ClientIsNotInActiveMode())
-        )
-    }
-    
+    public internal(set) var connectionStatus: ConnectionStatus = .initialized
+
     /// The config object of the `ChatClient` instance.
     ///
     /// This value can't be mutated and can only be set when initializing a new `ChatClient` instance.
@@ -397,6 +393,8 @@ extension ClientError {
 /// its `RequestEncoder`.
 extension _ChatClient: ConnectionStateDelegate {
     func webSocketClient(_ client: WebSocketClient, didUpdateConectionState state: WebSocketConnectionState) {
+        connectionStatus = .init(webSocketConnectionState: state)
+
         _connectionId.mutate { mutableConnectionId in
             _connectionIdWaiters.mutate { connectionIdWaiters in
                 
--- a/Sources_v3/StreamChat/ChatClient.swift
+++ b/Sources_v3/StreamChat/ChatClient.swift
@@ -211,6 +211,9 @@ public class _ChatClient<ExtraData: ExtraDataTypes> {
     
     /// An array of requests waiting for the connection id
     @Atomic var connectionIdWaiters: [(String?) -> Void] = []
+
+    /// An array of requests waiting for the token
+    @Atomic var tokenWaiters: [(Token?) -> Void] = []
     
     /// The token of the current user. If the current user is anonymous, the token is `nil`.
     @Atomic var currentToken: Token?
@@ -277,10 +280,8 @@ public class _ChatClient<ExtraData: ExtraDataTypes> {
     }
     
     deinit {
-        _connectionIdWaiters.mutate { waiters in
-            waiters.forEach { $0(nil) }
-            waiters.removeAll()
-        }
+        completeConnectionIdWaiters(connectionId: nil)
+        completeTokenWaiters(token: nil)
     }
     
     // TODO: Not used & tested yet -> CIS-224
@@ -309,12 +310,19 @@ public class _ChatClient<ExtraData: ExtraDataTypes> {
             builder(self.databaseContainer, self.eventNotificationCenter, self.apiClient)
         }
     }
-    
-    /// Before CIS-555, `webSocketClient` was initialized in `createBackgroundWorkers` in `init` since it was passed
-    /// to the background workers. Now, background workers do not access `webSocketClient` and it's not being
-    /// initialized in `init`, so we initialize it here explicitly.
-    private func createWebSocketClient() {
-        _ = webSocketClient
+
+    func completeConnectionIdWaiters(connectionId: String?) {
+        _connectionIdWaiters.mutate { waiters in
+            waiters.forEach { $0(connectionId) }
+            waiters.removeAll()
+        }
+    }
+
+    func completeTokenWaiters(token: Token?) {
+        _tokenWaiters.mutate { waiters in
+            waiters.forEach { $0(token) }
+            waiters.removeAll()
+        }
     }
 }
 
@@ -413,8 +421,14 @@ extension _ChatClient: ConnectionStateDelegate {
 
 /// `Client` provides connection details for the `RequestEncoder`s it creates.
 extension _ChatClient: ConnectionDetailsProviderDelegate {
-    func provideToken() -> Token? {
-        currentToken
+    func provideToken(completion: @escaping (_ token: Token?) -> Void) {
+        if let token = currentToken {
+            completion(token)
+        } else {
+            _tokenWaiters.mutate {
+                $0.append(completion)
+            }
+        }
     }
     
     func provideConnectionId(completion: @escaping (String?) -> Void) {
--- a/Sources_v3/StreamChat/ChatClient.swift
+++ b/Sources_v3/StreamChat/ChatClient.swift
@@ -88,7 +88,7 @@ public class _ChatClient<ExtraData: ExtraDataTypes> {
     ///
     /// `ChatClient` initializes a set of background workers that keep observing the current state of the system and perform
     /// work if needed (i.e. when a new message pending sent appears in the database, a worker tries to send it.)
-    private(set) var backgroundWorkers: [Worker]!
+    private(set) var backgroundWorkers: [Worker] = []
     
     /// Builder blocks used for creating `backgroundWorker`s when needed.
     private let workerBuilders: [WorkerBuilder]
@@ -270,9 +270,6 @@ public class _ChatClient<ExtraData: ExtraDataTypes> {
         self.environment = environment
         self.workerBuilders = workerBuilders
         self.eventWorkerBuilders = eventWorkerBuilders
-        
-        createBackgroundWorkers()
-        createWebSocketClient()
     }
     
     deinit {
--- a/Sources_v3/StreamChat/ChatClient_Tests.swift
+++ b/Sources_v3/StreamChat/ChatClient_Tests.swift
@@ -33,6 +33,7 @@ class ChatClient_Tests: StressTestCase {
     
     override func tearDown() {
         testEnv.apiClient?.cleanUp()
+        testEnv.clientUpdater?.cleanUp()
         AssertAsync.canBeReleased(&testEnv)
         super.tearDown()
     }
@@ -63,6 +64,7 @@ class ChatClient_Tests: StressTestCase {
         
         // Create env object with custom database builder
         var env = ChatClient.Environment()
+        env.clientUpdaterBuilder = ChatClientUpdaterMock.init
         env.databaseContainerBuilder = { kind, shouldFlushOnStart, shouldResetEphemeralValuesOnStart in
             usedDatabaseKind = kind
             shouldFlushDBOnStart = shouldFlushOnStart
@@ -96,6 +98,7 @@ class ChatClient_Tests: StressTestCase {
         
         // Create env object with custom database builder
         var env = ChatClient.Environment()
+        env.clientUpdaterBuilder = ChatClientUpdaterMock.init
         env.databaseContainerBuilder = { kind, _, _ in
             usedDatabaseKind = kind
             return DatabaseContainerMock()
@@ -131,6 +134,7 @@ class ChatClient_Tests: StressTestCase {
 
         // Create env object and store all `kinds it's called with.
         var env = ChatClient.Environment()
+        env.clientUpdaterBuilder = ChatClientUpdaterMock.init
         env.databaseContainerBuilder = { kind, _, _ in
             usedDatabaseKinds.append(kind)
             // Return error for the first time
@@ -251,7 +255,9 @@ class ChatClient_Tests: StressTestCase {
             }
         )
 
-        // Assert `init` completion hasn't been called yet.
+        // Assert `init` completion is called synchronously since there is no
+        // async job to be done (we don't try to establish and wait for connection since
+        // `shouldConnectAutomatically` == false).
         XCTAssertTrue(initCompletionCalled)
         // Assert connection status is `.initialized`
         XCTAssertEqual(client.connectionStatus, .initialized)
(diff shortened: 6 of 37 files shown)


Sources_v3/StreamChat/APIClient/Endpoints/Endpoint.swift
```diff
@@ -9,7 +9,24 @@ struct Endpoint<ResponseType: Decodable> {
...
+        requiresToken: Bool = true,
+        body: Encodable? = nil
+    ) {
+        self.path = path
+        self.method = method
+        self.queryItems = queryItems
+        self.requiresConnectionId = requiresConnectionId
+        self.requiresToken = requiresToken
+        self.body = body
+    }
```
evsaev, line 29: The init with default params is just a temp trick to reduce # of changes in the PR. Not sure it's the best way to go

Sources_v3/StreamChat/APIClient/Endpoints/WebSocketConnectEndpoint.swift
```diff
@@ -1,15 +1,15 @@
...
     static func webSocketConnect<UserData: UserExtraData>(
         userId: UserId,
-        name: String?,
-        imageURL: URL?,
-        role: UserRole = .user,
+        name: String? = nil,
+        imageURL: URL? = nil,
+        role: UserRole? = nil,
         extraData: UserData? = nil
     ) -> Endpoint<EmptyResponse> {
```
evsaev, line 14: While working on this PR learned that passing `name/imageURL/role/extraData` when connecting to the web-socket is **not required** and is just a **shorthand** for updating the user data before the connection is established OR creating a new user with the provided values, not the default (nil?) ones.

Sources_v3/StreamChat/ChatClient.swift
```diff
@@ -69,19 +69,15 @@ public typealias ChatClient = _ChatClient<DefaultExtraData>
...
     /// callbacks when the connection status changes.
     ///
-    public var connectionStatus: ConnectionStatus {
-        ConnectionStatus(
-            webSocketConnectionState: webSocketClient?.connectionState ?? .disconnected(error: .ClientIsNotInActiveMode())
-        )
-    }
-    
+    public private(set) var connectionStatus: ConnectionStatus = .initialized
+
```
evsaev, line 80: There was a need to have some state the **ChatClient** stays in after instantiation until the `reconnectIfNeeded` is invoked.

DemoApp/ChatPresenter.swift
```diff
@@ -13,16 +13,8 @@ extension UIViewController {
...
-        // Log in the current user
-        let currentUserController = client.currentUserController()
-        currentUserController.setUser(
-            userId: userCredentials.id,
-            name: userCredentials.name,
-            imageURL: userCredentials.avatarURL,
-            token: userCredentials.token
-        ) { error in
+        let client = ChatClient(config: config, tokenProvider: .hardCoded(userCredentials.token))
+        client.reconnectIfNeeded { error in
```
b-onc, line 17: Why don't we just `connect` since this is the initial connection part?

Sources_v3/StreamChat/APIClient/Endpoints/WebSocketConnectEndpoint.swift
b-onc: it saves an `updateUser` call but at what cost... We can remove these params from this func

Sources_v3/StreamChat/APIClient/RequestEncoder.swift
```diff
@@ -81,46 +81,25 @@ struct DefaultRequestEncoder: RequestEncoder {
...
+        addAuthorizationHeader(request: request, endpoint: endpoint) { authResult in
+            switch authResult {
+            case let .success(requestWithAuth):
+                self.addConnectionIdIfNeeded(request: requestWithAuth, endpoint: endpoint) { connectionResult in
+                    switch connectionResult {
+                    case let .success(requestWithAuthAndConnectionId):
+                        completion(.success(requestWithAuthAndConnectionId))
+                    case let .failure(error):
+                        completion(.failure(error))
+                    }
```
b-onc, line 98:
```suggestion
completion(connectionResult)
```

b-onc: I wonder if there's a solution with `map`

Sources_v3/StreamChat/ChatClient.swift
```diff
@@ -69,19 +69,15 @@ public typealias ChatClient = _ChatClient<DefaultExtraData>
 ///
 public class _ChatClient<ExtraData: ExtraDataTypes> {
     /// The `UserId` of the currently logged in user.
-    @Atomic public var currentUserId: UserId = .anonymous
-    
+    @Atomic public var currentUserId: UserId?
```
b-onc, line 72: Why don't we / are we going to:
- Parse the last logged user id from DB (I think we're going to do this anyway)
- If there's no last logged in user, currentUserId is `anonymous`
so that we don't have to deal with optional here?

Sources_v3/StreamChat/ChatClient.swift
```diff
@@ -271,54 +271,195 @@ public class _ChatClient<ExtraData: ExtraDataTypes> {
...
     func createBackgroundWorkers() {
         backgroundWorkers = workerBuilders.map { builder in
             builder(self.databaseContainer, self.apiClient)
         } + eventWorkerBuilders.map { builder in
             builder(self.databaseContainer, self.eventNotificationCenter, self.apiClient)
         }
     }
-    
+
+    func removeConnectionWaitersPassing(_ connectionId: String?) {
```
b-onc, line 307: This and its `tokenWaiters` friend's naming confused me, maybe `removeConnectionIdWaiters(passing value: nil)`?

Sources_v3/StreamChat/ChatClient.swift
```diff
@@ -271,54 +271,195 @@ public class _ChatClient<ExtraData: ExtraDataTypes> {
...
+
+extension _ChatClient {
+    /// Connects the chat client the controller represents to the chat servers.
+    ///
+    /// When the connection is established, `ChatClient` starts receiving chat updates, and `currentUser` variable is available.
+    ///
+    /// - Parameter completion: Called when the connection is established. If the connection fails, the completion is
+    /// called with an error.
+    ///
+    func connect(completion: ((Error?) -> Void)? = nil) {
```
b-onc, line 391: Shouldn't this also refresh token? Feels like actual connection code resides in `reconnectIfNeeded` and this is a helper method

Sources_v3/StreamChat/ChatClient.swift
```diff
@@ -394,6 +533,8 @@ extension ClientError {
 /// its `RequestEncoder`.
 extension _ChatClient: ConnectionStateDelegate {
     func webSocketClient(_ client: WebSocketClient, didUpdateConectionState state: WebSocketConnectionState) {
+        connectionStatus = .init(webSocketConnectionState: state)
```
b-onc, line 536: This will never run for passive clients so we'll always have `initializing` state, right?

Sources_v3/StreamChat/Config/Token.swift
```diff
@@ -1,40 +1,41 @@
...
+    case missingUserId
+}
 
-/// A token provider is a function in which you send a request to your own backend to get a Stream Chat API token.
-///
-/// Set your custom `TokenProvider` in `ChatClientConfig` when creating a `ChatClient` instance. `ChatClient` will use it whenever
-/// it needs to get a new token for the given user.
-///
-public typealias TokenProvider = (_ apiKey: APIKey, _ userId: UserId, _ completion: @escaping (Token?) -> Void) -> Void
+public struct Token: Decodable, Equatable {
```
b-onc, line 11: Are we going to handle expiring tokens in a different PR?

Sources_v3/StreamChat/APIClient/Endpoints/WebSocketConnectEndpoint.swift
evsaev: Yeah, was gonna preserve passing user data with the request but has a word with Vojta and we decided to connect with just a token. Maybe @VojtaStavik has smth to add here

DemoApp/ChatPresenter.swift
VojtaStavik: There's no initial connection guys, it happens implicitly ⚠️ This name should be `reloadUserIfNeeded` or something like that. Managing connection manually is an advanced topic, we shouldn't even use anything related to that in naming, so it's not confusing.

Sources_v3/StreamChat/APIClient/Endpoints/Endpoint.swift
VojtaStavik: I think it's actually a good approach with the default param. Only a very few endpoints don't require a token.

Sources_v3/StreamChat/APIClient/Endpoints/WebSocketConnectEndpoint.swift
VojtaStavik: Yes please, let's remove everything but the user id.

Sources_v3/StreamChat/ChatClient.swift
VojtaStavik: I think this approach is more correct and @evsaev's explanation of this in the description makes sense. I like the change.

Sources_v3/StreamChat/ChatClient.swift
VojtaStavik: The client should start connection immediately unless it's not specified in config otherwise ⚠️

DemoApp/ChatPresenter.swift
evsaev: Currently `connect/disconnect` API is occupied for web-socket actions but we can use it here for sure.

Named it `reconnectIfNeeded` to highight it can be safely called for a couple of times and none unnecessary job will be done (e.g. re-establish the connection) if the current user is not changed.

It is also expected to be used by **passive** clients but but `passiveClient.connect` feels too active 😄

Sources_v3/StreamChat/ChatClient.swift
```diff
@@ -124,26 +120,24 @@ public class _ChatClient<ExtraData: ExtraDataTypes> {
...
             eventNotificationCenter,
             internetConnection
         )
+
+        if let currentUserId = currentUserId {
+            webSocketClient?.connectEndpoint = .webSocketConnect(
+                userId: currentUserId,
+                extraData: nil as ExtraData.User?
+            )
+        }
```
VojtaStavik, line 140: This has to be handled differently. We shouldn't even try to connect when we don't have user id.

Sources_v3/StreamChat/APIClient/RequestEncoder.swift
evsaev: Hm, you mean chaining request building steps or some hashmap usage? 🤔

Sources_v3/StreamChat/ChatClient.swift
VojtaStavik: +1 on confusing naming. Why not just: `completeConnectionIdWaiters(value: String?)`? It work nicely with both an actual value and `nil`, too.

Sources_v3/StreamChat/ChatClient.swift
evsaev: Please take a look at the `Making the current user ID optional` section in the PR description, I gave some reasoning for making current user id non-optional.

Sources_v3/StreamChat/ChatClient.swift
evsaev: Great idea, will fix it

Sources_v3/StreamChat/ChatClient.swift
```diff
@@ -271,54 +271,195 @@ public class _ChatClient<ExtraData: ExtraDataTypes> {
...
+        _connectionIdWaiters.mutate {
+            $0.removeAll()
+        }
+    }
+
+    func removeTokenWaiters() {
+        _tokenWaiters.mutate {
+            $0.removeAll()
+        }
+    }
```
VojtaStavik, line 331: This is not correct IMO. We can't just remove the waiters. We always must call something on them.

Sources_v3/StreamChat/ChatClient.swift
evsaev: Yep. Does it make sense to you or ..? 😄

Sources_v3/StreamChat/ChatClient.swift
```diff
@@ -271,54 +271,195 @@ public class _ChatClient<ExtraData: ExtraDataTypes> {
...
+        createBackgroundWorkers()
+
+        // Cancel all API requests since they are related to the previous user.
+        removeTokenWaitersPassing(nil)
+
+        // Update the current token with the new one.
+        _currentToken.mutate { $0 = newToken }
+
+        // Update the current user id to the new one.
+        _currentUserId.mutate { $0 = newToken.userId }
```
VojtaStavik, line 371: You don't need to do this if the wrapped type is not a collection. `currentToken = newToken` and `currentUserId = newToken.userId` are equally thread-safe to this.

Sources_v3/StreamChat/ChatClient.swift
VojtaStavik: Good point @b-onc. We should probably check for the config here and set the value to `notConnected` instead of `initialized`.

Sources_v3/StreamChat/Config/Token.swift
VojtaStavik: @b-onc there's even a JIRA ticket for that already :)

evsaev: Would be nice it have it done inside this PR.

@VojtaStavik wdyt?

Sources_v3/StreamChat/Config/Token.swift
```diff
@@ -1,40 +1,41 @@
...
-            return true
+    init(rawValueWithUserID: String) throws {
+        guard let userId = rawValueWithUserID.jwtPayload?["user_id"] as? String else {
+            throw TokenError.missingUserId
         }
-        
-        return !userId.isEmpty && (payload?["user_id"] as? String) == userId
+
+        self.init(rawValue: rawValueWithUserID, userId: userId)
     }
```
VojtaStavik, line 21: nit 1: this should be public
nit 2: let's just name it `init(rawValue:)`, the other option is an implementation detail and shouldn't affect the public API in any way.

Sources_v3/StreamChat/Config/Token.swift
```diff
@@ -1,40 +1,41 @@
...
+// Copyright © 2021 Stream.io Inc. All rights reserved.
 //
 
 import Foundation
 
-/// A token is used to authenticate a user.
-public typealias Token = String
+enum TokenError: Error {
+    case missingUserId
+}
```
VojtaStavik, line 9: nit: Please use the `ClientError` approach we use for all other errors.

VojtaStavik (commented):
There's a lot happening but it looks great 🎉

Sources_v3/StreamChat/ChatClient.swift
evsaev: fixed

Sources_v3/StreamChat/ChatClient.swift
evsaev: fixed

Sources_v3/StreamChat/ChatClient.swift
evsaev: fixed

DemoApp/ChatPresenter.swift
```diff
@@ -13,16 +13,8 @@ extension UIViewController {
...
-        
-        // Log in the current user
-        let currentUserController = client.currentUserController()
-        currentUserController.setUser(
-            userId: userCredentials.id,
-            name: userCredentials.name,
-            imageURL: userCredentials.avatarURL,
-            token: userCredentials.token
-        ) { error in
+        let client = ChatClient(config: config, tokenProvider: .hardCoded(userCredentials.token))
```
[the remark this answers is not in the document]
evsaev: renamed

Sources_v3/StreamChat/ChatClient.swift
evsaev: moved to `prepareEnvironment` where user id is known

Sources_v3/StreamChat/ChatClient.swift
evsaev: now we call `completeConnectionId/TokenWaiters` instead

Sources_v3/StreamChat/ChatClient.swift
evsaev: fixed

Sources_v3/StreamChat/Config/Token.swift
evsaev: done

Sources_v3/StreamChat/Config/Token.swift
evsaev: done

Sources_v3/StreamChat/ChatClient.swift
```diff
@@ -271,54 +264,214 @@ public class _ChatClient<ExtraData: ExtraDataTypes> {
     ///
     init(
         config: ChatClientConfig,
+        tokenProvider: _TokenProvider<ExtraData>,
         workerBuilders: [WorkerBuilder],
         eventWorkerBuilders: [EventWorkerBuilder],
         environment: Environment
     ) {
+        log.assert(Thread.isMainThread, "ChatClient.init must be called on main thread.")
+
```
VojtaStavik, line 273: Is this really a condition we need?

Sources_v3/StreamChat/ChatClient.swift
```diff
@@ -271,54 +264,214 @@ public class _ChatClient<ExtraData: ExtraDataTypes> {
...
+        guard config.isClientInActiveMode else {
+            completion(ClientError.ClientIsNotInActiveMode())
+            return
+        }
+
+        // Cancel all API requests since they are related to the previous user.
+        completeTokenWaiters(passing: nil)
+
+        // Re-create backgroundWorker's since they are related to the previous user.
+        createBackgroundWorkers()
```
VojtaStavik, line 351: You also create background workers on L334

Sources_v3/StreamChat/ChatClient.swift
```diff
@@ -271,54 +264,214 @@ public class _ChatClient<ExtraData: ExtraDataTypes> {
...
+        }
+
+        // Setting a new user is not possible in connectionless mode.
+        guard config.isClientInActiveMode else {
+            completion(ClientError.ClientIsNotInActiveMode())
+            return
+        }
+
+        // Cancel all API requests since they are related to the previous user.
+        completeTokenWaiters(passing: nil)
```
VojtaStavik, line 348: This needs to be done for passive client, too

Sources_v3/StreamChat/ChatClient.swift
```diff
@@ -271,54 +264,214 @@ public class _ChatClient<ExtraData: ExtraDataTypes> {
...
     /// to the background workers. Now, background workers do not access `webSocketClient` and it's not being
     /// initialized in `init`, so we initialize it here explicitly.
-    private func createWebSocketClient() {
+    func createWebSocketClient() {
         _ = webSocketClient
     }
+
+    func prepareEnvironment(newToken: Token, completion: @escaping (Error?) -> Void) {
+        // Update web-socket endpoint.
+        webSocketClient?.connectEndpoint = .webSocketConnect(userId: newToken.userId)
```
VojtaStavik, line 323: This doesn't need to be done if the new user id is same as the current one.

Sources_v3/StreamChat/ChatClient.swift
evsaev: That's because of
```
currentUserId = databaseContainer.viewContext.currentUser()?.user.id
```
Checking the thread allows us to load the current user id synchronously since we are sure the thread is `main` we can access `viewContext` safely. Without this assertation, we have to load the current user via bg context asynchronously.

Why would some need to instantiate **ChatClient** on the background thread? 🤔

Sources_v3/StreamChat/ChatClient.swift
evsaev: This does need to be done (at least in current PR shape) since we removed the 
```swift
if let currentUserId = currentUserId {
 webSocketClient?.connectEndpoint = .webSocketConnect(userId: currentUserId)
}
```
from `lazy var webScoketClient` so initially `webScoketClient.connectEndpoint == nil` and has to be set somewhere.

_In `webSocketClient` we now have `createOrReuseEngine` which builds the new engine only if the request is different (e.g. the `user_id` is changed) so we can set the endpoint multiple times without extra job._

Sources_v3/StreamChat/ChatClient.swift
evsaev: Indeed! 👍 

Will also add forwarding the updated token. Someone might call the API right after the client is instantiated and the request will be waiting for the token:
```swift
guard newToken.userId != currentUserId else {
 ...
 // Forward the new token to waiting requests.
 completeTokenWaiters(passing: newToken)
 ...
}
```

Sources_v3/StreamChat/ChatClient.swift
evsaev: To me setting the endpoint when web-socket is created and resetting if the user has changed makes more sense.

Sources_v3/StreamChat/ChatClient.swift
VojtaStavik: Great 👏

Sources_v3/StreamChat/ChatClient.swift
VojtaStavik: It's quite a common practice to speed up app launch to move dependency instantiation you don't need immediately to a background thread. What about moving the handling of this situation to our side with:
```swift
        var existingUserId: UserId?
        
        if Thread.isMainThread {
            existingUserId = databaseContainer.viewContext.currentUser()?.user.id
        } else {
            databaseContainer.viewContext.performAndWait { [context = databaseContainer.viewContext] in
                existingUserId = context.currentUser()?.user.id
            }
        }
        
        print(existingUserId)
```

Sources_v3/StreamChat/ChatClient.swift
```diff
@@ -219,15 +210,26 @@ public class _ChatClient<ExtraData: ExtraDataTypes> {
...
-    public convenience init(config: ChatClientConfig) {
+    /// - Parameters:
+    ///   - config: The config object for the `Client`. See `ChatClientConfig` for all configuration options.
+    ///   - tokenProvider: The `_TokenProvider<ExtraData>` instance that incapsulates the logic of obtaining a JWT
+    ///   token used to communicate with REST-API.
+    ///   - didReloadUser: The closure that will be called once the **first** session for the token is obtained and user session is setup.
+    public convenience init(
+        config: ChatClientConfig,
+        tokenProvider: _TokenProvider<ExtraData>,
+        didReloadUser: @escaping (Error?) -> Void = { _ in }
```
VojtaStavik, line 231: Nice idea. Can we maybe rename it just to `completion` and also make it optional? We will call it once we're done with set up.

Sources_v3/StreamChat/ChatClient.swift
```diff
@@ -271,54 +275,231 @@ public class _ChatClient<ExtraData: ExtraDataTypes> {
...
         self.config = config
+        self.tokenProvider = tokenProvider
         self.environment = environment
         self.workerBuilders = workerBuilders
         self.eventWorkerBuilders = eventWorkerBuilders
-        
-        createBackgroundWorkers()
-        createWebSocketClient()
+
+        loadCurrentUserFromDatabase { userId in
```
VojtaStavik, line 290: `loadCurrentUserFromDatabase` -> `fetchCurrentUserIdFromDB`?

Sources_v3/StreamChat/ChatClient.swift
```diff
@@ -271,54 +275,231 @@ public class _ChatClient<ExtraData: ExtraDataTypes> {
...
+            self.currentUserId = userId
+            self.createWebSocketClient()
```
VojtaStavik, line 292: In this form, calling this on a different queue is not safe. `lazy` variable intialization is not thread safe. However, I assume this will be refactored anyway and we will create the WSClient only when we actually need it.

Sources_v3/StreamChat/ChatClient.swift
VojtaStavik: On the other hand - check my Slack comments. This will need to be benchmarked.

Sources_v3/StreamChat/ChatClient.swift
```diff
@@ -259,59 +272,234 @@ public class _ChatClient<ExtraData: ExtraDataTypes> {
...
+
+            // It makes more sense to create background workers here
+            // other than in `init` because workers without currently logged-in
+            // user do nothing.
+            if backgroundWorkers.isEmpty {
+                createBackgroundWorkers()
+            }
+
+            completion(nil)
+            return
```
VojtaStavik, line 360: Do we also need to recreate the WS connect endpoint here to reflect the new token? Or maybe even disconnect the WS?

Sources_v3/StreamChat/ChatClient.swift
```diff
@@ -259,59 +272,234 @@ public class _ChatClient<ExtraData: ExtraDataTypes> {
...
-    private func createWebSocketClient() {
+    func createWebSocketClient() {
         _ = webSocketClient
     }
+
+    func prepareEnvironment(newToken: Token, completion: @escaping (Error?) -> Void) {
+        // Check token is for current user.
+        guard newToken.userId != currentUserId else {
+            // Update the current token with the new one.
+            currentToken = newToken
```
VojtaStavik, line 347: We can also check whether the token is different. If not, we can just move on

Sources_v3/StreamChat/ChatClient_Tests.swift
```diff
@@ -373,58 +374,345 @@ class ChatClient_Tests: StressTestCase {
...
+
+        // Call `reloadUserIfNeeded` and catch the error.
+        var reloadUserIfNeededCompletionCalled = false
+        client.reloadUserIfNeeded { error in
+            XCTAssertNil(error)
+            reloadUserIfNeededCompletionCalled = true
+        }
+
+        // Wait `reloadUserIfNeeded` to be completed.
+        AssertAsync.willBeTrue(reloadUserIfNeededCompletionCalled)
```
VojtaStavik, line 446: You can just do:
```swift
let error = try await { client.reloadUserIfNeeded($0) }
XCTAssertNil(error)
```

Sources_v3/StreamChat/ChatClient_Tests.swift
```diff
@@ -373,58 +374,345 @@ class ChatClient_Tests: StressTestCase {
...
+        // Assert token is updated.
+        XCTAssertEqual(client.currentToken, lastProvidedToken)
```
VojtaStavik, line 448: This assumes that the token provider was called but there's no guarantee that actually happened. Maybe we should reset `lastProvidedToken` before calling `client.reloadUserIfNeeded`?

Sources_v3/StreamChat/ChatClient_Tests.swift
```diff
@@ -373,58 +374,345 @@ class ChatClient_Tests: StressTestCase {
...
+        // Assert current user is not changed.
+        XCTAssertEqual(client.currentUserId, currentUserId)
+        // Assert background workers are not recreated.
+        XCTAssertEqual(client.backgroundWorkers.count, 2)
```
VojtaStavik, line 452: How do you know the workers weren't recreated? What if these are two different workers? :)

Sources_v3/StreamChat/ChatClient_Tests.swift
```diff
@@ -373,58 +374,345 @@ class ChatClient_Tests: StressTestCase {
...
+            },
+            workerBuilders: [TestWorker.init],
+            eventWorkerBuilders: [TestEventWorker.init],
+            environment: .mock,
+            completion: { error in
+                XCTAssertNil(error)
+                completionCalled = true
+            }
+        )
+
```
VojtaStavik, line 483: I'd add here:
```swift
// Assert the completion wasn't called yet
XCTAssertFalse(completionCalled)
```

Sources_v3/StreamChat/ChatClient_Tests.swift
```diff
@@ -373,58 +374,345 @@ class ChatClient_Tests: StressTestCase {
...
+        // Call `reloadUserIfNeeded` and catch the error.
+        var reloadUserIfNeededError: Error?
+        client.reloadUserIfNeeded {
+            reloadUserIfNeededError = $0
+        }
+
+        // Assert `tokenProviderError` is propagated.
+        AssertAsync.willBeEqual(reloadUserIfNeededError as? TestError, tokenProviderError)
```
VojtaStavik, line 567: You can use `await { }` here too (and other similar cases in the file)

Sources_v3/StreamChat/ChatClient_Tests.swift
```diff
@@ -373,58 +374,345 @@ class ChatClient_Tests: StressTestCase {
...
+        // and giving any token for any user will fails since passive clients don't allow
+        // switching the user.
+        var environment = ChatClient.Environment.mock
+        let database = try environment.databaseContainerBuilder(.inMemory, true, false)
+        try database.createCurrentUser(id: currentUserId)
+        environment.databaseContainerBuilder = { _, _, _ in database }
+
+        // Create passive client
+        let passiveClient = ChatClient(
+            config: inactiveInMemoryStorageConfig,
```
VojtaStavik, line 691: nit: can we please rename `inactiveInMemoryStorageConfig` -> `passiveInMemoryStorageConfig`

Sources_v3/StreamChat/ChatClient.swift
```diff
@@ -76,12 +76,8 @@ public class _ChatClient<ExtraData: ExtraDataTypes> {
     /// To observe changes in the connection status, create an instance of `CurrentChatUserController`, and use it to receive
     /// callbacks when the connection status changes.
     ///
-    public var connectionStatus: ConnectionStatus {
-        ConnectionStatus(
-            webSocketConnectionState: webSocketClient?.connectionState ?? .disconnected(error: .ClientIsNotInActiveMode())
-        )
-    }
-    
+    public private(set) var connectionStatus: ConnectionStatus = .initialized
```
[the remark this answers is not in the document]
evsaev: It reflects the web-socket since we change the `connectionStatus` each time the `webSocketClient(_, didUpdateConectionState)` is invoked.

We also discussed that `connectionStatus` should be changed from `initialized` -> `disconnected` only when the current user session is restored but if we do:

```swift
public var connectionStatus: ConnectionStatus {
 guard let webSocket = webSocketClient else { .initilized }
 return ConnectionStatus(
  webSocketConnectionState: webSocketClient.connectionState
 )
}
```

for **active** client it becomes `disconnected` right after `init` is finished (instead of `initilized` which tells `client.currentUserId` and `currentUserController.currentUser?.id` can be different)

Sources_v3/StreamChat/ChatClient.swift
evsaev: >Do we also need to recreate the WS connect endpoint here to reflect the new token

no, since the `webSocketConnect` endpoint requires `userId` which's not changed.

> Or maybe even disconnect the WS?

that is a good question 👍 Looks like we should because the previous token could expire

Sources_v3/StreamChat/ChatClient.swift
```diff
@@ -259,43 +272,45 @@ public class _ChatClient<ExtraData: ExtraDataTypes> {
...
-        
-        createBackgroundWorkers()
+
+        currentUserId = fetchCurrentUserIdFromDatabase()
         createWebSocketClient()
+        createAPIClient()
+
+        environment
+            .clientUpdaterBuilder(self)
+            .reloadUserIfNeeded(completion: completion)
```
VojtaStavik, line 293: It took me a while to understand...
```swift
let updater = environment.clientUpdaterBuilder(self)
updater.reloadUserIfNeeded(completion: completion)
```

Sources_v3/StreamChat/ChatClient.swift
```diff
@@ -259,43 +272,45 @@ public class _ChatClient<ExtraData: ExtraDataTypes> {
...
+        self.tokenProvider = tokenProvider
         self.environment = environment
         self.workerBuilders = workerBuilders
         self.eventWorkerBuilders = eventWorkerBuilders
-        
-        createBackgroundWorkers()
+
+        currentUserId = fetchCurrentUserIdFromDatabase()
         createWebSocketClient()
+        createAPIClient()
```
VojtaStavik, line 289: Do we need to initialize API and WS clients in `init`?

Sources_v3/StreamChat/ChatClient.swift
```diff
@@ -305,13 +320,34 @@ public class _ChatClient<ExtraData: ExtraDataTypes> {
...
+    func createWebSocketClient() {
         _ = webSocketClient
     }
+
+    /// Before CIS-574, `apiClient` was initialized in `createBackgroundWorkers` in `init` since it was passed
+    /// to the background workers. Now, background workers do not access `webSocketClient` and it's not being
+    /// initialized in `init`, so we initialize it here explicitly.
+    func createAPIClient() {
+        _ = apiClient
+    }
```
VojtaStavik, line 350: PLEASE let's remove this madness here for both WS and API clients 😅

Sources_v3/StreamChat/ChatClient_Tests.swift
```diff
@@ -370,61 +375,97 @@ class ChatClient_Tests: StressTestCase {
...
+        }
     }
     
+    func test_reloadUserIfNeededIsCalled_whenClientIsInitialized_andErrorIsPropagated() {
+        for error in [nil, TestError()] {
+            var clientInitCompletionCalled = false
+            var clientInitCompletionError: Error?
+
+            // Create a client, provide the completion and catch the result.
+            // Create a Client instance and check the workers are initialized properly
```
VojtaStavik, line 435: nit: which comment is the right one? :)

Sources_v3/StreamChat/Workers/ChatClientUpdater_Tests.swift
```diff
@@ -0,0 +1,523 @@
...
+        XCTAssertEqual(client.mockWebSocketClient.disconnect_calledCounter, 0)
+        // Assert `completeTokenWaiters` was not called since token is not changed.
+        XCTAssertFalse(client.completeTokenWaiters_called)
+        // Assert background workers stay the same.
+        XCTAssertEqual(client.testBackgroundWorkerIDs, oldWorkderIDs)
+        // Assert database is not flushed.
+        XCTAssertFalse(client.mockDatabaseContainer.removeAllData_called)
+    }
+
+    func test_reloadUserIfNeeded_happyPaths() throws {
```
VojtaStavik, line 215: This is one hell of a test 🤯

Sources_v3/StreamChat/ChatClient.swift
b-onc: @evsaev We need to re-think tests so creation of APIClient/WSClient is not necessary in `Client.init`

Sources_v3/StreamChat/Workers/ChatClientUpdater_Tests.swift
evsaev: That's not surprising 😄 Changing a user is indeed the most complex part of the SDK I've ever faced with

Sources_v3/StreamChat/ChatClient_Tests.swift
evsaev: removed one

Sources_v3/StreamChat/ChatClient.swift
evsaev: done in cc4bab1787e818d3a2cfcfe3613f6f9f093a0e56

Sources_v3/StreamChat/ChatClient.swift
evsaev: no really, fixed and updated tests in cc4bab1787e818d3a2cfcfe3613f6f9f093a0e56

Sources_v3/StreamChat/ChatClient.swift
evsaev: done in c45940aca0a817c617215c7950bdf8554aa6d932

Sources_v3/StreamChat/Workers/ChatClientUpdater.swift
```diff
@@ -0,0 +1,162 @@
...
+    /// are received.
+    func disconnect() {
+        // Disconnecting is not possible in connectionless mode (duh)
+        guard client.config.isClientInActiveMode else {
+            log.error(ClientError.ClientIsNotInActiveMode().localizedDescription)
+            return
+        }
+
+        guard client.connectionId != nil else {
+            log.warning("The client is already disconnected. Skipping the `connect` call.")
```
VojtaStavik, line 148: `DISconnect call`

Sources_v3/StreamChat/Database/DatabaseContainer_Mock.swift
```diff
@@ -38,7 +39,12 @@ class DatabaseContainerMock: DatabaseContainer {
     }
     
     override func removeAllData(force: Bool = true) throws {
-        flush_called = true
+        removeAllData_called = true
+
+        if let error = removeAllData_errorResponse {
+            throw error
+        }
```
VojtaStavik, line 46: 👍

Sources_v3/StreamChat/ChatClient_Tests.swift
```diff
@@ -403,6 +418,9 @@ class ChatClient_Tests: StressTestCase {
             eventWorkerBuilders: [TestEventWorker.init],
             environment: testEnv.environment
         )
+
+        // Simulate `createBackgroundWorkers` so workers are created.
+        client.createBackgroundWorkers()
```
VojtaStavik, line 429: I think this solution is great 👍

Sources_v3/StreamChat/Config/ChatClientConfig.swift
```diff
@@ -45,17 +45,19 @@ public struct ChatClientConfig {
...
     /// This flag is automatically set to `false` for app extensions
     /// **Warning**: There should be at max 1 active client at the same time, else it can lead to undefined behavior.
     public var isClientInActiveMode: Bool
+
+    /// If set to `true` the `ChatClient` will restore the last
+    /// logged-in user session and establishing a web-socket connection automatically
+    /// when it is instantiated.
+    ///
+    /// This flag `true` by default.
+    public var shouldConnectAutomatically = true
```
VojtaStavik, line 60: 🎉

Sources_v3/StreamChat/Controllers/ChannelController/ChannelController_Tests.swift
```diff
@@ -185,10 +185,17 @@ class ChannelController_Tests: StressTestCase {
     }
 
     func test_channelControllerForNewChannel_throwsError_ifCurrentUserDoesNotExist() throws {
+        let clientWithoutCurrentUser = ChatClient(
+            config: .init(apiKeyString: .unique),
+            tokenProvider: .closure {
+                $1(.failure(TestError()))
+            }
```
VojtaStavik, line 192: Maybe we can introduce some kind of `.invalid` helper for tests?

VojtaStavik (approved):
Looks great. Let's merge this 🚢 🇮🇹 Thank you!

Sources_v3/StreamChat/Config/Token.swift
```diff
@@ -1,40 +1,61 @@
...
+    }
+}
+
+extension Token {
+    static var anonymous: Self {
+        .init(rawValue: "", userId: .anonymous)
+    }
+
+    static func development(userId: UserId) -> Self {
+        .init(rawValue: "development", userId: userId)
```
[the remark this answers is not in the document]
evsaev: This was in token provider until I started updating the `Sample` app. There we need a way to create a token to store it in user-defaults without having a token provider that's why how this extension appeared. 

I don't see much harm in it since some users might want to go with:
```swift
.closure { completion in
 sendTokenRequest { result in
   if error {
      completion(.anonymous)
   } else {
      completion(result.token)
   }
 }
}
```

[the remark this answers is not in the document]
VojtaStavik: @evsaev Makes sense 👍 Lets then add comments explaining when this should be used.

Sources_v3/StreamChat/Config/Token.swift
```diff
@@ -1,40 +1,61 @@
...
-/// A token provider is a function in which you send a request to your own backend to get a Stream Chat API token.
-///
-/// Set your custom `TokenProvider` in `ChatClientConfig` when creating a `ChatClient` instance. `ChatClient` will use it whenever
-/// it needs to get a new token for the given user.
-///
-public typealias TokenProvider = (_ apiKey: APIKey, _ userId: UserId, _ completion: @escaping (Token?) -> Void) -> Void
+    /// Created a new `Token` instance.
+    /// - Parameter value: The JWT string value. It must be in valid format and contain `user_id` in payload.
+    public init(stringLiteral value: StringLiteralType) {
+        try! self.init(rawValue: value)
```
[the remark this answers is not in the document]
evsaev: Used `fatalError` instead of our `log.assertion` since that's the `init` and it must have `self` with set `rawValue/userId` at the time it returns. But returning a token, in this case, doesn't make sense and we can/should crash (since it's used when the token is hardcoded as a string inside the app so we just notify the developer the token is invalid) 

e55f970b97ce750da86b4692041b7a03c1ee30b7

Tests_v3/Shared/TemporaryData.swift
```diff
@@ -31,6 +31,13 @@ extension String {
     static var unique: String { UUID().uuidString }
 }
 
+extension Token {
+    /// Returns a new unique string
+    static func unique(userId: UserId = .unique) -> Self {
+        .init(rawValue: .unique, userId: userId)
+    }
+}
+
```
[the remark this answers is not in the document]
evsaev: mentioned in e55f970b97ce750da86b4692041b7a03c1ee30b7

Sources_v3/StreamChat/Config/Token.swift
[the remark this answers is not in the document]
evsaev: added in e55f970b97ce750da86b4692041b7a03c1ee30b7

Sources_v3/StreamChat/ChatClient_Tests.swift
```diff
@@ -194,18 +194,63 @@ class ChatClient_Tests: StressTestCase {
...
+            environment: env,
+            completion: { error in
+                XCTAssertNil(error)
+                initCompletionCalled = true
+            }
         )
 
-        // Simulate connection state change of WSClient
+        // Assert `init` completion hasn't been called yet.
+        XCTAssertTrue(initCompletionCalled)
```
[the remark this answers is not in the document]
evsaev: fixed in 2701a7e8f7b4cf05a99ed5c73446b198c404abd3

Sources_v3/StreamChat/Workers/ChatClientUpdater.swift
evsaev: fixed in 6309a297de5b5882f7a60485741ce64093b57b6a

Sources_v3/StreamChat/Controllers/ChannelController/ChannelController_Tests.swift
evsaev: introduced in 381b4b8f2dd4633078258599584e7a0e7758fb76

PR #401, opened 2020-08-12
[CIS-279] Open a TCP connection while WS is connecting

VojtaStavik:
### In this PR

To improve request latency, we use the time while the WS connection is establishing to open a new TCP connection, which can then later be reused by other requests. With the backend increasing the `keep:alive` timeout, this should improve the request latency significantly.

PR #733, opened 2021-01-12
Fix QueryUsers is used for TCP heatup

### LINKED ISSUE #729: Malformed call to /users endpoint without team filter on SDK launch
## What did you do?
Everytime the SDK tries to connect, it does a call to the `/users` API endpoint, which result to the following `403` error :

```
HTTP/1.1 403 Forbidden
Server: fasthttp
Date: Mon, 11 Jan 2021 17:57:58 GMT
Content-Type: application/json;charset=utf-8
Content-Length: 190
Access-Control-Allow-Origin: *
Access-Control-Max-Age: 86400
Access-Control-Allow-Headers: x-requested-with, content-type, accept, origin, authorization, x-csrftoken, x-stream-client, stream-auth-type, x-csrf-token, content-encoding
Access-Control-Allow-Methods: GET, POST, PUT, PATCH, DELETE, OPTIONS
Cache-Control: no-cache
X-Ratelimit-Limit: 1000
X-Ratelimit-Remaining: 998
X-Ratelimit-Reset: 1610387880
Content-Encoding: gzip
Keep-Alive: timeout=58, max=1000
Connection: keep-alive

{"code":17,"message":"QueryUsers failed with error: \"you must include a team filter to use this endpoint client-side\"","StatusCode":403,"duration":"0.00ms","more_info":"https://getstream.io/chat/docs/api_errors_response"}
```

No API call to this endpoint on SDK init.

## GetStream Environment
**GetStream Chat version:**
2.6.2

**GetStream Chat frameworks:**
StreamChatClient

 **iOS version:**
14.3

 **Xcode version:**
12.3

**Device:**
All devices

## Additional context

Here is the line which make the API call without team filter :
https://github.com/GetStream/stream-chat-swift/blob/8d9ef364e9685a02d601bad9ef50a249f17c08b5/Sources/Client/Client/Client.swift#L260

gz#8547

b-onc:
In #401, heatUpTCPConnection was introduced to improve latency, but for unknown reasons the endpoint used was changed to queryUsers in #442. This fixes that and resolves #729

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Upcoming
 
-### 🔄 Changed
+### 🐞 Fixed
+- Make an OPTIONS request to `connect` instead of `QueryUsers` request to heat up HTTP connection. [#733](https://github.com/GetStream/stream-chat-swift/issues/733)
 
 # [2.6.2](https://github.com/GetStream/stream-chat-swift/releases/tag/2.6.2)
 _December 31, 2020_
--- a/Sources/Client/Client/Client.swift
+++ b/Sources/Client/Client/Client.swift
@@ -257,7 +257,7 @@ public final class Client: Uploader {
             webSocket.connect()
             
             // The improve latency, call the preheat endpoint to set up TCP connection which can be reused.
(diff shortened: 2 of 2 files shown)


VojtaStavik (approved):

VojtaStavik (commented):
Before merging this, let's verify the preheating still works. I think there were some issues with using `OPTIONS`. The connection wasn't being reused.

PR #735, opened 2021-01-12
Fix photo library permission asked unnecessarily

### LINKED ISSUE #657: ChatViewController. ‘View photo library' access setting in iOS 14
When user tries to add photos to message and selects option to allow selected photos,  all library's photos are opened instead.
It is related to changes in iOS 14. 
https://developer.apple.com/forums/thread/653414

## GetStream Environment
**iOS version:** 14
**Swift version:** 5
**Xcode version:** 12.2
**Device:** iPhone 12 Pro Max

gz#8059

b-onc:
Closes #657

Tested and works. SDK can access assets selected by user, and camera asks for permission automatically.

This makes me wonder why so many apps require full access while they could get by with this approach... (Nearly all messaging apps, marketplace apps, etc...)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Upcoming
 
-### 🔄 Changed
+### 🐞 Fixed
+- Photo library permission was asked unnecessarily. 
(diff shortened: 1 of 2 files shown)


VojtaStavik (approved):
Nice 👍

cardoso (approved):

PR #736, opened 2021-01-12
[CIS-505] Quoted messages

mantergo:
In this PR:

- Add LLC support for quoted messages sending and receiving

UI part will be done in `CIS-575`.

### DIFF
--- a/Sources_v3/StreamChat/Controllers/ChannelController/ChannelController.swift
+++ b/Sources_v3/StreamChat/Controllers/ChannelController/ChannelController.swift
@@ -657,6 +657,7 @@ public extension _ChatChannelController {
     ///   - text: Text of the message.
     ///   - extraData: Additional extra data of the message object.
     ///   - attachments: An array of the attachments for the message.
+    ///   - quotedMessageId: An id of the message new message quotes. (inline reply)
     ///   - completion: Called when saving the message to the local DB finishes.
     ///
     func createNewMessage(
@@ -664,6 +665,7 @@ public extension _ChatChannelController {
 //        command: String? = nil,
 //        arguments: String? = nil,
         attachments: [_ChatMessageAttachment<ExtraData>.Seed] = [],
+        quotedMessageId: MessageId? = nil,
         extraData: ExtraData.Message = .defaultValue,
         completion: ((Result<MessageId, Error>) -> Void)? = nil
     ) {
@@ -684,6 +686,7 @@ public extension _ChatChannelController {
             command: nil,
             arguments: nil,
             attachments: attachments,
+            quotedMessageId: quotedMessageId,
             extraData: extraData
         ) { result in
             self.callback {
--- a/Sources_v3/StreamChat/Models/Message.swift
+++ b/Sources_v3/StreamChat/Models/Message.swift
@@ -1,5 +1,5 @@
 //
-// Copyright © 2020 Stream.io Inc. All rights reserved.
+// Copyright © 2021 Stream.io Inc. All rights reserved.
 //
 
 import Foundation
@@ -67,6 +67,12 @@ public struct _ChatMessage<ExtraData: ExtraDataTypes> {
     /// Learn more about using custom extra data in our [cheat sheet](https://github.com/GetStream/stream-chat-swift/wiki/StreamChat-SDK-Cheat-Sheet#working-with-extra-data).
     ///
     public let extraData: ExtraData.Message
+    
+    /// Quoted message id.
+    ///
+    /// If message is inline reply this property will contain id of the message quoted by this reply.
+    ///
+    public let quotedMessageId: MessageId?
     
     /// A flag indicating whether the message is a silent message.
     ///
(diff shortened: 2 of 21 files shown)


Sources_v3/StreamChat/Models/Message.swift
```diff
@@ -68,6 +68,17 @@ public struct _ChatMessage<ExtraData: ExtraDataTypes> {
...
+    // Workaround for storing values of the same type in struct.
+    var _quotedMessage: [_ChatMessage] = []
+    
+    /// Quoted message.
+    ///
+    /// If message is inline reply this property will contain message quoted by this reply.
+    ///
+    public var quotedMessage: _ChatMessage? {
+        _quotedMessage.first
+    }
```
mantergo, line 80: I'm not happy with this but the alternative is to switch to `class` that will introduce inconsistency in public facing API and need to provide initializer here as well as adjust Initializer from `MessageDTO` in `MessageDTO.swift`.
That's not really big deal but I just can't tell which option is better.

Sources_v3/StreamChat/Workers/MessageUpdater.swift
```diff
@@ -136,6 +137,7 @@ class MessageUpdater<ExtraData: ExtraDataTypes>: Worker {
         parentMessageId: MessageId,
         attachments: [_ChatMessageAttachment<ExtraData>.Seed],
         showReplyInChannel: Bool,
+        quotedMessageId: MessageId?,
```
VojtaStavik, line 140: Hmmm, I'm not sure about this. Also it seems we have too many parameters in the method, `cid` and `parentMessageId` shouldn't be there since they are the ones provided by the controller. 

I'd suggest having two separate methods for this with the following arguments:
```
func createNewThreadReply(
        text: String,
        command: String?,
        arguments: String?,
        attachments: [_ChatMessageAttachment<ExtraData>.Seed],
        showReplyInChannel: Bool,
        extraData: ExtraData.Message,
        completion: ((Result<MessageId, Error>) -> Void)? = nil
    ) 
     
func createNewQuoteReply(
        text: String,
        command: String?,
        arguments: String?,
        attachments: [_ChatMessageAttachment<ExtraData>.Seed],
        extraData: ExtraData.Message,
        completion: ((Result<MessageId, Error>) -> Void)? = nil
     ) 
```

What do you think?

Sources_v3/StreamChat/Workers/ChannelUpdater.swift
```diff
@@ -103,6 +104,7 @@ class ChannelUpdater<ExtraData: ExtraDataTypes>: Worker {
         command: String?,
         arguments: String?,
         attachments: [_ChatMessageAttachment<ExtraData>.Seed] = [],
+        quotedMessageId: MessageId?,
```
VojtaStavik, line 107: ... here it can stay as is since this is not public API and we can make sure we handled all edge cases.

Sources_v3/StreamChat/Controllers/ChannelController/ChannelController.swift
```diff
@@ -657,13 +657,15 @@ public extension _ChatChannelController {
...
     ///   - attachments: An array of the attachments for the message.
+    ///   - quotedMessageId: An id of the message new message quotes. (inline reply)
     ///   - completion: Called when saving the message to the local DB finishes.
     ///
     func createNewMessage(
         text: String,
 //        command: String? = nil,
 //        arguments: String? = nil,
         attachments: [_ChatMessageAttachment<ExtraData>.Seed] = [],
+        quotedMessageId: MessageId? = nil,
```
VojtaStavik, line 668: The replies should be created only via message controller of the parent message.

Sources_v3/StreamChat/Models/Message.swift
VojtaStavik: I think this is actually dangerous in our use case. The backend makes sure the "quoted message depth" doesn't go more than 1 level but we can't do it locally. It means if we had a really long conversation with quoted replies we could end up wit serializing quite a lot of objects :) 
```
let originalMessageId = message.quotedMessage?.quotedMessage?.quotedMessage?.quotedMessage?.quotedMessage?.id
```

I think all we need to expose in this case is `quotedMessageId`. Users can easily fetch the data if needed via our `DataStore` API: 
```swift
let quotedMessage = client.dataStore.message(id: message.quotedMessageId!)
````

VojtaStavik (commented):
Looking really good 👍 My main comment is about the public API. I think we should expose the methods for creating thread reply and quoted reply only on the message controller of the parent message.

Sources_v3/StreamChat/Workers/MessageUpdater.swift
mantergo: The thing is it's possible to send quoted thread reply.
That is why I decided that adjusting existing public API will be least confusing option.

mantergo:
Current approach after review changes and clarifications:
- `MessageDTO` doesn't have `quotedMessageId` property. When created, `quotedMessage` for id is loaded and linked
- `quotedMessageId` property from backend response is ignored in `MessagePayload`
- both `ChannelController` and `MessageController` have API to create quoted messages
- in `ChatMessage` only `quotedMessageId` is exposed

VojtaStavik (approved):
B E A U T I FU L ❤️

PR #737, opened 2021-01-13
Introduce custom layout for message list

ManWithBear:
Custom Table View like layout that position item at index path 0-0 on bottom of the list. 
Unlike `UICollectionViewFlowLayout` we ignore some invalidation calls and persist items attributes between updates.
This resolves problem when on item reload layout would change content offset and user ends up on completely different item.
Layout intended for batch updates and right now I have no idea how it will react to `collectionView.reloadData()` (so far fine).

During testing on iPhone SE (1st gen) iOS 14.2 I was able to achieve hitch ratio:
- initial scroll from bottom to top: 49 ms/s
- scroll from top to bottom on already layouted list: 46 ms/s

So whole collection view layout related logic bring almost no overhead. (This are still awful numbers, Apple do not recommend ratio over 10 ms/s)

Known issues:
> when message list first time appears on screen, all items are in estimated height state (system just doesn't call autolayout updates until later)

Worked around by hiding message list till first layout reload

> messages that contains just small text return wrong size in `preferredLayoutAttributesFitting` initially and proper one after sometime. This cause unpleasant jump when cell in the middle of screen resize.

Solved by returning zero size for cells that haven't been injected to tree yet.

---
Hitch - any time a frame appears on screen later than expected
Hitch duration - time past deadline for frame preparation (in ms)
Hitch ratio = combined duration of all hitches / total test duration (in ms/s)
https://developer.apple.com/videos/play/tech-talks/10855

### DIFF
--- a/Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageCollectionViewCell.swift
+++ b/Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageCollectionViewCell.swift
@@ -53,6 +53,14 @@ open class _СhatMessageCollectionViewCell<ExtraData: ExtraDataTypes>: Collectio
     override open func preferredLayoutAttributesFitting(
         _ layoutAttributes: UICollectionViewLayoutAttributes
     ) -> UICollectionViewLayoutAttributes {
+        guard hasCompletedStreamSetup else {
+            // We cannot calculate size properly right now, because our view hierarchy is not ready yet.
+            // If we just return default size, small text bubbles would not resize itself properly for no reason.
+            let attributes = layoutAttributes.copy() as! UICollectionViewLayoutAttributes
+            attributes.frame.size = .zero
+            return attributes
+        }
+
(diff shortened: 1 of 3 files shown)


evsaev (commented):
The way the message list behaves when I add/remove a reaction is soooo nice! ❤️ However when I switched to the branch I found some glitches when chat is opened (1) and broken link previews (2). Not sure it's because of updated layout 🤔 

(1)
Simulator Screen Shot - iPhone 12 - 2021-01-13 at 12 15 57
(2)
Screen Shot 2021-01-13 at 12 16 20 PM

VojtaStavik (commented):
I must say I have no idea what's happening here but as long as we have the magic encapsulated in the layout object I'm fine with that 👍 😅  I can also see some glitches (expected) but in general the behavior looks to be really nice for 250 LOC. Great work 🎉 Let's continue and see what we can get our of it 🚀

VojtaStavik (approved):
Great, it seems this performs better than the current solution. Great work @ManWithBear 🎉 Thank you. Let's 🚢 🇮🇹

PR #739, opened 2021-01-13
[CIS-235] Update current user

nuno-vieira:
**In this PR:**
Adds a new method `updateCurrentUser()` to the LLC for updating the current user data. 
It uses the [partial update endpoint](https://getstream.io/chat/docs_rest/#partial-update-users). The endpoint supports resetting user data but is out of the scope of this PR.
Adds support to update the user data in the Sample app.

- [X] Updating partially new user data
- [x] Add support to change user data in the Sample App

https://user-images.githubusercontent.com/12814114/104375662-a4113080-551b-11eb-8f63-de2ef940e8db.mov


VojtaStavik (changes requested):
@nuno-vieira , congratulation to your first PR for Stream! 🎉 In general the PR looks good. I left a couple of comments, mostly regarding removing unused code, and simplifying a couple of things.

Sources_v3/StreamChat/APIClient/Endpoints/CurrentUserEndpoints.swift
```diff
@@ -0,0 +1,27 @@
...
+// Copyright © 2021 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+extension Endpoint {
+    static func updateCurrentUser<ExtraData: UserExtraData>(
+        id: UserId,
+        payload: CurrentUserUpdateRequestBody<ExtraData>
+    ) -> Endpoint<CurrentUserUpdateResponse<ExtraData>> {
```
VojtaStavik, line 11: Since we provide `UserId` I don't think this needs to be named `updateCurrentUser`. Technically, you can update any user you want if you have rights to do that. So maybe just `updateUser`?

Sources_v3/StreamChat/APIClient/Endpoints/Payloads/CurrentUserPayloads.swift
```diff
@@ -61,8 +61,46 @@ class CurrentUserPayload<ExtraData: UserExtraData>: UserPayload<ExtraData> {
 }
 
 /// An object describing the outgoing user JSON payload.
-class CurrentUserRequestBody<ExtraData: UserExtraData>: Encodable {
-    // TODO: Add more fields while working on CIS-235
+struct CurrentUserUpdateRequestBody<ExtraData: UserExtraData>: Encodable {
```
VojtaStavik, line 64: This can be just `UserUpdateRequestBody`

Sources_v3/StreamChat/APIClient/Endpoints/Payloads/CurrentUserPayloads.swift
```diff
@@ -61,8 +61,46 @@ class CurrentUserPayload<ExtraData: UserExtraData>: UserPayload<ExtraData> {
...
+    
+    func encode(to encoder: Encoder) throws {
+        var container = encoder.container(keyedBy: UserPayloadsCodingKeys.self)
+        try container.encodeIfPresent(name, forKey: .name)
+        try container.encodeIfPresent(imageURL, forKey: .imageURL)
+        try extraData?.encode(to: encoder)
+    }
+}
+
+struct CurrentUserUpdateResponse<ExtraData: UserExtraData>: Decodable {
```
VojtaStavik, line 83: Same here, this should just be `UserUpdateResponse`

Sources_v3/StreamChat/APIClient/Endpoints/CurrentUserEndpoints_Tests.swift
```diff
@@ -0,0 +1,40 @@
...
+            requiresConnectionId: false,
+            body: body
+        )
+        
+        let endpoint: Endpoint<CurrentUserUpdateResponse> = .updateCurrentUser(id: userId, payload: payload)
+        
+        XCTAssertEqual(AnyEndpoint(expectedEndpoint), AnyEndpoint(endpoint))
+    }
+}
+
```
VojtaStavik, line 36: The commit message says `Add update current user endpoint` but I don't see you doing this in the commit :)

Sources_v3/StreamChat/Controllers/CurrentUserController/CurrentUserController.swift
```diff
@@ -364,6 +364,47 @@ public extension _CurrentChatUserController {
...
+    /// Updates the current user.
+    ///
+    /// By default all data is `nil`, and it won't be updated unless a value is provided.
+    ///
+    /// - Parameters:
+    ///   - name: Optionally provide a new name to be updated.
+    ///   - imageURL: Optionally provide a new image to be updated.
+    ///   - userExtraData: Optionally provide new user extra data to be updated.
+    ///   - completion: Called when user is successfuly updated, or with error.
+    func updateUser(
```
VojtaStavik, line 376: `updateUser` -> `updateUserData` ? Update users can be understood like "changing" the actual logged-in user.

Sources_v3/StreamChat/Controllers/CurrentUserController/CurrentUserController.swift
```diff
@@ -364,6 +364,47 @@ public extension _CurrentChatUserController {
...
+        name: String? = nil,
+        imageURL: URL? = nil,
+        userExtraData: ExtraData.User? = nil,
+        completion: ((Error?) -> Void)? = nil
+    ) {
+        guard let currentUserId = currentUser?.id else {
+            completion?(ClientError.CurrentUserDoesNotExist())
+            return
+        }
+
```
VojtaStavik, line 386: We should probably also add a check that at least one of the fields is not `nil`. In that case we can just call `completion(nil)` but we should log a warning that nothing happened.

VojtaStavik (changes requested):
Looking good 👍 Just a couple of more suggestions regarding naming.

Sources_v3/StreamChat/APIClient/Endpoints/CurrentUserEndpoints.swift
nuno-vieira: Then, should this be moved to the UserEndpoints? Instead of having a `CurrentUserEndpoints`?

Sources_v3/StreamChat/APIClient/Endpoints/Payloads/CurrentUserPayloads.swift
nuno-vieira: Same as above, should this move to `UserPayloads` ?

Sources_v3/StreamChat/APIClient/Endpoints/Payloads/CurrentUserPayloads.swift
nuno-vieira: same as above

Sources_v3/StreamChat/APIClient/Endpoints/Payloads/CurrentUserPayloads.swift
VojtaStavik: @nuno-vieira I'd say so 🤷‍♂️

Sources_v3/StreamChat/Controllers/CurrentUserController/CurrentUserController.swift
```diff
@@ -364,6 +364,53 @@ public extension _CurrentChatUserController {
...
+    ) {
+        guard let currentUserId = currentUser?.id else {
+            completion?(ClientError.CurrentUserDoesNotExist())
+            return
+        }
+        
+        let params: [Any?] = [name, imageURL, userExtraData]
+        guard !params.allSatisfy({ $0 == nil }) else {
+            log.warning("Update user request not performed. All provided data was nil.")
+            return
```
VojtaStavik, line 390: You should call the completion here, too.

Sources_v3/StreamChat/Controllers/CurrentUserController/CurrentUserController_Tests.swift
```diff
@@ -565,6 +565,182 @@ final class CurrentUserController_Tests: StressTestCase {
...
+    
+    func test_updateUser_whenNoDataProvided_shouldNotMakeRequest() throws {
+        _ = try await {
+            controller.updateUserData(
+                name: nil,
+                imageURL: nil,
+                userExtraData: nil,
+                completion: $0
+            )
+        }
```
VojtaStavik, line 707: How come this test passes? You never call completion so the `await` should time-out 🤔

VojtaStavik (approved):
Looks good to me 👍 Thanks! Let's just fix the missing completion call.

Sources_v3/StreamChat/Controllers/CurrentUserController/CurrentUserController.swift
nuno-vieira: Fixed 👍

Sources_v3/StreamChat/Controllers/CurrentUserController/CurrentUserController_Tests.swift
nuno-vieira: The test was actually bad. I had to first simulate that a user was already logged in to properly test this. So in fact, the await did not timeout because the completion was actually being called, but in the first error case, and not this one.

PR #740, opened 2021-01-13
[CIS-577] Optional current user id

evsaev:
**This PR** :
- makes `currentUserId` optional in `ChatClient` so we can understand when there is no currnrt logged-in user
- excludes connect endpoint from `WebSocketClient.init` (since there can be NO user when it is created)
- updates other parts of LLC to deal with optional `currentUserId`

_This PR is one of the preparation PRs required for [CIS-574 Implement token-based user setting](https://stream-io.atlassian.net/secure/RapidBoard.jspa?rapidView=61&projectKey=CIS&modal=detail&selectedIssue=CIS-574) where we will restore the current user session and provide different way of token managment._


Sources_v3/StreamChat/ChatClient.swift
```diff
@@ -250,7 +241,7 @@ public class _ChatClient<ExtraData: ExtraDataTypes> {
         } else {
             workerBuilders = []
             eventWorkerBuilders = []
-            environment.webSocketClientBuilder = { _, _, _, _, _, _ in nil }
+            environment.webSocketClientBuilder = { _, _, _, _, _ in nil }
```
VojtaStavik, line 244: I just realized, imagine `webSocketClientBuilder` being optional instead of the returning value being optional 🤯

Sources_v3/StreamChat/WebSocketClient/WebSocketClient.swift
```diff
@@ -92,22 +88,24 @@ class WebSocketClient {
         return pingController
     }()
     
-    private func createEngine() -> WebSocketEngine {
+    private func createOrReuseEngine(for connectEndpoint: Endpoint<EmptyResponse>) -> WebSocketEngine {
```
VojtaStavik, line 91: `createEngineIfNeeded`? 🤯

Sources_v3/StreamChat/WebSocketClient/WebSocketClient.swift
```diff
@@ -134,27 +131,27 @@ class WebSocketClient {
...
+        engine = createOrReuseEngine(for: endpoint)
         
         // Cancel the reconnection timer if exists
         reconnectionTimer?.cancel()
         
         connectionState = .connecting
         
         engineQueue.async { [engine] in
-            engine.connect()
+            engine?.connect()
```
VojtaStavik, line 154: This looks like a potential silent failure point to me. Either `!` or unwrap properly and log an error if `engine` is `nil`. It's probably OK for `disconnect` and `sendPing` but here in `connect` we should treat it differently.

Sources_v3/StreamChat/WebSocketClient/WebSocketClient_Tests.swift
```diff
@@ -273,9 +318,11 @@ class WebSocketClient_Tests: StressTestCase {
     }
     
     func test_webSocketPingController_ping_callsEngineWithPing() {
-        assert(engine.sendPing_calledCount == 0)
```
VojtaStavik, line 276: This assert was important. We need to make sure the `sendPing_calledCount` is `0` when we call `sendPing()`. We can reset the counter manually after `test_connectionFlow()` like `engine!.sendPing_calledCount = 0`. Same thing for all other deleted asserts in this commit.

Sources_v3/StreamChat/WebSocketClient/WebSocketClient_Tests.swift
```diff
@@ -273,9 +318,11 @@ class WebSocketClient_Tests: StressTestCase {
...
+        // Simulate connection
+        test_connectionFlow()
+
```
VojtaStavik, line 323: `// Simulate connection to make sure web socket engine exists` ?

Sources_v3/StreamChat/Controllers/ChannelController/ChannelController.swift
```diff
@@ -508,8 +508,13 @@ public extension _ChatChannelController {
...
             return
         }
+
+        guard let currentUserId = client.currentUserId else {
+            callback { completion?(ClientError.CurrentUserDoesNotExist()) }
+            return
+        }
         
-        updater.hideChannel(cid: cid, userId: client.currentUserId, clearHistory: clearHistory) { error in
+        updater.hideChannel(cid: cid, userId: currentUserId, clearHistory: clearHistory) { error in
```
VojtaStavik, line 517: This is strange 🤔 Do we really need `currentUserId` to hide a channel? Maybe the `HideChannelRequest` doesn't have to contain the current user id? The backend should get the current user id from the token, as it does for all other requests.

DemoApp/CreateChatViewController.swift
```diff
@@ -285,7 +285,7 @@ extension CreateChatViewController: UITableViewDelegate, UITableViewDataSource {
         let client = searchController.client
         composerView.controller = try? client
             .channelController(
-                createDirectMessageChannelWith: selectedUserIds.union([client.currentUserId]),
+                createDirectMessageChannelWith: selectedUserIds.union(client.currentUserId.map { [$0] } ?? []),
```
VojtaStavik, line 288: We should handle this properly and show to our user what to do in these cases. When the current user is `nil` it's clear something is wrong and it doesn't make sense to use the fallback. The request will fail aynway.

VojtaStavik (approved):
Looks great, thank you! Left quite a few comments but nothing serious.

DemoApp/CreateChatViewController.swift
evsaev: Shouldn't it be the `LLC` responsibility to include `currentUserID` automatically when `createDirectMessageChannel` API is used? 🤔

VojtaStavik: @evsaev right. It possible we don't need to include the current user id here, too, and the backend will add it automatically.

Sources_v3/StreamChat/ChatClient.swift
evsaev: fixed in c732d658ca9625bd74d35c32c4d80b3a70871a0e

Sources_v3/StreamChat/WebSocketClient/WebSocketClient.swift
evsaev: fixed in c732d658ca9625bd74d35c32c4d80b3a70871a0e

Sources_v3/StreamChat/WebSocketClient/WebSocketClient.swift
evsaev: fixed in c732d658ca9625bd74d35c32c4d80b3a70871a0e. Used force uwrap cause after `createEngineIfNeeded` is call the engine should never become `nil`

Sources_v3/StreamChat/Controllers/ChannelController/ChannelController.swift
evsaev: removed and provided the reason in bcab130e7e58c4a2bd87df2c991488555d208e41

Sources_v3/StreamChat/WebSocketClient/WebSocketClient_Tests.swift
evsaev: fixed in c732d658ca9625bd74d35c32c4d80b3a70871a0e

Sources_v3/StreamChat/WebSocketClient/WebSocketClient_Tests.swift
evsaev: fixed in c732d658ca9625bd74d35c32c4d80b3a70871a0e and also used force-unwrap for `engine` when we expect it to exist

PR #743, opened 2021-01-14
V2: Add channel property to message

### LINKED ISSUE #650: Search. Can't find any reference to Channel
I use _Client.shared.search_ function with _SearchQuery_ paramater to find messages. The answer is [StreamChatClient.Message]. And it doesn’t have the property Channel. But the answer to the request, which I see in Log contains Channel Structure. Is it possible to get the channel in this case? I need it to present the channel with the message on UI. 

 It would be nice to have the Channel data in the parsed response to the Client.shared.search(query: query)

## GetStream Environment
**iOS version:** 14
**Swift version:** 5
**Xcode version:** 12.2
**Device:** iPhone 12 Pro Max

cardoso:
Had to add a fallback for `Config` because it doesn't come in result of message search and I didn't want to break API by making it optional.

Closes #650

### DIFF
--- a/Sources/Client/Model/Channel/Channel.swift
+++ b/Sources/Client/Model/Channel/Channel.swift
@@ -197,6 +197,7 @@ public final class Channel: Codable {
         let members = try container.decodeIfPresent([Member].self, forKey: .members) ?? []
         self.members = Set<Member>(members)
         invitedMembers = Set<Member>()
+        // Fallback because config doesn't come in message search and to avoid breaking API by making it optional
         let config = (try? container.decode(Config.self, forKey: .config)) ?? Config()
         self.config = config
         created = try container.decodeIfPresent(Date.self, forKey: .created) ?? config.created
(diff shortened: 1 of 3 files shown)


Sources/Client/Model/Message/Message.swift
```diff
@@ -199,6 +204,7 @@ public struct Message: Codable {
         id = try container.decode(String.self, forKey: .id)
         type = try container.decode(MessageType.self, forKey: .type)
         user = try container.decode(User.self, forKey: .user)
+        channel = try container.decode(Channel.self, forKey: .channel)
```
b-onc, line 207: This should be `decodeIfPossible`

Sources/Client/Model/Channel/Channel.swift
```diff
@@ -197,7 +197,7 @@ public final class Channel: Codable {
         let members = try container.decodeIfPresent([Member].self, forKey: .members) ?? []
         self.members = Set<Member>(members)
         invitedMembers = Set<Member>()
-        let config = try container.decode(Config.self, forKey: .config)
+        let config = (try? container.decode(Config.self, forKey: .config)) ?? Config()
```
b-onc, line 200: Can you comment the reason, as put in the description?

VojtaStavik (commented):
I believe the missing config is a bug. Let's double-check with the backend team.

VojtaStavik (approved):
I think we can merge this. However, we should probably add a "known issue" note to the change log that it's possible that for a short period of time `message.channel.config` won't be up to date and the users should always use the config from channel query.

PR #744, opened 2021-01-14
V2: Fix table view not at bottom when last message is large

### LINKED ISSUE #653: ChatViewController.  Conversation screen isn't opened from its bottom if the last message is large.
When we open ChatViewController and the last message is long, then the chat isn't scrolled to the bottom initially. 

IMG_4479

There is a video in zip
[RPReplay_Final1608023660.MP4.zip](https://github.com/GetStream/stream-chat-swift/files/5694637/RPReplay_Final1608023660.MP4.zip)

## GetStream Environment
**iOS version:** 14
**Swift version:** 5
**Xcode version:** 12.2
**Device:** iPhone XS

gz#8052

cardoso:
When scroll to row command is for the last item, it's safer to scroll to the bottom of it (happens after view is loaded). When it's for another item, top is better (happens when loading more messages). 

Closes #653

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 # Upcoming
 
 ### 🐞 Fixed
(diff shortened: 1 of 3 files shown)


b-onc (approved):
CHANGELOG then 🚢 🇮🇹

PR #745, opened 2021-01-14
[CIS-431] Use Int~32~64 for database

VojtaStavik:
We used all variants of ints for storing numbers in the DB. To avoid overthinking in what kind of `IntXX` to use, let's use ~`Int32`~ `Int64` everywhere ( `Int16` can sometimes not be enough for certain cases - like our demo app).

We also convert any incoming number to ~`Int32`~ `Int64` safely using ~`Int32(clamping:)`~`Int34(clamping:)`. So even when the number (due to a bug on the backend for example) exceeds ~`Int32.max`~ `Int64.max`, we will just clamp it,


DominikBucher12 (approved):
I approve this, but honestly I think we can use `Int64` just to be sure and it doesn't cost us too much. Let's see...

PR #746, opened 2021-01-14
[CIS-575] Quoted messages

mantergo:
**In this PR:**

- Add possibility to quote message
- Rename `ChatReplyBubbleView` -> `MessageComposerQuoteBubbleView`
- Adjust logic and UI for displaying `quotedMessage`

Screenshot 2021-01-15 at 00 52 24


Sources_v3/StreamChatUI/ChatChannel/ChatMessage/ChatMessageQuoteBubbleView.swift
```diff
@@ -0,0 +1,44 @@
...
+    }
+    
+    override open func setUpLayout() {
+        super.setUpLayout()
+        
+        // In ChatMessage view is added to superview in `setUpLayout` but all constraint are activated in
+        // `updateContent`
+        // This workaround is preventing ambiguous layout when constraints for this view will be activated
+        // before superview constraints.
+        container.removeFromSuperview()
```
mantergo, line 22: Either I'm missing something or it's alarming.
Basically our views from composer and from message are not really compatible.

Sources_v3/StreamChatUI/ChatChannel/ChatMessage/ChatMessageQuoteBubbleView.swift
```diff
@@ -0,0 +1,44 @@
...
+    
+    override open func updateContent() {
+        super.updateContent()
+    
+        if message == nil {
+            container.removeFromSuperview()
+            return
+        } else if container.superview == nil {
+            embed(container)
+        }
```
mantergo, line 36: I'm putting it back here when constraint for superview will be activated 🙄

Sources_v3/StreamChatUI/ChatChannel/ChatChannelVC.swift
```diff
@@ -61,22 +61,7 @@ open class ChatChannelVC<ExtraData: ExtraDataTypes>: ChatVC<ExtraData> {
...
-        messageController.synchronize { [weak self, messageController] _ in
-            guard let self = self, let parentMessage = messageController.message else { return }
-            self.channelController(
-                self.channelController,
-                didUpdateMessages: [.update(parentMessage, index: IndexPath(item: index, section: 0))]
-            )
-        }
-        return nil
+        message.quotedMessageId.flatMap { channelController.dataStore.message(id: $0) }
     }
```
VojtaStavik, line 65: It seems that this whole method is redundant now. Let's remove it?

Sources_v3/StreamChatUI/ChatChannel/ChatMessage/ChatMessageQuoteBubbleView.swift
VojtaStavik: This is strange. `updateContent` is called immediately after `setUpLayout` for this exact reason. All lifecycle methods are called immediately one after each other.

VojtaStavik (changes requested):
Looking good but the strange things we do we adding/removing container in `ChatMessageQuoteBubleView` should be merged imo. Let's investigate and fix properly. Thanks!

Sources_v3/StreamChatUI/ChatChannel/ChatMessage/ChatMessageQuoteBubbleView.swift
```diff
@@ -0,0 +1,44 @@
...
+        
+        if isParentMessageSentByCurrentUser == true {
```
ManWithBear, line 38: Why are we showing incoming and outgoing messages in composer same way (colors and avatar positioning)?

mantergo: Colors are different - you can see it on the screenshot.
It's set inside `ChatMessageQuoteBubbleView`.
Regarding avatar position - there was a message from designer team that avatar should be always on the left.
Let me check if I can find it.

Sources_v3/StreamChatUI/ChatChannel/ChatChannelVC.swift
mantergo: We need it cause we need to have `channelController`: 
`.flatMap { channelController.dataStore.message(id: $0) }`.

Sources_v3/StreamChatUI/ChatChannel/ChatMessage/ChatMessageQuoteBubbleView.swift
mantergo: Seems like I confused it with different comment so thanks for spotting this 😉

Screenshot 2021-01-15 at 14 16 52

VojtaStavik (approved):

PR #747, opened 2021-01-15
[CIS-581] Introduce StreamTextStyles, fix minor UI issues in ChannelList

DominikBucher12:
# What this PR does:
- Introduces correct TextStyles according to design team, also fixes some minor UI issues in ChannelList
# How to test it
- Launch the application, play with the fontSize in Accessibility settings in iOS settings, probably wait for Design to finish their QA cycle :) 
# Where you can start
- `UIConfig.swift`, then all the files. 

NOTE: Maybe check `ChatMessageInteractiveAttachmentView` and tell me if the italic headline feels wrong, we probably need to introduce new font to our line-up.

### DIFF
--- a/Sources_v3/StreamChatUI/ChatChannel/Attachments/ChatMessageImageGallery.swift
+++ b/Sources_v3/StreamChatUI/ChatChannel/Attachments/ChatMessageImageGallery.swift
@@ -21,7 +21,7 @@ open class ChatMessageImageGallery<ExtraData: ExtraDataTypes>: View, UIConfigPro
 
     public private(set) lazy var moreImagesOverlay: UILabel = {
         let label = UILabel()
-        label.font = UIFont.preferredFont(forTextStyle: .largeTitle).bold
+        label.font = uiConfig.font.title
         label.adjustsFontForContentSizeCategory = true
         label.textAlignment = .center
         return label.withoutAutoresizingMaskConstraints
--- a/Sources_v3/StreamChatUI/ChatChannel/Attachments/ChatMessageInteractiveAttachmentView.swift
+++ b/Sources_v3/StreamChatUI/ChatChannel/Attachments/ChatMessageInteractiveAttachmentView.swift
@@ -22,7 +22,7 @@ open class ChatMessageInteractiveAttachmentView<ExtraData: ExtraDataTypes>: View
 
     public private(set) lazy var titleLabel: UILabel = {
         let label = UILabel()
-        label.font = UIFont.preferredFont(forTextStyle: .headline).italic
+        label.font = uiConfig.font.bodyItalic
         label.adjustsFontForContentSizeCategory = true
         label.textAlignment = .center
         return label.withoutAutoresizingMaskConstraints
--- a/Sources_v3/StreamChatUI/UIConfig.swift
+++ b/Sources_v3/StreamChatUI/UIConfig.swift
@@ -59,6 +59,8 @@ public extension UIConfig {
         public var channelListIndicatorBorderColor: UIColor = .streamWhiteSnow
         public var channelListActionDeleteChannel: UIColor = .streamAccentRed
         public var channelListAvatarOnlineIndicator: UIColor = .streamAccentGreen
+        public var channelListUnreadCountView: UIColor = .streamAccentRed
+        public var channelListUnreadCountLabel: UIColor = .streamWhite
 
         // MARK: - Text interactions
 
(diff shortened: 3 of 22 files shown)


Sources_v3/StreamChatUI/ChatChannel/Attachments/ChatMessageImageGallery.swift
```diff
@@ -21,7 +21,7 @@ open class ChatMessageImageGallery<ExtraData: ExtraDataTypes>: View, UIConfigPro
 
     public private(set) lazy var moreImagesOverlay: UILabel = {
         let label = UILabel()
-        label.font = UIFont.preferredFont(forTextStyle: .largeTitle).bold
+        label.font = .streamTitle // TODO: Not sure about this, check design for more fontStyles?
```
VojtaStavik, line 24: Just ask in #mobile-design in Slack, they usually reply very quickly

Sources_v3/StreamChatUI/ChatChannel/Attachments/ChatMessageInteractiveAttachmentView.swift
```diff
@@ -22,7 +22,7 @@ open class ChatMessageInteractiveAttachmentView<ExtraData: ExtraDataTypes>: View
 
     public private(set) lazy var titleLabel: UILabel = {
         let label = UILabel()
-        label.font = UIFont.preferredFont(forTextStyle: .headline).italic
+        label.font = UIFont.streamHeadline.italic // TODO: Check this what with italic? Create another font styles?
```
VojtaStavik, line 25: NOW is the time to check ;-)

VojtaStavik (approved):
Looks good! Let's ask designers now so we don't need to have another follow up PR for this :)

Sources_v3/StreamChatUI/UIConfig.swift
```diff
@@ -384,3 +384,16 @@ private extension UIColor {
         }
     }
 }
+
+// Typography defined in Figma design.
+// Because our design guidelines don't really fit the `prefferedFont`, we need to redaclare this.
+public extension UIFont {
```
ManWithBear, line 392: Do we want to expose our fonts to user? 
I don't like when 3rd parties pollute namespaces in my project.

Perhaps we could use same approach as with colors: private extension, usage via `UIConfig.ColorPalette`?
Or overkill?

b-onc: I'm onboard with this. It'll cover the usacase "I want all titles to look like this instead"

Sources_v3/StreamChatUI/ChatChannel/Attachments/ChatMessageImageGallery.swift
DominikBucher12: thanks, this is fixed now :)

Sources_v3/StreamChatUI/ChatChannel/Attachments/ChatMessageInteractiveAttachmentView.swift
DominikBucher12: thanks, this is fixed now :)

Sources_v3/StreamChatUI/UIConfig.swift
DominikBucher12: Thanks, for faster solution, I made it the same way as UIColor is done now. Otherwise as my 2️⃣ I will take a look how we can make `UIConfig` greater :)

Sources_v3/StreamChatUI/UIConfig.swift
```diff
@@ -106,6 +107,20 @@ public extension UIConfig {
...
+    struct Font {
+        public let captionBold: UIFont = .streamCaptionBold
+        public let footnote: UIFont = .streamFootnote
+        public let footnoteBold: UIFont = .streamFootnoteBold
+        public let body: UIFont = .streamBody
+        public let bodyBold: UIFont = .streamBodyBold
+        public let bodyItalic: UIFont = .streamBodyItalic
+        public let headline: UIFont = .streamHeadline
+        public let headlineBold: UIFont = .streamHeadlineBold
+        public let title: UIFont = .streamTitle
```
ManWithBear, line 122: `var` instead of `let`, otherwise it destroys whole point of `UIConfig`

ManWithBear (approved):
Please change `Font` properties to `var`s and we ready to go 👍🏾

Sources_v3/StreamChatUI/UIConfig.swift
DominikBucher12: Thanks :) This is done.

PR #748, opened 2021-01-15
[CIS-583] Resolve design issues in chat creation

b-onc:
All High and most of Mid issues resolved, with a sprinkle of Lows here and there


VojtaStavik (approved):
FRIDAY MERGEFEST! 🚢 🇮🇹

PR #751, opened 2021-01-21
Replace `DefaultExtraData` with `NoExtraData`

evsaev:
**This PR** 
- replaces `DefaultExtra` data with `NoExtraData` since by default there are no custom fields (thanks to the backend team and @b-onc, `imageURL/name` became model fields)
- uses `NoExtraData` instead of nested `NoExtraData.User/Channel/Message/Attachment`

There are NO documentation changes in this PR so we keep using `default extra data` wording in typealiases description. It looks fine since we didn't refer to the specific type (`DefaultExtraData`) anywhere.


Sources_v3/StreamChat/Models/ExtraData.swift
```diff
@@ -11,8 +11,14 @@ public protocol ExtraData: Codable & Hashable {
...
-public struct NoExtraData: Codable, Hashable, UserExtraData, ChannelExtraData, MessageExtraData, MessageReactionExtraData,
-    AttachmentExtraData {
+public struct NoExtraData: Codable,
+                           Hashable,
+                           UserExtraData,
+                           ChannelExtraData,
+                           MessageExtraData,
+                           MessageReactionExtraData,
+                           AttachmentExtraData,
+                           ExtraDataTypes {
```
evsaev, line 21: Implementing `ExtraDataTypes` by the type creates a recursive dependency when nested types can be accessed from any nested one since all of them are `NoExtraData` (e.g. `NoExtraData.User.Message.Channel`). Are we OK with that? (looks like something we don't want IMHO)

One way to avoid this is having concrete separate `XXXNoExtraData` types for each model which will implement just the `XXXExtraDataType`.

Sources_v3/StreamChat/APIClient/Endpoints/Payloads/GuestUserTokenPayload_Tests.swift
```diff
@@ -7,7 +7,6 @@ import XCTest
 
 final class GuestUserTokenPayload_Tests: XCTestCase {
     let guestUserDefaultExtraDataJSON = XCTestCase.mockData(fromFile: "GuestUser+DefaultExtraData")
-    let guestUserNoExtraDataJSON = XCTestCase.mockData(fromFile: "GuestUser+NoExtraData")
     let guestUserCustomExtraDataJSON = XCTestCase.mockData(fromFile: "GuestUser+CustomExtraData")
     let guestUserInvalidTokenJSON = XCTestCase.mockData(fromFile: "GuestUser+InvalidToken")
```
VojtaStavik, line 11: when `DefaultExtraData` became `NoExtraData`, should the `defaultExtraData` or `noExtraData` tests be removed?

image

VojtaStavik (approved):
Looks great, thank you!

Sources_v3/StreamChat/Models/ExtraData.swift
VojtaStavik: I think we can live with this for now 👍

PR #752, opened 2021-01-21
[CIS-582] Prevent unnecessary DB updates

b-onc:
Since User, Channel and Message composes of nearly all updated entities, this PR focuses on these 3.

### DIFF
--- a/Sources/StreamChat/APIClient/Endpoints/Payloads/UserPayloads.swift
+++ b/Sources/StreamChat/APIClient/Endpoints/Payloads/UserPayloads.swift
@@ -26,7 +26,7 @@ enum UserPayloadsCodingKeys: String, CodingKey {
 // MARK: - GET users
 
 /// An object describing the incoming user JSON payload.
-class UserPayload<ExtraData: UserExtraData>: Decodable {
+class UserPayload<ExtraData: UserExtraData>: Decodable, ChangeHashable {
     let id: String
     let name: String?
     let imageURL: URL?
@@ -39,6 +39,27 @@ class UserPayload<ExtraData: UserExtraData>: Decodable {
     let isBanned: Bool
     let teams: [String]
     let extraData: ExtraData
+    
+    var hasher: ChangeHasher {
+        UserHasher(
+            id: id,
+            name: name,
+            imageURL: imageURL,
+            role: role.rawValue,
+            createdAt: createdAt,
+            updatedAt: updatedAt,
+            lastActiveAt: lastActiveAt,
+            isOnline: isOnline,
+            isBanned: isBanned,
+            extraData: (try? JSONEncoder.default.encode(extraData)) ?? .init()
+        )
+    }
+    
+    // This needs to be explicit so we can override it in
+    // `UserDTO_Tests.test_DTO_skipsUnnecessarySave`
+    var changeHash: Int {
+        hasher.changeHash
+    }
     
     init(
         id: String,
--- a/Sources/StreamChat/Controllers/ListDatabaseObserver_Tests.swift
+++ b/Sources/StreamChat/Controllers/ListDatabaseObserver_Tests.swift
@@ -91,7 +91,7 @@ class ListDatabaseObserver_Tests: XCTestCase {
         super.setUp()
         
         fetchRequest = NSFetchRequest(entityName: "TestManagedObject")
-        fetchRequest.sortDescriptors = [.init(key: "id", ascending: true)]
+        fetchRequest.sortDescriptors = [.init(key: "testId", ascending: true)]
         database = try! DatabaseContainerMock(
             kind: .inMemory,
             modelName: "TestDataModel",
@@ -161,6 +161,45 @@ class ListDatabaseObserver_Tests: XCTestCase {
         assert(testFRC.test_performFetchCalled == false)
         try observer.startObserving()
         XCTAssertTrue(testFRC.test_performFetchCalled)
+    }
+    
+    func test_updateNotReported_whenSamePropertyAssignedViaAssignIfDifferent() throws {
+        // For this test, we need an actual NSFetchedResultsController, not the test one
+        let observer = ListDatabaseObserver<TestManagedObject, TestManagedObject>(
+            context: database.viewContext,
+            fetchRequest: fetchRequest,
+            itemCreator: { $0 }
+        )
+        
+        var receivedChanges: [ListChange<TestManagedObject>]?
+        observer.onChange = { receivedChanges = $0 }
+        
+        // Call startObserving to set everything up
+        try observer.startObserving()
+        
+        // Insert the test object
+        let testValue = String.unique
+        var item: TestManagedObject!
+        try database.writeSynchronously { _ in
+            let context = self.database.writableContext
+            item = NSEntityDescription.insertNewObject(forEntityName: "TestManagedObject", into: context) as? TestManagedObject
+            item.testId = testValue
+            item.testValue = testValue
+        }
+        
+        XCTAssertEqual(receivedChanges?.first?.item.testId, testValue)
+        
+        let oldChanges = receivedChanges
+        
+        // Assign the same testValue to the same entity
+        try database.writeSynchronously { _ in
+            // We use similar logic in DTO `save` methods
+            // to avoid unnecessary updates to properties
+            assignIfDifferent(item, \.testValue, testValue)
+        }
+        
+        // Assert no new change is reported
+        AssertAsync.staysEqual(receivedChanges, oldChanges)
     }
     
     func test_allItemsAreRemoved_whenDatabaseContainerRemovesAllData() throws {
--- a/Sources/StreamChat/Database/DTOs/ChannelDTO.swift
+++ b/Sources/StreamChat/Database/DTOs/ChannelDTO.swift
@@ -25,6 +25,8 @@ class ChannelDTO: NSManagedObject {
     
     @NSManaged var isFrozen: Bool
     
+    @NSManaged var savedChangeHash: Int64
+    
     // MARK: - Relationships
     
     @NSManaged var createdBy: UserDTO
@@ -63,6 +65,41 @@ class ChannelDTO: NSManagedObject {
         new.cid = cid.rawValue
         return new
     }
+    
+    override func willSave() {
+        super.willSave()
+        
+        // Save the current hash of the entity
+        // This is used during `save` calls to compare DTOs `changeHash` and
+        // corresponding payload's `changeHash` to avoid unnecessary assignment of values.
+        // Direct assignment cannot be used
+        // Since it'll generate new `willSave` calls causing recursion
+        assignIfDifferent(self, \.savedChangeHash, Int64(hasher.changeHash))
+    }
+}
+
+extension ChannelDTO: ChangeHashable {
+    var hasher: ChangeHasher {
+        ChannelHasher(
+            cid: cid,
+            name: name,
+            imageURL: imageURL,
+            extraData: extraData,
+            typeRawValue: typeRawValue,
+            lastMessageAt: lastMessageAt,
+            createdAt: createdAt,
+            deletedAt: deletedAt,
+            updatedAt: updatedAt,
+            createdByChangeHash: createdBy.changeHash,
+            config: config,
+            isFrozen: isFrozen,
+            memberCount: Int(memberCount)
+        )
+    }
+    
+    var changeHash: Int {
+        Int(savedChangeHash)
+    }
 }
 
 // MARK: - EphemeralValuesContainer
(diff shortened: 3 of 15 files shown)


Sources_v3/StreamChat/Database/DTOs/UserDTO.swift
```diff
@@ -21,13 +21,35 @@ class UserDTO: NSManagedObject {
...
+        hasher.combine(userUpdatedAt)
+        hasher.combine(lastActivityAt)
+        hasher.combine(isOnline)
+        // hasher.combine(isInvisible)
+        hasher.combine(isBanned)
+        // hasher.combine(teams)
+        hasher.combine(extraData)
+        // Direct assignment cannot be used
+        // Since it'll generate new `willSave` calls causing recursion
+        assignIfDifferent(self, \.payloadHash, Int64(hasher.finalize()))
```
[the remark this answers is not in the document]
b-onc: I wanted to convey "this should be the same as payload's hash with the same values" :D Since logically `dto.hashValue` won't be the same as `payload.hashValue`, they have differing properties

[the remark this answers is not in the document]
VojtaStavik: Interesting ... my mental model for this is that values `A` `B` `C` from DTO has the same hash as values `A` `B` `C` from the payload

Sources/StreamChat/Controllers/ListDatabaseObserver_Tests.swift
```diff
@@ -163,6 +163,43 @@ class ListDatabaseObserver_Tests: XCTestCase {
...
+        
+        // Assign the same testValue to the same entity
+        try database.writeSynchronously { _ in
+            // We use this function instead of direct assignment (`=`) in our `save` methods
+            assignIfDifferent(item, \.testValue, testValue)
+        }
+        
+        if case .update = receivedChanges?.first {
+            XCTFail("Update shouldn't be reported!")
+        }
```
VojtaStavik, line 200: It think it's safer to use `AssertAsync.staysEqual()` here

Sources/StreamChat/APIClient/Endpoints/Payloads/UserPayloads.swift
```diff
@@ -40,6 +40,27 @@ class UserPayload<ExtraData: UserExtraData>: Decodable {
...
+        hasher.combine(lastActiveAt)
+        hasher.combine(isOnline)
+        // hasher.combine(isInvisible)
+        hasher.combine(isBanned)
+        // hasher.combine(teams)
+        if let extraData = try? JSONEncoder.default.encode(extraData) {
+            hasher.combine(extraData)
+        }
+    }
+
```
VojtaStavik, line 63: There are some big statements here. Implementing `==` just as `id == id` in terms of `UserPayload` is IMO not correct. This is a payload describing an entity, not the entity itself (where `id == id` makes more sense).

I think we shouldn't conform to hashable explicitly unless it's a _true_ hash -> like the one generated by the compiler.

Isn't all we need just some kind of method like `computeHashForDTOSerializing()`?

Sources/StreamChat/Database/DTOs/UserDTO.swift
```diff
@@ -21,13 +21,35 @@ class UserDTO: NSManagedObject {
...
+        hasher.combine(lastActivityAt)
+        hasher.combine(isOnline)
+        // hasher.combine(isInvisible)
+        hasher.combine(isBanned)
+        // hasher.combine(teams)
+        hasher.combine(extraData)
+        // Direct assignment cannot be used
+        // Since it'll generate new `willSave` calls causing recursion
+        assignIfDifferent(self, \.payloadHash, Int64(hasher.finalize()))
+    }
```
VojtaStavik, line 52: ... or even better - what if we created something like `struct UserHasher: Hashable { ... }` and use it just to get the hash from that? When a new field is added, it's initializer will help us to add it everywhere where needed.

VojtaStavik (changes requested):
The general idea is good 👍 I think we could execute it with less code and in a more compile-time safe way.

Sources/StreamChat/APIClient/Endpoints/Payloads/UserPayloads.swift
b-onc: Very good fundemantel remarks here. I also came to the realization that `Hashable` is actually not correct in this case. We're using hashes, but for our own purpose. I think we need to have our own protocol, so both DTO and payload and implement it.

Sources/StreamChat/Database/DTOs/UserDTO.swift
b-onc: actually, great idea

Sources/StreamChat/Controllers/ListDatabaseObserver_Tests.swift
```diff
@@ -163,6 +163,47 @@ class ListDatabaseObserver_Tests: XCTestCase {
...
+        try database.writeSynchronously { _ in
+            // We use this function instead of direct assignment (`=`) in our `save` methods
+            assignIfDifferent(item, \.testValue, testValue)
+        }
+        
+        AssertAsync.staysEqual(receivedChanges, oldChanges)
+        
+        if case .update = receivedChanges?.first {
+            XCTFail("Update shouldn't be reported!")
+        }
```
VojtaStavik, line 204: Isn't this assertion redundant when you already assert `staysEqual`?

Sources/StreamChat/Utils/ChatHashable.swift
```diff
@@ -0,0 +1,137 @@
+//
+// Copyright © 2021 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+protocol ChatHasher {
```
VojtaStavik, line 7: I know this is not public API but some kind of basic documentation for our future selves would be nice.

Sources/StreamChat/APIClient/Endpoints/Payloads/UserPayloads.swift
```diff
@@ -40,6 +40,27 @@ class UserPayload<ExtraData: UserExtraData>: Decodable {
...
+            extraData: (try? JSONEncoder.default.encode(extraData)) ?? .init()
+        )
+    }
+    
+    // This needs to be explicit so we can override it in
+    // `UserDTO_Tests.test_DTO_skipsUnnecessarySave`
+    var chatHash: Int {
+        hasher.chatHash
+    }
+
```
VojtaStavik, line 63: Again, not a public API but no one will know why we have this 2 months from now.

Sources/StreamChat/APIClient/Endpoints/Payloads/UserPayloads.swift
```diff
@@ -40,6 +40,27 @@ class UserPayload<ExtraData: UserExtraData>: Decodable {
...
+            lastActiveAt: lastActiveAt,
+            isOnline: isOnline,
+            isBanned: isBanned,
+            extraData: (try? JSONEncoder.default.encode(extraData)) ?? .init()
+        )
+    }
+    
+    // This needs to be explicit so we can override it in
+    // `UserDTO_Tests.test_DTO_skipsUnnecessarySave`
+    var chatHash: Int {
```
VojtaStavik, line 60: nit: why `ChatHash`? :) Isn't' everything "Chat" when we're building a chat SDK? 🤔 `changesHash` `versionHash`?

Sources/StreamChat/Database/DTOs/UserDTO.swift
```diff
@@ -21,13 +21,44 @@ class UserDTO: NSManagedObject {
...
         request.predicate = NSPredicate(format: "id == %@", userId)
         return request
     }
+    
+    override func willSave() {
+        super.willSave()
+        
+        // Direct assignment cannot be used
+        // Since it'll generate new `willSave` calls causing recursion
+        assignIfDifferent(self, \.savedChatHash, Int64(hasher.chatHash))
```
VojtaStavik, line 39: but @b-onc , why are we doing this? A comment would be so nice

Sources/StreamChat/Database/DTOs/UserDTO_Tests.swift
```diff
@@ -141,6 +141,108 @@ class UserDTO_Tests: XCTestCase {
...
+        XCTAssertNotEqual(changedPayload.lastActiveAt, loadedUserModel?.lastActiveAt)
+        
+        let newPayload: UserPayload<NoExtraData> = .dummy(userId: userId)
+        
+        // Save the changed payload with the same hash
+        try database.writeSynchronously { session in
+            try session.saveUser(payload: newPayload)
+        }
+        
+        XCTAssertEqual(loadedUserModel?.name, newPayload.name)
```
VojtaStavik, line 243: `// Save the changed payload with the same hash` ... and then what? :D Please don't make me think when reading tests. Just explain to me what you're doing in the comments, thanks!

Sources/StreamChat/Database/DTOs/ChannelDTO_Tests.swift
```diff
@@ -124,6 +124,29 @@ class ChannelDTO_Tests: XCTestCase {
...
+        }
+        
+        // Assert that hash is not changed
+        guard let dtoHash = loadedChannel?.chatHash else {
+            XCTFail("DTO is missing hash!")
+            return
+        }
+        XCTAssertEqual(Int(dtoHash), payload.chatHash)
+    }
+
```
VojtaStavik, line 149: Isn't there a missing test here? The one that actually checks the DTO is not saved when not needed? It seems to me that if I remove `if dto.chatHash != payload.chatHash {` from the DTO code everything will keep being green.

VojtaStavik (changes requested):
Looks much better! I desperately miss more documentation. This is an extremely uncommon and very custom feature. We should be sure to document everything very precisely.

Also, did you make some performance measurements we talked about? (i.e. saving the same payload 500x). What's the effect of this change?

Sources/StreamChat/APIClient/Endpoints/Payloads/UserPayloads.swift
b-onc: My mind slipped to our own custom hash (instead of built-in, well known hash), another approach of mine was `updateHash` but I like `changesHash` too

Sources/StreamChat/Database/DTOs/ChannelDTO_Tests.swift
b-onc: Yes, since ChannelDetailPayload is a struct (Message and User payloads are classes) I couldn't subclass them to introduce `explicitHash`, I got stuck how to write a similar test tbh

Sources/StreamChat/Controllers/ListDatabaseObserver_Tests.swift
b-onc: fixed in e1c35788df1afd10341996a90a43d21e32d4c39d

Sources/StreamChat/Utils/ChatHashable.swift
b-onc: fixed in d1a4ed6f97a6a61cec7487f6fe07b07120e78617

Sources/StreamChat/APIClient/Endpoints/Payloads/UserPayloads.swift
b-onc: fixed in d1a4ed6f97a6a61cec7487f6fe07b07120e78617

Sources/StreamChat/APIClient/Endpoints/Payloads/UserPayloads.swift
b-onc: fixed in d1a4ed6f97a6a61cec7487f6fe07b07120e78617

Sources/StreamChat/Database/DTOs/UserDTO_Tests.swift
b-onc: fixed in 293c1ee258489fb76a6981f2a809decb044453a0

Sources/StreamChat/Database/DTOs/ChannelDTO_Tests.swift
b-onc: Suggestions welcome!

VojtaStavik (approved):

b-onc:
@VojtaStavik performance test:
```swift
    let otherUserJSON = XCTestCase.mockData(fromFile: "OtherUser")
    var database: DatabaseContainerMock!

    override func setUpWithError() throws {
        try super.setUpWithError()
        database = try DatabaseContainerMock(kind: .onDisk(databaseFileURL: .newTemporaryFileURL()))
    }

    func testPerformance() throws {
        self.measure {
            do {
                let payload = try JSONDecoder.default.decode(UserPayload<NoExtraData>.self, from: otherUserJSON)
                
                for _ in 0...1000 {
                    try database.writeSynchronously { session in
                        try session.saveUser(payload: payload)
                    }
                }
            } catch {
                XCTFail("\(error)")
            }
        }
    }
```

Result
Before PR: Average 0.640 sec
After PR: Average 0.368 sec
**%42 improvement**

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
```diff
@@ -81,23 +118,26 @@ extension NSManagedObjectContext {
...
-        dto.lastMessageAt = payload.lastMessageAt
-        dto.memberCount = Int64(clamping: payload.memberCount)
-        
-        dto.isFrozen = payload.isFrozen
-        
-        if let createdByPayload = payload.createdBy {
-            let creatorDTO = try saveUser(payload: createdByPayload)
-            dto.createdBy = creatorDTO
+        
+        if dto.changeHash != payload.changeHash {
```
nuno-vieira, line 122: Do we really need to come up with this logic ourselves? Doesn't `hasChanges` from core data check if all properties weren't changed? @b-onc

PR #755, opened 2021-01-22
Introduce `register_new_device` lane

evsaev:


fastlane/Fastfile
```diff
@@ -102,6 +102,27 @@ end
...
+  )
+
+  run_match_me(readonly: false)
+end
+
+# Runs `match` for development and adhoc profiles of iOS apps/frameworks.
+# If `readonly: true`, it just donwloads the certs and profiles
+# If `readonly: false`, certs and profiles will be recreated, the new ones then
+# will be downloaded locally and automatically pushed to git repo.
+def run_match_me(readonly: true)
```
evsaev, line 125: The `readonly` can also be a `lane match` option.

evsaev: There was an idea to name this helper based on what it does (recreate_profiles...) but the name appeared to be long and the `fastlane` terminology was lost

b-onc: `readonly` being a `lane match_me` option with defaulting to `true` makes more sense

fastlane/Fastfile
```diff
@@ -102,6 +102,27 @@ end
...
+  device_name = prompt(text: "Enter the device name: ")
+  device_udid = prompt(text: "Enter the device UDID: ")
+
+  register_device(
+    name: device_name,
+    udid: device_udid
+  )
+
+  run_match_me(readonly: false)
+end
```
b-onc, line 119: I like this lane! However the name should reflect that profiles will be recreated

VojtaStavik (commented):
Great idea. I let @b-onc to review since this is his area of expertise 👍

fastlane/Fastfile
```diff
@@ -102,6 +102,27 @@ end
 
 desc "Installs all Certs and Profiles necessary for development and ad-hoc"
 lane :match_me do
+  run_match_me(readonly: true)
+end
+
+desc "Register new device, regenerates profiles, updates them remotely and locally"
+lane :register_new_device do
+  device_name = prompt(text: "Enter the device name: ")
+  device_udid = prompt(text: "Enter the device UDID: ")
```
DominikBucher12, line 111: If you wanna be fancy, you can print `xcrun xctrace list devices ` where this lists connected devices (and simulators :( ) with their UDIDs for faster copy-past action

evsaev: I have an infinite list of simulators and getting the device UDID is simpler by Finder, IMO 🤔
@b-onc WDYT?

DominikBucher12: Okay, you can do this:

`xcrun instruments -s devices | grep -v Simulator`, this will show only physical devices, however xcrun instruments is deprecated now...

PR #757, opened 2021-01-22
Run v2 Actions only on v2 PRs

VojtaStavik:
- v2 actions now run only when the target branch is `main_v2`
- we no longer run nightly tests for v2. We don't expect having many v2 PRs so it's OK when the CI runs a bit longer for them


.github/workflows/ci_v2.yml
```diff
@@ -3,17 +3,10 @@ name: CI v2
...
+      - main_v2
   pull_request:
     branches:
-      - main
-    paths:
-      - 'Sources/**' # Only run if a file in /Sources changes (v2 folder)
-  release:
-    types:
-      - created
+      - main_m2
```
b-onc, line 9: typo

VojtaStavik: thanks! fixed

PR #758, opened 2021-01-22
Make main v2 free

b-onc:
Tried to keep commits as separate as possible...


CHANGELOG.md
```diff
@@ -6,829 +6,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Upcoming
 
-### 🔄 Changed
-
```
VojtaStavik, line 10: I don't think we should remove the old stuff! Just add a new version :)

VojtaStavik (approved):
🚢 🇮🇹

CHANGELOG.md
b-onc: it'll be interesting then, old CHANGELOG will continue (2.6.4, 2.7, 2.7.1....) but this changelog will jump from 2.6.2 -> 3.0.0

VojtaStavik: All right with me 👍

PR #760, opened 2021-01-25
Fix development token

cardoso:
As discussed in Slack

### DIFF
--- a/Sources/StreamChat/Config/TokenProvider_Tests.swift
+++ b/Sources/StreamChat/Config/TokenProvider_Tests.swift
@@ -25,8 +25,10 @@ final class TokenProvider_Tests: StressTestCase {
         }.get()
 
         // Assert token is correct.
-        XCTAssertEqual(token.rawValue, "development")
-        XCTAssertEqual(token.userId, userId)
+        let tokenParts = token.rawValue.split(separator: ".")
+        XCTAssertEqual(tokenParts[0], "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9") // header
+        XCTAssertEqual(token.userId, userId) // payload
+        XCTAssertEqual(tokenParts[2], "devtoken") // signature
     }
 
     func test_staticProvider_propagatesToken() throws {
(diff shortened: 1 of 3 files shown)


Sources/StreamChat/Config/TokenProvider_Tests.swift
```diff
@@ -17,15 +17,16 @@ final class TokenProvider_Tests: StressTestCase {
 
     func test_developmentProvider_propagatesToken() throws {
         // Create a user identifier.
-        let userId: UserId = .unique
+        let userId: UserId = "luke_skywalker"
```
VojtaStavik, line 20: Can we keep the `.unique` here and rather check only for the static parts of the token string explicitly?

VojtaStavik (changes requested):

VojtaStavik:
@cardoso Since v3 is public now, you don't need to add the `v3` label + please add a changelog entry.

VojtaStavik (approved):
LGTM 🎉 Thank you for fixing this @cardoso 🙏

Sources/StreamChat/Config/TokenProvider_Tests.swift
```diff
@@ -17,15 +17,16 @@ final class TokenProvider_Tests: StressTestCase {
...
 
         // Get a token from `development` token provider.
         let token = try await {
             TokenProvider.development(userId: userId).getToken(.mock, $0)
         }.get()
 
         // Assert token is correct.
-        let expectedToken = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1c2VyX2lkIjoibHVrZV9za3l3YWxrZXIifQ==.devtoken"
+        let expectedPayload = #"{"user_id":"\#(userId)"}"# .data(using: .utf8)!.base64EncodedString()
+        let expectedToken = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.\(expectedPayload).devtoken"
```
VojtaStavik, line 29: (feel free to ignore) nit: The `user_id` payload is already covered by `XCTAssertEqual(token.userId, userId)`. Here you could do just:
```swift
let tokenParts = token.split(separator: ".")
XCTAssertEqual(tokenParts[0], "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9")
XCTAssertEqual(tokenParts[2], "devtoken")
```

VojtaStavik:
Let's wait with merging this until we sort our the situation with CHANGELOGs

PR #761, opened 2021-01-26
Move code from setupAppearance to defaultAppearance

DominikBucher12:
# What this PR do:
- Moves remaining code from `setupAppearance` to `defaultAppearance` inside SDK so the changes happen in correct functions across the SDK


nuno-vieira (approved):
LGTM!

VojtaStavik (approved):
Is this the right time to add a custom SwiftLint rule matching `override public func defaultAppearance()`?

PR #762, opened 2021-01-26
[CIS-589] UI config

DominikBucher12:
# What this PR do:
1. Improves UIConfig structure by separating concerns in files with extensions
2. Introduces Images struct in UIConfig  so it's easier to change the default ones 
3. Replaces images in application for the ones defined in UIConfig
3. Renames colors to smaller palette so it's non-component-based and easier to change

# Where you can start
-  `UIConfig+Anything.swift`


DemoApp/ChatPresenter.swift
```diff
@@ -21,6 +21,7 @@ extension UIViewController {
         // Config
         var uiConfig = UIConfig<NoExtraData>()
         uiConfig.navigation.channelListRouter = DemoChatChannelListRouter.self
+        UIConfig.default = uiConfig
```
VojtaStavik, line 24: there three lines are not correct. It should be just:
```swift
UIConfig.default.navigation.channelListRouter = DemoChatChannelListRouter.self
```

Sources/StreamChatUI/UIConfig+ColorTheme.swift
```diff
@@ -0,0 +1,90 @@
...
+        public var unselectedColorForColor: (UIColor) -> UIColor = { _ in .lightGray }
+
+        // MARK: - Background
+
+        public var background: UIColor = .streamWhiteSnow
+        public var background1: UIColor = .streamWhiteSmoke
+        public var background2: UIColor = .streamGrayGainsboro
+        public var background3: UIColor = .streamOverlay
+        public var background4: UIColor = .streamOverlayDark
+        public var background5: UIColor = .streamWhite
```
VojtaStavik, line 32: I actually kinda like this pattern but we should just be sure the naming makes sense, i.e. `background1` is lighter than `background2` etc ...

Sources/StreamChatUI/UIConfig+ColorTheme.swift
```diff
@@ -0,0 +1,90 @@
...
+
+        public var popoverBackground: UIColor = .streamWhite
```
VojtaStavik, line 34: how is this different from `background5`?

Sources/StreamChatUI/UIConfig+ColorTheme.swift
DominikBucher12: Actually this kinda magically works. i.e. If you look into our designs, it is exactly like you say (1 is lighter than 2 etc.)

DemoApp/ChatPresenter.swift
DominikBucher12: Done.

Sources/StreamChatUI/UIConfig+ColorTheme.swift
DominikBucher12: Deleted `background5`

Sources/StreamChatUI/UIConfig+MessageList.swift
```diff
@@ -0,0 +1,97 @@
...
+            СhatIncomingMessageCollectionViewCell<ExtraData>.self
+        public var outgoingMessageCell: _СhatMessageCollectionViewCell<ExtraData>.Type =
+            СhatOutgoingMessageCollectionViewCell<ExtraData>.self
+
+        public var collectionView: ChatMessageListCollectionView.Type = ChatMessageListCollectionView.self
+        public var collectionLayout: ChatMessageListCollectionViewLayout.Type = ChatMessageListCollectionViewLayout.self
+        public var channelNamer: ChatChannelNamer.Type = ChatChannelNamer.self
+        public var minTimeInvteralBetweenMessagesInGroup: TimeInterval = 10
+        /// Vertical contentOffset for message list, when next message batch should be requested
+        public var offsetToPreloadMoreMessages: CGFloat = 100
```
VojtaStavik, line 21: If we say that the rule is that only contestants that are shared between multiple views should be in `UIConfig`, I think these two probably shouldn't be here.

Sources/StreamChatUI/UIConfig+MessageList.swift
```diff
@@ -0,0 +1,97 @@
...
+        public var fileAttachmentActionIcons: [LocalAttachmentState?: UIImage] = [
+            .uploaded: UIImage(named: "uploaded", in: .streamChatUI)!,
+            .uploadingFailed: UIImage(named: "restart", in: .streamChatUI)!,
+            nil: UIImage(named: "download_and_open", in: .streamChatUI)!
+        ]
+        // Images
+        public var imageGallery: _ChatMessageImageGallery<ExtraData>.Type = _ChatMessageImageGallery<ExtraData>.self
+        public var imageGalleryItem: _ChatMessageImageGallery<ExtraData>.ImagePreview.Type =
+            _ChatMessageImageGallery<ExtraData>.ImagePreview.self
+        public var imageGalleryInteritemSpacing: CGFloat = 2
```
VojtaStavik, line 82: No spacings please

Sources/StreamChatUI/UIConfig+MessageList.swift
```diff
@@ -0,0 +1,97 @@
...
+        public var interactiveAttachmentView: _ChatMessageInteractiveAttachmentView<ExtraData>.Type =
+            _ChatMessageInteractiveAttachmentView<ExtraData>.self
+        public var interactiveAttachmentActionButton: _ChatMessageInteractiveAttachmentView<ExtraData>.ActionButton.Type =
+            _ChatMessageInteractiveAttachmentView<ExtraData>.ActionButton.self
+        // Giphy
+        public var giphyAttachmentView: _ChatMessageGiphyView<ExtraData>.Type =
+            _ChatMessageGiphyView<ExtraData>.self
+        public var giphyBadgeView: _ChatMessageGiphyView<ExtraData>.GiphyBadge.Type = _ChatMessageGiphyView<ExtraData>.GiphyBadge
+            .self
+        public var giphyBadgeIcon: UIImage? = UIImage(named: "icon_giphy", in: .streamChatUI)
```
VojtaStavik, line 95: Images should be used directly

Sources/StreamChatUI/UIConfig+MessageList.swift
```diff
@@ -0,0 +1,97 @@
...
+            .Type = _ChatMessageFileAttachmentListView<ExtraData>.self
+        public var fileAttachmentItemView: _ChatMessageFileAttachmentListView<ExtraData>.ItemView.Type =
+            _ChatMessageFileAttachmentListView<ExtraData>.ItemView.self
+        public var fileFallbackIcon = UIImage(named: "generic", in: .streamChatUI)!
+        public var fileIcons = [AttachmentFileType: UIImage](
+            uniqueKeysWithValues: AttachmentFileType.allCases.compactMap {
+                guard let icon = UIImage(named: $0.rawValue, in: .streamChatUI) else { return nil }
+                return ($0, icon)
+            }
+        )
```
VojtaStavik, line 72: Isn't this already defined in `documentPreviews`?

Sources/StreamChatUI/UIConfig+MessageList.swift
```diff
@@ -0,0 +1,97 @@
...
+        public var fileAttachmentActionIcons: [LocalAttachmentState?: UIImage] = [
+            .uploaded: UIImage(named: "uploaded", in: .streamChatUI)!,
+            .uploadingFailed: UIImage(named: "restart", in: .streamChatUI)!,
+            nil: UIImage(named: "download_and_open", in: .streamChatUI)!
```
VojtaStavik, line 76: These images should be probably defined in `Images`

VojtaStavik (approved):
Looks good, just a couple of forgotten things. Totally understandable given the size of these changes. Great work @DominikBucher12 ! 🎉 

🚢 🇮🇹 

---

Did you apply the latest changes by @mantergo? Isn't this overriding his font changes?

Sources/StreamChatUI/UIConfig+Images.swift
```diff
@@ -0,0 +1,197 @@
...
+        public var messageActionDelete: UIImage = UIImage(named: "icn_delete", in: .streamChatUI)!
+        public var messageActionResend: UIImage = UIImage(named: "icn_resend", in: .streamChatUI)!
+
+        // MARK: - Placeholders
+
+        public var userAvatarPlaceholder1: UIImage = UIImage(named: "pattern1", in: .streamChatUI)!
+        public var userAvatarPlaceholder2: UIImage = UIImage(named: "pattern2", in: .streamChatUI)!
+        public var userAvatarPlaceholder3: UIImage = UIImage(named: "pattern3", in: .streamChatUI)!
+        public var userAvatarPlaceholder4: UIImage = UIImage(named: "pattern4", in: .streamChatUI)!
+        public var userAvatarPlaceholder5: UIImage = UIImage(named: "pattern5", in: .streamChatUI)!
```
nuno-vieira, line 138: Just like we have the views separated by structs, can't we have the images too? Like `Image.files`, `Image.reactions`, `Images.messageAction`. Just to avoid having a Huge `Images` struct.

VojtaStavik: I see what you mean @nuno-vieira but I'd suggest a small modification. Since this should serve as the main pool of images you can use across the app, rather than organizing it by "where it's used" it can be organized by "what is it". So like `Images.icons`, `Images.reactions`, or simply just `Images.backButton` for images without caterogy.

Sources/StreamChatUI/UIConfig+MessageList.swift
DominikBucher12: They are already there, just didn't delete this.  Now it's done.

Sources/StreamChatUI/UIConfig+MessageList.swift
DominikBucher12: Okay somehow when recreating the commit this stayed in there. This should definitely not be there but it's quite different, documentPreviews do not have the dictionary key of `AttachmentFileType`... I moved this to `UIConfig+Images` too it's a different case a bit.

Sources/StreamChatUI/UIConfig+MessageList.swift
DominikBucher12: Fixed.

Sources/StreamChatUI/UIConfig+MessageList.swift
DominikBucher12: Deleted.

Sources/StreamChatUI/UIConfig+MessageList.swift
DominikBucher12: Fixed. Thanks :)

PR #768, opened 2021-01-27
[CIS-603] Fix stability issues when reordering channel list

nuno-vieira:
# In this PR
- Fixes a crash when reordering the channel lists that was caused by moving and updating the same IndexPath in the collection view.
- When creating a new message, the order of the channel list is now updated instantly

### DIFF
--- a/Sources/StreamChat/Controllers/ListDatabaseObserver.swift
+++ b/Sources/StreamChat/Controllers/ListDatabaseObserver.swift
@@ -1,5 +1,5 @@
 //
-// Copyright © 2020 Stream.io Inc. All rights reserved.
+// Copyright © 2021 Stream.io Inc. All rights reserved.
 //
 
 import CoreData
@@ -292,6 +292,23 @@ class ListChangeAggregator<DTO: NSManagedObject, Item>: NSObject, NSFetchedResul
     }
     
     func controllerDidChangeContent(_ controller: NSFetchedResultsController<NSFetchRequestResult>) {
+        // All destination indices of `move` changes
+        let moveToIndexChanges: [IndexPath] = currentChanges.compactMap {
+            if case let .move(_, _, toIndex) = $0 {
+                return toIndex
+            }
+            return nil
+        }
+
+        // Remove `update` operations with the same index path as move's `toIndex`changes.
+        currentChanges = currentChanges.filter {
+            if case let .update(_, index) = $0 {
+                // Include only if the update `index` is not a `move` change destination index.
+                return moveToIndexChanges.contains(index) == false
+            }
+            return true
+        }
+        
         onChange?(currentChanges)
     }
 }
--- a/Sources/StreamChat/Controllers/ListDatabaseObserver_Tests.swift
+++ b/Sources/StreamChat/Controllers/ListDatabaseObserver_Tests.swift
@@ -420,6 +420,58 @@ class ListChangeAggregator_Tests: XCTestCase {
             .update(updatedObject.uniqueValue, index: [2, 0])
         ])
     }
+    
+    func test_controllerWillChangeContent_whenUpdatesAndMovesWithSameIndexPath_removeThoseUpdates() {
+        var result: [ListChange<String>]?
+        aggregator.onChange = { result = $0 }
+        
+        // Simulate FRC starts updating
+        aggregator.controllerWillChangeContent(fakeController)
+        
+        let addedObject = TestManagedObject()
+        let movedObject = TestManagedObject()
+        let updatedObject = TestManagedObject()
+        
+        aggregator.controller(
+            fakeController,
+            didChange: addedObject,
+            at: nil,
+            for: .insert,
+            newIndexPath: [1, 0]
+        )
+        
+        aggregator.controller(
+            fakeController,
+            didChange: updatedObject,
+            at: [2, 0],
+            for: .update,
+            newIndexPath: nil
+        )
+        
+        aggregator.controller(
+            fakeController,
+            didChange: movedObject,
+            at: [4, 0],
+            for: .move,
+            newIndexPath: [2, 0]
+        )
+        
+        aggregator.controller(
+            fakeController,
+            didChange: updatedObject,
+            at: [2, 0],
+            for: .update,
+            newIndexPath: nil
+        )
+        
+        // Simulate FRC finishes updating
+        aggregator.controllerDidChangeContent(fakeController)
+        
+        XCTAssertEqual(result, [
+            .insert(addedObject.uniqueValue, index: [1, 0]),
+            .move(movedObject.uniqueValue, fromIndex: [4, 0], toIndex: [2, 0])
+        ])
+    }
 }
 
 class TestFetchedResultsController: NSFetchedResultsController<TestManagedObject> {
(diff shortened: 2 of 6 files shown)


Sources/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -239,6 +239,10 @@ extension NSManagedObjectContext: MessageDatabaseSession {
             parentMessageDTO.replies.insert(message)
         }
         
+        channelDTO.lastMessageAt = createdDate
+        channelDTO.updatedAt = createdDate
+        channelDTO.defaultSortingAt = createdDate
```
VojtaStavik, line 244: I just realized @nuno-vieira , this is not a wrong approach. When you send a message and you're offline the channel should still jump to the topmost position. So we actually really need something like this 👍 

Please add a comment there explainig why we do this + tests

Sources/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -239,6 +239,10 @@ extension NSManagedObjectContext: MessageDatabaseSession {
             parentMessageDTO.replies.insert(message)
         }
         
+        channelDTO.lastMessageAt = createdDate
+        channelDTO.updatedAt = createdDate
```
VojtaStavik, line 243: The `updatedAt` shouldn't be changed. I think this value doesn't change with new messages, it's related to members and channel details.

Sources/StreamChat/Controllers/ListDatabaseObserver.swift
```diff
@@ -272,11 +272,25 @@ class ListChangeAggregator<DTO: NSManagedObject, Item>: NSObject, NSFetchedResul
...
                 return
             }
+            
+            // If a move already exists, do not append an update change
+            let updateAlreadyExists = currentChanges.contains(where: { change -> Bool in
+                guard case let .move(_, _, indexPath) = change else { return false }
+                return index == indexPath
+            })
+            guard !updateAlreadyExists else { return }
+
```
VojtaStavik, line 293: Instead of doing it twice, wouldn't it be easier to do it once as the final step before publishing the changes?

Sources/StreamChat/Controllers/ListDatabaseObserver_Tests.swift
```diff
@@ -420,6 +420,96 @@ class ListChangeAggregator_Tests: XCTestCase {
             .update(updatedObject.uniqueValue, index: [2, 0])
         ])
     }
+    
+    func test_removeUpdateWhenMoveWithSameIndexPath() {
```
VojtaStavik, line 424: Please make sure you follow the correct test naming pattern.

Sources/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -233,6 +233,8 @@ extension NSManagedObjectContext: MessageDatabaseSession {
         
         message.user = currentUserDTO.user
         message.channel = channelDTO
+        channelDTO.lastMessageAt = createdDate
+        channelDTO.defaultSortingAt = createdDate
```
VojtaStavik, line 237: I asked you to add a comment here explaining why we're doing this.

Sources/StreamChat/Database/DTOs/MessageDTO_Tests.swift
```diff
@@ -627,6 +627,61 @@ class MessageDTO_Tests: XCTestCase {
         XCTAssert(result is ClientError.ChannelDoesNotExist)
     }
     
+    func test_creatingNewMessage_shouldUpdateChannelDatesToTriggerSorting() throws {
```
VojtaStavik, line 630: -> `test_creatingNewMessage_updatesRelatedChannelFields`

Sources/StreamChat/Database/DTOs/MessageDTO_Tests.swift
```diff
@@ -627,6 +627,61 @@ class MessageDTO_Tests: XCTestCase {
...
+                try session.saveCurrentUser(payload: .dummy(
+                    userId: currentUserId,
+                    role: .admin,
+                    extraData: NoExtraData.defaultValue
+                ))
+                
+                try session.saveChannel(payload: self.dummyPayload(with: cid))
+                
+            }, completion: completion)
+        }
```
VojtaStavik, line 646: You can just use `database.writeSynchronously { }`

Sources/StreamChat/Database/DTOs/MessageDTO_Tests.swift
```diff
@@ -627,6 +627,61 @@ class MessageDTO_Tests: XCTestCase {
...
+        var newMessageId: MessageId!
+        
+        let newMessageText: String = .unique
+        let newMessageCommand: String = .unique
+        let newMessageArguments: String = .unique
+        let newMessageParentMessageId: String = .unique
+        let newMessageAttachmentSeeds: [_ChatMessageAttachment<NoExtraDataTypes>.Seed] = [
+            .dummy(),
+            .dummy()
+        ]
```
VojtaStavik, line 658: Why bother we these parameters when they are optional and they are not part of the test?

Sources/StreamChat/Database/DTOs/MessageDTO_Tests.swift
```diff
@@ -627,6 +627,61 @@ class MessageDTO_Tests: XCTestCase {
...
+                
+        _ = try await { completion in
+            database.write({
```
VojtaStavik, line 661: ... `writeSynchronously { }`

Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
```diff
@@ -114,19 +114,24 @@ extension ChatChannelListVC: _ChatChannelListControllerDelegate {
...
+                    case let .remove(_, index):
+                        collectionView.deleteItems(at: [index])
+                    case let .update(_, index):
+                        collectionView.reloadItems(at: [index])
+                    }
                 }
-            }
-        }, completion: nil)
+            },
+            completion: { _ in self.collectionView.reloadItems(at: movedItems) }
```
VojtaStavik, line 134: Please add a comment explaining why we're doing this.

VojtaStavik (changes requested):

Sources/StreamChat/Controllers/ListDatabaseObserver.swift
```diff
@@ -292,6 +292,34 @@ class ListChangeAggregator<DTO: NSManagedObject, Item>: NSObject, NSFetchedResul
...
+                if updateAlreadyExists {
+                    continue
+                }
+            }
+            
+            changesWithoutDuplicateUpdates.append(change)
+        }
+        
+        currentChanges = changesWithoutDuplicateUpdates
         onChange?(currentChanges)
```
VojtaStavik, line 323: What if we did it in a bit more simple way? 
```swift
// All destination indices of `move` changes
        let toIndicesOfMoveChanges: [IndexPath] = currentChanges.compactMap {
            if case let .move(_, _, toIndex) = $0 {
                return toIndex
            } else {
                return nil
            }
        }

        // Remove `update` operations with the same index path is existing move `toIndex` changes.
        currentChanges = currentChanges.filter {
            if case let .update(_, index) = $0 {
                // Include only if the reload `index` is not a `move` change destination index.
                return toIndicesOfMoveChanges.contains(index) == false
            } else {
                return true
            }
        }
```

VojtaStavik (approved):
Suggested one more improvement...

PR #775, opened 2021-01-29
Increase fake size for non-sized cells

ManWithBear:
Follow up for #737 

While using zero size for non-sizided cells, we forced layout to iterate over ALL cells (who would expect from zero height cells :D ).
After quick test, this magic size worked well, and I haven't noticed any weird text cell flicks.


VojtaStavik (approved):
Please remove the last commit ;-)

PR #776, opened 2021-01-29
[CIS-256] Lazy model transformation for LLC

ManWithBear:
Improve performance of CoreData observers by introducing `LazyCachedMapCollection`, data structure that applies transformations to underlying sequence on read access.

Decrease amount of DB queries on DTO -> Model transformation when possible.

Huge FPS drop on channel open happens not because of message list layout or message load (affects it, but not critical).
Apparently when we open channel, channel list observer overload CPU by making useless job.
Screenshot 2021-01-29 at 4 20 39

### DIFF
--- a/DemoApp/CreateGroupViewController.swift
+++ b/DemoApp/CreateGroupViewController.swift
@@ -49,7 +49,7 @@ class CreateGroupViewController: UIViewController {
     
     var searchController: ChatUserSearchController!
     
-    var users: [ChatUser] {
+    var users: LazyCachedMapCollection<ChatUser> {
         searchController.users
     }
     
--- a/Sources/StreamChat/Controllers/ChannelController/ChannelController+SwiftUI.swift
+++ b/Sources/StreamChat/Controllers/ChannelController/ChannelController+SwiftUI.swift
@@ -1,5 +1,5 @@
 //
-// Copyright © 2020 Stream.io Inc. All rights reserved.
+// Copyright © 2021 Stream.io Inc. All rights reserved.
 //
 
 import Foundation
@@ -19,7 +19,7 @@ extension _ChatChannelController {
         @Published public private(set) var channel: _ChatChannel<ExtraData>?
         
         /// The messages related to the channel.
-        @Published public private(set) var messages: [_ChatMessage<ExtraData>] = []
+        @Published public private(set) var messages: LazyCachedMapCollection<_ChatMessage<ExtraData>> = []
         
         /// The current state of the Controller.
         @Published public private(set) var state: DataController.State
--- a/Sources/StreamChat/Controllers/ChannelController/ChannelController+SwiftUI_Tests.swift
+++ b/Sources/StreamChat/Controllers/ChannelController/ChannelController+SwiftUI_Tests.swift
@@ -55,7 +55,7 @@ class ChannelController_SwiftUI_Tests: iOS13TestCase {
             )
         }
         
-        AssertAsync.willBeEqual(observableObject.messages, [newMessage])
+        AssertAsync.willBeEqual(Array(observableObject.messages), [newMessage])
     }
     
     func test_observableObject_reactsToDelegateStateChangesCallback() {
@@ -117,8 +117,8 @@ class ChannelControllerMock: ChatChannelController {
     }
     
     var messages_simulated: [_ChatMessage<NoExtraData>]?
-    override var messages: [_ChatMessage<NoExtraData>] {
-        messages_simulated ?? super.messages
+    override var messages: LazyCachedMapCollection<_ChatMessage<NoExtraData>> {
+        messages_simulated.map { $0.lazyCachedMap { $0 } } ?? super.messages
     }
 
     var state_simulated: DataController.State?
--- a/Sources/StreamChat/Controllers/ListDatabaseObserver.swift
+++ b/Sources/StreamChat/Controllers/ListDatabaseObserver.swift
@@ -75,7 +75,7 @@ extension ListChange: Equatable where Item: Equatable {}
 class ListDatabaseObserver<Item, DTO: NSManagedObject> {
     /// The current collection of items matching the provided fetch request. To receive granular updates to this collection,
     /// you can use the `onChange` callback.
-    @Cached var items: [Item]
+    @Cached var items: LazyCachedMapCollection<Item>
     
     /// Called with the aggregated changes after the internal `NSFetchResultsController` calls `controllerDidChangeContent`
     /// on its delegate.
@@ -95,7 +95,7 @@ class ListDatabaseObserver<Item, DTO: NSManagedObject> {
     /// Used for observing the changes in the DB.
     private(set) var frc: NSFetchedResultsController<DTO>!
     
-    let itemCreator: (DTO) -> Item?
+    let itemCreator: (DTO) -> Item
     let request: NSFetchRequest<DTO>
     let context: NSManagedObjectContext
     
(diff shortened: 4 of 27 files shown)


Sources/StreamChat/Controllers/ChannelController/LazyChannelController.swift
```diff
@@ -0,0 +1,1301 @@
...
+            return observer
+        }
+    }
+
+    private func setMessagesObserver() {
+        _messagesObserver.computeValue = { [unowned self] in
+            guard let cid = self.cid else { return nil }
+            let sortAscending = self.listOrdering == .topToBottom ? false : true
+            let request = MessageDTO.messagesFetchRequest(for: cid, sortAscending: sortAscending)
+            request.fetchBatchSize = 20
```
ManWithBear, line 545: This one is super important

Sources/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -441,26 +441,37 @@ private extension _ChatMessage {
         mentionedUsers = Set(dto.mentionedUsers.map { $0.asModel() })
         threadParticipants = Set(dto.threadParticipants.map(\.id))
 
-        latestReplies = MessageDTO
-            .loadReplies(for: dto.id, limit: 25, context: context)
-            .map(_ChatMessage.init)
-        
+        if dto.replies.isEmpty {
```
ManWithBear, line 444: Don;t do DB queries, when here no need. It's super expensive.

Sources/StreamChat/Utils/LazyMapCollection.swift
```diff
@@ -0,0 +1,66 @@
...
+// Copyright © 2021 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+/// Read-only collection that applies transformation to element on first access.
+///
+/// Compared to `LazyMapCollection` do not perform transformation on `count` call.
+public struct LazyCachedMapCollection<Element, SourceElement>: RandomAccessCollection {
+    public typealias Index = Array<SourceElement>.Index
```
VojtaStavik, line 11: Isn't this just a fancy way of saying `public typealias Index = Int`?

Sources/StreamChat/Utils/LazyMapCollection.swift
```diff
@@ -0,0 +1,66 @@
+//
+// Copyright © 2021 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+/// Read-only collection that applies transformation to element on first access.
+///
+/// Compared to `LazyMapCollection` do not perform transformation on `count` call.
+public struct LazyCachedMapCollection<Element, SourceElement>: RandomAccessCollection {
```
VojtaStavik, line 10: This collection needs to be tested.

Sources/StreamChat/Controllers/ChannelController/ChannelController+SwiftUI.swift
```diff
@@ -19,7 +19,7 @@ extension _ChatChannelController {
         @Published public private(set) var channel: _ChatChannel<ExtraData>?
         
         /// The messages related to the channel.
-        @Published public private(set) var messages: [_ChatMessage<ExtraData>] = []
+        @Published public private(set) var messages: AnyRandomAccessCollection<_ChatMessage<ExtraData>> = .init([])
```
VojtaStavik, line 22: We shouldn't expose this via `AnyRandomAccessCollection`, users then can't use `Int` for the index of the collection. Ideally, we'd expose the `LazyCachedMapCollection` as is, but we can't do that because it leaks the DTO type. So we probably need some kind of custom `AnyLazyCachedMapCollection<Element>: Collection` wrapper type for this, where the `Index == Int`.

DemoApp/CreateGroupViewController.swift
```diff
@@ -50,7 +50,7 @@ class CreateGroupViewController: UIViewController {
     var searchController: ChatUserSearchController!
     
     var users: [ChatUser] {
-        searchController.users
+        Array(searchController.users)
     }
```
VojtaStavik, line 54: -> `AnyLazyCachedMapCollection`

Sources/StreamChat/Controllers/ListDatabaseObserver.swift
```diff
@@ -131,7 +131,9 @@ class ListDatabaseObserver<Item, DTO: NSManagedObject> {
             cacheName: nil
         )
         
-        _items.computeValue = { [unowned self] in (self.frc.fetchedObjects ?? []).lazy.compactMap(self.itemCreator) }
+        _items.computeValue = { [unowned self] in
+            LazyCachedMapCollection(source: self.frc.fetchedObjects ?? [], map: self.itemCreator)
+        }
```
VojtaStavik, line 136: All collections in `Foundation` can be constructed using some convenience extension method. I think we should follow this practice so the code looks more natural. So maybe something like `.lazyCacheMap { }` 🤔

Sources/StreamChat/Controllers/ChannelController/ChannelController+SwiftUI.swift
VojtaStavik: This comment is valid for all similar cases in the PR

Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
```diff
@@ -78,15 +78,15 @@ open class _ChatChannelListVC<ExtraData: ExtraDataTypes>: ViewController,
         ) as! _ChatChannelListCollectionViewCell<ExtraData>
     
         cell.uiConfig = uiConfig
-        cell.channelView.channelAndUserId = (controller.channels[indexPath.row], controller.client.currentUserId)
+        cell.channelView.channelAndUserId = (controller.channels[AnyIndex(indexPath.row)], controller.client.currentUserId)
```
VojtaStavik, line 81: I'm surprised this didn't bother you :)

Sources/StreamChat/Controllers/ChannelController/ChannelController+SwiftUI_Tests.swift
```diff
@@ -23,7 +23,7 @@ class ChannelController_SwiftUI_Tests: iOS13TestCase {
         
         XCTAssertEqual(observableObject.state, channelController.state)
         XCTAssertEqual(observableObject.channel, channelController.channel)
-        XCTAssertEqual(observableObject.messages, channelController.messages)
+        XCTAssertEqual(Array(observableObject.messages), Array(channelController.messages))
```
VojtaStavik, line 26: We use our own `XCTAssertEqual` overload. Isn't there a way how to solve this by introducing some specific overload so we don't have to cast it everywhere?

VojtaStavik (changes requested):

Sources/StreamChat/Controllers/ChannelController/ChannelController+SwiftUI_Tests.swift
ManWithBear: Half solved by making collection to conform to `Equatable`. Solves cases when you compare one Lazy to another Lazy.
Cases when you compare it to other datatype (Array) are to rare to bother right now.

Sources/StreamChat/Utils/LazyCachedMapCollection.swift
```diff
@@ -0,0 +1,80 @@
...
+
+    public var endIndex: Index { cache.values.endIndex }
+
+    public var count: Index { cache.values.count }
+
+    public init<Collection: RandomAccessCollection, SourceElement>(
+        source: Collection,
+        map: @escaping (SourceElement) -> Element
+    ) where Collection.Element == SourceElement {
+        let erased = Array(source)
```
ManWithBear, line 31: If we use `AnyRandomAccessCollection` for type erasure, we need to find way how to map `Int` index to `Index` type of source `Collection`. Otherwise usage of `AnyIndex` leads to runtime error.
I let stdlib to solve it.

VojtaStavik: Is this safe to do here? For example, if the source is some kind of other lazy collection, this step means it will be finalized.

VojtaStavik:
@b-onc Since Denis has a day off today, I finished the PR. Can you please review it?

b-onc (approved):
:shipit:

PR #777, opened 2021-01-29
Fix command button in composer

DominikBucher12:
# What this PR do:
- Fixes issue with forgotten and commented out code while merging UIConfig...
# How to test it
- Go to some conversation, open composer, add some slash command and expect the badge to look according to designs.

Before:
Snímek obrazovky 2021-01-29 v 10 42 43

After:
Snímek obrazovky 2021-01-29 v 12 56 51


VojtaStavik (approved):
If you could put a before/after screenshot to the PR description when making a UI change, that would be great 🙏

PR #778, opened 2021-01-29
Replace passing of userSearchController in ChatVC

DominikBucher12:
# What this PR do:
- UserSearchController is now created from `channelController` in `ChatVC`


VojtaStavik (approved):
Najs 👍

PR #779, opened 2021-01-29
[CIS-617] Use resource_bundles for StreamChatUI

### LINKED ISSUE #774: Multiple commands produce Assets.car
## What did you do?

Installed StreamChatUI "3.0-beta.2" through CocoaPods

Be able to build my app without build errors.

## What happened instead?

Received a `Multiple commands produce Assets.car error`
This is due to StreamChatUI using the `resources` property instead of `resource_bundles` in the StreamChatUI.podspec which has issues when building static libraries

```
    spec.resources = ["Sources/StreamChatUI/Resources/**/*"]
```

Can this be changed to use the resource_bundles property. https://guides.cocoapods.org/syntax/podspec.html#resource_bundles

e.g.
```
   spec.resource_bundles = ["Sources/StreamChatUI/*.{xcassets,strings}]"
```

## GetStream Environment
**GetStream Chat version:**3.0
**iOS version:**14.2
**Swift version:**5
**Xcode version:**Xcode 12.3
**Device:**Simulator

## Additional context

gz#8967

b-onc:
Resolves #774

I don't know a way to resolve this issue without the compiler check, suggestions welcome

Tested and works for both direct import (DemoApp) and Cocoapods

### DIFF
--- a/Sources/StreamChatUI/Utils/Bundle+Extensions.swift
+++ b/Sources/StreamChatUI/Utils/Bundle+Extensions.swift
@@ -1,5 +1,5 @@
 //
-// Copyright © 2020 Stream.io Inc. All rights reserved.
+// Copyright © 2021 Stream.io Inc. All rights reserved.
 //
 
 import Foundation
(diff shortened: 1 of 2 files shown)


VojtaStavik (approved):
LGTM 🚢 🇮🇹 I think we will have to have a similar solution for SPM resource bundle.

PR #784, opened 2021-02-01
[CIS-623] Add text truncating in Suggestion cell

DominikBucher12:
# What this PR do:
- Fixes text overlapping the mention image and continuing beyond screen in suggestion cells (see screenshots)
# How to test it
- Run suggestions in any chat screen
# The issue

The problem in this case is that the collectionView cells ignored the `estimatedRowSize` width and rather use the maximum width based on cell content. because we missed the method `preferredLayoutAttributesFitting` where we set the width to be required.

So before introducing this method and changing the text stackView trailing to the image on the right,  it was actually a go for cell autoresizing width. Also without us knowing, the cell was much wider because of the long id text (had no trailing and stack view caused to expand wider).

Basically this issue could have been solved with using explicit `itemSize` with some magic height constraint instead of `estimatedSize` in `CollectionViewFlowLayout` and implementing `preferredLayoutAttributesFitting`.

I was also trying to somehow make work `itemSize` and `estimatedItemSize` together so I would set `itemSize.width` to collectionView width and height to approx 60 but this caused UI loop... So this is probably best available solution I could came up 

Before: 
Snímek obrazovky 2021-02-01 v 13 07 11

After: 
Snímek obrazovky 2021-02-01 v 13 31 32


VojtaStavik (approved):
Thanks 👍

PR #786, opened 2021-02-01
Fix incorrect extra data cheat sheet link

VojtaStavik:
The link we used in the documentation was incorrect.


nuno-vieira (approved):
LGTM

PR #787, opened 2021-02-01
[CIS-629] Fix mint failing run phase if swiftlint is not installed

### LINKED ISSUE #783: DemoApp - Standard Library Issues
## What did you do?
Attempted to install and run the DemoApp

Demo application would run after SPM install.

## What happened instead?
Build failed with error "<unknown>:1:1: unable to load standard library for target 'x86_64-apple-macosx10.15'" and "SwiftLint:1:1: manifest parse error(s):". 

## GetStream Environment
**GetStream Chat version:**
3.0
**iOS version:**
14.3
**Swift version:**
5.3.2
**Xcode version:**
12.4
**Device:**
Simulator (iPhone 12)

## Additional context
If my memory recalls correctly this was a problem when using the Xcode 12 beta's when they were released... It had to do something with excluding certain architectures in the build settings (excluding arm64 for example). However, my memory fails me and I do need some help.

Thanks.

gz#9020

b-onc:
Fixes #783
Cause: https://github.com/yonaskolb/Mint/issues/179

Since `mint run` always installs a package, and installing `swiftlint` takes around 3-4 minutes (on my computer), I've decided it's best that developers don't have to install swiftlint to run the project.

#skip_changelog

### DIFF
--- a/Sources/StreamChatUI/Utils/Bundle+Extensions.swift
+++ b/Sources/StreamChatUI/Utils/Bundle+Extensions.swift
@@ -8,14 +8,14 @@ private class BundleIdentifyingClass {}
 
 extension Bundle {
     static var streamChatUI: Bundle {
-        /// We're using `resource_bundles` to export our resources in the podspec file
-        /// (See https://guides.cocoapods.org/syntax/podspec.html#resource_bundles)
-        /// since we need to support building pod as a static library.
-        /// This attribute causes cocoapods to build a resource bundle and put all our resources inside, during `pod install`
-        /// But this bundle exists only for cocoapods builds, and for other methods (Carthage, git submodule) we directly export
-        /// assets.
-        /// So we need this compiler check to decide which bundle to use.
-        /// See https://github.com/GetStream/stream-chat-swift/issues/774
+        // We're using `resource_bundles` to export our resources in the podspec file
+        // (See https://guides.cocoapods.org/syntax/podspec.html#resource_bundles)
+        // since we need to support building pod as a static library.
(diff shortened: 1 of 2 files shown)


Sources/StreamChatUI/Utils/Bundle+Extensions.swift
```diff
@@ -8,14 +8,14 @@ private class BundleIdentifyingClass {}
...
-        /// So we need this compiler check to decide which bundle to use.
-        /// See https://github.com/GetStream/stream-chat-swift/issues/774
+        // We're using `resource_bundles` to export our resources in the podspec file
+        // (See https://guides.cocoapods.org/syntax/podspec.html#resource_bundles)
+        // since we need to support building pod as a static library.
+        // This attribute causes cocoapods to build a resource bundle and put all our resources inside, during `pod install`
+        // But this bundle exists only for cocoapods builds, and for other methods (Carthage, git submodule) we directly export
+        // assets.
+        // So we need this compiler check to decide which bundle to use.
+        // See https://github.com/GetStream/stream-chat-swift/issues/774
```
VojtaStavik, line 18: I remember SwiftLint complained about this in the previous PR and it was ignored :trollface:

VojtaStavik (approved):
I'm not sure about this solution. The fact that you have `mint` installed doesn't mean we should run all these checks.

Maybe we should introduce some kind of `I_AM_STREAM_DEV` ENV value and run these things only when the value is set? :)

nuno-vieira (approved):
LGTM

PR #789, opened 2021-02-01
[CIS-417] StreamChatUI Tests Environment

nuno-vieira:
# In this PR
Adds Snapshot tests to the StreamChatUI SDK.
Fixes Online Indicator View logic.

- Each Snapshot Variant tests light or dark trait, and different font sizes.
- Currently, all tests run in a fixed size, so the simulator will not influence the tests.
- Maybe later we can add an iPad size
- Test plan has been added and included in the CI Pipeline. Currently, only Debug config has been added.

Note: The current ChatChannelListVC test has UI bugs. This should be fixed in a separate PR.


Sources/StreamChatUI/ChatChannelList/ChatChannelAvatarView.swift
```diff
@@ -43,7 +43,7 @@ open class _ChatChannelAvatarView<ExtraData: ExtraDataTypes>: ChatAvatarView, UI
 
         if channel.isDirectMessageChannel,
             let currentUserId = channelAndUserId.currentUserId,
-            let otherMember = channel.cachedMembers.first(where: { $0.id == currentUserId }),
+            let otherMember = channel.cachedMembers.first(where: { $0.id != currentUserId }),
```
nuno-vieira, line 46: Does this seem it was wrong before? Or Am I missing something?

Sources/StreamChatUI/ChatChannelList/Tests/ChatChannelListVC_SnapshotTests.swift
```diff
@@ -0,0 +1,66 @@
...
+                    lastMessageAt: .init(timeIntervalSince1970: 1_611_951_529_000),
+                    latestMessages: [
+                        ChatMessage.mock(id: "2", text: "Hello", author: .mock(id: "Vader")),
+                        ChatMessage.mock(id: "1", text: "Hello2", author: .mock(id: "Vader2"))
+                    ]
+                )
+            ],
+            changes: []
+        )
+        AssertSnapshot(vc, isEmbededInNavigationController: true)
```
nuno-vieira, line 60: Here, I tried to test every possible logic in this view controller snapshot. With this approach, we reduce the number of screenshots we need to save in comparison if we did the tests per component. So the advantage of this, is that in a single view controller test we test all the components logic and all trait variants. This results already in +/- 8 pictures, which is all the variants we currently support. If we had to do this per component, it would increase a lot the number of pictures stored.

Tests/StreamChatUITests/AssertSnapshot.swift
```diff
@@ -0,0 +1,71 @@
...
+            matching: viewController,
+            as: .image(traits: variant.trait),
+            named: variant.rawValue,
+            record: record,
+            file: file,
+            testName: function,
+            line: line
+        )
+    }
+}
```
nuno-vieira, line 40: If we can, we should try to always snapshot view controllers, It is a more realistic test, and we can group a lot of tests at once. Also, it really only makes sense to test all dynamic font trait variants in a view controller, since it represents real screen sizes.

Tests/StreamChatUITests/AssertSnapshot.swift
```diff
@@ -0,0 +1,71 @@
...
+            matching: view,
+            as: .image(traits: variant.trait),
+            named: variant.rawValue,
+            record: record,
+            file: file,
+            testName: function,
+            line: line
+        )
+    }
+}
```
nuno-vieira, line 71: If there's a view component that has too many logic, then, we could test it in isolation. For example, a message cell type, probably it will not be a good idea to create the MessageListVC with all types of messages, because they will not even fit in the screen.

StreamChat.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -19,6 +19,15 @@
...
       },
+      {
+        "package": "SnapshotTesting",
+        "repositoryURL": "https://github.com/pointfreeco/swift-snapshot-testing.git",
+        "state": {
+          "branch": null,
+          "revision": "c466812aa2e22898f27557e2e780d3aad7a27203",
+          "version": "1.8.2"
+        }
+      },
```
VojtaStavik, line 30: It's not clear from the diff but did you add the target just to the test target?

StreamChat.xcodeproj/project.pbxproj
```diff
@@ -157,7 +156,6 @@
 		794927ED249E0BE2009D7EB7 /* ExtraData.swift in Sources */ = {isa = PBXBuildFile; fileRef = 794927EC249E0BE2009D7EB7 /* ExtraData.swift */; };
 		794927F3249E3F77009D7EB7 /* NotificationAddedToChannel.json in Resources */ = {isa = PBXBuildFile; fileRef = 794927EE249E3D37009D7EB7 /* NotificationAddedToChannel.json */; };
 		794E20F52577DF4D00790DAB /* NameGroupViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 794E20F42577DF4D00790DAB /* NameGroupViewController.swift */; };
-		795296C12582494000435B2E /* UIConfigProvider_Tests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 795296C02582494000435B2E /* UIConfigProvider_Tests.swift */; };
```
VojtaStavik, line 160: Why did you remove this?

Sources/StreamChatUI/ChatChannelList/Tests/ChatChannelListVC_SnapshotTests.swift
```diff
@@ -0,0 +1,66 @@
...
+
+class ChatChannelListVC_SnapshotTests: XCTestCase {
+    var view: ChatChannelListItemView!
+    var vc: ChatChannelListVC!
+    var mockedChannelListController: ChatChannelListController_Mock<NoExtraData>!
+    
+    override func setUp() {
+        super.setUp()
+        UIConfig.default.channelList.channelListItemSubviews.avatarView = MockAvatarView.self
+        UIConfig.default.currentUser.currentUserViewAvatarView = MockCurrentChatUserAvatarView.self
```
VojtaStavik, line 17: In tests we should rather use `UIConfig` provided explicitly. So rather something like this:
```swift
var uiConfig = UIConfig()
uiConfgi.channelList.channelListItemSubviews.avatarView = MockAvatarView.self
uiConfig.currentUser.currentUserViewAvatarView = MockCurrentChatUserAvatarView.self

vc = ChatChannelListVC()
vc.uiConfig = uiConfig
```

Sources/StreamChatUI/ChatChannelList/Tests/ChatChannelListVC_SnapshotTests.swift
VojtaStavik: Great 👍

Tests/StreamChatUITests/AssertSnapshot.swift
```diff
@@ -0,0 +1,71 @@
...
+///
+/// - Parameters:
+///   - view: The component to be tested.
+///   - variants: The variants that a snapshot will be taken.
+///   Only the light and dark variants should be tested in small components.
+///   - record: True if a new reference should be saved. False by default,
+///   so that the newly captured snapshot is compared with the current reference.
+func AssertSnapshot(
+    _ view: UIView,
+    variants: [SnapshotVariation] = [.light, .dark],
```
VojtaStavik, line 54: why just these two variants?

StreamChat.xcodeproj/project.pbxproj
```diff
@@ -4136,7 +4136,7 @@
 				CODE_SIGN_STYLE = Automatic;
 				DEVELOPMENT_TEAM = EHV7XZLAHA;
 				INFOPLIST_FILE = Tests/Info.plist;
-				IPHONEOS_DEPLOYMENT_TARGET = 11.0;
+				IPHONEOS_DEPLOYMENT_TARGET = 12.1;
```
VojtaStavik, line 4183: > This simplifies a lot the code in SnapshotVariant since there’s no need for optionals and Macros.

What kind of optionals and macros? Can you be more specific please?

Tests/StreamChatUITests/MockViews/MockAvatarView.swift
```diff
@@ -0,0 +1,13 @@
...
+
+@testable import StreamChatUI
+import XCTest
+
+class MockAvatarView: ChatChannelAvatarView {
+    override func updateContent() {
+        super.updateContent()
+        imageView.image = UIImage(named: "pattern1", in: .streamChatUI)
+    }
+}
```
VojtaStavik, line 13: This is a workaround for a random image in `ChatChannelAvatarView` when the avatar URL is missing, right? Why don't you just assign it the model object?

Tests/StreamChatUITests/MockViews/MockCurrentChatUserAvatarView.swift
```diff
@@ -0,0 +1,13 @@
...
+
+@testable import StreamChatUI
+import XCTest
+
+class MockCurrentChatUserAvatarView: CurrentChatUserAvatarView {
+    override func updateContent() {
+        super.updateContent()
+        avatarView.imageView.image = UIImage(named: "pattern2", in: .streamChatUI)
+    }
+}
```
VojtaStavik, line 13: same here

Tests/StreamChatUITests/SnapshotVariation.swift
```diff
@@ -0,0 +1,28 @@
...
+        case .small: return UITraitCollection(preferredContentSizeCategory: .small)
+        case .medium: return UITraitCollection(preferredContentSizeCategory: .medium)
+        case .large: return UITraitCollection(preferredContentSizeCategory: .large)
+        case .extraLarge: return UITraitCollection(preferredContentSizeCategory: .extraLarge)
+        case .extraExtraExtraLarge: return UITraitCollection(preferredContentSizeCategory: .extraExtraExtraLarge)
+        case .light: return UITraitCollection(userInterfaceStyle: .light)
+        case .dark: return UITraitCollection(userInterfaceStyle: .dark)
+        }
+    }
+}
```
VojtaStavik, line 28: This is great 👍 Let's make it even better:
- we can combine all those with a `displayScale = 1` trait and make the screenshots smaller -> faster comparison
- let's try to keep the number of screenshots to a minimum and combine those traits together. Maybe we just need [small + dark], [large + light],[xxl + dark]?
- any other interesting trait collecitons we should test for?

StreamChat.xcodeproj/project.pbxproj
```diff
@@ -3454,7 +3476,29 @@
...
+				ADCAA33C25C48DDA003BE403 /* TestRunnerEnvironment.swift in Sources */,
+				AD9F7B7525C820E700DF8CD2 /* MockAvatarView.swift in Sources */,
+				ADCAA31E25C473C0003BE403 /* ChatChannel_Mock.swift in Sources */,
+				ADCAA33E25C48DDA003BE403 /* iOS13TestCase.swift in Sources */,
+				ADCAA32225C473C0003BE403 /* ChatUser_Mock.swift in Sources */,
+				ADCAA32025C473C0003BE403 /* ChatChannelMember_Mock.swift in Sources */,
+				ADCAA33D25C48DDA003BE403 /* Await.swift in Sources */,
+				ADCAA33125C48DDA003BE403 /* TestDispatchQueue.swift in Sources */,
+				ADCAA32325C473C0003BE403 /* ChatClient_Mock.swift in Sources */,
+				ADCAA33A25C48DDA003BE403 /* StressTestCase.swift in Sources */,
```
VojtaStavik, line 3501: You just need to import `StreamChatTestTools`

Sources/StreamChat/Models/Channel.swift
```diff
@@ -215,4 +215,9 @@ public struct ChannelUnreadCount: Decodable, Equatable {
     
     /// The number of unread messages that mention the current user.
     public internal(set) var mentionedMessages: Int
+    
+    public init(messages: Int, mentionedMessages: Int) {
+        self.messages = messages
+        self.mentionedMessages = mentionedMessages
+    }
```
VojtaStavik, line 222: Nope, there's a reason why this is not public. The way to expose this is via `StreamChatTestTools` - see other mocks.

Sources/StreamChatTestTools/ChatClient_Mock.swift
```diff
@@ -11,23 +11,7 @@ public extension _ChatClient {
...
-                        sessionConfiguration: $1,
-                        requestEncoder: $2,
-                        eventDecoder: $3,
-                        eventNotificationCenter: $4,
-                        internetConnection: $5
-                    )
-                }
-            )
-        )
+        return .init(config: config, tokenProvider: .development(userId: .anonymous))
```
VojtaStavik, line 14: You can't just change a piece of code without understanding why was it here ❗ What about the `environment` field and `workerBuilders`?

VojtaStavik (commented):
Good start 👍 Please be more careful when changing existing code.

Tests/StreamChatUITests/AssertSnapshot.swift
```diff
@@ -0,0 +1,71 @@
...
+    record: Bool = false,
+    line: UInt = #line,
+    file: StaticString = #file,
+    function: String = #function
+) {
+    let viewController = isEmbededInNavigationController ? UINavigationController(rootViewController: vc) : vc
+    variants.forEach { variant in
+        assertSnapshot(
+            matching: viewController,
+            as: .image(traits: variant.trait),
```
VojtaStavik, line 32: What's the default size for this? Shouldn't we specify the size explicitly?

StreamChat.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
nuno-vieira: Yes, I've added the snapshot framework only to the test target. But I was expecting this to show in the diff? 🤔 Maybe in package.swift? But I can see that we are not specifying there the dependencies of the test targets, so maybe it is the project file, but it is not clear in the diff?

StreamChat.xcodeproj/project.pbxproj
nuno-vieira: This will be reverted, it was a temporary easy fix to avoid compilation issues

StreamChat.xcodeproj/project.pbxproj
nuno-vieira: image
The light and dark trait collections are only supported by iOS 12+. If we have the deployment target to iOS 11.0, we need the `trait: UITraitCollection` property to be optional, because we would need a `#if` macro condition for iOS <12 and return `nil` in this case.

Tests/StreamChatUITests/MockViews/MockAvatarView.swift
nuno-vieira: The model object expects an URL, this way, we won't need a GET Request, and the snapshot will be faster. But maybe we want to test that?

VojtaStavik: It can be a local file URL

StreamChat.xcodeproj/project.pbxproj
VojtaStavik: Well isn't it just a problem with the fact that we chose `enum` to represent this? It's easy to skip the `userInterfaceStyle` trait collection when running on iOS11. Maybe it should rather be something like:
```swift
extension UITraitCollection { 
  static let case1: UITraitCollection = ... // small + dark (if applicable)
  ...
} 
```

Tests/StreamChatUITests/AssertSnapshot.swift
nuno-vieira: My initial idea was that for small components, we would not care about the font sizes, because that should be tested in view controllers. But thinking about it better now, that is probably a bad assumption 🤔 In the message cells, we might need to test the font sizes without them being in a view controller. So probably is better to switch back these to all cases. I guess we could also test each message cell, in a view controller with one cell, not sure... I think we need to explore and see what works best for us.

VojtaStavik: Yeah, we'll see once we start using it for real.

Tests/StreamChatUITests/SnapshotVariation.swift
nuno-vieira: 1. Yes great idea 👍 That will also reduce the size of the screenshots 🚀 
2. Yeah we can for sure join some of them, to reduce the number of screenshots 👍 
3. For now I think we have the most important ones, and later when we have better accessibility features, we can use this [plugin](https://github.com/cashapp/AccessibilitySnapshot) of the snapshot framework to test more accessibility features. But maybe there's more that we can add right now? Thoughts?  @GetStream/ios-developers

Tests/StreamChatUITests/AssertSnapshot.swift
nuno-vieira: For view controllers, the size will be the one from the device that we are currently running on. Which means, that the CI and us when running locally, should use the same simulator device. We can improve this by running assertions on `setUp` to make sure that we are running in the correct device. We can create a `SnapshotTestCase: XCTestCase` that will make sure that when we run locally, that we are running in the correct device, because if do not, the tests will fail. 

Anyways, I will try to see if it is possible to change the width and height of the view controller to a custom one independent of the running device. But this adds another question, what if we have designed specific per iPad/iPhone ? Then, in this case, the device matters, and not the size of view controller.

Sources/StreamChatTestTools/ChatClient_Mock.swift
nuno-vieira: Yes totally missed this, this was also a quick fix, and fix it later, but forgot about this one, will include everything 👍

Tests/StreamChatUITests/AssertSnapshot.swift
VojtaStavik: AFAIK you can provide an explicit size, you can even use something like `.image(traits: variant.trait, device: .iPhoneX)`. We need something like this. We can't rely on everybody using the same simulator for the snapshots.

Tests/StreamChatUITests/SnapshotVariation.swift
VojtaStavik: OK, let's start simple for now, we can always add more things as we need them.

ManWithBear:
> In order to snapshot view controllers, the device where the test is run should always be the same. Maybe there's away to define a custom with and height for the view controller? 🤔 (Investigating)
> Should we add iPhone and iPad Tests? Or portrait vs landscape?
@DominikBucher12 ^

Sources/StreamChatUI/ChatChannelList/Tests/ChatChannelListVC_SnapshotTests.swift
ManWithBear: On other hand you lose "single reason to fail" test property.
So if something goes wrong, you need to find yourself from diff which exactly state is bugged and etc.

Considering size of SDK, I'm sure amount of images will growth anyway. And for such cases Git-LFS exists 
https://git-lfs.github.com

Tests/StreamChatUITests/SnapshotVariation.swift
```diff
@@ -0,0 +1,28 @@
+//
+// Copyright © 2021 Stream.io Inc. All rights reserved.
+//
+
+@testable import StreamChatUI
+import XCTest
+
+enum SnapshotVariation: String, Hashable, CaseIterable {
```
ManWithBear, line 8: IMHO it's weird to joint Font Size and User Interface Style.
I would expect SnapshotVariation be at least pair:
```
static let default = (.large, .light)
static let bigAndDark = (.extraExtraExtraLarge, .dark)
...
```

Sources/StreamChatUI/ChatChannelList/Tests/ChatChannelListVC_SnapshotTests.swift
VojtaStavik: @ManWithBear I hear you but isn't the failure ambiguity an inherent characteristic of snapshot tests? I think about them more like integration tests than unit tests. I guess we could try to reduce this. For example in `ChannelListVC` by introducing some kind of "dummy" view instead of real channel items. But that would also mean we'd have to add more actual unit tests to test that the cells are loaded with correct data. However, we'd probably still need to have a snapshot of the "real" cell just to test that the sizing and layout work correctly as a whole.

So yeah, maybe this can be reduced by testing the different cell states separately but we will still need some "master" snapshot that will just verify everything works well together, too. IMO the approach proposed by @nuno-vieira is a good-enough start. We can always introduce more granularity if we see the lack of it causes issues.

Sources/StreamChatUI/ChatChannelList/Tests/ChatChannelListVC_SnapshotTests.swift
```diff
@@ -0,0 +1,66 @@
...
+        vc = ChatChannelListVC()
+        mockedChannelListController = ChatChannelListController_Mock.mock()
+        vc.controller = mockedChannelListController
+    }
+    
+    func test_populated() {
```
DominikBucher12, line 23: This just a nit since this is WIP, just not to forget...Probably it would be better to create some convention Or use something like `test_chatChannelList_isPopulated` as you can see with the rest of the tests... Or somehow start `test_snapshotTest_...`

Sources/StreamChatUI/ChatChannelList/Tests/ChatChannelListVC_SnapshotTests.swift
nuno-vieira: @ManWithBear I think that snapshot testing each component is fine for some cases (Especially for the messages cells), but for the most part, it is more realistic to test the views in a view controller, and we will find problems that we would not with just testing a view in isolation. 

Yes I'm aware of git LFS and we probably will need to use it when this scales a lot, but for now, is better if we can minimize the images we save. I also agree with @VojtaStavik that I see more advantages in treating snapshot tests as integration tests than unit tests.

It should be easy to see what went wrong, if the test fails, we can see the Diff image, so that should not be a problem:
image

Sources/StreamChatUI/ChatChannelList/Tests/ChatChannelListVC_SnapshotTests.swift
nuno-vieira: The snapshots are saved in folders with the name of the view controller already. From what I've seen, this is the pattern used in Snapshot tests, but I also prefer putting the name of what we are testing in the function name, even if in the folder/picture path it will be a bit repetitive, so I will use `test_chatChannelList_isPopulated`, thanks for the suggestion 👌

Sources/StreamChatUI/ChatChannelList/Tests/ChatChannelListVC_SnapshotTests.swift
nuno-vieira: I tried doing this before, but it is not working properly, the config is not being overridden in that way, maybe we can investigate that in another ticket?

StreamChat.xcodeproj/project.pbxproj
nuno-vieira: I can convert this to an option set, but the advantage I see in this case to be an enum, is that we make we always snapshot for every variant with `allCases`. If in the future we introduce a new Variant, we will always test all variants automatically. Where if we use an OptionSet, we will need to add the new option to an `all` property manually, and maybe someone can forget to add it. I bumped the version because I thought it wouldn't be that important since I don't think we will run the snapshots on iOS 11 anyways. But let me know if you think it is better for this to be an OptionSet, and put the iOS 11 support.

Sources/StreamChatUI/ChatChannelList/Tests/ChatChannelListVC_SnapshotTests.swift
nuno-vieira: I will not use this anyways btw, since a local URL file should fix this, without the need to customize the views.

nuno-vieira: Nuno from the future: The local file did not work, I can see the local URL image exists, but setImage does not render the image properly 🤔 Maybe it tries to use URLSession to download the image under the hood (Nuke)? 

If we keep the MockViews, then can we address this later?
```
vc = ChatChannelListVC()
vc.uiConfig = uiConfig
```
Does not work and does not override the views.

Tests/StreamChatUITests/MockViews/MockAvatarView.swift
nuno-vieira: https://github.com/GetStream/stream-chat-swift/pull/789/commits/f93eaa1d8156efd74bf01b80cd7aaf237e8e2098#r568880558

Tests/StreamChatUITests/MockViews/MockCurrentChatUserAvatarView.swift
nuno-vieira: https://github.com/GetStream/stream-chat-swift/pull/789/commits/f93eaa1d8156efd74bf01b80cd7aaf237e8e2098#r568880558

StreamChat.xcodeproj/project.pbxproj
VojtaStavik: We support iOS and we will eventually run snapshot tests for iOS11, too. The `allCases` benefits are minimal so I'd suggest moving from `enum` into something more flexible.

Sources/StreamChatUI/ChatChannelList/Tests/ChatChannelListVC_SnapshotTests.swift
VojtaStavik: The local URL should work. The system doesn't care if the URL is local or remote. -> Slack

Sources/StreamChatTestTools/ChatClient_Mock.swift
```diff
@@ -13,17 +13,18 @@ public extension _ChatClient {
         
         return .init(
             config: config,
+            tokenProvider: .development(userId: .anonymous),
```
VojtaStavik, line 16: You can just do `.anonymous`

Sources/StreamChatTestTools/Models/ChatChannelMember_Mock.swift
```diff
@@ -9,22 +9,26 @@ public extension _ChatChannelMember {
...
-        userUpdatedAt: Date,
-        lastActiveAt: Date?,
-        extraData: ExtraData,
-        memberRole: MemberRole,
-        memberCreatedAt: Date,
-        memberUpdatedAt: Date,
-        isInvited: Bool,
-        inviteAcceptedAt: Date?,
-        inviteRejectedAt: Date?
+        name: String?,
```
VojtaStavik, line 12: Why doesn't `name` have the default parameter?

Tests/StreamChatTests/StreamChatTestPlan.xctestplan
```diff
@@ -1,32 +1,37 @@
...
-      "id" : "5358EB81-7D4D-42F7-B281-016C02706988",
-      "name" : "Configuration 1",
-      "options" : {
-
+      "id": "5358EB81-7D4D-42F7-B281-016C02706988",
+      "name": "Configuration 1",
+      "options": {
       }
     }
   ],
```
VojtaStavik, line 9: What is the purpose of changing this file?

Tests/StreamChatUITests/AssertSnapshot.swift
```diff
@@ -0,0 +1,71 @@
...
+    record: Bool = false,
+    line: UInt = #line,
+    file: StaticString = #file,
+    function: String = #function
+) {
+    let viewController = isEmbededInNavigationController ? UINavigationController(rootViewController: vc) : vc
+    variants.forEach { variant in
+        assertSnapshot(
+            matching: viewController,
+            as: .image(size: CGSize(width: 360, height: 700), traits: variant.traits),
```
VojtaStavik, line 32: this size should be in a constant somewhere. See similar constants like `defaultTimeout`...

Tests/StreamChatUITests/SnapshotVariation.swift
```diff
@@ -0,0 +1,52 @@
...
+            return UITraitCollection(preferredContentSizeCategory: .small)
+        case .medium:
+            return UITraitCollection(preferredContentSizeCategory: .medium)
+        case .large:
+            return UITraitCollection(preferredContentSizeCategory: .large)
+        case .extraLarge:
+            return UITraitCollection(preferredContentSizeCategory: .extraLarge)
+        case .extraExtraExtraLarge:
+            return UITraitCollection(preferredContentSizeCategory: .extraExtraExtraLarge)
+        }
```
VojtaStavik, line 40: I don't think we need that many sizes. Just `small`, `large`, and `extraExtraExtraLarge` is OK imo.

Tests/StreamChatUITests/StreamChatUITestPlan.xctestplan
```diff
@@ -0,0 +1,37 @@
...
+          "name": "StreamChatUI"
+        }
+      ]
+    },
+    "environmentVariableEntries": [
+      {
+        "key": "CI",
+        "value": "TRUE"
+      }
+    ]
```
VojtaStavik, line 25: This value should be set by Fastlane, it shouldn't be committed. When you run it locally, you don't want this value to be set.

VojtaStavik (changes requested):
All right, looking good 👍 Just a couple of final changes...

Tests/StreamChatTests/StreamChatTestPlan.xctestplan
nuno-vieira: I can see it is just format changes, probably it was done by the pre-commit hook? And because I messed with the testplan files 🤔 I can revert this if necessary

Sources/StreamChatTestTools/Models/ChatChannelMember_Mock.swift
nuno-vieira: Forgot to add it 👍 Will fix it

nuno-vieira:
> # [Codecov](https://codecov.io/gh/GetStream/stream-chat-swift/pull/789?src=pr&el=h1) Report
> > Merging [#789](https://codecov.io/gh/GetStream/stream-chat-swift/pull/789?src=pr&el=desc) ([1fc25a7](https://github.com/GetStream/stream-chat-swift/commit/1fc25a7c47ec023975372a5f7ca1f3b69549a1e8)) into [main](https://codecov.io/gh/GetStream/stream-chat-swift/commit/9a2a96f0f15c901753d9791a49550f671d46072a?el=desc) ([9a2a96f](https://github.com/GetStream/stream-chat-swift/commit/9a2a96f0f15c901753d9791a49550f671d46072a)) will **decrease** coverage by `75.14%`.
> > The diff coverage is `81.66%`.
> 
> [Impacted file tree graph](https://codecov.io/gh/GetStream/stream-chat-swift/pull/789?src=pr&el=tree)
> 
> ```diff
> @@             Coverage Diff             @@
> ##             main     #789       +/-   ##
> ===========================================
> - Coverage   89.16%   14.02%   -75.15%     
> ===========================================
>   Files         219      296       +77     
>   Lines        8576    12545     +3969     
> ===========================================
> - Hits         7647     1759     -5888     
> - Misses        929    10786     +9857     
> ```
> 
> [Impacted Files](https://codecov.io/gh/GetStream/stream-chat-swift/pull/789?src=pr&el=tree)	Coverage Δ	
> [Tests/StreamChatUITests/AssertSnapshot.swift](https://codecov.io/gh/GetStream/stream-chat-swift/pull/789/diff?src=pr&el=tree#diff-VGVzdHMvU3RyZWFtQ2hhdFVJVGVzdHMvQXNzZXJ0U25hcHNob3Quc3dpZnQ=)	`52.17% <52.17%> (ø)`	
> [...ChatUI/ChatChannelList/ChatChannelAvatarView.swift](https://codecov.io/gh/GetStream/stream-chat-swift/pull/789/diff?src=pr&el=tree#diff-U291cmNlcy9TdHJlYW1DaGF0VUkvQ2hhdENoYW5uZWxMaXN0L0NoYXRDaGFubmVsQXZhdGFyVmlldy5zd2lmdA==)	`86.95% <100.00%> (ø)`	
> [Tests/StreamChatUITests/SnapshotVariant.swift](https://codecov.io/gh/GetStream/stream-chat-swift/pull/789/diff?src=pr&el=tree#diff-VGVzdHMvU3RyZWFtQ2hhdFVJVGVzdHMvU25hcHNob3RWYXJpYW50LnN3aWZ0)	`100.00% <100.00%> (ø)`	
> [...ChatUITests/TestImages/XCTestCase+TestImages.swift](https://codecov.io/gh/GetStream/stream-chat-swift/pull/789/diff?src=pr&el=tree#diff-VGVzdHMvU3RyZWFtQ2hhdFVJVGVzdHMvVGVzdEltYWdlcy9YQ1Rlc3RDYXNlK1Rlc3RJbWFnZXMuc3dpZnQ=)	`100.00% <100.00%> (ø)`	
> [Sources/StreamChat/Models/Device.swift](https://codecov.io/gh/GetStream/stream-chat-swift/pull/789/diff?src=pr&el=tree#diff-U291cmNlcy9TdHJlYW1DaGF0L01vZGVscy9EZXZpY2Uuc3dpZnQ=)	`0.00% <0.00%> (-100.00%)`	⬇️
> [Sources/StreamChat/Utils/UniqueId.swift](https://codecov.io/gh/GetStream/stream-chat-swift/pull/789/diff?src=pr&el=tree#diff-U291cmNlcy9TdHJlYW1DaGF0L1V0aWxzL1VuaXF1ZUlkLnN3aWZ0)	`0.00% <0.00%> (-100.00%)`	⬇️
> [Sources/StreamChat/Database/DataStore.swift](https://codecov.io/gh/GetStream/stream-chat-swift/pull/789/diff?src=pr&el=tree#diff-U291cmNlcy9TdHJlYW1DaGF0L0RhdGFiYXNlL0RhdGFTdG9yZS5zd2lmdA==)	`0.00% <0.00%> (-100.00%)`	⬇️
> [Sources/StreamChat/Models/ChannelRead.swift](https://codecov.io/gh/GetStream/stream-chat-swift/pull/789/diff?src=pr&el=tree#diff-U291cmNlcy9TdHJlYW1DaGF0L01vZGVscy9DaGFubmVsUmVhZC5zd2lmdA==)	`0.00% <0.00%> (-100.00%)`	⬇️
> [Sources/StreamChat/Query/ChannelQuery.swift](https://codecov.io/gh/GetStream/stream-chat-swift/pull/789/diff?src=pr&el=tree#diff-U291cmNlcy9TdHJlYW1DaGF0L1F1ZXJ5L0NoYW5uZWxRdWVyeS5zd2lmdA==)	`0.00% <0.00%> (-100.00%)`	⬇️
> [Sources/StreamChat/Models/AttachmentId.swift](https://codecov.io/gh/GetStream/stream-chat-swift/pull/789/diff?src=pr&el=tree#diff-U291cmNlcy9TdHJlYW1DaGF0L01vZGVscy9BdHRhY2htZW50SWQuc3dpZnQ=)	`0.00% <0.00%> (-100.00%)`	⬇️
> ... and [311 more](https://codecov.io/gh/GetStream/stream-chat-swift/pull/789/diff?src=pr&el=tree-more)		
> [Continue to review full report at Codecov](https://codecov.io/gh/GetStream/stream-chat-swift/pull/789?src=pr&el=continue).
> 
> > **Legend** - [Click here to learn more](https://docs.codecov.io/docs/codecov-delta)
> > `Δ = absolute <relative> (impact)`, `ø = not affected`, `? = missing data`
> > Powered by [Codecov](https://codecov.io/gh/GetStream/stream-chat-swift/pull/789?src=pr&el=footer). Last update [9a2a96f...1fc25a7](https://codecov.io/gh/GetStream/stream-chat-swift/pull/789?src=pr&el=lastupdated). Read the [comment docs](https://docs.codecov.io/docs/pull-request-comments).

This is weird, I set the code coverage of StreamChatUITests to only StreamChatUI, but it looks it is counting StreamChat too ? 🤔

Sources/StreamChatTestTools/Controllers/ChatChannelController_Mock.swift
```diff
@@ -16,8 +16,8 @@ public class ChatChannelController_Mock<ExtraData: ExtraDataTypes>: _ChatChannel
         channel_mock ?? super.channel
     }
     
-    public private(set) var messages_mock: [_ChatMessage<ExtraData>]?
-    override public var messages: [_ChatMessage<ExtraData>] {
+    public private(set) var messages_mock: LazyCachedMapCollection<_ChatMessage<ExtraData>>?
+    override public var messages: LazyCachedMapCollection<_ChatMessage<ExtraData>> {
         messages_mock ?? super.messages
     }
```
VojtaStavik, line 22: This is not the ideal solution really. `messages_mock` should just be `[_ChatMessage<ExtraData>]` and then you can do:
```swift
message_mock.map { $0.lazyCachedMap { $0 } } ?? super.messages
```
Look at other mocks how it's done there.

Sources/StreamChatTestTools/Controllers/ChatChannelController_Mock.swift
```diff
@@ -30,7 +30,11 @@ public class ChatChannelController_Mock<ExtraData: ExtraDataTypes>: _ChatChannel
 
 public extension ChatChannelController_Mock {
     /// Simulates the initial conditions. Setting these values doesn't trigger any observer callback.
-    func simulateInitial(channel: _ChatChannel<ExtraData>, messages: [_ChatMessage<ExtraData>], state: DataController.State) {
+    func simulateInitial(
+        channel: _ChatChannel<ExtraData>,
+        messages: LazyCachedMapCollection<_ChatMessage<ExtraData>>,
```
VojtaStavik, line 35: -> yeah, let's please use just `Array`

VojtaStavik: same comment of all similar places

Tests/StreamChatTests/StreamChatTestPlan.xctestplan
```diff
@@ -1,32 +1,31 @@
 {
-  "configurations" : [
+  "configurations": [
```
VojtaStavik, line 2: I'd recommend removing all changes from `StreamChatTestPlan.xctestplan` completely.

fastlane/Fastfile
```diff
@@ -121,6 +121,21 @@ lane :test do |options|
     )
 end
 
+desc "Runs ui tests in Debug config"
+lane :test_ui do |options|
+
+  setCIEnvironmentVariable("../Tests/StreamChatUITests/StreamChatUITestPlan.xctestplan")
```
VojtaStavik, line 127: You suggested that this should be run only if Fastlane's `is_ci` is `true` and I think it's a great idea. We can maybe add the check inside `setCIEnvironmentVariable`?

VojtaStavik (approved):
Once you figure out how to make it work on CI we can merge this 👍

Tests/StreamChatUITests/AssertSnapshot.swift
```diff
@@ -0,0 +1,76 @@
...
```
VojtaStavik, line 37: It's great that you created the parameter but it's usually also good to use it :trollface: :D

VojtaStavik (approved):
Just one small fix and then :shipit: 🎉 Thank you!

b-onc (changes requested):
(As discussed in Slack) We don't need `IS_CI` as env value and ui tests should have their own job

b-onc (approved):
Happy to see codecov fixed! :shipit:

PR #791, opened 2021-02-02
[CIS-631] Disable changelog warning for StreamChatUI

DominikBucher12:
Disable changelog warning for StreamChatUI in DangerFile, hopefully this works on first try 😅

### DIFF
--- a/Dangerfile
+++ b/Dangerfile
@@ -23,7 +23,8 @@ if github.pr_body.length < 3 && git.lines_of_code > 50
 end
 
 ## Let's check if there are any changes in the project folder
-has_app_changes = !git.modified_files.grep(/Sources/).empty?
+has_app_changes = !git.modified_files.grep('/Sources/StreamChat/').empty?
+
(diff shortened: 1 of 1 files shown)


Dangerfile
```diff
@@ -24,6 +24,8 @@ end
 
 ## Let's check if there are any changes in the project folder
 has_app_changes = !git.modified_files.grep(/Sources/).empty?
```
b-onc, line 26: Just change this location to `/Sources/StreamChat/` and it'll work, no need to add additional flag

When we're ready (for StreamChatUI) we'll change it back to this

To avoid ruby error, put ' (make it `'/Sources/StreamChat/'`)

DominikBucher12: Great, thank you, will fix this in next commit (uploaded wrong again)

b-onc (approved):
:shipit: on green

VojtaStavik (approved):

PR #794, opened 2021-02-02
[CIS-426][CIS-615] Add StreamChat and StreamChatUI to SPM

DominikBucher12:
# What this PR do:
- Support resources for SPM, because Swift 5.3 allows us to do so.
# How to test it
- Run the Demo project (`Integration/SPM_Tests/SwiftPackageManager.xcodeproj`) try to load some resources via UIConfig (for `StreamChatUI` and check if the app doesn't crash after importing  `StreamChat` because it could happen the database model couldn't be find..)

Note: Currently checking out if I can solve the warnings on source files, probably shouldn't be so difficult.


Package.swift
```diff
@@ -2,50 +2,356 @@
 // The swift-tools-version declares the minimum version of Swift required to build this package.
 
 import PackageDescription
+import Foundation
+
+let streamChatFiles: [String] = [
```
ManWithBear, line 7: Are they magically generated or we would need to do it by hands?

b-onc: @DominikBucher12 We need some kind of logic here
pseudocode
```swift
let streamChatSourceFiles = FileManager.files(under: "/Sources/StreamChat").filter { $0.name.suffix == ".swift" && $0.name.suffix != "_Tests.swift" } // or Mock
// same for streamChatUI
```

VojtaStavik: Indeed, this will be generated eventually. The first goal is to make the SPM integration work 😅 Then we can talk about how to improve it.

The snippet you suggest @b-onc is of course our first try. `Package.swift` is not evaluated in the context of the root folder. You can't iterate over the files from the directly. This will have to be generated by some kind of GH action, until SPM has support for `*_Tests.swift` etc.

Sources/StreamChatUI/Utils/Bundle+Extensions.swift
```diff
@@ -20,6 +20,28 @@ extension Bundle {
...
+        // But this bundle exists only for cocoapods builds, and for other methods (Carthage, git submodule) we directly export
+        // assets.
+        // So we need this compiler check to decide which bundle to use.
+        // See https://github.com/GetStream/stream-chat-swift/issues/774
+        #if COCOAPODS
+        return Bundle(for: BundleIdentifyingClass.self)
+            .url(forResource: "StreamChat", withExtension: "bundle")
+            .flatMap(Bundle.init(url:))!
+        #elseif SWIFT_PACKAGE
+        return Bundle.module
```
VojtaStavik: This file lives in `StreamChatUI` target, so the bundle for `streamChat` should be here.

Sources/StreamChatUI/Utils/Bundle+Extensions.swift
```diff
@@ -20,6 +20,28 @@ extension Bundle {
...
+        // since we need to support building pod as a static library.
+        // This attribute causes cocoapods to build a resource bundle and put all our resources inside, during `pod install`
+        // But this bundle exists only for cocoapods builds, and for other methods (Carthage, git submodule) we directly export
+        // assets.
+        // So we need this compiler check to decide which bundle to use.
+        // See https://github.com/GetStream/stream-chat-swift/issues/774
+        #if COCOAPODS
+        return Bundle(for: BundleIdentifyingClass.self)
+            .url(forResource: "StreamChat", withExtension: "bundle")
+            .flatMap(Bundle.init(url:))!
```
VojtaStavik: For this to work we must also enable resource bundles in the podspec for `StreamChat`

StreamChat.xcodeproj/project.pbxproj
```diff
@@ -131,7 +131,6 @@
 		792DDA66256FB69F001DB91B /* LaunchScreen.storyboard in Resources */ = {isa = PBXBuildFile; fileRef = 792DDA64256FB69F001DB91B /* LaunchScreen.storyboard */; };
 		792DDAA125711AF2001DB91B /* CreateChatViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 792DDAA025711AF2001DB91B /* CreateChatViewController.swift */; };
 		792DDAA825753BEA001DB91B /* CreateGroupViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 792DDAA725753BEA001DB91B /* CreateGroupViewController.swift */; };
-		792E3D6B25C97D920040B0C2 /* Package.swift in Sources */ = {isa = PBXBuildFile; fileRef = 792E3D6A25C97D920040B0C2 /* Package.swift */; };
```
VojtaStavik: Why? It was on purpose to make editing of `Package.swift` simpler

hooks/regenerage-spm-package-if-needed.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+# Was there a new file added or deleted?
+if [ ! -z "$(git diff --staged | grep  -e 'deleted\|new file mode')" ];
+then
+    # Regenerate list of files in Package.swift
+    ./GenerateSPMFileLists.swift
+
+    # Add changes to the current commit
```
VojtaStavik, line 9: This solution is a bit naive. It can happen that also the changes that weren't staged will be added. We can improve later if it's a problem.

.github/workflows/ci.yml
```diff
@@ -221,3 +221,36 @@ jobs:
...
+        restore-keys: ${{ runner.os }}-gem-
+    - name: Cache Mint
+      uses: actions/cache@v2
+      id: mint-cache
+      with:
+        path: /usr/local/lib/mint
+        key: ${{ runner.os }}-mint-${{ hashFiles('./Mintfile') }}
+        restore-keys: ${{ runner.os }}-mint-
+    - name: Run bootstrap.sh
+      run: ./bootstrap.sh
```
b-onc, line 245: Why do we run bootstrap in CI?

Package.swift
```diff
@@ -13,35 +15,193 @@ let package = Package(
...
             dependencies: ["Starscream"],
-            path: "Sources/Client")
+            exclude: ["README.md", "Info.plist"] + streamChatSourcesExcluded,
+            resources: [.copy("Database/StreamChatModel.xcdatamodeld")]
+        ),
+        .target(
+            name: "StreamChatUI",
+            dependencies: ["StreamChat", "Nuke", "SwiftyGif"],
+            exclude: ["README.md", "Info.plist"] + streamChatUISourcesExcluded,
+            resources: [.process("Sources/StreamChatUI/Resources")]
```
b-onc, line 41: Does it work when you don't put `path: "Sources/StreamChatUI"`? I didn't know SPM was that smart

b-onc (approved):
Looks great! :shipit:

.github/workflows/ci.yml
VojtaStavik: SwiftLint mostly.

b-onc: Why do we need swiftlint in CI?

Package.swift
VojtaStavik: Yes, the default path for sources is `Sources/<target name>`. fun fact: It used to be the only way the first couple of years of SPM. You couldn't even specify your custom folder name.

PR #796, opened 2021-02-03
[CIS-632] Fix race conditions on database observers

VojtaStavik:
We already have a fix like this at other places in the observers. I just extended it to cover the other exposed cases, too.

---

Fixed exactly as @cardoso suggested :trollface: 

Screen Shot 2021-02-03 at 14 40 02

### DIFF
--- a/Sources/StreamChat/Controllers/EntityDatabaseObserver.swift
+++ b/Sources/StreamChat/Controllers/EntityDatabaseObserver.swift
@@ -1,5 +1,5 @@
 //
-// Copyright © 2020 Stream.io Inc. All rights reserved.
+// Copyright © 2021 Stream.io Inc. All rights reserved.
 //
 
 import CoreData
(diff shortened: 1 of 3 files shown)


Sources/StreamChat/Controllers/EntityDatabaseObserver.swift
```diff
@@ -133,7 +137,11 @@ class EntityDatabaseObserver<Item, DTO: NSManagedObject> {
             cacheName: nil
         )
         
-        _item.computeValue = { [unowned self] in
+        _item.computeValue = { [weak self] in
+            // Ideally, this should rather be `unowned`, however, `deinit` is not always called on the same thread as this
+            // callback which can cause a race condition when the object is already being deinited on a different thread.
+            guard let self = self else { return nil }
```
b-onc, line 143: We don't need `[weak self]` here, we can get by with `[weak frc]` and the `guard let fetchedObjects = frc?.fetchedObjects` will take care of it

VojtaStavik: Correct 👍

b-onc (approved):

cardoso (approved):

Sources/StreamChat/Controllers/EntityDatabaseObserver.swift
VojtaStavik: @b-onc whad do you think about 320a228 ?

PR #802, opened 2021-02-07
[CIS-638] Decrease channel launch times

ManWithBear:
Average channel open time (duration from tap to idle state) is 1.3s.
A lot of time takes data base updates and table reloads.
But in most cases such reloads are redundant because object haven;t been changed, only touched and marked as dirty.

By reseting dirty, but not changed, objects I was able to reduce open time from 1.3s to 0.6s.

Rest of time is initial layout setup (VC and message list).  It doesn't make much sense to invest time in this, since layout policy is about to change.

### DIFF
--- a/Sources/StreamChat/Database/DTOs/AttachmentDTO.swift
+++ b/Sources/StreamChat/Database/DTOs/AttachmentDTO.swift
@@ -41,6 +41,16 @@ class AttachmentDTO: NSManagedObject {
     
     @NSManaged var message: MessageDTO
     @NSManaged var channel: ChannelDTO
+
+    override func willSave() {
+        super.willSave()
+
+        // When attachment changed, we need to propagate this change up to holding message
+        if hasPersistentChangedValues, !message.hasChanges {
+            // this will not change object, but mark it as dirty, triggering updates
+            message.id = message.id
+        }
+    }
     
     static func load(id: AttachmentId, context: NSManagedObjectContext) -> AttachmentDTO? {
         let request = NSFetchRequest<AttachmentDTO>(entityName: AttachmentDTO.entityName)
--- a/Sources/StreamChat/Database/DTOs/UserDTO.swift
+++ b/Sources/StreamChat/Database/DTOs/UserDTO.swift
@@ -20,6 +20,9 @@ class UserDTO: NSManagedObject {
     @NSManaged var userUpdatedAt: Date
     
     @NSManaged var flaggedBy: CurrentUserDTO?
+
+    @NSManaged var members: Set<MemberDTO>?
+    @NSManaged var currentUser: CurrentUserDTO?
     
     @NSManaged var savedChangeHash: Int64
     
@@ -30,10 +33,25 @@ class UserDTO: NSManagedObject {
         request.predicate = NSPredicate(format: "id == %@", userId)
         return request
     }
-    
+
     override func willSave() {
         super.willSave()
-        
+
+        // When user changed, we need to propagate this change to members and current user
+        if hasPersistentChangedValues {
+            if let currentUser = currentUser, !currentUser.hasChanges {
+                // this will not change object, but mark it as dirty, triggering updates
+                let assigningPropertyToItself = currentUser.unreadChannelsCount
+                currentUser.unreadChannelsCount = assigningPropertyToItself
+            }
+            for member in members ?? [] {
+                guard !member.hasChanges else { continue }
+                // this will not change object, but mark it as dirty, triggering updates
+                let assigningPropertyToItself = member.channelRoleRaw
+                member.channelRoleRaw = assigningPropertyToItself
+            }
+        }
+
         // Save the current hash of the entity
         // This is used during `save` calls to compare DTOs `changeHash` and
         // corresponding payload's `changeHash` to avoid unnecessary assignment of values.
--- a/Sources/StreamChat/Database/DTOs/UserDTO_Tests.swift
+++ b/Sources/StreamChat/Database/DTOs/UserDTO_Tests.swift
@@ -373,4 +373,82 @@ class UserDTO_Tests: XCTestCase {
             NSSortDescriptor(key: "userRoleRaw", ascending: true)
         )
     }
+
+    func test_userChange_triggerMembersUpdate() throws {
+        // Arrange: Store member and user in database
+        let userId: UserId = .unique
+        let channelId: ChannelId = .unique
+
+        let userPayload: UserPayload<NoExtraData> = .dummy(userId: userId)
+
+        let payload: MemberPayload<NoExtraData> = .init(
+            user: userPayload,
+            role: .moderator,
+            createdAt: .init(timeIntervalSince1970: 4000),
+            updatedAt: .init(timeIntervalSince1970: 5000)
+        )
+
+        try database.writeSynchronously { session in
+            try session.saveMember(payload: payload, channelId: channelId)
+        }
+
+        // Arrange: Observe changes on members
+        let observer = EntityDatabaseObserver<MemberDTO, MemberDTO>(
+            context: database.viewContext,
+            fetchRequest: MemberDTO.member(userId, in: channelId),
+            itemCreator: { $0 }
+        )
+        try observer.startObserving()
+
+        var receivedChange: EntityChange<MemberDTO>?
+        observer.onChange { receivedChange = $0 }
+
+        // Act: Update user
+        try database.writeSynchronously { session in
+            let loadedUser: UserDTO = try XCTUnwrap(session.user(id: userId))
+            loadedUser.name = "Jo Jo"
+        }
+
+        // Assert: Members should be updated
+        XCTAssertNotNil(receivedChange)
+    }
+
+    func test_userChange_triggerCurrentUserUpdate() throws {
+        // Arrange: Store current user in database
+        let userId: UserId = .unique
+
+        let payload: CurrentUserPayload<NoExtraData> = .dummy(
+            userId: userId,
+            role: .admin,
+            extraData: .defaultValue,
+            devices: [DevicePayload.dummy],
+            mutedUsers: [
+                .dummy(userId: .unique)
+            ]
+        )
+
+        try database.writeSynchronously { session in
+            try session.saveCurrentUser(payload: payload)
+        }
+
+        // Arrange: Observe changes on current user
+        let observer = EntityDatabaseObserver<CurrentUserDTO, CurrentUserDTO>(
+            context: database.viewContext,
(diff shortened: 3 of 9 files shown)


Sources/StreamChat/Database/DatabaseContainer.swift
```diff
@@ -152,6 +152,14 @@ class DatabaseContainer: NSPersistentContainer {
             log.debug("Starting a database session.")
             do {
                 try actions(self.writableContext)
+                // If you touch ManagedObject and update one of it properties to same value
+                // Object will be marked as `updated` even it hasn't changed.
+                // By reseting such objects we remove updates that are not updates.
+                for object in self.writableContext.updatedObjects {
+                    if object.changedValues().isEmpty {
+                        self.writableContext.refresh(object, mergeChanges: false)
+                    }
```
VojtaStavik, line 161: Do I understand it correctly, that `object.hasChanges` is set whenever you set a value on the object, but `object.changedValues()` actually evaluates the value and returns only when the value was actually changed?

Does it mean that we can actually revert the changes from https://github.com/GetStream/stream-chat-swift/pull/752 ? cc @b-onc

VojtaStavik (approved):

Sources/StreamChat/Database/DatabaseContainer.swift
b-onc: If that's ⬆️ the case, we definitely don't need that persisted hash workaround, this would be a much cleaner one.
Also, the existence of this PR proves that #752 did not resolve the issue fully.

ManWithBear: Basically yes. `hasChanges` is just dirty flag, so whenever you touch object, it gets marked as with changes.
`changedValues` contains pairs of keys-values of actually changed values.
Here exist additional property `hasPersistentChangedValues` that should work as you would expect from `hasChanges`, but documentation is lucking, so I went with current solution

Sources/StreamChat/Database/DTOs/AttachmentDTO.swift
```diff
@@ -41,6 +41,16 @@ class AttachmentDTO: NSManagedObject {
...
     @NSManaged var message: MessageDTO
     @NSManaged var channel: ChannelDTO
+
+    override func willSave() {
+        super.willSave()
+
+        // When attachment changed, we need to propagate this change up to holding message
+        if hasPersistentChangedValues, !message.hasChanges {
+            // this will not change object, but mark it as dirty, triggering updates
+            message.id = message.id
```
VojtaStavik, line 51: How come this change is not removed by https://github.com/GetStream/stream-chat-swift/pull/802/commits/df7983b65dac26188897ea48caad9bf8cf4d0b03 ?

Sources/StreamChat/Database/DTOs/UserDTO.swift
```diff
@@ -30,10 +33,24 @@ class UserDTO: NSManagedObject {
...
+        if hasPersistentChangedValues {
+            if !(currentUser?.hasChanges ?? false) {
+                // this will not change object, but mark it as dirty, triggering updates
+                currentUser?.unreadChannelsCount = currentUser?.unreadChannelsCount ?? 0
+            }
+            for member in members ?? [] {
+                guard !member.hasChanges else { continue }
+                // this will not change object, but mark it as dirty, triggering updates
+                let swiftBelieveMeIKnowWhatImDoing = member.channelRoleRaw
+                member.channelRoleRaw = swiftBelieveMeIKnowWhatImDoing
```
VojtaStavik, line 50: I understand this is a workaround for the compiler warning. I wouldn't mind if you named it the way such that it's more obvious.

Sources/StreamChat/Database/DTOs/AttachmentDTO.swift
```diff
@@ -41,6 +41,16 @@ class AttachmentDTO: NSManagedObject {
     
     @NSManaged var message: MessageDTO
     @NSManaged var channel: ChannelDTO
+
+    override func willSave() {
+        super.willSave()
+
+        // When attachment changed, we need to propagate this change up to holding message
```
VojtaStavik, line 48: Can you add a test for this, please?

Sources/StreamChat/Database/DTOs/UserDTO.swift
```diff
@@ -30,10 +33,24 @@ class UserDTO: NSManagedObject {
...
     override func willSave() {
         super.willSave()
-        
+
+        // When user changed, we need to propagate this change to members and current user
+        if hasPersistentChangedValues {
+            if !(currentUser?.hasChanges ?? false) {
+                // this will not change object, but mark it as dirty, triggering updates
+                currentUser?.unreadChannelsCount = currentUser?.unreadChannelsCount ?? 0
+            }
```
VojtaStavik, line 45: wouldn't this be more readable and simple using `if let currentUser = currentUser, currentUSer.hasChanges { }` ?

Sources/StreamChat/Database/DTOs/UserDTO_Tests.swift
```diff
@@ -373,4 +373,82 @@ class UserDTO_Tests: XCTestCase {
             NSSortDescriptor(key: "userRoleRaw", ascending: true)
         )
     }
+
+    func test_DTO_triggerMembersUpdate() throws {
+        // Arrange: Store member and user in database
+        let userId = UUID().uuidString
+        let channelId = ChannelId(type: .init(rawValue: "messaging"), id: UUID().uuidString)
```
VojtaStavik, line 380: nit: you can do:
```
let userId: UserId = .unique
let channelId: ChannelId = .unique
```

Sources/StreamChat/Database/DTOs/UserDTO_Tests.swift
```diff
@@ -373,4 +373,82 @@ class UserDTO_Tests: XCTestCase {
...
+
+        let payload: MemberPayload<NoExtraData> = .init(
+            user: userPayload,
+            role: .moderator,
+            createdAt: .init(timeIntervalSince1970: 4000),
+            updatedAt: .init(timeIntervalSince1970: 5000)
+        )
+
+        try database.writeSynchronously { session in
+            try! session.saveMember(payload: payload, channelId: channelId)
```
VojtaStavik, line 392: why `try!` here when the whole method is `throws`?

Sources/StreamChat/Database/DTOs/UserDTO_Tests.swift
```diff
@@ -373,4 +373,82 @@ class UserDTO_Tests: XCTestCase {
...
+        }
+
+        // Arrange: Observe changes on members
+        let observer = ListDatabaseObserver<MemberDTO, MemberDTO>(
+            context: database.viewContext,
+            fetchRequest: MemberDTO.member(userId, in: channelId),
+            itemCreator: { $0 }
+        )
```
VojtaStavik, line 400: nit: You can use `EntityDatabaseObserver`, the functionality is similar it just makes your intentions clearer.

Sources/StreamChat/Database/DTOs/UserDTO_Tests.swift
```diff
@@ -373,4 +373,82 @@ class UserDTO_Tests: XCTestCase {
...
+        try observer.startObserving()
+
+        var receivedChanges: [ListChange<MemberDTO>]?
+        observer.onChange = { receivedChanges = $0 }
+
+        // Act: Update user
+        try database.writeSynchronously { session in
+            let loadedUser: UserDTO = try XCTUnwrap(session.user(id: userId))
+            loadedUser.name = "Jo"
```
VojtaStavik, line 409: nit: did you mean Jo Jo? https://starwars.fandom.com/wiki/Jo_Jo

Sources/StreamChat/Database/DTOs/UserDTO_Tests.swift
```diff
@@ -373,4 +373,82 @@ class UserDTO_Tests: XCTestCase {
             NSSortDescriptor(key: "userRoleRaw", ascending: true)
         )
     }
+
+    func test_DTO_triggerMembersUpdate() throws {
```
VojtaStavik, line 377: The name of the method should reflect what you test. Maybe something like `test_changingUserDTO_triggersMemberDTOUpdate`?

Sources/StreamChat/Database/DTOs/UserDTO_Tests.swift
```diff
@@ -373,4 +373,82 @@ class UserDTO_Tests: XCTestCase {
...
+        try database.writeSynchronously { session in
+            let loadedUser: UserDTO = try XCTUnwrap(session.user(id: userId))
+            loadedUser.name = "Jo"
+        }
+
+        // Assert: Members should be updated
+        XCTAssertFalse(receivedChanges?.isEmpty ?? true)
+    }
+
+    func test_DTO_triggerCurrentUserUpdate() throws {
```
VojtaStavik, line 416: -> `test_changingUserDTO_triggersCurrentUserDTOUpdate` ?

VojtaStavik (commented):
Looking good, I have mostly nit comments. One thing I don't understand is
> How come this change is not removed by df7983b ?

Sources/StreamChat/Database/DTOs/UserDTO.swift
VojtaStavik: WAIT, we shouldn't check `currentUser`'s changes here, right?

Sources/StreamChat/Database/DTOs/AttachmentDTO.swift
ManWithBear: `willSave` is called inside `context.save()`, so our "cleanup" logic is already performed.

Sources/StreamChat/Database/DTOs/AttachmentDTO.swift
ManWithBear: This behavior is already tested by `MessageSender`, e.g. in `test_sender_sendsMessage_withBothAttachmentSeedsAndUploadedAttachments`

VojtaStavik: I see. Seems to me that having an extra test here for this specific behavior will be beneficial for future use cases. If you don't mind adding it it would great.

Sources/StreamChat/Database/DTOs/AttachmentDTO.swift
VojtaStavik: Gotcha 👍

VojtaStavik:
@ManWithBear All right, let's merge this!

PR #803, opened 2021-02-08
[CIS-643] Example Audit for CurrentChatUserAvatarView

DominikBucher12:
# What this PR do:
- Audits `CurrentChatUserAvatarView` UI Components so it's ready for production

# What to do with this?

Very carefully check if this component is ready, tested and works properly and expected within our `StreamChatUI` framework.

 You can go through this checklist: 

- [X] Cleaning up the implementation, removing dead code, make sure it follows the "open by default" rule
- [X] Looking at the component implementation, making sure it implements life-cycle methods correctly
- [x] Add snapshot tests for default UI
- [X] Add tests to verify customization via `defaultAppearace.addRule { }`
- [X] Add tests to verify customization via subclassing (Use custom subclass in tests which implements overriding `View` functions)
- [X] Audit naming and change it if needed to follow common patterns 
- [X] **Add documentation to the component + UIConfig components**


Sources/StreamChatUI/Common Views/CurrentChatUserAvatarView_Tests.swift
```diff
@@ -0,0 +1,23 @@
...
+import StreamChat
+@testable import StreamChatUI
+import XCTest
+
+class CurrentChatUserAvatarView_Tests: XCTestCase {
+    var avatarView: CurrentChatUserAvatarView! = CurrentChatUserAvatarView().withoutAutoresizingMaskConstraints
+
+    func test_default_currentChatUserAvatarView_withImage_looksAsExpected() {
+        let testImage = TestImages.vader.image
+        avatarView.accessibilityPerformMagicTap()
```
VojtaStavik, line 14: What kind of sourcery is this? :)

Sources/StreamChatUI/Common Views/CurrentChatUserAvatarView_Tests.swift
```diff
@@ -3,21 +3,69 @@
...
+
+class CustomCurrentChatUserAvatarView: CurrentChatUserAvatarView {
+    override func setUpAppearance() {
+        super.setUpAppearance()
+
+        backgroundColor = .green
+        avatarView.imageView.backgroundColor = .blue
+        avatarView.imageView.alpha = 0.2
     }
 }
```
VojtaStavik: Ideally in the subclass, you should change something you can't change using the appearance hook, like layout for example.

DominikBucher12: Yep you are right, Will do this in next commit.

Sources/StreamChatUI/Common Views/ChatAvatarView_Tests.swift
```diff
@@ -0,0 +1,23 @@
...
+        avatarView.imageView.image = testImage
+        AssertSnapshot(avatarView)
+    }
+
+    func test_default_avatarView_withoutImage_looksAsExpected() {
+        // What is default behavior of AvatarView when no image?
+        // Shouldn't be there some default placeholder image?
+        AssertSnapshot(avatarView)
+    }
+}
```
DominikBucher12, line 23: This file should be deleted, `ChatAvatarView` is superclass which is not used directly in our codebase.

Sources/StreamChatUI/Common Views/CurrentChatUserAvatarView_Tests.swift
```diff
@@ -10,27 +10,39 @@ import XCTest
 class CurrentChatUserAvatarView_Tests: XCTestCase {
     var defaultAvatarView: CurrentChatUserAvatarView! = CurrentChatUserAvatarView().withoutAutoresizingMaskConstraints
     var customAvatarView: CustomCurrentChatUserAvatarView! = CustomCurrentChatUserAvatarView().withoutAutoresizingMaskConstraints
+    var defaultAppearanceHookAvatarView: CurrentChatUserAvatarView! = CurrentChatUserAvatarView().withoutAutoresizingMaskConstraints
```
VojtaStavik: I don't think we need to store all these variants when they are valid just for one test.

Sources/StreamChatUI/Common Views/CurrentChatUserAvatarView_Tests.swift
```diff
@@ -0,0 +1,86 @@
...
+
+        AssertSnapshot(defaultAppearanceHookAvatarView)
+    }
+
+    func test_customApperanceHook_currentChatUserAvatarView_withoutImage_looksAsExpected() {
+        UIConfig.default.currentUser.currentUserViewAvatarView.defaultAppearance.addRule {
+            $0.avatarView.imageView.isOpaque = true
+            $0.backgroundColor = .brown
+            $0.avatarView.imageView.backgroundColor = .cyan
+        }
```
DominikBucher12, line 43: Will be refactored

Sources/StreamChatUI/Common Views/CurrentChatUserAvatarView_Tests.swift
```diff
@@ -0,0 +1,86 @@
...
+class CustomCurrentChatUserAvatarView: CurrentChatUserAvatarView {
+    override func setUpAppearance() {
+        super.setUpAppearance()
+
+        avatarView.imageView.isOpaque = true
+        backgroundColor = .green
+        avatarView.imageView.backgroundColor = .blue
+        avatarView.imageView.alpha = 0.2
+    }
+}
```
DominikBucher12, line 84: Will be changed to `setupLayout`.

Sources/StreamChatUI/Common Views/CurrentChatUserAvatarView_Tests.swift
```diff
@@ -0,0 +1,86 @@
...
+
+class DefaultAppearanceHookCurrentChatUserAvatarView: CurrentChatUserAvatarView {}
```
DominikBucher12, line 86: Will be deleted :D

Sources/StreamChatUI/Common Views/CurrentChatUserAvatarView_Tests.swift
```diff
@@ -10,27 +10,39 @@ import XCTest
...
     var mockedCurrentUserController: CurrentChatUserController_Mock<NoExtraData>! = .mock()
 
-//    test_defaultAppearance() ✅
-//    test_contentIsLoaded() // klidne vic casu, treba "s" a "bez" online indicator
-//    test_customDefaultAppearance()
-//    test_customSubclassing() ✅
+    override func setUp() {
+        super.setUp()
+        UIConfig.default.currentUser.currentUserViewAvatarView.defaultAppearance.rules.removeAll()
+    }
```
VojtaStavik, line 20: This is because of the appearance hooks tests? Maybe it will be easier to do something like:
```
class AppearanceHookTestAvatarView: CurrentChatUserAvatarView {}
AppearanceHookTestAvatarView.defaultApperance.addRule { ... }
```
so we don't have to mess with the global config

Sources/StreamChatUI/Common Views/CurrentChatUserAvatarView_Tests.swift
```diff
@@ -10,27 +10,39 @@ import XCTest
...
             imageURL: TestImages.yoda.url
         )
-        defaultAvatarView.controller = mockedCurrentUserController
+        defaultAppearanceHookAvatarView.controller = mockedCurrentUserController
 
-        defaultAvatarView.applyDefaultAppearance()
-
-        AssertSnapshot(defaultAvatarView, record: true)
+        AssertSnapshot(defaultAppearanceHookAvatarView)
     }
```
VojtaStavik: Maybe this can be simplified like:
```swift
func test_customizationUsingDefaultAppearanceRule() {
    class TestAvatarView: CurrentChatUserAvatarView {}
    TestAvatarView.defaultAppearance {
        $0.avatarView.imageView.isOpaque = true
        $0.backgroundColor = .brown
        $0.avatarView.imageView.backgroundColor = .cyan
    }

    mockedCurrentUserController.currentUser_mock = .init(
        id: "Yoda",
        imageURL: TestImages.yoda.url
    )

    let view = TestAvatarView()
    view.controller = mockedCurrentUserController

    AssertSnapshot(view, record: true)
}
```

Sources/StreamChatUI/Common Views/CurrentChatUserAvatarView_Tests.swift
VojtaStavik: You can't do this in tests. These values must be set and reset in `setUp` and `tearDown` methods. Please look at LLC tests how it's done there.

DominikBucher12: Hmm. Why even keep those locally in test? I changed this a bit so the values are created just in scope of the test. Let's see...

Sources/StreamChatUI/Common Views/CurrentChatUserAvatarView_Tests.swift
DominikBucher12: Thanks :)

Sources/StreamChatUI/Common Views/CurrentChatUserAvatarView_Tests.swift
DominikBucher12: I think resetting the config is no longer needed as I can do this in a way as you suggested with class at function scope 1 comment below.

Sources/StreamChatUI/Common Views/CurrentChatUserAvatarView_Tests.swift
DominikBucher12: See other comments

Sources/StreamChatUI/Common Views/CurrentChatUserAvatarView_Tests.swift
DominikBucher12: Changed to `setUpLayout` and `setUpAppearance`

Sources/StreamChatUI/Common Views/CurrentChatUserAvatarView_Tests.swift
DominikBucher12: A resolved one :D

DominikBucher12:
Need to regenerate the snapshot tests for iPhone 12 device too as pointed out by @nuno-vieira , thanks!

Sources/StreamChatUI/Common Views/CurrentChatUserAvatarView_Tests.swift
```diff
@@ -0,0 +1,107 @@
...
+
+        mockedCurrentUserController.currentUser_mock = .init(
+            id: "Yoda",
+            imageURL: TestImages.yoda.url
+        )
+
+        let view = TestAvatarView().withoutAutoresizingMaskConstraints
+        view.controller = mockedCurrentUserController
+
+        AssertSnapshot(view)
```
nuno-vieira, line 30:
```suggestion
        AssertSnapshot(view, varaints = [.small])
```
Because here we are testing an image only, I don't think it is worth to test all the variants, since there's no text. Can you change the variants to only one please?

Sources/StreamChatUI/Common Views/CurrentChatUserAvatarView_Tests.swift
```diff
@@ -0,0 +1,107 @@
...
+        TestAvatarView.defaultAppearance {
+            $0.avatarView.imageView.isOpaque = true
+            $0.backgroundColor = .brown
+            $0.avatarView.imageView.backgroundColor = .cyan
+        }
+
+        let customAppearanceAvatarView = TestAvatarView().withoutAutoresizingMaskConstraints
+        customAppearanceAvatarView.controller = mockedCurrentUserController
+
+        AssertSnapshot(customAppearanceAvatarView)
```
nuno-vieira, line 45: same here

Sources/StreamChatUI/Common Views/CurrentChatUserAvatarView_Tests.swift
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    func test_defaultAppearance_currentChatUserAvatarView_withImage_looksAsExpected() {
+        mockedCurrentUserController.currentUser_mock = .init(
+            id: "Yoda",
+            imageURL: TestImages.yoda.url
+        )
+        let defaultAppearanceAvatarView = CurrentChatUserAvatarView().withoutAutoresizingMaskConstraints
+        defaultAppearanceAvatarView.controller = mockedCurrentUserController
+        AssertSnapshot(defaultAppearanceAvatarView)
```
nuno-vieira, line 55: same here

Sources/StreamChatUI/Common Views/CurrentChatUserAvatarView_Tests.swift
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    func test_defaultAppearance_currentChatUserAvatarView_withoutImage_looksAsExpected() {
+        let defaultAppearanceAvatarView = CurrentChatUserAvatarView().withoutAutoresizingMaskConstraints
+        AssertSnapshot(defaultAppearanceAvatarView)
```
nuno-vieira, line 60: same here

Sources/StreamChatUI/Common Views/CurrentChatUserAvatarView_Tests.swift
```diff
@@ -0,0 +1,107 @@
...
+        }
+
+        let customAvatarView = CustomCurrentChatUserAvatarView().withoutAutoresizingMaskConstraints
+
+        mockedCurrentUserController.currentUser_mock = .init(
+            id: "Yoda",
+            imageURL: TestImages.yoda.url
+        )
+        customAvatarView.controller = mockedCurrentUserController
+        AssertSnapshot(customAvatarView)
```
nuno-vieira, line 85: same here

Sources/StreamChatUI/Common Views/CurrentChatUserAvatarView_Tests.swift
```diff
@@ -0,0 +1,107 @@
...
+            override open func setUpLayout() {
+                widthAnchor.pin(equalToConstant: 30).isActive = true
+                heightAnchor.pin(equalToConstant: 50).isActive = true
+                embed(avatarView)
+            }
+        }
+
+        let customAvatarView = CustomCurrentChatUserAvatarView().withoutAutoresizingMaskConstraints
+
+        AssertSnapshot(customAvatarView)
```
nuno-vieira, line 105: And here 👍

Sources/StreamChatUI/Common Views/CurrentChatUserAvatarView_Tests.swift
DominikBucher12: Thanks :) Done

Sources/StreamChatUI/Common Views/CurrentChatUserAvatarView_Tests.swift
DominikBucher12: Done

Sources/StreamChatUI/Common Views/CurrentChatUserAvatarView_Tests.swift
DominikBucher12: Done

Sources/StreamChatUI/Common Views/CurrentChatUserAvatarView_Tests.swift
DominikBucher12: Done

Sources/StreamChatUI/Common Views/CurrentChatUserAvatarView_Tests.swift
DominikBucher12: Done

Sources/StreamChatUI/Common Views/CurrentChatUserAvatarView_Tests.swift
DominikBucher12: ✔️

Sources/StreamChatUI/ChatChannelList/ChatChannelAvatarView.swift
```diff
@@ -10,6 +10,8 @@ public typealias ChatChannelAvatarView = _ChatChannelAvatarView<NoExtraData>
 open class _ChatChannelAvatarView<ExtraData: ExtraDataTypes>: ChatAvatarView, UIConfigProvider {
     // MARK: - Properties
 
+    open var onlineIndicatorWidthRatio: CGFloat = 0.3
```
VojtaStavik, line 13: missing documentation

Sources/StreamChatUI/Common Views/ChatAvatarView.swift
```diff
@@ -7,11 +7,7 @@ import UIKit
 open class ChatAvatarView: View {
     // MARK: - Subviews
     
-    public let imageView: UIImageView = {
-        let imageView = UIImageView()
-        imageView.translatesAutoresizingMaskIntoConstraints = false
-        return imageView
-    }()
+    open var imageView: UIImageView = UIImageView().withoutAutoresizingMaskConstraints
```
VojtaStavik, line 14: 👍 for making this open, but to make sure it still works correctly, shouldn't we make it `private(set)`? Aka - you can provide your custom object in subclasses but you can't modify it during the runtime.

Sources/StreamChatUI/Common Views/CurrentChatUserAvatarView.swift
```diff
@@ -60,6 +61,8 @@ open class _CurrentChatUserAvatarView<ExtraData: ExtraDataTypes>: Control, UICon
     override open func setUpLayout() {
         super.setUpLayout()
 
+        defaultIntrinsicContentSize = .init(width: 32, height: 32)
+
```
VojtaStavik, line 69: Hmm, this looks more correct but that means the order of the lifecycle methods is wrong. We should probably call them like this:
```swift
        setUp()
        setUpLayout()
        (self as! Self).applyDefaultAppearance()
        setUpAppearance()
        updateContent()

```

Sources/StreamChatUI/Common Views/CurrentChatUserAvatarView_Tests.swift
```diff
@@ -0,0 +1,107 @@
...
+        TestAvatarView.defaultAppearance {
+            $0.avatarView.imageView.isOpaque = true
+            $0.backgroundColor = .brown
+            $0.avatarView.imageView.backgroundColor = .cyan
+        }
+
+        mockedCurrentUserController.currentUser_mock = .init(
+            id: "Yoda",
+            imageURL: TestImages.yoda.url
+        )
```
VojtaStavik, line 25: this is kind of code that should be in `setUp() { }`

VojtaStavik (approved):
Looks really good @DominikBucher12, I have just some nit comments, thank you 👍 Can you please also add documentation to all non-system public methods and variables we expose?

nuno-vieira (approved):
LGTM!

Sources/StreamChatUI/ChatChannelList/ChatChannelAvatarView.swift
DominikBucher12: Fixed.

Sources/StreamChatUI/Common Views/CurrentChatUserAvatarView_Tests.swift
DominikBucher12: Yep, thanks done :)

Sources/StreamChatUI/Common Views/CurrentChatUserAvatarView.swift
DominikBucher12: Thanks, you ar right. Changed for all...

Sources/StreamChatUI/Common Views/ChatAvatarView.swift
DominikBucher12: Thanks, you are right :) Changed.

Sources/StreamChatUI/Common Views/BaseViews.swift
```diff
@@ -50,8 +50,8 @@ open class View: UIView, AppearanceSetting, Customizable {
         guard superview != nil else { return }
         
         setUp()
-        (self as! Self).applyDefaultAppearance()
         setUpAppearance()
+        (self as! Self).applyDefaultAppearance()
         setUpLayout()
         updateContent()
```
VojtaStavik, line 56: This is not the order I was suggesting in the comment 🙃

DominikBucher12: Edited the commit, now it should be 😉

DominikBucher12:
Looks like change introduced in 8369024675021cd189159086a2b9f46d84bd9bea  breaks a bit the UI and caused bad side effect (maybe not side effect at all?) 

Difference:
difference_3_C2391D92-AF43-4C3E-A895-FF17F6009F8F

Let me explain:

In `ChatChannelListItemView` we set  `subtitleLabel.font = uiConfig.font.subheadline` in `defaultAppearance` Customizable method. The order of calling those customizable method is after the commit mentioned above changed

What happens now is that we setup layout which is dependent in some cases like `ChatChannelListItemView` on Appearance sooner than the appearance is set and therefore we have bad values for spacings in the `ChatChannelListItemView`

Sources/StreamChatUI/Common Views/CurrentChatUserAvatarView_Tests.swift
nuno-vieira: It actually makes sense to change to `setupLayout` ?  setup is 1 word not 2 🤔

Sources/StreamChatUI/Utils/UIImageView+Extensions.swift
```diff
@@ -1,12 +1,13 @@
...
 
 import Nuke
 import UIKit
 
 extension UIImageView {
-    func setImage(from url: URL, placeholder: UIImage? = nil) {
+    func loadImage(from url: URL?, placeholder: UIImage? = nil) {
+        guard let url = url else { image = nil; return }
         let options = ImageLoadingOptions(placeholder: placeholder)
         Nuke.loadImage(with: url, options: options, into: self)
```
nuno-vieira, line 12: I remember we had big images in some attachments, and It looks that Nuke does not downsample the images by default. In order to do it, we should add: `processors: [ImageProcessors.Resize(size: imageView.bounds.size)]`. This is out of the scope of this PR, but it will probably improve the performance of the chat significantly 🤔

Sources/StreamChatUI/ChatChannelList/ChatChannelAvatarView_Tests.swift
```diff
@@ -0,0 +1,131 @@
...
+        ])
+    }
+    
+    func test_emptyAppearance() {
+        let view = ChatChannelAvatarView().withoutAutoresizingMaskConstraints
+        view.addSizeConstraints()
+        AssertSnapshot(view, variants: [.small])
+    }
+
+    func test_defaultAppearance_withDirectMessageChannel() {
```
nuno-vieira, line 30: on `ChannelListVC_Tests` we have the pattern `test_channelListVC_defaultAppearance`. We should have the same pattern in all places. Although the latter is a bit repetitive, I think it generates a better snapshot name, especially when thinking of storing the artefacts on Github, it will be easier to implement, and it will be easier to understand what view the snapshot belongs to. I'm fine with both strategies, we just should make sure that we have the same pattern always 🙂 I guess if we have this strategy, we can just store the failure snapshots in a folder with the name of the view, so the decision will be just a question of taste 👍

Sources/StreamChatUI/ChatChannelList/ChatChannelAvatarView_Tests.swift
```diff
@@ -0,0 +1,131 @@
...
+    func test_appearanceCustomization_usingAppearanceHook() {
+        class TestView: ChatChannelAvatarView {}
+        TestView.defaultAppearance {
+            // Modify appearance
+            $0.onlineIndicatorView.backgroundColor = .red
+            $0.backgroundColor = .yellow
+
+            // Modify layout
+            NSLayoutConstraint.activate([
+                // swiftlint:disable constraints_forbidden
```
nuno-vieira, line 81: Maybe better to disable this rule for UI_Tests ?

nuno-vieira (approved):
Looks great 👌 Just some small questions :)

Tests/Shared/ChannelId.swift
```diff
@@ -1,5 +1,5 @@
 //
-// Copyright © 2020 Stream.io Inc. All rights reserved.
+// Copyright © 2021 Stream.io Inc. All rights reserved.
```
DominikBucher12, line 2: Why is this here :(

.swiftlint.yml
```diff
@@ -1,4 +1,6 @@
 excluded:
+  - "Scripts"
```
DominikBucher12, line 2: <3

Sources/StreamChatTestTools/Models/ChatChannel_Mock.swift
```diff
@@ -102,4 +102,48 @@ public extension _ChatChannel {
...
+        name: String? = nil,
+        imageURL: URL? = nil,
+        lastMessageAt: Date? = nil,
+        createdAt: Date = .init(),
+        updatedAt: Date = .init(),
+        deletedAt: Date? = nil,
+        createdBy: _ChatUser<ExtraData.User>? = nil,
+        config: ChannelConfig = .mock(),
+        isFrozen: Bool = false,
+        members: Set<_ChatChannelMember<ExtraData.User>> = [],
```
DominikBucher12, line 117: A corner for thinking and probably nit: 

DM should be always 2 people, what happens when you set 3 members to DM Channel? 
I know this is mock and everything, but a bit more convenient from API user point would be if you would have just `member` argument and set the 1-element set in this function...

Sources/StreamChatUI/Utils/UIImageView+Extensions.swift
VojtaStavik: Good idea @nuno-vieira , can you please put it to JIRA?

Sources/StreamChatUI/ChatChannelList/ChatChannelAvatarView_Tests.swift
VojtaStavik: Hmm, I don't think we need to use the name of the view we test in the test name, since it's already in the call name. The snapshot are organized by the test case name in folders and it seems to me it's clear enough. WDYT?

Screen Shot 2021-02-11 at 12 35 27

Sources/StreamChatUI/ChatChannelList/ChatChannelAvatarView_Tests.swift
VojtaStavik: Yeah, I was thinking about it but wasn't sure. It probably makes sense. Implemented in 26f2db2

Tests/Shared/ChannelId.swift
VojtaStavik: The file was moved and the formatter updates the header automatically. What's wrong with it? :)

Sources/StreamChatTestTools/Models/ChatChannel_Mock.swift
VojtaStavik: Actually, that's not entirely true. We call it direct message but it should be rather called something like member-defined groups. They are channels implicitly identified by their members. Aka, if you can have only one channel with the given combination of members.

Tests/Shared/ChannelId.swift
DominikBucher12: Ok it was moved, It was suspicious that it remained untouched and only header changed :)

Sources/StreamChatUI/ChatChannelList/ChatChannelAvatarView_Tests.swift
nuno-vieira: Yes for me its good like it is 👍

PR #806, opened 2021-02-08
[CIS-636] Validate database flushing logic

nuno-vieira:
# In this PR
Adds test coverage to validate that the Database is flushed when it fails to load the persistent store (For example, when there are significant DB changes that would require a migration).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Improve serialization performance by exposing items as `LazyCachedMapCollection` instead of `Array` [#776](https://github.com/GetStream/stream-chat-swift/pull/776)
 - Reduce amount of fake updates by erasing touched objects [#802](https://github.com/GetStream/stream-chat-swift/pull/802)
 - Trigger members and current user updates on UserDTO changes [#802](https://github.com/GetStream/stream-chat-swift/pull/802)
+- Extracts the connection handling responsibility of `CurrentUserController` to a new `ChatConnectionController`. [#804](https://github.com/GetStream/stream-chat-swift/pull/804)
 
 ### 🐞 Fixed
 - Fix race conditions in database observers [#796](https://github.com/GetStream/stream-chat-swift/pull/796)
--- a/Sources/StreamChat/Database/DatabaseContainer_Tests.swift
+++ b/Sources/StreamChat/Database/DatabaseContainer_Tests.swift
@@ -175,6 +175,26 @@ class DatabaseContainer_Tests: StressTestCase {
         XCTAssertNil(testObject)
     }
     
+    func test_databaseContainer_createsNewDatabase_whenPersistentStoreFailsToLoad() throws {
+        // Create a new on-disc database with the test data model
+        let dbURL = URL.newTemporaryFileURL()
+        _ = try DatabaseContainerMock(
+            kind: .onDisk(databaseFileURL: dbURL),
+            modelName: "TestDataModel",
+            bundle: Bundle(for: DatabaseContainer_Tests.self)
+        )
+                
+        // Create a new database with the same url but totally different models
+        // Should re-create the database
+        XCTAssertNoThrow(
+            try DatabaseContainerMock(
+                kind: .onDisk(databaseFileURL: dbURL),
+                modelName: "TestDataModel2",
+                bundle: Bundle(for: DatabaseContainer_Tests.self)
+            )
+        )
+    }
+    
     func test_databaseContainer_hasDefinedBehaviorForInMemoryStore_whenShouldFlushOnStartIsTrue() throws {
         // Create a new in-memory database that should flush on start and assert no error is thrown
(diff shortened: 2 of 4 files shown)


Sources/StreamChat/Database/DatabaseContainer_Mock.swift
```diff
@@ -71,6 +74,19 @@ class DatabaseContainerMock: DatabaseContainer {
         resetEphemeralValues_called = true
         super.resetEphemeralValues()
     }
+    
+    override func loadPersistentStores(completionHandler block: @escaping (NSPersistentStoreDescription, Error?) -> Void) {
+        // If shouldFailLoadingStore is true and the recreating of the store has not been called yet, it should return error
+        if shouldFailLoadingStore, !recreatePersistentStore_called {
+            super.loadPersistentStores { store, _ in
+                block(store, NSError(domain: "test", code: 1, userInfo: nil))
```
VojtaStavik, line 82: nit: You can use `TestError()` for cases like this.

VojtaStavik (commented):
A huge 👍 for creativity when figuring out how to test this. The only thing I'm afraid of is that the test uses too much knowledge of the inner working of `DatabaseContainer`.

What if we structured the test a bit differently:
1. Create two different CoreData models (like we do with `TestDataModel` in other tests) and have them completely different (i.e. containing a different entity)
2. We create `DatabaseContainer` with the first model
3. We try to create another container with the second model but using the same URL
4. We just need to check the container is created correctly.

Sources/StreamChat/Database/DatabaseContainer_Mock.swift
nuno-vieira: > A huge 👍 for creativity when figuring out how to test this. The only thing I'm afraid of is that the test uses too much knowledge of the inner working of `DatabaseContainer`.
> 
> What if we structured the test a bit differently:
> 
> 1. Create two different CoreData models (like we do with `TestDataModel` in other tests) and have them completely different (i.e. containing a different entity)
> 2. We create `DatabaseContainer` with the first model
> 3. We try to create another container with the second model but using the same URL
> 4. We just need to check the container is created correctly.

Yes, I agree, when the system under test is also being mocked, it is very difficult to test some stuff. In this case, your suggestion is better since it reflects the real use case better than the creative approach 👍 Also, adding logic to mocks is a red flag (but I case in this case there wouldn't be another way)

@VojtaStavik Please re-review :)

Sources/StreamChat/Database/DatabaseContainer_Tests.swift
```diff
@@ -175,6 +175,26 @@ class DatabaseContainer_Tests: StressTestCase {
...
+                
+        // Create a new database with the same url but totally different models
+        database = try DatabaseContainerMock(
+            kind: .onDisk(databaseFileURL: dbURL),
+            modelName: "TestDataModel2",
+            bundle: Bundle(for: DatabaseContainer_Tests.self)
+        )
+        
+        // Should re-create the database
+        XCTAssertNotNil(database)
```
VojtaStavik, line 195: I think it's not possible for the program to actually reach this point when `database` is `nil`. So, this test basically tests that the DB init doesn't throw. So maybe we can even do something like this:
```swift
// Create a new on-disc database with the test data model
         let dbURL = URL.newTemporaryFileURL()
         _ = try DatabaseContainerMock(
             kind: .onDisk(databaseFileURL: dbURL),
             modelName: "TestDataModel",
             bundle: Bundle(for: DatabaseContainer_Tests.self)
         )

        // Create a new database with the same url but totally different models and assert it doesn't throw
        XCTAssertNoThrow(
            try DatabaseContainerMock(
                kind: .onDisk(databaseFileURL: dbURL),
                modelName: "TestDataModel2",
                bundle: Bundle(for: DatabaseContainer_Tests.self)
            )
        )
```

VojtaStavik (approved):
Just a small suggestion on how to improve the test syntax. Once fixed, good to merge 👍

Sources/StreamChat/Database/DatabaseContainer_Tests.swift
nuno-vieira: Good catch! 🤦

nuno-vieira: Done ✅

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### 🔄 Changed
 - Improve serialization performance by exposing items as `LazyCachedMapCollection` instead of `Array` [#776](https://github.com/GetStream/stream-chat-swift/pull/776)
+- Extracts the connection handling responsibility of `CurrentUserController` to a new `ConnectionController`. [#804](https://github.com/GetStream/stream-chat-swift/pull/804)
```
VojtaStavik, line 19: `ChatConnectionController` right?

nuno-vieira: ✅

PR #808, opened 2021-02-09
[CIS-640] Separate `ChatChannelDefaultAttachment` into concrete attachment type models

mantergo:
**In this PR:**

- Remove obsolete `AttachmentDTO` field
- Use `ChatMessageAttachmentSeed` for uploading attachment
- Reorganise current attachment models
- Introduce `ChatMessageLinkAttachment`
- Introduce `ChatMessageImageAttachment`
- Introduce `ChatMessageFileAttachment`
- Introduce `ChatMessageGiphyAttachment`
- Make use of separated attachment models
- Group attachment related models


Sources/StreamChat/Models/AttachmentTypes.swift
```diff
@@ -7,205 +7,14 @@ import Foundation
 /// A protocol for attachment model objects that are created from database `AttachmentDTO` objects and exposed.
 public protocol ChatMessageAttachment {
     var type: AttachmentType { get }
-    var id: AttachmentId? { get }
+    var id: AttachmentId? { get set }
```
VojtaStavik: Why this change?

Sources/StreamChat/Models/ChatMessageLinkAttachment.swift
```diff
@@ -0,0 +1,66 @@
...
+        case author = "author_name"
+        case ogURL = "og_scrape_url"
+        case titleLink = "title_link"
+        case imageURL = "image_url"
+        case thumbURL = "thumb_url"
+        case assetURL = "asset_url"
+    }
+    
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
```
VojtaStavik: We should also check that the type matches the expected one. It's possible that there will be a custom attachment with the same structure but different type.

Sources/StreamChat/Models/ChatMessageLinkAttachment.swift
```diff
@@ -0,0 +1,66 @@
...
+
+import Foundation
+
+/// A type representing an attachment of type `link`.
+/// More info about enriched `URLs` can be found here:
+/// https://getstream.io/chat/docs/node/message_format/?language=swift#attachment-format
+public struct ChatMessageLinkAttachment: ChatMessageAttachment, AttachmentEnvelope, Decodable {
+    public var type: AttachmentType { .link }
+    /// A unique identifier of the attachment.
+    public var id: AttachmentId?
```
VojtaStavik: Why is this `var`? 🤔

Sources/StreamChat/Models/ChatMessageImageAttachment.swift
```diff
@@ -0,0 +1,71 @@
...
+        case title
+        case type
+        case image
+        case url
+        case name
+        case thumbURL = "thumb_url"
+        case fallback
+        case imageURL = "image_url"
+        case assetURL = "asset_url"
+    }
```
VojtaStavik: Maybe we should have a common `CodingKeys` for all built-in types? Like we have it for user payloads, message payloads, etc.

Sources/StreamChat/Database/DTOs/AttachmentDTO.swift
```diff
@@ -130,16 +130,31 @@ extension NSManagedObjectContext: AttachmentDatabaseSession {
...
+        if isAttachmentModelSeparationChangesApplied {
+            var attachment: Encodable
+            
+            switch seed.type {
+            case .image:
+                attachment = ChatMessageImageAttachment(title: seed.fileName)
+            case .file:
+                attachment = ChatMessageFileAttachment(title: seed.fileName, file: seed.file)
+            default:
+                throw ClientError.AttachmentSeedUploading(id: id)
```
VojtaStavik, line 142: This will go away, right?

VojtaStavik (approved):
Looks like you had a lot of fun :) I must say I have a hard time wrapping my head around having both systems implemented at the same time. I have some questions but as far as I understand it looks good 👍 I'll review properly in the PR where we will remove the old implementation and everything will be clearer. Thanks!

Sources/StreamChat/Models/AttachmentTypes.swift
mantergo: Cause the object itself is decoded from the original payload and `id` is generated and set on the DB level.
Before introducing concrete types we were manipulating the `ChatMessageDefaultAttachment` type and the `id` property was `var` already.

Sources/StreamChat/Models/ChatMessageLinkAttachment.swift
mantergo: Nice, thanks!

Sources/StreamChat/Models/ChatMessageImageAttachment.swift
mantergo: Nice catch, thanks!

Sources/StreamChat/Database/DTOs/AttachmentDTO.swift
mantergo: I'm not sure. 
We should somehow throw an error when `Seed` is created with the type that we don't know how to upload.

mantergo:
Can you also please check the last commit with the `.link` type improvement, @VojtaStavik ?

PR #814, opened 2021-02-11
[CIS-628] Make CurrentUserController a DataController

nuno-vieira:
# In this PR
Makes `CurrentUserController` a `DataController`.


Sources/StreamChat/Workers/CurrentUserUpdater.swift
```diff
@@ -0,0 +1,166 @@
...
+
+/// Updates current user data to the backend and updates local storage.
+class CurrentUserUpdater<ExtraData: ExtraDataTypes>: Worker {
+    /// Load the most recent current user data.
+    ///
+    /// - Parameters:
+    ///   - currentUserId: The current user identifier.
+    ///   - completion: Called when the API call is finished. Called with `Error` if the remote update fails.
+    ///
+    func loadUser(_ currentUserId: UserId, completion: ((Error?) -> Void)? = nil) {
```
VojtaStavik, line 16: I thought we agreed we're not going to make this call since the local data are always updated.

Sources/StreamChat/Controllers/CurrentUserController/CurrentUserController.swift
```diff
@@ -94,17 +98,39 @@ public class _CurrentChatUserController<ExtraData: ExtraDataTypes>: Controller,
...
+        
+        guard let currentUserId = currentUser?.id else {
+            completion?(ClientError.CurrentUserDoesNotExist())
+            return
+        }
+        
+        currentUserUpdater.loadUser(currentUserId) { error in
+            self.state = error == nil ? .remoteDataFetched : .remoteDataFetchFailed(ClientError(with: error))
+            self.callback { completion?(error) }
+        }
```
VojtaStavik, line 119: We agreed we're not going to do this and make just the call for devices, since that the only thing from the user data that's not being updated via events.

Sources/StreamChat/Controllers/CurrentUserController/CurrentUserController.swift
```diff
@@ -257,21 +234,12 @@ public extension _CurrentChatUserController {
             completion?(ClientError.CurrentUserDoesNotExist())
             return
         }
-        let completion: ((Error?) -> Void) = { error in
+        
+        currentUserUpdater.updateDevices(currentUserId: currentUserId) { error in
```
VojtaStavik, line 238: `updateDevices` shouldn't be even exposed in public API

VojtaStavik (changes requested):
I like the refactoring and introduction of `CurrentUserUpdater`. However, we talked about what the controller should behave like when calling `synchronize` and this implementation is different.

Sources/StreamChat/Controllers/CurrentUserController/CurrentUserController.swift
nuno-vieira: Discussed on slack, will change to updateDevices() :)

Sources/StreamChat/Workers/CurrentUserUpdater.swift
nuno-vieira: Discussed on a call, will change to updateDevices() :)

Sources/StreamChat/Controllers/CurrentUserController/CurrentUserController_Tests.swift
```diff
@@ -40,47 +41,94 @@ final class CurrentUserController_Tests: StressTestCase {
...
-        // Save user to the db
-        try client.databaseContainer.writeSynchronously {
-            try $0.saveCurrentUser(payload: userPayload)
-        }
+    
+    // MARK: - Synchronize tests
+    
+    func test_synchronize_localDataIsAvailable() {
+        let expectedId = UserId.unique
+        let expectedUnreadCount = UnreadCount(channels: 10, messages: 212)
```
VojtaStavik, line 49: nit: It would be better to use something like `.unique` here, too

Sources/StreamChat/Controllers/CurrentUserController/CurrentUserController_Tests.swift
```diff
@@ -40,47 +41,94 @@ final class CurrentUserController_Tests: StressTestCase {
...
+        XCTAssertEqual(controller.state, .initialized)
+        
+        // Simulate no current user
+        env.currentUserObserverItem = nil
+        
+        // Simulate `synchronize` call.
+        controller.synchronize()
+        
+        // Check if state changed after successful network call.
+        XCTAssertEqual(controller.state, .localDataFetched)
```
VojtaStavik, line 94: This comment doesn't seem correct. Also, you should check that the synchronize call returns an error when there's no current user.

VojtaStavik (approved):
Great, just a couple of nits 👍

PR #815, opened 2021-02-12
[CIS-661] Simplify attachment send API

mantergo:
**In this PR:**

- Simplify attachment send API

### DIFF
--- a/Sources/StreamChat/Models/Attachments/ChatMessageAttachmentSeed.swift
+++ b/Sources/StreamChat/Models/Attachments/ChatMessageAttachmentSeed.swift
@@ -5,7 +5,7 @@
 import Foundation
 
 /// A type designed to combine all the information required to create new attachment that needs to be uploaded before sending.
-public struct ChatMessageAttachmentSeed: Hashable {
+public struct ChatMessageAttachmentSeed: AttachmentEnvelope, Hashable {
     /// A local url the data for uploading will be taken from.
     /// When an attachment in uploaded and a message is sent the `localURL` of resulting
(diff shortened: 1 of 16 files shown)


b-onc (approved):
IMHO it looks much nicer now 🎉 
Pro: Simplified API, unified flow
Con: Docs are basically required to understand these things, but `seeds` were not self-explanatory anyway

Sources/StreamChat/Models/Attachments/ChatMessageAttachmentSeed.swift
```diff
@@ -55,6 +55,9 @@ public struct ChatMessageAttachmentSeed: Hashable {
         self.type = type
         self.localState = localState
     }
+    
+    // Dummy encodable conformance to satisfy `AttachmentEnvelope` protocol.
+    public func encode(to encoder: Encoder) throws {}
```
VojtaStavik, line 60: This is an interesting side effect. I'm thinking - what does this mean? Maybe we're missing something?

VojtaStavik (approved):
Good idea @mantergo and @b-onc  👍

Sources/StreamChat/Models/Attachments/ChatMessageAttachmentSeed.swift
mantergo: `Encodable` is the main constraint for the custom attachments cause it is the minimum we need to blindly send it to the backend.
Dummy conformance means that we can't blindly send it, some pre-processing is required and we know how to handle that.
I think that's acceptable trade-off.

PR #816, opened 2021-02-12
[CIS-662] Audit ChatChannelCreateNewButton

nuno-vieira:
# In this PR

Audit of ChatChannelCreateNewButton:
- [ ]  The empty state is tested (N/A ?)


Tests/StreamChatUITests/AssertSnapshot.swift
```diff
@@ -5,6 +5,12 @@
 import SnapshotTesting
 import XCTest
 
+/// Whether or not to record all new references.
+var isRecording: Bool {
+    get { SnapshotTesting.isRecording }
+    set { SnapshotTesting.isRecording = true }
```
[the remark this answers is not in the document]
nuno-vieira: 🤦

Sources/StreamChatUI/Common Views/ChatChannelCreateNewButton.swift
```diff
@@ -5,16 +5,19 @@
...
+open class _ChatChannelCreateNewButton<ExtraData: ExtraDataTypes>: Button, UIConfigProvider {
     override public func defaultAppearance() {
-        defaultIntrinsicContentSize = .init(width: 44, height: 44)
+        super.defaultAppearance()
         setImage(uiConfig.images.newChat, for: .normal)
     }
-
-    open var defaultIntrinsicContentSize: CGSize?
-    override open var intrinsicContentSize: CGSize {
-        defaultIntrinsicContentSize ?? super.intrinsicContentSize
```
nuno-vieira, line 18: Don’t know how I feel about the intrinsicContentSize, I think it is adding more options to change the layout, and we should try to follow the same strategy always. Also, the customer can override the intrinsic content size if they want. WDYT ?

Tests/StreamChatUITests/SnapshotVariant.swift
```diff
@@ -44,3 +48,13 @@ enum SnapshotVariant: String, Hashable, CaseIterable {
...
+
+extension Array where Element == SnapshotVariant {
+    static var allCases: [SnapshotVariant] {
+        SnapshotVariant.allCases
+    }
+    
+    static var lightAndDarkCases: [SnapshotVariant] {
+        SnapshotVariant.lightAndDarkCases
+    }
+}
```
nuno-vieira, line 60: I think these make it easier to use the static properties, instead of using `SnapshotVariant.allCases`, we can simply use `.allCases` in the argument. Also, to add more combinations is also cool IMO.

Tests/StreamChatUITests/SnapshotVariant.swift
```diff
@@ -10,6 +10,10 @@ enum SnapshotVariant: String, Hashable, CaseIterable {
     case large
     case extraExtraExtraLarge
     
+    static var lightAndDarkCases: [SnapshotVariant] {
+        [small, large]
+    }
```
nuno-vieira, line 15: I think this is useful if we are just interested In testing the dark and light variation.
 Another thing I that bug me is that right now there's no way to know which variant is dark or light. We can change this by renaming the cases to include this, for example: `case largeAndLight, smallAndDark`. 

Or at least, we could change the `rawValue: String` to include the dark and light variant? So it is printed in the snapshot file name? Since if the aim of the snapshot is to check how it looks on dark and light, it could be useful to know which one is dark and light, in case they render the same way 🤔

Sources/StreamChatUI/UIConfig+ChannelList.swift
```diff
@@ -13,7 +13,7 @@ public extension _UIConfig {
         public var channelListItemView: _ChatChannelListItemView<ExtraData>.Type = _ChatChannelListItemView<ExtraData>.self
         public var channelViewCell: _ChatChannelListCollectionViewCell<ExtraData>.Type =
             _ChatChannelListCollectionViewCell<ExtraData>.self
-        public var newChannelButton: ChatChannelCreateNewButton<ExtraData>.Type = ChatChannelCreateNewButton<ExtraData>.self
+        public var newChannelButton: _ChatChannelCreateNewButton<ExtraData>.Type = _ChatChannelCreateNewButton<ExtraData>.self
```
DominikBucher12, line 16: Just a reminder that you should document this as well.

Sources/StreamChatUI/Common Views/ChatChannelCreateNewButton.swift
DominikBucher12: You can always override `setupLayout`, not call super and not set any constraints if you wanna use just `intrinsicContentSize` :)

Tests/StreamChatUITests/SnapshotVariant.swift
DominikBucher12: Hmm we discussed in slack that we could use optionSet for the `SnapshotVariants` and make this easily set like `variant: [.small, .dark]` and so, maybe we could create ticket for this and rewrite this a bit

nuno-vieira: Yeah I think the OptionSet is looking to be a better candidate for this. I changed my mind, again 😅

nuno-vieira: Maybe is just better to do it here? This one is pretty small PR, and it will be helpful to change this fast, so the next Audits will get the OptionSet already. WDYT?

DominikBucher12: Sounds good, however we need to carefully check this if this doesn't break anything.

Sources/StreamChatUI/UIConfig+ChannelList.swift
nuno-vieira: ✅ Done!

Tests/StreamChatUITests/SnapshotVariant.swift
nuno-vieira: Yes, that is a good point. Probably is better to make that PR isolated. It should be a quick one though. I'll remove here the commits changing the SnapshotVariant stuff if we agree on migrating to an OptionSet. CC @VojtaStavik

Sources/StreamChatUI/UIConfig+ChannelList.swift
nuno-vieira: Also added on more test, for UIConfig changes if you want to take a look at it 👀

Sources/StreamChatUI/Common Views/ChatChannelCreateNewButton_Tests.swift
```diff
@@ -0,0 +1,80 @@
...
+// Copyright © 2021 Stream.io Inc. All rights reserved.
+//
+
+import StreamChat
+import StreamChatTestTools
+@testable import StreamChatUI
+import XCTest
+
+class ChatChannelCreateNewButton_Tests: XCTestCase {
+    let containerView: UIView = {
```
DominikBucher12, line 11: Hmm I would swear I wrote and published this comment.

The question is... Do we need the containerView?  Why don't we just snapshot the views created inside the tests? IMHO this is redundant.. Or not?

Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
```diff
@@ -34,7 +34,7 @@ open class _ChatChannelListVC<ExtraData: ExtraDataTypes>: ViewController,
         return collection
     }()
     
-    public private(set) lazy var createNewChannelButton: ChatChannelCreateNewButton<ExtraData> = {
+    public private(set) lazy var createNewChannelButton: _ChatChannelCreateNewButton<ExtraData> = {
```
VojtaStavik, line 37: This doesn't have to be `_ChatChannelCreateNewButton<ExtraData>` but it can be just `UIButton`, right? -> less public API surface area and more flexibility for users.

Sources/StreamChatUI/UIConfig+ChannelList.swift
VojtaStavik: Same here, this can also be just `UIButton`

Tests/StreamChatUITests/AssertSnapshot.swift
```diff
@@ -5,6 +5,12 @@
 import SnapshotTesting
 import XCTest
 
+/// Whether or not to record all new references.
+var isRecording: Bool {
+    get { SnapshotTesting.isRecording }
+    set { SnapshotTesting.isRecording = newValue }
+}
+
```
VojtaStavik, line 13: I don't like this tbh. This is dangerous since it can override all existing snapshots. This should be done with extreme caution and adding a convenient API is the exact opposite.

Tests/StreamChatUITests/SnapshotVariant.swift
VojtaStavik: We want to have a unified set of snapshots for every component. So far it seems that we just need to distinguish between views the react to size changes and the ones that don't. I'd suggest creating `userInterfaceStyleCases` as an addition to `allCases` we already have, and use `userInterfaceStyleCases` as the default for the snapshots.

Tests/StreamChatUITests/SnapshotVariant.swift
```diff
@@ -44,3 +48,13 @@ enum SnapshotVariant: String, Hashable, CaseIterable {
         }
     }
 }
+
+extension Array where Element == SnapshotVariant {
+    static var allCases: [SnapshotVariant] {
+        SnapshotVariant.allCases
+    }
+    
+    static var lightAndDarkCases: [SnapshotVariant] {
```
VojtaStavik, line 57: Nice, rename this to `userInterfaceStyleCases` and use it as the default value for snapshots, and I'm happy :)

Sources/StreamChatUI/Common Views/ChatChannelCreateNewButton.swift
```diff
@@ -11,12 +11,13 @@ public typealias ChatChannelCreateNewButton = _ChatChannelCreateNewButton<NoExtr
...
+    
+    override open func setUpLayout() {
+        super.setUpLayout()
+        widthAnchor.pin(equalToConstant: 60).isActive = true
+        heightAnchor.pin(equalToConstant: 60).isActive = true
```
VojtaStavik: I don't think we should do this at all. Let the underlying UIButton instance decide its size.

Sources/StreamChatUI/Common Views/ChatChannelCreateNewButton_Tests.swift
VojtaStavik: I also don't get this

Sources/StreamChatUI/Common Views/ChatChannelCreateNewButton_Tests.swift
```diff
@@ -0,0 +1,80 @@
...
+        return view
+    }()
+    
+    let defaultSize = CGSize(width: 44, height: 44)
+    
+    func test_defaultAppearance() {
+        let view = ChatChannelCreateNewButton().withoutAutoresizingMaskConstraints
+        containerView.addSubview(view)
+        AssertSnapshot(containerView, variants: [.small], size: defaultSize)
+    }
```
VojtaStavik, line 23: look at the test for avatar view, when the element doesn't have a fixed size, you can just add sizing constraints to make it resize properly. I don't think we should use the `size` parameter here.

Sources/StreamChatUI/Common Views/ChatChannelCreateNewButton_Tests.swift
```diff
@@ -0,0 +1,80 @@
...
+class ChatChannelCreateNewButton_Tests: XCTestCase {
+    let containerView: UIView = {
+        let view = UIView(frame: .init(x: 0, y: 0, width: 44, height: 44))
+        view.backgroundColor = UIColor.white
+        return view
+    }()
+    
+    let defaultSize = CGSize(width: 44, height: 44)
+    
+    func test_defaultAppearance() {
```
VojtaStavik, line 19: what about the highlighted and disabled states of the button?

VojtaStavik (changes requested):
Cool, left a couple of comments there but nothing serious :)

Tests/StreamChatUITests/AssertSnapshot.swift
nuno-vieira: This is part of the snapshot framework API, I just exposed it, so we don't need to import the framework everywhere. And yes it should be used carefully, but is very useful when writing new tests and we want to record all tests for that View Controller. But I'll remove it for extra safety then 👍

Sources/StreamChatUI/Common Views/ChatChannelCreateNewButton_Tests.swift
nuno-vieira: This was more to see the change of size of the button. Without the container view it is not possible to understand that the button has successfully changed size. Should I remove it anyways?

Sources/StreamChatUI/Common Views/ChatChannelCreateNewButton_Tests.swift
nuno-vieira: Isn't the same thing? I think what we are doing in the avatar view is what "size" argument is doing, shouldn't we just use this instead? Since it is already available for us 🤔

Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
```diff
@@ -34,7 +34,7 @@ open class _ChatChannelListVC<ExtraData: ExtraDataTypes>: ViewController,
         return collection
     }()
     
-    public private(set) lazy var createNewChannelButton: ChatChannelCreateNewButton<ExtraData> = {
+    public private(set) lazy var createNewChannelButton: UIButton = {
         let button = uiConfig.channelList.newChannelButton.init()
         button.translatesAutoresizingMaskIntoConstraints = false
         button.addTarget(self, action: #selector(didTapCreateNewChannel), for: .touchUpInside)
```
nuno-vieira, line 40: @VojtaStavik Because now uiConfig only knows that is an UIButton, here we need to set that it is a UIButton. If the client is customizing the button of channelListVC, or wants to populate extra data with it, isn't this an issue?

Tests/StreamChatUITests/SnapshotVariant.swift
```diff
@@ -10,6 +10,10 @@ enum SnapshotVariant: String, Hashable, CaseIterable {
     case large
     case extraExtraExtraLarge
     
+    static var userInterfaceStyleCases: [SnapshotVariant] {
+        [small, large]
+    }
```
nuno-vieira, line 15: @VojtaStavik @DominikBucher12 I still think this is not optimal, because the screenshots will print "small" and "large", so it will be difficult to understand which one is dark, and which one is light unless we override the `rawValue: String` of the `SnapshotVariant`. I still think the `OptionSet` is a bit more flexible, and probably a better solution for our needs. WDYT?

DominikBucher12: I vote for the `OptionSet` to have this clear and flexible. Also I like rebasing my pull requests :D

Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
VojtaStavik: Nope, we don't assign any data to if be the default. This is exactly how it should be 👍

Tests/StreamChatUITests/SnapshotVariant.swift
VojtaStavik: -> Slack

Sources/StreamChatUI/Common Views/ChatChannelCreateNewButton.swift
```diff
@@ -11,10 +11,11 @@ public typealias ChatChannelCreateNewButton = _ChatChannelCreateNewButton<NoExtr
 /// A Button subclass that should be used for creating new channels.
 open class _ChatChannelCreateNewButton<ExtraData: ExtraDataTypes>: Button, UIConfigProvider {
     override public func defaultAppearance() {
+        super.defaultAppearance()
         defaultIntrinsicContentSize = .init(width: 44, height: 44)
         setImage(uiConfig.images.newChat, for: .normal)
```
VojtaStavik, line 16: There's no reason why the button should have this intrinsic content size, let's remove it completely.

Sources/StreamChatUI/Common Views/ChatChannelCreateNewButton_Tests.swift
```diff
@@ -0,0 +1,66 @@
...
+
+    func test_customizationUsingAppearanceHook() {
+        class TestView: ChatChannelCreateNewButton {}
+        TestView.defaultAppearance {
+            // Modify appearance
+            $0.layer.borderWidth = 1
+            $0.layer.borderColor = UIColor.green.cgColor
+            $0.backgroundColor = .black
+            $0.tintColor = .lightGray
+        }
```
VojtaStavik, line 36: it'd be nice to add a couple of constraints here to modify the size of the button, too

Sources/StreamChatUI/Common Views/ChatChannelCreateNewButton_Tests.swift
```diff
@@ -0,0 +1,66 @@
...
+    func test_customizationUsingSubclassingHook() {
+        class TestView: ChatChannelCreateNewButton {
+            override func defaultAppearance() {
+                super.defaultAppearance()
+                layer.borderWidth = 1
+                layer.borderColor = UIColor.green.cgColor
+                backgroundColor = .black
+                tintColor = .lightGray
+                setImage(uiConfig.images.close, for: .normal)
+            }
```
VojtaStavik, line 51: same here, please modify the layout here, too

VojtaStavik (approved):
Looking good, just some minor comments.

Sources/StreamChatUI/Common Views/ChatChannelCreateNewButton.swift
nuno-vieira: Keep in mind this will change the ChanneListVC snapshot tests, because this pushes the button of the navigation bar to the left a little. That is why I had put before a width of 60, so it would have the same style as before. But I don't know why we wouldn't change the default layout of the button in the Nav bar 🤔

nuno-vieira: ✅ Done

Sources/StreamChatUI/Common Views/ChatChannelCreateNewButton_Tests.swift
nuno-vieira: ✅ Done

Sources/StreamChatUI/Common Views/ChatChannelCreateNewButton_Tests.swift
nuno-vieira: ✅

PR #817, opened 2021-02-15
[CIS-658] Introduce TraitCollectionWatcher

ManWithBear:
System calls `traitCollectionDidChange` in parent-first way, but our SDK relies on parent be able to override child properties.
To achieve it we wait for system call and on next runloop iteration travel whole view & view controller tree and perform appearance reset.

Known drawback of such approach:
- if parent's `resetAppearance` call `updateContent` on child, it performed twice (first by child's `resetAppearance`, then by parent)

### DIFF
--- a/Sources/StreamChatUI/Common Views/BaseViews.swift
+++ b/Sources/StreamChatUI/Common Views/BaseViews.swift
@@ -22,6 +22,17 @@ public protocol Customizable {
     /// Main point of updating views with the latest data.
     /// Calling super is recommended in most of the cases.
     func updateContent()
+
+    /// It's common for appearance to depend on system settings. When system settings are changed,
+    /// component needs to reset its appearance.
+    func resetAppearance()
+}
+
+public extension Customizable {
+    func resetAppearance() {
+        setUpAppearance()
+        updateContent()
+    }
 }
 
 public extension Customizable where Self: UIView {
@@ -57,32 +68,14 @@ open class View: UIView, AppearanceSetting, Customizable {
     }
     
     public func defaultAppearance() { /* default empty implementation */ }
-    open func setUp() { /* default empty implementation */ }
+    open func setUp() { TraitCollectionWatcher.activate() }
     open func setUpAppearance() { /* default empty implementation */ }
     open func setUpLayout() { /* default empty implementation */ }
     open func updateContent() { /* default empty implementation */ }
-
-    override open func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
-        super.traitCollectionDidChange(previousTraitCollection)
-
-        guard #available(iOS 12, *) else { return }
-        guard previousTraitCollection?.userInterfaceStyle != traitCollection.userInterfaceStyle else { return }
-
-        // Execute previous pending updates
-        pendingUpdates?()
-
-        // Add this view's updates to pending updates
-        pendingUpdates = {
-            (self as! Self).applyDefaultAppearance()
-            self.setUpAppearance()
-            self.updateContent()
-        }
-
-        // Make sure all pending updates are executed eventually
-        DispatchQueue.main.async {
-            pendingUpdates?()
-            pendingUpdates = nil
-        }
+    open func resetAppearance() {
+        (self as! Self).applyDefaultAppearance()
+        setUpAppearance()
+        updateContent()
     }
 }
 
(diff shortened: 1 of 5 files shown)


Sources/StreamChatUI/Common Views/BaseViews.swift
```diff
@@ -274,46 +195,69 @@ open class ViewController: UIViewController, AppearanceSetting, Customizable {
...
+        }
+    }
 
-        guard #available(iOS 12, *) else { return }
-        guard previousTraitCollection?.userInterfaceStyle != traitCollection.userInterfaceStyle else { return }
+    private static var activeTap: Wiretap!
 
-        // Execute previous pending updates
-        pendingUpdates?()
+    static func activate() {
```
DominikBucher12, line 226: Could you document please how this function works, so it's clear from the outside?

Sources/StreamChatUI/Common Views/BaseViews.swift
```diff
@@ -274,46 +195,69 @@ open class ViewController: UIViewController, AppearanceSetting, Customizable {
...
+            DispatchQueue.main.async {
+                TraitCollectionWatcher.dfsAppearanceReset()
+            }
         }
 
-        // Make sure all pending updates are executed eventually
-        DispatchQueue.main.async {
-            pendingUpdates?()
-            pendingUpdates = nil
+        UIApplication.shared.keyWindow?.addSubview(activeTap)
```
DominikBucher12, line 237: So... this view will last forever until application is shut down. I have few questions :D 

1. What are the drawbacks of this? 
2. Doesn't it break something?

Sources/StreamChatUI/Common Views/BaseViews.swift
```diff
@@ -274,46 +195,69 @@ open class ViewController: UIViewController, AppearanceSetting, Customizable {
...
+
+            guard #available(iOS 12, *) else { return }
+            guard previousTraitCollection?.userInterfaceStyle != traitCollection.userInterfaceStyle else { return }
+            onTraitChange()
+        }
+    }
 
-        guard #available(iOS 12, *) else { return }
-        guard previousTraitCollection?.userInterfaceStyle != traitCollection.userInterfaceStyle else { return }
+    private static var activeTap: Wiretap!
```
VojtaStavik, line 224: Maybe it will be better for this to be optional? If the key window changes, the mechanism will stop working but this reference will still keep the instance alive. If this would be `Wiretap?`, once the window is destroyed, the view will be recreated again.

Sources/StreamChatUI/Common Views/BaseViews.swift
```diff
@@ -274,46 +195,69 @@ open class ViewController: UIViewController, AppearanceSetting, Customizable {
...
+        updateContent()
+    }
+}
 
-    override open func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
-        super.traitCollectionDidChange(previousTraitCollection)
+/// Observes trait collection changes and calls `Customizable.resetAppearance` in child-first DFS way
+/// on view & view controller hierarchies.
+private enum TraitCollectionWatcher {
+    private class Wiretap: View {
```
VojtaStavik, line 212: Not a fan of this naming tbh. I think our future selves would benefit from more clear naming like `TraitCollectionChangesObservingView` 🤷‍♂️

DominikBucher12 (approved):
For me this looks good. However please add documentation and make the UI Tests working. I need this ASAP

Sources/StreamChatUI/Common Views/BaseViews.swift
```diff
@@ -274,46 +195,69 @@ open class ViewController: UIViewController, AppearanceSetting, Customizable {
...
-        pendingUpdates?()
+    static func activate() {
+        guard activeTap == nil else { return }
 
-        // Add this view's updates to pending updates
-        pendingUpdates = {
-            (self as! Self).applyDefaultAppearance()
-            self.setUpAppearance()
-            self.updateContent()
+        activeTap = Wiretap(frame: CGRect(x: -10, y: -10, width: 1, height: 1))
```
VojtaStavik, line 229: Looks like there was some thinking behind this, would be nice to capture it in a comment for future generations.

VojtaStavik (approved):
Looks good 👍 Main comments are about naming and missing documentation.

Sources/StreamChatUI/Common Views/BaseViews.swift
ManWithBear: Unless someone heavily rely on special subviews order in active window (why would you?), it shouldn't matter.

ManWithBear:
This solution do not work well with simulated trait changes for UITests.
We decide to stick with current solution for now

PR #818, opened 2021-02-15
[CIS-658] Introduce trait collection reload stack

ManWithBear:
System calls `traitCollectionDidChange` in parent-first way, but our SDK relies on parent be able to override child properties.
To achieve it we wait store appearance reload closures into stack and on next runloop iteration call them in FILO order.

Known drawback of such approach:
- if parent's appearance reload calls `updateContent` on child, it will be performed twice (first by child's appearance reload, then by parent)


VojtaStavik (approved):
LGTM if it actually works :)

DominikBucher12 (approved):
nice

DominikBucher12 (approved):
Awesome

Sources/StreamChatUI/AppearanceSetting.swift
```diff
@@ -58,6 +58,25 @@ private class _AppearanceStorage {
...
+    ///   1. setUp()
+    ///   2. setUpLayout()
+    ///   3. defaultAppearance()
+    ///   4. 👉 <custom rules>
+    ///   5. setUpAppearance()
+    ///   6. updateContent()
+    /// ```
+    ///
+    /// - Important: The closure can be executed multiple times for the same instance or not executed at all. All changes
+    /// done in the closure should be idenpotent.
```
nuno-vieira, line 75:
```suggestion
    /// done in the closure should be idempotent.
```

nuno-vieira (approved):
Small typo, other then that LGTM 👍

PR #819, opened 2021-02-15
[CIS-660] Audit ChatMessageComposerMentionCellView

DominikBucher12:
- [X]  The implementation is cleaned up, unused code is removed, everything follows the "open by default" rule.
 - [X]  The view uses lifecycle methods correctly
 - [X]  The empty state is tested
 - [X]  Default appearance with data is tested + all other possible states of the view (snapshot and/or unit tests)
 - [X]  `UIConfig` customization is tested
 - [X]  Customization using `defaultAppearace { }` is tested
 - [X]  Customization using subclassing is tested
 - [X]  The naming of the view and methods follow the convention
 - [X]  All public-facing API is documented, including used ****`UIConfig` ****components
 - [X] UIConfig uses the most generic super-type possible. For example, don't make the type ChatChannelCreateNewButton<ExtraData>.Type when it can be just UIButton.Type. This is easily applicable for all views that don't use generic content.  ->> **No way in here to put there `AvatarView` thanks to `content` tuple.**
 - [X] Support dynamic fontSizes and resizing of the view with any font Size.


Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerMentionCollectionViewCell.swift
```diff
@@ -5,61 +5,74 @@
 import StreamChat
 import UIKit
 
+/// A View that is embed inside `UICollectionViewCell`  which shows information about user which we want to tag in suggestions
 public typealias ChatMessageComposerMentionCellView = _ChatMessageComposerMentionCellView<NoExtraData>
 
+/// A View that is embed inside `UICollectionViewCell`  which shows information about user which we want to tag in suggestions
 open class _ChatMessageComposerMentionCellView<ExtraData: ExtraDataTypes>: View, UIConfigProvider {
     // MARK: Properties
```
VojtaStavik, line 13: nit: Let's remove noise like this, especially for such small classes

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerMentionCollectionViewCell.swift
```diff
@@ -5,61 +5,74 @@
...
+    /// This is tuple which contains:
+    /// - title: By default name of user
+    /// - subtitle: By default tag of user
+    /// - imageURL: URL of image for the user
+    /// - isUserOnline: Indicates if user is online to set onlineIndicator visible
     open var content: (title: String, subtitle: String, imageURL: URL?, isUserOnline: Bool)? {
         didSet {
             updateContentIfNeeded()
         }
     }
```
VojtaStavik, line 25: One beautiful day when we will have time, this will become:
```
@Content title: String?
@Content subtitle: String?
@Content imageURL: URL?
@Content isUserOnline: Bool
```

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerMentionCollectionViewCell.swift
```diff
@@ -5,61 +5,74 @@
...
 
+    /// `_ChatChannelAvatarView` instance which holds photo of user for tagging.
     open private(set) lazy var avatarView = uiConfig
```
VojtaStavik, line 28: Let's agree on specifying the types explicitly, it will make the lives of us and our users much easier.

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerMentionCollectionViewCell.swift
```diff
@@ -80,43 +93,43 @@ open class _ChatMessageComposerMentionCellView<ExtraData: ExtraDataTypes>: View,
...
 
+/// `UICollectionView` subclass which embeds inside `ChatMessageComposerMentionCellView`
 open class _ChatMessageComposerMentionCollectionViewCell<ExtraData: ExtraDataTypes>: CollectionViewCell, UIConfigProvider {
     // MARK: Properties
 
+    /// Reuse identifier for the cell used in `collectionView(cellForItem:)`
     open class var reuseId: String { String(describing: self) }
 
+    /// Instance of `ChatMessageComposerMentionCellView` which shows information about the mentioned user.
     public private(set) lazy var mentionView = uiConfig
```
VojtaStavik, line 118: explicit type here please

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerMentionCollectionViewCell.swift
```diff
@@ -80,43 +93,43 @@ open class _ChatMessageComposerMentionCellView<ExtraData: ExtraDataTypes>: View,
...
         .messageComposer
         .suggestionsMentionCellView.init()
         .withoutAutoresizingMaskConstraints
 
+    // MARK: - Lifecycle
+
```
VojtaStavik, line 124: ~// MARK: - Lifecycle~

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerMentionCellView_Tests.swift
```diff
@@ -0,0 +1,236 @@
...
+//
+
+import StreamChat
+import StreamChatTestTools
+@testable import StreamChatUI
+import XCTest
+
+class ChatMessageComposerMentionCellView_Tests: XCTestCase {
+    /// Default reference width for the cell on current tested device
+    private static var defaultCellWidth = UIScreen.main.bounds.width - 16
```
VojtaStavik, line 12: The goal for the tests is to be isolated and reproducible. This is the exact opposite of this. Just replace it with a constant and it will be ok 👍

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerMentionCellView_Tests.swift
```diff
@@ -0,0 +1,236 @@
...
+        AssertSnapshot(view, variants: [.small])
+    }
+
+    func test_appearanceCustomization_usingUIConfig_withUserOnline() {
+        class RectIndicator: UIView {
+            override func didMoveToSuperview() {
+                super.didMoveToSuperview()
+                backgroundColor = .green
+                widthAnchor.constraint(equalTo: heightAnchor, multiplier: 1).isActive = true
+            }
```
VojtaStavik, line 47: why not put this to `init`?

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerMentionCellView_Tests.swift
```diff
@@ -0,0 +1,236 @@
...
+        }
+
+        class CustomAvatarView: ChatMessageComposerMentionAvatarView {
+            override func didMoveToSuperview() {
+                super.didMoveToSuperview()
+                backgroundColor = .green
+                widthAnchor.constraint(equalTo: heightAnchor, multiplier: 1).isActive = true
+            }
```
VojtaStavik, line 55: if only we had a dedicated set of methods you should override to achieve certain customization ... oh wait!

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerMentionCellView_Tests.swift
```diff
@@ -0,0 +1,236 @@
...
+    func test_appearanceCustomization_usingAppearanceHook_whenUserOffline() {
+        class TestView: ChatMessageComposerMentionCellView {}
+        TestView.defaultAppearance {
+            // Modify appearance
+            $0.backgroundColor = .white
+            $0.usernameLabel.textColor = .blue
+            $0.usernameTagLabel.textColor = .cyan
+            $0.textStackView.distribution = .equalSpacing
+
+            $0.removeConstraints($0.textStackView.constraints)
```
VojtaStavik, line 108: Why? This shouldn't be needed. That's why we use default low constraint priority.

VojtaStavik (changes requested):
Looking really good 👍 Let's just work a bit on the constraints and their priorities. It should be easier to modify the layout. Let's discuss in Slack

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerMentionCollectionViewCell.swift
DominikBucher12: Done. Thanks

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerMentionCollectionViewCell.swift
DominikBucher12: Done.

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerMentionCollectionViewCell.swift
DominikBucher12: Done. :(

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerMentionCellView_Tests.swift
DominikBucher12: Done :)

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerMentionCellView_Tests.swift
DominikBucher12: Resolved in slack, didn't add init because of unnecessary pollution with required init, this is straightforward.

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerMentionCellView_Tests.swift
DominikBucher12: Thanks but resolved in Slack channel :trollface:

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerMentionCellView_Tests.swift
DominikBucher12: Yeah you are right. I was overthinking this a bit and ended up with not removing these...

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerMentionCollectionViewCell.swift
DominikBucher12: Deleted.

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerMentionCellView_Tests.swift
```diff
@@ -0,0 +1,234 @@
...
+//
+
+import StreamChat
+import StreamChatTestTools
+@testable import StreamChatUI
+import XCTest
+
+class ChatMessageComposerMentionCellView_Tests: XCTestCase {
+    /// Default reference width for the cell.
+    private static var defaultCellWidth: CGFloat = 374
```
VojtaStavik, line 12: nit: You know this could be any artificial number? :D Why 374? Why not 300? 400? 450? 😅

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerMentionCellView_Tests.swift
```diff
@@ -0,0 +1,234 @@
...
+            ])
+        }
+
+        let view = TestView().withoutAutoresizingMaskConstraints
+        view.widthAnchor.constraint(equalToConstant: Self.defaultCellWidth).isActive = true
+        view.content = ("Mr Always LightVader", "darkforce37", TestImages.vader.url, true)
+        AssertSnapshot(view, variants: [.small])
+    }
+
+    func test_appearanceCustomization_usingSubclassing_whenUserOnline() {
```
VojtaStavik, line 149: What if we had just `test_appearanceCustomization_usingSubclassing` and then just make two snapshots inside with different `isUserOnline` value? We already have this pattern in `test_defaultAppearance_withDirectMessageChannel`, check it out.

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerMentionCellView_Tests.swift
```diff
@@ -0,0 +1,234 @@
...
+        class TestView: ChatMessageComposerMentionCellView {
+            override func setUpAppearance() {
+                usernameLabel.textColor = .black
+                usernameTagLabel.textColor = .darkGray
+                backgroundColor = .white
+            }
```
VojtaStavik, line 155: You should call `super.setUpAppearance()`

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerMentionCellView_Tests.swift
```diff
@@ -0,0 +1,234 @@
...
+
+            override func setUpLayout() {
+                addSubview(mentionSymbolImageView)
+                addSubview(textStackView)
+                addSubview(avatarView)
```
VojtaStavik, line 160: You should call `super` here, too. We doesn't really test how flexible the layout is when we completely replace it 🤷‍♂️

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerMentionCollectionViewCell.swift
VojtaStavik: It's not

VojtaStavik (changes requested):
The tests for subclasses shouldn't completely replace the layout but only adjust it.

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerMentionCollectionViewCell.swift
DominikBucher12: Now it is.

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerMentionCellView_Tests.swift
DominikBucher12: It's the same value without needing of re-recording snapshots :trollface: Ended up re-recording snapshot anyways but this kept around :D

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerMentionCellView_Tests.swift
DominikBucher12: Thank you, I changed this so we have less tests but with suffixes

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerMentionCellView_Tests.swift
DominikBucher12: Added.

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerMentionCellView_Tests.swift
DominikBucher12: Fixed.

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerMentionCellView_Tests.swift
```diff
@@ -0,0 +1,136 @@
...
+// Copyright © 2021 Stream.io Inc. All rights reserved.
+//
+
+import StreamChat
+import StreamChatTestTools
+@testable import StreamChatUI
+import XCTest
+
+class ChatMessageComposerMentionCellView_Tests: XCTestCase {
+    /// Default reference width for the cell. This references
```
VojtaStavik, line 11: looks like the other part of the comment is missing.

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerMentionCellView_Tests.swift
VojtaStavik: Please change into a better number. This looks oddly specific and raises questions.

VojtaStavik (commented):
Looking good. I just realized, this view contains text so we should also cover its behavior with different content sizes.

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerMentionCellView_Tests.swift
```diff
@@ -0,0 +1,136 @@
...
+
+    func test_defaultAppearance() {
+        let view = ChatMessageComposerMentionCellView().withoutAutoresizingMaskConstraints
+        view.widthAnchor.constraint(equalToConstant: Self.defaultCellWidth).isActive = true
+
+        view.content = ("Mr Vader", "darkforce37", TestImages.vader.url, true)
+        AssertSnapshot(view, variants: [.small], suffix: "online indicator visible")
+
+        view.content = ("Mr Vader", "darkforce37", TestImages.vader.url, false)
+        AssertSnapshot(view, variants: [.small])
```
nuno-vieira, line 29: With the new refactor you should now use `.onlyUserInterfaceStyle`. And don't forget to re-record the screenshots to update the names too 😛

DominikBucher12: Thanks <3 Works flawlessly, now done :)

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerMentionCellView_Tests.swift
DominikBucher12: Thanks, actuall added constant for height as well.. This is done.

DominikBucher12:
Thanks @VojtaStavik for mentioning the FontSizes. I have redone the constraints on this view a bit and now this view supports flawlessly dynamic fontSizes and resizing of the view according to it. Should be very fine now, check the snapshots. 🥳

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerMentionCellView_Tests.swift
```diff
@@ -0,0 +1,134 @@
...
+    /// To simulate cell behavior with autoresizing height according to content
+    /// we create this height value as well.
+    private static var defaultCellHeight: CGFloat = 50
+
+    func test_emptyAppearance() {
+        let view = ChatMessageComposerMentionCellView().withoutAutoresizingMaskConstraints
+        view.widthAnchor.constraint(equalToConstant: Self.defaultCellWidth).isActive = true
+        view.heightAnchor.constraint(greaterThanOrEqualToConstant: Self.defaultCellHeight).isActive = true
+
+        AssertSnapshot(view, variants: .onlyUserInterfaceStyles)
```
VojtaStavik, line 22: This should be just `AssertSnapshot(view)`, you want to test different content sizes, not just light/dark mode.

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerMentionCellView_Tests.swift
```diff
@@ -0,0 +1,134 @@
...
+    func test_appearanceCustomization_usingAppearanceHook() {
+        class TestView: ChatMessageComposerMentionCellView {}
+        TestView.defaultAppearance {
+            // Modify appearance
+            $0.backgroundColor = .white
+            $0.usernameLabel.textColor = .blue
+            $0.usernameTagLabel.textColor = .cyan
+            $0.textStackView.distribution = .equalSpacing
+            $0.usernameTagLabel.font = .systemFont(ofSize: 24)
+            $0.usernameLabel.font = .systemFont(ofSize: 36)
```
VojtaStavik, line 79: That's not how we want to test dynamic font sizes 🙃 -> Slack

VojtaStavik (changes requested):
Are you sure the snapshots are correct? There are "dark" snapshots with white background and other strange things...

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerMentionCellView_Tests.swift
```diff
@@ -0,0 +1,168 @@
...
+            isOnline: true
+        )
+    }
+
+    func test_emptyAppearance() {
+        let view = ChatMessageComposerMentionCellView().withoutAutoresizingMaskConstraints
+        view.widthAnchor.constraint(equalToConstant: Self.defaultCellWidth).isActive = true
+        view.heightAnchor.constraint(greaterThanOrEqualToConstant: Self.defaultCellHeight).isActive = true
+
+        AssertSnapshot(view)
```
nuno-vieira, line 49: I still think we should be using the size argument of the `AssertSnapshot(size:)`. We are basically repeating what the framework is already doing inside. Also, defining the constraints is more extensive than just passing a CGSize. WDYT @VojtaStavik ? Either way, I think we should try to reuse this since this will be very repetitive, but then again, I think this is already the aim of the `size` property. But if for some reason we still want to add our own constraints, it should happen inside the `AssertSnapshot` so we can reuse this.

VojtaStavik: -> slack

DominikBucher12: This is fixed with https://github.com/GetStream/stream-chat-swift/pull/819/commits/768c6431fd102be9480ee6cea809c5e353a55c01 after our Slack conversation)

VojtaStavik (approved):
LGTM 👍

Sources/StreamChatUI/Utils/UILabel+Extensions.swift
```diff
@@ -0,0 +1,12 @@
...
+// Copyright © 2021 Stream.io Inc. All rights reserved.
+//
+
+import UIKit
+
+extension UILabel {
+    var adjustingFontForContentSizeCategory: Self {
+        adjustsFontForContentSizeCategory = true
+        return self
+    }
```
nuno-vieira, line 11:
```suggestion
    var withAdjustingFontForContentSizeCategory: Self {
        adjustsFontForContentSizeCategory = true
        return self
    }
```
Maybe this follows the convention better?

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerMentionCellView_Tests.swift
```diff
@@ -0,0 +1,160 @@
...
+            override func didMoveToSuperview() {
+                super.didMoveToSuperview()
+                backgroundColor = .green
+                widthAnchor.constraint(equalTo: heightAnchor, multiplier: 1).isActive = true
+            }
+        }
+
+        var config = UIConfig()
+        config.channelList.channelListItemSubviews.onlineIndicator = RectIndicator.self
+        config.messageComposer.mentionAvatarView = CustomAvatarView.self
```
nuno-vieira, line 81: If this is related to the message composer why are we using channeList views? 🤔  Maybe the onlineIndicator should be in a more generic place?

DominikBucher12: Hmm actually very good comment. However please create ticket for this, I figured out that the structure of `UIConfig` is object to change 😅  I mean with this change we would probably need to change the structure a bit, because it's quite different from `ChannelList`...

Sources/StreamChatUI/Utils/UILabel+Extensions.swift
DominikBucher12: Fixed. thanks :)

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerMentionCellView_Tests.swift
```diff
@@ -0,0 +1,160 @@
...
+
+        view.uiConfig = config
+        view.content = chatUserOnline
+        AssertSnapshot(view, suffix: "online indicator visible")
+
+        view.content = chatUserOffline
+        AssertSnapshot(view)
+
+        view.content = chatUserNoName
+        AssertSnapshot(view, suffix: "user name not set")
```
nuno-vieira, line 94: Does it make sense to test here all variants? 🤔 We already test them in the defaultAppearance. Here we are only interested in testing if the customization works,right?

DominikBucher12: Actually we are interested how customization work in various scenarios and with different traits. Now that we finally made it with `contentSize` and Fonts working in tests I think this is quite okay.

PR #820, opened 2021-02-15
[CIS-665] Refactor SnapshotVariant

nuno-vieira:
# In this PR
Refactors the `SnapshotVariant` so it has more flexibility and also it generates more explicit snapshot names since it includes all relevant traits in the name.

We now have a `SnapshotVariant` and a `SnapshotTrait`. A `SnapshotVariant` includes a collection of `SnapshotTrait`'s. 
The `SnapshotVariant` is a combination of traits that a snapshot image is testing.  The `SnapshotTrait` should not be used directly, and only to build variants.

**Explanation for some decisions:**
- `.small` -> `.smallAndDark`. A `SnapshotVariant` has a collection of traits. Making the name `smallAndDark` is explicit of which traits it has. Also, having just a `.dark` variant is not very useful since this only works iOS 12+, which means that for some tests, we would need to do conditional logic. This way, if it supports, it will use the dark and light traits, if it doesn't, it simply uses the content size trait.
- Introduction of `SnapshotTrait`. This is useful to have a name and a raw trait combination. This simplifies the naming of each snapshot, instead of manually naming each Variant, it is done automatically. Each Variant has the name of each Trait that it has. Only DisplayScale trait is not accounted for, since it is present in every Variant (Injected when creating the Variant).


Sources/StreamChatUI/ChatChannelList/ChatChannelAvatarView_Tests.swift
```diff
@@ -24,14 +24,14 @@ class ChatChannelAvatarView_Tests: XCTestCase {
     func test_emptyAppearance() {
         let view = ChatChannelAvatarView().withoutAutoresizingMaskConstraints
         view.addSizeConstraints()
-        AssertSnapshot(view, variants: [.small])
+        AssertSnapshot(view, variants: [.smallAndDark])
```
VojtaStavik, line 27: This doesn't seem to work well. Here we have a view that's content-size agnostic and we want to test only different user interface style values. So, from the API point of view, this should look something like:
```
AssertSnapshot(view, variants: .allUserInterfaceStyles)

... or

AssertSnapshot(view, variants: .onlyUserInterfaceStyles)

... or we can even flip the logic around and specify the snapshots you don't want:

AssertSnapshot(view, variants: .withoutContentSizeChanges)
```

I'd assume it will create 2 snapshots (just 1 on iOS <12) with the default content size.

For views where the content size matters, the API could look like just:
```
AssertSnapshot(view) 
```
which means "snapshot the full matrix of options".

VojtaStavik (changes requested):
Let's work on the API ergonomics a bit more.

Sources/StreamChatUI/ChatChannelList/ChatChannelAvatarView_Tests.swift
nuno-vieira: Actually, in the case of ChatAvatarView, we are testing an UIImage, so the dark and light traits are not very useful (I guess? 🤔) Unless we have border colours etc, in that case testing the default `.allUserInterfaceStyles` is actually useful. That is why didn't use the default arguments here. So maybe we don't need here to specify the variant at all and just use the default which tests all interface styles. 

If not, we can actually create variants without the dark and light trait, and just name them "small" and "large" or "default".

But currently `AssertSnapshot` for viewControllers use `.all` by default. When testing views, `AssertSnapshot` uses `.allUserInterfaceStyles`. Do you want me, in this case, to not provide any custom argument and just use the `.allUserInterfaceStyles` ?

VojtaStavik: > Actually, in the case of ChatAvatarView, we are testing an UIImage, so the dark and light traits are not very useful (I guess? 🤔) 

Correct, there's nothing that should react to interface style changes in `ChatAvatarView` but, for example, `ChatChannelAvatarView` has the online indicator which background changes based on the style.

On the other hand, I think it makes sense to test dark/light variant for all views. It will help to prevent regressions when we introduced different appearances by accident.

I don't think it's a good idea to introduce different default values for view and view controllers. It will be confusing. So, what if we did it like this:
```
AssertSnapshot(view) 
// 👆 by default creates the full matrix:
// small + dark
// large + light
// xxl + dark

AssertSnapshot(view, variants: . onlyUserInterfaceStyles) 
//  👆creates only two snaphots (light + dark) with default content size
```

nuno-vieira: @VojtaStavik Changes done ✅

Tests/StreamChatUITests/SnapshotVariant.swift
```diff
@@ -5,56 +5,89 @@
...
+
+extension SnapshotVariant {
+    // MARK: - Combinations
+    
+    static let all: [SnapshotVariant] = {
+        [smallAndDark, largeAndLight, extraExtraExtraLargeAndLight]
+    }()
+    
+    static let onlyUserInterfaceStyles: [SnapshotVariant] = {
+        [smallAndDark, largeAndLight]
```
VojtaStavik, line 34: This is not correct. This should be `largeAndLight` and `largeAndDark` -> only user inteface style should change, not the size

VojtaStavik (changes requested):

Tests/StreamChatUITests/SnapshotVariant.swift
nuno-vieira: Changes done ✅

VojtaStavik (approved):
Cool 🎉 😎 :shipit:

PR #822, opened 2021-02-16
Fix failing devices test

VojtaStavik:
The implementation was incorrect and the test wasn't structured properly enough to capture it.

### DIFF
--- a/Sources/StreamChat/Controllers/CurrentUserController/CurrentUserController.swift
+++ b/Sources/StreamChat/Controllers/CurrentUserController/CurrentUserController.swift
@@ -113,7 +113,7 @@ public class _CurrentChatUserController<ExtraData: ExtraDataTypes>: DataControll
             return
         }
         
-        currentUserUpdater.updateDevices(currentUserId: currentUserId) { error in
+        currentUserUpdater.fetchDevices(currentUserId: currentUserId) { error in
             self.state = error == nil ? .remoteDataFetched : .remoteDataFetchFailed(ClientError(with: error))
             self.callback { completion?(error) }
         }
@@ -212,7 +212,7 @@ public extension _CurrentChatUserController {
     /// `setUser` must be called before calling this.
     /// - Parameters:
     ///   - id: Device id to be removed. You can obtain registered devices via `currentUser.devices`.
-    ///   If `currentUser.devices` is not up-to-date, please make an `updateDevices` call.
+    ///   If `currentUser.devices` is not up-to-date, please make an `synchronize` call.
     ///   - completion: Called when device is successfully deregistered, or with error.
     func removeDevice(id: String, completion: ((Error?) -> Void)? = nil) {
         guard let currentUserId = currentUser?.id else {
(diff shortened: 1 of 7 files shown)


Sources/StreamChat/Workers/CurrentUserUpdater.swift
```diff
@@ -116,7 +116,9 @@ class CurrentUserUpdater<ExtraData: ExtraDataTypes>: Worker {
             do {
                 let devicesPayload = try result.get()
                 self?.database.write({ (session) in
-                    try session.saveCurrentUserDevices(devicesPayload.devices)
+                    // Since this call always return all device, we want' to clear the existing ones
+                    // to remove the deleted devices.
+                    try session.saveCurrentUserDevices(devicesPayload.devices, clearExisting: true)
```
nuno-vieira, line 121: I've actually removed this on propose, I've tested the call and it returns an empty array if there are no updates. So, this will delete all devices locally.

Sources/StreamChat/Workers/CurrentUserUpdater_Tests.swift
```diff
@@ -494,31 +496,41 @@ final class CurrentUserUpdater_Tests: StressTestCase {
...
         // Previous devices should not be cleared
         AssertAsync {
-            Assert.willBeEqual(currentUser?.devices.count, 2)
-            Assert.willBeEqual(currentUser?.devices.last?.id, dummyDevices.devices.last?.id)
+            Assert.willBeEqual(
+                currentUser?.devices.map(\.id).sorted(),
+                dummyDevices.devices.map(\.id).sorted()
+            )
+            Assert.willBeTrue(callbackCalled)
         }
```
nuno-vieira, line 534: This test actually meant to verify that if we had previous devices already added, they were not deleted, so I stored 1 device first, and then called updateDevices, and it should not clear them. Because the request of update devices was returning empty array. So if we have stored devices, this will clear all devices.

nuno-vieira (approved):
LGTM

Sources/StreamChat/Controllers/CurrentUserController/CurrentUserController.swift
```diff
@@ -212,7 +212,7 @@ public extension _CurrentChatUserController {
     /// `setUser` must be called before calling this.
     /// - Parameters:
     ///   - id: Device id to be removed. You can obtain registered devices via `currentUser.devices`.
-    ///   If `currentUser.devices` is not up-to-date, please make an `updateDevices` call.
+    ///   If `currentUser.devices` is not up-to-date, please make an `fetchDevices` call.
```
VojtaStavik, line 215: This seems to be outdated, shouldn't it rather be `synchronize`?

Sources/StreamChat/Workers/CurrentUserUpdater.swift
```diff
@@ -86,7 +86,7 @@ class CurrentUserUpdater<ExtraData: ExtraDataTypes>: Worker {
     /// - Parameters:
     ///   - id: Device id to be removed. You can obtain registered devices via `currentUser.devices`.
     ///   - currentUserId: The current user identifier.
-    ///   If `currentUser.devices` is not up-to-date, please make an `updateDevices` call.
+    ///   If `currentUser.devices` is not up-to-date, please make an `fetchDevices` call.
```
VojtaStavik, line 89: same here

VojtaStavik (commented):
@nuno-vieira Thank you!

Sources/StreamChat/Controllers/CurrentUserController/CurrentUserController.swift
VojtaStavik: ping @nuno-vieira

nuno-vieira: @VojtaStavik Fixed it

PR #823, opened 2021-02-17
[CIS-607] Rename folders and files with spaces and add swiftlint rule

DominikBucher12:
# What this PR do:
- Renames all folders, groups and files in project with spaces in their name and replaces with camelCase.
# How to test it
- Just take a look at the PR. And at the Swiftlint rule, if you run `mint run swiftlint` and grep out `file_name_no_space` you shouldn't see any error. 
# Where you can start
-  Taking a look at `.swiftlint.yml`. And if I have missed any directory via some magic bash command :)  

This has been issue for me for some time because working with paths with spaces in them really hurts. Thankfully there is also another Swiftlint rule for directories with spaces on it's way at https://github.com/realm/SwiftLint/pull/3523 so when it's merged, we can update Swiftlint and add this too. :) 

btw Please don't judge me guys, I got really mad while working with files with spaces in my PRs so I decided to do this while everyone sleeps :D


VojtaStavik (approved):
👍 LGTM, thanks for doing this 🎉

nuno-vieira (approved):
LGTM

VojtaStavik:
:shipit: @DominikBucher12

PR #826, opened 2021-02-18
[CIS-667] Audit ChatMessageComposerCommandViewCell

nuno-vieira:
# In this PR

Every `StreamUI` SDK view component should have the following in order to be marked as "audited" and therefore production-ready:

- [x]  The empty state is tested
- [x]  `UIConfig` uses the most generic super-type possible. For example, don't use the type `ChatChannelCreateNewButton<ExtraData>.Type` when it can be just `UIButton.Type`. This is easily applicable to all views that don't use generic `content`.
- [x]  Don't use `UIImage` where `URL` is a better fit. For images where we expect that the content could be easily customized by users and/or loaded remotely, use `URL` instead of `UIImage` directly.
- [x]  If the view is not content-size specific, use `AssertSnapshot(view, variants: .onlyUserInterfaceStyles)` for snapshots. This will generate two snapshots of the view (light + dark UI). 
If the view is content-size specific, use the default `AssertSnapshot(view)` variant. This will generate the full matrix of snapshots testing the view for different combinations of `userInterfaceStyle` and `contentSizeCategory`.
- [x]  Remove `// MARK` comments


Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerCommandCollectionViewCell_Tests.swift
```diff
@@ -0,0 +1,26 @@
...
+    func test_defaultAppearance_withNonDMChannel() {
+        // TODO: https://stream-io.atlassian.net/browse/CIS-652
+    }
+
+    func test_appearanceCustomization_usingUIConfig() {}
+
+    func test_appearanceCustomization_usingAppearanceHook() {}
+
+    func test_appearanceCustomization_usingSubclassing() {}
+}
```
DominikBucher12: This might be nit / unnecessary comment so I am probably sorry :D But this actually isn't part of Removing dead code, it's adding it :trollface:

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerCommandCollectionViewCell_Tests.swift
```diff
@@ -0,0 +1,26 @@
...
+//
+
+@testable import StreamChatUI
+import XCTest
+
+class ChatMessageComposerCommandCellView_Tests: XCTestCase {
+    func test_defaultAppearance() {
+        let view = ChatMessageComposerCommandCellView()
+        view.command = .init(name: "Hello", description: "what", set: "is this", args: "WHAT")
+        AssertSnapshot(view, size: CGSize(width: 300, height: 40), record: true)
```
VojtaStavik: It's important for cells to test they work correctly with dynamic sizing, so IMO it makes more sense to just add one fixed constraint (typically a width constraint) and then let the other dimension dynamic.

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerCommandCollectionViewCell_Tests.swift
```diff
@@ -0,0 +1,26 @@
...
+// Copyright © 2021 Stream.io Inc. All rights reserved.
+//
+
+@testable import StreamChatUI
+import XCTest
+
+class ChatMessageComposerCommandCellView_Tests: XCTestCase {
+    func test_defaultAppearance() {
+        let view = ChatMessageComposerCommandCellView()
+        view.command = .init(name: "Hello", description: "what", set: "is this", args: "WHAT")
```
VojtaStavik: This is a strict NO-GO. None of these words is related to StarWars.

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerCommandCollectionViewCell_Tests.swift
```diff
@@ -2,25 +2,94 @@
...
+        AssertSnapshot(cell, size: defaultCellSize)
     }
 
-    func test_defaultAppearance_withDirectMessageChannel() {}
+    func test_defaultAppearance_allAvailableCommands() {
+        let cell = ChatMessageComposerCommandCollectionViewCell()
+        cell.uiConfig.images.commandIcons.forEach { command in
+            cell.commandView.content = .init(name: command.key, description: "", set: "", args: "[text]")
+            AssertSnapshot(cell, variants: [.defaultLight], size: defaultCellSize, suffix: command.key)
+        }
```
VojtaStavik, line 28: What exactly is this test testing?

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerCommandCollectionViewCell.swift
```diff
@@ -39,11 +39,10 @@ open class _ChatMessageComposerCommandCellView<ExtraData: ExtraDataTypes>: View,
...
         commandNameLabel.font = uiConfig.font.bodyBold
+        commandNameLabel.textColor = uiConfig.colorPalette.text
 
         commandNameSubtitleLabel.font = uiConfig.font.body
         commandNameSubtitleLabel.textColor = uiConfig.colorPalette.subtitleText
-
-        commandNameLabel.textColor = uiConfig.colorPalette.text
     }
 
     override open func setUpLayout() {
```
DominikBucher12, line 48: nit: can be oneliner aka `setUpLayout() {}`

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerCommandCollectionViewCell_Tests.swift
```diff
@@ -2,25 +2,94 @@
...
-    func test_appearanceCustomization_usingUIConfig() {}
+    func test_appearanceCustomization_usingAppearanceHook() {
+        class CustomView: ChatMessageComposerCommandCollectionViewCell {}
+        CustomView.defaultAppearance {
+            $0.commandView.backgroundColor = .systemGray
+            $0.commandView.commandNameSubtitleLabel.textColor = UIColor.white
+            $0.commandView.commandNameLabel.textColor = UIColor.darkGray
+        }
+        
+        let view = CustomView()
```
VojtaStavik, line 66: We use `TestView` in all other tests :)

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerCommandCollectionViewCell_Tests.swift
```diff
@@ -2,25 +2,94 @@
...
+                commandNameSubtitleLabel
+                    .trailingAnchor
+                    .constraint(equalTo: layoutMarginsGuide.trailingAnchor)
+                    .isActive = true
+            }
+        }
+        
+        let cell = ChatMessageComposerCommandCollectionViewCell()
+        cell.uiConfig.messageComposer.suggestionsCommandCellView = CustomView.self
+        cell.commandView.content = defaultCommand
```
VojtaStavik, line 53: You're modifying the default `UIConfig` here. A safer way of doing this is what we do in other tests like this:
```swift
var config = UIConfig()
config.messageComposer.suggestionsCommandCellView = CustomView.self
cell.uiConfig = config
```

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerCommandCollectionViewCell_Tests.swift
```diff
@@ -2,25 +2,94 @@
...
+        class CustomView: ChatMessageComposerCommandCollectionViewCell {}
+        CustomView.defaultAppearance {
+            $0.commandView.backgroundColor = .systemGray
+            $0.commandView.commandNameSubtitleLabel.textColor = UIColor.white
+            $0.commandView.commandNameLabel.textColor = UIColor.darkGray
+        }
+        
+        let view = CustomView()
+        view.commandView.content = defaultCommand
+        AssertSnapshot(view, variants: [.defaultLight], size: defaultCellSize)
```
VojtaStavik, line 68: Why only `defaultLight`? What if subclassing breaks dynamic fonts? 😱

VojtaStavik (changes requested):
Looking very good 👍 I have some minor comments regarding naming and tests.

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerCommandCollectionViewCell.swift
```diff
@@ -39,11 +39,10 @@ open class _ChatMessageComposerCommandCellView<ExtraData: ExtraDataTypes>: View,
         backgroundColor = .clear
```
DominikBucher12, line 39: `backgroundColor = uiConfig.colorPalette.popoverBackground` Please set the background color to this, to be synched with `ChatMessageComposerMentionCollectionViewCell`

DominikBucher12 (changes requested):

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerCommandCollectionViewCell.swift
```diff
@@ -45,21 +54,20 @@ open class _ChatMessageComposerCommandCellView<ExtraData: ExtraDataTypes>: View,
...
     }
 
-    // MARK: Private
-
     private func setupLeftImageViewConstraints() {
         commandImageView.leadingAnchor.pin(equalTo: layoutMarginsGuide.leadingAnchor).isActive = true
         commandImageView.topAnchor.pin(equalTo: layoutMarginsGuide.topAnchor).isActive = true
         commandImageView.bottomAnchor.pin(equalTo: layoutMarginsGuide.bottomAnchor).isActive = true
         commandImageView.widthAnchor.pin(equalTo: commandImageView.heightAnchor).isActive = true
+        commandImageView.centerYAnchor.pin(equalTo: layoutMarginsGuide.centerYAnchor).isActive = true
```
DominikBucher12, line 70: I am just curious why it's `layoutMarginsGuide.centerYAnchor` and not just `centerYAnchor` :)

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerCommandCollectionViewCell_Tests.swift
nuno-vieira: Oh nooo... probably a bad rebase 😅

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerCommandCollectionViewCell_Tests.swift
nuno-vieira: Yes this commit had old code, forgot to edit it 👍  I've pushed the changed, re-review the commit please

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerCommandCollectionViewCell_Tests.swift
nuno-vieira: In this case, the Suggestions View Controller, which I'm testing right now too, the command cell's have actually a fixed size 🤔 In this case in particular, I think the idea of the design is that if the command is too long, it should just truncate the text. It would look kinda weird that the size of the cell, in this case, would grow, that is why I ignored the self-sizing, WDYT ?

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerCommandCollectionViewCell_Tests.swift
nuno-vieira: Yeah Forgot to edit this commit 🤦 Fixed it

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerCommandCollectionViewCell_Tests.swift
nuno-vieira: This is testing all the commands we support. Imagine, if we edit the commands in the future, and for a "Key" the image is corrupted or something, this test will catch that. This is making sure that all commands have a corresponding image

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerCommandCollectionViewCell_Tests.swift
nuno-vieira: Yes good point, will rename it, to be consistent 👍

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerCommandCollectionViewCell_Tests.swift
nuno-vieira: I was trying to reduce the test duplication since all variants are already tested in previous tests. How would subclassing break the dynamic fonts? 🤔 Unless we overrode the UILabel and removed the adjust property, but that would be a client's issue. The way I see it is that subclassing and appearance hook tests should be focusing on testing if the customization works. If dynamic fonts or light/dark modes are compromised is probably something that the client is doing wrong? Or maybe I'm missing something? 🤔

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerCommandCollectionViewCell.swift
nuno-vieira: To be honest, I don't know 😅 It is probably the same thing? just noticed I used the layoutMarginsGuide, probably copy paste problem 😄

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerCommandCollectionViewCell_Tests.swift
DominikBucher12: IMHO those snapshot tests should catch only layout stuff and how we are doing with handling dynamic fonts and accessibility. I get what you are trying to solve with the corresponding image corrupted/different, but I think it's unnecessary and crosses the scope of Snapshot tests with something else.

I mean you have really good point, but this should be tested somewhere else / somehow else.  But this is interesting topic actually 🤔

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerCommandCollectionViewCell.swift
nuno-vieira: Is this really needed? The background color will be the one from the suggestions container. Why should we force the cell to have a bg color?

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerCommandCollectionViewCell_Tests.swift
nuno-vieira: I think you can use snapshot tests to test whatever you want. That is why we should also test state differences. You can even test other things, not only images. I don't think we should limit ourselves to test only layout and accessibility issues with snapshots. We can test also business logic. And in this case actually, by testing all commands we support, we are checking that each command does not change the layout in someway 🤔

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerCommandCollectionViewCell.swift
nuno-vieira: This is not empty btw 😄 You probably got misled by the Github's review 😛

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerCommandCollectionViewCell_Tests.swift
nuno-vieira: ✅ Resolved

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerCommandCollectionViewCell_Tests.swift
nuno-vieira: ✅ Resolved

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerCommandCollectionViewCell_Tests.swift
VojtaStavik: All right, I'm probably too paranoid 👍 Leave it as is

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerCommandCollectionViewCell.swift
DominikBucher12: ahh okay it doesn't matter, thanks :)

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerCommandCollectionViewCell.swift
DominikBucher12: Hmm it's good question, I mean I wrote mainly because of consistency and for the snapshots to look ok while taking a look at them. Maybe it's not that important.

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerCommandCollectionViewCell.swift
DominikBucher12: When I am looking at the method name I feel ashamed :D :D :D Thanks, this shouldn't really be empty in any case :D :D

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerCommandCollectionViewCell_Tests.swift
nuno-vieira: Resolved, removed the test ✅

DominikBucher12 (approved):
For me looks good :)

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerCommandCollectionViewCell_Tests.swift
nuno-vieira: New test added ✅  @VojtaStavik

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerCommandCollectionViewCell_Tests.swift
nuno-vieira: @VojtaStavik Fixed ✅

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerCommandCollectionViewCell_Tests.swift
```diff
@@ -0,0 +1,113 @@
...
+        config.images.messageComposerCommandFallback = TestImages.vader.image
+        
+        let view = ChatMessageComposerCommandCellView()
+        view.uiConfig = config
+        view.content = .init(name: "R2", description: "", set: "", args: "[text]")
+        
+        view.updateContent()
+        
+        XCTAssertEqual(view.commandImageView.image, config.images.messageComposerCommandFallback)
+    }
```
VojtaStavik, line 47: nit: You can combine these tests and just re-assing the content.

VojtaStavik (approved):
Looking good 👍 🚢 🇮🇹

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerCommandCollectionViewCell_Tests.swift
nuno-vieira: Ups, I only saw this now, merging 😅 Only saw the "ship it" in my email. I'll refactor it in my current branch 👍

VojtaStavik: Haha, ok 👍

PR #832, opened 2021-02-19
Improve base view naming + fix CIS-509

VojtaStavik:
We have a collision between `StreamChatUI.View` and `SwiftUI.View`. We don't want to break the existing SwiftUI code just by importing `StreamChatUI`.  This PR adds `_` prefix to all base views we use.

Adding `_` works nicely with the existing `_XXX` types because the same "don't use directly" rule applies to the base types, too.

---

I also fixed CIS-509 (formatting of generated files), since it took me just a minute and it was extremely annoying :)

### DIFF
--- a/Sources/StreamChatUI/Generated/L10n.swift
+++ b/Sources/StreamChatUI/Generated/L10n.swift
@@ -1,7 +1,3 @@
-//
-// Copyright © 2021 Stream.io Inc. All rights reserved.
-//
-
 // swiftlint:disable all
 // Generated using SwiftGen — https://github.com/SwiftGen/SwiftGen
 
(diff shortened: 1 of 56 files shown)


Sources/StreamChatUI/Generated/L10n.swift
```diff
@@ -14,139 +10,131 @@ import Foundation
 // swiftlint:disable explicit_type_interface function_parameter_count identifier_name line_length
 // swiftlint:disable nesting type_body_length type_name vertical_whitespace_opening_braces
 internal enum L10n {
```
DominikBucher12, line 12: This re-generated by accident or not? Otherwise I don't understand this :(

VojtaStavik: Nope, this is the version of the file without the formatter. It was committed formatted originally.

VojtaStavik: From now on you should see changes in this file only if you actually change something there.

DominikBucher12: Great! Thank you❗️

DominikBucher12 (approved):
For me this is awesome <3

b-onc (approved):
:shipit: 
And especially thankful for the L10n.swift file

PR #833, opened 2021-02-19
[CIS-669] Audit ChatMessageComposerSuggestionsViewController

nuno-vieira:
# In this PR

Every `StreamUI` SDK view component should have the following in order to be marked as "audited" and therefore production-ready:

- [x]  The empty state is tested
- [x]  `UIConfig` uses the most generic super-type possible. For example, don't use the type `ChatChannelCreateNewButton<ExtraData>.Type` when it can be just `UIButton.Type`. This is easily applicable for all views that don't use generic `content`.
- [x]  Don't use `UIImage` where `URL` is a better fit. For images where we expect that the content could be easily customized by users and/or loaded remotely, use `URL` instead of `UIImage` directly.
- [x]  If the view is not content-size specific, use `AssertSnapshot(view, variants: .onlyUserInterfaceStyles)` for snapshots. This will generate two snapshots of the view (light + dark UI). 
If the view is content-size specific, use the default `AssertSnapshot(view)` variant. This will generate the full matrix of snapshots testing the view for different combinations of `userInterfaceStyle` and `contentSizeCategory` .
- [x]  Remove `// MARK` comments


Sources/StreamChatTestTools/Controllers/ChatUserSearchController_Mock.swift
```diff
@@ -9,4 +9,9 @@ public class ChatUserSearchController_Mock<ExtraData: ExtraDataTypes>: _ChatUser
     public static func mock() -> ChatUserSearchController_Mock<ExtraData> {
         .init(client: .mock())
     }
+    
+    public var users_mock: [_ChatUser<ExtraData.User>]?
+    public override var users: LazyCachedMapCollection<_ChatUser<ExtraData.User>> {
```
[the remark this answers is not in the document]
DominikBucher12: 🤔

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerSuggestionsViewController.swift
```diff
@@ -38,27 +37,34 @@ open class _ChatMessageComposerSuggestionsViewController<ExtraData: ExtraDataTyp
...
         view.superview != nil
     }
-
-    private var collectionViewHeightObserver: NSKeyValueObservation?
-
-    // MARK: - Subviews
-
+    
+    /// The collection view of the commands.
     open private(set) lazy var collectionView = uiConfig
```
DominikBucher12, line 50: Maybe add here explicit type?

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerSuggestionsViewController.swift
```diff
@@ -38,27 +37,34 @@ open class _ChatMessageComposerSuggestionsViewController<ExtraData: ExtraDataTyp
...
-
+    
+    /// The container view where collectionView is embedded.
     open private(set) lazy var containerView: UIView = UIView().withoutAutoresizingMaskConstraints
 
-    // MARK: - Overrides
-
+    // Height for suggestion cell, this value should never be 0
+    // otherwise it causes loop for height of this controller and as a result this controller height will be 0 as well.
+    // Note: This value can be 1, it's just for purpose of 1 cell being visible.
```
DominikBucher12, line 61: You could add the documentation that this is for purpose of seeing just 1 cell to adjust height and not looping to 0 according to our Slack conversation.

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerSuggestionsViewController.swift
```diff
@@ -38,27 +37,34 @@ open class _ChatMessageComposerSuggestionsViewController<ExtraData: ExtraDataTyp
...
+    private let defaultRowHeight: CGFloat = 44
+    
+    private var frameObserver: NSKeyValueObservation?
+    
+    private var collectionViewHeightObserver: NSKeyValueObservation?
```
DominikBucher12, line 66: Even it's private I would add some little documentation. But I guess it's up to you and it's obvious.

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerSuggestionsCommandsHeaderView.swift
```diff
@@ -5,26 +5,36 @@
...
 }
 
+/// The header view of the suggestion collection view.
 public typealias ChatMessageComposerSuggestionsCommandsHeaderView = _ChatMessageComposerSuggestionsCommandsHeaderView<NoExtraData>
 
+/// The header view of the suggestion collection view.
 open class _ChatMessageComposerSuggestionsCommandsHeaderView<ExtraData: ExtraDataTypes>: _View, UIConfigProvider {
+    
+    /// The image icon of the commands header view.
     public private(set) lazy var commandImageView = UIImageView()
```
DominikBucher12, line 34: Explicit type?

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerSuggestionsCommandsHeaderView.swift
```diff
@@ -5,26 +5,36 @@
...
         .withoutAutoresizingMaskConstraints
-
+    
+    /// The text label of the commands header view.
     public private(set) lazy var headerLabel = UILabel()
```
DominikBucher12, line 38: Missing explicit type

DominikBucher12 (changes requested):

DominikBucher12:
Looks good, just some minor stuff from me + Could you please add tests where we can see Separators of the cells? Would be really cool stuff to test if we can do this easily

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerSuggestionsCommandsHeaderView.swift
nuno-vieira: Do we really need the explicit type here? 🤔 It is already described right after

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerSuggestionsCommandsHeaderView.swift
nuno-vieira: same here? 🤔

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerSuggestionsViewController.swift
nuno-vieira: The document that is already present seems enough? Or not? 🤔

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerSuggestionsViewController.swift
nuno-vieira: I think the documentation above and also inside setupLayout looks enough to me :)

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerSuggestionsViewController.swift
DominikBucher12: Ahh in that case it's okay :D Sorry I was confused from our Slack conversation... This is good actually 👍

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerSuggestionsViewController_Tests.swift
```diff
@@ -0,0 +1,238 @@
...
+    private let defaultMentionSuggestionsSize = CGSize(width: 360, height: 110)
+    
+    // MARK: - Mock Data
+    
+    private let commands: [Command] = [
+        .init(name: "giphy", description: "", set: "", args: "[text]"),
+        .init(name: "ban", description: "", set: "", args: "[@username] [text]"),
+        .init(name: "mute", description: "", set: "", args: "[@username] [text]"),
+        .init(name: "unmute", description: "", set: "", args: "[@username] [text]")
+    ]
```
VojtaStavik, line 25: when using real comments, the test will for example fail when we change the giphy icon. I'd suggest creating two test-only commands and add some test images for them to `UIConfig`. The goal is to make the test as focused as possible.

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerSuggestionsViewController_Tests.swift
```diff
@@ -0,0 +1,238 @@
...
+@testable import StreamChatUI
+import XCTest
+
+class ChatMessageComposerSuggestionsViewController_Tests: XCTestCase {
+    
+    // We need to provide a size to the suggestions view since here we are testing the view in isolation,
+    // and so we can't attach it to a bottomAnchorView. The test to verify the height calculation dependent
+    // on the rows should be done in the parent view controller tests.
+    private let defaultCommandSuggestionsSize = CGSize(width: 360, height: 210)
+    private let defaultMentionSuggestionsSize = CGSize(width: 360, height: 110)
```
VojtaStavik, line 16: Why not use just one size for both? :)

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerSuggestionsCommandsHeaderView.swift
VojtaStavik: Yeah, it might seem redundant, but let's make a habit out of it. We can even have a SwiftLint rule for that :) If you want to remove the duplicate naming you can do ` public private(set) lazy var commandImageView: UIImageView = .init()`

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerSuggestionsCommandsHeaderView.swift
```diff
@@ -5,26 +5,36 @@
...
     class var reuseId: String { String(describing: self) }
-
+    
+    /// The suggestions header view.
     public lazy var suggestionsHeader: _ChatMessageComposerSuggestionsCommandsHeaderView<ExtraData> = {
         let header = uiConfig.messageComposer.suggestionsHeaderView.init().withoutAutoresizingMaskConstraints
         embed(header)
         return header
     }()
 }
```
VojtaStavik, line 25: Can we make all these properties `open`?

VojtaStavik (changes requested):
Looking good 👍 Just some minor suggestions and questions. I'd normally put `Approve` but you like to merge PRs without checking the comments first, so I decided to go with the safest "Request changes" option 😅

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerSuggestionsViewController_Tests.swift
nuno-vieira: One had 4 commands, and the other and 2 mentions 😄 but since I'm going to use new mocked commands now, I can change that to be the same 👍

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerSuggestionsViewController.swift
DominikBucher12: Ok 👍🏼

DominikBucher12 (approved):
LGTM. But what about the separators for cells?

Tests/StreamChatUITests/TestImages/XCTestCase+TestImages.swift
```diff
@@ -16,6 +16,14 @@ extension XCTestCase {
...
         }()
         
+        static let commandGiphy: (url: URL, image: UIImage) = {
+            getImage(withName: "Giphy", fileExtension: "png")
+        }()
+        
+        static let commandFallback: (url: URL, image: UIImage) = {
+            getImage(withName: "Fallback", fileExtension: "png")
+        }()
+
```
VojtaStavik, line 26: It's confusing when you use real images :) It must be clear to everyone these are TEST images. Maybe we need more images than Yoda and Vader ?

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerSuggestionsViewController_Tests.swift
```diff
@@ -0,0 +1,246 @@
...
+        )
+    ]
+    
+    var vc: ChatMessageComposerSuggestionsViewController!
+    
+    override func setUp() {
+        super.setUp()
+        vc = ChatMessageComposerSuggestionsViewController()
+        vc.uiConfig.images.commandIcons["giphy"] = TestImages.commandGiphy.image
+        vc.uiConfig.images.messageComposerCommandFallback = TestImages.commandFallback.image
```
VojtaStavik, line 45: You can't do this. This modifies the default `UIConfig`. You need to create a new `UIConfig` instance and assign it explicitly to the VC. We do it already in all other tests like that.

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerSuggestionsViewController_Tests.swift
```diff
@@ -0,0 +1,246 @@
...
+                super.setUpLayout()
+
+                widthAnchor.constraint(equalTo: heightAnchor, multiplier: 1).isActive = true
+            }
+        }
+        
+        var config = UIConfig()
+        config.messageComposer.suggestionsMentionCellView = TestView.self
+        config.images.commandIcons["giphy"] = TestImages.yoda.image
+        config.images.messageComposerCommandFallback = TestImages.vader.image
```
VojtaStavik, line 189: 👍

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerSuggestionsViewController.swift
```diff
@@ -211,13 +219,28 @@ public typealias ChatMessageComposerSuggestionsMentionDataSource = _ChatMessageC
...
+    /// Data Source Initialiser
+    ///
+    /// - Parameters:
+    ///   - commands: The list of commands.
+    ///   - collectionView: The collection view of the commands.
+    
+    /// Data Source Initialiser
+    /// - Parameters:
+    ///   - collectionView: The collection view of the mentions.
+    ///   - searchController: The search controller to find mentions.
```
VojtaStavik, line 243: probably a copy/paste error?

Sources/StreamChatUI/UIConfig+MessageComposer.swift
```diff
@@ -48,7 +48,7 @@ public extension _UIConfig {
         public var suggestionsCommandCellView: _ChatMessageComposerCommandCellView<ExtraData>.Type =
             _ChatMessageComposerCommandCellView<ExtraData>.self
         /// The collection view layout of the suggestions collection view.
-        public var suggestionsCollectionViewLayout: ChatMessageComposerSuggestionsCollectionViewLayout.Type =
+        public var suggestionsCollectionViewLayout: UICollectionViewLayout.Type =
```
VojtaStavik: 👍

VojtaStavik (approved):

DominikBucher12 (approved):

DominikBucher12:
I will just leave here comment for future generations: 

https://github.com/pointfreeco/swift-snapshot-testing/issues/424 

There is an issue with architectures recording snapshots a bit differently so it causes failures on x86 mac when recorded from  M1 macbooks (ARMv8)

VojtaStavik (approved):

VojtaStavik (changes requested):
-> slack

VojtaStavik (approved):

PR #835, opened 2021-02-22
[CIS-676] Fix crash when creating new channel

nuno-vieira:
# In this PR
Fixes a crash when creating a new channel by Verifying that `previousItems` are not empty when `finalLayoutAttributes` is called. This function can be called randomly before `prepare(forCollectionViewUpdates)` where `previousItems` is populated. Sometimes ChatVC calls `keyboardFrameWillChange` that triggers the `finalLayoutAttributes` function before the  `prepare(forCollectionViewUpdates)`. So we need to safely check first if the `previousItems` are not empty.


Sources/StreamChatUI/ChatMessageList/ChatMessageListCollectionViewLayout.swift
```diff
@@ -269,6 +269,7 @@ open class ChatMessageListCollectionViewLayout: UICollectionViewLayout {
     }
 
     override open func finalLayoutAttributesForDisappearingItem(at itemIndexPath: IndexPath) -> UICollectionViewLayoutAttributes? {
+        guard previousItems.isEmpty else { return nil }
```
VojtaStavik, line 272: Is this the only place where we should guard for this?

VojtaStavik (approved):

PR #845, opened 2021-02-25
Fix incorrect bundle for strings with CocoaPods integration

### LINKED ISSUE #839: Localized strings do not work with Cocoapods for StreamChatUI v3
## What did you do?
* Integrated StreamChatUI (`3.2-beta` though looking at git history I imagine it applies to all versions)

* Strings with their localization to display

## What happened instead?
* String keys are displayed

## GetStream Environment
**GetStream Chat version: 3.2-beta**
**GetStream Chat frameworks:** StreamChatUI
**iOS version: <=12.0**
**Swift version: 5.x**
**Xcode version: 12.2**
**Device: Simulator**

## Additional context

I looked through your code and it seems like you are creating a separate resource bundle than the one referenced in `L10n.swift`. I would submit a PR but it seems like `L10n.swift` is generated and you need to keep using this separate bundle for static frameworks, so I'm not quite sure what the solution here is...

VojtaStavik:
The generated strings file didn't use the correct resources bundle for CocoaPods.

I also added an integration test for CocoaPods which will be run on CI for every PR.

---

Closes #839

### DIFF
--- a/.github/workflows/beta.yml
+++ b/.github/workflows/beta.yml
@@ -19,13 +19,15 @@ jobs:
       with:
         ssh-private-key: ${{ secrets.BOT_SSH_PRIVATE_KEY }}
     - uses: actions/checkout@v1
+    - name: Set build image var
+      run: echo "ImageVersion=$ImageVersion"  >> $GITHUB_ENV
     - name: Cache RubyGems
       uses: actions/cache@v2
       id: rubygem-cache
       with:
         path: vendor/bundle
-        key: ${{ runner.os }}-gem-${{ hashFiles('**/Gemfile.lock') }}
-        restore-keys: ${{ runner.os }}-gem-
+        key: ${{ runner.os }}-${{ env.ImageVersion }}-gem-${{ hashFiles('**/Gemfile.lock') }}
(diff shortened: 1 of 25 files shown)


Sample/Carthage/ChatExample.xcodeproj/project.xcworkspace/xcuserdata/vojta.xcuserdatad/IDEFindNavigatorScopes.plist
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<array/>
+</plist>
```
VojtaStavik, line 5: whoops, sorry for the noise. Will remove it

VojtaStavik:
@b-onc ping

b-onc (approved):
I thought I already approved!
:shipit:

PR #846, opened 2021-02-26
[CIS-679] Removes ContainerStackView from ChatChannelListItemView

nuno-vieira:
# In this PR

- Removes the ContainerStackView from ChatChannelListItemView.
- Changes the fonts to the correct ones from Figma.
- Fixes the fonts not adapting properly for different font sizes.
- Sets properly the height of the cell separators.


Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
```diff
@@ -40,10 +40,13 @@ open class _ChatChannelListItemView<ExtraData: ExtraDataTypes>: _ChatChannelSwip
         .unreadCountView.init()
         .withoutAutoresizingMaskConstraints
 
+    /*
+        TODO: ReadStatusView, Missing LLC API
     /// The view showing indicator for read status of the last message in channel.
     open private(set) lazy var readStatusView: _ChatChannelReadStatusCheckmarkView<ExtraData> = uiConfigSubviews
         .readStatusView.init()
         .withoutAutoresizingMaskConstraints
+     */
```
DominikBucher12, line 53: hmmph I always follow rule if you have need to comment something, you should delete it. Unless it's some gaming in your WIP commit whatever... I think let's leave it there and add documentation that it is not currently supported?

Sources/StreamChatUI/UIConfig+Fonts.swift
```diff
@@ -8,6 +8,7 @@ public extension _UIConfig {
     struct Font {
         public var caption1 = UIFont.preferredFont(forTextStyle: .caption1)
         public var footnoteBold = UIFont.preferredFont(forTextStyle: .footnote).bold
+        public var footnote = UIFont.preferredFont(forTextStyle: .footnote)
```
DominikBucher12, line 11: Could you explain somehow? Some design changes were made?

Sources/StreamChatUI/ChatChannelList/ChatChannelSwipeableListItemView.swift
```diff
@@ -59,7 +59,7 @@ open class _ChatChannelSwipeableListItemView<ExtraData: ExtraDataTypes>: _View,
         trailingConstraint?.isActive = true
         
         addSubview(bottomSeparatorView)
-        bottomSeparatorView.heightAnchor.pin(equalToConstant: 0.4).isActive = true
+        bottomSeparatorView.heightAnchor.pin(equalToConstant: 1 / contentScaleFactor).isActive = true
```
DominikBucher12, line 62: What is `contentScaleFactor`? Would you mind adding documentation there?

Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
nuno-vieira: The reason I did not delete it, is because the ReadStatusView already exists. If I remove the comment, I'm afraid, the person developing this, will not use that view, because it won't know the view already exists. So I think for now this is the best approach I guess

Sources/StreamChatUI/UIConfig+Fonts.swift
nuno-vieira: Yes the figma design now uses footnote for the body of the Last message of the channel list :)

Sources/StreamChatUI/ChatChannelList/ChatChannelSwipeableListItemView.swift
nuno-vieira: The documentation is already provided in the property itself, but I think it is a good idea to comment on what is that doing and why it is need, so I'll grab the commit's description and add it as a comment 👍

nuno-vieira: Added the comment :)

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerCommandCollectionViewCell.swift
```diff
@@ -51,6 +51,7 @@ open class _ChatMessageComposerCommandCellView<ExtraData: ExtraDataTypes>: _View
 
         addSubview(textStackView)
         setupStack()
+        commandNameLabel.setContentCompressionResistancePriority(.streamRequire, for: .horizontal)
```
VojtaStavik, line 54: Better if you set it to `.streamLow` so it doesn't mess with other `streamRequire` constraints.

Sources/StreamChatUI/CommonViews/ChatOnlineIndicatorView.swift
```diff
@@ -24,7 +24,6 @@ open class _ChatOnlineIndicatorView<ExtraData: ExtraDataTypes>: _View, UIConfigP
 
     override open func layoutSubviews() {
         super.layoutSubviews()
-        layer.frame = layer.frame.inset(by: .init(top: -1, left: -1, bottom: -1, right: -1))
```
VojtaStavik, line 27: this was making the indicator smaller with every `layoutSubviews` call? :)

Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
```diff
@@ -40,84 +44,126 @@ open class _ChatChannelListItemView<ExtraData: ExtraDataTypes>: _ChatChannelSwip
...
+        ]
+
+        // Set subtitleLabel compression resistance smaller than the timestamp label
+        subtitleLabel.setContentCompressionResistancePriority(.streamLow, for: .horizontal)
+        timestampLabel.setContentCompressionResistancePriority(.streamRequire, for: .horizontal)
+
+        NSLayoutConstraint.activate(constraintsToActivate)
     }
     
     override open func updateContent() {
```
VojtaStavik, line 169: The "NoChannel" text is strange when the channel is missing. We should show just a black view when the channel is missing. Don't forget to reset all subviews not just the title label, the item view gets reused.

VojtaStavik (approved):
Looks great. I have some suggestions but they are minor. Thanks 🎉

Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
nuno-vieira: @VojtaStavik ✅

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerCommandCollectionViewCell.swift
nuno-vieira: ✅

VojtaStavik (approved):

PR #847, opened 2021-02-26
[CIS-683] Fix crash when rotating screen

VojtaStavik:
The checks were there but they were inverted. cc @nuno-vieira


nuno-vieira (approved):
🤦 LGTM. We really need to add tests to the MessageList 😄

PR #849, opened 2021-03-01
[CIS-690] Expose ChatChanneltemView constraints in the Layout object

fortmarek:
# Submit a pull request

## CLA

- [x] The code changes follow best practices
- [x] Code changes are not tested => this PR is only a refactoring and thus snapshot tests should suffice (LMK if you think there should actually be some tests added)

Changing a location of a view without not calling `super.setUpLayout()` has been so far mildly challenging as there were too many constraints from `super.setUpLayout` influencing it. 

To make such customizations easier, we can introduce `Layout` struct which will contain properties of grouped constraints. Therefore, if a client wants to change a location of eg. `timestampLabel`, he could do `NSLayoutConstraint.deactivate(layout.timestampLabelConstraints)` and then add their own constraints to make the process more straightforward.

### DIFF
--- a/Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
+++ b/Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
@@ -15,6 +15,20 @@ open class _ChatChannelListItemView<ExtraData: ExtraDataTypes>: _ChatChannelSwip
     public var content: (channel: _ChatChannel<ExtraData>?, currentUserId: UserId?) {
         didSet { updateContentIfNeeded() }
     }
+    
+    /// Properties tied to `ChatChannelListItemView` layout
+    public struct Layout {
+        /// Constraints of `timestampLabel`
+        public fileprivate(set) var timestampLabelConstraints: [NSLayoutConstraint] = []
+        /// Constraints of `avatarView`
+        public fileprivate(set) var avatarViewConstraints: [NSLayoutConstraint] = []
+        /// Constraints of `titleLabel`
+        public fileprivate(set) var titleLabelConstraints: [NSLayoutConstraint] = []
+        /// Constraints of `subtitleLabel`
+        public fileprivate(set) var subtitleLabelConstraints: [NSLayoutConstraint] = []
+        /// Constraints of `unreadCountView`
+        public fileprivate(set) var unreadCountViewConstraints: [NSLayoutConstraint] = []
+    }
         
     private lazy var uiConfigSubviews: _UIConfig.ChannelListItemSubviews = uiConfig.channelList.channelListItemSubviews
     
@@ -43,6 +57,9 @@ open class _ChatChannelListItemView<ExtraData: ExtraDataTypes>: _ChatChannelSwip
     open private(set) lazy var unreadCountView: _ChatChannelUnreadCountView<ExtraData> = uiConfigSubviews
         .unreadCountView.init()
         .withoutAutoresizingMaskConstraints
+    
+    /// Layout properties of this view
+    public private(set) var layout = Layout()
 
     /*
         TODO: ReadStatusView, Missing LLC API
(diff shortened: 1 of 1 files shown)


Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
```diff
@@ -92,7 +104,7 @@ open class _ChatChannelListItemView<ExtraData: ExtraDataTypes>: _ChatChannelSwip
         ]
         
         // Avatar view
-        constraintsToActivate += [
+        layout.avatarViewConstraints += [
```
VojtaStavik, line 107: IMO we should `= []` rather than `+= []`

Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
```diff
@@ -163,7 +175,14 @@ open class _ChatChannelListItemView<ExtraData: ExtraDataTypes>: _ChatChannelSwip
         subtitleLabel.setContentCompressionResistancePriority(.streamLow, for: .horizontal)
         timestampLabel.setContentCompressionResistancePriority(.streamRequire, for: .horizontal)
 
-        NSLayoutConstraint.activate(constraintsToActivate)
+        NSLayoutConstraint.activate(
+            constraintsToActivate
```
VojtaStavik, line 179: Since `constraintsToActive` now contains only `visualCenterGuide` constraints, maybe we should rename it to `visualCenterGuideConstriant`?

Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
```diff
@@ -15,6 +15,15 @@ open class _ChatChannelListItemView<ExtraData: ExtraDataTypes>: _ChatChannelSwip
...
         didSet { updateContentIfNeeded() }
     }
+    
+    /// Properties tied to `ChatChannelListItemView` layout
+    public struct Layout {
+        public fileprivate(set) var timestampLabelConstraints: [NSLayoutConstraint] = []
+        public fileprivate(set) var avatarViewConstraints: [NSLayoutConstraint] = []
+        public fileprivate(set) var titleLabelConstraints: [NSLayoutConstraint] = []
+        public fileprivate(set) var subtitleLabelConstraints: [NSLayoutConstraint] = []
+        public fileprivate(set) var unreadCountViewConstraints: [NSLayoutConstraint] = []
```
VojtaStavik, line 25: Even though it seems obvious from the name of the variable, it should still be documented properly.

VojtaStavik (changes requested):
Looks good 👍 Just some minor comments. Also, you don't need to use the CLA PR template as an internal contributor.

Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
fortmarek: good point, missed it

Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
fortmarek: eg. "Constraints of `unreadCountView`" or something different / more detailed?

Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
fortmarek: makes sense, it will be more consistent anyways

Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
VojtaStavik: Yep, something like that. We have this rule internally to prevent discussions if something is worth/needed to be documented. This way you don't have to think about it - is it `public` -> should be documented.

VojtaStavik (approved):
Looks good 👍 Let's merge this and see how it feels

DominikBucher12 (approved):
LGTM 🎉 I was just thinking if there is some way we could enforce the `Layout` struct (or maybe some other form?) to different views via some mechanism...

fortmarek:
> I was just thinking if there is some way we could enforce the Layout struct (or maybe some other form?) to different views via some mechanism...

There should be some way but I'd say let's see first where this will be necessary and if such optimization is desired at all

PR #850, opened 2021-03-01
[CIS-682] Use decoration view for cell separators

nuno-vieira:
# In this PR

Introduces a CellSeperatorView + CellSeparatorCollectionViewLayout to replace the bottomSeperatorView.


Sources/StreamChatUI/UIConfig+ChannelList.swift
```diff
@@ -7,7 +7,7 @@ import UIKit.UICollectionViewLayout
 public extension _UIConfig {
     struct ChannelListUI {
         public var channelCollectionView: ChatChannelListCollectionView.Type = ChatChannelListCollectionView.self
-        public var channelCollectionLayout: UICollectionViewLayout.Type = ChatChannelListCollectionViewLayout.self
+        public var channelCollectionLayout: UICollectionViewLayout.Type = ChatChannelListCollectionViewLayout<ExtraData>.self
```
nuno-vieira, line 10: Not sure if we should put the CellSeperatorView in the UIConfig? 🤔 I guess it would be easier to customize. But should we put it here in the ChannelListUI config, or in a more generic place? I guess we should create a CommonUI Config section ?

PR #852, opened 2021-03-01
[CIS-682] Use decoration view for cell separators

nuno-vieira:
# In this PR

Removes the bottom separator view and replaces it with a reusable decoration view to show the cell separators.


Sources/StreamChatUI/CommonViews/CellSeperatorView/CellSeparatorCollectionViewLayout.swift
```diff
@@ -0,0 +1,67 @@
...
+    open var separatorHeight: CGFloat = 0.5
+
+    /// The reuse identifier of the separator cell view view.
+    open var separatorIdentifier = "SeparatorCell"
+
+    /// Closure handler to decide what indexPaths's should be rendered.
+    /// By default it is rendered for every cell.
+    open var shouldHideSeparator: (IndexPath, UICollectionView) -> Bool = { _, _ in
+        return false
+    }
```
nuno-vieira, line 21: What do you guys think about this? It is the most flexible API have come up with hide the cells that the customer wants. An example to hide the last cell would be:
`return indexPath.row == collectionView.numberOfItems(section: indexPath.section) - 1`

Sources/StreamChatUI/CommonViews/CellSeperatorView/CellSeparatorCollectionViewLayout.swift
```diff
@@ -0,0 +1,66 @@
...
+    /// The height of the cell separator view.
+    open var separatorHeight: CGFloat = 0.5
+
+    /// The reuse identifier of the cell separator view.
+    open var separatorIdentifier = "CellSeparator"
+
+    /// Closure handler to decide what indexPaths's the separator should be hidden.
+    open var shouldHideSeparator: (IndexPath, UICollectionView) -> Bool = { _, _ in
+        return false
+    }
```
nuno-vieira, line 20: What do you guys think about this? It is the most flexible API have come up with to hide the cells that the customer wants. An example to hide the last cell would be:
return indexPath.row == collectionView.numberOfItems(section: indexPath.section) - 1

DominikBucher12: For me simple and cool, actually this looks pretty powerful at first sight. :D

Sources/StreamChatUI/CommonViews/CellSeperatorView/CellSeparatorCollectionViewLayout.swift
```diff
@@ -0,0 +1,66 @@
...
+//
+
+import StreamChat
+import UIKit
+
+/// A reusable collection view layout that supports rendering separator views in the collection cells.
+open class CellSeparatorCollectionViewLayout<ExtraData: ExtraDataTypes>: UICollectionViewFlowLayout {
+
+    /// The height of the cell separator view.
+    open var separatorHeight: CGFloat = 0.5
```
DominikBucher12, line 12: nit: Probably should be `0.33`? 

Snímek obrazovky 2021-03-02 v 0 22 24
 as default separator height is 0.33?

DominikBucher12 (approved):
For me looks great :) Just noticed that there is no separator at the top of channelList but in the same time there is one at the bottom. Is this intentional?

Sources/StreamChatUI/CommonViews/CellSeperatorView/CellSeparatorCollectionViewLayout.swift
VojtaStavik: I think the default actually is 1 / display scale

Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
```diff
@@ -50,10 +50,14 @@ open class _ChatChannelListVC<ExtraData: ExtraDataTypes>: _ViewController,
...
-            uiConfig.channelList.channelCellSeparatorReusableView,
-            forDecorationViewOfKind: collectionViewLayout.separatorIdentifier
-        )
+
+        if let cellSeparatorIdentifier = (collectionViewLayout as? CellSeparatorContainingLayout)?.separatorIdentifier {
+            collectionViewLayout.register(
+                uiConfig.channelList.channelCellSeparatorReusableView,
+                forDecorationViewOfKind: cellSeparatorIdentifier
+            )
+        }
```
nuno-vieira, line 59: Does the `collectionViewLayout` property really need to be `UICollectionViewLayout`? I thought we should use the views of UIKit, only if we didn't need any specific data. 

Another problem with this is that we will be repeating this a lot. There are other views that require the cell separator, and it feels we making this less reusable and less obvious to reuse 🤔

Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
```diff
@@ -77,6 +81,17 @@ open class _ChatChannelListVC<ExtraData: ExtraDataTypes>: _ViewController,
...
+        if let flowLayout = collectionViewLayout as? UICollectionViewFlowLayout {
+            // The default separator height is the hair height (1 physical px)
+            flowLayout.minimumLineSpacing = 1 / traitCollection.displayScale
+
+            flowLayout.itemSize = UICollectionViewFlowLayout.automaticSize
+            flowLayout.estimatedItemSize = .init(
+                width: collectionView.bounds.width,
+                height: 64
+            )
+        }
```
nuno-vieira, line 94: The same here, whenever we need the separator behaviour we will need to cast it and set always the flowLayout properties, repeating this code in multiple places, and also easier to make a mistake and not put a consistent value with the rest of the UI

Sources/StreamChatUI/CommonViews/CellSeperatorView/CellSeparatorCollectionViewLayout.swift
```diff
@@ -5,62 +5,63 @@
...
-            let cellFrame = layoutAttribute.frame
+
+            let cellFrame = cellAttribute.frame
+
             separatorAttribute.frame = CGRect(
                 x: cellFrame.origin.x,
                 y: cellFrame.origin.y + cellFrame.size.height,
                 width: cellFrame.size.width,
-                height: separatorHeight
+                height: minimumLineSpacing // use `minimumLineSpacing` as the separator height
```
nuno-vieira, line 61: I still think this is an implementation detail. It would be better to provide a more explicit API. The customers will need to check this implementation to understand what changes the separator height.

Sources/StreamChatUI/CommonViews/CellSeperatorView/CellSeparatorView.swift
```diff
@@ -5,36 +5,22 @@
...
 import UIKit
 
-/// The cell separator reusable view that acts as container of the content view.
+/// The cell separator reusable view that acts as container of the visible part of the separator view.
 public typealias CellSeparatorReusableView = _CellSeparatorReusableView<NoExtraData>
 
-/// The cell separator reusable view that acts as container of the content view.
-open class _CellSeparatorReusableView<ExtraData: ExtraDataTypes>: UICollectionReusableView, UIConfigProvider {
+/// The cell separator reusable view that acts as container of the visible part of the separator view.
+open class _CellSeparatorReusableView<ExtraData: ExtraDataTypes>: _CollectionReusableView, UIConfigProvider {
```
nuno-vieira, line 12: Nice one 👍 We probably need to do this to the commands suggestion header reusable view, and the other `CollectionReusableView`'s we are using, so every reusable view follows the same pattern

Sources/StreamChatUI/CommonViews/CellSeperatorView/CellSeparatorCollectionViewLayout.swift
```diff
@@ -0,0 +1,67 @@
...
+
+// Default implementation of CellSeparatorContainingLayoutDelegate
+
+public extension CellSeparatorContainingLayoutDelegate {
+    func collectionView(
+        _ collectionView: UICollectionView,
+        layout collectionViewLayout: CellSeparatorContainingLayout & UICollectionViewFlowLayout,
+        shouldShowSeparatorForCellAtIndexPath indexPath: IndexPath
+    ) -> Bool { true }
+}
```
nuno-vieira, line 30: The delegate makes sense, but the other approach would make it possible to change this through UIConfig. But make we don't want that? 🤔

Sources/StreamChatUI/ChatChannelList/ChatChannelListCollectionViewLayout.swift
```diff
@@ -3,47 +3,19 @@
...
     override open func prepare() {
         super.prepare()
-        
+
+        // We should always propose the full width of the collection view for the cell width
         estimatedItemSize = .init(
             width: collectionView?.bounds.width ?? 0,
-            height: 64
+            height: estimatedItemSize.height
         )
```
VojtaStavik, line 18: This can live in `ListCollectionViewLayout`

Sources/StreamChatUI/CommonViews/ListCollectionViewLayout/ListCollectionViewLayout.swift
```diff
@@ -0,0 +1,72 @@
...
+/// An `UICollectionViewFlowLayout` implementation to make the collection view behave as a `UITableView`.
+open class ListCollectionViewLayout: UICollectionViewFlowLayout{
+
+    /// The reuse identifier of the cell separator view.
+    open var separatorIdentifier: String = "CellSeparatorIdentifier"
+
+    /// The height of the cell separator view. By default is the hair height, one physical pixel (1 / displayScale).
+    open var separatorHeight: CGFloat {
+        1 / (collectionView?.traitCollection.displayScale ?? 1)
+    }
```
VojtaStavik, line 27: This can be ` open lazy var separatorHeight: CGFloat = 1 / (collectionView?.traitCollection.displayScale ?? 1)`

Sources/StreamChatUI/CommonViews/ListCollectionViewLayout/ListCollectionViewLayout.swift
```diff
@@ -0,0 +1,72 @@
...
+
+    weak var delegate: ListCollectionViewLayoutDelegate?
```
VojtaStavik, line 29: Let's do it the same way `UICollectionViewDelegateFlowLayout` does it, by accessing `collectionView.delegate`

Sources/StreamChatUI/CommonViews/ListCollectionViewLayout/ListCollectionViewLayout.swift
```diff
@@ -0,0 +1,72 @@
...
+    ) -> Bool
+}
+
+/// An `UICollectionViewFlowLayout` implementation to make the collection view behave as a `UITableView`.
+open class ListCollectionViewLayout: UICollectionViewFlowLayout{
+
+    /// The reuse identifier of the cell separator view.
+    open var separatorIdentifier: String = "CellSeparatorIdentifier"
+
+    /// The height of the cell separator view. By default is the hair height, one physical pixel (1 / displayScale).
```
VojtaStavik, line 24: Please also add a note that this changes `minimumLineSpacing` too

VojtaStavik (changes requested):
The last set of changes and we're there 👍

Sources/StreamChatUI/ChatChannelList/ChatChannelListCollectionViewLayout.swift
nuno-vieira: I guess then we don't actually need `ChatChannelListCollectionViewLayout` anymore right? Will delete it

VojtaStavik: @nuno-vieira True 👍

Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
```diff
@@ -73,6 +81,14 @@ open class _ChatChannelListVC<ExtraData: ExtraDataTypes>: _ViewController,
...
         
         collectionView.backgroundColor = uiConfig.colorPalette.background
+
+        if let flowLayout = collectionViewLayout as? ListCollectionViewLayout {
+            flowLayout.itemSize = UICollectionViewFlowLayout.automaticSize
+            flowLayout.estimatedItemSize = .init(
+                width: collectionView.bounds.width,
+                height: 64
+            )
+        }
```
nuno-vieira, line 91: This is probably not needed anymore ? 🤔

Sources/StreamChatUI/CommonViews/ListCollectionViewLayout/ListCollectionViewLayout.swift
```diff
@@ -0,0 +1,79 @@
+//
+//  Copyright © 2021 Stream.io Inc. All rights reserved.
+//
+
+import StreamChat
+import UIKit
+
+/// The `ListCollectionViewLayout` delegate to control how to display the list.
+public protocol ListCollectionViewLayoutDelegate: class, UICollectionViewDelegate  {
```
VojtaStavik, line 9: Is the `class` here actually needed?

VojtaStavik (approved):
NICE 🎉

PR #854, opened 2021-03-02
[CIS-650] Create flexible mother cell

b-onc:
Simplifies the way we create cells. All cell subviews are instantiated lazily which speeds up the initial setup time.

This is not production ready yet. The implementation has to cleaned-up and properly tested.


VojtaStavik (approved):
OK, let's fix the CI and merge this. @b-onc can you please prepare some JIRA tickets with the next steps? Ideally, we'd clean this up enough for the Friday release.

PR #853, opened 2021-03-01
[CIS-689] Refactor ChatChannelNamer

DominikBucher12:
# What this PR do:
- Makes `ChatChannelNamer` more customizable by changing the class into one global function with same purpose as suggested by @VojtaStavik in our slack Channel
# How to test it
- The tests should cover all cases, you can run the demo application and check if the channel names are the same. 
# Where you can start
- `ChatChannelNamer.swift`

Note: Maybe we could create some different `ChannelNamer` closure to show how the customization is done?

PR #855, opened 2021-03-02
[CIS-688] Audit ChatChannelListVC Components

nuno-vieira:
# In this PR

Final Audit to all ChatChannelListVC's components.


Sources/StreamChatUI/UIConfig+ChannelList.swift
```diff
@@ -5,47 +5,66 @@
...
-        /// A `UICollectionViewCell` subclass that shows channel information.
-        public var channelViewCell: _ChatChannelListCollectionViewCell<ExtraData>.Type =
-            _ChatChannelListCollectionViewCell<ExtraData>.self
+        /// The subviews that compose the `itemView`.
+        public var itemSubviews = ItemSubviews()
 
-        /// A button used for creating new channels.
-        public var newChannelButton: UIButton.Type = _ChatChannelCreateNewButton<ExtraData>.self
+        /// The subviews of that compose the `itemView`.
+        public struct ItemSubviews {
```
nuno-vieira, line 42: I think the subviews is useful to better organize each view and to not overpopulate the config. This will be especially useful for the message list.

Sources/StreamChatUI/ChatChannelList/ChatChannelListVC_Tests.swift
```diff
@@ -46,50 +86,104 @@ class ChatChannelListVC_Tests: XCTestCase {
...
+        }
+
+        class TestView: TestChatChannelListVC {}
+        TestView.defaultAppearance {
+            $0.createNewChannelButton.tintColor = UIColor.orange
+            if let listLayout = $0.collectionViewLayout as? ListCollectionViewLayout {
+                listLayout.separatorHeight = 4
+                listLayout.register(TestSeparatorView.self, forDecorationViewOfKind: listLayout.separatorIdentifier)
+            }
+        }
```
nuno-vieira, line 148: Not sure how I feel about this, it is not very friendly right now to change both the color and separator height, there are two steps needed, which I guess is not expected, I think we should be able to customize both the color and height in the same place. WDYT? 🤔

nuno-vieira: Btw, this is true only for subclassing. When using UIConfig, it is easier.

Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
```diff
@@ -225,4 +221,11 @@ extension _ChatChannelListItemView {
             return "No messages"
         }
     }
+
+    /// The formatted string containing the typing member.
+    var typingMemberString: String? {
+        guard let members = content.channel?.currentlyTypingMembers, !members.isEmpty else { return nil }
+        let names = members.compactMap(\.name).sorted()
+        return names.joined(separator: ", ") + " \(names.count == 1 ? "is" : "are") typing..."
```
VojtaStavik, line 229: This needs to be localized

Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
```diff
@@ -209,12 +209,8 @@ open class _ChatChannelListItemView<ExtraData: ExtraDataTypes>: _ChatChannelSwip
 }
 
 extension _ChatChannelListItemView {
-    var typingMemberString: String? {
-        guard let members = content.channel?.currentlyTypingMembers, !members.isEmpty else { return nil }
-        let names = members.compactMap(\.name).sorted()
-        return names.joined(separator: ", ") + " \(names.count == 1 ? "is" : "are") typing..."
-    }
-    
+
```
VojtaStavik, line 212: the `getFormattedDate` helper in `updateContent` needs to be improved. The formatter object should be cached + the time should be formatted according to the current locale.

Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
```diff
@@ -225,4 +221,11 @@ extension _ChatChannelListItemView {
             return "No messages"
```
VojtaStavik, line 221: a) This is not localized
b) Is this correct according to Figma?

Sources/StreamChatUI/UIConfig+CurrentUser.swift
```diff
@@ -0,0 +1,13 @@
...
+//  Copyright © 2021 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+public extension _UIConfig {
+    struct CurrentUser {
+        public var currentUserViewAvatarView: _CurrentChatUserAvatarView<ExtraData>.Type = _CurrentChatUserAvatarView<ExtraData>
+            .self
+        public var avatarView: ChatAvatarView.Type = ChatAvatarView.self
```
VojtaStavik, line 11: What is the difference between those two?

Sources/StreamChatUI/ChatChannelList/ChatChannelListVC_Tests.swift
```diff
@@ -13,6 +13,47 @@ class ChatChannelListVC_Tests: XCTestCase {
...
+                ChatMessage.mock(id: "1", text: "Hello2", author: .mock(id: "Vader2"))
+            ]
+        )
+        return [
+            channel,
+            channelWithOnlineIndicator,
+            channelWithLongTextAndUnreadCount,
+            channelWithMultipleMessages
+        ]
+    }
```
VojtaStavik, line 56: This is a computed property which means that every time you access it, a new set of channels (some of them with a unique id every time) is created.

I kinda like the idea of having a set of various channels ready to use for tests. Maybe it would make sense to put this code into some kind of `dummy` helper (like we do it for LLC tests) and then just assign the property local property in `setUp()` as we do it everywhere else.

I already commented on a similar this here https://github.com/GetStream/stream-chat-swift/pull/853#discussion_r585358400 . Please get used to the fact that unless you _really want_ the variable to be constant (i.e. snapshot size), it better to set it in `setUp`. It will save you headaches in the future.

Sources/StreamChatUI/ChatChannelList/ChatChannelListVC_Tests.swift
```diff
@@ -46,50 +86,104 @@ class ChatChannelListVC_Tests: XCTestCase {
...
-            cid: .unique,
-            name: "Channel 1",
-            imageURL: TestImages.yoda.url,
-            lastMessageAt: .init(timeIntervalSince1970: 1_611_951_526_000)
+
+    func test_emptyAppearance() {
+        mockedChannelListController.simulate(
+            channels: [],
+            changes: []
         )
```
VojtaStavik, line 94: `simulate` is meant to be used to simulate changes. It seems to me what you're looking for here is `simulateInitial`. The behavior is different, and if used properly, it better simulates the real-world behavior.

Sources/StreamChatUI/ChatChannelList/ChatChannelListVC_Tests.swift
VojtaStavik: I see what you mean @nuno-vieira. You could probably do this if you really needed to have this in one place:
```swift
        TestView.defaultAppearance {
             $0.createNewChannelButton.tintColor = UIColor.orange
             if let listLayout = $0.collectionViewLayout as? ListCollectionViewLayout {
                 listLayout.separatorHeight = 4
                 listLayout.register(TestSeparatorView.self, forDecorationViewOfKind: listLayout.separatorIdentifier)
             }
             TestSeparatorView.defaultAppearance { 
               $0.separatorView.backgroundColor = .orange
             }
         }
```
We don't have access to the separator instances directly AFAIK. The collection view manages them automatically.

If it's ugly/dangerous/confusing you need to decide for yourself 🤷‍♂️

VojtaStavik (commented):
Look very good 👍 I commented on a lot of things so let's talk in Slack and decide what to do next.

Sources/StreamChatUI/UIConfig+ChannelList.swift
```diff
@@ -5,47 +5,66 @@
 import UIKit.UICollectionViewLayout
 
 public extension _UIConfig {
-    struct ChannelListUI {
+    struct ChannelList {
+        /// A button used for creating new channels.
+        public var newChannelButton: UIButton.Type = _ChatChannelCreateNewButton<ExtraData>.self
+
+        /// The logic to generate a name for the given channel.
+        public var channelNamer: ChatChannelNamer.Type = ChatChannelNamer.self
```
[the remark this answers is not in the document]
nuno-vieira: Yes, I talked about this before, we already have some common views, so it makes sense to add a section for the common views in the UIConfig. I was thinking `UIConfig.Common`, WDYT? And there we would put the channelNamer, maybe the CellSeperator as well, the avatars, etc... and the other sections would the views from the common, but could be overridden like you are saying 👍

Sources/StreamChatUI/ChatChannelList/ChatChannelListCollectionViewCell.swift
```diff
@@ -11,24 +11,25 @@ public typealias ChatChannelListCollectionViewCell = _ChatChannelListCollectionV
...
 open class _ChatChannelListCollectionViewCell<ExtraData: ExtraDataTypes>: _CollectionViewCell, UIConfigProvider {
+
     /// The `ChatChannelListItemView` instance used as content view.
-    open private(set) lazy var channelView: _ChatChannelListItemView<ExtraData> = uiConfig.channelList.itemView.init()
+    open private(set) lazy var itemView: _ChatChannelListItemView<ExtraData> = uiConfig.channelList.itemView.init()
 
     override public func prepareForReuse() {
         super.prepareForReuse()
-        channelView.trailingConstraint?.constant = 0
+        itemView.trailingConstraint?.constant = 0
```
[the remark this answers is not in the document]
nuno-vieira: The reason I didn't change this view that much is because @DominikBucher12 AFAIK is working on it right? 🤔

Sources/StreamChatUI/UIConfig+ChannelList.swift
```diff
@@ -5,47 +5,66 @@
...
+            _ChatChannelListCollectionViewCell<ExtraData>.self
 
         /// The Cell Separator View.
-        public var channelCellSeparatorReusableView: UICollectionReusableView.Type = _CellSeparatorReusableView<ExtraData>.self
+        public var cellSeparatorReusableView: UICollectionReusableView.Type = _CellSeparatorReusableView<ExtraData>.self
 
         /// The base view of the channel list item to support swipeable gestures.
-        public var channelListSwipeableItemView: _ChatChannelSwipeableListItemView<ExtraData>.Type =
+        public var swipeableItemView: _ChatChannelSwipeableListItemView<ExtraData>.Type =
             _ChatChannelSwipeableListItemView<ExtraData>.self
```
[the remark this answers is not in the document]
nuno-vieira: Yes this is true 🤔 I'll improve this

Sources/StreamChatUI/UIConfig+CurrentUser.swift
nuno-vieira: I will have to investigate

Sources/StreamChatUI/ChatChannelList/ChatChannelListVC_Tests.swift
nuno-vieira: I don't think the dummy helper here is very beneficial. Because it is not dummy data, and it is useful to see in this place all the data that will be displayed in the snapshot. But the setup of the data has been changed 👍

Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
nuno-vieira: ✅ Done. It is cached now and it is possible to change the format by the customer 👍

VojtaStavik: @nuno-vieira the format shouldn't be hardcoded like this. The time format depends on the device locale. We should use some of the existing time-format presets because they respect the locale of the device.

Sources/StreamChatUI/Utils/DateFormatter+Extensions.swift
```diff
@@ -0,0 +1,15 @@
...
+import Foundation
+
+extension DateFormatter {
+    static func makeDefault() -> DateFormatter {
+        let formatter = DateFormatter()
+        formatter.dateStyle = .none
+        formatter.timeStyle = .short
+        formatter.locale = Locale.autoupdatingCurrent
+        return formatter
+    }
```
VojtaStavik, line 14: nit: I think this should rather be `static let default: DateFormatter`. We only need one instance of the default formatter.

DocsSnippets/01-UX/06-Customizing-Views/02-Changing-Appearance.swift
```diff
@@ -7,7 +7,7 @@ func snippets_ux_customizing_views_changing_appearance() {
     // > import UIKit
     // > import StreamChatUI
 
-    UIConfig.default.channelList.channelListItemView.defaultAppearance.addRule {
+    UIConfig.default.channelList.itemView.defaultAppearance.addRule {
         $0.backgroundColor = UIColor.yellow.withAlphaComponent(0.2)
         $0.titleLabel.textColor = .darkGray
     }
```
VojtaStavik, line 13: Hmmm, this is not really what the snippets should look like. It's confusing. I'll fix it later when working on finalizing the documentation.

VojtaStavik (approved):
Looks good + our discussion in Slack

PR #856, opened 2021-03-03
[CIS-702] Add custom view modifier for SwiftUI usage of UIConfig

IgorRosocha:

This PR adds custom view modifier `setUpStreamChatUIEnvironment()` which is to be called when user desires to setup `UIConfig` environment object for usage in `SwiftUI` app.

### DIFF
--- a/Sources/StreamChatUI/UIConfig+SwiftUI.swift
+++ b/Sources/StreamChatUI/UIConfig+SwiftUI.swift
@@ -1,0 +1,57 @@
+//
+// Copyright © 2021 Stream.io Inc. All rights reserved.
+//
+
+import StreamChat
+import SwiftUI
+
+@available(iOS 13.0, *)
+extension _UIConfig {
+
+    /// Used to initialize `_UIConfig` as `ObservableObject`.
+    public var asObservableObject: ObservableObject { .init(self) }
+
+    @dynamicMemberLookup
+    /// `_UIConfig` represented as `ObservableObject` class for SwiftUI requirements.
+    public class ObservableObject: SwiftUI.ObservableObject {
+
+        private let wrappedConfig: _UIConfig<ExtraData>
+
+        public subscript<T>(dynamicMember keyPath: KeyPath<_UIConfig<ExtraData>, T>) -> T {
+            wrappedConfig[keyPath: keyPath]
+        }
+
+        fileprivate init(_ wrappedConfig: _UIConfig<ExtraData>) {
+            self.wrappedConfig = wrappedConfig
+        }
+    }
+}
+
+@available(iOS 13.0, *)
+/// Modifier for setting `UIConfig` environment object.
+private struct SwiftUIEnvironment<ExtraData: ExtraDataTypes>: ViewModifier {
+    /// Custom `ObservableObject` of `UIConfig`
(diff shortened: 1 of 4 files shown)


Sources/StreamChatUI/UIConfig+SwiftUI.swift
```diff
@@ -0,0 +1,31 @@
+//
+// Copyright © 2021 Stream.io Inc. All rights reserved.
+//
+
+import SwiftUI
+
+@available(iOS 13.0, *)
+@dynamicMemberLookup
+/// `UIConfig` represented as `ObservableObject` class for SwiftUI requirements.
+public class SwiftUIConfig: ObservableObject {
```
VojtaStavik, line 10: `UIConfig` is generic over extra data. Let's follow the same naming pattern we use elsewhere:
```swift
public typealias SwiftUIConfig = _SwiftUIConfig<NoExtraData>
public class _SwiftUIConfig<ExtraData: ExtraDataTypes> {}
```

Maybe instead of a top-level type, we could namespace the wrapper?
```swift
extension UIConfig {
    class ObservableObject: SwiftUI.ObservableObject {}
}
```

Sources/StreamChatUI/UIConfig+SwiftUI.swift
```diff
@@ -0,0 +1,31 @@
...
+        content.environmentObject(SwiftUIConfig())
+    }
+}
+
+@available(iOS 13.0, *)
+public extension View {
+    /// Sets up UI environment.
+    func setUpStreamChatUIEnvironment() -> some View {
+        self.modifier(SwiftUIEnvironment())
+    }
```
VojtaStavik, line 30: Which `SwiftUIConfig` is used needs to be customizable, this always uses `_UIConfig<NopExtraData>.default` which should ideally be the default value.

Maybe we should rename this to `setStreamChatUIConfig()`?

VojtaStavik (commented):
Yep, looks like a good start. Is it possible to test this somehow? For example, that when you use custom UIConfig it uses that one properly.

Sources/StreamChatUI/UIConfig+SwiftUI.swift
IgorRosocha: Adjusted in d41dac45bd4ebf0c747a410ded539c4210bba194.

Sources/StreamChatUI/UIConfig+SwiftUI.swift
IgorRosocha: Adjusted in 20235c54ffa29f90506859e1123c29b03957f826.

IgorRosocha:
> Yep, looks like a good start. Is it possible to test this somehow? For example, that when you use custom UIConfig it uses that one properly.

Actually, I'm not quite sure how to properly test it. Do you have some thoughts/ideas how to approach this? 🙂

VojtaStavik:
@IgorRosocha I tried something. It's not the most beautiful code I've ever written but it works :)

IgorRosocha:
> @IgorRosocha I tried something. It's not the most beautiful code I've ever written but it works :)

Maybe not the most beautiful code, but definitely exciting. Great job 👍

VojtaStavik (approved):

PR #859, opened 2021-03-04
[CIS-691] Add support for cooldown/throttling

b-onc:
Not doing:
- Not introducing check for sending a new message

### DIFF
--- a/Sources/StreamChat/APIClient/Endpoints/Payloads/ChannelListPayload.swift
+++ b/Sources/StreamChat/APIClient/Endpoints/Payloads/ChannelListPayload.swift
@@ -94,6 +94,10 @@ struct ChannelDetailPayload<ExtraData: ExtraDataTypes>: Decodable {
     /// Refer to [docs](https://getstream.io/chat/docs/multi_tenant_chat/?language=swift) for more info.
     public let team: String
     
+    /// Cooldown duration for the channel, if it's in slow mode.
+    /// This value will be 0 if the channel is not in slow mode.
+    let cooldownDuration: Int
+    
     init(from decoder: Decoder) throws {
         let container = try decoder.container(keyedBy: ChannelCodingKeys.self)
         typeRawValue = try container.decode(String.self, forKey: .typeRawValue)
@@ -115,6 +119,8 @@ struct ChannelDetailPayload<ExtraData: ExtraDataTypes>: Decodable {
         
         members = try container.decodeIfPresent([MemberPayload<ExtraData.User>].self, forKey: .members)
         
+        cooldownDuration = try container.decodeIfPresent(Int.self, forKey: .cooldownDuration) ?? 0
+        
         extraData = try ExtraData.Channel(from: decoder)
     }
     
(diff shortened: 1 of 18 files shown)


Sources/StreamChat/APIClient/Endpoints/Payloads/ChannelListPayload.swift
```diff
@@ -94,6 +94,10 @@ struct ChannelDetailPayload<ExtraData: ExtraDataTypes>: Decodable {
     /// Refer to [docs](https://getstream.io/chat/docs/multi_tenant_chat/?language=swift) for more info.
     public let team: String
     
+    /// Cooldown duration for the channel, if it's in slow mode.
+    /// This value will be 0 if the channel is not in slow mode.
+    let cooldownDuration: Int
+
```
VojtaStavik, line 100: nit: I get this value is in whole seconds but still - maybe it's better to use `TimeInterval` for this?

Sources/StreamChat/Models/Channel.swift
```diff
@@ -100,6 +100,12 @@ public struct _ChatChannel<ExtraData: ExtraDataTypes> {
     ///
     public let reads: [_ChatChannelRead<ExtraData>]
     
+    /// Cooldown duration for the channel, if it's in slow mode.
+    /// This value will be 0 if the channel is not in slow mode.
+    /// This value is in seconds.
+    /// For more information, please check [documentation](https://getstream.io/chat/docs/javascript/slow_mode/?language=swift).
+    public let cooldownDuration: Int
+
```
VojtaStavik, line 108: Same here - the `Int` here hurts my eyes. IMO it should be `TimeInterval`

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
```diff
@@ -781,6 +781,54 @@ public extension _ChatChannelController {
...
+        guard let cid = cid, isChannelAlreadyCreated else {
+            channelModificationFailed(completion)
+            return
+        }
+        guard cooldownDuration >= 1 && cooldownDuration <= 120 else {
+            callback {
+                completion?(ClientError.InvalidCooldownDuration())
+            }
+            return
+        }
```
VojtaStavik, line 806: Maybe this should rather be an assert? In debug this will fail fast, in release the backend will handle the error.

VojtaStavik (approved):
TOP NOTCH WORK 💯

Sources/StreamChat/APIClient/Endpoints/Payloads/ChannelListPayload.swift
b-onc: backend only accepts seconds, if we use `TimeInterval` users who skip documentation will assume it's ms...

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
b-onc: harder to write tests though? works for me

b-onc (commented):
I'm against `TimeInterval` vs `Int` since backend purposefully restricts it...

PR #863, opened 2021-03-04
[CIS-699] Make separators supplementary views

fortmarek:

- fixed a bug where separators would disappear when the height of the cell would be different from estimated size
- make separators supplementary views instead of decorators

Possible Qs:
- I was not completely sure how to handle the kind and reuse identifier. As reuse identifier is quite contained, I had hard coded it in `ChatChannelListVC`. The kind identifier is then reused via a property defined in `ListCollectionViewLayout`. LMK if you'd have done it differently

### DIFF
--- a/Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
+++ b/Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
@@ -43,20 +43,28 @@ open class _ChatChannelListVC<ExtraData: ExtraDataTypes>: _ViewController,
         .currentUserViewAvatarView.init()
         .withoutAutoresizingMaskConstraints
     
+    /// Reuse identifier of separator
+    open var separatorReuseIdentifier: String { "CellSeparatorIdentifier" }
+    
+    /// Reuse identiifer of `collectionViewCell`
+    open var collectionViewCellReuseIdentifier: String { "Cell" }
+    
     override open func setUp() {
         super.setUp()
         
         controller.setDelegate(self)
         controller.synchronize()
         
-        collectionView.register(uiConfig.channelList.collectionViewCell.self, forCellWithReuseIdentifier: "Cell")
-
-        if let cellSeparatorIdentifier = (collectionViewLayout as? ListCollectionViewLayout)?.separatorIdentifier {
-            collectionViewLayout.register(
-                uiConfig.channelList.cellSeparatorReusableView,
-                forDecorationViewOfKind: cellSeparatorIdentifier
-            )
-        }
+        collectionView.register(
+            uiConfig.channelList.collectionViewCell.self,
+            forCellWithReuseIdentifier: collectionViewCellReuseIdentifier
+        )
+        
+        collectionView.register(
+            uiConfig.channelList.cellSeparatorReusableView,
+            forSupplementaryViewOfKind: ListCollectionViewLayout.separatorKind,
+            withReuseIdentifier: separatorReuseIdentifier
+        )
 
         collectionView.dataSource = self
         collectionView.delegate = self
@@ -100,7 +108,7 @@ open class _ChatChannelListVC<ExtraData: ExtraDataTypes>: _ViewController,
         cellForItemAt indexPath: IndexPath
     ) -> UICollectionViewCell {
         let cell = collectionView.dequeueReusableCell(
-            withReuseIdentifier: "Cell",
+            withReuseIdentifier: collectionViewCellReuseIdentifier,
             for: indexPath
         ) as! _ChatChannelListCollectionViewCell<ExtraData>
     
@@ -108,6 +116,18 @@ open class _ChatChannelListVC<ExtraData: ExtraDataTypes>: _ViewController,
         cell.itemView.content = (controller.channels[indexPath.row], controller.client.currentUserId)
         
         return cell
+    }
+    
+    public func collectionView(
+        _ collectionView: UICollectionView,
+        viewForSupplementaryElementOfKind kind: String,
+        at indexPath: IndexPath
+    ) -> UICollectionReusableView {
+        collectionView.dequeueReusableSupplementaryView(
+            ofKind: ListCollectionViewLayout.separatorKind,
+            withReuseIdentifier: separatorReuseIdentifier,
+            for: indexPath
+        )
     }
         
     open func collectionView(_ collectionView: UICollectionView, didSelectItemAt indexPath: IndexPath) {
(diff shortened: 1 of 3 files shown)


Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
```diff
@@ -50,11 +50,12 @@ open class _ChatChannelListVC<ExtraData: ExtraDataTypes>: _ViewController,
...
-            collectionViewLayout.register(
+        
+        if let cellSeparatorKind = (collectionViewLayout as? ListCollectionViewLayout)?.separatorKind {
+            collectionView.register(
                 uiConfig.channelList.cellSeparatorReusableView,
-                forDecorationViewOfKind: cellSeparatorIdentifier
+                forSupplementaryViewOfKind: cellSeparatorKind,
+                withReuseIdentifier: "CellSeparatorIdentifier"
             )
         }
```
VojtaStavik, line 60: Hmmm, let's do it the same way UICollectionView does it with section headers and footers:
- make the separator kind a class var `ListCollectionViewLayout.separatorKind`
- the reuse identifier could either be a local constant used within the VC or we could use `uiConfig.channelList.cellSeparatorReusableView.reuseIdentifier`

Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
```diff
@@ -109,6 +110,19 @@ open class _ChatChannelListVC<ExtraData: ExtraDataTypes>: _ViewController,
...
+        _ collectionView: UICollectionView,
+        viewForSupplementaryElementOfKind kind: String,
+        at indexPath: IndexPath
+    ) -> UICollectionReusableView {
+        let cellSeparatorKind = (collectionViewLayout as? ListCollectionViewLayout)?.separatorKind ?? "CellSeparator"
+        return collectionView.dequeueReusableSupplementaryView(
+            ofKind: cellSeparatorKind,
+            withReuseIdentifier: "CellSeparatorIdentifier",
+            for: indexPath
+        ) as! _CellSeparatorReusableView<ExtraData>
```
VojtaStavik, line 124: Maybe we can do it like this?
```
collectionView.dequeueReusableSupplementaryView(
    ofKind: ListCollectionViewLayout.cellSeparatorKind,
    withReuseIdentifier: uiConfig.channelList.cellSeparatorReusableView.reuseIdentifier,
    for: indexPath
)
```

VojtaStavik (changes requested):
Looks good, just a couple of ideas on how to make it bit more flexible and cleaner :)

Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
fortmarek: yep, the `static let` sounds better 🙌 

as for the `reuseIdentifier`, do we want people to access it? afaict it's the same case as for the cell's reuse identifier, so I'd be for keeping it local

VojtaStavik: I think we don't need to specify the reuse identifier locally by default. It's ok to use `uiConfig.channelList.cellSeparatorReusableView.reuseIdentifier` here too, isn't it?

VojtaStavik: we can even remove the `if` check since you can register supplementary views as you want. worst case is that the layout won't use them

fortmarek: sure, we can do it like that)

Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
```diff
@@ -43,20 +43,21 @@ open class _ChatChannelListVC<ExtraData: ExtraDataTypes>: _ViewController,
         .currentUserViewAvatarView.init()
         .withoutAutoresizingMaskConstraints
     
+    private let separatorReuseIdentifier = "CellSeparatorIdentifier"
+
```
VojtaStavik, line 47: There's no reason why this should be private :) Please make it `open`

Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
```diff
@@ -109,6 +110,18 @@ open class _ChatChannelListVC<ExtraData: ExtraDataTypes>: _ViewController,
...
+    public func collectionView(
+        _ collectionView: UICollectionView,
+        viewForSupplementaryElementOfKind kind: String,
+        at indexPath: IndexPath
+    ) -> UICollectionReusableView {
+        collectionView.dequeueReusableSupplementaryView(
+            ofKind: ListCollectionViewLayout.separatorKind,
+            withReuseIdentifier: separatorReuseIdentifier,
+            for: indexPath
+        ) as! _CellSeparatorReusableView<ExtraData>
```
VojtaStavik, line 123: Why do you cast the type here? It should work without the cast, too, shouldn't it?

Sources/StreamChatUI/CommonViews/ListCollectionViewLayout/ListCollectionViewLayout.swift
```diff
@@ -16,10 +16,10 @@ public protocol ListCollectionViewLayoutDelegate: class, UICollectionViewDelegat
 }
 
 /// An `UICollectionViewFlowLayout` implementation to make the collection view behave as a `UITableView`.
-open class ListCollectionViewLayout: UICollectionViewFlowLayout{
+open class ListCollectionViewLayout: UICollectionViewFlowLayout {
 
-    /// The reuse identifier of the cell separator view.
-    open var separatorIdentifier: String = "CellSeparatorIdentifier"
+    /// The kind identifier of the cell separator view.
+    public static let separatorKind: String = "CellSeparator"
```
VojtaStavik, line 22: `public static`-> `open class`

Sources/StreamChatUI/CommonViews/ListCollectionViewLayout/ListCollectionViewLayout.swift
```diff
@@ -38,6 +38,20 @@ open class ListCollectionViewLayout: UICollectionViewFlowLayout{
...
+    override open func invalidateLayout(with context: UICollectionViewLayoutInvalidationContext) {
+        guard let indexPaths = context.invalidatedItemIndexPaths else {
+            super.invalidateLayout(with: context)
+            return
+        }
+        context.invalidateSupplementaryElements(
+            ofKind: Self.separatorKind,
+            at: indexPaths
+        )
+        super.invalidateLayout(with: context)
```
VojtaStavik, line 53: we can replace with:
```swift
override open func invalidateLayout(with context: UICollectionViewLayoutInvalidationContext) {
    if let indexPaths = context.invalidatedItemIndexPaths {
        context.invalidateSupplementaryElements(
            ofKind: Self.separatorKind,
            at: indexPaths
        )
    }
    super.invalidateLayout(with: context)
}
```

VojtaStavik (approved):
Looks good, just some tiny details. Please also take a look here https://www.notion.so/Core-Principles-1c9862955c8d461d9120546d76861c84 , it will help you with future PRs :)

nuno-vieira (approved):
LGTM

PR #864, opened 2021-03-04
[CIS-692] Implement Channel Truncating

VojtaStavik:
When the channel is truncated, it means all messages sent before the truncation disappear. When you query the channel, it has no messages and even the `last_message_at` value is reset.

### Possible ways to implement this locally

a) Delete all messages from the DB -> hard deleting entities from CoreData is not trivial because of our complex data model scheme. This would be the most complicated way to implement this feature. 🔴 

b) Soft-delete messages by `message.type = .deleted`-> this is less complex but not ideal. Deleted messages have different semantics from the truncated messages, especially when it's the messages sent by the current user. Almost there... 🔴 

c) Do it the same way the backed does it -> I added `truncatedAt` field to `ChannelDTO` and adjusted the message predicates to ignore the message created before this date. This solution is the most simple one and works nicely  ✅ 

### Issues

- We will have to introduce some kind of DB cleanup worker that will take care of "physically" removing the truncated messages from the DB. This worker can run only when there are no observers attached to the DB in order to prevent invalid updates. However, the clean-up worker is needed anyway otherwise the DB size would just grow and grow with more incoming messages and new channels.

- Our current custom layout crashes when all messages are removed. I found the issue but the fix was not obvious. I didn't want to spend more time on it since @olejnjak is working on fixing the layout anyway.

### DIFF
--- a/Sources/StreamChat/Controllers/ChannelController/ChannelController_Tests.swift
+++ b/Sources/StreamChat/Controllers/ChannelController/ChannelController_Tests.swift
@@ -494,6 +494,27 @@ class ChannelController_Tests: StressTestCase {
         XCTAssertEqual(controller.messages.map(\.id), [outgoingDeletedMessage.id])
     }
 
+    func test_truncatedMessages_areNotVisible() throws {
+        // Prepare channel with 10 messages
+        try client.databaseContainer.writeSynchronously {
+            try $0.saveChannel(payload: self.dummyPayload(with: self.channelId, numberOfMessages: 10))
+        }
+
+        // Simulate `synchronize` call and check all messages are fetched
+        controller.synchronize()
+        XCTAssertEqual(controller.messages.count, 10)
+
+        // Set channel `truncatedAt` date before the 5th message
+        let truncatedAtDate = self.controller.messages[4].createdAt.addingTimeInterval(-0.1)
+        try client.databaseContainer.writeSynchronously {
+            $0.channel(cid: self.channelId)?.truncatedAt = truncatedAtDate
+        }
+
+        // Check only the 5 messages after the truncatedAt date are visible
+        XCTAssertEqual(controller.messages.count, 5)
+        XCTAssert(controller.messages.allSatisfy { $0.createdAt > truncatedAtDate })
+    }
+
     // MARK: - Delegate tests
     
     func test_settingDelegate_leadsToFetchingLocalData() {
--- a/Sources/StreamChat/Controllers/ChannelController/ChannelController_Tests.swift
+++ b/Sources/StreamChat/Controllers/ChannelController/ChannelController_Tests.swift
@@ -1094,6 +1094,84 @@ class ChannelController_Tests: StressTestCase {
         AssertAsync.willBeEqual(completionCalledError as? TestError, testError)
     }
     
+    // MARK: - Truncating channel
+
+    func test_truncateChannel_failsForNewChannels() throws {
+        //  Create `ChannelController` for new channel
+        let query = _ChannelQuery(channelPayload: .unique)
+        setupControllerForNewChannel(query: query)
+
+        // Simulate `truncateChannel` call and assert error is returned
+        var error: Error? = try await { [callbackQueueID] completion in
+            controller.truncateChannel { error in
+                AssertTestQueue(withId: callbackQueueID)
+                completion(error)
+            }
+        }
+        XCTAssert(error is ClientError.ChannelNotCreatedYet)
+
+        // Simulate successful backend channel creation
+        env.channelUpdater!.update_channelCreatedCallback?(query.cid!)
+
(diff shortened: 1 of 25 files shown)


Sources/StreamChat/Controllers/ChannelController/ChannelController_Tests.swift
```diff
@@ -494,6 +494,31 @@ class ChannelController_Tests: StressTestCase {
...
+            try $0.saveChannel(payload: self.dummyPayload(with: self.channelId, numberOfMessages: 10))
+        }
+
+        // Simulate `synchronize` call and check all messages are fetched
+        controller.synchronize()
+        XCTAssertEqual(controller.messages.count, 10)
+
+        controller.messages.forEach {
+            print($0.createdAt)
+        }
```
b-onc, line 509: Forgot from debugging? :D

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
```diff
@@ -35,7 +41,23 @@ class ChannelDTO: NSManagedObject {
...
+        if changedValues().keys.contains("truncatedAt") {
+            messages
+                .filter { !$0.hasChanges }
+                .forEach {
+                    // Simulate an update
+                    $0.willChangeValue(for: \.id)
+                    $0.didChangeValue(for: \.id)
+                }
+        }
+    }
```
b-onc, line 59: I didn't understand what's going on here... What are the updates we need to trigger?

Sources/StreamChat/Utils/Logger/Logger.swift
```diff
@@ -260,3 +260,16 @@ private extension Logger {
         }
     }
 }
+
+public extension Data {
```
b-onc, line 264: Why `public`?

Sources/StreamChat/WebSocketClient/Events/EventType.swift
```diff
@@ -36,7 +36,9 @@ enum EventType: String, Codable {
     case channelDeleted = "channel.deleted"
     /// When a channel was hidden.
     case channelHidden = "channel.hidden"
-    
+    /// When a channel was trucnated.
```
b-onc, line 39:
```suggestion
    /// When a channel was truncated.
```

Sources/StreamChat/Workers/ChannelUpdater_Tests.swift
```diff
@@ -307,7 +307,7 @@ class ChannelUpdater_Tests: StressTestCase {
     func test_deleteChannel_makesCorrectAPICall() {
         let channelID = ChannelId.unique
 
-        // Simulate `deleteChannel(cid:, completion:)` call
+        // Simulate `truncateChannel(cid:, completion:)` call
```
b-onc, line 310: Why this change?

b-onc (commented):
LGTM overall 🎉

Sources/StreamChat/Controllers/ChannelController/ChannelController_Tests.swift
VojtaStavik: Yes :)  🤦‍♂️

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
VojtaStavik: I created CIS-695 which improves this. It's the same trick we use for attachments and messages. Fetched results controller doesn't observe changes in related objects, even when you specify them in the predicate. So for example:
```
predicate = ALL messages WHERE message.createdAt > channel.truncatedAt
```
will get evaluated properly when you fetch the first time. However, when you change the `channel.truncatedAt` value, the FRC ignores it because it observers only changes to `MessageDTO` entity. This way we force the FRC to re-evaluate the predicate and update the results.

Sources/StreamChat/WebSocketClient/EventMiddlewares/ChannelTruncatedEventMiddleware.swift
```diff
@@ -0,0 +1,29 @@
...
+
+        database.write { session in
+            if let channelDTO = session.channel(cid: truncatedEvent.cid) {
+                channelDTO.truncatedAt = payload.createdAt
+            } else {
+                throw ClientError.ChannelDoesNotExist(cid: truncatedEvent.cid)
+            }
+        } completion: { error in
+            if let error = error {
+                log.error("Failed to update message reaction in the database, error: \(error)")
```
b-onc, line 24: (While copying your code for my Watcher Middleware) I've realized you forgot to change the error message here :trollface:

VojtaStavik: Nice 😅

PR #870, opened 2021-03-08
[CIS-706] Fix custom channel path parameters

olejnjak:
Add tests and fix custom channel path parameters

### DIFF
--- a/Sources/StreamChat/Models/ChannelType.swift
+++ b/Sources/StreamChat/Models/ChannelType.swift
@@ -22,6 +22,8 @@ public enum ChannelType: Codable, Hashable {
     case commerce
     
     /// The type of the channel is custom.
+    ///
+    /// Only small letters, underscore and numbers should be used
     case custom(String)
     
     /// A channel type title.
@@ -35,7 +37,9 @@ public enum ChannelType: Codable, Hashable {
         case .team: return "team"
         case .gaming: return "gaming"
         case .commerce: return "commerce"
-        case let .custom(value): return value
+        case let .custom(value):
+            Self.assertCustomTypeValue(value)
+            return value
         }
     }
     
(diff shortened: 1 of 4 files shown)


Sources/StreamChat/Query/ChannelQuery_Tests.swift
```diff
@@ -57,6 +57,32 @@ class ChannelQuery_Tests: XCTestCase {
...
+            imageURL: .unique(),
+            team: nil,
+            members: [.unique],
+            invites: [],
+            extraData: .defaultValue
+        ))
+        XCTAssertEqual(query.pathParameters, "custom_type")
+    }
+    
+    func test_pathParameters_with_customType_and_id() {
```
VojtaStavik, line 76: nit: we try to follow this pattern when naming tests

olejnjak: 👌

DominikBucher12 (approved):
LGTM 🎉 

Maybe just a nit I noticed in the ticket attachment - shouldn't we add to documentation inside code for `type` that there are only numbers, small letters and underscore allowed in the channel types? 

Or maybe this is unnecessary, but we could restrict the custom type to only allowed characters? Or at least add to documentation of `ChannelType`, I didn't know this until this PR 😅

DominikBucher12 (approved):
<3

Sources/StreamChat/Models/ChannelType.swift
```diff
@@ -49,18 +53,34 @@ public enum ChannelType: Codable, Hashable {
...
         try container.encode(rawValue)
     }
+    
+    /// Check that custom types have valid `rawValue`
+    private static func assertCustomTypeValue(_ value: String) {
+        let allowedCharacters = CharacterSet.lowercaseLetters.union(CharacterSet(charactersIn: "0123456789_"))
+        let valueCharacters = CharacterSet(charactersIn: value)
+        log.assert(
+            valueCharacters.isSubset(of: allowedCharacters),
+            "Allowed characters for `ChannelType.custom` are small letters, numbers and underscore."
```
VojtaStavik, line 83: nit: Can we print the actual value in the error message? It will help with debugging

VojtaStavik (approved):
Looks good 👍 One nit fix and you can merge it :shipit:

PR #871, opened 2021-03-09
[CIS-709] Replace SwiftLint with SwiftFormat --lint

nuno-vieira:
# In this PR

Replaces swiftlint with swiftformat --lint. 
Removes swiftlint from Xcode build phases.
Formats all files.


Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
```diff
@@ -819,7 +819,7 @@ public extension _ChatChannelController {
             channelModificationFailed(completion)
             return
         }
-        guard cooldownDuration >= 1 && cooldownDuration <= 120 else {
+        guard cooldownDuration >= 1, cooldownDuration <= 120 else {
```
DominikBucher12, line 822: Are we sure we want this configuration? i.e. using `,` instead of good old `&&` for conditions? imho in this particular case `&&` makes sense... What about opting-out this rule maybe?

DominikBucher12 (approved):
LGTM just one small comment...

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
nuno-vieira: Done ✅  Disabled the rule 👍

Scripts/swiftformat.sh
```diff
@@ -0,0 +1,19 @@
...
+echo -e "👉 Linting Tests..."
+mint run swiftformat --lint --config .swiftformat Tests
+echo -e "👉 Linting Sample..."
+mint run swiftformat --lint --config .swiftformat Sample
+echo -e "👉 Linting DemoApp..."
+mint run swiftformat --lint --config .swiftformat DemoApp
+echo -e "👉 Linting Integration..."
+mint run swiftformat --lint --config .swiftformat Integration
+echo -e "👉 Linting DocsSnippets..."
+mint run swiftformat --lint --config .swiftformat-snippets DocsSnippets
```
VojtaStavik, line 19: The name of this script is confusing. Maybe it should rather be `run_linter.sh` or something like that?

.github/workflows/ci.yml
```diff
@@ -43,6 +43,32 @@ jobs:
...
+      uses: actions/cache@v2
+      id: mint-cache
+      with:
+        path: /usr/local/lib/mint
+        key: ${{ runner.os }}-mint-${{ hashFiles('./Mintfile') }}
+        restore-keys: ${{ runner.os }}-mint-
+    - name: Run bootstrap.sh
+      run: ./bootstrap.sh
+    - name: Run SwiftFormat
+      run: ./Scripts/swiftformat.sh
```
VojtaStavik, line 71: Hmmm, I don't think we should create a new worker for this. It literally spends minutes initializing and 2 seconds running the actual script :) Can we move this to the DANGER action instead?

VojtaStavik (approved):
Looks like a good first step. What was wrong with the Danger plugin? It would be cool to have it unified via Danger.

Gemfile
```diff
@@ -7,7 +7,7 @@ git_source(:github) {|repo_name| "https://github.com/#{repo_name}" }
 gem "fastlane"
 gem "cocoapods"
 gem "danger"
-gem "danger-swiftlint"	
+gem "danger-swiftformat"
```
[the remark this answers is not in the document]
nuno-vieira: Forgot to delete this 🤦‍♂️

Scripts/swiftformat.sh
nuno-vieira: Yes it is better to not confuse with the Swiftformat config

.github/workflows/ci.yml
nuno-vieira: Yes I can move to the danger job, but maybe that is confusing and people think this is part of danger ? Because it is not, if linting fails, danger won't post a comment anywhere. Also this linting is not run by danger so the job name is would be a bit misleading 🤔 What if we renamed the job to "Automated PR Review", or something like that

Gemfile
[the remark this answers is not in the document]
nuno-vieira: ✅

Scripts/swiftformat.sh
nuno-vieira: ✅

.github/workflows/ci.yml
nuno-vieira: @VojtaStavik Let me know if this is okay now: https://github.com/GetStream/stream-chat-swift/pull/871/commits/199b50f4cc2d08e39b7760ba4be8bbc69944b328

I could not setup swift format in danger, because there's not much information about that plugin. It looks it expects to set the binary path of swiftformat in the config file, and they recommend putting the binary in our repo, but we use Mint to manage swiftformat. So it is not clear on danger plugin' repo on how we can set it up with Mint.  

Maybe we can create a Ticket to investigate this further?

VojtaStavik: Sounds good regarding the ticket 👍 Regarding the linter script location - I see your point. It's up to you. I wouldn't worry about it too much since we will solve it in CIS-684 🤷 So yeah, either keep it as it, or move it and rename.

PR #874, opened 2021-03-09
[CIS-713] Prefix SDK Version in Client Header with 'v'

nuno-vieira:
# In this PR

Send the client header version prefixed with ´v´ so we can identify the new SDK versions in the Dashboard.


VojtaStavik (approved):
Cool 👍

PR #875, opened 2021-03-10
[CIS-718] Fix message actions not visible

nuno-vieira:
# In this PR

Fixes hidden message actions. When getting the `currentUser` from `CurrentUserController`, it was missing the start observing. So the `currentUser` would only be retrieved if `synchronize` was called.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### 🐞 Fixed
 - Fix encoding of channels with custom type [#872](https://github.com/GetStream/stream-chat-swift/pull/872)
+- Fix `CurreUserController.currentUser` returning nil before `synchronize()` is called [#875](https://github.com/GetStream/stream-chat-swift/pull/875)
 
 # [3.1.1](https://github.com/GetStream/stream-chat-swift/releases/tag/3.1.1)
 _February 26, 2021_
--- a/Sources/StreamChat/Controllers/CurrentUserController/CurrentUserController.swift
+++ b/Sources/StreamChat/Controllers/CurrentUserController/CurrentUserController.swift
@@ -70,7 +70,8 @@ public class _CurrentChatUserController<ExtraData: ExtraDataTypes>: DataControll
     /// The currently logged-in user. `nil` if the connection hasn't been fully established yet, or the connection
     /// wasn't successful.
     public var currentUser: _CurrentChatUser<ExtraData.User>? {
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### 🐞 Fixed
 - Fix encoding of channels with custom type [#872](https://github.com/GetStream/stream-chat-swift/pull/872)
+- Fix currentUser returning nil if `synchronize` was not called [#875](https://github.com/GetStream/stream-chat-swift/pull/875)
```
VojtaStavik, line 18: Changelog should be written from the user's perspective. So this should be something like `Fix message actions not visible on long press`

Sources/StreamChat/Controllers/CurrentUserController/CurrentUserController.swift
```diff
@@ -70,7 +70,8 @@ public class _CurrentChatUserController<ExtraData: ExtraDataTypes>: DataControll
     /// The currently logged-in user. `nil` if the connection hasn't been fully established yet, or the connection
     /// wasn't successful.
     public var currentUser: _CurrentChatUser<ExtraData.User>? {
-        currentUserObserver.item
+        startObservingIfNeeded()
+        return currentUserObserver.item
```
VojtaStavik, line 74: So was this just a copy-paste mistake?

VojtaStavik (approved):

CHANGELOG.md
nuno-vieira: Yes the reason I did this is because although we identified the issue in the actions. It is actually a broader problem and is related with LLC. If someone was using LLC now, the messages hidden would not make sense if they were looking to fix an issue where they could not get the current user. 🤔

Sources/StreamChat/Controllers/CurrentUserController/CurrentUserController.swift
nuno-vieira: Yes exactly, forgot to put this when converting this to a DataController

CHANGELOG.md
VojtaStavik: I see, in that case, it could be something like: 
`Fix `CurrentUserController.currentUser` incorrectly returning `nil` before `synchronize` is called`?

nuno-vieira: Yes that is a bit more clear, I'll change it 👍

nuno-vieira: ✅

PR #876, opened 2021-03-10
WIP Fix codecov

nuno-vieira:
# In this PR

WIP


.github/workflows/ci.yml
```diff
@@ -108,8 +112,6 @@ jobs:
       run: ./bootstrap.sh
     - name: Run v3 UI Tests (Debug)
       run: bundle exec fastlane test_ui device:"iPhone 12"
-    - name: Post Codecov report
-      run: bash <(curl -s https://codecov.io/bash) -t ${{ secrets.CODECOV_TOKEN }}
```
b-onc, line 112: Why don't we post codecov for UI tests?

.github/workflows/ci.yml
```diff
@@ -71,8 +71,12 @@ jobs:
       run: ./bootstrap.sh
     - name: Run v3 Tests (Debug)
       run: bundle exec fastlane test
-    - name: Post Codecov report
-      run: bash <(curl -s https://codecov.io/bash) -t ${{ secrets.CODECOV_TOKEN }}
+    - uses: codecov/codecov-action@v1
+      with:
+        token: ${{ secrets.CODECOV_TOKEN }}
+        flags: unittests-ios14
+        fail_ci_if_error: true # if codecov upload fails, should fail ci
```
b-onc, line 78: How often does the upload fail? :D

.github/workflows/ci.yml
b-onc: Why don't we post it with a flag, too?

b-onc (approved):
I didn't know codecov already has a github action, and the flags! That's awesome. If CI is green & looks good, we can merge it

.github/workflows/ci.yml
nuno-vieira: We had the code coverage off when running the UI Tests with fastlane's scan because we were in the early stages of the Audits. But I guess we can turn it on now, so we can track the progress 🤔  I think it is a good idea 👍  WDYT?  cc @VojtaStavik

.github/workflows/ci.yml
nuno-vieira: I don't know actually 😄 But they say this can happen randomly 🤷‍♂️  These unexpected changes are hard to debug, so better to fix every variable we can. Even with this PR, I'm not yet 100% this will fix it.

.github/workflows/ci.yml
b-onc: This flag basically guarantees we don't mess up coverages, right? It's safe to enable in that case

.github/workflows/ci.yml
b-onc: fair enough, we can merge and see

.github/workflows/ci.yml
nuno-vieira: Yes exactly

.github/workflows/ci.yml
nuno-vieira: Okay, let's merge it then 👍 I'll add the UI tests in the Clean CI ticket I'm currently working on

VojtaStavik (approved):

PR #880, opened 2021-03-11
[CIS-726] Introduce MemberEventMiddleware

b-onc:
I was surprised to see `saveMember` doesn't actually link the member to the Channel! Is that intentional? Introducing this logic didn't break any test.


VojtaStavik (approved):
Nice 👍

PR #881, opened 2021-03-11
Add cancel workflow action

b-onc:
Thanks to @DominikBucher12

### DIFF
--- a/.github/workflows/cancel.yml
+++ b/.github/workflows/cancel.yml
@@ -1,0 +1,14 @@
+name: Cancel obsolete workflows
+
+on: [push]
+
+jobs:
+  cancel-obsolete-workflows:
+    name: 'Cancel obsolete workflows'
+    runs-on: macos-latest
+    timeout-minutes: 3
+    steps:
+      - uses: styfle/cancel-workflow-action@0.8.0
+        with:
+          workflow_id: "smoke-checks.yml, distribute-test-build.yml" # Workflow file names to cancel
+          access_token: ${{ github.token }}


DominikBucher12 (approved):
🙌 Noice <3

.github/workflows/cancel.yml
```diff
@@ -0,0 +1,14 @@
...
+jobs:
+  cancel-obsolete-workflows:
+    name: 'Cancel obsolete workflows'
+    runs-on: macos-latest
+    timeout-minutes: 3
+    steps:
+      - uses: styfle/cancel-workflow-action@0.8.0
+        with:
+          workflow_id: beta, ci # Workflow file names to cancel
+          access_token: ${{ github.token }}
```
VojtaStavik, line 14: Can we somehow bump the priority of this action? If there are already 20 actions in the queue, it won't help that this will be executed as the 21st :) Or am I missing somethign?

nuno-vieira (approved):
LGTM

VojtaStavik (approved):

.github/workflows/cancel.yml
VojtaStavik: Probably makes sense? cc @nuno-vieira

nuno-vieira: Yes I'll add this to the other jobs 👍

VojtaStavik:
@b-onc Can we merge this?

VojtaStavik:
@b-onc I thought you want to add this job as the dependency for all others...

b-onc:
Merging this as is, since there's currently no way to prioritize jobs. Creator suggests using this action in every job's first step, but that might be overkill, let's see.
https://github.com/styfle/cancel-workflow-action/issues/49

PR #884, opened 2021-03-11
[CIS-684] Clean CI Jobs

nuno-vieira:
# In this PR
- Restructures the workflows
- CI Checks are now Smoke-Checks
- Nighly Checks are now Full-Checks
- Creates bootstrap action, so it is reused across workflows
- Creates prepare-ios-simulator action, so it is reused and configurable across workflows


fastlane/Fastfile
```diff
@@ -216,9 +216,23 @@ lane :test_ui do |options|
...
-    code_coverage: false
-    )
+    devices: options[:device]
+  )
+end
+
+desc "Runs ui tests in Release config"
+lane :test_ui_release do |options|
+
+  setCIEnvironmentVariable("../Tests/StreamChatUITests/StreamChatUITestPlan.xctestplan")
```
b-onc, line 226: Can we also add the `if !is_ci {}` check to this action in this PR?

.github/workflows/full-checks.yml
```diff
@@ -0,0 +1,202 @@
...
+
+jobs:
+  build-and-test-release:
+    name: Test LLC + UI (Release)
+    runs-on: macos-latest
+    steps:
+    - uses: actions/checkout@v1
+    - name: Set build image var
+      run: echo "ImageVersion=$ImageVersion"  >> $GITHUB_ENV
+    - name: Cache RubyGems
```
b-onc, line 19: Why don't we add an action for this too?

.github/workflows/full-checks.yml
```diff
@@ -0,0 +1,202 @@
...
+      uses: actions/cache@v2
+      id: rubygem-cache
+      with:
+        path: vendor/bundle
+        key: ${{ runner.os }}-${{ env.ImageVersion }}-gem-${{ hashFiles('**/Gemfile.lock') }}
+        restore-keys: ${{ runner.os }}-${{ env.ImageVersion }}-gem-
+    - name: Cache Mint
```
b-onc, line 26: Why don't we add an action for this too?

.github/workflows/full-checks.yml
```diff
@@ -0,0 +1,202 @@
...
+    - name: Run UI Tests (Release)
+      run: bundle exec fastlane test_ui_release device:"iPhone 12"
+    - uses: 8398a7/action-slack@v3
+      with:
+        status: ${{ job.status }}
+        text: "You shall not pass!"
+        fields: message,commit,author,action,eventName,workflow,job,took
+      env:
+        SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
+        MATRIX_CONTEXT: ${{ toJson(matrix) }}
```
b-onc, line 45: Why don't we add an action for this too?

fastlane/Fastfile
nuno-vieira: Yes makes sense 👍

.github/workflows/full-checks.yml
nuno-vieira: The build-image-var action we can create it yes. 👍

.github/workflows/full-checks.yml
nuno-vieira: We can only create actions that run bash scripts. We can't use external actions on our internal actions, unfortunately

.github/workflows/full-checks.yml
nuno-vieira: Same as above

.github/workflows/full-checks.yml
nuno-vieira: That is why me and @VojtaStavik were discussing on generating the workflows from swift, and reuse the actions in that way.

.github/workflows/demo-distribution.yml
```diff
@@ -1,4 +1,4 @@
-name: Firebase Test Builds
+name: Distribute Demo App
```
VojtaStavik, line 1: Can we name this `Distribute Firebase Test Build`? We will later have a TestFlight version of this, too

.github/workflows/smoke-checks.yml
```diff
@@ -232,52 +120,11 @@ jobs:
...
-      run: |
-        sudo mkdir -p /Library/Developer/CoreSimulator/Profiles/Runtimes
-        sudo ln -s /Applications/Xcode_10.3.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/Library/CoreSimulator/Profiles/Runtimes/iOS.simruntime /Library/Developer/CoreSimulator/Profiles/Runtimes/iOS\ 12.4.simruntime
-        xcrun simctl list runtimes
-        xcrun simctl create custom-test-device "iPhone 7" "com.apple.CoreSimulator.SimRuntime.iOS-12-4"
-        xcrun simctl list devices 12.4
-    - name: Build Sample
-      run: bundle exec fastlane build_sample device:"iPhone 7 (12.4)"
+    - name: Build Demo App
+      run: bundle exec fastlane build_demo
```
VojtaStavik, line 127: Do we want `Build Docs Snippets` as a part of smoke checks? Ideally, we'd keep the number of jobs <= 5

fastlane/Fastfile
VojtaStavik: We have a whole separate ticket for Fastfile changes https://stream-io.atlassian.net/secure/RapidBoard.jspa?rapidView=52&projectKey=CIS&modal=detail&selectedIssue=CIS-233

VojtaStavik (approved):
LGTM 👍 I'll leave the implementation details review to @b-onc  and @DominikBucher12

.github/workflows/smoke-checks.yml
nuno-vieira: This one is pretty fast to run and it is useful. I've already broken the docs snippets, and it was very easy to fix them because I had context on what broke. It would take much more time for someone else to fix it. So all the things that a PR can break, we should make sure it gets fixed there, so someone else won't be losing time before a release or something like that

fastlane/Fastfile
nuno-vieira: We will add this change in the Fastfile cleanup then.

.github/workflows/full-checks.yml
nuno-vieira: ✅

b-onc (approved):

.github/actions/prepare-ios-simulator/action.yml
```diff
@@ -0,0 +1,21 @@
...
+runs:
+  using: "composite"
+  steps:
+    - run: 
+        sudo mkdir -p /Library/Developer/CoreSimulator/Profiles/Runtimes
+        sudo ln -s /Applications/Xcode_12.4.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/Library/CoreSimulator/Profiles/Runtimes/iOS.simruntime /Library/Developer/CoreSimulator/Profiles/Runtimes/iOS\ ${{ inputs.ios }}.simruntime
+        xcrun simctl list runtimes
+        ios=${{ inputs.ios }}
+        ios_runtime=${ios//./-}
+        xcrun simctl create custom-test-device "${{ inputs.device }}" "com.apple.CoreSimulator.SimRuntime.iOS-$ios_runtime"
```
DominikBucher12, line 18: This probably creates new device all the time, possibly leading to space issues  in some time... Are devices deleted somewhere in the pipeline??  
Snímek obrazovky 2021-03-15 v 17 52 42


Snímek obrazovky 2021-03-15 v 17 51 50

DominikBucher12: What you can do is to `xcrun simctl delete custom-test-device` and `xcrun simctl delete unavailable` to cleanup. problem with `xcrun simctl delete custom-test-device`  is that it will delete only last one because there are more `custom-test-device` instances with different UDID(UUIDs?) :D Always get those wrong :D

PR #890, opened 2021-03-15
Add cleanup to stress tests

olejnjak:
- adds cleanup after each test suite so disk space issue should be resolved
- prefers test failure to crash if `DatabaseContainer_Mock` in `ChatClient_Mock` cannot be instantiated

### DIFF
--- a/Tests/Shared/StressTestCase.swift
+++ b/Tests/Shared/StressTestCase.swift
@@ -5,6 +5,12 @@
 import XCTest
 
 class StressTestCase: XCTestCase {
+    override func setUp() {
+        super.setUp()
+        
+        continueAfterFailure = false
+    }
+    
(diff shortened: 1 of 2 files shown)


Tests/Shared/StressTestCase.swift
```diff
@@ -4,13 +4,42 @@
...
+/// - by default ends test when test failure occurs
 class StressTestCase: XCTestCase {
     override func setUp() {
         super.setUp()
         
         continueAfterFailure = false
     }
     
+    override class func tearDown() {
+        super.tearDown()
```
VojtaStavik, line 20: nit: `super.tearDown` should be called at the end of the method. However, it's probably OK for this particular scenario.

VojtaStavik (approved):

PR #891, opened 2021-03-16
Cleanup CI Full Checks Tests

nuno-vieira:
# In This PR

- Add iOS 13 to building sample apps
- Adds latest iOS to Stress tests
- Runs Stress tests only in Release


VojtaStavik (approved):
👍

PR #893, opened 2021-03-16
Fix watcher events received for current user during query

b-onc:
This happens because when we query channels with watch: true, our current user watches the channels and backend sends the `userStartedWatching` event before we save the channels that we just watched to the DB.

### DIFF
--- a/Sources/StreamChat/WebSocketClient/EventMiddlewares/UserWatchingEventMiddleware.swift
+++ b/Sources/StreamChat/WebSocketClient/EventMiddlewares/UserWatchingEventMiddleware.swift
@@ -16,8 +16,22 @@ struct UserWatchingEventMiddleware<ExtraData: ExtraDataTypes>: EventMiddleware {
         
         database.write { session in
             guard let channelDTO = session.channel(cid: userWatchingEvent.cid) else {
-                throw ClientError.ChannelDoesNotExist(cid: userWatchingEvent.cid)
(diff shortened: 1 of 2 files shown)


Sources/StreamChat/WebSocketClient/EventMiddlewares/UserWatchingEventMiddleware.swift
```diff
@@ -16,19 +16,35 @@ struct UserWatchingEventMiddleware<ExtraData: ExtraDataTypes>: EventMiddleware {
...
-            guard let userDTO = session.user(id: userWatchingEvent.userId) else {
-                throw ClientError.UserDoesNotExist(userId: userWatchingEvent.userId)
-            }
+            channelDTO.watcherCount = Int64(userWatchingEvent.watcherCount)
             
-            if userWatchingEvent.isStarted {
-                channelDTO.watchers.insert(userDTO)
+            if let eventPayload = userWatchingEvent.payload as? EventPayload<ExtraData>,
+               let userPayload = eventPayload.user {
+                let userDTO = try session.saveUser(payload: userPayload)
```
VojtaStavik, line 38: Isn't the payload already saved in `EventDataProcessorMiddleware`? IMO the original version of this was completely fine.

VojtaStavik (commented):
Either I don't understand something or this is more code than needed :)

VojtaStavik (approved):

PR #895, opened 2021-03-17
[CIS-710] Fix flaky LLC tests

olejnjak:
Wait for release of database so CoreData tests do no affect each other. I also reduced stress tests cycles to 5 for now so we pass the 6h job limit on CI.

Seems to work well as tonight stress tests [passed](https://github.com/GetStream/stream-chat-swift/actions/runs/659391394) 🎉

### DIFF
--- a/Tests/Shared/TestRunnerEnvironment.swift
+++ b/Tests/Shared/TestRunnerEnvironment.swift
@@ -4,16 +4,18 @@
 
 import Foundation
 
-struct TestRunnerEnvironment {
+enum TestRunnerEnvironment {
     /// `true` if the tests are currently running on the CI. We get this information by checking for the custom `CI` environment
     /// variable passed in to the process.
     static var isCI: Bool {
         ProcessInfo.processInfo.environment["CI"] == "TRUE"
     }
     
-    /// `true` if the current tests are stress test. We get this information by checking for the custom `STRESS_TESTS` environment
-    /// variable passed in to the process.
-    static var isStressTest: Bool {
-        ProcessInfo.processInfo.environment["STRESS_TESTS"] == "TRUE"
+    /// Number of stress test invocations
+    static var testInvocations: Int {
+        ProcessInfo.processInfo.environment["TEST_INVOCATIONS"].flatMap(Int.init) ?? 1
     }
+    
+    /// `true` if we invoke stress tests more than a single time
+    static var isStressTest: Bool { testInvocations > 1 }
 }
--- a/Tests/StreamChatTests/StreamChatStressTestPlan.xctestplan
+++ b/Tests/StreamChatTests/StreamChatStressTestPlan.xctestplan
@@ -12,10 +12,11 @@
     "codeCoverage" : false,
     "environmentVariableEntries" : [
       {
-        "key" : "STRESS_TESTS",
-        "value" : "TRUE"
+        "key" : "TEST_INVOCATIONS",
+        "value" : "10"
       }
-    ]
+    ],
+    "testExecutionOrdering" : "random"
   },
   "testTargets" : [
     {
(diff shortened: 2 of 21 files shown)


fastlane/Fastfile
```diff
@@ -4,7 +4,7 @@ default_platform :ios
 require 'json'
 
 # The number of times the stress test suite is ran
-stress_tests_cycles = 10
+stress_tests_cycles = 5
```
VojtaStavik, line 7: Instead of this, can you please change the `StressTestCase` number of cycles to say 10 + make the tests run randomly + increase this to ~50?

Basically, more cycles with random tests but less reparations of each tests

VojtaStavik (approved):

fastlane/Fastfile
olejnjak: What do you think about converting the `STRESS_TESTS` environment variable to something like `TEST_INVOCATIONS` so eventually we are able to set number of invocation from each CI job?

VojtaStavik: I like it! I actually thought about running it at least once 100x and this would solve it perfectly 🎉

Tests/Shared/TestRunnerEnvironment.swift
```diff
@@ -4,16 +4,18 @@
 
 import Foundation
 
-struct TestRunnerEnvironment {
+enum TestRunnerEnvironment {
```
DominikBucher12, line 7: <3

Tests/Shared/TestRunnerEnvironment.swift
```diff
@@ -4,16 +4,18 @@
...
         ProcessInfo.processInfo.environment["CI"] == "TRUE"
     }
     
-    /// `true` if the current tests are stress test. We get this information by checking for the custom `STRESS_TESTS` environment
-    /// variable passed in to the process.
-    static var isStressTest: Bool {
-        ProcessInfo.processInfo.environment["STRESS_TESTS"] == "TRUE"
+    /// Number of stress test invocations
+    static var testInvocations: Int {
+        ProcessInfo.processInfo.environment["TEST_INVOCATIONS"].flatMap(Int.init) ?? 1
```
DominikBucher12, line 16: Just curious... why 1 test invocation when there are no tests??

Tests/StreamChatTests/StreamChatStressTestPlan.xctestplan
```diff
@@ -12,10 +12,11 @@
     "codeCoverage" : false,
     "environmentVariableEntries" : [
       {
-        "key" : "STRESS_TESTS",
-        "value" : "TRUE"
+        "key" : "TEST_INVOCATIONS",
+        "value" : "10"
```
DominikBucher12, line 16: super nit: `version` property somewhere below is `1` . This value should support Integers AFAIK...

DominikBucher12 (approved):
Feel free to ignore comments, because they are probably irelevant (Just my curiousity/nitpicking) 

Thanks <3

Tests/Shared/TestRunnerEnvironment.swift
olejnjak: Not sure I follow, what do you mean by "there are no tests"? 😄 1 is default value if environment variables do not contain `TEST_INVOCATIONS` (e.g. default test plan) or the value is not `Int` convertible

Tests/StreamChatTests/StreamChatStressTestPlan.xctestplan
olejnjak: Well it does not, as [environment variables](https://developer.apple.com/documentation/foundation/processinfo/1417911-environment) are always `String`

DominikBucher12: Wow now on second look I saw the dict in the file... Thanks :)

PR #896, opened 2021-03-17
[CIS-578] Support Pinned Messages

nuno-vieira:
# In this PR

- Add pin support when creating new messages
- Add pin and unpin support for existing messages
- Add support for getting latest 10 pinnedMessages

### DIFF
--- a/Sources/StreamChat/APIClient/Endpoints/Payloads/ChannelListPayload_Tests.swift
+++ b/Sources/StreamChat/APIClient/Endpoints/Payloads/ChannelListPayload_Tests.swift
@@ -49,6 +49,8 @@ class ChannelPayload_Tests: XCTestCase {
         XCTAssert(firstMessage.reactionScores.isEmpty)
         XCTAssertEqual(firstMessage.replyCount, 0)
         XCTAssertFalse(firstMessage.isSilent)
+
+        XCTAssertEqual(payload.pinnedMessages.map(\.id), ["broken-waterfall-5-7aede36b-b89f-4f45-baff-c40c7c1875d9"])
         
         let channel = payload.channel
         XCTAssertEqual(channel.cid, try! ChannelId(cid: "messaging:general"))
--- a/Sources/StreamChat/APIClient/Endpoints/Payloads/MessagePayloads.swift
+++ b/Sources/StreamChat/APIClient/Endpoints/Payloads/MessagePayloads.swift
@@ -28,6 +28,10 @@ enum MessagePayloadsCodingKeys: String, CodingKey {
     case reactionScores = "reaction_scores"
     case isSilent = "silent"
     case channel
+    case pinned
+    case pinnedBy = "pinned_by"
+    case pinnedAt = "pinned_at"
+    case pinExpires = "pin_expires"
     //        case i18n
 }
 
@@ -64,6 +68,11 @@ class MessagePayload<ExtraData: ExtraDataTypes>: Decodable {
     let attachments: [AttachmentPayload]
     let isSilent: Bool
 
+    var pinned: Bool
+    var pinnedBy: UserPayload<ExtraData.User>?
+    var pinnedAt: Date?
+    var pinExpires: Date?
+
     /// Only message payload from `getMessage` endpoint contains channel data. It's a convenience workaround for having to
     /// make an extra call do get channel details.
     let channel: ChannelDetailPayload<ExtraData>?
@@ -99,6 +108,11 @@ class MessagePayload<ExtraData: ExtraDataTypes>: Decodable {
         
         // Some endpoints return also channel payload data for convenience
         channel = try container.decodeIfPresent(ChannelDetailPayload<ExtraData>.self, forKey: .channel)
+
+        pinned = try container.decodeIfPresent(Bool.self, forKey: .pinned) ?? false
+        pinnedBy = try container.decodeIfPresent(UserPayload<ExtraData.User>.self, forKey: .pinnedBy)
+        pinnedAt = try container.decodeIfPresent(Date.self, forKey: .pinnedAt)
+        pinExpires = try container.decodeIfPresent(Date.self, forKey: .pinExpires)
     }
     
     init(
@@ -124,7 +138,11 @@ class MessagePayload<ExtraData: ExtraDataTypes>: Decodable {
         reactionScores: [MessageReactionType: Int],
         isSilent: Bool,
         attachments: [AttachmentPayload],
-        channel: ChannelDetailPayload<ExtraData>? = nil
+        channel: ChannelDetailPayload<ExtraData>? = nil,
+        pinned: Bool = false,
+        pinnedBy: UserPayload<ExtraData.User>? = nil,
+        pinnedAt: Date? = nil,
+        pinExpires: Date? = nil
     ) {
         self.id = id
         self.type = type
@@ -148,6 +166,10 @@ class MessagePayload<ExtraData: ExtraDataTypes>: Decodable {
         self.isSilent = isSilent
         self.attachments = attachments
         self.channel = channel
+        self.pinned = pinned
+        self.pinnedBy = pinnedBy
+        self.pinnedAt = pinnedAt
+        self.pinExpires = pinExpires
     }
 }
 
@@ -162,6 +184,8 @@ struct MessageRequestBody<ExtraData: ExtraDataTypes>: Encodable {
     let showReplyInChannel: Bool
     let quotedMessageId: String?
     let attachments: [Encodable]
+    var pinned: Bool
+    var pinExpires: Date?
     let extraData: ExtraData.Message
     
     init(
@@ -174,6 +198,8 @@ struct MessageRequestBody<ExtraData: ExtraDataTypes>: Encodable {
         showReplyInChannel: Bool = false,
         quotedMessageId: String? = nil,
         attachments: [Encodable] = [],
+        pinned: Bool = false,
+        pinExpires: Date? = nil,
         extraData: ExtraData.Message
     ) {
         self.id = id
@@ -185,6 +211,8 @@ struct MessageRequestBody<ExtraData: ExtraDataTypes>: Encodable {
         self.showReplyInChannel = showReplyInChannel
         self.quotedMessageId = quotedMessageId
         self.attachments = attachments
+        self.pinned = pinned
+        self.pinExpires = pinExpires
         self.extraData = extraData
     }
     
@@ -197,7 +225,9 @@ struct MessageRequestBody<ExtraData: ExtraDataTypes>: Encodable {
         try container.encodeIfPresent(parentId, forKey: .parentId)
         try container.encodeIfPresent(showReplyInChannel, forKey: .showReplyInChannel)
         try container.encodeIfPresent(quotedMessageId, forKey: .quotedMessageId)
-        
+        try container.encode(pinned, forKey: .pinned)
+        try container.encodeIfPresent(pinExpires, forKey: .pinExpires)
+
         if !attachments.isEmpty {
             try container.encode(attachments.map(AnyEncodable.init), forKey: .attachments)
         }
--- a/Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
+++ b/Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
@@ -690,15 +690,17 @@ public extension _ChatChannelController {
     ///
     /// - Parameters:
     ///   - text: Text of the message.
-    ///   - extraData: Additional extra data of the message object.
+    ///   - pinning: Pins the new message. `nil` if should not be pinned.
     ///   - attachments: An array of the attachments for the message.
     ///     `Note`: can be built-in types, custom attachment types conforming to `AttachmentEnvelope` protocol
     ///     and `ChatMessageAttachmentSeed`s.
     ///   - quotedMessageId: An id of the message new message quotes. (inline reply)
+    ///   - extraData: Additional extra data of the message object.
     ///   - completion: Called when saving the message to the local DB finishes.
     ///
     func createNewMessage(
         text: String,
+        pinning: MessagePinning? = nil,
 //        command: String? = nil,
 //        arguments: String? = nil,
         attachments: [AttachmentEnvelope] = [],
@@ -720,6 +722,7 @@ public extension _ChatChannelController {
         updater.createNewMessage(
             in: cid,
             text: text,
+            pinning: pinning,
             command: nil,
             arguments: nil,
             attachments: attachments,
@@ -849,7 +852,7 @@ public extension _ChatChannelController {
             }
         }
     }
-    
+
     /// Start watching a channel
     ///
     /// Watching a channel is defined as observing notifications about this channel.
--- a/Sources/StreamChat/Controllers/ChannelController/ChannelController_Tests.swift
+++ b/Sources/StreamChat/Controllers/ChannelController/ChannelController_Tests.swift
@@ -384,7 +384,11 @@ class ChannelController_Tests: StressTestCase {
         
         // Simulate an incoming message
         let newMessageId: MessageId = .unique
-        let newMessagePayload: MessagePayload<NoExtraData> = .dummy(messageId: newMessageId, authorUserId: .unique)
+        let newMessagePayload: MessagePayload<NoExtraData> = .dummy(
+            messageId: newMessageId,
+            authorUserId: .unique,
+            createdAt: Date()
+        )
         _ = try await {
             client.databaseContainer.write({ session in
                 try session.saveMessage(payload: newMessagePayload, for: self.channelId)
@@ -414,7 +418,7 @@ class ChannelController_Tests: StressTestCase {
         // Check the order of messages is correct
         let topToBottomIds = [message1, message2].sorted { $0.createdAt > $1.createdAt }.map(\.id)
         XCTAssertEqual(controller.messages.map(\.id), topToBottomIds)
-        
+
         // Set bottom-to-top ordering
         controller.listOrdering = .bottomToTop
         
@@ -594,8 +598,8 @@ class ChannelController_Tests: StressTestCase {
         }
         XCTAssertNil(error)
         let channel: ChatChannel = client.databaseContainer.viewContext.channel(cid: channelId)!.asModel()
-        assert(channel.latestMessages.count == 1)
-        let message: ChatMessage = channel.latestMessages.first!
+        XCTAssertEqual(channel.latestMessages.count, 1)
+        let message: ChatMessage = try XCTUnwrap(channel.latestMessages.first)
 
         // Assert DB observers call delegate updates
         AssertAsync {
@@ -737,8 +741,8 @@ class ChannelController_Tests: StressTestCase {
         }
         XCTAssertNil(error)
         let channel: ChatChannel = client.databaseContainer.viewContext.channel(cid: channelId)!.asModel()
-        assert(channel.latestMessages.count == 1)
-        let message: ChatMessage = channel.latestMessages.first!
+        XCTAssertEqual(channel.latestMessages.count, 1)
+        let message: ChatMessage = try XCTUnwrap(channel.latestMessages.first)
         
         AssertAsync {
             Assert.willBeEqual(delegate.didUpdateChannel_channel, .create(channel))
@@ -760,8 +764,8 @@ class ChannelController_Tests: StressTestCase {
             }, completion: $0)
         }
         let channel: ChatChannel = client.databaseContainer.viewContext.channel(cid: channelId)!.asModel()
-        assert(channel.latestMessages.count == 1)
-        let message: ChatMessage = channel.latestMessages.first!
+        XCTAssertEqual(channel.latestMessages.count, 1)
+        let message: ChatMessage = try XCTUnwrap(channel.latestMessages.first)
         
         AssertAsync {
             Assert.willBeEqual(delegate.didUpdateChannel_channel, .create(channel))
@@ -1339,6 +1343,7 @@ class ChannelController_Tests: StressTestCase {
         let message = try session.createNewMessage(
             in: channelId,
             text: "Message",
+            pinning: nil,
             quotedMessageId: nil,
             attachmentSeeds: [
                 ChatMessageAttachmentSeed.dummy(),
@@ -1636,11 +1641,13 @@ class ChannelController_Tests: StressTestCase {
             .dummy()
         ]
         let quotedMessageId: MessageId = .unique
+        let pin = MessagePinning(expirationDate: .unique)
         
         // Simulate `createNewMessage` calls and catch the completion
         var completionCalled = false
         controller.createNewMessage(
             text: text,
+            pinning: pin,
 //            command: command,
 //            arguments: arguments,
             attachments: attachments + attachmentSeeds,
@@ -1675,6 +1682,7 @@ class ChannelController_Tests: StressTestCase {
             attachmentSeeds
         )
         XCTAssertEqual(env.channelUpdater?.createNewMessage_quotedMessageId, quotedMessageId)
+        XCTAssertEqual(env.channelUpdater?.createNewMessage_pinning?.expirationDate, pin.expirationDate!)
         
         // Simulate successful update
         env.channelUpdater?.createNewMessage_completion?(.success(newMessageId))
@@ -1711,7 +1719,7 @@ class ChannelController_Tests: StressTestCase {
             XCTFail("Expected .failure but received \(result)")
         }
     }
-    
+
     // MARK: - Adding members
     
     func test_addMembers_failsForNewChannels() throws {
--- a/Sources/StreamChat/Controllers/MessageController/MessageController.swift
+++ b/Sources/StreamChat/Controllers/MessageController/MessageController.swift
@@ -350,6 +350,29 @@ public extension _ChatMessageController {
             }
         }
     }
+
+    /// Pin the message this controller manages.
+    ///  - Parameters:
+    ///   - pinning: The pinning expiration information. It supports setting an infinite expiration, setting a date, or the amount of time a message is pinned.
+    ///   - completion: A completion block with an error if the request was failed.
+    func pin(_ pinning: MessagePinning, completion: ((Error?) -> Void)? = nil) {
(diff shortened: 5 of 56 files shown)


Sources/StreamChat/Query/ChannelListQuery.swift
```diff
@@ -74,6 +74,9 @@ public extension FilterKey where Scope: AnyChannelListFilterScope {
 
     /// A filter key for matching the `memberCount` value.
     static var memberCount: FilterKey<Scope, Int> { "member_count" }
+
+    /// A filter key for matching the `pinned` value.
+    static var pinned: FilterKey<Scope, Bool> { "pinned" }
```
VojtaStavik, line 79: As talked in Slack, this is probably not correct

Sources/StreamChat/WebSocketClient/EventMiddlewares/ChannelPinnedMessagesUpdaterMiddleware.swift
```diff
@@ -3,3 +3,83 @@
...
+                event: event,
+                cid: cid,
+                messageId: messageId,
+                handle: { (channel, message) in
+                    if message.pinned {
+                        channel.pinnedMessages.insert(message)
+                    }
+                },
+                completion: completion
+            )
```
VojtaStavik, line 54: I think ideally this should be done in the payload saving code.

Sources/StreamChat/WebSocketClient/EventMiddlewares/ChannelPinnedMessagesUpdaterMiddleware.swift
```diff
@@ -3,3 +3,83 @@
...
+                event: event,
+                cid: cid,
+                messageId: messageId,
+                handle: { (channel, message) in
+                    if message.pinned {
+                        channel.pinnedMessages.remove(message)
+                    }
+                },
+                completion: completion
+            )
```
VojtaStavik, line 40: This is probably not needed. Ideally, we should approach this as any other message -> the message is still visible, but it's marked deleted.

VojtaStavik (commented):
I really like this PR 👍 I'm not sure we need `ChannelPinnedMessagesUpdaterMiddleware`, seems to me like this should rather be handled by the serialization code. Let's talk in Slack

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
```diff
@@ -944,6 +969,13 @@ public enum ListOrdering {
     case bottomToTop
 }
 
+// Describes the pin expiration
+public enum Pin {
+    case noExpiration
+    case expirationDate(Date)
+    case expirationTime(TimeInterval)
+}
+
```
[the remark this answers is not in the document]
nuno-vieira: I think this simplifies a bit the API by adding this abstraction. Some customers might not know How to use the `Date` API. So with this abstraction, everyone will understand.

Also, we need to provide a way to `pin` and `unpin` a message. We can't just provide a date. Because if `nil` would mean `noExpiration` then how would you not pin a message, or unpin a message? This would require us to have a different property, like `pinned = true`. This is actually what the other platforms are doing but I think it is bad because it makes it possible to make a mistake and write `pinned = false` and `expirationDate = Date()`, which is a wrong state.

With this enum, we can describe 4 states. 
- No pin (`Pin? = nil`)
- Pin without expiration
- Pin with an expiration date
- Pin with the amount of time.

Sources/StreamChat/Models/Message.swift
```diff
@@ -162,6 +171,18 @@ public enum MessageType: String, Codable {
...
+// The pinning information of a message.
+public struct _MessagePinDetails<ExtraData: ExtraDataTypes> {
+    /// Date when the message got pinned
+    var pinnedAt: Date?
+
+    /// The user that pinned the message
+    var pinnedBy: _ChatUser<ExtraData.User>?
+
+    /// Date when the message pin expires. An nil value means that message does not expire
+    var expiresAt: Date?
```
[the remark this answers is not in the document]
nuno-vieira: The backend says they are optional. But I guess that if the message is pinned, these properties can't be optional. I played safe here and let them optional anyways. 

Other approaches:
1. In the parsing I can assume that if I can't parse these values, the message is not pinned, even if it says "pinned = true".
2. If we find "pinned = true" we parse those values and make sure they exist, if not, it fails the parsing and crashes.

Maybe option 1. is the best approach?

Sources/StreamChat/APIClient/Endpoints/Payloads/MessagePayloads.swift
```diff
@@ -64,6 +68,11 @@ class MessagePayload<ExtraData: ExtraDataTypes>: Decodable {
     let attachments: [AttachmentPayload]
     let isSilent: Bool
 
+    var pinned: Bool
```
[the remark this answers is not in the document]
nuno-vieira: It possible according to the Backend API, but I would say it would be a bug. Better to confirm this tho.

Sources/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -191,12 +196,35 @@ extension MessageDTO {
...
+            pinExpires = pinnedDate.addingTimeInterval(time)
+        }
+    }
+
+    func unpin() {
+        pinned = false
+        pinnedAt = nil
+        pinnedBy = nil
+        pinExpires = nil
+    }
```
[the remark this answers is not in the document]
nuno-vieira: Thanks, will do 👍

Sources/StreamChat/WebSocketClient/EventMiddlewares/ChannelPinnedMessagesUpdaterMiddleware.swift
nuno-vieira: Yes makes totally sense 👍

Tests/StreamChatTests/MockEnpointResponses/Events/Notification/NotificationMessageNew.json
```diff
@@ -148,7 +148,11 @@
         "mentioned_users" : [
             
         ],
-        "html" : "<p>Hi<\/p>\n"
+        "html" : "<p>Hi<\/p>\n",
+        "pinned": false,
+        "pinned_at": null,
+        "pin_expires": null,
+        "pinned_by": null
```
VojtaStavik, line 155: Thank you for updating all payloads! nit: Ideally these changes would be in a separate commit so it's easier to review.

Tests/StreamChatTests/MockEnpointResponses/Channel.json
```diff
@@ -1191,6 +1191,59 @@
...
+        ],
+        "updated_at" : "2020-06-09T08:10:40.800912Z",
+        "reply_count" : 0,
+        "user" : {
+          "id" : "broken-waterfall-5",
+          "banned" : false,
+          "unread_channels" : 0,
+          "extraData" : {
+            "name" : "Tester"
+          },
```
VojtaStavik, line 1224: Where did you get these data from? This is not the way we save `extraData` :)

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
```diff
@@ -941,6 +944,21 @@ public enum ListOrdering {
     case bottomToTop
 }
 
+// Describes the pinning expiration
+public struct Pin: Equatable {
```
VojtaStavik, line 973: Can we rename this to `Pinning`? Also, it would be great to add documentation for public APIs

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
```diff
@@ -735,6 +735,31 @@ public extension _ChatChannelController {
...
+    ///   - messageId: The message identifier.
+    ///   - completion: A completion block with an error if the request was failed.
+    func unpinMessage(messageId: MessageId, completion: ((Error?) -> Void)? = nil) {
+        updater.unpinMessage(messageId: messageId) { result in
+            self.callback {
+                completion?(result)
+            }
+        }
+    }
+
```
VojtaStavik, line 762: This modifies the messages so shouldn't this rather be on `MessageController`?

Sources/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -375,6 +375,24 @@ extension NSManagedObjectContext: MessageDatabaseSession {
     func delete(message: MessageDTO) {
         delete(message)
     }
+
+    func pin(message: MessageDTO, _ pin: Pin) throws {
```
VojtaStavik, line 403: Why is it `_ pin` and not `pin`? 🤷‍♂️ :) 

I think with the adjusted naming it will work nicely `func pin(message: MessageDTO, pinning: MessagePinning)`

Sources/StreamChat/Controllers/ChannelController/ChannelController_Tests.swift
```diff
@@ -594,8 +594,8 @@ class ChannelController_Tests: StressTestCase {
         }
         XCTAssertNil(error)
         let channel: ChatChannel = client.databaseContainer.viewContext.channel(cid: channelId)!.asModel()
-        assert(channel.latestMessages.count == 1)
-        let message: ChatMessage = channel.latestMessages.first!
+        XCTAssertEqual(channel.latestMessages.count, 1)
+        let message: ChatMessage = try XCTUnwrap(channel.latestMessages.first)
```
VojtaStavik, line 598: The semantic difference between `XCTAssert` and `assert` is that by using `assert` you're saying - this is how it should be and it's not the thing we're testing here. However, I do understand the original implementation wasn't the best because it started crashing once you introduced pinned messages. Thanks for improving 👍

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
```diff
@@ -20,6 +20,10 @@ class ChannelDTO: NSManagedObject {
     @NSManaged var updatedAt: Date
     @NSManaged var lastMessageAt: Date?
 
+    // The oldest message in the current batch of messages. This property only lives locally, and it is useful
+    // to filter out older pinned messages that do not belong to the current batch of messages in the channel.
+    @NSManaged var oldestMessageAt: Date
+
```
VojtaStavik, line 26: `// The oldest message in the current batch of messages.` is not correct, it's the oldest message of the channel we have locally coming from a regular channel query.

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
```diff
@@ -124,6 +128,10 @@ extension NSManagedObjectContext {
         dto.lastMessageAt = payload.lastMessageAt
         dto.memberCount = Int64(clamping: payload.memberCount)
 
+        if let oldestMessageAt = dto.messages.map(\.createdAt).min() {
+            dto.oldestMessageAt = oldestMessageAt
+        }
```
VojtaStavik, line 133: We should override only if `dto.oldestMessageAt < oldestMessageAt`. This would be good to test, too, btw.

Sources/StreamChat/Database/DTOs/ChannelDTO_Tests.swift
```diff
@@ -7,6 +7,8 @@ import XCTest
 
 class ChannelDTO_Tests: XCTestCase {
     var database: DatabaseContainer!
+
+    static let channelCreatedDate = "2019-12-12T15:33:46.488935Z".toDate()
```
VojtaStavik, line 11: To make this more fun, you can make this `.unique`, too. It will be unique for the duration of the tests, but it will change with every test suite run to provide additional variability in the tests.

Sources/StreamChat/Database/DTOs/ChannelDTO_Tests.swift
```diff
@@ -209,6 +211,72 @@ class ChannelDTO_Tests: XCTestCase {
         XCTAssertEqual(channel?.latestMessages.count, 10)
     }
 
+    func test_channelPayload_pinnedMessagesOlderThanOldestMessageAtAreIgnored() throws {
+        // Save a channel payload with 100 messages
```
VojtaStavik, line 215: that `100` there is probably a copy/paste mistake? Also, why even use 20? Wouldn't just 1+1 be better and more obvious? :)

Sources/StreamChat/Database/DTOs/ChannelDTO_Tests.swift
```diff
@@ -209,6 +211,72 @@ class ChannelDTO_Tests: XCTestCase {
...
+        let channelId: ChannelId = .unique
+        let oldPinnedMessage: MessagePayload<NoExtraData> = MessagePayload(
+            id: .unique,
+            type: .regular,
+            user: dummyUser,
+            createdAt: "2018-12-12T15:33:46.488935Z".toDate(),
```
VojtaStavik, line 221: nit: Why not just use some more readable version of the date? The actual value doesn't matter, it just has to be older than a specific date.

Sources/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -117,11 +117,18 @@ class MessageDTO: NSManagedObject {
             .init(format: "createdAt > channel.truncatedAt")
         ])
 
+        // Some pinned messages might be in the local database, but should not be fetched
+        // if they are not in the current batch of messages.
```
VojtaStavik, line 121: Same here - the "current batch of messages" is imo a little bit confusing wording.

Sources/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -331,15 +338,32 @@ extension NSManagedObjectContext: MessageDatabaseSession {
...
+        var channelDTO: ChannelDTO?
+
         if let channelPayload = payload.channel {
-            let channelDTO = try saveChannel(payload: channelPayload, query: nil)
-            dto.channel = channelDTO
+            channelDTO = try saveChannel(payload: channelPayload, query: nil)
         } else if let cid = cid {
-            let channelDTO = ChannelDTO.loadOrCreate(cid: cid, context: self)
-            dto.channel = channelDTO
+            channelDTO = ChannelDTO.loadOrCreate(cid: cid, context: self)
```
VojtaStavik, line 346: nit: We talked about the fact that this should be just `load`

Sources/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -331,15 +338,32 @@ extension NSManagedObjectContext: MessageDatabaseSession {
...
+            } else {
+                // If message is not pinned, or was unpinned, should be removed from pinnedMessages
+                channelDTO.pinnedMessages.remove(dto)
+                // If the current message dto is pinned, should not be included to oldestMessagetAt
+                messagesCreatedAt.append(dto.createdAt)
+            }
+            if let oldestMessageAt = messagesCreatedAt.min() {
+                channelDTO.oldestMessageAt = oldestMessageAt
+            }
+            dto.channel = channelDTO
```
VojtaStavik, line 365: I'm confused by this. I think we shouldn't touch the `channelDTO.oldestMessageAt` here at all. That's the main point of it - to update it only if it comes from the regular channel query. Implemented like this, if a message that's not presented locally is updated, it will adjust `channelDTO.oldestMessageAt` and we have exactly the problem we're trying to work around. What we do in `ChannelDTO` is enough IMO.

Sources/StreamChat/Database/DTOs/MessageDTO_Tests.swift
```diff
@@ -242,7 +242,59 @@ class MessageDTO_Tests: XCTestCase {
...
+                // Save the message
+                let messageDTO = try! session.saveMessage(payload: payload, for: channelId)
+                completion((channelDTO, messageDTO))
+            }
+        }
+
+        XCTAssertEqual(channelDTO.oldestMessageAt, payload.createdAt)
+        XCTAssertFalse(channelDTO.pinnedMessages.contains(messageDTO))
+    }
+
```
VojtaStavik, line 297: Yep, I think this is actually behavior we don't want

VojtaStavik (changes requested):
This PR is full of interesting and complex solutions. Let's talk in Slack about the details.

Sources/StreamChat/APIClient/Endpoints/Payloads/ChannelListPayload_Tests.swift
```diff
@@ -49,6 +53,7 @@ class ChannelPayload_Tests: XCTestCase {
         XCTAssert(firstMessage.reactionScores.isEmpty)
         XCTAssertEqual(firstMessage.replyCount, 0)
         XCTAssertFalse(firstMessage.isSilent)
+        XCTAssertEqual(firstPinnedMessage.id, "broken-waterfall-5-7aede36b-b89f-4f45-baff-c40c7c1875d9")
```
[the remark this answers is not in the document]
nuno-vieira: 🤦 An assertNotNil here is enough 😄

Tests/StreamChatTests/MockEnpointResponses/Channel.json
nuno-vieira: I only changed data related to Pin information, so this extradata was already in this json 🤔

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
nuno-vieira: Yes I added the docs to the functions, but forgot to add documentation directly to Pin 👍

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
nuno-vieira: In the documentation all other platforms pin and unpin the message in the channelController. So this makes it consistent with the other platforms, and also kinda makes sense, because we pinning a message in a channel, and not pinning a message in all channels, don't know if this last sentence made sense 😄

Sources/StreamChat/Database/DTOs/MessageDTO.swift
nuno-vieira: Yeah the hidden labelling was so it was not so repetitive. `pin(message: someMessage, .noExpirationDate)` instead of repeating `pin` again. But `MessagePinning` sounds good to me 👍

Sources/StreamChat/Controllers/ChannelController/ChannelController_Tests.swift
nuno-vieira: Yes that was exactly why I changed this, but I understood why it was there in the first place. But it was kinda annoying to always crash when was trying to debug, so that is why I changed 👍

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
nuno-vieira: Makes sense, thanks 👍

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
nuno-vieira: Why is that check necessary? 🤔

Sources/StreamChat/Database/DTOs/ChannelDTO_Tests.swift
nuno-vieira: Yes copy paste issue 😄 Yeah you right, 1+1 will be easier to understand what we are testing here, will change it 👍

Tests/StreamChatTests/MockEnpointResponses/Channel.json
nuno-vieira: I extracted a message, from another payload, so that payload has the extraData like this, probably it is a old payload. I'm going to remove it from here

nuno-vieira: We have it in a lot of places tho :D Do you want me to remove them all?
image

VojtaStavik: @nuno-vieira if you can remove it please do, it will only cause confusion in the future.

Sources/StreamChat/APIClient/Endpoints/Payloads/ChannelListPayload_Tests.swift
```diff
@@ -49,6 +53,7 @@ class ChannelPayload_Tests: XCTestCase {
         XCTAssert(firstMessage.reactionScores.isEmpty)
         XCTAssertEqual(firstMessage.replyCount, 0)
         XCTAssertFalse(firstMessage.isSilent)
+        XCTAssertNotNil(firstPinnedMessage)
```
VojtaStavik, line 56: Maybe we can replace
```swift
XCTAssertEqual(payload.pinnedMessages.count, 1)
let firstPinnedMessage = payload.pinnedMessages
    .first(where: { $0.id == "broken-waterfall-5-7aede36b-b89f-4f45-baff-c40c7c1875d9" })
XCTAssertNotNil(firstPinnedMessage)
```

with just

```swift
XCTAssertEqual(
    payload.pinnedMessages.map { $0.id }, 
    ["broken-waterfall-5-7aede36b-b89f-4f45-baff-c40c7c1875d9"]
)
```
??

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
```diff
@@ -153,6 +154,10 @@ extension NSManagedObjectContext {
         let dto = try saveChannel(payload: payload.channel, query: query)
         
         try payload.messages.forEach { _ = try saveMessage(payload: $0, for: payload.channel.cid) }
+        try payload.pinnedMessages.forEach {
+            let message = try saveMessage(payload: $0, for: payload.channel.cid)
+            dto.pinnedMessages.insert(message)
```
VojtaStavik, line 173: Is this needed? This is already happening inside `saveMessage`

Sources/StreamChat/Models/Channel.swift
```diff
@@ -96,6 +96,13 @@ public struct _ChatChannel<ExtraData: ExtraDataTypes> {
     /// and using a `ChatChannelController` for this channel id.
     ///
     public let latestMessages: [_ChatMessage<ExtraData>]
+
+    /// Pinned messages present on the channel.
+    ///
+    /// This field contains only the pinned messages of the channel. You can get all existing messages in the channel by creating
+    /// and using a `ChatChannelController` for this channel id.
+    ///
```
VojtaStavik, line 104: Now when we have it, it would make sense to actually use the `@Cached` mechanism for `pinnedMessages`, too.

Sources/StreamChat/Database/DTOs/ChannelDTO_Tests.swift
```diff
@@ -166,6 +175,19 @@ class ChannelDTO_Tests: XCTestCase {
...
+        // Save a channel payload with 100 messages
+        let channelId: ChannelId = .unique
+        let payload = dummyPayload(with: channelId, numberOfMessages: 100)
+
+        try database.writeSynchronously { session in
+            try session.saveChannel(payload: payload)
+        }
+
+        let channel: ChatChannel? = database.viewContext.channel(cid: channelId)?.asModel()
+        XCTAssertEqual(channel?.pinnedMessages.count, 1)
```
VojtaStavik, line 188: For the test here it's not clear why `pinnedMessages.count` should be 1

Sources/StreamChat/Database/DTOs/ChannelDTO_Tests.swift
```diff
@@ -7,6 +7,8 @@ import XCTest
 
 class ChannelDTO_Tests: XCTestCase {
     var database: DatabaseContainer!
+
+    static let channelCreatedDate = Date.unique
```
VojtaStavik, line 11: Can you please add a comment explaining why this is needed and what problem it solves?

Sources/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -326,15 +326,24 @@ extension NSManagedObjectContext: MessageDatabaseSession {
...
         }
+
+        if let channelDTO = channelDTO {
+            if dto.pinned {
+                channelDTO.pinnedMessages.insert(dto)
+            } else {
+                channelDTO.pinnedMessages.remove(dto)
+            }
+            dto.channel = channelDTO
+        }
```
VojtaStavik, line 358: if `channelDTO` is `nil` we silently fail but in fact, it's a very serious issue.

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
```diff
@@ -690,15 +690,17 @@ public extension _ChatChannelController {
     ///
     /// - Parameters:
     ///   - text: Text of the message.
-    ///   - extraData: Additional extra data of the message object.
+    ///   - pinning: Pins the new message. Nil if should not be pinned.
```
VojtaStavik, line 693: nit: `Nil` -> `` `nil` ``

Sources/StreamChat/Controllers/MessageController/MessageController.swift
```diff
@@ -193,6 +193,7 @@ public extension _ChatMessageController {
     ///
     /// - Parameters:
     ///   - text: Text of the message.
+    ///   - pinning: Pins the new message. Nil if should not be pinned.
```
VojtaStavik, line 196: same here

Sources/StreamChat/Models/MessagePinning.swift
```diff
@@ -0,0 +1,31 @@
...
+    public static func expirationDate(_ date: Date) -> Self {
+        .init(expirationDate: date)
+    }
+
+    /// Pins a message.
+    /// - Parameter time: The amount of seconds the message will be pinned.
+    public static func expirationTime(_ time: TimeInterval) -> Self {
+        .init(expirationDate: Date().addingTimeInterval(time))
+    }
+}
```
VojtaStavik, line 31: Nice 🎉

VojtaStavik (approved):
Looks great 👍 Just some minor comments to test + the issue with overriding `oldestMessageAt` and it's good to 🚢 🎉

VojtaStavik (approved):
All right! Let's 🚢 this beauty 🎉

VojtaStavik:
@nuno-vieira before merging, please add a CHANGELOG entry for this

PR #898, opened 2021-03-18
[CIS-729] Remove setting of TAMIC inside pin methods

DominikBucher12:
# What this PR do:
- Removes TAMIC from pin view extensions and replaces the 
# How to test it
- Probably CI can take care of UI Tests, if they pass, we are fine :) 
# Where you can start
- `UIView+Extensions.swift`

### DIFF
--- a/Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView+SwiftUI_Tests.swift
+++ b/Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView+SwiftUI_Tests.swift
@@ -52,8 +52,6 @@ class ChatChannelListItemView_SwiftUI_Tests: XCTestCase {
 
 private extension UIView {
     func addWidthConstraint() {
-        translatesAutoresizingMaskIntoConstraints = false
-
         NSLayoutConstraint.activate([
             widthAnchor.constraint(equalToConstant: 350)
         ])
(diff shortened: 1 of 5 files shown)


Sources/StreamChatUI/CommonViews/AvatarView/ChatChannelAvatarView+SwiftUI.swift
```diff
@@ -39,6 +39,7 @@ extension _ChatChannelAvatarView {
         }
 
         override public func setUpLayout() {
+            hostingController?.view.translatesAutoresizingMaskIntoConstraints = false
             embed(hostingController!.view)
```
VojtaStavik, line 43: I searched our codabase for `embed(hostingController!.view)` and it found 3 cases. Why do we change it just for this one?

VojtaStavik (approved):
LGTM, however, just a simple text search found places where the value is not set correctly. Can you please double-check and make sure we **really** set the value where needed?

Sources/StreamChatUI/CommonViews/AvatarView/ChatChannelAvatarView+SwiftUI.swift
DominikBucher12: Maybe the others are not tested yet :) I will check, thank you. :)

DominikBucher12: Wow very thanks for this inside, I actually overlooked that `TAMIC = false` is being set in tests for SwiftUI views, which was part of strange fails when I was trying to figure out. `HostingVC.view` doesn't need to have this property set to false... I deleted the setting from tests and now we should be fine. :)

b-onc (approved):
:shipit:

PR #900, opened 2021-03-19
[CIS-659] Resize images to fit UIImageView bounds

VojtaStavik:
If it's possible and the size of the `UIImageView` is known, we automatically resize the image to fit the image view. This resized image is also cached automatically by Nuke.

### DIFF
--- a/Sources/StreamChatUI/Utils/UIImageView+Extensions.swift
+++ b/Sources/StreamChatUI/Utils/UIImageView+Extensions.swift
@@ -6,7 +6,7 @@ import Nuke
 import UIKit
 
 extension UIImageView {
-    func loadImage(from url: URL?, placeholder: UIImage? = nil) {
+    func loadImage(from url: URL?, placeholder: UIImage? = nil, resizeAutomatically: Bool = true) {
         guard !SystemEnvironment.isTests else {
             // When running tests, we load the images synchronously
             image = url.flatMap { UIImage(data: try! Data(contentsOf: $0)) }
@@ -17,6 +17,12 @@ extension UIImageView {
         currentImageLoadingTask?.cancel()
 
         guard let url = url else { image = nil; return }
+
+        let preprocessors: [ImageProcessing] = resizeAutomatically && bounds.size != .zero
+            ? [ImageProcessors.Resize(size: bounds.size, contentMode: .aspectFill, crop: true)]
+            : []
+ 
+        let request = ImageRequest(url: url, processors: preprocessors)
         let options = ImageLoadingOptions(placeholder: placeholder)
 
         currentImageLoadingTask = Nuke.loadImage(with: request, options: options, into: self)
--- a/Sources/StreamChatUI/Utils/UIImageView+Extensions.swift
+++ b/Sources/StreamChatUI/Utils/UIImageView+Extensions.swift
@@ -7,6 +7,11 @@ import UIKit
 
 extension UIImageView {
     func loadImage(from url: URL?, placeholder: UIImage? = nil) {
+        guard !SystemEnvironment.isTests else {
+            // When running tests, we load the images synchronously
+            image = url.flatMap { UIImage(data: try! Data(contentsOf: $0)) }
+            return
+        }
         guard let url = url else { image = nil; return }
         let options = ImageLoadingOptions(placeholder: placeholder)
         Nuke.loadImage(with: url, options: options, into: self)
--- a/Sources/StreamChatUI/Utils/UIImageView+Extensions.swift
+++ b/Sources/StreamChatUI/Utils/UIImageView+Extensions.swift
@@ -12,8 +12,22 @@ extension UIImageView {
             image = url.flatMap { UIImage(data: try! Data(contentsOf: $0)) }
             return
         }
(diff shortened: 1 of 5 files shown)


Sources/StreamChatUI/Utils/UIImageView+Extensions.swift
```diff
@@ -7,8 +7,20 @@ import UIKit
...
+        currentImageLoadingTask = Nuke.loadImage(with: request, options: options, into: self)
+    }
+}
+
+private extension UIImageView {
+    static var nukeLoadingTaskKey: UInt8 = 0
+
+    var currentImageLoadingTask: ImageTask? {
+        get { objc_getAssociatedObject(self, &Self.nukeLoadingTaskKey) as? ImageTask }
+        set { objc_setAssociatedObject(self, &Self.nukeLoadingTaskKey, newValue, .OBJC_ASSOCIATION_RETAIN) }
```
DominikBucher12, line 31: wow :)

Sources/StreamChatUI/Utils/UIImageView+Extensions.swift
```diff
@@ -6,10 +6,17 @@ import Nuke
...
 extension UIImageView {
-    func loadImage(from url: URL?, placeholder: UIImage? = nil) {
+    func loadImage(from url: URL?, placeholder: UIImage? = nil, resizeAutomatically: Bool = true) {
         // Cancel any previous loading task
         currentImageLoadingTask?.cancel()
+
         guard let url = url else { image = nil; return }
+
+        let preprocessors: [ImageProcessing] = resizeAutomatically && bounds.size != .zero
+            ? [ImageProcessors.Resize(size: bounds.size, contentMode: .aspectFill, crop: true)]
```
DominikBucher12, line 16: 🤔 Shouldn't this also be somehow tested? Because it's out of Nuke scope basically and we have some custom logic in here....  But maybe it's too much :D I ask because this is not implemented in any tests AFAIK, so there is no way to know if this changes something visually . But on the other hand this would be probably more complex (and probably unnecessary work?) to test (Different set of Images, already tested by `Nuke`...)

DominikBucher12 (approved):
LGTM. Just curious question whether we should test this 🤔

Sources/StreamChatUI/Utils/UIImageView+Extensions.swift
DominikBucher12: Or maybe the `UITests` already discovered :trollface:

VojtaStavik: Ideally yes, but I actually did the opposite in 17ece73 to bypass Nuke completely. We need to have the image set synchronously. Previously we pre-loaded the image to Nuke's cache to force it to load synchronusly. With the various sizes now this becomes more complex so I removed it. So yeah, ideally we will have this tested eventually but I'm not sure exactly how to approach this. Also, it's not that big of a deal as long as all image setting logic is funneled via this method.

Sources/StreamChatUI/Utils/UIImageView+Extensions.swift
```diff
@@ -6,9 +6,34 @@ import Nuke
...
+        currentImageLoadingTask = Nuke.loadImage(with: request, options: options, into: self)
+    }
+}
+
+private extension UIImageView {
+    static var nukeLoadingTaskKey: UInt8 = 0
+
+    var currentImageLoadingTask: ImageTask? {
+        get { objc_getAssociatedObject(self, &Self.nukeLoadingTaskKey) as? ImageTask }
+        set { objc_setAssociatedObject(self, &Self.nukeLoadingTaskKey, newValue, .OBJC_ASSOCIATION_RETAIN) }
```
nuno-vieira, line 37: Why do we need this? 🤔

VojtaStavik: To be able to cancel previous tasks. It's not related to resizing. If we don't have this the following can happen:
```
loadImage A -----X-----------------------------------------------------------X
            start loading A                                              image = loaded Image A

cell is reused

loadImage B -----------------------X--------------------------------X----------
                       start loading B            image = loaded Image B
```
If we don't cancel the loading task for A explicitly, what can happen is that it will finish after the loading for another task has finished.

nuno-vieira: Ah nevermind, I see, this is an extension of UIImageView, so it is a way to add a new property to UIImageView, without making a subclass right?

nuno-vieira (approved):
LGTM!

Sources/StreamChatUI/Utils/UIImageView+Extensions.swift
VojtaStavik: Exactly, we don't want to limit the users to use our specific `UIImageView` subclass for this.

nuno-vieira: Yes my question was not about why we needed the cancel, that I understood, my questions was why we needed the swizzling 👍

PR #903, opened 2021-03-22
[CIS-657] Slack clone backbone

fortmarek:

In this PR I am adding a Slack clone app.

Besides that, I have also changed `set` to `public` for properties in `ChatMessageContentView` as there was no other way not to use the default constraints.

I am also wondering if it's OK to make `layoutOptions` and `textContent` `public`? Right now I had to copy & paste it since I don't want to use `super` in `updateContent` in `SlackChatMessageContentView`

Known inconsistencies:
- we should not show avatar and metadata view for each message but only for first in group. This is not trivial to implement at the moment, so it will be added in a subsequent PR


VojtaStavik (approved):
Cool 👍 Let's merge this and we will take care of details later.

PR #906, opened 2021-03-22
[CIS-733] Load expensive CoreData properties lazily

olejnjak:
Improve some `asModel` function performance by loading expensive top level properties lazily. This means that affected model structs are now thread-confined which should be addressed later as it does not make much sense for value types.

Some other changes were necessary:
- `@Cached` wrapper now has `computeValue` closure as `projectedValue` as otherwise it cannot be set from outside as `_cachedWrapper` property is private ➡️ this will not be possible as `__mentionedUsers` is inaccessible because it is private automatically
```
__mentionedUsers.computeValue = { Set(dto.mentionedUsers.map { $0.asModel() }) }
```
- some model inits needed to be reorder to make compiler happy

I consider the semantic change of `_ChatChannel` and `_ChatMessage` to be breaking so it is marked this way in changelog, even though the API remains unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### 🐞 Fixed
 - Fix `ChannelDoesNotExist` error is logged by `UserWatchingEventMiddleware` when channels are fetched for the first time [#893](https://github.com/GetStream/stream-chat-swift/issues/893)
+- Improve model loading performance by lazy loading expensive properties [#906](https://github.com/GetStream/stream-chat-swift/issues/906)
 
 # [3.1.3](https://github.com/GetStream/stream-chat-swift/releases/tag/3.1.3)
 _March 12, 2021_
--- a/Sources/StreamChat/Controllers/ChannelController/ChannelController+SwiftUI_Tests.swift
+++ b/Sources/StreamChat/Controllers/ChannelController/ChannelController+SwiftUI_Tests.swift
@@ -158,15 +158,15 @@ extension _ChatMessage {
             quotedMessageId: nil,
             isSilent: false,
             reactionScores: ["": 1],
-            author: .init(id: .unique),
-            mentionedUsers: [],
+            author: { .init(id: .unique) },
+            mentionedUsers: { [] },
             threadParticipants: [],
-            attachments: [],
-            latestReplies: [],
+            attachments: { [] },
+            latestReplies: { [] },
             localState: nil,
             isFlaggedByCurrentUser: false,
-            latestReactions: [],
-            currentUserReactions: [],
+            latestReactions: { [] },
+            currentUserReactions: { [] },
             isSentByCurrentUser: false
         )
     }
--- a/Sources/StreamChat/Database/DTOs/ChannelDTO.swift
+++ b/Sources/StreamChat/Database/DTOs/ChannelDTO.swift
@@ -244,11 +244,6 @@ extension _ChatChannel {
         
         let context = dto.managedObjectContext!
         
-        // TODO: make messagesLimit a param
-        let latestMessages: [_ChatMessage<ExtraData>] = MessageDTO
-            .load(for: dto.cid, limit: 25, context: context)
-            .map { $0.asModel() }
-        
         let reads: [_ChatChannelRead<ExtraData>] = dto.reads.map { $0.asModel() }
         
         var unreadCount = ChannelUnreadCount.noUnread
@@ -268,6 +263,12 @@ extension _ChatChannel {
                 messages: currentUserChannelRead.unreadMessagesCount,
                 mentionedMessages: mentionedMessagesCount
             )
+        }
+        
+        let fetchMessages: (Int) -> [_ChatMessage<ExtraData>] = {
+            MessageDTO
+                .load(for: dto.cid, limit: $0, context: context)
+                .map { $0.asModel() }
         }
         
         return _ChatChannel(
@@ -294,7 +295,8 @@ extension _ChatChannel {
             cooldownDuration: Int(dto.cooldownDuration),
             extraData: extraData,
 //            invitedMembers: [],
-            latestMessages: latestMessages
+            latestMessages: { fetchMessages(25) }, // TODO: make messagesLimit a param
+            lastMessage: fetchMessages(1).first
         )
     }
 }
--- a/Sources/StreamChat/Database/DTOs/MessageDTO.swift
+++ b/Sources/StreamChat/Database/DTOs/MessageDTO.swift
@@ -415,56 +415,63 @@ private extension _ChatMessage {
         }
         self.extraData = extraData
         
-        author = dto.user.asModel()
-        mentionedUsers = Set(dto.mentionedUsers.map { $0.asModel() })
         threadParticipants = Set(dto.threadParticipants.map(\.id))
-
-        if dto.replies.isEmpty {
-            latestReplies = []
-        } else {
-            latestReplies = MessageDTO
-                .loadReplies(for: dto.id, limit: 5, context: context)
-                .map(_ChatMessage.init)
-        }
         localState = dto.localMessageState
-        
         isFlaggedByCurrentUser = dto.flaggedBy != nil
-
-        if dto.reactions.isEmpty {
-            latestReactions = []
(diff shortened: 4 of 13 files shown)


Sources/StreamChat/Models/Channel.swift
```diff
@@ -97,6 +97,8 @@ public struct _ChatChannel<ExtraData: ExtraDataTypes> {
     ///
     public let latestMessages: [_ChatMessage<ExtraData>]
     
+    public let lastMessage: _ChatMessage<ExtraData>?
+
```
VojtaStavik: This is public API so it needs to be documented.

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
```diff
@@ -294,7 +295,10 @@ extension _ChatChannel {
             cooldownDuration: Int(dto.cooldownDuration),
             extraData: extraData,
 //            invitedMembers: [],
-            latestMessages: latestMessages
+            latestMessages: fetchMessages(25), // TODO: make messagesLimit a param
+            lastMessage: MessageDTO
+                .load(for: dto.cid, limit: 1, context: context)
+                .map { $0.asModel() }.first
```
VojtaStavik: Why you don't use the same `fetchMessages` helper you use for `latestMessages`?

Sources/StreamChat/Models/Channel.swift
```diff
@@ -149,7 +151,7 @@ public struct _ChatChannel<ExtraData: ExtraDataTypes> {
         cooldownDuration: Int = 0,
         extraData: ExtraData.Channel,
 //        invitedMembers: Set<_ChatChannelMember<ExtraData.User>> = [],
-        latestMessages: [_ChatMessage<ExtraData>] = []
+        latestMessages: @escaping @autoclosure (() -> [_ChatMessage<ExtraData>]) = [],
```
VojtaStavik, line 154: I must say I find this API quite confusing for the caller. For the optimization to work correctly, you need to know the fields are loaded lazily and you have to make sure you don't them prematurely. 

So, rather then `@escaping @autoclosure` it's IMO better and much more safer to have `() -> [_ChatMessage<ExtraData>]` explicitly here.

Sources/StreamChat/Utils/Cached.swift
```diff
@@ -10,8 +10,10 @@ import Foundation
 /// to cache it for certain period of time to prevent unnecessary recomputing of the value.
 @propertyWrapper
 class Cached<T> {
+    typealias ComputeValue = (() -> T)
+
```
VojtaStavik, line 14: Maybe a nit but my opinion is that typealiases like this make it easier to write but harder to read. I'd prefer having it the other way around tbh.

For example, when looking at the following samples, the latter is IMO easier to understand.

```swift
    var projectedValue: ComputeValue { }
    @Atomic private var _cached: T?
```

```swift
    var projectedValue: () -> T { }
    @Atomic private var _cached: T?
```

Sources/StreamChat/Controllers/ChannelController/ChannelController+SwiftUI_Tests.swift
```diff
@@ -158,16 +158,19 @@ extension _ChatMessage {
...
         )
+        
+        message.$_author = { .init(id: .unique) }
+        message.$_mentionedUsers = { [] }
+        message.$_attachments = { [] }
+        message.$_latestReplies = { [] }
+        message.$_latestReactions = { [] }
+        message.$_currentUserReactions = { [] }
+        
+        return message
```
VojtaStavik, line 174: Do I understand it correctly by introducing the `@Cahced` wrappers we lost the compiler support for those values? That's a HUGE issue. The initializer was generated, right? We need to fix this and write the initializer manually. It can't stay like this.

Sources/StreamChat/Models/Message.swift
```diff
@@ -84,19 +84,35 @@ public struct _ChatMessage<ExtraData: ExtraDataTypes> {
     public let reactionScores: [MessageReactionType: Int]
     
     /// The user which is the author of the message.
-    public let author: _ChatUser<ExtraData.User>
+    ///
+    /// Author is loaded lazily on demand to maintain high performance, so this property is not thread safe
```
VojtaStavik, line 88: Great that you put those comments there. Can we just tweak them a bit? Something like:
```
    /// The user who is the author of the message.
    ///
    /// - Important: The `author` property is loaded and evaluated lazily to maintain high performance.
    ///
```

There's no need to mention the potential thread issue, we will fix it shortly. It's just a temporary thing.

CHANGELOG.md
```diff
@@ -8,6 +8,19 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
+    - `_ChatChannel.latestMessages`
+    - `_ChatMessage.author`
+    - `_ChatMessage.mentionedUsers`
+    - `_ChatMessage.attachments`
+    - `_ChatMessage.latestReplies`
+    - `_ChatMessage.latestReactions`
+    - `_ChatMessage.currentUserReactions`
+
+
+
```
VojtaStavik, line 23: This just a temporary downside, no need to mention it in the changelog. However, you should mention there that the performance is improved due to some properties being loaded lazily :)

VojtaStavik (changes requested):
Great 🎉 Before merging this, let's address the regressions to our internal API.

Sources/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -415,56 +415,63 @@ private extension _ChatMessage {
...
-        quotedMessageId = dto.quotedMessage.map(\.id)
+        }
+        
+        if dto.reactions.isEmpty {
+            $_latestReactions = { [] }
+        } else {
+            $_latestReactions = {
+                Set(
+                    MessageReactionDTO
+                        .loadLatestReactions(for: dto.id, limit: 5, context: context)
```
DominikBucher12, line 470: Hardcoded limit is okay? Same for line 459.

DominikBucher12 (approved):
At first, this spinned my head around.  After seeing the bigger picture of `@Cached` property  wrapper and the implementation I see this is awesome 👍

Sources/StreamChat/Database/DTOs/MessageDTO.swift
olejnjak: Not sure, just moved the original code to make compiler happy. In `ChannelDTO` there is originally a TODO for that.

Sources/StreamChat/Models/Channel.swift
```diff
@@ -95,8 +95,11 @@ public struct _ChatChannel<ExtraData: ExtraDataTypes> {
     /// This field contains only the latest messages of the channel. You can get all existing messages in the channel by creating
     /// and using a `ChatChannelController` for this channel id.
     ///
-    public let latestMessages: [_ChatMessage<ExtraData>]
+    /// - Important: The `latestMessages` property is loaded and evaluated lazily to maintain high performance.
+    public var latestMessages: [_ChatMessage<ExtraData>] { _latestMessages }
+    @Cached private var _latestMessages: [_ChatMessage<ExtraData>]
     
+    /// Last message present on the channel
     public let lastMessage: _ChatMessage<ExtraData>?
```
VojtaStavik: nit 🇬🇧 
```
/// The newest message of the channel. `nil` if there are no messages in the channel.
```

Sources/StreamChat/Models/Message.swift
```diff
@@ -114,12 +130,76 @@ public struct _ChatMessage<ExtraData: ExtraDataTypes> {
...
-    public let currentUserReactions: Set<_ChatMessageReaction<ExtraData>>
+    ///
+    /// - Important: The `currentUserReactions` property is loaded and evaluated lazily to maintain high performance.
+    public var currentUserReactions: Set<_ChatMessageReaction<ExtraData>> { _currentUserReactions }
+    
+    @Cached internal var _currentUserReactions: Set<_ChatMessageReaction<ExtraData>>
     
     public let isSentByCurrentUser: Bool
+    
+    public init(
```
VojtaStavik, line 147: The initializer should be `internal`

VojtaStavik (approved):
Great 👍 Just a couple of nit comments and then please merge 🎉

PR #908, opened 2021-03-24
Create UIViewRepresentable wrapper

fortmarek:

This PR adds `SwiftUIRepresentable` protocol which we can add to types we want users to be able to leverage in SwiftUI - so far, I have added it to some views used in `ChatChannelList`, but adding it to more types is just a matter of adding the conformance itself.

Besides that, there is also `SwiftUIViewRepresentable` that is a concrete type conforming to `UIViewRepresentable` and is what users will use in their SwiftUI views.

To create the wrapper for eg. `ChatChannelAvatarView`, in `body` of your SwiftUI view you can call `uiConfig.channelList.itemSubviews.avatarView.asView(content)` where `content` corresponds to the content of `ChatChannelAvatarView`


Sources/StreamChatUI/CommonViews/SwiftUIViewRepresentable_Tests.swift
```diff
@@ -0,0 +1,40 @@
...
+        let channel = ChatChannel.mock(cid: .unique)
+        
+        var config = UIConfig()
+        config.channelList.itemSubviews.avatarView = CustomAvatarView.self
+        let view = CustomView(content: (channel, .unique))
+            .environmentObject(config.asObservableObject)
+        
+        AssertSnapshot(view)
+    }
+}
```
VojtaStavik, line 40: tbh I don't think this is needed here. We will probably add snapshot tests for each view.

VojtaStavik (approved):
LGTM 👍 I think we don't need the generic tests here, so feel free to remove them.

Sources/StreamChatUI/CommonViews/SwiftUIViewRepresentable_Tests.swift
fortmarek: I have already added `SwiftUIRepresentable` conformances to a bunch of views - should I add tests for those then? I felt that adding a snapshot test for each view we add this conformance to might be a little bit unnecessary as what we will essentially test is the wrapper itself over and over again - but maybe it might help unravel some bug in the future

VojtaStavik: Exactly, I'm sure we will find some behavior we won't expect :) Let's test the result not the tooling

PR #909, opened 2021-03-24
ChatVC refactoring

LukasHromadnik:
# Submit a pull request

## CLA

- [x] I have signed the [Stream CLA](https://docs.google.com/forms/d/e/1FAIpQLScFKsKkAJI7mhCr7K9rEIOpqIDThrWxuvxnwUq2XkHyG154vQ/viewform) (required).
- [ ] The code changes follow best practices
- [ ] Code changes are tested (add some information if not applicable)

In this PR I tried to refactor the current state of `ChatVC` and its children. There were `ChatChannelVC` and `ChatThreadVC`, both controllers have the same parent `ChatVC`.

I removed the base class `ChatVC` and make those child controllers separate. With this approach it's easier for the end user to customize the UI.

Some of the shared logic was moved to separate entities so the final implementation of VCs contains only "boring" code.


Sources/StreamChatUI/UIConfig.swift
```diff
@@ -16,6 +16,8 @@ public struct _UIConfig<ExtraData: ExtraDataTypes> {
 
     /// An avatar view with an online indicator.
     public var presenceAvatarView: _ChatPresenceAvatarView<ExtraData>.Type = _ChatPresenceAvatarView<ExtraData>.self
+    
+    public var titleView: _ChatMessageListTitleView<ExtraData>.Type = _ChatMessageListTitleView<ExtraData>.self
```
VojtaStavik, line 20: This should ideally do inside the `MessageListUI` section

Sources/StreamChatUI/ChatMessageList/ChatMessageListTitleView.swift
```diff
@@ -5,43 +5,41 @@
...
+    }
+    
+    open override func setUpLayout() {
+        super.setUpLayout()
+        
         let titleView = UIStackView(arrangedSubviews: [titleLabel, subtitleLabel])
         titleView.axis = .vertical
         addSubview(titleView)
         titleView.translatesAutoresizingMaskIntoConstraints = false
         titleView.pin(to: self)
```
VojtaStavik: I think it would be beneficial to have the stack view as a variable so it's easier to tweak it when you subclass this.

LukasHromadnik: 👍

Sources/StreamChatUI/UIConfig.swift
LukasHromadnik: 👍

LukasHromadnik:
Closing draft, now we need to split this into separate PRs

PR #910, opened 2021-03-24
[CIS-736] Improve Demo App experience 🤔

DominikBucher12:
Improves overal experience of Demo App:

1. Adds loading indicator to channel loading
2. Adds Haptic feedback on longpress of message and improves animation
3. New chat screen add button now supports dark mode
4. Advanced options menu is now presented modally 
5. TableView on user selection is now bouncing

### DIFF
--- a/Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
+++ b/Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
@@ -16,6 +16,14 @@ open class _ChatChannelListVC<ExtraData: ExtraDataTypes>: _ViewController,
     SwipeableViewDelegate {
     /// The `ChatChannelListController` instance that provides channels data.
     public var controller: _ChatChannelListController<ExtraData>!
+
+    open private(set) lazy var loadingIndicator: UIActivityIndicatorView = {
+        if #available(iOS 13.0, *) {
+            return UIActivityIndicatorView(style: .large).withoutAutoresizingMaskConstraints
+        } else {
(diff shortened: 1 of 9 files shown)


VojtaStavik (approved):
Works good enough as quick polishing fixes 👍 Thank you

nuno-vieira (approved):
LGTM 👍

Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
```diff
@@ -244,3 +254,13 @@ extension _ChatChannelListVC: _ChatChannelListControllerDelegate {
...
     }
 }
+
+extension _ChatChannelListVC: DataControllerStateDelegate {
+    public func controller(_ controller: DataController, didChangeState state: DataController.State) {
+        if self.controller.channels.isEmpty {
+            loadingIndicator.startAnimating()
+        } else {
+            loadingIndicator.stopAnimating()
+        }
```
VojtaStavik, line 264: We need to check the actual state here. Like this will just keep spinning if there are no channels. We have the following cases:
```
/// The controller is created but no data fetched.
case initialized
/// The controllers already fetched local data if any.
case localDataFetched
/// The controller failed to fetch local data.
case localDataFetchFailed(ClientError)
/// The controller fetched remote data.
case remoteDataFetched
/// The controller failed to fetch remote data.
case remoteDataFetchFailed(ClientError)
```
The first two should cause the indicator to be visible + the `isEmpty` check, the rest should hide the indicator.

VojtaStavik (changes requested):
The indicator logic is missing/incorrect.

VojtaStavik (approved):

PR #911, opened 2021-03-24
[CIS-727] Channel members & watchers limit

olejnjak:
Create `lastActiveMembers` and `lastActiveWatchers` properties on `Channel` which use only a subset of members/watchers.

Currently I used some arbitrary number (25) for it, but generally I don't like it being hard-coded but I've seen it in some other files. I think good place for it will be `ChatClientConfig.channel` but I see it currently commented out.

Also I was thinking about implementing those properties as CoreData fetched properties but was unable to implement limit of how many items is returned and I thought that doing it partially in DB and partially in code would be confusing.


Sources/StreamChat/Models/Channel.swift
```diff
@@ -63,7 +63,16 @@ public struct _ChatChannel<ExtraData: ExtraDataTypes> {
     /// - Important: This list doesn't have to contain all members of the channel. To access the full list of members, create
     /// a `ChatChannelController` for this channel and use it to query all channel members.
     ///
-    public let cachedMembers: Set<_ChatChannelMember<ExtraData.User>>
+    @available(*, renamed: "lastActiveMembers")
+    public var cachedMembers: Set<_ChatChannelMember<ExtraData.User>> { lastActiveMembers }
```
VojtaStavik, line 67: Can you please move this to some kind of extension that we can then easily delete once we decide to drop this?

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
```diff
@@ -226,7 +226,10 @@ extension ChannelDTO {
 extension _ChatChannel {
     /// Create a ChannelModel struct from its DTO
     fileprivate static func create(fromDTO dto: ChannelDTO) -> _ChatChannel {
-        let members: [_ChatChannelMember<ExtraData.User>] = dto.members.map { $0.asModel() }
+        let memberCompare: (MemberDTO, MemberDTO) -> Bool = { lhs, rhs in
+            lhs.user.lastActivityAt ?? Date(timeIntervalSince1970: 0) < rhs.user.lastActivityAt ?? Date(timeIntervalSince1970: 0)
+        }
+        let members: [_ChatChannelMember<ExtraData.User>] = dto.members.sorted(by: memberCompare).prefix(25).map { $0.asModel() }
```
VojtaStavik, line 233: The reason why we didn't have this parametrized was that we could find a simple way how to pass the value from the config here. But then it came to me and of course, Apple didn't let us down! https://developer.apple.com/documentation/coredata/nsmanagedobjectcontext/1506740-userinfo

This will probably have to set in `DatabaseContainer` based on the values provided by `ChatCient`.

Sources/StreamChat/Models/Channel.swift
```diff
@@ -63,7 +63,16 @@ public struct _ChatChannel<ExtraData: ExtraDataTypes> {
...
+    
+    /// A list of locally cached last active members objects.
+    ///
+    /// - Important: This list doesn't have to contain all members of the channel. To access the full list of members, create
+    /// a `ChatChannelController` for this channel and use it to query all channel members.
+    ///
+    /// - Note: This property will contain no more than 25 members
+    public let lastActiveMembers: Set<_ChatChannelMember<ExtraData.User>>
```
VojtaStavik, line 75: This should idealy be `@Cached`

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
VojtaStavik, line 233: Are you sure about this?
```
dto.members.sorted(by: memberCompare).prefix(25)
```
There can be thousands of members in the channel. I don't think it's a good idea to iterate over all of them. Maybe it would be better to create a fetch request for this? You can also specify the sorting there directly.

VojtaStavik (changes requested):
I commented only on the "members" commit but the same things apply to watchers, too.

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
olejnjak: Well that's why I've been thinking about fetched property, which is basically a fetch request...and then prefixing that, do you have any preference?

VojtaStavik: Solely for the purpose of having better control over it, I think a fetch request is better for this

Sources/StreamChat/Deprecations.swift
```diff
@@ -0,0 +1,24 @@
+//
+// Copyright © 2021 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+public extension Channel {
```
VojtaStavik, line 7: This has to be `ChatChannel_`

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
```diff
@@ -271,6 +270,18 @@ extension _ChatChannel {
...
+                .loadLastActiveWatchers(for: cid, context: context)
+                .map { $0.asModel() }
+        }
+        
+        let fetchMembers: () -> [_ChatChannelMember<ExtraData.User>] = {
+            MemberDTO
+                .loadLastActiveMembers(in: cid, context: context)
+                .map { $0.asModel() }
+        }
+
```
VojtaStavik, line 284: We should capture the behavior of those two variables in tests - does it return the correct entities in the correct order with the correct limit.

VojtaStavik (approved):
I already pushed my review changes directly to the branch

Sources/StreamChat/Config/ChatClientConfig.swift
```diff
@@ -83,20 +83,19 @@ extension ChatClientConfig {
...
-//
+extension ChatClientConfig {
+    /// `ChatChannel` specific settings.
+    public struct Channel {
+        /// Limit how many watchers are returned in `_ChatChannel.lastActiveWatchers`
+        public var lastActiveWatchersLimit = 25
+        /// Limit how many members are returned in `_ChatChannel.lastActiveMembers`
+        public var lastActiveMembersLimit = 25
+    }
+
```
b-onc: I think these 2 should live in `ChannelController` as params, maybe `ChannelQuery` even (they already are present in `ChannelQuery`

Sources/StreamChat/Database/DTOs/MemberModelDTO.swift
```diff
@@ -76,6 +76,14 @@ extension MemberDTO {
         new.id = Self.createId(userId: id, channeldId: channelId)
         return new
     }
+    
+    static func loadLastActiveMembers(cid: ChannelId, context: NSManagedObjectContext) -> [MemberDTO] {
+        let request = NSFetchRequest<MemberDTO>(entityName: MemberDTO.entityName)
+        request.predicate = NSPredicate(format: "ANY channel.cid == %@", cid.rawValue)
+        request.sortDescriptors = [ChannelMemberListSortingKey.lastActiveSortDescriptor]
+        request.fetchLimit = context.channelConfig?.lastActiveMembersLimit ?? 25
```
b-onc, line 84: IMHO context shouldn't be aware of `channelConfig` (I even don't think there should be a `channelConfig`, but that was another comment), `loadLastAciveMembers` should accept a `limit` param and it should be passed down (possibly from `ChannelController`, `itemCreator` block?)

b-onc (changes requested):
I think the arch approach needs a re-discussion internally

b-onc (approved):

VojtaStavik:
Nooooooooooooooooooooooooo

VojtaStavik:
I said I was going to take care of this @olejnjak 😭  This wasn't ready to merge

PR #905, opened 2021-03-22
[CIS-259] Implement teams

DominikBucher12:
# What this PR do:
- Adds support for user teams and channel team inside those payloads
# How to test it
- Run tests if `teams` property is parsed from payload.
# Where you can start
- `UserPayload` and `ChannelListPayload`, check if there is correct relationship between objects.

PR #912, opened 2021-03-24
Add team filter keys

VojtaStavik:
Just something we forgot to add in #905. This PR adds the possibility to filter users and channels using the `team(s)` keys.


DominikBucher12 (approved):
Ahh so this was the missing piece! :D Thanks, good for me.

PR #914, opened 2021-03-25
Feature/messenger clone

fortmarek:

Create basic MessengerClone.

Main missing things:
- composer view (waiting for audit)
- metadata should be only shown for the first message in a group, I'll fix this once I add appropriate property for that


MessengerClone/ChatClient.swift
```diff
@@ -0,0 +1,27 @@
+import UIKit
+import StreamChat
+import StreamChatUI
```
VojtaStavik, line 3: The files are not properly formatted -> missing file header with copyright

MessengerClone/FetchImage.swift
```diff
@@ -0,0 +1,149 @@
+import SwiftUI
+import Nuke
+
+/// Taken from: https://github.com/kean/FetchImage
```
VojtaStavik, line 4: We should also include the original copyright

hooks/pre-commit.sh
```diff
@@ -6,6 +6,7 @@
 ./hooks/git-format-staged --formatter 'mint run swiftformat --config .swiftformat stdin' 'DemoApp/*.swift'
 ./hooks/git-format-staged --formatter 'mint run swiftformat --config .swiftformat stdin' 'SlackClone/*.swift'
 ./hooks/git-format-staged --formatter 'mint run swiftformat --config .swiftformat stdin' 'iMessageClone/*.swift'
+./hooks/git-format-staged --formatter 'mint run swiftformat --config .swiftformat stdin' 'MessengerClone/*.swift'
```
VojtaStavik, line 9: It seems you added this too late :)

VojtaStavik (approved):
It seems the formatter didn't work correctly. Can you please reformat the existing code manually

hooks/pre-commit.sh
fortmarek: yep, you're right, my bad 🙇

MessengerClone/FetchImage.swift
fortmarek: I added the header comments from the original file above this comment. I was not able to change the top-level header because that is rewritten by `swiftformat`. I could probably find a way how to exclude but I believe the current situation might be fine

VojtaStavik: Yep, works 👍

PR #916, opened 2021-03-25
[CIS-753] First message in group

fortmarek:

Added property to `ChatMessageGroupPart` for `isFirstInGroup` - similar to `isLastInGroup`.

I have two following questions:
- any recommendations for making tests for this?
- I was asked to look at how to refactor `ChatMessageGroupPart` - and honestly, I am not really sure how. I was thinking of adding a delegate but views should only be configurable via `content` property, so that did not seem like a good path to take. Possible refactor would probably be better to do in a different PR anyways, in order for this to be purely additional


VojtaStavik (approved):
There's no need to test this at this point, we're auditing this component atm and will add all the tests needed.

PR #918, opened 2021-03-25
Setup docusaurus

b-onc:
Please check https://www.notion.so/Docusaurus-System-ad84e69b2837430c9c5cfeaf65d0c82f for info about the system.

I tried to test the github action locally using nektos/act but couldn't get it pushing... The bot should have push permission so the action should work.


.github/workflows/docusaurus.yml
```diff
@@ -0,0 +1,38 @@
+name: Deploy Docusaurus
+
+on:
+  push:
+    branches:
+      - main
```
VojtaStavik, line 6: So, do we want to run this on every push to main or on every release? 🤔

.github/workflows/docusaurus.yml
```diff
@@ -0,0 +1,38 @@
...
+  deploy-docusaurus:
+    name: 'Deploy Docusaurus'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - name: Install Bot SSH Key
+        uses: webfactory/ssh-agent@v0.4.1
+        with:
+          ssh-private-key: ${{ secrets.BOT_SSH_PRIVATE_KEY }}
+      - name: Install yarn (Somehow it isn't pre-installed?)
```
VojtaStavik, line 21: IMO it's better to channel our frustrations in our Slack channel rather than here ;-) Or at least make it a comment and don't include it into the action name?

VojtaStavik (approved):
Cool 👍

.github/workflows/docusaurus.yml
b-onc: Not on every push - only if there are changes to affecting files (`paths`), to always have an up-to-date documentation

.github/workflows/docusaurus.yml
b-onc: totally, apologizes 😅 most probably it _is_ preinstalled and we'll remove this action, somehow it didn't work locally

PR #919, opened 2021-03-25
Fix CI not running tests below iOS 14

nuno-vieira:
# In This PR

The tests were not running below < iOS 14. It would fall back to the default device 🤦‍♂️

Issues found:
- The Xcode version was hardcoded, and it can't be because each Xcode includes different default ios simulator runtimes.
- There was missing a pipe "|" for the action to actually work -.-
- The stress tests were not even running on the specified versions because it was missing the `options[:device]` parameter on the lanes. Even if the prepare-simulator-action was working, the stress tests were always running on the default device.

The script needs to have a condition because of how the Xcode path is different on 10.3 vs >10.3
More details here: https://github.com/actions/virtual-environments/issues/551#issuecomment-637344435

Now I finally got why the iOS 12 coverage was the same as the iOS 14 😄 

It probably would be a good idea to open source this action actually 🤔 It took a lot of time to make it work properly for every case.

### DIFF
--- a/.github/actions/prepare-ios-simulator/action.yml
+++ b/.github/actions/prepare-ios-simulator/action.yml
@@ -6,16 +6,27 @@ inputs:
   ios:
     description: 'The iOS version'
     required: true
+  xcode:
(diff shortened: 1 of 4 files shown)


VojtaStavik (approved):
This is cool 👍 🎉 Thank you

DominikBucher12 (approved):
🎉

PR #920, opened 2021-03-25
[CIS-674] Fix channel list inconsistency

olejnjak:
Well the inconsistency error we are getting is actually [docummented](https://developer.apple.com/documentation/uikit/uicollectionview/1618045-performbatchupdates#discussion).

> If the collection view's layout is not up to date before you call this method, a reload may occur. To avoid problems, you should update your data model inside the updates block or ensure the layout is updated before you call performBatchUpdates(_:completion:).

This is actually what is happening in test I added (will be removed as it only logs the error but it is not thrown 🤷)

Screen Shot 2021-03-25 at 23 07 39

So we need to make sure to update channel (especially its count) inside `performBatchUpdates`. Currently I thought about mirroring `controller.channels` to local property in `ChatChannelListVC`, the implementation is definitely to be discussed as there are other options, this is the most straight-forward solution I could think of.

I can imagine that we will just copy `controller.channels.count` property and update only that in `performBatchUpdates`, we could improve this mirroring solution and we might force the `collectionView` to lay out, but that is definitely not my favorite.

Got any more ideas?

### DIFF
--- a/Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
+++ b/Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
@@ -16,6 +16,9 @@ open class _ChatChannelListVC<ExtraData: ExtraDataTypes>: _ViewController,
     SwipeableViewDelegate {
     /// The `ChatChannelListController` instance that provides channels data.
     public var controller: _ChatChannelListController<ExtraData>!
+    
+    /// To prevent inconsistencies during `collectionView.performBatchUpdates` we need to update are dataSource count in updates block
+    private var channelsCount = 0
 
     open private(set) lazy var loadingIndicator: UIActivityIndicatorView = {
         if #available(iOS 13.0, *) {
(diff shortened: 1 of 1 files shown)


VojtaStavik (approved):
LGTM 👍 Any idea why the CI is failing @olejnjak?

olejnjak:
Well looks like fixing tests on older systems introduced this as it fails on iOS 13, but locally it still works for me 🙈

Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
```diff
@@ -62,6 +65,7 @@ open class _ChatChannelListVC<ExtraData: ExtraDataTypes>: _ViewController,
     override open func setUp() {
         super.setUp()
 
+        channelsCount = controller.channels.count
```
b-onc, line 68: This needs to be after `setDelegate` else it'll always be 0, right?

VojtaStavik: Not really. Here it's just about getting the initial value. Accessing channels loads the local channels lazily.

PR #922, opened 2021-03-26
Fix ChannelUpdatedEvent failing to parse

b-onc:
This was due to missing `user` field. Backend doesn't send `user` field (it's optional in v2) for every event. ChannelUpdated event hasn't received much love yet


VojtaStavik (approved):
👍

PR #924, opened 2021-03-27
Add missing options argument to fastlane lanes

nuno-vieira:
# In this PR

The options were missing from some release test lanes.

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -239,7 +239,7 @@ lane :test_ui_release do |options|
 end
 
 desc "Runs tests in Release config"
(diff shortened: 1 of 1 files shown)


VojtaStavik:
@nuno-vieira This is the reason why the full check failed here https://github.com/GetStream/stream-chat-swift/actions/runs/696306894

VojtaStavik (approved):

nuno-vieira:
> @nuno-vieira This is the reason why the full check failed here https://github.com/GetStream/stream-chat-swift/actions/runs/696306894

Yes 👍 This fixes that

PR #925, opened 2021-03-27
Remove building demo app iOS < 14 in Full Checks

nuno-vieira:
# In this PR
~Now, with the CI finally running on iOS < 14, another issue found is that the Demo App had min target of iOS 14, so the CI failed when trying to run on iOS 12 and iOS 13.~

~The issue is that changing the min target to iOS 12 has a lot of compilation errors, especially because of the SceneDelegate etc... And also other APIs that we use in the DemoApp are only iOS 13+. For now, probably the best idea is to only change the min target to 13? And actually is there any benefit in compiling the Demo Apps on iOS below 14? Since we already build the SDKs below iOS 14, what is the aim of compiling the demo apps below 14?~

Update:
Removes building demo app iOS < 14

### DIFF
--- a/.github/workflows/full-checks.yml
+++ b/.github/workflows/full-checks.yml
@@ -69,7 +69,7 @@ jobs:
       run: bundle exec fastlane stress_test_release device:"iPhone 12"
 
   stress-tests-ios13:
-    name: Stress Test LLC - iOS 13.4 (Release)
+    name: Stress Test LLC - iOS 13.5 (Release)
     runs-on: macos-latest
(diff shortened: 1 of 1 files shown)


VojtaStavik:
@nuno-vieira It was a deliberate decision to keep the Demo app iOS14 only. It's possible we will change this decision eventually but so far there was no reason for it. So I'd say it's not worth it do it now, especially when the UI SDK is not fully finished yet.

VojtaStavik (approved):

PR #923, opened 2021-03-26
[CIS-727] `ChatChannel` limits refinements

VojtaStavik:
This is a follow-up PR on #911 . It improves the naming and adds a limit for `ChatChannel.latestMesages`.

PR #926, opened 2021-03-28
 Use temp folder instead of /dev/null for test DB

VojtaStavik:
We see some issues with using `/dev/null` where writes to the test DB are getting locked by another test database. This change makes sure every test has its own _real_ sqlite db file in the temp directory.

### DIFF
--- a/Sources/StreamChat/Database/DatabaseContainer_Mock.swift
+++ b/Sources/StreamChat/Database/DatabaseContainer_Mock.swift
@@ -17,8 +17,8 @@ class DatabaseContainerMock: DatabaseContainer {
     @Atomic var recreatePersistentStore_errorResponse: Error?
     @Atomic var resetEphemeralValues_called = false
     
-    convenience init() {
-        try! self.init(kind: .onDisk(databaseFileURL: .newTemporaryFileURL()))
+    convenience init(channelConfig: ChatClientConfig.Channel? = nil) {
+        try! self.init(kind: .onDisk(databaseFileURL: .newTemporaryFileURL()), channelConfig: channelConfig)
     }
     
     override init(
--- a/Sources/StreamChat/Database/DatabaseContainer_Mock.swift
+++ b/Sources/StreamChat/Database/DatabaseContainer_Mock.swift
@@ -18,7 +18,7 @@ class DatabaseContainerMock: DatabaseContainer {
     @Atomic var resetEphemeralValues_called = false
     
     convenience init() {
-        try! self.init(kind: .inMemory)
+        try! self.init(kind: .onDisk(databaseFileURL: .newTemporaryFileURL()))
     }
     
     override init(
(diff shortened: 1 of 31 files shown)


Sources/StreamChat/Database/DatabaseContainer_Mock.swift
```diff
@@ -18,7 +18,7 @@ class DatabaseContainerMock: DatabaseContainer {
     @Atomic var resetEphemeralValues_called = false
     
     convenience init() {
-        try! self.init(kind: .inMemory)
+        try! self.init(kind: .onDisk(databaseFileURL: .newTemporaryFileURL()))
```
DominikBucher12, line 21: Shouldn't we try-catch this instead to know what's going on? This would surely bring some rewriting of the inits 😅

Tests/Shared/StressTestCase.swift
```diff
@@ -17,23 +17,22 @@ class StressTestCase: XCTestCase {
...
+                at: tempDirectoryURL,
+                includingPropertiesForKeys: nil,
+                options: .skipsHiddenFiles
+            )
+            
+            try tempFileURLs.forEach {
+                try fileManager.removeItem(at: $0)
+            }
+        } catch {
+            XCTFail("Failed to clean up after tests: \(error)")
```
olejnjak, line 35: I don't like relying on `StressTestCase` that it will cleanup temp directory, e.g. originally tests in #923 were failing because of that and `ChannelDTO_Tests` became `StressTestCase` because of my incorrect commit as I just used it for debugging, but never intended to push it on purpose. 

I'd prefer deleting DB file in `DatabaseContainer_Mock.deinit`, to generally prevent orphaned DB files. I know you're aware of general tests performance, so if you're aware of this limitation, it's good with me.

Sources/StreamChat/Database/DatabaseContainer_Mock.swift
VojtaStavik: tl;dr: You're probably right but I think it's OK.

---

Good comment. I think one of the main reasons for the convenience init is to hide the fact it throws.

However, this `init` is used only in tests that don't test the actual DB functionality. I think it's OK to hide the throwing here. Even if I added some inner `do { } catch { }` and log the error nicely, I'd still have to add a `fatalError` because it's a non-failable initializer. This container init will fail only if the CD stack fails to init which in our setup is a programmer error.

Tests/Shared/StressTestCase.swift
VojtaStavik: @olejnjak Makes sense 👍 I added the general cleanup to `TestsEnvironmentSetup` to make sure we always start with a clean temp folder. Then I added a flag to `DatabaseContainerMock` that makes it delete its own DB files once it's deinited. I restricted this functionality only to the convenience init since there are some tests that rely on the fact the file stays there. The assumption is: if you use the convenience init, you don't really on the DB details in the tests.

b-onc (approved):
:shipit:

PR #930, opened 2021-03-29
[CIS-681] Log error when UIConfig is assigned after the view is initialized

LukasHromadnik:

Assigning `UIConfig` instance explicitly is possible only before view's subviews are instantiated. When someone assigns `UIConfig` after the view has been initialized, we should log an error. You can use the base views' `isInitialized` for this.

### DIFF
--- a/Sources/StreamChatUI/CommonViews/BaseViews.swift
+++ b/Sources/StreamChatUI/CommonViews/BaseViews.swift
@@ -3,6 +3,7 @@
 //
 
 import Foundation
+import StreamChat
 import UIKit
 
 // Just a protocol to formalize the methods required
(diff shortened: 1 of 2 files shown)


Sources/StreamChatUI/CommonViews/BaseViews.swift
```diff
@@ -49,11 +50,19 @@ public extension Customizable where Self: UIViewController {
     }
 }
 
+extension UIConfigProvider where Self: _View {
+    public func uiConfigDidRegister() {
+        if isInitialized {
+            log(.error, message: "UIConfig assigned after the view is initialized.")
```
VojtaStavik, line 56: Let's throw an assertion failure here. This is a serious error. You can use `log.assertionFailure` for that. The message can be something like:
```
`UIConfig` was assigned after the view has been already initialized. This is most likely caused by assigning the custom `UIConfig` instance after the view has been added to the view hierarchy, or after the view's subviews have been initialized already. This is undefined behavior and should be avoided.
```

VojtaStavik (approved):

Sources/StreamChatUI/CommonViews/BaseViews.swift
LukasHromadnik: ✅

Sources/StreamChatUI/CommonViews/BaseViews.swift
```diff
@@ -49,11 +50,22 @@ public extension Customizable where Self: UIViewController {
     }
 }
 
+extension UIConfigProvider where Self: _View {
+    public func uiConfigDidRegister() {
+        if isInitialized {
+            log
+                .assertationFailure(
+                    "`UIConfig` was assigned after the view has been already initialized. This is most likely caused by assigning the custom `UIConfig` instance after the view has been added to the view hierarchy, or after the view's subviews have been initialized already. This is undefined behavior and should be avoided."
+                )
```
VojtaStavik, line 59: nit: We tend to format things like this by multiple concatenated strings to keep the readability up:
```swift
log.assertationFailure(
    "`UIConfig` was assigned after the view has been already initialized. This is most likely caused by assigning " +
    "the custom `UIConfig` instance after the view has been added to the view hierarchy, or after the view's subviews " +
    "have been initialized already. This is undefined behavior and should be avoided."
)
```

VojtaStavik (approved):
LGTM, just one tiny nit and you can merge 👍 Don't forget that this contains `log.assertationFailure` so you will have to adjust this PR or #935 to fix this, depending on which one you'll merge first.

Sources/StreamChatUI/CommonViews/BaseViews.swift
LukasHromadnik: message ✅  let's merge the #935 first 👍

PR #931, opened 2021-03-29
[CIS-748] Audit `ChatMessageQuoteView`

nuno-vieira:
# In this PR

- Refactors the message quote views. Removes `ChatMessageComposerQuoteBubbleView` and `ChatMessageQuoteBubbleView`, and replaces with only one `ChatMessageQuoteView`.
- Audits the `ChatMessageQuoteView` component.


Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerQuoteBubbleView.swift
```diff
@@ -205,4 +214,16 @@ open class _ChatMessageComposerQuoteBubbleView<ExtraData: ExtraDataTypes>: _View
...
+        if hidden {
+            layout.attachmentPreviewConstraints.prefix(3).forEach {
+                $0.constant = 0
+            }
+        } else {
+            layout.attachmentPreviewConstraints[0].constant = attachmentPreviewSize.width
+            layout.attachmentPreviewConstraints[1].constant = attachmentPreviewSize.height
+            layout.attachmentPreviewConstraints[2].constant = contentView.directionalLayoutMargins.leading
+        }
+    }
```
nuno-vieira, line 228: I guess this is one of the cons of having an array of constraints instead of each constraint in the `Layout` struct? Maybe for this special case, we can store the exact constraints? Using the indices to hide or unhide this view is of course not a good idea because the array can change by the customer, or not do anything at all, if they are deactivated.

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerQuoteBubbleView_Tests.swift
```diff
@@ -0,0 +1,141 @@
...
+                fatalError("init(coder:) has not been implemented")
+            }
+        }
+
+        var config = UIConfig()
+        config.messageComposer.quotedMessageBubbleSubviews.contentView = TestView.self
+
+        view.uiConfig = config
+
+        AssertSnapshot(view, variants: [.defaultLight])
```
nuno-vieira, line 70: @VojtaStavik This does not work properly, the border color will be overridden by the appearance life cycle. Should the views of UIConfig really be the most generic ones? In this case there's actually no option, but maybe a better option is to make the UIConfigs conform to `Customizable`? Or the appearance of the views should always be customized in the appearance life cycle of the Parent Customizable view?

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerQuoteBubbleView.swift
nuno-vieira: Attempt here: https://github.com/GetStream/stream-chat-swift/pull/931/commits/1a5ac598099bfb40062b7bcd688bbdca2fafd648

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerQuoteBubbleView.swift
```diff
@@ -8,44 +8,66 @@ import UIKit
...
-    
-    public var avatarViewSize: CGSize = .init(width: 24, height: 24)
-    public var attachmentPreviewSize: CGSize = .init(width: 34, height: 34)
-    
-    public var message: _ChatMessage<ExtraData>? {
+    /// Default avatar view size. To override this, subclass `setUpLayout` and add width and height constraints.
+    private var avatarViewSize: CGSize = .init(width: 24, height: 24)
+
+    /// Default attachment preview size. To override this, subclass `setUpLayout` and add width and height constraints.
+    private var attachmentPreviewSize: CGSize = .init(width: 34, height: 34)
```
nuno-vieira, line 15: I've made this private, but probably is not the best idea, especially if you just want to change the size of the avatar and the attachments. If the customer changes the size in the `setUpLayout` the hiding will not work properly, since it will use the old width and height.

Sources/StreamChatUI/CommonViews/ChatMessageQuoteView/ChatMessageQuoteView.swift
```diff
@@ -0,0 +1,224 @@
...
+import StreamChat
+import UIKit
+
+/// The `ChatMessageQuoteView` content.
+public struct ChatMessageQuoteViewContent<ExtraData: ExtraDataTypes> {
+    /// The quoted message.
+    let message: _ChatMessage<ExtraData>?
+    /// The avatar position in relation with the text message.
+    let avatarAlignment: ChatMessageQuoteAvatarAlignment
+}
```
nuno-vieira, line 14: I think this is a pattern we could adopt. I think using tuples for the content of the views is a bit limiting and not very scalable if the tuple starts to grow. Also, auto-completion with tuples is worse than using a regular type. WDYGT?

VojtaStavik (approved):

Sources/StreamChatUI/CommonViews/ChatMessageQuoteView/ChatMessageQuoteView.swift
nuno-vieira: The problem with the Nested type inside the `ChatMessageQuoteView` is that it makes it harder to type because of ExtraData 😕 In order to access the subtype, the customer will have to write `ChatMessageQuoteView<ExtraData>.Content`. Probably at first, you will write just `ChatMessageQuoteView.Content` and the compiler will warn you that this is not possible, if the customer knows about ExtraData, he will be able to tell what is wrong, but is not very intuitive.

PR #934, opened 2021-03-30
[CIS-776] Extract ChatVC logic to separate components

LukasHromadnik:
# Description of the pull request

Extracted common logic for `ChatVC`'s children to separate components:

* `ChatMessageListTitleView` (audit will be in future PR)
* `KeyboardFrameObserver`

### DIFF
--- a/Sources/StreamChatUI/ChatMessageList/ChatMessageListTitleView.swift
+++ b/Sources/StreamChatUI/ChatMessageList/ChatMessageListTitleView.swift
@@ -1,0 +1,48 @@
+//
+// Copyright © 2021 Stream.io Inc. All rights reserved.
+//
+
+import StreamChat
+import UIKit
+
+// TODO: Audit
+public class ChatMessageListTitleView<ExtraData: ExtraDataTypes>: UIView, UIConfigProvider {
+    public var title: String? {
+        get { titleLabel.text }
+        set { titleLabel.text = newValue }
+    }
+    
+    public var subtitle: String? {
+        get { subtitleLabel.text }
+        set { subtitleLabel.text = newValue }
+    }
+    
+    private weak var titleLabel: UILabel!
+    private weak var subtitleLabel: UILabel!
+    
+    public init() {
+        super.init(frame: .zero)
+        
+        let titleLabel = UILabel()
+        titleLabel.textAlignment = .center
+        titleLabel.font = uiConfig.font.headlineBold
+        self.titleLabel = titleLabel
+
+        let subtitleLabel = UILabel()
+        subtitleLabel.textAlignment = .center
+        subtitleLabel.font = uiConfig.font.caption1
+        subtitleLabel.textColor = uiConfig.colorPalette.subtitleText
+        self.subtitleLabel = subtitleLabel
+
+        let titleView = UIStackView(arrangedSubviews: [titleLabel, subtitleLabel])
+        titleView.axis = .vertical
+        addSubview(titleView)
+        titleView.translatesAutoresizingMaskIntoConstraints = false
+        titleView.pin(to: self)
+    }
+    
+    @available(*, unavailable)
+    public required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+}
--- a/Sources/StreamChatUI/Utils/KeyboardFrameObserver.swift
+++ b/Sources/StreamChatUI/Utils/KeyboardFrameObserver.swift
@@ -1,0 +1,87 @@
+//
+// Copyright © 2021 Stream.io Inc. All rights reserved.
+//
+
+import StreamChat
+import UIKit
+
+final class KeyboardFrameObserver {
+    weak var containerView: UIView!
+    weak var scrollView: UIScrollView!
+    weak var composerBottomConstraint: NSLayoutConstraint?
+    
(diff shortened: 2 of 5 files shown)


Sources/StreamChatUI/Utils/KeyboardFrameObserver.swift
```diff
@@ -0,0 +1,87 @@
...
+final class KeyboardFrameObserver {
+    weak var containerView: UIView!
+    weak var scrollView: UIScrollView!
+    weak var composerBottomConstraint: NSLayoutConstraint?
+    
+    init(containerView: UIView, scrollView: UIScrollView, composerBottomConstraint: NSLayoutConstraint?) {
+        self.containerView = containerView
+        self.scrollView = scrollView
+        self.composerBottomConstraint = composerBottomConstraint
+    }
```
VojtaStavik, line 17: This class takes `containerView`, `scrollView`, and `composerBottomConstraint` as input parameters. It seems to me it's very closely tightened to the message list controllers. Maybe we should reflect it in the name of the class + it should not live in `/Utils`?

Another option would be to more generalize the naming here if you think it's possible and it makes sense.

VojtaStavik:
The commit and the PR names are a little bit confusing. This doesn't _actually_ extract any code. It's more about adding some common components that will be used later in the message list VCs.

Sources/StreamChatUI/Utils/KeyboardFrameObserver.swift
LukasHromadnik: Currently I don't see any reason to make it general. So I renamed it to `ChatMessageListKeyboardObserver` and moved it to `ChatMessage` folder.

LukasHromadnik:
I extracted the logic from `ChatVC`, hopefully now the PR makes more sense

Sources/StreamChatUI/ChatMessageList/ChatMessageListTitleView.swift
```diff
@@ -0,0 +1,48 @@
...
+        set { titleLabel.text = newValue }
+    }
+    
+    public var subtitle: String? {
+        get { subtitleLabel.text }
+        set { subtitleLabel.text = newValue }
+    }
+    
+    private weak var titleLabel: UILabel!
+    private weak var subtitleLabel: UILabel!
```
DominikBucher12, line 21: Just curious... Why are those weak?

Sources/StreamChatUI/ChatMessageList/ChatMessageListTitleView.swift
```diff
@@ -0,0 +1,48 @@
+//
+// Copyright © 2021 Stream.io Inc. All rights reserved.
+//
+
+import StreamChat
+import UIKit
+
+// TODO: Audit
+public class ChatMessageListTitleView<ExtraData: ExtraDataTypes>: UIView, UIConfigProvider {
```
DominikBucher12, line 9: This view is a bit different than all others because it's the titleView of navigationBar, right? I am still convinced that it should inherit our `_View` class. AKA `ChatMessageListTitleView<ExtraData: ExtraDataTypes>: _View`?

VojtaStavik (approved):
Nice 🎉

Sources/StreamChatUI/ChatMessageList/ChatMessageListTitleView.swift
```diff
@@ -0,0 +1,48 @@
...
+        subtitleLabel.font = uiConfig.font.caption1
+        subtitleLabel.textColor = uiConfig.colorPalette.subtitleText
+        self.subtitleLabel = subtitleLabel
+
+        let titleView = UIStackView(arrangedSubviews: [titleLabel, subtitleLabel])
+        titleView.axis = .vertical
+        addSubview(titleView)
+        titleView.translatesAutoresizingMaskIntoConstraints = false
+        titleView.pin(to: self)
+    }
```
DominikBucher12, line 42: Actually if you subclass the `_View` you can follow our lifecycle logic with `setUpLayout` and `setUpAppearance` which then makes it easier for our customers to adopt it and set custom behaviour :)

Sources/StreamChatUI/ChatMessageList/ChatMessageListTitleView.swift
```diff
@@ -0,0 +1,48 @@
...
+    
+    @available(*, unavailable)
+    public required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+}
```
DominikBucher12, line 48: Not actually sure if this is a good idea if anyone would want to use this in interface builder-based pace.

Sources/StreamChatUI/ChatMessageList/ChatMessageListTitleView.swift
VojtaStavik: @DominikBucher12 FYI this PR is just about moving the functionality, the audit will come in follow-up PRs

LukasHromadnik:
@DominikBucher12 thanks for the comments. 👍 The audit is coming in the next PR where all those things are already fixed. 🙂

DominikBucher12 (approved):
Anyway LGTM, I also just realised there is `TODO: Audit` :D Nice <3

PR #935, opened 2021-03-30
[CIS-757] Rename `assertationFailure` -> `assertionFailure`

LukasHromadnik:
# Description of the pull request

> The first one is not correct. Don't forget that it's already part of the public API, so it should be renamed in a non-breaking way.

Proposed solution will raise an error if the old method is used. Xcode will then suggest the correct new name for the method and the "Fix" button.

We can also do it with the `deprecated` annotation which will raise only a warning.


Sources/StreamChat/Utils/Logger/Logger.swift
```diff
@@ -234,11 +234,24 @@ public class Logger {
     ///
     /// - Parameters:
     ///   - message: A custom message to log if `condition` is evaluated to false.
+    @available(*, unavailable, renamed: "assertionFailure")
     public func assertationFailure(
         _ message: @autoclosure () -> Any,
         functionName: StaticString = #function,
         fileName: StaticString = #file,
         lineNumber: UInt = #line
+    ) { }
```
VojtaStavik, line 243: This shouldn't be empty but it should rather call the new implementation.

Sources/StreamChat/Utils/Logger/Logger.swift
```diff
@@ -234,11 +234,24 @@ public class Logger {
...
+    
+    /// Stops program execution with `Swift.assertationFailure`. In RELEASE builds only
```
VojtaStavik, line 245: please fix the name in the comment, too

VojtaStavik (changes requested):

Sources/StreamChat/Utils/Logger/Logger.swift
VojtaStavik: Also, can you please move this method to `Deprecations.swift`?

DominikBucher12 (changes requested):
> The first one is not correct. Don't forget that it's already part of the public API, so it should be renamed in a non-breaking way.

In this case, the changelog is missing. 🤔

Sources/StreamChat/Utils/Logger/Logger.swift
LukasHromadnik: I changed the `unavailable` mark to `deprecated` and moved the whole method to `Deprecations.swift`

Sources/StreamChat/Utils/Logger/Logger.swift
LukasHromadnik: ✅

LukasHromadnik:
> > The first one is not correct. Don't forget that it's already part of the public API, so it should be renamed in a non-breaking way.
> 
> In this case, the changelog is missing. 🤔

CHANGELOG is updated ✅

DominikBucher12 (approved):
LGTM 🎉

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 # Upcoming
 
 ### 🔄 Changed
+- Add `Logger.assertionFailure` to replace `Logger.assertationFailure` which has incorrect name and it's now deprecated [#935](https://github.com/GetStream/stream-chat-swift/pull/935)
```
VojtaStavik, line 9: Can we just say "`Logger.assertationFailure` was renamed to `Logger.assertionFailure`"? :) I think it's enough :)

VojtaStavik (approved):
LGTM 👍 Please just one minor fix and you can apply to fixmes and merge 🎉

CHANGELOG.md
LukasHromadnik: done ✅

PR #939, opened 2021-03-30
[CIS-654] Mask indicator view in avatar view.

skorepak:
Until now, the online indicator view (green dot) at avatar had a border with the same colour as a background, to appear like a cut-out. This PR is replacing that with proper masking to actually cut the circle out of the avatar. 

## API Changes
Added `maskingPath: CGPath?` property to `ChatOnlineIndicatorView` in case clients will want to customise the cut-out shape. 

Added `showOnlineIndicator: Bool` property to `ChatPresenceAvatarView` to move the setup code from `ChatChannelAvatarView` to `ChatPresenceAvatarView` as the `ChatPresenceAvatarView` is direct parent of the affected subviews and IMO the logic should go there.

## Links
https://stream-io.atlassian.net/browse/CIS-654


Sources/StreamChatUI/CommonViews/AvatarView/ChatPresenceAvatarView/ChatPresenceAvatarView.swift
```diff
@@ -35,4 +43,26 @@ open class _ChatPresenceAvatarView<ExtraData: ExtraDataTypes>: _View, UIConfigPr
...
+        super.layoutSubviews()
+        
+        setUpMask(indicatorVisible: showOnlineIndicator)
+    }
+    
+    /// Creates space for indicator view in avatar view by masking path provided by the indicator view.
+    /// - Parameter visible: Bool to determine if the indicator should be shown. The avatar view won't be masked if the indicator is not visible.
+    open func setUpMask(indicatorVisible: Bool) {
+        guard
+            indicatorVisible,
```
VojtaStavik, line 57: Why not just check `onlineIndicatorView.isVisible`?

VojtaStavik (approved):
Congratulations on your first PR here 🎉 Looks really good 👍  I left a couple of questions and nit comments.

Sources/StreamChatUI/CommonViews/AvatarView/ChatPresenceAvatarView/ChatOnlineIndicatorView.swift
```diff
@@ -10,27 +10,37 @@ public typealias ChatOnlineIndicatorView = _ChatOnlineIndicatorView<NoExtraData>
 
 /// A view used to indicate the presence of a user.
 open class _ChatOnlineIndicatorView<ExtraData: ExtraDataTypes>: _View, UIConfigProvider {
+    
+    open private(set) lazy var circleView: UIView = UIView()
+        .withoutAutoresizingMaskConstraints
+
```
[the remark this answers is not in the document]
skorepak: Is `indicatorView` ok? I wasn't sure because it's part of the class name, to not be confused.

Sources/StreamChatUI/CommonViews/AvatarView/ChatPresenceAvatarView/ChatPresenceAvatarView.swift
skorepak: I don't want to depend on `onlineIndicatorView.isVisible` because then the `setUpMask` function would always have to be called *after* you set the `onlineIndicatorView.isVisible`, giving the consumer a bigger chance to make a mistake by changing the call order.

Sources/StreamChatUI/CommonViews/AvatarView/ChatPresenceAvatarView/ChatPresenceAvatarView.swift
```diff
@@ -16,9 +16,17 @@ open class _ChatPresenceAvatarView<ExtraData: ExtraDataTypes>: _View, UIConfigPr
         .withoutAutoresizingMaskConstraints
 
     /// A view indicating whether the user this view represents is online.
-    open private(set) lazy var onlineIndicatorView: UIView = uiConfig
+    open private(set) lazy var onlineIndicatorView: _ChatOnlineIndicatorView<ExtraData> = uiConfig
         .onlineIndicatorView.init()
         .withoutAutoresizingMaskConstraints
+    
+    /// Bool to determine if the indicator should be shown.
+    open var showOnlineIndicator: Bool = false {
```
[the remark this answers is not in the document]
nuno-vieira: what about just `isOnline` ?

Sources/StreamChatUI/UIConfig.swift
```diff
@@ -9,7 +9,7 @@ public typealias UIConfig = _UIConfig<NoExtraData>
 
 public struct _UIConfig<ExtraData: ExtraDataTypes> {
     /// A view used as an online activity indicator (online/offline).
-    public var onlineIndicatorView: UIView.Type = _ChatOnlineIndicatorView<ExtraData>.self
```
nuno-vieira, line 12: The aim is for the type to have the most generic possible. Any reason why this was changed? 🤔

DominikBucher12 (approved):
For me looks good! <3

Sources/StreamChatUI/CommonViews/AvatarView/ChatPresenceAvatarView/ChatPresenceAvatarView.swift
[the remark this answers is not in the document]
skorepak: @nuno-vieira  I had isOnline before, but for me, it didn't match iOS conventions. `showOnlineIndicator` sounded the most "natural" to me, but I will take what you guys think is the best.

nuno-vieira:
Btw, I've just noticed something from the images diff, that the indicator now looks different, the green part is bigger, do we want that? 🤔

skorepak:
@nuno-vieira Yes, we want that :) The indicator is bigger and shifted more out of the avatar image. I had a call with Francesco Mastrogiacomo yesterday to confirm that. Here's the design source of truth https://www.figma.com/file/ekifwChR9tR7zRJg1QEzSM/Stream-Mobile-Chat?node-id=10158%3A22790

Sources/StreamChatUI/CommonViews/AvatarView/ChatPresenceAvatarView/ChatOnlineIndicatorView.swift
```diff
@@ -5,16 +5,22 @@
 import StreamChat
 import UIKit
 
+public protocol MaskProviding {
+    /// Path used to mask space in super view.
+    ///
+    /// No mask is used when nil is returned
+    var maskingPath: CGPath? { get }
+}
+
```
VojtaStavik, line 14: I'm not 100% convinced about this approach. We usually use subclassing for cases like this. Would be great to get an opinion from the rest of the team: cc @b-onc @DominikBucher12 @nuno-vieira

Sources/StreamChatUI/CommonViews/AvatarView/ChatPresenceAvatarView/ChatPresenceAvatarView.swift
```diff
@@ -16,9 +16,20 @@ open class _ChatPresenceAvatarView<ExtraData: ExtraDataTypes>: _View, UIConfigPr
         .withoutAutoresizingMaskConstraints
 
     /// A view indicating whether the user this view represents is online.
+    ///
+    /// The type of `onlineIndicatorView` is UIView & MaskProviding in UIConfig.
+    /// Xcode is failing to compile due to `Segmentation fault: 11` when used here.
     open private(set) lazy var onlineIndicatorView: UIView = uiConfig
```
VojtaStavik, line 22: 🤷‍♂️

Sources/StreamChatUI/UIConfig.swift
```diff
@@ -9,7 +9,7 @@ public typealias UIConfig = _UIConfig<NoExtraData>
...
+    public var onlineIndicatorView: (UIView & MaskProviding).Type = _ChatOnlineIndicatorView<ExtraData>.self
```
VojtaStavik, line 12: Maybe this can stay `UIView`? There's not reason to force users to provide a mask if the don't need to cut out the superview.

VojtaStavik (approved):
LGTM 👍 I'm surprised to see how many snapshots were recorded incorrectly! We must pay much bigger attention to it.

Please wait for the team feedback on the masking protocol and if all good, feel free to merge 👍

Sources/StreamChatUI/UIConfig.swift
skorepak: I was thinking about this a lot. If we don't request the mask, users might be surprised why the custom view is not masked and could take them a good amount of time to realize how to achieve that from our implementation. By asking them to conform to this protocol, it's explicit and easy to just return nil when they don't want to use the mask.

DominikBucher12 (approved):
Nice  🎉 😍 Thanks you for resolving this

nuno-vieira (approved):
Looks really good 👏 Only thing missing is the `MaskProviding` documentation, other then that LGTM ✅

PR #950, opened 2021-04-01
[CIS-754] Batch processing of events

VojtaStavik:
This PR is a little bit bigger but most of the changes there are just mechanical changes so bear with me :)

**Original state**
- Every middleware created its own write session in the database. This means that each event could generate several db observer updates and cause UI re-rendering.
- This situation got even worse when the app started. The SDK asks the backend for the missing events while it was offline. In combination with the initial channel list query which also generates a lot of watcher events, it was very common that the SDK generated several hundred (❗)  UI update callbacks within the first second or two.

**Changes in the PR**
- All middlewares now share a single database session so it's guaranteed that every event will cause at max one database write.
- We now batch the processing of the events. Currently, the batching window is 0.5s but we should tweak this value and find the sweet spot. Thanks to this change, we have 2-3 UI updates when the app starts, rather than 200-300.

---

_Note:_ The current way we process the events is not exactly safe and we need to change it. The syntax makes it possible to keep the reference to the session and use it in an escaping way. This will be addressed in a follow-up PR.

### DIFF
--- a/Sources/StreamChat/Database/DatabaseSession_Mock.swift
+++ b/Sources/StreamChat/Database/DatabaseSession_Mock.swift
@@ -1,0 +1,214 @@
+//
+// Copyright © 2021 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+@testable import StreamChat
+
+/// This class allows you to wrap an existing `DatabaseSession` and adjust the behavior of its methods.
+class DatabaseSessionMock: DatabaseSession {
+    /// The wrapped session
+    let underlyingSession: DatabaseSession
+
+    /// If set to a non-nil value, the error is returned from all throwing methods of the session
+    @Atomic var errorToReturn: Error?
+    
+    init(underlyingSession: DatabaseSession) {
+        self.underlyingSession = underlyingSession
+    }
+}
+
+// Here start the boilerplate that forwards and intercepts the session calls if needed
+
+extension DatabaseSessionMock {
+    func saveCurrentUserDevices(_ devices: [DevicePayload], clearExisting: Bool) throws {
+        try throwErrorIfNeeded()
+        try underlyingSession.saveCurrentUserDevices(devices, clearExisting: clearExisting)
+    }
+    
+    func saveChannel<ExtraData>(
+        payload: ChannelDetailPayload<ExtraData>,
+        query: _ChannelListQuery<ExtraData.Channel>?
+    ) throws -> ChannelDTO where ExtraData: ExtraDataTypes {
+        try throwErrorIfNeeded()
+        return try underlyingSession.saveChannel(payload: payload, query: query)
+    }
+    
+    func saveUser<ExtraData>(payload: UserPayload<ExtraData>, query: _UserListQuery<ExtraData>?) throws -> UserDTO
+        where ExtraData: UserExtraData {
+        try throwErrorIfNeeded()
+        return try underlyingSession.saveUser(payload: payload, query: query)
+    }
+    
+    func saveQuery<ExtraData>(query: _UserListQuery<ExtraData>) throws -> UserListQueryDTO? where ExtraData: UserExtraData {
+        try throwErrorIfNeeded()
+        return try underlyingSession.saveQuery(query: query)
+    }
+    
+    func user(id: UserId) -> UserDTO? {
+        underlyingSession.user(id: id)
+    }
+    
+    func deleteQuery<ExtraData>(_ query: _UserListQuery<ExtraData>) where ExtraData: UserExtraData {
+        underlyingSession.deleteQuery(query)
+    }
+    
+    func saveCurrentUser<ExtraData>(payload: CurrentUserPayload<ExtraData>) throws -> CurrentUserDTO
+        where ExtraData: UserExtraData {
+        try throwErrorIfNeeded()
+        return try saveCurrentUser(payload: payload)
+    }
+    
+    func saveCurrentUserUnreadCount(count: UnreadCount) throws {
+        try throwErrorIfNeeded()
+        try saveCurrentUserUnreadCount(count: count)
+    }
+    
+    func deleteDevice(id: DeviceId) {
+        underlyingSession.deleteDevice(id: id)
+    }
+    
+    func currentUser() -> CurrentUserDTO? {
+        underlyingSession.currentUser()
+    }
+    
+    func createNewMessage<ExtraData>(
+        in cid: ChannelId,
+        text: String,
+        pinning: MessagePinning?,
+        command: String?,
+        arguments: String?,
+        parentMessageId: MessageId?,
+        attachments: [AttachmentEnvelope],
+        showReplyInChannel: Bool,
+        quotedMessageId: MessageId?,
+        extraData: ExtraData
+    ) throws -> MessageDTO where ExtraData: MessageExtraData {
+        try throwErrorIfNeeded()
+
+        return try underlyingSession.createNewMessage(
+            in: cid,
+            text: text,
+            pinning: pinning,
+            command: command,
+            arguments: arguments,
+            parentMessageId: parentMessageId,
+            attachments: attachments,
+            showReplyInChannel: showReplyInChannel,
+            quotedMessageId: quotedMessageId,
+            extraData: extraData
+        )
+    }
+    
+    func saveMessage<ExtraData>(payload: MessagePayload<ExtraData>, for cid: ChannelId?) throws -> MessageDTO
+        where ExtraData: ExtraDataTypes {
+        try throwErrorIfNeeded()
+        return try underlyingSession.saveMessage(payload: payload, for: cid)
+    }
+    
+    func pin(message: MessageDTO, pinning: MessagePinning) throws {
+        try throwErrorIfNeeded()
+        return try underlyingSession.pin(message: message, pinning: pinning)
+    }
+    
+    func unpin(message: MessageDTO) {
+        underlyingSession.unpin(message: message)
+    }
+    
+    func message(id: MessageId) -> MessageDTO? {
+        underlyingSession.message(id: id)
+    }
+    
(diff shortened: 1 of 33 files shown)


Sources/StreamChat/WebSocketClient/EventMiddlewares/EventMiddleware_Tests.swift
```diff
@@ -2,16 +2,17 @@
...
 class EventMiddleware_Tests: XCTestCase {
     /// A test middleware that can be initiated with a closure/
     struct ClosureBasedMiddleware: EventMiddleware {
-        let closure: (_ event: Event, _ completion: @escaping (Event?) -> Void) -> Void
+        let closure: (_ event: Event, _ session: DatabaseSession, _ completion: @escaping (Event?) -> Void) -> Void
         
-        func handle(event: Event, completion: @escaping (Event?) -> Void) {
-            closure(event, completion)
+        func handle(event: Event, session: DatabaseSession, completion: @escaping (Event?) -> Void) {
+            closure(event, session, completion)
```
evsaev, line 15: Maybe we can use `EventMiddlewareMock` here and get rid of `ClosureBasedMiddleware`?

Sources/StreamChat/Database/DatabaseSession_Mock.swift
```diff
@@ -0,0 +1,257 @@
...
+// Here start the boilerplate that forwards and intercepts the session calls if needed
+
+extension DatabaseSessionMock {
+    func saveCurrentUserDevices(_ devices: [DevicePayload], clearExisting: Bool) throws {
+        if let error = errorToReturn {
+            throw error
+        } else {
+            return try underlyingSession.saveCurrentUserDevices(devices, clearExisting: clearExisting)
+        }
+    }
```
evsaev, line 30: WDYT about some private helper?

```swift
private extension DatabaseSessionMock {
    func checkErrorFirst<Output>(action: () throws -> Output) throws -> Output {
        if let error = errorToReturn {
            throw error
        } else {
            return try action()
        }
    }
}

extension DatabaseSessionMock {
    func saveCurrentUserDevices(_ devices: [DevicePayload], clearExisting: Bool) throws {
        try checkErrorFirst {
            try underlyingSession.saveCurrentUserDevices(devices, clearExisting: clearExisting)
        }
    }
    // the same for all other methods where we check the error first
}
```

Sources/StreamChat/WebSocketClient/EventMiddlewares/ChannelMemberTypingStateUpdaterMiddleware.swift
```diff
@@ -25,12 +23,10 @@ struct ChannelMemberTypingStateUpdaterMiddleware<ExtraData: ExtraDataTypes>: Eve
...
-        }, completion: { error in
-            if let error = error {
-                log.error("Failed saving incoming `TypingEvent` data to DB. Error: \(error)")
-            }
-            
-            completion(event)
-        })
+        } catch {
+            log.error("Failed saving incoming `TypingEvent` data to DB. Error: \(error)")
+        }
```
evsaev, line 28: The compiler complains that there's nothing to catch here 🙂

Sources/StreamChat/WebSocketClient/EventMiddlewares/ChannelReadUpdaterMiddleware.swift
```diff
@@ -6,41 +6,39 @@ import Foundation
...
-                log.error("Failed to update channel read for cid \(cid) and userId \(userId), error: \(error)")
-            }
-            completion()
-        })
+    private func updateReadEvent(
+        for cid: ChannelId,
+        userId: UserId,
+        lastReadAt: Date,
+        session: DatabaseSession,
+        completion: @escaping () -> Void
```
evsaev, line 33: The method works in a sync manner, do we need a `completion`? 🤔 Or it's for readability?

Sources/StreamChat/WebSocketClient/EventMiddlewares/MessageReactionsMiddleware.swift
```diff
@@ -18,7 +16,7 @@ struct MessageReactionsMiddleware<ExtraData: ExtraDataTypes>: EventMiddleware {
             return
         }
         
-        database.write({ session in
+        do {
```
evsaev, line 19: There's a regression in case of `ReactionDeletedEvent` with the missing local dto: the `completion` won't be invoked

Sources/StreamChat/WebSocketClient/EventMiddlewares/UserChannelBanEventsMiddleware.swift
```diff
@@ -18,41 +16,33 @@ struct UserChannelBanEventsMiddleware<ExtraData: ExtraDataTypes>: EventMiddlewar
...
+                break
             }
-
-        default:
-            completion(event)
+        } catch {
+            log.error("Error handling `\(type(of: event))` event: \(error)")
         }
+        
+        completion(event)
```
evsaev, line 45: nit: I really like `defer { completion(event) }` approach used in some middlewares, it's really clear that an event will be forwarded in any case, no matter if the middleware is able to process it. Maybe we can update all the middlewares to follow this pattern (not necessarily in this PR)?

Sources/StreamChat/Workers/EventNotificationCenter.swift
```diff
@@ -11,6 +11,15 @@ class EventNotificationCenter: NotificationCenter {
     /// The database used when evaluating middlewares.
     let database: DatabaseContainer
     
+    /// Events are processed in batches to reduce database writes. This interval is the max number of seconds
+    /// an event can wait before its processing started.
+    ///
+    /// Mutating this value doesn't affect the existing batch and the new value is applied for the following batch.
+    ///
+    var eventBatchPeriod: TimeInterval = 0.5
```
evsaev, line 19: Feels like smth we could put into `ChatClientConfig` and make configurable 🤔

Sources/StreamChat/Workers/EventNotificationCenter_Tests.swift
```diff
@@ -89,6 +90,87 @@ final class EventNotificationCenter_Tests: XCTestCase {
...
+        XCTAssertEqual(database.writeSessionCounter, 2)
+        XCTAssertEqual(eventLogger.events as! [TestEvent], testEvents)
+    }
+}
+
+private extension EventNotificationCenter_Tests {
+    func wait(_ time: TimeInterval) {
+        let start = Date()
+        AssertAsync.willBeTrue(Date().timeIntervalSince(start) >= time)
+    }
```
evsaev, line 173: ❤️

evsaev (approved):
Great! 🚢

Sources/StreamChat/WebSocketClient/EventMiddlewares/UserChannelBanEventsMiddleware.swift
```diff
@@ -18,41 +16,33 @@ struct UserChannelBanEventsMiddleware<ExtraData: ExtraDataTypes>: EventMiddlewar
...
                 memberDTO.isBanned = false
                 memberDTO.banExpiresAt = nil
-            } completion: { error in
-                if let error = error {
-                    log.error("Failed to write `UserUnbannedEvent` updates to the DB: \(error)")
-                }
-                completion(event)
+                
+            default:
+                break
```
DominikBucher12, line 39: maybe add here some assertion or some log that this is unhandled case?

DominikBucher12 (approved):
LGTM 👍

Sources/StreamChat/WebSocketClient/EventMiddlewares/UserChannelBanEventsMiddleware.swift
evsaev: Hmm 🤔 The only requirement for the incoming event to be processed is to have a `payload` with `cid`. There are many events that pass these rules besides `UserBannedEvent/UserUnbannedEvent` -> traping into `default` is not a failure but just a normal scenario -> I'd not put any log there.

Sources/StreamChat/WebSocketClient/EventMiddlewares/EventMiddleware.swift
```diff
@@ -6,42 +6,31 @@ import Foundation
...
-    func handle(event: Event, completion: @escaping (Event?) -> Void)
+    ///   - session: The database session the middleware works with.
+    /// - Returns: The original `event` passed via params OR `nil` if the incoming event was consumed by the middleware.
+    func handle(event: Event, session: DatabaseSession) -> Event?
 }
 
 extension Array where Element == EventMiddleware {
     /// Evaluates an array of `EventMiddleware`s in the order they're specified in the array. It's not guaranteed that
-    /// all middlewares are called. If a middleware returns `nil`, no middlewares down in the chain are called.
+    /// all middlewares are called. If a middleware calls the completion with `nil`, no middlewares down in the chain are called.
```
VojtaStavik, line 20: wasn't the original version actually the correct one? 🤔

Sources/StreamChat/Workers/EventNotificationCenter_Tests.swift
```diff
@@ -26,8 +26,9 @@ final class EventNotificationCenter_Tests: XCTestCase {
             .init()
         ]
         
-        // Create notication center with middlewares
-        let center = EventNotificationCenter(middlewares: middlewares, database: database)
+        // Create notification center with middlewares
+        let center = EventNotificationCenter(database: database)
+        middlewares.forEach(center.add)
```
VojtaStavik, line 32: Maybe we can just introduce a method for adding multiple middlewares at once? ;-)

VojtaStavik (commented):
Cool 👍 Just some small nits otherwise good to 🚢 🎉 Thank you

Issue #77, opened 2020-01-27
CIS-40 Automize release process with fastlane


b-onc:
I’ve checked how we can create a fastlane lane for our release flow, I’m dumping my findings here:
each step corresponds to the steps from `releasing.md`:
 1. ,  2.  `increment_version_number(version_number: <input>)` https://docs.fastlane.tools/actions/increment_version_number/
 3. `version_bump_podspec(path: <podspec>, version_number: <input>)` https://docs.fastlane.tools/actions/version_bump_podspec/
 4.  ,  5.  `git_add` , `add_git_tag(tag: <input>`),` git_commit, push_to_git_remote(tags: true)`  https://docs.fastlane.tools/actions/
 6.  ,  7.  `jazzy https://docs.fastlane.tools/actions/jazzy/`
 8. `git_add`, `git_commit`, `push_to_git_remote`
 9. this is a bit tricky. I’m considering adding a `latest_release_notes.md` file, which we’ll change for each release, and make fastlane parse this file’s contents, append it to CHANGELOG.md and also use the contents for github release: `set_github_release` https://docs.fastlane.tools/actions/set_github_release/
 10. `pod_push` https://docs.fastlane.tools/actions/pod_push/
obviously we can remove step 8 by moving the steps 6 and 7 up 1 step.

PR #952, opened 2021-04-02
Fix `teams` for `CurrentChatUser` and `ChatChannelMember`

nuno-vieira:
# In this PR
- Fixes `teams` property not being populated in `CurrentChatUser` and `ChatChannelMember`.
- Makes `User.init` without default arguments, to avoid future similar issues.


evsaev (approved):
It's great to have this pitfall fixed! 🎉 

Though there are some places where mocks from **StreamChatTestTools** could be used:

**ChannelController_SwiftUI/Combine_Tests:**
```swift
let typingMember = ChatChannelMember(
 id: .unique,
 name: .unique,
 ...
)
```
**MemberController_Combine_Tests**:
```swift
extension _ChatChannelMember {
  static var dummy: _ChatChannelMember { ... } // a duplication of mock
}
```
**UserController_SwiftUI_Tests**
```swift
extension _ChatUser {
 static var unique: _ChatUser { .mock } // .mock can be used directly instead of .unique
}
```

... I realized you didn't change those parts intentionally since `Use test tools where possible` feels like a separate PR 😄

Great job!

StreamChat.xcodeproj/project.pbxproj
```diff
@@ -1004,6 +1004,13 @@
 			remoteGlobalIDString = 799C941A247D2F80001F1104;
 			remoteInfo = StreamChat;
 		};
+		ADE787132617A410007129ED /* PBXContainerItemProxy */ = {
+			isa = PBXContainerItemProxy;
+			containerPortal = 8AD5EC8622E9A3E8005CFAC9 /* Project object */;
+			proxyType = 1;
+			remoteGlobalIDString = 793060E525778896005CF846;
+			remoteInfo = StreamChatTestTools;
+		};
```
evsaev, line 1013: nit: would be nice to have this commit as the first one so the reviewer expects `import StreamChatTestTools` later in the PR

Sources/StreamChat/Database/DTOs/CurrentUserDTO_Tests.swift
```diff
@@ -62,7 +62,8 @@ class CurrentUserModelDTO_Tests: XCTestCase {
             Assert.willBeEqual(mutedUserIDs, Set(loadedCurrentUser?.mutedUsers.map(\.id) ?? []))
             Assert.willBeEqual(payload.devices.count, loadedCurrentUser?.devices.count)
             Assert.willBeEqual(payload.devices.first?.id, loadedCurrentUser?.devices.first?.id)
-            // TODO: Teams
+            Assert.willBeEqual(payload.teams, loadedCurrentUser?.user.teams?.map(\.id))
+            Assert.willBeFalse(payload.teams.isEmpty)
```
evsaev, line 66: Do we need the insurance the test data (payload) is valid?

AFAIU it will fail if someone breaks the` CurrentUserPayload.dummy` by removing the `teams: ["GREEN", "RED"]` inside. 

Wouldn't it be nicer to update the `dummy` to get `teams: [String]` as input param and explicitly provide ["GREEN", "RED"] when we create a payload in **test_currentUserPayload_isStoredAndLoadedFromDB**?

Sources/StreamChat/Database/DTOs/MemberModelDTO_Tests.swift
```diff
@@ -72,6 +73,8 @@ class MemberModelDTO_Tests: XCTestCase {
             Assert.willBeEqual(payload.user.updatedAt, loadedMember?.userUpdatedAt)
             Assert.willBeEqual(payload.user.lastActiveAt, loadedMember?.lastActiveAt)
             Assert.willBeEqual(payload.user.extraData, loadedMember?.extraData)
+            Assert.willBeEqual(Set(payload.user.teams), loadedMember?.teams)
+            Assert.willBeFalse(payload.user.teams.isEmpty)
```
evsaev, line 77: not sure we need to verify that test data is valid since we create it manually.

Sources/StreamChat/Models/CurrentUser.swift
```diff
@@ -92,10 +93,12 @@ public class _CurrentChatUser<ExtraData: UserExtraData>: _ChatUser<ExtraData> {
             imageURL: imageURL,
             isOnline: isOnline,
             isBanned: isBanned,
+            isFlaggedByCurrentUser: false,
```
evsaev, line 96: Is it possible to flag yourself? 😄

Sources/StreamChat/WebSocketClient/EventMiddlewares/TypingStartCleanupMiddleware_Tests.swift
```diff
@@ -38,7 +39,7 @@ class TypingStartCleanupMiddleware_Tests: XCTestCase {
     
     func test_stopTypingEvent_sentAfterTimeout() {
         // Simulate some user started typing
-        let otherUser = ChatUser(id: UUID().uuidString)
+        let otherUser = _ChatUser<NoExtraData>.mock(id: UUID().uuidString)
```
evsaev, line 42:
```suggestion
        let otherUser: ChatUser = .mock(id: .unique)
```

nit: Should we provide a default value for `id` in ChatUser.mock if we are not interested in a specific value? 🤔

Sources/StreamChat/Models/CurrentUser.swift
nuno-vieira: That is a question that I have but I'm assuming you can't, that would be weird 😂

StreamChat.xcodeproj/project.pbxproj
nuno-vieira: Yes makes sense 👍 Although this deserves a PR on its own, I think it is not worth it adding new mocks to the StreamChat SDK, or new dummies, if we should be using the StreamChatTestTools. So in this case for it made sense to start using SteamChatTestTools for LLC too, so we can slowly start using it, and deprecating the old dummies/mocks etc..

Sources/StreamChat/Database/DTOs/MemberModelDTO_Tests.swift
nuno-vieira: Sorry what do you mean? 🤔

Sources/StreamChat/Database/DTOs/CurrentUserDTO_Tests.swift
nuno-vieira: ✅

Sources/StreamChat/WebSocketClient/EventMiddlewares/TypingStartCleanupMiddleware_Tests.swift
nuno-vieira: ✅

StreamChat.xcodeproj/project.pbxproj
nuno-vieira: ✅

Sources/StreamChat/Database/DTOs/MemberModelDTO_Tests.swift
evsaev: That's all about line #77 😄 Basically, the same thing mentioned in here https://github.com/GetStream/stream-chat-swift/pull/952#discussion_r607024331

Sources/StreamChat/Database/DTOs/CurrentUserDTO_Tests.swift
evsaev: I'd say we also don't need `Assert.willBeFalse(payload.teams.isEmpty)` since we are testing the input data here. If it fails it's a developer mistake, not related to `CurrentUserDTO`

Sources/StreamChat/Database/DTOs/MemberModelDTO_Tests.swift
nuno-vieira: ✅

Sources/StreamChat/Models/CurrentUser.swift
```diff
@@ -71,6 +71,7 @@ public class _CurrentChatUser<ExtraData: UserExtraData>: _ChatUser<ExtraData> {
         createdAt: Date = .init(),
         updatedAt: Date = .init(),
         lastActiveAt: Date? = nil,
+        teams: Set<TeamId> = [],
         extraData: ExtraData = .defaultValue,
         devices: [Device] = [],
         currentDevice: Device? = nil,
```
VojtaStavik, line 77: The `init` of `CurrentUser` shouldn't be `public`. I think we should remove the default values from this init, too.

Sources/StreamChat/Models/CurrentUser.swift
VojtaStavik: Maybe when you make a really stupid joke and you want to self-acknowledge that? 🤔

VojtaStavik (approved):
Looks good 👍 Thank you for fixing this. What do you think about the `CurrentChatUser` model init?

Sources/StreamChat/Models/CurrentUser.swift
nuno-vieira: Ah okay, I thought it was public for some reason. If it shouldn't be public, I'll remove it, and the default values too 👍

nuno-vieira: ✅

nuno-vieira: @VojtaStavik Fixed here: https://github.com/GetStream/stream-chat-swift/pull/952/commits/a66c2900f01c37e693892a939c5e273e55eac4b5

PR #959, opened 2021-04-06
Don't update collection view unless it's fully laid out

VojtaStavik:
Yet another try to fix the issue with incorrect collection view update...


b-onc (approved):
:shipit:

PR #960, opened 2021-04-07
[CIS-782] Use channel.membership instead of explicit currentUserId in ChatChannelListItem

dmigach:
We don't need to pass the value explicitly, it's already contained in `channel.membership`


Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
```diff
@@ -11,10 +11,14 @@ public typealias ChatChannelListItemView = _ChatChannelListItemView<NoExtraData>
 /// An `UIView` subclass that shows summary and preview information about a given channel.
 open class _ChatChannelListItemView<ExtraData: ExtraDataTypes>: _View, UIConfigProvider, SwiftUIRepresentable {
     /// The data this view component shows.
-    public var content: (channel: _ChatChannel<ExtraData>?, currentUserId: UserId?) {
+    public var content: _ChatChannel<ExtraData>? {
         didSet { updateContentIfNeeded() }
     }
 
+    private var channel: _ChatChannel<ExtraData>? {
```
dmigach, line 18: Just a minor readability improvement since there used to be a label in the tuple and now it's gone. Please let me know if you think it's redundant

Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
```diff
@@ -11,10 +11,14 @@ public typealias ChatChannelListItemView = _ChatChannelListItemView<NoExtraData>
 /// An `UIView` subclass that shows summary and preview information about a given channel.
 open class _ChatChannelListItemView<ExtraData: ExtraDataTypes>: _View, UIConfigProvider, SwiftUIRepresentable {
     /// The data this view component shows.
-    public var content: (channel: _ChatChannel<ExtraData>?, currentUserId: UserId?) {
```
dmigach, line 14: I wonder what's the policy in terms of breaking changes in API's. I believe there is a best practice that you can only break compilation when you release a major version, but ChatUI is not really in prod yet, right?

VojtaStavik: The public API is currently available only in beta, so it's OK to change it.

Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
VojtaStavik: I see what you mean but it's IMO not needed. We use the `content` everywhere and it's quite easy to get used to the pattern. I actually find seeing `channel` here more confusing than `content` :)

VojtaStavik (approved):
Looks great 👍 Congrats on your first PR here 🎉 Just a minor fix and we're good to merge.

Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
dmigach: Ok, let's keep it in sync with the rest of the codebase 👌

VojtaStavik:
@nuno-vieira are you interested in reviewing this or can we merge?

PR #961, opened 2021-04-07
Container view

LukasHromadnik:
# Description

In this PR a custom `UIStackView`-like container is implemented.

Main purpose of this container is to have more control over the layout inside stack-like containers.


VojtaStavik (approved):
Let's 🚢 this one 🎉

PR #962, opened 2021-04-07
Message layout options evaluation

evsaev:


Sources/StreamChatUI/ChatMessageList/RETHINK/MessageListVC.swift
```diff
@@ -50,7 +106,7 @@ class MessageListVC<ExtraData: ExtraDataTypes>: _ViewController, UICollectionVie
         let message = channelController.messages[indexPath.item]
         
         let reuseId = cellReuseIdentifier(for: message)
```
VojtaStavik, line 108: Should we also change this to the index path version?

VojtaStavik (approved):

Sources/StreamChatUI/ChatMessageList/RETHINK/MessageListVC.swift
evsaev: Most likely but decided not to touch the reuse logic part in this PR

PR #964, opened 2021-04-07
Message cell rendering

evsaev:


Sources/StreamChatUI/ChatMessageList/RETHINK/MessageCell.swift
```diff
@@ -86,3 +101,30 @@ class MessageCell<ExtraData: ExtraDataTypes>: _CollectionViewCell {
         return preferredAttributes
     }
 }
+
+private extension MessageCell {
+    func updateTextView() {
+        textView.textAlignment = content?.isSentByCurrentUser == true ? .right : .left
```
VojtaStavik, line 107: The text alignment is always left ;-)

Sources/StreamChatUI/ChatMessageList/RETHINK/MessageCell.swift
```diff
@@ -54,17 +58,28 @@ class MessageCell<ExtraData: ExtraDataTypes>: _CollectionViewCell {
...
-            containerView.addArrangedSubview(textView)
+            containerView.addArrangedSubview(textView, respectsLayoutMargins: true)
         }
     }
     
     override func updateContent() {
-        // TODO: this shouldn't intialize the text view
-        textView.text = content?.text
+        updateTextView()
+        updateBubbleView()
```
VojtaStavik, line 82: These "updateXXX" methods should work the way that they won't initialize the subview if not needed. Only the layout options should have the power of initializing the subviews.

VojtaStavik (approved):

Sources/StreamChatUI/ChatMessageList/RETHINK/MessageCell.swift
evsaev: hm, but those are private and are called from `updateContent` which is invoked when the view is added to the superview and has access to `uiconfig` 🤔 So it looks there's no chance to call them earlier than expected

Sources/StreamChatUI/ChatMessageList/RETHINK/MessageCell.swift
evsaev: lol 🤦

Sources/StreamChatUI/ChatMessageList/RETHINK/MessageCell.swift
VojtaStavik: What I mean is that this way `textView` will be initialized even when layout options don't have `.text`. We want to avoid this to keep everything as fast as possible.

evsaev: @VojtaStavik hey, could you have a look at the latest commit and confirm it's what we want?

Sources/StreamChatUI/ChatMessageList/RETHINK/MessageCell.swift
```diff
@@ -176,15 +183,21 @@ class MessageCell<ExtraData: ExtraDataTypes>: _CollectionViewCell, UIConfigProvi
...
+        let attachments: _ChatMessageAttachmentListViewData<ExtraData>? = content.map {
             .init(
                 attachments: $0.attachments.compactMap { $0 as? ChatMessageDefaultAttachment },
                 didTapOnAttachment: nil,
                 didTapOnAttachmentAction: nil
             )
         }
+
+        // Photo preview
+        photoPreviewView?.isVisible = layoutOptions?.contains(.photoPreview) ?? false
```
VojtaStavik, line 208: why do we need to do this? If the preview is not in layout options it won't be visible 🤷‍♂️

Sources/StreamChatUI/ChatMessageList/RETHINK/MessageCell.swift
```diff
@@ -121,6 +122,16 @@ class MessageCell<ExtraData: ExtraDataTypes>: _CollectionViewCell, UIConfigProvi
             ]
         }
 
+        // Giphy
+        if options.contains(.giphy) {
+            let giphyView = createGiphyView()
+            bubbleContainer.addArrangedSubview(giphyView, respectsLayoutMargins: false)
+            constraintsToActivate += [
+                // This is ugly. Ideally the photo preview should be updated to fill all available space.
+                giphyView.widthAnchor.constraint(equalToConstant: window!.bounds.width * 0.75).almostRequired
```
VojtaStavik, line 131: Yeah, let's figure it out :)

PR #965, opened 2021-04-07
[CIS-766] Add Working with Channel List guide

b-onc:


VojtaStavik (approved):
I think this is great 👍 I don't see any reasons why not to merge it for now. We will work on it later but it's better to have it in `main` than here :)

PR #967, opened 2021-04-07
[CIS-750] Audit `ChatMessageComposerImageAttachmentsView` 

nuno-vieira:
# In this PR

- Adds more testing images to StreamChatUITests
- Audits the ChatMessageComposerImageAttachmentsView


Sources/StreamChatUI/UIConfig+MessageComposer.swift
```diff
@@ -12,14 +12,29 @@ public extension _UIConfig {
...
+        /// The collection view layout of the image attachments collection view.
+        public var imageAttachmentsCollectionViewLayout: UICollectionViewFlowLayout.Type =
+            UICollectionViewFlowLayout.self
+        /// A view that displays the image attachment.
+        public var imageAttachmentCellView: _ChatMessageComposerImageAttachmentView<ExtraData>.Type =
+            _ChatMessageComposerImageAttachmentView<ExtraData>.self
+        /// The view cell that displays the image attachment.
+        public var imageAttachmentCollectionViewCell: _ChatMessageComposerImageAttachmentCollectionViewCell<ExtraData>.Type =
+            _ChatMessageComposerImageAttachmentCollectionViewCell<ExtraData>.self
+
```
VojtaStavik, line 37: Shouldn't some of these views be at the root and some shouldn't be here at all?

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerImageAttachmentsView.swift
```diff
@@ -5,44 +5,39 @@
...
     UIConfigProvider,
     UICollectionViewDelegate,
     UICollectionViewDataSource {
-    // MARK: - Properties
-    
-    open var imagePreviewItemSize: CGSize = .init(width: 100, height: 100)
-    
-    public var images: [UIImage] = [] {
+    /// The image attachment item size. Can be overridden in `setUpLayout()`.
+    private var imagePreviewItemSize: CGSize = .init(width: 100, height: 100)
```
VojtaStavik, line 17: If I understand it correctly, the height of the inner collection view is the same as the height of this component, right? This also limits the max height of the cell while the width of the cell should (eventually) be defined by the ratio of the attachment preview.

Seems to me that this view shouldn't have any inners sizing information and it's height can be set from the outside by its superview :)

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerImageAttachmentsView.swift
```diff
@@ -5,44 +5,39 @@
...
+
+    /// The images data source.
+    public var content: [UIImage] = [] {
         didSet {
             updateContentIfNeeded()
         }
```
VojtaStavik, line 23: It seems like this should be abstracted somehow. Rather than `UIImage` we should probably introduce some kind of protocol that any attachment can conform to. This will be extremely important for custom attachment types. 

Maybe something like this? We could also add more options to specify the source of the preview. Later we can extend it by `previewTitle` `previewSize` etc.
```
protocol PreviewProvider { 
  var previewImageURL: URL? { get }
  var previewImage: UIImage? { get }
}
```

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerImageAttachmentsView.swift
```diff
@@ -55,39 +50,43 @@ open class _ChatMessageComposerImageAttachmentsView<ExtraData: ExtraDataTypes>:
...
     override open func setUpLayout() {
-        collectionView.translatesAutoresizingMaskIntoConstraints = false
-        embed(collectionView)
-        
+        flowLayout.scrollDirection = .horizontal
         flowLayout.itemSize = imagePreviewItemSize
         flowLayout.sectionInset = layoutMargins
+
+        embed(collectionView, insets: .init(top: layoutMargins.top, leading: 0, bottom: layoutMargins.bottom, trailing: 0))
+        collectionView.heightAnchor.pin(equalToConstant: imagePreviewItemSize.height).isActive = true
```
VojtaStavik, line 64: I think this is not needed.

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerImageAttachmentCollectionViewCell.swift
```diff
@@ -5,57 +5,85 @@
...
-    // MARK: - Properties
-    
-    class var reuseId: String { String(describing: self) }
-    
+    open class var reuseId: String { String(describing: self) }
+
+    public private(set) lazy var imageAttachmentView = uiConfig
+        .messageComposer
+        .imageAttachmentCellView.init()
+        .withoutAutoresizingMaskConstraints
```
VojtaStavik, line 20: `open`

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerImageAttachmentsView.swift
```diff
@@ -5,44 +5,39 @@
...
+    /// The images data source.
+    public var content: [UIImage] = [] {
         didSet {
             updateContentIfNeeded()
         }
     }
-    
+
+    /// The closure handler when an image attachment has been removed.
     public var didTapRemoveItemButton: ((Int) -> Void)?
```
VojtaStavik, line 27: `open`

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerImageAttachmentsView.swift
```diff
@@ -5,44 +5,39 @@
...
-    // MARK: - Properties
-    
-    open var imagePreviewItemSize: CGSize = .init(width: 100, height: 100)
-    
-    public var images: [UIImage] = [] {
+    /// The image attachment item size. Can be overridden in `setUpLayout()`.
+    private var imagePreviewItemSize: CGSize = .init(width: 100, height: 100)
+
+    /// The images data source.
+    public var content: [UIImage] = [] {
```
VojtaStavik, line 20: `open`

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerImageAttachmentCollectionViewCell.swift
```diff
@@ -5,57 +5,85 @@
...
-    
+
+    /// The `ChatMessageComposerImageAttachmentView` layout constraints.
+    public struct Layout {
+        public var discardButtonConstraints: [NSLayoutConstraint] = []
+    }
+
+    /// The `ChatMessageComposerImageAttachmentView` layout constraints.
+    public private(set) var layout = Layout()
+
```
VojtaStavik, line 54: Do you think we still need this? Complex layouts will be done via `ContainerView`, for simple layouts like this, it's OK to not have it.

VojtaStavik (commented):
Looks good 👍 I noticed there's `public` instead of `open`. I also added a couple of questions that are not related to this specific component but to all of them in general.

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerImageAttachmentsView.swift
DominikBucher12: Just to fill in, this size is needed for `flowLayout.itemSize` to setup the size in the `collectionView` otherwise this just doesn't resize properly and we cannot see the cells... Also if this stays, I would definitely make it public and somehow highlight that this value basically defines height of the view... 

The last sentence feels a bit bad attitude, so maybe reconsider some bigger change as Vojta suggested.

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerImageAttachmentsView.swift
```diff
@@ -55,39 +50,43 @@ open class _ChatMessageComposerImageAttachmentsView<ExtraData: ExtraDataTypes>:
...
     
     override open func updateContent() {
         collectionView.reloadData()
     }
-
-    // MARK: - UICollectionView
     
     public func collectionView(_ collectionView: UICollectionView, numberOfItemsInSection section: Int) -> Int {
-        images.count
+        content.count
```
DominikBucher12, line 72: 👍

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerImageAttachmentsView.swift
```diff
@@ -55,39 +50,43 @@ open class _ChatMessageComposerImageAttachmentsView<ExtraData: ExtraDataTypes>:
...
             ) as? _ChatMessageComposerImageAttachmentCollectionViewCell<ExtraData>
-        else { return UICollectionViewCell() }
-        
-        cell.imageView.image = images[indexPath.row]
-        cell.discardButtonHandler = { [weak self] in self?.didTapRemoveItemButton?(indexPath.row) }
+        else {
+            return UICollectionViewCell()
+        }
+
+        cell.uiConfig = uiConfig
```
DominikBucher12, line 85: nit: Maybe overthinking, but I think it would be nice to put here some comment about `UIConfig` and why it is needed to pass it on `UITableViewCell` / `UICollectionViewCell` instances...

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerImageAttachmentsView.swift
```diff
@@ -55,39 +50,43 @@ open class _ChatMessageComposerImageAttachmentsView<ExtraData: ExtraDataTypes>:
...
-                withReuseIdentifier: _ChatMessageComposerImageAttachmentCollectionViewCell<ExtraData>.reuseId,
+                withReuseIdentifier: uiConfig.messageComposer.imageAttachmentCollectionViewCell.reuseId,
                 for: indexPath
             ) as? _ChatMessageComposerImageAttachmentCollectionViewCell<ExtraData>
-        else { return UICollectionViewCell() }
-        
-        cell.imageView.image = images[indexPath.row]
-        cell.discardButtonHandler = { [weak self] in self?.didTapRemoveItemButton?(indexPath.row) }
+        else {
+            return UICollectionViewCell()
```
DominikBucher12, line 82: Just don't know how much we care about this but when the casting fails during some change, we don't know why... maybe some assert or log for programmer error...

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerImageAttachmentCollectionViewCell.swift
```diff
@@ -5,57 +5,85 @@
...
+/// The view cell that displays the image attachment.
 open class _ChatMessageComposerImageAttachmentCollectionViewCell<ExtraData: ExtraDataTypes>: _CollectionViewCell,
     UIConfigProvider {
-    // MARK: - Properties
-    
-    class var reuseId: String { String(describing: self) }
-    
+    open class var reuseId: String { String(describing: self) }
+
+    public private(set) lazy var imageAttachmentView = uiConfig
```
DominikBucher12, line 17: Missing documentation?

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerImageAttachmentsView.swift
DominikBucher12: Actually it is needed, it resizes the whole collectionView according to the height of `itemSize`... Probably should be overwritten to something else...

DominikBucher12 (changes requested):
IMHO the logic around sizing the collectionView feels a bit strange... Maybe we can rethink this a bit.

Sources/StreamChatUI/UIConfig+MessageComposer.swift
nuno-vieira: Yes, as I said in Slack, I prefer to do these restructures when auditing the ChatMessageComposerVC since I will have a better overview of how the composer works, and so will make better decisions on the refactorings 🙂

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerImageAttachmentsView.swift
nuno-vieira: Yes like @DominikBucher12 this size is to ser the `collectionView.itemSize`. I also reuse it to set the collection view height. But you are right, the collection view can have constraints to the super view, and then the height is defined by the superview, which makes more sense 👍

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerImageAttachmentsView.swift
nuno-vieira: We also have this on the documents attachment view: 
image
So we have some other "preview types", maybe just calling this `ImagePreviewProvider` ? Also, if the customer wants to add his own type, he would need to cast to his type to use other properties and fill them in the updateContent right?  🤔

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerImageAttachmentsView.swift
nuno-vieira: What I think Vojta means is that the embed of the collection view is enough. And the height of the view should be defined by the super view, right ? @DominikBucher12 @VojtaStavik

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerImageAttachmentCollectionViewCell.swift
nuno-vieira: The reason I did this, is that somehow, and I couldn't find the reason, the only way to override properly the constraints for discard button was to deactivate them. If I didn't, and for example set the discard button to the leading, it would go to the middle of the view 🤷‍♂️

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerImageAttachmentsView.swift
VojtaStavik: Correct @nuno-vieira 👍 Basically, this component should just accept whatever size its superview suggests.

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerImageAttachmentsView.swift
nuno-vieira: I actually thought about creating our own extension to reuse cells, where we can simplify the use of the native `dequeueCell` and in that extension we provide the uiConfig always. Because this is already the 2nd time I lose a lot of time to understand why the uiconfig overriding is not working 😕 So if we have that always done, it will be much safer.

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerImageAttachmentsView.swift
```diff
@@ -55,39 +47,39 @@ open class _ChatMessageComposerImageAttachmentsView<ExtraData: ExtraDataTypes>:
...
-            ) as? _ChatMessageComposerImageAttachmentCollectionViewCell<ExtraData>
-        else { return UICollectionViewCell() }
-        
-        cell.imageView.image = images[indexPath.row]
-        cell.discardButtonHandler = { [weak self] in self?.didTapRemoveItemButton?(indexPath.row) }
+        let reuseId = uiConfig.messageComposer.imageAttachmentCollectionViewCell.reuseId
+        let cell = collectionView.dequeueReusableCell(
+            withReuseIdentifier: reuseId,
+            for: indexPath
+        ) as! _ChatMessageComposerImageAttachmentCollectionViewCell<ExtraData>
```
DominikBucher12, line 76: ok I guess this solves the logging issue :trollface:

DominikBucher12 (approved):
👍

VojtaStavik (approved):
Let's merge and finish the details once we have the full context.

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerImageAttachmentsView.swift
nuno-vieira: Ahah, yes I did this because we are doing this in all other places. So it is to keep consistency. And ofc this will only happen if the dev made a mistake, so no big deal

PR #969, opened 2021-04-08
[CIS-784] Fix channel unread count calculations for new messages

VojtaStavik:
This PR fixes a lot of small issues which together caused very strange behavior of channel unread count number.

**The most important changes:**
- Changes to `ChannelReadDTO` now trigger `ChannelDTO` updated callbacks
- `MessageReadEvent` incorrectly had a mandatory `unreadCount` parameter which made it impossible to parse and handle for other than the current user.
- After consulting with the backend team [in Slack](https://getstream.slack.com/archives/CE5N802GP/p1617885760181100?thread_ts=1617740759.146000&cid=CE5N802GP), `ChannelReadUpdaterMiddleware` now increases the read unread count for new incoming messages.

---

https://user-images.githubusercontent.com/6388510/114058792-10d66580-9894-11eb-99a6-4def4f109a78.mov


Sources/StreamChat/Models/Channel.swift
```diff
@@ -164,7 +165,7 @@ public struct _ChatChannel<ExtraData: ExtraDataTypes> {
         currentlyTypingMembers: Set<_ChatChannelMember<ExtraData.User>> = [],
         lastActiveWatchers: @escaping (() -> [_ChatUser<ExtraData.User>]) = { [] },
         team: TeamId? = nil,
-        unreadCount: ChannelUnreadCount = .noUnread,
+        unreadCount: @escaping () -> ChannelUnreadCount = { .noUnread },
```
b-onc, line 168: so `unreadCount` is not a value but a dynamic thingy now.. This makes us move further away from "snapshot models". Since we send `channelUpdated` delegate callbacks when unread count updates, why don't we still use static unread counts in models?

Sources/StreamChat/WebSocketClient/Events/MessageEvents.swift
```diff
@@ -59,19 +59,17 @@ public struct MessageDeletedEvent<ExtraData: ExtraDataTypes>: EventWithUserPaylo
...
-    public let unreadCount: UnreadCount
     
     let payload: Any
     
-    init(from response: EventPayload<ExtraData>) throws {
+    init<ExtraData: ExtraDataTypes>(from response: EventPayload<ExtraData>) throws {
         userId = try response.value(at: \.user?.id)
         cid = try response.value(at: \.cid)
         readAt = try response.value(at: \.createdAt)
-        unreadCount = try response.value(at: \.unreadCount)
```
b-onc, line 74: I'm confused, backend doesn't send this info anymore, or it never sent this, or this is not useful?

Sources/StreamChat/WebSocketClient/Events/NotificationEvents.swift
```diff
@@ -4,44 +4,44 @@
...
     let payload: Any
     
-    init(from response: EventPayload<ExtraData>) throws {
+    init<ExtraData: ExtraDataTypes>(from response: EventPayload<ExtraData>) throws {
         userId = try response.value(at: \.message?.user.id)
         cid = try response.value(at: \.channel?.cid)
         messageId = try response.value(at: \.message?.id)
         createdAt = try response.value(at: \.message?.createdAt)
-        unreadCount = try response.value(at: \.unreadCount)
+        unreadCount = try? response.value(at: \.unreadCount)
```
b-onc, line 20: In which cases backend doesn't send this value? would be nice to comment this

b-onc (commented):
I'm really not sure about `unreadCount` becoming a `Cached` value...

Sources/StreamChat/Models/Channel.swift
VojtaStavik: Nope, this just means it's computed lazily. Getting unread count requires iterating over channel reads + an additional DB query. That's why it's computed lazily. See other `@Cached` properties in our models.

Sources/StreamChat/WebSocketClient/Events/MessageEvents.swift
VojtaStavik: This was actually a complete misunderstanding of how `MessageReadEvent` work :) The unread count is always zero hence is not sent. This unread count is the total unread count and is sent only for the current user -> see the PR description for more info.

Sources/StreamChat/Models/Channel.swift
b-onc: OK, explanation + re-reading code made it clearer. Makes sense 👌

Sources/StreamChat/WebSocketClient/Events/MessageEvents.swift
b-onc: Shouldn't it be optional and useful when available? If it exists, we can use it to update current user's unread count

Sources/StreamChat/WebSocketClient/Events/NotificationEvents.swift
VojtaStavik: There are several conditions. See https://getstream.slack.com/archives/CE5N802GP/p1617740759146000 for more info. This whole thing about unread counts is still not ideal and will need a rethink.

b-onc (approved):

PR #970, opened 2021-04-09
[CIS-528] Add connection status banner to DemoApp

dmigach:
Mostly this change is about:
* Adding a view for displaying the banner
* Fixing an issue with `internetIsOfflineError` that prevented web socket client from reconnecting
* Fixing URLSession configuration to make reconnection logic work
* Implementing banner logic in DemoApp
* Rehauling navigation in DemoApp a bit to support presenting banner on any screen
banner

### DIFF
--- a/DemoApp/BannerShowingConnectionDelegate.swift
+++ b/DemoApp/BannerShowingConnectionDelegate.swift
@@ -1,0 +1,78 @@
+//
+// Copyright © 2021 Stream.io Inc. All rights reserved.
+//
+
+import StreamChat
+import UIKit
+
+final class BannerShowingConnectionDelegate {
+    // MARK: - Private Properties
+    
+    private let view: UIView
+    private let bannerView = BannerView()
+    private let bannerAppearanceDuration: TimeInterval = 0.5
+    
+    // MARK: -
+    
+    init(showUnder view: UIView) {
+        self.view = view
+        setupViews()
+    }
+}
+
+// MARK: - ChatConnectionControllerDelegate
+
+extension BannerShowingConnectionDelegate: ChatConnectionControllerDelegate {
+    public func connectionController(_ controller: ChatConnectionController, didUpdateConnectionStatus status: ConnectionStatus) {
+        switch status {
+        case .disconnected:
+            showBanner()
+        case .connected:
+            hideBanner()
+        case .initialized,
+             .disconnecting,
+             .connecting:
+            break
+        }
+    }
+}
+
+// MARK: - Private Methods
+
+private extension BannerShowingConnectionDelegate {
+    func setupViews() {
+        attachToTopViewIfNeeded()
+        bannerView.alpha = 0
+        bannerView.update(text: "Connecting...")
+    }
+    
+    func attachToTopViewIfNeeded() {
+        guard bannerView.superview != view else { return }
+        
+        view.addSubview(bannerView)
+        
+        bannerView.translatesAutoresizingMaskIntoConstraints = false
+        NSLayoutConstraint.activate(
+            [
+                bannerView.topAnchor.constraint(equalTo: view.bottomAnchor),
+                bannerView.widthAnchor.constraint(equalTo: view.widthAnchor),
+                bannerView.heightAnchor.constraint(equalToConstant: 28)
+            ]
+        )
+    }
+    
+    func showBanner() {
+        attachToTopViewIfNeeded()
+        animateBannerAlpha(to: 1)
+    }
+    
+    func hideBanner() {
+        animateBannerAlpha(to: 0)
+    }
+    
+    func animateBannerAlpha(to value: CGFloat) {
+        UIView.animate(withDuration: bannerAppearanceDuration) {
+            self.bannerView.alpha = value
+        }
+    }
+}
--- a/DemoApp/BannerView.swift
+++ b/DemoApp/BannerView.swift
@@ -8,28 +8,30 @@ import UIKit
 final class BannerView: UIView {
     private let label = UILabel()
     
-    override func didMoveToSuperview() {
-        super.didMoveToSuperview()
-        setUpLayout()
-        defaultAppearance()
+    override init(frame: CGRect) {
+        super.init(frame: frame)
+        commonInit()
+    }
+    
+    @available(*, unavailable)
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
     }
     
     func update(text: String) {
         label.text = text
     }
     
+    private func commonInit() {
+        setUpLayout()
+        defaultAppearance()
+    }
+    
     private func setUpLayout() {
-        guard let superview = superview else { return }
-        
         addSubview(label)
-        
-        translatesAutoresizingMaskIntoConstraints = false
         label.translatesAutoresizingMaskIntoConstraints = false
-        
         NSLayoutConstraint.activate(
             [
-                widthAnchor.constraint(equalTo: superview.widthAnchor),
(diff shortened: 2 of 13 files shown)


DemoApp/BannerShowingConnectionDelegate.swift
```diff
@@ -0,0 +1,73 @@
...
+//
+
+import StreamChat
+import UIKit
+
+final class BannerShowingConnectionDelegate {
+    // MARK: - Private Properties
+    
+    private let navigationController: UINavigationController
+    private var connectionEstablishmentTime: Date?
```
dmigach, line 12: This property improves UX by allowing us to ignore momentary states changes, plus Vojta mentioned that ConnectionController might fire several consecutive connection/disconnection events. 
So this way we can throttle banner appearance

DemoApp/BannerShowingConnectionDelegate.swift
```diff
@@ -0,0 +1,73 @@
...
+        switch status {
+        case .connected:
+            connectionEstablishmentTime = Date()
+        case .disconnecting:
+            break
+        case .connecting:
+            if let time = connectionEstablishmentTime {
+                let elapsedTime = time.distance(to: Date())
+                if elapsedTime > 5 {
+                    // TODO: We need to reset the property we make desicion on whether to show the banner or not
```
dmigach, line 34: I left this comment because I'm not sure that `.connected` event ever fires second time for example when you recover after a disconnect, but I'm going to double check that

DemoApp/BannerShowingConnectionDelegate.swift
```diff
@@ -0,0 +1,73 @@
...
+
+// MARK: - Private Methods
+
+private extension BannerShowingConnectionDelegate {
+    func showBanner() {
+        guard let view = navigationController.topViewController?.view else { return }
+        let bannerView = BannerView()
+        view.addSubview(bannerView)
+        bannerView.alpha = 0
+        bannerView.update(text: "Reconnecting...")
```
dmigach, line 52: Is it fine to leave this unlocalised?

DemoApp/BannerShowingConnectionDelegate.swift
```diff
@@ -0,0 +1,73 @@
...
+        
+        UIView.animate(withDuration: 0.5) {
+            bannerView.alpha = 1
```
dmigach, line 55: Actually Figma says the banner should slide in from under the nav bar and then slide back, but it isn't that straightforward to implement, so for now I made if fade in/fade out

DemoApp/DemoAppCoordinator.swift
```diff
@@ -0,0 +1,53 @@
+//
+// Copyright © 2021 Stream.io Inc. All rights reserved.
+//
+
+import StreamChat
+import StreamChatUI
+import UIKit
+
+final class DemoAppCoordinator {
+    static var shared: DemoAppCoordinator!
```
dmigach, line 10: It is this ugly for two reasons:
* Initial nav controller lives in a storyboard
* LoginViewController needs that to present Chat 
* We through out LogicViewController as soon as we present Chat

DemoApp/BannerShowingConnectionDelegate.swift
```diff
@@ -0,0 +1,73 @@
...
+        }
+    }
+}
+
+// MARK: - Private Methods
+
+private extension BannerShowingConnectionDelegate {
+    func showBanner() {
+        guard let view = navigationController.topViewController?.view else { return }
+        let bannerView = BannerView()
```
dmigach, line 49: Just realised this needs to be changed a bit, we need to only create it once and then show/hide it

Sources/StreamChat/ChatClient.swift
```diff
@@ -193,7 +193,7 @@ public class _ChatClient<ExtraData: ExtraDataTypes> {
     /// required header auth parameters to make a successful request.
     private var urlSessionConfiguration: URLSessionConfiguration {
         let config = URLSessionConfiguration.default
-        config.waitsForConnectivity = true
+        config.waitsForConnectivity = false
```
VojtaStavik, line 196: This change should be covered by a test since it's a requirement for the system to work correctly.

VojtaStavik (changes requested):
Looks good 👍 I left a couple of comments and questions...

VojtaStavik:
WHAT? Almost all my comments disappeared 😭

DemoApp/BannerView.swift
```diff
@@ -0,0 +1,47 @@
...
+import UIKit
+
+final class BannerView: UIView {
+    private let label = UILabel()
+    
+    override func didMoveToSuperview() {
+        super.didMoveToSuperview()
+        setUpLayout()
+        defaultAppearance()
+    }
```
VojtaStavik, line 15: This a pattern from the SDK and I don't think we should use it here. The SDK for example takes care of the fact that this is called just once, which here doesn't happen. I think it's better to call these methods from a method which is guaranteed to be called just once.

DemoApp/BannerView.swift
```diff
@@ -0,0 +1,47 @@
...
+    func update(text: String) {
+        label.text = text
+    }
+    
+    private func setUpLayout() {
+        guard let superview = superview else { return }
+        
+        addSubview(label)
+        
+        translatesAutoresizingMaskIntoConstraints = false
```
VojtaStavik, line 26: The view's superview should decide about this, not the view itself.

DemoApp/BannerShowingConnectionDelegate.swift
```diff
@@ -0,0 +1,77 @@
...
+        
+        topView.addSubview(bannerView)
+        
+        let guide = topView.safeAreaLayoutGuide
+        NSLayoutConstraint.activate(
+            [
+                bannerView.topAnchor.constraint(equalTo: guide.topAnchor)
+            ]
+        )
+    }
```
VojtaStavik, line 61: Rather than messing up with the view, can't we just add the banner to the navbar view?

DemoApp/DemoAppCoordinator.swift
VojtaStavik: If it's ugly why don't you fix it? 🤔 🤷‍♂️

dmigach: What I meant is that we don't have much choice here, so it's ugly comparing to conventional coordinators.
Or do you mean fixing the reason I enumerated? I believe you mentioned something about storyboard being used on purpose

dmigach: Reworked it

Sources/StreamChat/ChatClient.swift
dmigach: Added a test

DemoApp/BannerView.swift
dmigach: Now it's called once from init

DemoApp/BannerView.swift
dmigach: Reworked

DemoApp/BannerShowingConnectionDelegate.swift
dmigach: Now it works like this. I was having doubts on this approach when I first considered how to implement it because it is based on an assumption that navbar has clipsToBounds == false and also that it has the highest Z position, but probably it should be fine for the DemoApp. Anyway it's easy to rework if there will be issues

VojtaStavik (approved):
Great 🎉

VojtaStavik:
@dmigach please merge when you're ready 👍

PR #972, opened 2021-04-09
[CIS-778] Call `markRead` on `ChatVC.viewDidAppear`

VojtaStavik:
All work was done in #969, this is just the final piece 🍒 🍰


DominikBucher12 (approved):
LGTM, however I am a bit confused why you didn't implement the scrollView delegate to `_ChatMessageListVC` and the implementation is not on scroll... Perhaps it is a bit more complex to implement?

b-onc (approved):

PR #974, opened 2021-04-09
[CIS-783] Fix missing `startObservingIfNeeded` calls in some controllers

VojtaStavik:
This was a bug in `UserListController` reported by a customer. I added tests for this case to all controllers and found the same issue in more controllers.


b-onc (approved):
I was so close to finding this bug during my debugging phase in `ChannelListController` but I tangent-missed it :D I did realize we wouldn't call `startObservingIfNeeded` if we call `synchronize` before accessing the properties but I incorrectly assumed they'd be called when user accesses properties in delegate callbacks.
Thanks for finding it! :shipit:

PR #975, opened 2021-04-09
Message cell actions forwarding

evsaev:


Sources/StreamChatUI/ChatMessageList/RETHINK/MessageCell.swift
```diff
@@ -14,6 +24,15 @@ class MessageCell<ExtraData: ExtraDataTypes>: _CollectionViewCell, UIConfigProvi
         }
     }
 
+    var delegate: Delegate? {
+        didSet {
+            // We need this since the action closures for `photo/file` previews
+            // are set up in `updateContent`. After dequeuing the cell we set content AND the delegate which results
+            // in 2 `updateContent` executions and should be fixed/reimplemented.
+            updateContentIfNeeded()
```
VojtaStavik, line 32: I have a WIP solution for this somewhere. Let's me find it and finish it :)

Sources/StreamChatUI/ChatMessageList/RETHINK/MessageListVC.swift
```diff
@@ -256,7 +256,24 @@ class MessageListVC<ExtraData: ExtraDataTypes>: _ViewController, UICollectionVie
...
+            didTapOnAttachment: { [weak self] in
+                self?.handleTapOnAttachment($0, forCellAt: indexPath)
+            },
+            didTapOnAttachmentAction: { [weak self] attachment, action in
+                self?.handleTapOnAttachmentAction(action, for: attachment, forCellAt: indexPath)
+            },
+            didTapOnQuotedMessage: { [weak self] in
+                self?.handleTapOnQuotedMessage($0, forCellAt: indexPath)
+            }
+        )
```
VojtaStavik, line 276: I think we can provide some kind of `setDelegate` method that will wrap this boilerplate

Sources/StreamChatUI/ChatMessageList/RETHINK/MessageListVC.swift
```diff
@@ -392,6 +409,63 @@ class MessageListVC<ExtraData: ExtraDataTypes>: _ViewController, UICollectionVie
             reactionsController: reactionsController
         )
     }
+
+    private func restartUploading(for attachment: ChatMessageDefaultAttachment) {
```
VojtaStavik, line 413: Why `private`?

VojtaStavik (approved):

Sources/StreamChatUI/ChatMessageList/RETHINK/MessageListVC.swift
evsaev: haha, that is not that straightforward. I find it more flexible to work with `indexPath` instead of `cell/message` but the index path is a context variable which we capture via action closures. We can have a dedicated helper on VC level but on cell level it might look weired

Sources/StreamChatUI/ChatMessageList/RETHINK/MessageListVC.swift
evsaev: hmm, didn't expect this to be changed/overriden in client code 🤔 At the same time we are `open by default` so let's update the access level

PR #982, opened 2021-04-12
Message list thread

LukasHromadnik:
# Description

This PR creates detail of thread.

* Message cell now contains thread reply button to display the detail
* Messages in ThreadVC don't contain `threadInfo` layout option
* Only replies are used in ThreadVC as dataSource. Parent message is inserted as a subview to the collectionView and its content has inset equal to the parent message height + spacing

### DIFF
--- a/Sources/StreamChatUI/ChatMessageList/RETHINK/DefaultMessageLayoutOptionsResolver.swift
+++ b/Sources/StreamChatUI/ChatMessageList/RETHINK/DefaultMessageLayoutOptionsResolver.swift
@@ -5,10 +5,26 @@
 import StreamChat
 
 public typealias _LayoutOptionsResolver<ExtraData: ExtraDataTypes> =
-    (_ message: _ChatMessage<ExtraData>, _ isLastInGroup: Bool) -> ChatMessageLayoutOptions
+    (_ indexPath: IndexPath, _ messages: [_ChatMessage<ExtraData>]) -> ChatMessageLayoutOptions
 
-public func DefaultLayoutOptionsResolver<ExtraData: ExtraDataTypes>() -> _LayoutOptionsResolver<ExtraData> {
-    { message, isLastInGroup in
+public func DefaultLayoutOptionsResolver<ExtraData: ExtraDataTypes>(
+    minTimeIntervalBetweenMessagesInGroup: Double = 10
+) -> _LayoutOptionsResolver<ExtraData> {
+    { indexPath, messages in
+        let message = messages[indexPath.item]
+
+        let isLastInGroup: Bool = {
+            guard indexPath.item > 0 else { return true }
+            
+            let nextMessage = messages[indexPath.item - 1]
+
+            guard nextMessage.author == message.author else { return true }
+            
+            let delay = nextMessage.createdAt.timeIntervalSince(message.createdAt)
+            
+            return delay > minTimeIntervalBetweenMessagesInGroup
+        }()
+        
         var options: ChatMessageLayoutOptions = []
         
         if message.isSentByCurrentUser {
--- a/Sources/StreamChatUI/ChatMessageList/RETHINK/DefaultMessageLayoutOptionsResolver.swift
+++ b/Sources/StreamChatUI/ChatMessageList/RETHINK/DefaultMessageLayoutOptionsResolver.swift
@@ -1,0 +1,81 @@
+//
+// Copyright © 2021 Stream.io Inc. All rights reserved.
+//
+
+import StreamChat
+
+public typealias _LayoutOptionsResolver<ExtraData: ExtraDataTypes> =
+    (_ message: _ChatMessage<ExtraData>, _ isLastInGroup: Bool) -> ChatMessageLayoutOptions
+
+public func DefaultLayoutOptionsResolver<ExtraData: ExtraDataTypes>() -> _LayoutOptionsResolver<ExtraData> {
+    { message, isLastInGroup in
+        var options: ChatMessageLayoutOptions = []
+        
+        if message.isSentByCurrentUser {
+            options.insert(.flipped)
+        }
+        if !isLastInGroup {
+            options.insert(.continuousBubble)
+        }
+        if !isLastInGroup && !message.isSentByCurrentUser {
+            options.insert(.avatarSizePadding)
+        }
+        if isLastInGroup {
+            options.insert(.metadata)
+        }
+        if !message.textContent.isEmpty {
+            options.insert(.text)
+        }
+        
+        guard message.deletedAt == nil else {
+            return options
+        }
+        
+        if isLastInGroup && !message.isSentByCurrentUser {
+            options.insert(.avatar)
+        }
+        if message.quotedMessageId != nil {
+            options.insert(.quotedMessage)
+        }
+        if message.isPartOfThread {
+            options.insert(.threadInfo)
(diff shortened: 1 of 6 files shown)


Sources/StreamChatUI/ChatMessageList/RETHINK/MessageCell.swift
```diff
@@ -29,10 +29,12 @@ class MessageCell<ExtraData: ExtraDataTypes>: _CollectionViewCell, UIConfigProvi
     var reactionsView: _ReactionsCompactView<ExtraData>?
     var errorIndicatorView: _ChatMessageErrorIndicator<ExtraData>?
 
-    var threadReplyCountLabel: UILabel?
+    var threadReplyCountButton: UIButton?
     var threadAvatarView: ChatAvatarView?
     var threadArrowView: _SimpleChatMessageThreadArrowView<ExtraData>?
     
+    var onThreadReplyTapped: (() -> Void)!
+
```
VojtaStavik, line 37: I think @evsaev already solved this? 🤔

Sources/StreamChatUI/ChatMessageList/RETHINK/MessageThreadVC.swift
```diff
@@ -0,0 +1,323 @@
...
+// Copyright © 2021 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+import StreamChat
+import UIKit
+
+class MessageThreadVC<ExtraData: ExtraDataTypes>: _ViewController, UICollectionViewDelegate, UICollectionViewDataSource,
+    UIConfigProvider {
+    var channelController: _ChatChannelController<ExtraData>!
```
VojtaStavik, line 11: I don't think we ween the channel controller

VojtaStavik (approved):
OK, there seems to be some redundancy with already merged the work by @evsaev. -> let's talk in Slack

Sources/StreamChatUI/ChatMessageList/RETHINK/MessageCell.swift
evsaev: Yeah, this action has already been forwarded to the controller via **MessageCell.Delegate.didTapOnThread**. Now it's just the matter of changing the implementation of `handleTapOnThread(forCellAt:)` in `MessageListVC`

Sources/StreamChatUI/ChatMessageList/RETHINK/DefaultMessageLayoutOptionsResolver.swift
```diff
@@ -0,0 +1,81 @@
+//
+// Copyright © 2021 Stream.io Inc. All rights reserved.
+//
+
+import StreamChat
+
+public typealias _LayoutOptionsResolver<ExtraData: ExtraDataTypes> =
+    (_ message: _ChatMessage<ExtraData>, _ isLastInGroup: Bool) -> ChatMessageLayoutOptions
```
VojtaStavik, line 8: I think this is too restricting. The "group" is what should happen inside the resolver. Maybe the API can rather look something like this:
```swift
public typealias MessageLayoutOptionsResolver = (_ indexPath: IndexPath, _ message: (IndexPath) -> Message?) -> ChatMessageLayoutOptions
```
which would make it possible to access all messages you need.

The default resolver could look like:
```swift
 func DefaultMessageLayoutOptionsResolver(
    minTimeInteravlBetweenMessagesInGroup: TimeInterval = 10
) -> MessageLayoutOptionsResolver
```

Sources/StreamChatUI/ChatMessageList/RETHINK/MessageListVC.swift
```diff
@@ -57,6 +57,8 @@ class MessageListVC<ExtraData: ExtraDataTypes>: _ViewController, UICollectionVie
 //    public lazy var router = uiConfig.navigation.messageListRouter.init(rootViewController: self)
     public lazy var router = MessageListRouter(rootViewController: self)
     
+    public private(set) lazy var layoutOptionsResolver = uiConfig.messageList.layoutOptionsResolver
+
```
VojtaStavik, line 61: This doesn't need to be stored

Sources/StreamChatUI/ChatMessageList/RETHINK/MessageListVC.swift
```diff
@@ -167,80 +169,6 @@ class MessageListVC<ExtraData: ExtraDataTypes>: _ViewController, UICollectionVie
         return delay > minTimeIntervalBetweenMessagesInGroup
     }
     
-    func cellLayoutOptionsForMessage(at indexPath: IndexPath) -> ChatMessageLayoutOptions {
```
VojtaStavik, line 166: As mentioned in Slack "it's important to keep the cellLayoutOptionsForMessage method so users can easily override it"

Sources/StreamChatUI/ChatMessageList/RETHINK/MessageThreadVC.swift
```diff
@@ -225,7 +157,10 @@ class MessageThreadVC<ExtraData: ExtraDataTypes>: _ViewController, UICollectionV
         let message = chatMessage(for: indexPath)
         
         let reuseId = cellReuseIdentifier(for: message)
-        let layoutOptions = cellLayoutOptionsForMessage(at: indexPath)
+        let isLastInGroup = isMessageLastInGroup(at: indexPath)
+        var layoutOptions = layoutOptionsResolver(message, isLastInGroup)
+        layoutOptions.remove(.threadInfo)
+        layoutOptions.remove(.continuousBubble)
```
VojtaStavik, line 163: This will break grouping for threads which is not correct.

Sources/StreamChatUI/ChatMessageList/RETHINK/MessageThreadVC.swift
LukasHromadnik: I think it's needed in `MessageComposerVC`

Sources/StreamChatUI/ChatMessageList/RETHINK/DefaultMessageLayoutOptionsResolver.swift
LukasHromadnik: updated ✅

Sources/StreamChatUI/ChatMessageList/RETHINK/MessageListVC.swift
LukasHromadnik: 👍

Sources/StreamChatUI/ChatMessageList/RETHINK/MessageListVC.swift
LukasHromadnik: ✅

Sources/StreamChatUI/ChatMessageList/RETHINK/MessageThreadVC.swift
LukasHromadnik: fixed ✅

Sources/StreamChatUI/ChatMessageList/RETHINK/MessageCell.swift
LukasHromadnik: Let's merge the correct long press implementation first, then I will rebase it and connect the implementation with the above solution

PR #983, opened 2021-04-12
Use ContainerView in ChatChannelListItemView

DominikBucher12:
# What this PR do:
- Uses `ContainerView` inside `ChatChannelListItemView` instead of constrained layout. This gives us more flexibility in the code and also fixe some minor issues with the layout in clone applications.
# How to test it
- Run test applications, see if they look OK, see if CI doesn't fail and mostly check the DEMO application. This replacement is actually fixing issue with `unreadCountView` being wrongly aligned when there is no title to the channel.-


Sources/StreamChatUI/CommonViews/ContainerView.swift
```diff
@@ -28,15 +28,16 @@ extension UILayoutPriority {
...
 
 public class ContainerView: UIView {
-    convenience init(
+
+    public convenience init(
         axis: NSLayoutConstraint.Axis = .horizontal,
         alignment: Alignment = .fill,
         spacing: CGFloat = .auto,
-        views: [UIView] = []
+        views: [UIView]
```
VojtaStavik, line 40: This doesn't look correct :)

Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
```diff
@@ -134,106 +136,38 @@ open class _ChatChannelListItemView<ExtraData: ExtraDataTypes>: _View, UIConfigP
...
         super.setUpLayout()
+
+        /// Default layout:
+        /// ```
+        /// |----------------------------------------------------|
+        /// |            | titleLabel          | unreadCountView |
+        /// | avatarView | --------------------------------------|
+        /// |            | subtitleLabel        | timestampLabel |
+        /// |----------------------------------------------------|
+        /// ```
```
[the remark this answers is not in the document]
VojtaStavik: I think it should rather be this https://github.com/GetStream/stream-chat-swift/pull/983/commits/fa7fb10bb7b36354363ed6bcd6a8c261cc6bdddf#diff-edd8fac22492735b2760382253d2edc80d5bdb2047c54af717267f882c558b79R21-R31 that should be deleted

iMessageClone/iMessageChatChannelListItemView.swift
```diff
@@ -24,34 +24,15 @@ final class iMessageChatChannelListItemView: ChatChannelListItemView {
     
     override func setUpLayout() {
         super.setUpLayout()
-        
+
+        // Add left unread view blue dot.
         unreadView.translatesAutoresizingMaskIntoConstraints = false
         addSubview(unreadView)
```
VojtaStavik, line 30: Why not insert this as the 1st view of the main container?

VojtaStavik (approved):
Cool 👍

Sources/StreamChatUI/CommonViews/ContainerView.swift
DominikBucher12: Haha now I see what was in the conflict while rebasing :D Fixed now...

Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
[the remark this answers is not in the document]
DominikBucher12: Deleted.

iMessageClone/iMessageChatChannelListItemView.swift
DominikBucher12: FIxed

Sources/StreamChatUI/ChatChannelList/ChatChannelUnreadCountView.swift
```diff
@@ -47,7 +52,7 @@ open class _ChatChannelUnreadCountView<ExtraData: ExtraDataTypes>: _View, UIConf
     }
     
     override open func updateContent() {
-        isHidden = content.mentionedMessages == 0 && content.messages == 0
+        isHidden = shouldBeHidden
```
nuno-vieira, line 55: How do we need two properties? Can't we just have:
```
open var isHidden: Bool {
        content.mentionedMessages == 0 && content.messages == 0
    }
```

nuno-vieira: Hum I see, isHidden is from UIView. But the naming here is a bit confusing and redundant.
I think this is enough: 
```
isHidden = content == .noUnread
```
We don't need an extra property. Because if the customer can access `isHidden` and `shouldBeHidden`, what that means? Simply accessing the `.content` property and check that it is `noUnread` is enough

nuno-vieira (changes requested):
Looks good in general 👍  Just made one comment. And also, there might have conflicts in ContainerView which now is `ContainerStackView`. Requested changes, just to keep sure the resolve of the conflicts is properly solved, since it might be tricky

Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
```diff
@@ -17,20 +17,16 @@ open class _ChatChannelListItemView<ExtraData: ExtraDataTypes>: _View, UIConfigP
...
-        /// Constraints of `unreadCountView`
-        public fileprivate(set) var unreadCountViewConstraints: [NSLayoutConstraint] = []
-    }
+
+    /// Main container which holds `avatarView` and two horizontal containers `title` and `unreadCount` and
+    /// `subtitle` and `timestampLabel`
+    open private(set) lazy var mainContainer: ContainerView = ContainerView()
+
+    /// By default contains `title` and `unreadCount`.
+    open private(set) lazy var topContainer: ContainerView = ContainerView()
```
fortmarek, line 26: I'd maybe add here a comment about the fact that this is above `mainContainer`. Although it has `top` in the name, it might not be immediately obvious in relation to which element it is "top". Same for `bottomContainer`

SlackClone/SlackChatChannelListItemView.swift
```diff
@@ -8,47 +8,20 @@ import UIKit
...
-        
-        NSLayoutConstraint.deactivate(layout.unreadCountViewConstraints)
-        NSLayoutConstraint.activate([
-            unreadCountView.topAnchor.constraint(equalTo: timestampLabel.bottomAnchor),
-            unreadCountView.trailingAnchor.constraint(equalTo: topStackView.trailingAnchor)
-        ])
-        
+
+        // Switch unreadCount and timestamp
+        topContainer.insertArrangedSubview(timestampLabel, at: 1)
```
fortmarek, line 18: can't this be just `addArrangedSubview`?

Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
```diff
@@ -134,106 +127,38 @@ open class _ChatChannelListItemView<ExtraData: ExtraDataTypes>: _View, UIConfigP
...
-                + layout.timestampLabelConstraints
-                + layout.titleLabelConstraints
-                + layout.subtitleLabelConstraints
-                + layout.unreadCountViewConstraints
-        )
+        mainContainer.addArrangedSubviews([
+            avatarView, ContainerView(axis: .vertical, spacing: 4, views: [topContainer, bottomContainer])
+        ])
+        
+        mainContainer.alignment = .center
```
fortmarek, line 157: Is there a reason why this is not specified in its initializer? If user override `mainContainer` and initialize it with `ContainerView(alignment: .axisLeading)` then it will be overridden by this which might be a little bit counterintuitive.

Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
```diff
@@ -134,106 +127,38 @@ open class _ChatChannelListItemView<ExtraData: ExtraDataTypes>: _View, UIConfigP
...
+        mainContainer.isLayoutMarginsRelativeArrangement = true
+        
+        embed(mainContainer.withoutAutoresizingMaskConstraints)
```
fortmarek, line 160: afaict `withoutAutoresizingMaskConstraints` is usually defined right after declaring the view - but it's just a matter of convention (and I might be wrong)

Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
```diff
@@ -134,106 +127,38 @@ open class _ChatChannelListItemView<ExtraData: ExtraDataTypes>: _View, UIConfigP
...
+        mainContainer.pin(to: self)
```
fortmarek, line 161: how does this differ from `embed`?

Sources/StreamChatUI/CommonViews/ContainerView.swift
```diff
@@ -5,7 +5,7 @@
 import UIKit
 
 extension UIView {
-    func flexible(axis: NSLayoutConstraint.Axis) -> Self {
+    public func flexible(axis: NSLayoutConstraint.Axis) -> Self {
```
fortmarek, line 8: does this need to be public? I'd probably be a little bit surprised to find new methods on `UIView` after importing `StreamChatUI`. But maybe we do this already somewhere else. Also, if it shall stay public, it should be documented. The same comment is for `CGFloat.auto`

Sources/StreamChatUI/CommonViews/ContainerView.swift
```diff
@@ -28,11 +28,11 @@ extension UILayoutPriority {
 }
 
 extension CGFloat {
-    static let auto: CGFloat = .infinity
+    public static let auto: CGFloat = .infinity
 }
 
 public class ContainerView: UIView {
-    convenience init(
+    public convenience init(
```
fortmarek, line 35: hmm, this class is not documented at all, although it is public ... is it planned to document this later ... ?

VojtaStavik: @fortmarek yes, still a bit WIP

Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
VojtaStavik: @fortmarek  There's no other reason than having all layout code in one place. I think you could say the same thing about all other elements :) If you override the variable and provide some custom values there, it will be overridden by customization methods...

SlackClone/SlackChatChannelListItemView.swift
DominikBucher12: Yes, replaced. :)

nuno-vieira (commented):
I can see we did not change the tests, but it would be nice to manipulate the Container View in the tests and check how customisable it is. Since this is the point of this PR, is to make it easier to customise the view. So we should be doing that on the tests.

nuno-vieira (approved):
LGTM!

PR #984, opened 2021-04-13
[CIS-792] Make models thread safe

VojtaStavik:
There was a regression introduced by #906 which made our models thread-unsafe. This PR fixes the problem. Instead of the `@Cached` wrapper, we're using `@CoreDataLazy` which evaluates the values on the correct `NSManagedObjectContext`.

### DIFF
--- a/Sources/StreamChat/Utils/CoreDataLazy.swift
+++ b/Sources/StreamChat/Utils/CoreDataLazy.swift
@@ -1,0 +1,65 @@
+//
+// Copyright © 2021 Stream.io Inc. All rights reserved.
+//
+
+import CoreData
+import Foundation
+
+/// Makes it possible to lazily evaluate properties of `NSManagedObject`s.
+///
+/// Works like a "classic" lazy wrapper. The only different is that the `computeValue` closure is evaluated on the queue
+/// of the provided `NSManagedObjectContext`.
+///
+@propertyWrapper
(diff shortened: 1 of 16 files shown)


Sources/StreamChat/Utils/CoreDataLazy.swift
```diff
@@ -0,0 +1,65 @@
...
+            let perform = {
+                log.assert(self.computeValue != nil, "You must set the `computeValue` closure before accessing the cached value.")
+                returnValue = self.computeValue()
+            }
+            
+            if let context = context {
+                context.performAndWait { perform() }
+            } else {
+                // This is really a fallback for cases like tests, mocks, and edge cases. Even though `NSManagedObject.managedObjectContext`
+                // is optional, we probably won't reach this point during a normal production run.
```
olejnjak, line 41: If we consider this to be unreachable in production, it might be better to add some assertion?

olejnjak (approved):

Sources/StreamChat/Utils/CoreDataLazy.swift
VojtaStavik: @olejnjak I didn't say it's "unreachable" I said "we probably won't reach" 😅 The comment is actually outdated. MessageDTO uses this for the cases where the results are empty and we don't need to use the context queue. I'll update it.

PR #985, opened 2021-04-13
Message content view

evsaev:
**This PR**:
- extracts content from `MessageCell` to `MessageContentView`
- makes use of `MessageContentView` throughout message rethink


Sources/StreamChatUI/ChatMessageList/RETHINK/MessageCell.swift
```diff
@@ -5,331 +5,61 @@
...
-                ),
-                bubbleView.topAnchor.pin(equalTo: reactionsView.centerYAnchor)
-            ]
-        }
-        
-        NSLayoutConstraint.activate(constraintsToActivate)
+    var content: _ChatMessage<ExtraData>? {
+        get { messageContentView.content }
+        set { messageContentView.content = newValue }
     }
```
VojtaStavik, line 22: Let's not do this. These values should be accessed directly on the view

Sources/StreamChatUI/ChatMessageList/RETHINK/MessageCell.swift
```diff
@@ -5,331 +5,61 @@
...
-                    self?.delegate?.didTapOnAttachmentAction($0, $1)
-                }
-            )
+    func setUpLayoutIfNeeded(options: ChatMessageLayoutOptions) {
+        if messageContentView.layoutOptions == nil {
+            let alignment = options.contains(.flipped) ?
+                messageContentView.trailingAnchor.pin(equalTo: contentView.trailingAnchor) :
+                messageContentView.leadingAnchor.pin(equalTo: contentView.leadingAnchor)
+            alignment.isActive = true
         }
```
evsaev, line 35: To me it feels the container should be responsible for content alignment, not the content view. That's why we set the leading/trailing constraint in a cell. However, the `if options == nil` check doesn't look nice but didn't find a better way to set the constraint just once

VojtaStavik, line 35: Can we keep all alignment stuff in one place -> inside the content view?

Sources/StreamChatUI/ChatMessageList/RETHINK/MessageCell.swift
```diff
@@ -5,331 +5,61 @@
...
+            messageContentView
+                .topAnchor
+                .pin(equalTo: contentView.topAnchor),
+            messageContentView
+                .bottomAnchor
+                .pin(equalTo: contentView.bottomAnchor),
+            messageContentView
+                .widthAnchor
+                .pin(lessThanOrEqualTo: contentView.widthAnchor, multiplier: 0.75)
+        ])
```
VojtaStavik, line 53: The content should decide about its ideal width.

VojtaStavik (approved):

Sources/StreamChatUI/ChatMessageList/RETHINK/MessageCell.swift
evsaev: done

Sources/StreamChatUI/ChatMessageList/RETHINK/MessageCell.swift
evsaev: fixed 👍

PR #986, opened 2021-04-13
[CIS-793] Implement typing indicator to messageList

DominikBucher12:
# What this PR do:
- Implements new UI Component for indicating that other users are typing inside Channel
# How to test it
- Run the app on 2 different devices and choose 2 different accounts, then test when you are writing from one account that another account is receiving the events.

https://user-images.githubusercontent.com/17381941/114886980-518c2c80-9e08-11eb-8cf7-02459a8ac215.mov

https://user-images.githubusercontent.com/17381941/114887014-581aa400-9e08-11eb-909f-35220c2f49fb.mov


Sources/StreamChatUI/ChatMessageList/ChatTypingIndicatorView.swift
```diff
@@ -0,0 +1,150 @@
...
+
+// TODO: Decide which IndicatorView to Use:
+// A) AChatTypingIndicatorView
+// B) ChatTypingIndicatorView
+
+/// A `UIView` subclass with 3 dots which can be animated with fading out effect.
+public typealias AChatTypingIndicatorView = _ChatTypingIndicatorView<NoExtraData>
+
+/// A `UIView` subclass with 3 dots which can be animated with fading out effect.
+public typealias ChatTypingIndicatorView = _ChatTypingIndicatorView<NoExtraData>
```
DominikBucher12, line 16: IMHO the `ChatTypingIndicatorView` is much better looking + memory friendly. It doesn't use any complicated API and the layer is easily replicated. Also it animates properly, the `AChatTypingIndicatorView` sometimes doesn't finish the animationGroup and behaves a lot strange..

I tried to implement this via `UIViewPropertyAnimator` but there is no simple way of repeating the animations, hence read this: https://developer.apple.com/forums/thread/122209

DominikBucher12: Please let me know which of the 2 indicators you prefer ^^

Sources/StreamChatUI/ChatMessageList/ChatChannelVC.swift
```diff
@@ -99,4 +99,27 @@ extension _ChatChannelVC: _ChatChannelControllerDelegate {
...
+
+    public func channelController(
+        _ channelController: _ChatChannelController<ExtraData>,
+        didChangeTypingMembers typingMembers: Set<_ChatChannelMember<ExtraData.User>>
+    ) {
+        guard let currentMember = channelController.channel?.membership else {
+            typingIndicatorView.setAnimatedly(hidden: true)
+            log.error("Current member not found in ChatChannelVC.")
+            return
+        }
```
VojtaStavik, line 111: I wouldn't check for a membership here. Users don't have to be members to create typing events. Better to check for the current logged in user directly.

Sources/StreamChatUI/ChatMessageList/ChatChannelVC.swift
```diff
@@ -99,4 +99,27 @@ extension _ChatChannelVC: _ChatChannelControllerDelegate {
...
         messageList.updateMessages(with: changes)
     }
+
+    public func channelController(
+        _ channelController: _ChatChannelController<ExtraData>,
+        didChangeTypingMembers typingMembers: Set<_ChatChannelMember<ExtraData.User>>
+    ) {
+        guard let currentMember = channelController.channel?.membership else {
+            typingIndicatorView.setAnimatedly(hidden: true)
+            log.error("Current member not found in ChatChannelVC.")
```
VojtaStavik, line 109: This is not an error at all but a perfectly valid case ;-)

Sources/StreamChatUI/ChatMessageList/ChatChannelVC.swift
```diff
@@ -99,4 +99,27 @@ extension _ChatChannelVC: _ChatChannelControllerDelegate {
...
+            return
+        }
+
+        let typingMembersWithoutCurrentUser = typingMembers.subtracting([currentMember])
+
+        if typingMembersWithoutCurrentUser.isEmpty {
+            typingIndicatorView.setAnimatedly(hidden: true)
+        } else if typingMembersWithoutCurrentUser.count == 1 {
+            guard let member = typingMembers.first, let name = member.name else { return }
```
VojtaStavik, line 118: Maybe we can just fallback to `Someone is typing`?

Sources/StreamChatUI/ChatMessageList/ChatChannelVC.swift
```diff
@@ -99,4 +99,27 @@ extension _ChatChannelVC: _ChatChannelControllerDelegate {
...
+            typingIndicatorView.content = "\(name) " + L10n.MessageList.TypingIndicator.typingSingular
+        } else {
+            guard let member = typingMembers.first, let name = member.name else { return }
```
VojtaStavik, line 121: Same here

Sources/StreamChatUI/ChatMessageList/ChatMessageListIndicatorOverlayView.swift
```diff
@@ -0,0 +1,61 @@
...
+//
+
+import StreamChat
+import UIKit
+
+/// An `UIView` subclass indicating that user or multiple users are currently typing.
+public typealias ChatMessageTypingIndicatorOverlayView = _ChatMessageListIndicatorOverlayView<NoExtraData>
+
+/// An `UIView` subclass indicating that user or multiple users are currently typing.
+open class _ChatMessageListIndicatorOverlayView<ExtraData: ExtraDataTypes>: _View, UIConfigProvider {
```
VojtaStavik, line 12: This is not _just_ overlay view. IMO this is _the_ typing indicator view.

Sources/StreamChatUI/ChatMessageList/ChatMessageListIndicatorOverlayView.swift
```diff
@@ -0,0 +1,61 @@
...
+            updateContentIfNeeded()
+        }
+    }
+
+    /// The animated view with three dots indicating that someone is typing.
+    open private(set) lazy var typingIndicatorView: _ChatTypingIndicatorView<ExtraData> = uiConfig
+        .messageList
+        .typingIndicatorView
+        .init()
+        .withoutAutoresizingMaskConstraints
```
VojtaStavik, line 24: `typingIndicatorView` -> `typingAnimationView`?

Sources/StreamChatUI/ChatMessageList/ChatTypingIndicatorView.swift
```diff
@@ -0,0 +1,150 @@
...
+        layer.backgroundColor = UIColor.red.cgColor
+        layer.cornerRadius = dotSize.height / 2
+        return layer
+    }()
+
+    open private(set) lazy var replicatorLayer = CAReplicatorLayer()
+
+    override open func setUpLayout() {
+        super.setUpLayout()
+        embed(innerView, insets: directionalLayoutMargins)
```
VojtaStavik, line 42: We need to remove this API 😱 It's wrong. You want to do `innerView.pin(to: layoutMarginsGuide)`

Sources/StreamChatUI/ChatMessageList/ChatMessageListIndicatorOverlayView.swift
```diff
@@ -0,0 +1,61 @@
...
+        componentStackView.addArrangedSubview(informationLabel)
+        componentStackView.distribution = .fill
+        componentStackView.alignment = .center
+
+        typingIndicatorView.heightAnchor.pin(equalToConstant: 5).isActive = true
+        typingIndicatorView.centerYAnchor.pin(equalTo: centerYAnchor).isActive = true
+    }
+
+    override public func defaultAppearance() {
+        layer.opacity = 0.8
```
VojtaStavik, line 52: This opacity affects also the label and the animation view. Is that intentional?

Sources/StreamChatUI/ChatMessageList/ChatTypingIndicatorView.swift
```diff
@@ -0,0 +1,150 @@
...
+public typealias ChatTypingIndicatorView = _ChatTypingIndicatorView<NoExtraData>
+
+/// A `UIView` subclass with 3 dots which can be animated with fading out effect.
+open class _ChatTypingIndicatorView<ExtraData: ExtraDataTypes>: _View, UIConfigProvider {
+    open var dotSize: CGSize = .init(width: 5, height: 5)
+
+    private(set) lazy var innerView: UIView = UIView().withoutAutoresizingMaskConstraints
+
+    private var viewWidthConstant: CGFloat {
+        dotLayer.frame.size.width * 1.5 * 3 + directionalLayoutMargins.leading * 2
```
VojtaStavik, line 25: would be nice to extract these values into some constants

Sources/StreamChatUI/UIConfig+MessageList.swift
```diff
@@ -12,6 +12,13 @@ public extension _UIConfig {
         public var defaultMessageCell: _СhatMessageCollectionViewCell<ExtraData>.Type =
             _СhatMessageCollectionViewCell<ExtraData>.self
 
+        /// A `UIView` subclass with animated 3 dots for indicating that user is typing.
+        public var typingIndicatorView: _ChatTypingIndicatorView<ExtraData>.Type = _ChatTypingIndicatorView<ExtraData>.self
+
+        /// A `UIView` subclass which serves as container for `typingIndicator` and `UILabel` describing who is currently typing
+        public var typingIndicatorOverlayView: _ChatMessageListIndicatorOverlayView<ExtraData>.Type =
+            _ChatMessageListIndicatorOverlayView<ExtraData>.self
+
```
VojtaStavik, line 21: Please check out this document and make sure the changes to UIConfig follow it https://www.notion.so/UI-Config-Rules-Re-Think-5f58357b1f324879927aaeea1aa980a0

Sources/StreamChatUI/ChatMessageList/ChatTypingIndicatorView.swift
```diff
@@ -0,0 +1,150 @@
...
+// B) ChatTypingIndicatorView
+
+/// A `UIView` subclass with 3 dots which can be animated with fading out effect.
+public typealias AChatTypingIndicatorView = _ChatTypingIndicatorView<NoExtraData>
+
+/// A `UIView` subclass with 3 dots which can be animated with fading out effect.
+public typealias ChatTypingIndicatorView = _ChatTypingIndicatorView<NoExtraData>
+
+/// A `UIView` subclass with 3 dots which can be animated with fading out effect.
+open class _ChatTypingIndicatorView<ExtraData: ExtraDataTypes>: _View, UIConfigProvider {
```
VojtaStavik, line 19: This one seems simpler to me 👍

Sources/StreamChatUI/ChatMessageList/ChatTypingIndicatorView.swift
```diff
@@ -0,0 +1,150 @@
...
+    open var dotSize: CGSize = .init(width: 5, height: 5)
+
+    private(set) lazy var innerView: UIView = UIView().withoutAutoresizingMaskConstraints
```
VojtaStavik, line 22: why do we need this?

VojtaStavik (changes requested):
Almost there ;-)

Sources/StreamChatUI/Resources/en.lproj/Localizable.strings
```diff
@@ -22,6 +22,9 @@
 "message.actions.user-mute" = "Mute User";
 "message.actions.resend" = "Resend";
 
+"messageList.typingIndicator.typing-singular" = "is typing";
+"messageList.typingIndicator.typing-plural" = "more are typing";
```
DominikBucher12, line 26: Add localization for `Someone is typing` as a fallback.

Sources/StreamChatUI/ChatMessageList/ChatMessageListIndicatorOverlayView.swift
DominikBucher12: Nope, I actually fixed this with creating new color ( discovered that we are missing color `overlayLight` which is contrast to `overlayDark`), this will be fixed in upcoming commits.

Sources/StreamChatUI/ChatMessageList/ChatChannelVC.swift
```diff
@@ -99,4 +99,27 @@ extension _ChatChannelVC: _ChatChannelControllerDelegate {
...
+            typingIndicatorView.content = "\(name) and \(typingMembers.count - 1) " + L10n.MessageList.TypingIndicator.typingPlural
```
skorepak, line 122: I would suggest passing the parameters into the localised string. 

We should enable the clients to change the positions of the parameters in the localised string. So they can change from: "Dominik and 3 others are typing" to "Typing: Dominik and 3 others" for example.

Sources/StreamChatUI/ChatMessageList/ChatTypingIndicatorView.swift
DominikBucher12: I could achieve this other ways mentioned:
 1.  Adding some borders to the layer in the view. 
 2. Using frame based layout for the layer inside the view.  

However IMHO is the most easier way to add `insets` to the layer and add the replicator layer inside with no fear it will touch the leading and trailing. Because on layer, you cannot add the insets.

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerInputTextView.swift
```diff
@@ -90,6 +90,10 @@ open class _ChatMessageComposerInputTextView<ExtraData: ExtraDataTypes>: UITextV
     }
     
     open func updateContent() {}
+
+    open func setUpHeightConstraint() {
```
skorepak, line 94: Naming suggestion: `update` over `setUp`

Sources/StreamChatUI/Resources/en.lproj/Localizable.strings
skorepak, line 26: Using singular for one and plural for 2 and more is sufficient in English, but not for other languages. A proper pluralisation would be nice here. I'm not sure how L10n works, but the standards iOS way is to use stringsdict file.

Sources/StreamChatUI/ChatMessageList/ChatChannelVC.swift
DominikBucher12: Changed the logic a bit, now searching for any member that has name, otherwise `Someone is typing`

Sources/StreamChatUI/ChatMessageList/ChatChannelVC.swift
DominikBucher12: Thanks, fixed)

Sources/StreamChatUI/ChatMessageList/ChatChannelVC.swift
DominikBucher12: Extracted the code above, however question is if `Someone is typing` in DM Channel isn't strange 😅

Sources/StreamChatUI/ChatMessageList/ChatChannelVC.swift
DominikBucher12: Updated, thanks for the suggestion... I tried to implement this via Plural localized string but no luck, more info in internal discussion.

skorepak (approved):
LGTM from the localisation point of view 👍🏻 
The rest of my knowledge is still limited, please wait for other approvals

Sources/StreamChatUI/ChatMessageList/ChatMessageListIndicatorOverlayView.swift
```diff
@@ -0,0 +1,59 @@
...
+//
+
+import StreamChat
+import UIKit
+
+/// An `UIView` subclass indicating that user or multiple users are currently typing.
+public typealias ChatMessageListTypingIndicatorView = _ChatMessageListTypingIndicatorView<NoExtraData>
+
+/// An `UIView` subclass indicating that user or multiple users are currently typing.
+open class _ChatMessageListTypingIndicatorView<ExtraData: ExtraDataTypes>: _View, UIConfigProvider {
```
VojtaStavik, line 12: No more crazy long names please 😅  `TypingIndicatorView` is enough for our needs

Sources/StreamChatUI/ChatMessageList/ChatTypingIndicatorView.swift
VojtaStavik: Can we name this `TypingAnimationView`?

Sources/StreamChatUI/ChatMessageList/ChatTypingIndicatorView.swift
```diff
@@ -0,0 +1,69 @@
...
+
+    private let numberOfDots: Int = 3
+    private let dotSpacing: CGFloat = 1.5
```
VojtaStavik, line 18: `open`?

Sources/StreamChatUI/ChatMessageList/ChatTypingIndicatorView.swift
```diff
@@ -0,0 +1,69 @@
...
+    override public func defaultAppearance() {
+        backgroundColor = .clear
+        dotLayer.backgroundColor = uiConfig.colorPalette.text.cgColor
+    }
+
+    open func startAnimating() {
+        let opacityAnimation = CABasicAnimation(keyPath: "opacity")
+        opacityAnimation.fromValue = 0.9
+        opacityAnimation.toValue = 0.3
+        opacityAnimation.duration = 1.0
```
VojtaStavik, line 55: Why not extract these values, too?

Sources/StreamChatUI/UIConfig+MessageList.swift
```diff
@@ -12,6 +12,10 @@ public extension _UIConfig {
         public var defaultMessageCell: _СhatMessageCollectionViewCell<ExtraData>.Type =
             _СhatMessageCollectionViewCell<ExtraData>.self
 
+        /// A `UIView` subclass which serves as container for `typingIndicator` and `UILabel` describing who is currently typing
+        public var typingIndicatorOverlayView: _ChatMessageListTypingIndicatorView<ExtraData>.Type =
+            _ChatMessageListTypingIndicatorView<ExtraData>.self
+
```
VojtaStavik, line 18: This should be in the root https://www.notion.so/UI-Config-Rules-Re-Think-5f58357b1f324879927aaeea1aa980a0 + the name of the variable is off

Sources/StreamChatUI/ChatMessageList/ChatMessageListIndicatorOverlayView.swift
```diff
@@ -0,0 +1,59 @@
...
+    var content: String? {
```
VojtaStavik, line 13: missing documentation

Sources/StreamChatUI/ChatMessageList/ChatMessageListIndicatorOverlayView.swift
```diff
@@ -0,0 +1,59 @@
...
+        .init()
+        .withoutAutoresizingMaskConstraints
+
+    /// Label describing who is currently typing
+    /// `User is typing`
+    /// `User and 3 more are typing`
+    open private(set) lazy var informationLabel: UILabel = UILabel().withoutAutoresizingMaskConstraints
+
+    /// StackView holding `typingIndicatorView` and `informationLabel`
+    open private(set) lazy var componentStackView: UIStackView = UIStackView().withoutAutoresizingMaskConstraints
```
VojtaStavik, line 31: Let's use our container view instead

VojtaStavik (approved):
Some final touches and we're good to :ship:

Sources/StreamChatUI/ChatMessageList/ChatTypingIndicatorView.swift
DominikBucher12: Okay I deleted the innerView because it was quite complex (and ugly) to add the margins to it and make it work properly. Now the view has no margins and they need to be setup manually from outside (which IMHO makes more sense)

PR #987, opened 2021-04-14
Composer components rename

b-onc:
Also includes creation of folders for the components.

Did not touch Attachment stuff (ImageAttachments/DocumentAttachment) since they'll be reworked completely.


Sources/StreamChatUI/UIConfig+MessageComposer.swift
```diff
@@ -10,8 +10,8 @@ public extension _UIConfig {
             _ChatMessageComposerVC<ExtraData>.self
         public var messageComposerView: _ChatMessageComposerView<ExtraData>.Type =
             _ChatMessageComposerView<ExtraData>.self
-        public var messageInputView: _ChatMessageComposerInputContainerView<ExtraData>
-            .Type = _ChatMessageComposerInputContainerView<ExtraData>.self
+        public var messageInputView: _ChatInputContainerView<ExtraData>
+            .Type = _ChatInputContainerView<ExtraData>.self
```
nuno-vieira, line 14: I think we need to move this to the root of the UIConfig. Only `messageComposerView` should live here

Sources/StreamChatUI/UIConfig+MessageComposer.swift
```diff
@@ -45,8 +45,8 @@ public extension _UIConfig {
 
         public var checkmarkControl: _ChatMessageComposerCheckmarkControl<ExtraData>.Type =
             _ChatMessageComposerCheckmarkControl<ExtraData>.self
-        public var slashCommandView: _ChatMessageInputSlashCommandView<ExtraData>
-            .Type = _ChatMessageInputSlashCommandView<ExtraData>.self
+        public var slashCommandView: _ChatInputSlashCommandView<ExtraData>
+            .Type = _ChatInputSlashCommandView<ExtraData>.self
```
nuno-vieira, line 49: This is one should also go to the root of the UIConfig

Sources/StreamChatUI/CommonViews/ChatSuggestions/ChatSuggestionsCollectionView.swift
```diff
@@ -6,10 +6,10 @@ import StreamChat
 import UIKit
 
 /// The collection view of the suggestions view controller.
-public typealias ChatMessageComposerSuggestionsCollectionView = _ChatMessageComposerSuggestionsCollectionView<NoExtraData>
+public typealias ChatSuggestionsCollectionView = _ChatSuggestionsCollectionView<NoExtraData>
 
 /// The collection view of the suggestions view controller.
-open class _ChatMessageComposerSuggestionsCollectionView<ExtraData: ExtraDataTypes>: UICollectionView,
+open class _ChatSuggestionsCollectionView<ExtraData: ExtraDataTypes>: UICollectionView,
```
nuno-vieira, line 12: I think this name is a bit generic, and hard to understand what are these suggestions about. WDYT of `ChatInputSuggestions...` prefix ?

Sources/StreamChatUI/CommonViews/ChatSuggestions/ChatSuggestionsCommandCollectionViewCell.swift
```diff
@@ -6,10 +6,10 @@ import StreamChat
 import UIKit
 
 /// A view that displays the command name, image and arguments.
-public typealias ChatMessageComposerCommandCellView = _ChatMessageComposerCommandCellView<NoExtraData>
+public typealias ChatSuggestionsCommandCellView = _ChatSuggestionsCommandCellView<NoExtraData>
 
 /// A view that displays the command name, image and arguments.
-open class _ChatMessageComposerCommandCellView<ExtraData: ExtraDataTypes>: _View, UIConfigProvider {
+open class _ChatSuggestionsCommandCellView<ExtraData: ExtraDataTypes>: _View, UIConfigProvider {
```
nuno-vieira, line 12: I think we should remove this pattern of naming `CellView` suffix. It should just be `ChatSuggestionsCommandView` since you could simply use this view in another place other than inside a cell

Sources/StreamChatUI/UIConfig+MessageComposer.swift
```diff
@@ -43,8 +43,8 @@ public extension _UIConfig {
         public var composerButton: UIButton.Type = UIButton.self
         public var textView: _ChatMessageComposerInputTextView<ExtraData>.Type = _ChatMessageComposerInputTextView<ExtraData>.self
 
-        public var checkmarkControl: _ChatMessageComposerCheckmarkControl<ExtraData>.Type =
-            _ChatMessageComposerCheckmarkControl<ExtraData>.self
+        public var checkmarkControl: _ChatCheckmarkControl<ExtraData>.Type =
+            _ChatCheckmarkControl<ExtraData>.self
```
nuno-vieira, line 47: This can also go to the root of the config. But, this component is probably not even needed, since it is so simple, but on the refactor of composer view that will be decided, so no need to change anything here.

nuno-vieira, line 47: This can also go to the root of the config. But, this component is probably not even needed, since it is so simple, but on the refactor of composer view that will be decided, so no need to change anything here.

Sources/StreamChatUI/UIConfig+MessageComposer.swift
VojtaStavik: @nuno-vieira Ideally nothing lives here and the `UIConfig` is just one level deep.

Sources/StreamChatUI/UIConfig+MessageComposer.swift
```diff
@@ -41,7 +41,7 @@ public extension _UIConfig {
 
         /// Buttons used in the left side of the composer. Corresponds to attachment and actions buttons.
         public var composerButton: UIButton.Type = UIButton.self
-        public var textView: _ChatMessageComposerInputTextView<ExtraData>.Type = _ChatMessageComposerInputTextView<ExtraData>.self
+        public var textView: _ChatInputTextView<ExtraData>.Type = _ChatInputTextView<ExtraData>.self
```
nuno-vieira, line 44: Same here, this should go to the root of UIConfig, and probably rename to `inputTextView` ?

Sources/StreamChatUI/CommonViews/ChatSuggestions/ChatSuggestionsCollectionView.swift
VojtaStavik: I don't think we need/should have this type tbh. It's just a collection view with some layout settings. The setting should happen in the VC, aynways.

Sources/StreamChatUI/CommonViews/ChatSuggestions/ChatSuggestionsCommandCollectionViewCell.swift
VojtaStavik: `ChatSuggestionsCommandView` -> isn't `ChatCommandSuggestionsView` more correct?

Sources/StreamChatUI/CommonViews/ChatSuggestions/ChatSuggestionsCollectionView.swift
nuno-vieira: Yes, that is true, this one not necessary at all probably. My comment is only for the renaming of these components prefix, to be `InputSuggestions` instead of only `Suggestions`.

Sources/StreamChatUI/CommonViews/ChatSuggestions/ChatSuggestionsCommandCollectionViewCell.swift
nuno-vieira: Based on the comment above I would suggest `ChatInputSuggestionsCommandView` or `ChatCommandInputSuggestionsView`. Depends on the pattern that we chose.

Option 1:
`ChatInputSuggestionsViewController`
`ChatInputSuggestionsCommandView`
`ChatInputSuggestionsMentionView`

Option 2:
`ChatInputSuggestionsViewController`
`ChatCommandInputSuggestionsCommandView`
`ChatCommandInputSuggestionsMentionView`

I prefer Option 1, since it looks cleaner and easier to find because the prefix is always the same. WDYT?

Sources/StreamChatUI/UIConfig+MessageComposer.swift
nuno-vieira: Yes exactly, we should aim to remove all `UIConfig+X` sections. The `UIConfig` should just act like a design system, where each component can be used and composed in any part of the app 👍

b-onc: I aim to refactor UIConfig in a follow up PR, wdyt?

Sources/StreamChatUI/CommonViews/ChatSuggestions/ChatSuggestionsCollectionView.swift
b-onc: valid feedback, i think it's better suited for the UIConfig follow-up PR

Sources/StreamChatUI/CommonViews/ChatSuggestions/ChatSuggestionsCommandCollectionViewCell.swift
b-onc: These views can be used in non-input contexts, so I don't find `Input` in the name correct. I liked `ChatCommandSuggestionsView` 2️⃣ 💵

nuno-vieira (approved):
LGTM!

PR #988, opened 2021-04-15
Message reactions bubble

evsaev:
**This PR**:
- introduces new bubble for reactions view
- masks `MessageCell` to have space around reactions bubble
- introduces some `CGRect/CGPoint` helpers
- introduces `UIImage` helper to make a mask

### DIFF
--- a/Sources/StreamChatUI/ChatMessageList/RETHINK/MessageContentView.swift
+++ b/Sources/StreamChatUI/ChatMessageList/RETHINK/MessageContentView.swift
@@ -43,8 +43,10 @@ class MessageContentView<ExtraData: ExtraDataTypes>: _View, UIConfigProvider {
     var filePreviewView: _ChatMessageFileAttachmentListView<ExtraData>?
     var giphyView: _ChatMessageGiphyView<ExtraData>?
     var actionsView: _ChatMessageInteractiveAttachmentView<ExtraData>?
-    var reactionsView: _ReactionsCompactView<ExtraData>?
     var errorIndicatorView: _ChatMessageErrorIndicator<ExtraData>?
+
+    var reactionsView: _ReactionsCompactView<ExtraData>?
+    var reactionsBubbleView: ReactionsBubbleView<ExtraData>?
 
     var threadReplyCountButton: UIButton?
     var threadAvatarView: ChatAvatarView?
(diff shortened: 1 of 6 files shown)


Sources/StreamChatUI/ChatMessageList/RETHINK/MessageContentView.swift
```diff
@@ -43,15 +43,51 @@ class MessageContentView<ExtraData: ExtraDataTypes>: _View, UIConfigProvider {
...
+        let conainerMaskingPath = reactionsBubble.maskingPath
+        conainerMaskingPath.apply(moveToMainContainer)
+
+        let maskImage = UIGraphicsImageRenderer(size: mainContainer.bounds.size)
+            .image {
+                let layer = CAShapeLayer()
+                layer.path = conainerMaskingPath.cgPath
+                layer.render(in: $0.cgContext)
+            }
+            .asAlphaMask()
```
evsaev, line 87: This part is computationally heavy and will be replaced later when we figure out how to build a mask with `evenOdd` + `UIBezierPath` with the intersections (reactions bubble's tail has 2 overlapping circles and we have a hole in mask where circles intersect)

Sources/StreamChatUI/ChatMessageList/RETHINK/MessageContentView.swift
```diff
@@ -240,17 +276,21 @@ class MessageContentView<ExtraData: ExtraDataTypes>: _View, UIConfigProvider {
...
         if options.contains(.reactions) {
+            let reactionsBubbleView = createReactionsBubbleView()
+            addSubview(reactionsBubbleView)
+
             let reactionsView = createReactionsView()
-            reactionsView.setContentCompressionResistancePriority(.defaultHigh, for: .horizontal)
-            reactionsView.setContentCompressionResistancePriority(.defaultHigh, for: .vertical)
-            addSubview(reactionsView)
+            reactionsBubbleView.addSubview(reactionsView)
+            reactionsView.pin(to: reactionsBubbleView.layoutMarginsGuide)
```
evsaev, line 284: Previously this setup was inside the bubble but I tired to flatten this a bit. To me bubble view by its own looks weird since it's not clear how it should be used. However, the flattened structure is easier to follow then a nested `bubble -> content` one

Sources/StreamChatUI/ChatMessageList/Reactions/ChatMessageReactionsView.swift
```diff
@@ -7,6 +7,122 @@ import UIKit
...
+            dy: 14
+        )
+    }
+
+    var smallTailCirleCenter: CGPoint {
+        bigTailCirleCenter.offsetBy(
+            dx: tailDirection == .toTrailing ? 4 : -4,
+            dy: 6
+        )
+    }
```
evsaev, line 84: In `ReactionsBubbleView` there is a plenty of magic numbers (circle radii and offsets). 

Was trying to get rid of them but didn't succeed since **we have to have shape centres fixed so we can make scale them for the mask and they stay in the same place**

VojtaStavik (approved):
And you thought you wouldn't use calculus in your daily work! 🎉

PR #989, opened 2021-04-15
Regenerate snapshots for ChatMessageComposerImageAttachmentsView

fortmarek:
# Submit a pull request

`main` is currently broken because `ChatMessageComposerImageAttachmentsView` snapshot tests do not match the reference. AFAICT the only change was in the size of snapshots, so I just regenerated the snapshots as they seem to be the same. Let me know if you think otherwise.


fortmarek:
Seems like main has been fixed, closing

PR #990, opened 2021-04-15
Message cell adjustments

evsaev:


Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageLayoutOptions.swift
```diff
@@ -15,6 +15,7 @@ public struct ChatMessageLayoutOptions: OptionSet, Hashable {
 public extension ChatMessageLayoutOptions {
     /// Typically the messages sent by the current user have flipped content
     static let flipped = Self(rawValue: 1 << 0)
+    static let maxWidth = Self(rawValue: 1 << 15)
```
VojtaStavik, line 18: Very good idea 👍

Sources/StreamChatUI/ChatMessageList/RETHINK/MessageContentView.swift
```diff
@@ -293,6 +292,14 @@ class MessageContentView<ExtraData: ExtraDataTypes>: _View, UIConfigProvider {
...
             ]
         }
+
+        constraintsToActivate += fullWidthViews.map { view in
+            view.widthAnchor
+                .pin(equalTo: mainContainer.widthAnchor)
+                // should work with just .streamLow but it doesn't
+                // work for stream preview
+                .with(priority: .streamAlmostRequire)
+        }
```
evsaev, line 302: That's probably the most interesting part of the PR but not sure `maxWidth` option is valuable for the new attachments approach

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageMetadataView.swift
```diff
@@ -58,6 +67,9 @@ open class _ChatMessageMetadataView<ExtraData: ExtraDataTypes>: _View, UIConfigP
             timestampLabel.text = nil
         }
         currentUserVisabilityIndicator.isVisible = content?.onlyVisibleForCurrentUser ?? false
+
+        authorLabel.isHidden = content?.isSentByCurrentUser ?? true
```
VojtaStavik, line 71: Nope, this should be decided on the `ChatMessageLayoutOptions` level. For example, in 1:1 channel, the author name is never visible, only when there are more than the current user + 1 person there, we show the names for everyone but the current user.

Sources/StreamChatUI/ChatMessageList/RETHINK/MessageContentView.swift
```diff
@@ -98,17 +100,26 @@ class MessageContentView<ExtraData: ExtraDataTypes>: _View, UIConfigProvider {
         layoutOptions = options
         
         var constraintsToActivate: [NSLayoutConstraint] = []
+        var fullWidthViews: [UIView] = []
         
         // Main container
         mainContainer.alignment = .axisTrailing
```
VojtaStavik, line 106: I actually think we can achieve the full-width functionality just by changing this to `.fill`. It probably won't work yet but it should eventually :)

VojtaStavik (approved):

Sources/StreamChatUI/ChatMessageList/RETHINK/MessageContentView.swift
```diff
@@ -412,7 +419,7 @@ private extension MessageContentView {
             textView?.isScrollEnabled = false
             textView?.backgroundColor = .clear
             textView?.adjustsFontForContentSizeCategory = true
-            textView?.textContainerInset = .zero
+            textView?.textContainerInset = .init(top: 0, left: 8, bottom: 0, right: 8)
```
evsaev, line 422: @VojtaStavik  I was thinking about extending `ContainerView` so it's possible to provide custom layout margins for each arranged subview but decided to discuss `ContainerView` with you first and keep it simple in the PR

Sources/StreamChatUI/ChatMessageList/RETHINK/MessageContentView.swift
evsaev: The `fill` requires us to rework existing components so they can stretch but still look right

Sources/StreamChatUI/ChatMessageList/RETHINK/MessageContentView.swift
VojtaStavik: @evsaev sure, let's talk about the details during auditing 👍

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageMetadataView.swift
evsaev: Fixed, please have a look.

The fact we don't have access to `cid` from the message brings the complexity 🤔 Why not to change this:
```swift
extension _ChatChannel {
    var isDirectMessageChannel: Bool { cid.id.hasPrefix("!members") }
}
```
with:
```swift
struct _ChatMessage {
    let cid: ChannelId
}
extension ChannelId {
    var isDirectMessageChannel: Bool { id.hasPrefix("!members") }
}
```

VojtaStavik: @evsaev gotcha. Let's leave it as it, it's not _that_ bad :) Another thing to think about properly during the audit.

PR #991, opened 2021-04-15
Switch DemoApp users to SampleApp users

b-onc:
Where we use Star Wars themed people


VojtaStavik:
> Where we use Star Wars themed ~people~ beings

VojtaStavik (approved):
Thank you!

PR #993, opened 2021-04-16
MessageListVC documentation

LukasHromadnik:
# Description

Audit of the new `MessageListVC`. Currently only documentation and method order. Snapshots will follow.


VojtaStavik (commented):
👍 Don't be afraid to use `open` for everythig instead of `public`, unless there's a very strong reason not to have it overridable.

PR #995, opened 2021-04-16
Refactor composer view with new container view

nuno-vieira:
# In this PR
- Refactors the composer view to use the new container stack view.

**Note:** Right now the composer view controller needs to know about the implementation details of ComposerView in order to use it. This will change in the next PR.

### DIFF
--- a/Sources/StreamChatUI/CommonViews/ContainerStackView.swift
+++ b/Sources/StreamChatUI/CommonViews/ContainerStackView.swift
@@ -278,7 +278,8 @@ public class ContainerStackView: UIView {
     }
 
     public func hideSubview(_ subview: UIView, animated: Bool = true) {
-        assert(subviews.contains(subview))
+        guard subviews.contains(subview) else { return }
+        guard subview.alpha != 0 else { return }
 
         updateConstraintsIfNeeded()
 
@@ -289,7 +290,10 @@ public class ContainerStackView: UIView {
 
         Animate(isAnimated: animated) {
             subview.alpha = 0
-            self.hideConstraintsByView[subview]?.width.isActive = true
+
+            if self.axis == .horizontal {
+                self.hideConstraintsByView[subview]?.width.isActive = true
+            }
             self.hideConstraintsByView[subview]?.height.isActive = true
 
             self.spacingConstraintsByView[subview]?.setTemporaryConstant(0)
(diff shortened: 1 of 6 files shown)


Sources/StreamChatUI/CommonViews/ContainerStackView.swift
```diff
@@ -289,7 +290,10 @@ public class ContainerStackView: UIView {
 
         Animate(isAnimated: animated) {
             subview.alpha = 0
-            self.hideConstraintsByView[subview]?.width.isActive = true
+
+            if self.axis == .horizontal {
+                self.hideConstraintsByView[subview]?.width.isActive = true
+            }
             self.hideConstraintsByView[subview]?.height.isActive = true
```
VojtaStavik, line 297: Isn't there a missing `else` there?

Sources/StreamChatUI/CommonViews/ContainerStackView.swift
```diff
@@ -298,8 +302,9 @@ public class ContainerStackView: UIView {
         }
     }
 
-    func showSubview(_ subview: UIView, animated: Bool = true) {
-        assert(subviews.contains(subview))
+    public func showSubview(_ subview: UIView, animated: Bool = true) {
+        guard subviews.contains(subview) else { return }
+        guard isHidden(subview) else { return }
```
VojtaStavik, line 307: tbh: I'd find doing `guard subview.alpha == 0 else { return }` less confusing than this helper.

iMessageClone/iMessageChatMessageComposerView.swift
```diff
@@ -10,31 +10,58 @@ final class iMessageChatMessageComposerView: ChatMessageComposerView {
...
+
+        centerContentContainer.layer.cornerRadius = 18
         messageInputView.textView.font = .systemFont(ofSize: 15)
         
         attachmentButton.setImage(
             UIImage(systemName: "camera.fill"),
             for: .normal
         )
+        attachmentButton.widthAnchor.constraint(equalToConstant: 30).isActive = true
+        attachmentButton.widthAnchor.constraint(equalToConstant: 30).isActive = true
```
VojtaStavik, line 65: same line twice

VojtaStavik (approved):
OK 👍 Let's deal with the details in a follow up PR

Sources/StreamChatUI/CommonViews/ContainerStackView.swift
nuno-vieira: No, it is fine like this, only if horizontal we want to activate it, but if not, we don't need to anything more 👍

PR #997, opened 2021-04-17
[CIS-795] Remove defaultAppearance Hook

DominikBucher12:
# Motivation for PR:
- `DefaultAppearance` hook usage was ambigous and we agreed that we should delete the implementation because we cannot **restrict the usage of this method only to modify appearance** and not actually code for layout and other properties which could potentially lead to side effects because of the nature of `StreamChat._View` lifecycle. 

# What this PR does: 
- Removes `Customizable.defaultAppearance` method which was responsible for default `StreamChat` appearance. With this it also removes `defaultAppearance hook`  which was the root motivation for this PR. Instead of using `defaultAppearance` in our codebase, we switch to overriding `setUpAppearance` with `super.setUpAppearance` calls.
# How to test it
- The biggest prove should be the tests being green. Nevertheless always try to run the app and take a look if something is behaving strange or is different layout then before, if yes we should fix it and add test to it as it's not covered. 

Please check the code snippets and documentation if I should improve something or is not clear.


Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
```diff
@@ -104,7 +104,7 @@ open class _ChatChannelListVC<ExtraData: ExtraDataTypes>: _ViewController,
         loadingIndicator.pin(anchors: [.centerX, .centerY], to: view)
     }
     
-    override public func defaultAppearance() {
+    override open func setUpAppearance() {
```
VojtaStavik, line 107: We need to call `super` for `setUpAppearance` now

DominikBucher12: ✅

Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView_Tests.swift
```diff
@@ -50,21 +50,6 @@ class ChatChannelListItemView_Tests: XCTestCase {
...
-        }
-        
-        let view = TestView().withoutAutoresizingMaskConstraints
-        
-        view.addSizeConstraints()
-        
-        view.content = content
-        AssertSnapshot(view)
-    }
-
```
nuno-vieira, line 67: We need to remove all snapshot files that have `_usingAppearanceHook` and are now obsolete 😄

DominikBucher12: Yes :D Deleted :)

VojtaStavik (approved):
LGTM once you fix the snapshot zero-size snapshot tests 👍

Sources/StreamChatUI/ChatMessageList/ChatChannelVC.swift
```diff
@@ -34,9 +34,9 @@ open class _ChatChannelVC<ExtraData: ExtraDataTypes>: _ChatVC<ExtraData> {
         return navbarListener
     }
 
-    override public func defaultAppearance() {
-        super.defaultAppearance()
-
+    override open func setUpAppearance() {
+        super.setUpAppearance()
+        super.setUpAppearance()
```
skorepak, line 39: Double super 👀

skorepak (approved):

Sources/StreamChatUI/ChatMessageList/ChatChannelVC.swift
DominikBucher12: Fixed.

Sources/StreamChatUI/CommonViews/BaseViews.swift
```diff
@@ -76,12 +76,11 @@ open class _View: UIView, AppearanceSetting, Customizable {
         
         setUp()
         setUpLayout()
-        (self as! Self).applyDefaultAppearance()
+
```
DominikBucher12: Deleted as well 😅

PR #998, opened 2021-04-19
[CIS-526] Expose thread participants as a set of ChartUsers instead of a set of UserIds

dmigach:
Update the way we expose thread participants. Rather than UserId we should expose them as @CoreDataLazy set of _ChatUser.
Also a part of this task was to make sure we serialise and save participants correctly, this seems to be fine. The only thing I noticed is that backend doesn't include imageURLs for participants into MessagePayload


Sources/StreamChatUI/MessageActionsPopup/ChatMessageActionsVC.swift
```diff
@@ -88,6 +88,9 @@ open class _ChatMessageActionsVC<ExtraData: ExtraDataTypes>: _ViewController, UI
                 threadReplyActionItem(),
                 copyActionItem()
             ]
+            
+            let threadReply = threadReplyActionItem()
+            threadReply.action(threadReply)
```
VojtaStavik, line 93: what's happening here? 🤯

dmigach: It was a temp workaround for testing thread replies since message actions are broken

Sources/StreamChat/Models/Message.swift
```diff
@@ -217,6 +219,16 @@ extension _ChatMessage {
...
+    
+    public var lastActiveThreadParticipant: _ChatUser<ExtraData.User>? {
+        func sortingCriteriaDate(_ user: _ChatUser<ExtraData.User>) -> Date {
+            user.lastActiveAt ?? user.userUpdatedAt
+        }
+        
+        return threadParticipants
+            .sorted(by: { sortingCriteriaDate($0) > sortingCriteriaDate($1) })
+            .first
+    }
```
VojtaStavik, line 231: I'm not sure this is needed tbh. It's very specific for the cases where you want to show just one avatar there. What if your UI requires two avatars, there? :)

VojtaStavik (approved):
I'm not sure about having the `lastThreadParticipant` convenience variable on the model directly. Maybe move to the file where it's needed in the UI SDK? Even the `isPinned` there was probably a mistake :/ 

Once fixed I think we're good to merge this 👍  Also, please add CHANGELOG entry for this change ;-)

Sources/StreamChat/Models/Message.swift
dmigach: It does make sense to move it to StreamChatUI, I'll check if the same can be done for isPinned

PR #999, opened 2021-04-19
Fix message popup

fortmarek:

- fix views having bigger height than they should -> I am not sure why [this](https://github.com/GetStream/stream-chat-swift/compare/bugfix/reactions?expand=1#diff-a41621945186dd3f456888db631f0bdc157e35fedb332477c9f2d9790e922d17L200) was there before, but the behavior was different from `UIStackView` which I found weird in this case. 
- fix not tappable reactions -> gesture recognizer added to `scrollView` was being triggered instead of the reactions


Sources/StreamChatUI/CommonViews/ContainerStackView.swift
```diff
@@ -195,14 +195,14 @@ public class ContainerStackView: UIView {
...
                 customConstraints.append(
-                    lView.widthAnchor.constraint(equalTo: rView.widthAnchor).priority(distribution == .equal ? .required : .lowest)
+                    lView.widthAnchor.constraint(equalTo: rView.widthAnchor)
                 )
             } else {
                 customConstraints.append(
                     lView.heightAnchor.constraint(equalTo: rView.heightAnchor)
-                        .priority(distribution == .equal ? .required : .lowest)
                 )
             }
```
VojtaStavik, line 207: cc @nuno-vieira 😅

Sources/StreamChatUI/CommonViews/ContainerStackView.swift
```diff
@@ -195,14 +195,14 @@ public class ContainerStackView: UIView {
 
         // Add constraints for distribution
         zip(subviews, subviews.dropFirst()).forEach { lView, rView in
+            guard distribution == .equal else { return }
```
VojtaStavik, line 198: we can probably move this `guard` outside the loop ;-)

VojtaStavik (commented):
So what was the issue?

fortmarek:
> So what was the issue?

The issue with the height of actions was that they "tried" to be as big as the message (the preceding view). Even with very low priority it influenced the view (one could most likely fix it with `setContentHugging` but I don't it'd be a proper solution here)

Sources/StreamChatUI/CommonViews/ContainerStackView.swift
nuno-vieira:
```
if distribution == .equal {
            zip(subviews, subviews.dropFirst()).forEach { lView, rView in
                if axis == .horizontal {
                    customConstraints.append(
                        lView.widthAnchor.constraint(equalTo: rView.widthAnchor)
                    )
                } else {
                    customConstraints.append(
                        lView.heightAnchor.constraint(equalTo: rView.heightAnchor)
                    )
                }
            }
        }
```

fortmarek: yeah, I have just done that 👍

VojtaStavik (approved):

VojtaStavik:
@fortmarek do you have time to fix the failing test and merge today?

PR #1001, opened 2021-04-20
[CIS-800] Fix socket connection

olejnjak:
When event decoding fails in `WebSocketClient` we always(!) set `connectionState` to `.disconnecting(...)` which causes socket connection to disconnect.

On this line we have type `WebSocketErrorContainer??` with value `.some(.none)` which means that `map` to `ClientError.WebSocket(...)` succeeds with `nil` error.

https://github.com/GetStream/stream-chat-swift/blob/2870fa2beeede18bced4394d88b226274024fa95/Sources/StreamChat/WebSocketClient/WebSocketClient.swift#L275

Because of that, this `if let` always succeeds and `connectionState` is changed.

https://github.com/GetStream/stream-chat-swift/blob/2870fa2beeede18bced4394d88b226274024fa95/Sources/StreamChat/WebSocketClient/WebSocketClient.swift#L278

In general if we ever got to this catch branch, we always ended with `connectionState == .disconnecting`.

Just a sidenote - I am not sure why I have changes in _Package.swift_, looks like we have in `main` some code that was not checked by our githooks.


VojtaStavik (approved):
Nice, thank you for finding this.

Sources/StreamChat/WebSocketClient/WebSocketClient.swift
```diff
@@ -272,7 +272,7 @@ extension WebSocketClient: WebSocketEngineDelegate {
             // Check if the message contains an error object from the server
             let webSocketError = message
                 .data(using: .utf8)
-                .map { try? JSONDecoder.default.decode(WebSocketErrorContainer.self, from: $0) }
+                .flatMap { try? JSONDecoder.default.decode(WebSocketErrorContainer.self, from: $0) }
```
DominikBucher12, line 275: ❤️ nice catch

DominikBucher12 (approved):
Nice :) 

The Changes in Package.swift are actually ok, Those files are not existing and actually in my workflow the Package changed a lot of times and I didn't update it....

PR #1003, opened 2021-04-21
[CIS-449] Custom photo gallery

fortmarek:

If image is tapped, new view controller `ImageGalleryVC` is shown. There user can go through images that were sent in the particular message.

The current implementation is built on top of `UICollectionView` to ensure that if a message has a larger number of images, only the subset that the user views is loaded.

There may be some rough edges but the basic functionality should work as intended.


Sources/StreamChatUI/ChatMessageList/ChatMessageListRouter.swift
```diff
@@ -42,4 +42,13 @@ open class _ChatMessageListRouter<ExtraData: ExtraDataTypes>: ChatRouter<_ChatMe
...
         rootViewController.present(navigation, animated: true)
     }
+    
+    open func showImageGallery(for message: _ChatMessage<ExtraData>) {
+        let imageGalleryVC = _ImageGalleryVC<ExtraData>()
+        imageGalleryVC.modalPresentationStyle = .overFullScreen
+        imageGalleryVC.content = message
+        
+        // TODO: Animated transition?
+        rootViewController.present(imageGalleryVC, animated: true)
```
fortmarek, line 52: images are usually enlarged to the image gallery -> do we want that, too? My suggestion is to probably do that in some next iteration

Sources/StreamChatUI/ImageCollectionViewCell.swift
```diff
@@ -0,0 +1,84 @@
...
+        
+        contentView.embed(imageScrollView)
+        imageScrollView.embed(imageView)
+        imageView.pin(anchors: [.height, .width], to: contentView)
+    }
+    
+    override open func updateContent() {
+        super.updateContent()
+
+        imageView.loadImage(from: content.imageURL, resizeAutomatically: false)
```
fortmarek, line 78: do we want to load the preview image url (if available first) - and only then the full image? Also should I try to reuse the image already loaded from the message list?

Sources/StreamChatUI/ImageGalleryVC.swift
```diff
@@ -0,0 +1,232 @@
...
+    }
+    
+    @objc
+    open func closeButtonTapped() {
+        dismiss(animated: true, completion: nil)
+    }
+    
+    @objc
+    open func shareButtonTapped() {
+        guard let imageURL = images[currentPage].imageURL else { return }
```
fortmarek, line 166: do we want to share the image URL or the image itself? Sharing the image itself _might_ be a little trickier to do, but possible

Sources/StreamChatUI/ChatMessageList/ChatMessageListRouter.swift
```diff
@@ -42,4 +42,13 @@ open class _ChatMessageListRouter<ExtraData: ExtraDataTypes>: ChatRouter<_ChatMe
         let navigation = UINavigationController(rootViewController: preview)
         rootViewController.present(navigation, animated: true)
     }
+    
+    open func showImageGallery(for message: _ChatMessage<ExtraData>) {
```
VojtaStavik, line 46: I think we should also pass the index of the image the user tapped on because that should be the one that you see first.

Sources/StreamChatUI/ChatMessageList/ChatMessageListRouter.swift
VojtaStavik: I think it's OK to do it now. We already have a similar thing in the message actions popover. It's important to have it because we for sure want to the interaction where you can swipe the image to close it.

Sources/StreamChatUI/UIConfig+Images.swift
```diff
@@ -15,6 +15,7 @@ public extension _UIConfig {
         public var back: UIImage = UIImage(named: "icn_back", in: .streamChatUI)!
         public var onlyVisibleToCurrentUser = UIImage(named: "eye", in: .streamChatUI)!
         public var more = UIImage(named: "icn_more", in: .streamChatUI)!
+        public var share = UIImage(named: "share", in: .streamChatUI)!
```
VojtaStavik, line 18: I'd say we should use the default system icon for this and only when it's not available use the custom one.

Sources/StreamChatUI/ImageGalleryVC.swift
```diff
@@ -50,6 +50,10 @@ open class _ImageGalleryVC<ExtraData: ExtraDataTypes>: _ViewController, UIConfig
     public private(set) lazy var currentPhotoLabel = UILabel()
         .withoutAutoresizingMaskConstraints
     
+    public private(set) lazy var closeButton = _Button()
+    
+    public private(set) lazy var shareButton = _Button()
+
```
VojtaStavik, line 56: I wouldn't be afraid to make components for these buttons. I'm sure it's not the only place where we will need them.

VojtaStavik (commented):
Cool, looks very promising! 🎉

fortmarek:
@VojtaStavik this is ready for another round, I have implemented the animated and interactive transition. In the meantime, I'll start writing snapshot tests for `ImageGalleryVC` before this will be in mergeable state.

Sources/StreamChatUI/CommonViews/CloseButton.swift
```diff
@@ -0,0 +1,19 @@
...
+
+/// A Button subclass that should be used for closing.
+open class _CloseButton<ExtraData: ExtraDataTypes>: _Button, AppearanceProvider {
+    override open func setUpAppearance() {
+        super.setUpAppearance()
+        
+        setImage(appearance.images.close, for: .normal)
+        tintColor = appearance.colorPalette.text
+    }
+}
```
VojtaStavik, line 19: The button no longer needs to be generic over `ExtraData` ;-)

Sources/StreamChatUI/CommonViews/ShareButton.swift
```diff
@@ -0,0 +1,19 @@
...
+//
+
+import StreamChat
+import UIKit
+
+/// A Button subclass that should be used for sharing content
+public typealias ShareButton = _ShareButton<NoExtraData>
+
+/// A Button subclass that should be used for sharing content
+open class _ShareButton<ExtraData: ExtraDataTypes>: _Button, AppearanceProvider {
```
VojtaStavik, line 12: Same here...

Sources/StreamChatUI/CommonViews/CloseButton.swift
```diff
@@ -0,0 +1,19 @@
...
+/// A Button subclass that should be used for closing.
+public typealias CloseButton = _CloseButton<NoExtraData>
+
+/// A Button subclass that should be used for closing.
+open class _CloseButton<ExtraData: ExtraDataTypes>: _Button, AppearanceProvider {
+    override open func setUpAppearance() {
+        super.setUpAppearance()
+        
+        setImage(appearance.images.close, for: .normal)
+        tintColor = appearance.colorPalette.text
```
VojtaStavik, line 17: Can you please add the highlighted state image, too? Also for `ShareButton`

Sources/StreamChatUI/ImageGallery/ImageGalleryVC.swift
```diff
@@ -0,0 +1,305 @@
...
+        )
+        attachmentsCollectionView.setContentOffset(contentOffset, animated: false)
+    }
+    
+    override open func updateContent() {
+        super.updateContent()
+        
+        images = content.attachments
+            .filter { $0.type == .image }
+            .compactMap { $0 as? ChatMessageDefaultAttachment }
```
VojtaStavik, line 203: Since `updateContent` can be called multiple times, should we reload the CW when assigning the new `images` value?

Sources/StreamChatUI/ImageGallery/ImageGalleryVC.swift
```diff
@@ -0,0 +1,305 @@
...
+        
+        userLabel.text = content.author.name
+        dateLabel.text = L10n.sent + " " + dateFormatter.string(from: content.createdAt).firstLowercased
```
VojtaStavik, line 206: This should be "Seen ...." and it's about when the author was active the last time. We have the logic already in the channel list.

https://www.figma.com/file/ekifwChR9tR7zRJg1QEzSM/Chat-Sample-App?node-id=12494%3A8697

VojtaStavik (approved):
Amazing! 😍 Great work @fortmarek 🎉

VojtaStavik:
@fortmarek please merge when you're ready :)

PR #1004, opened 2021-04-21
`ContainerStackView` new hiding/showing API + Cleanup

nuno-vieira:
# In this PR
- `ContainerStackView` now observes `isHidden` property from its subviews, to hide and show them in the container.
- Re-organizes and adds documentation to the `ContainerStackView`
- Removes `addArrangedSubview(builder:)` redundant API
- Removes ordering API from `ContainerStackView` since it can just be achieved by placing the subviews in the order the user wants in the `addArrangedSubviews()`.
- Add `removeAllArrangedSubviews()` to `ContainerStackView`
- Add Subviews()` to `ContainerStackView`


Sources/StreamChatUI/CommonViews/ContainerStackView.swift
```diff
@@ -302,13 +314,14 @@ public class ContainerStackView: UIView {
...
-    public func showSubview(_ subview: UIView, animated: Bool = true) {
+    /// Shows the arranged subview by setting the width, height and spacing constraints to the original value before being hidden.
+    private func showArrangedSubview(_ subview: UIView) {
         guard subviews.contains(subview) else { return }
         guard subview.alpha == 0 else { return }
 
         updateConstraintsIfNeeded()
 
-        Animate(isAnimated: animated) {
+        Animate(isAnimated: true) {
```
evsaev, line 335: Do you think we need the implicit animations inside the container view? 🤔What if by default the hide/show works without any animation but if we need one we can be explicit:
```swift
Animate { subview.isHidden = true }
```

evsaev:
@nuno-vieira here is the code snippet that makes the container have the smallest possible size:

```swift
if alignment != .fill {
 // For the containers where subview anchors are not `=` to its layout guide (but are `>=` or `<=`)
 // we set low priority size constraint. The layout engine will try to minimise errors for those constraints.
 // As a result, the container will be the smallest possible that fits the content.
 customConstraints += [
  widthAnchor.constraint(equalToConstant: 0).with(priority: .init(1)),
  heightAnchor.constraint(equalToConstant: 0).with(priority: .init(1))
 ]
}
```

Sources/StreamChatUI/CommonViews/ContainerStackView.swift
nuno-vieira: Ah yes, you are totally right! Will change this 👍

Sources/StreamChatUI/CommonViews/ContainerStackView.swift
```diff
@@ -113,17 +114,27 @@ public class ContainerStackView: UIView {
...
 
+        hidingObserversByView[subview] = subview
+            .observe(\.isHidden, options: [.new]) { [weak self] (view, isHiddenChange) in
+                if isHiddenChange.newValue == true {
+                    self?.hideArrangedSubview(view)
+                } else {
+                    self?.showArrangedSubview(view)
+                }
+            }
+
```
evsaev, line 142: Great! 👍 I was thinking to also observe `superview` property. When the `superview` is changed the container is not responsible for this subview anymore and shouldn't track the `isHidden` field. It also will be super easy to move views between containers.

nuno-vieira: In order to remove the subview from the container, we need to call `removeArrangedSubview`, which will remove the observer. Although you can use `removeFromSuperview` in a view inside the container, that is not expected, and it is a misuse of the Container, probably this should also be mentioned in the `removeArrangedSubview` documentation, I will add this 👍 But for me this is enough, I don't think it necessary to observe the `superview` property, or I'm missing something else?

evsaev: With `superview` observer it would be safe to do `myView.removeFromSuperview` or `anotherView.addSubview(myView)` or `anotherContainer.addArrangedSubview(myView)`. The idea was to have the insurance from misuse but on the other hand, this encourages misuse since nothing will get broken 😄 You are right, we don't need it 👍

Sources/StreamChatUI/CommonViews/ContainerStackView.swift
```diff
@@ -4,70 +4,81 @@
...
+/// The aim of this view is to make UI customizability easier in the SDK.
 public class ContainerStackView: UIView {
+
+    private var customConstraints: [NSLayoutConstraint] = []
+
+    private var customSpacingByView: [UIView: CGFloat] = [:]
+
+    private var spacingConstraintsByView: [UIView: NSLayoutConstraint] = [:]
+
+    private var hideConstraintsByView: [UIView: (width: NSLayoutConstraint, height: NSLayoutConstraint)] = [:]
```
evsaev, line 44: Shouldn't setting either `height=0` OR `width=0` work? 
```swift
private var hideConstraintsByView: [UIView: NSLayoutConstraint] = [:]

// hideArrangedSubview
hideConstraintsByView[subview] = {
 if axis == .horizontal {
  return subview.widthAnchor.constraint(equalToConstant: 0)
 } else {
  return subview.heightAnchor.constraint(equalToConstant: 0)
 }
}()
```

Sources/StreamChatUI/ChatChannelList/ChatChannelUnreadCountView.swift
```diff
@@ -48,7 +48,7 @@ open class _ChatChannelUnreadCountView<ExtraData: ExtraDataTypes>: _View, UIConf
     }
     
     override open func updateContent() {
-        isHidden = content.mentionedMessages == 0 && content.messages == 0
+        alpha = content.mentionedMessages == 0 && content.messages == 0 ? 0 : 1
```
VojtaStavik, line 51: Why this change?

Sources/StreamChatUI/CommonViews/ContainerStackView.swift
nuno-vieira: Hum, maybe 🤔 But I was thinking right now... what Happens if the view that is being hidden has the bigger width in a vertical stack, and the bigger height in a horizontal stack? 🤔 Won't that stack the stack to have empty space?

VojtaStavik: I think we accidentally introduced a regression when changing the container in the past. I think both of the constraints are needed because every view can affect both axes.

Vertical container view like this:
```
┌ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┐
 ┌───────────────────────────────────────────┐ 
││                                           ││
 │                     A                     │ 
││                                           ││
 └───────────────────────────────────────────┘ 
│┌──────────────────────┐                     │
 │                      │                      
││          B           │                     │
 │                      │                      
│└──────────────────────┘                     │
 ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ 
```
should look like this when you hide the A. Even though the axis is vertical, the A affected the size of the container in both axes.

```           
┌ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─                      
 ┌──────────────────────┐│                     
││                      │                      
 │          B           ││                     
││                      │                      
 └──────────────────────┘│                     
└ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─                      
```

Sources/StreamChatUI/CommonViews/ContainerStackView.swift
```diff
@@ -336,6 +364,35 @@ public class ContainerStackView: UIView {
...
+    }
+}
+
+extension UILayoutPriority {
+    static let lowest = UILayoutPriority(defaultLow.rawValue / 2.0)
+}
+
+extension CGFloat {
+    public static let auto: CGFloat = .infinity
+}
```
VojtaStavik: can we rename this to `systemDefinedContainerSpacing`? `auto` is too general for a public extension on `CGFloat`

Sources/StreamChatUI/CommonViews/ContainerStackView.swift
```diff
@@ -311,11 +311,6 @@ public class ContainerStackView: UIView {
             : safeAreaLayoutGuide
     }
 
-    public func addArrangedSubview(_ builder: () -> UIView) {
-        addArrangedSubview(builder())
-    }
-
-
```
VojtaStavik: This reminds me, I think it would make sense to add
```
public func addArrangedSubviews(_ views: [UIView])
public func removeAllArrangedSubviews()
```

VojtaStavik (approved):
Great 👍 👏

Sources/StreamChatUI/CommonViews/ContainerStackView.swift
```diff
@@ -4,70 +4,81 @@
...
+    /// Describes the alignment of the arranged subviews in perpendicular to the container's axis.
+    public struct Alignment: Equatable {
+        /// Makes the arranged subviews so that they **fill** the available space perpendicular to the container’s axis.
+        public static let fill = Alignment(rawValue: 0)
+        /// Makes the arranged subviews align to the **leading edge** in a **vertical axis** container.
+        /// Makes the arranged subviews align to the **top edge** in a **horizontal axis** container.
+        public static let axisLeading = Alignment(rawValue: 1)
+        /// Makes the arranged subviews align to the **trailing edge** in a **vertical axis** container.
+        /// Makes the arranged subviews align to the **bottom edge** in a **horizontal axis** container.
+        public static let axisTrailing = Alignment(rawValue: 2)
```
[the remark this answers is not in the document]
nuno-vieira: Yes that is a good idea 👍 I like that way better, also is more similar with what we have in a normal UIStackView 👍

Sources/StreamChatUI/ChatChannelList/ChatChannelUnreadCountView.swift
nuno-vieira: The unreadCountView when is empty, we just want to hide the opacity. But because this view is present in a container stack view, that will hide the view from the container, and re-layout the views inside, make the timestamp label to the centre of the container, which we don't want.

Sources/StreamChatUI/CommonViews/ContainerStackView.swift
nuno-vieira: Yes the `public func addArrangedSubviews(_ views: [UIView])` we already have, but not the `removeAllArrangedSubviews()`. It makes sense to add it 👍 And maybe adding a method the replace all the arrangedSubviews with the provided array? 🤔

Sources/StreamChatUI/CommonViews/ContainerStackView.swift
nuno-vieira: Actually, I think it is a better idea to create a `ContainerStackView.Spacing`? So that we don't influence the `CGFloat` of the customer. And we can make `ContainerStackView.Spacing` conform to `ExpressibleByFloatLiteral` so that we can still use it like `spacing = 10`.

Example:
```Swift
   public struct Spacing: Equatable, ExpressibleByFloatLiteral {

        public static let auto = Spacing(rawValue: .infinity)

        public init(floatLiteral value: Double) {
            rawValue = CGFloat(value)
        }

        init(rawValue: CGFloat) {
            self.rawValue = rawValue
        }

        var rawValue: CGFloat
    }
```

WDYT?

VojtaStavik: It's a bit overkill but yeah, why not :) However, I think we should expose the raw value publicly then. There's no reason IMO why it shouldn't be.

nuno-vieira: Ah yes it should be public. I made a WIP Implementation here: https://github.com/GetStream/stream-chat-swift/pull/1004/commits/450338621aba6bf23b2134510f1bd5a49b14eb4a, let me know what you think please @VojtaStavik

Sources/StreamChatUI/ChatMessageList/TypingIndicatorView.swift
```diff
@@ -47,7 +47,7 @@ open class _TypingIndicatorView<ExtraData: ExtraDataTypes>: _View, UIConfigProvi
         componentContainerView.alignment = .center
         // UIStackView.spacingUseSystem returns 0, because we are not really in stackView in here,
         // so let's use `directionalLayoutMargins`
-        componentContainerView.spacing = directionalLayoutMargins.trailing
+        componentContainerView.spacing = .init(rawValue: directionalLayoutMargins.trailing)
```
VojtaStavik, line 50: btw I don't think this line is actually needed now

Sources/StreamChatUI/CommonViews/ContainerStackView.swift
VojtaStavik: @nuno-vieira all right, let's do it :)

Sources/StreamChatUI/ChatMessageList/TypingIndicatorView.swift
nuno-vieira: Yes, `auto` should be enough right?

VojtaStavik: Yep

Sources/StreamChatUI/CommonViews/ContainerStackView.swift
nuno-vieira: ✅  Done!

Sources/StreamChatUI/CommonViews/ContainerStackView.swift
nuno-vieira: ✅  Done!

Sources/StreamChatUI/CommonViews/ContainerStackView.swift
nuno-vieira: ✅  Done! The new hiding algorithm is implemented

Sources/StreamChatUI/CommonViews/ContainerStackView.swift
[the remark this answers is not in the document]
nuno-vieira: ✅  Done!

Sources/StreamChatUI/CommonViews/ContainerStackView.swift
```diff
@@ -32,6 +32,32 @@ extension ContainerStackView {
...
+
+        /// `Spacing` can be expressed by an `Int`.
+        /// Example: `spacing = 10`, instead of `spacing = Spacing(10)`.
+        public init(integerLiteral value: Int) {
+            rawValue = CGFloat(value)
+        }
+
+        init(_ rawValue: CGFloat) {
+            self.rawValue = rawValue
+        }
```
VojtaStavik, line 55: I believe this should be `public`, too. If you can do `spacing = 5`, you should be able to do `spacing = Spacing(myConstant)`, like you do in tests :)

Sources/StreamChatUI/CommonViews/ContainerStackView.swift
```diff
@@ -137,16 +179,17 @@ public class ContainerStackView: UIView {
...
-        setNeedsUpdateConstraints()
+    /// Changes the spacing after a specific view.
+    /// - Parameters:
+    ///   - spacing: The value of the spacing.
+    ///   - subview: The subview that the spacing will be applied (after this subview).
+    func setCustomSpacing(_ spacing: CGFloat, after subview: UIView) {
+        assert(subviews.contains(subview))
+        customSpacingByView[subview] = spacing
+        invalidateConstraints()
     }
```
VojtaStavik: this should be `public`

Sources/StreamChatUI/CommonViews/ContainerStackView.swift
```diff
@@ -222,18 +266,18 @@ public class ContainerStackView: UIView {
             }
         }
 
+        // Add constraints for the layout alignment.
         subviews.forEach { subview in
-
             if axis == .horizontal {
-                if alignment == .axisLeading || alignment == .fill {
+                if alignment == .leading || alignment == .top || alignment == .fill {
```
VojtaStavik: This is not really needed, `top` has the same raw value as `leading`. It's just syntax sugar :)

VojtaStavik (approved):
LGTM 🎉 🚀

DominikBucher12 (approved):
This looks great to me 👍

Sources/StreamChatUI/CommonViews/ContainerStackView.swift
nuno-vieira: Yes nice catch 👍 Thanks!

PR #1007, opened 2021-04-21
Fix Application Crash for ChatMessageCollectionViewCell unknown ReuseID

DominikBucher12:
TL;DR of this PR is That there is no Cell Reuse identifier for message cell with no `layoutOptions`. This PR add support for this. 

# What this PR do:
- Fixes issue where application could crash when the cell has no layout options and there is no reuse identifier for such cell because there is no explicit 0 rawValue option.
# How to test it
- Right now if you go to Dem App -> Luke Skywalker -> C-3PO conversation, you will experience a crash associated with no reuse identifier, try this PR and see if it works.


VojtaStavik (approved):
Thank you 👍 We'll be changing this part soon but it's great to have this fixed 👍

skorepak (approved):

PR #1013, opened 2021-04-23
[CIS-804] Implement channel hiding filtering logic

VojtaStavik:
The channel list controller didn't react in real-time to the changes in the visibility of the channels. This PR fixes this by handling the related events explicitly.


nuno-vieira (approved):
LGTM!

PR #1018, opened 2021-04-26
Fix `CurrentUserController.synchronize()` flaky Tests

nuno-vieira:
# In this PR

Fixes 2 flaky tests on `CurrentUserController.synchronize()`

### DIFF
--- a/Sources/StreamChat/Controllers/CurrentUserController/CurrentUserController_Tests.swift
+++ b/Sources/StreamChat/Controllers/CurrentUserController/CurrentUserController_Tests.swift
@@ -101,17 +101,20 @@ final class CurrentUserController_Tests: StressTestCase {
         // Simulate current user
         env.currentUserObserverItem = .mock(id: .unique)
 
-        var synchronizeError: Error?
+        var synchronizeCalled = false
         controller.synchronize { [callbackQueueID] error in
             AssertTestQueue(withId: callbackQueueID)
-            synchronizeError = error
+            XCTAssertNil(error)
+            synchronizeCalled = true
         }
 
         // Simulate connection successful
         client.simulateProvidedConnectionId(connectionId: .unique)
 
-        XCTAssertEqual(controller.state, .remoteDataFetched)
-        XCTAssertNil(synchronizeError)
+        AssertAsync {
+            Assert.willBeEqual(self.controller.state, .remoteDataFetched)
+            Assert.willBeTrue(synchronizeCalled)
+        }
     }
 
     func test_synchronize_changesControllerStateOnError() {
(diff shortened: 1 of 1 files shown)


Sources/StreamChat/Controllers/CurrentUserController/CurrentUserController_Tests.swift
```diff
@@ -122,16 +127,21 @@ final class CurrentUserController_Tests: StressTestCase {
...
 
         // Simulate connection not successful
         client.simulateProvidedConnectionId(connectionId: nil)
 
-        XCTAssertEqual(controller.state, .remoteDataFetchFailed(.ConnectionNotSuccessful()))
-        XCTAssertNotNil(synchronizeError)
+        AssertAsync {
+            Assert.willBeEqual(self.controller.state, .remoteDataFetchFailed(.ConnectionNotSuccessful()))
+            Assert.willBeEqual(synchronizeError as? ClientError, ClientError.ConnectionNotSuccessful())
+            Assert.willBeTrue(synchronizeCalled)
```
VojtaStavik, line 143: This is redundant because if `synchronizeError` is non-nil it means the synchronize block got called :)

Sources/StreamChat/Controllers/CurrentUserController/CurrentUserController_Tests.swift
```diff
@@ -102,16 +102,21 @@ final class CurrentUserController_Tests: StressTestCase {
...
 
         // Simulate connection successful
         client.simulateProvidedConnectionId(connectionId: .unique)
 
-        XCTAssertEqual(controller.state, .remoteDataFetched)
-        XCTAssertNil(synchronizeError)
+        AssertAsync {
+            Assert.willBeEqual(self.controller.state, .remoteDataFetched)
+            Assert.willBeTrue(synchronizeCalled)
+            Assert.willBeNil(synchronizeError)
```
VojtaStavik, line 118: `synchronizeError` value is `nil` by default which means that this assert will succeed and even when the callback returns non-nil error, it won't be caught here. There's a couple of options:
- You can check the error is `nil` synchronously within the `synchronize` callback.
- You can change this to `staysNil`. It not 100% correct because it suggests the synchronize callback could be called multiple times.
- You can just do `XCTAssertNil(synchronizeError)` after the `AssertAsync { }` block.

VojtaStavik (approved):
👍  Thank you for fixing this! Please check my comments with suggestions and adjust the way you think makes the most sense.

Sources/StreamChat/Controllers/CurrentUserController/CurrentUserController_Tests.swift
nuno-vieira: ✅ Done

Sources/StreamChat/Controllers/CurrentUserController/CurrentUserController_Tests.swift
nuno-vieira: ✅ Done

PR #1023, opened 2021-04-27
Optimize current user fetch

olejnjak:
- fetch `CurrentUserDTO` only once for each context, then rely on CoreData that its properties are up to date
- convert `func currentUser() -> CurrentUserDTO?` to `var currentUser: CurrentUserDTO?` to indicate that it is not fetched on each call
- use context's `userInfo` to store fetched user as I think this is exactly the kind of information that should be stored there (other option would be associated object)
- when `currentUser` is called for the first time (and the fetch from DB is performed) I register for the `DatabaseContainer.WillRemoveAllDataNotification` so I can clear cached user as leaving it there would create inconsistencies - there are apparently other options how to do this, but I think this responsibility should remain in this part of code

### DIFF
--- a/Sources/StreamChat/ChatClient.swift
+++ b/Sources/StreamChat/ChatClient.swift
@@ -302,10 +302,10 @@ public class _ChatClient<ExtraData: ExtraDataTypes> {
 
         let context = databaseContainer.viewContext
         if Thread.isMainThread {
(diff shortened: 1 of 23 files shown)


VojtaStavik (approved):
Very nice 👍 Thank you

nuno-vieira (approved):
LGTM! ✅

PR #1024, opened 2021-04-27
[CIS-807] Add `controllerWillChangeChannels` delegate callback

VojtaStavik:
This is mainly to avoid the crashes when `UICollectionView` is updated before it's fully laid out. We already try other (simpler) fixes but none of those worked properly. This is hopefully the last fix of this issue.

- I added `willChange` callback to `ListDatabaseObsever`
- I added `controllerWillChangeChannels` to `ChatChannelListController` delegate
- `ChatChannelListVC` now forces its `UICollectionView` to lay out before the update form the controller comes

---

If we see this pattern works and helps with the issue, we can extend other list controllers, too.

### DIFF
--- a/Sources/StreamChat/Controllers/ChannelListController/ChannelListController_Tests.swift
+++ b/Sources/StreamChat/Controllers/ChannelListController/ChannelListController_Tests.swift
@@ -283,7 +283,10 @@ class ChannelListController_Tests: StressTestCase {
 
         let channel: ChatChannel = client.databaseContainer.viewContext.channel(cid: cid)!.asModel()
         
-        AssertAsync.willBeEqual(delegate.didChangeChannels_changes, [.insert(channel, index: [0, 0])])
+        AssertAsync {
+            Assert.willBeTrue(delegate.willChangeChannels_called)
+            Assert.willBeEqual(delegate.didChangeChannels_changes, [.insert(channel, index: [0, 0])])
+        }
     }
     
     func test_genericDelegateMethodsAreCalled() throws {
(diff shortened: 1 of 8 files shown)


Sources/StreamChat/Controllers/ChannelListController/ChannelListController_Tests.swift
```diff
@@ -420,13 +473,19 @@ private class TestEnvironment {
...
     func controller(_ controller: DataController, didChangeState state: DataController.State) {
         self.state = state
         validateQueue()
     }
-    
+
+    func controllerWillChangeChannels(_ controller: ChatChannelListController) {
+        willChangeChannels_called = true
+        validateQueue()
+    }
```
olejnjak, line 487: Just for a consideration, I usually prefer to create a closure that allows to define what will be executed in function so for this case I'd have something like this:

```swift
var controllerWillChangeChannelsBody: (ChatChannelListController) -> Void = { _ in }

func controllerWillChangeChannels(_ controller: ChatChannelListController) {
    controllerWillChangeChannelsBody(controller)
    validateQueue() // not sure if it would come here or to the default body
}
```

This approach is a bit more verbose but generally gives you flexibility in tests. But as I mentioned, just for future considerations.

Generally it is annoying to switch to this approach, especially with 1100 tests, but it is quite simple and fast to use it in new tests.

olejnjak (approved):
LGTM!

Sources/StreamChat/Controllers/ChannelListController/ChannelListController_Tests.swift
VojtaStavik: @olejnjak I see the benefits of this approach in terms of better flexibility, on the other hand, it also requires bigger setup in the test:

```swift
controller.simulateSomething()
AssertAsync.willBeTrue(mock.willChangeChannels_called)

vs.

var wasCalled = false
mock.controllerWillChangeChannelsBody = { wasCalled = true  }

controller.simulateSomething()
AssertAsync.willBeTrue(wasCalled)
```

It's quite common for our current test setup to have multiple "mocks" for the same object but with different functionalities. So instead of having one almighty global mock, we have multiple local mocks. This basically replicates what you suggest but with concrete types rather than closures.

In this comparison, I don't see any clear winner there. However, thank you for pointing this pattern out. I'm sure there are cases where it works better than the "hardcoded" mock style.

PR #1025, opened 2021-04-27
Restructure the Input View + Remove manual height calculation

nuno-vieira:
# In this PR
The focus of this PR is to remove the manual height calculation of the old Input Text View, and re-organizes the Input Container View, although this still might change in the next PRs, please look below the current and possible design of `ChatMessageInputView`. Keep in mind the `content` property is still mostly in the `ComposerVC`, but in the next PR's the aim is for `ChatMessgeInputView` to have it's own `content` property.

### Summary:
- Removes manual height calculation of input text view.
- Changes `ChatMessageComposerInputTextView` -> `ChatInputTextView`
- Changes `ChatMessageInputSlashCommandView` -> `ChatCommandLabel`
- Changes `ChatMessageComposerInputContainerView` -> `ChatMessageInputView`

### Current design of `ChatMessageInputView`:
- `ChatMessageInputView`
    - `ChatInputTextView`
    - `ChatCommandLabel`
    
### Future design of `ChatMessageInputView`:
- `ChatMessageInputView`
    - `ChatInputTextView`
    - `ChatCommandLabel`
    - `ImageAttachmentsCollectionView`
    - `DocumentsAttachmentsCollectionView`
    - `MessageQuoteView`

The idea is that the `ChatMessageInputView` contains all the content that can be part of a new message. So the `ChatMessageInputView` is essentially represented by this part of the composer:


VojtaStavik (approved):
Looks straightforward and clean 👍 🎉

Sources/StreamChatUI/CommonViews/ChatInputView/ChatCommandLabel.swift
```diff
@@ -5,44 +5,43 @@
...
         didSet {
             updateContentIfNeeded()
         }
     }
 
-    // MARK: - Subviews
-    
-    public private(set) lazy var container = ContainerStackView()
+    /// The container stack view that layouts the label and the icon view.
+    public private(set) lazy var container: ContainerStackView = ContainerStackView()
```
evsaev, line 21: Can we let the compiler do its job? 😄
```swift
public private(set) lazy var container = ContainerStackView()
```

Sources/StreamChatUI/UIConfig.swift
```diff
@@ -28,6 +28,9 @@ public struct _UIConfig<ExtraData: ExtraDataTypes> {
 
     /// A view for inputting text with placeholder support.
     public var inputTextView: _ChatInputTextView<ExtraData>.Type = _ChatInputTextView<ExtraData>.self
+
+    /// A view that displays the command name and icon.
+    public var commandLabel: _ChatCommandLabel<ExtraData>.Type = _ChatCommandLabel<ExtraData>.self
```
evsaev, line 33: It looks the same can be done in config 
```swift
public var commandLabel = _ChatCommandLabel<ExtraData>.self
```

Sources/StreamChatUI/CommonViews/ChatInputView/ChatCommandLabel.swift
```diff
@@ -5,44 +5,43 @@
...
         .withoutAutoresizingMaskConstraints
-    
-    private lazy var commandLabel = UILabel()
+
+    /// An `UILabel` that displays the command name.
+    private lazy var commandLabel: UILabel = UILabel()
```
evsaev, line 25: Do we need dynamic fonts here? 🤔
```swift
UILabel()
```

Sources/StreamChatUI/CommonViews/ChatInputView/ChatCommandLabel.swift
```diff
@@ -52,25 +51,27 @@ open class _ChatMessageInputSlashCommandView<ExtraData: ExtraDataTypes>: _View,
...
+        iconView.isHidden = false
         commandLabel.isHidden = false
         
         iconView.contentMode = .scaleAspectFit
+        layer.cornerRadius = bounds.height / 2
     }
     
     override open func updateContent() {
-        commandLabel.text = commandName
+        commandLabel.text = content?.name.uppercased()
```
evsaev, line 75: Should we always invoke `super` in case we add smth there in the future?

Sources/StreamChatUI/CommonViews/ChatInputView/ChatCommandLabel.swift
nuno-vieira: This is part of the pattern with Audit of the components, that each subview should have its type explicit. I guess this is more important on the views that come from the `UIConfig` but to follow the same pattern, we do it everywhere if I'm not mistaken, right? CC @VojtaStavik Probably this is missing from the Audit Document, I'll add it

Sources/StreamChatUI/UIConfig.swift
nuno-vieira: Some as above ☝️

Sources/StreamChatUI/CommonViews/ChatInputView/ChatCommandLabel.swift
nuno-vieira: Good catch, yes I think we need 👍

Sources/StreamChatUI/CommonViews/ChatInputView/ChatCommandLabel.swift
nuno-vieira: I'm actually not sure 🤔 But now the pattern is that we should avoid at all costs to subclass views, and only use composition. But I think all views are calling super, so to follow the same pattern, I'll do the same, thanks 👍

evsaev (approved):
Great job 👍

Sources/StreamChatUI/CommonViews/ChatInputView/ChatCommandLabel.swift
VojtaStavik: Yes, we try to make it easier for the user by specifying the type explicitly. As @nuno-vieira said, for UIConfig views, it's important much more, here it doesn't matter that much :)

PR #1026, opened 2021-04-28
Fix CoreData concurrency

### LINKED ISSUE #954: Chat Client crashes upon initialization
## What did you do?
I upgraded from version 3.1.1 to version 3.1.4 of the StreamChat SDK.

I expected to see [new Channel properties](https://github.com/GetStream/stream-chat-swift/releases/tag/3.1.4)  working.

## What happened instead?

App crashed upon client initialization. It does happen both with clean installs and installing over previous data (BTW, it seems that a CoreData migration is needed upgrading from 3.1.1, and the crash happens right after migrating). 

Crash is in `NSManagedObjectContext+Extensions`:

```swift
extension NSManagedObjectContext {
    private static let localCachingKey = "io.getStream.StreamChat.local_caching_key"
    
    /// Provides the defaults for local caching and model serialization for this context.
    var localCachingSettings: ChatClientConfig.LocalCaching? {
        get { userInfo[Self.localCachingKey] as? ChatClientConfig.LocalCaching }
        set { userInfo[Self.localCachingKey] = newValue }
    }
}
```

The crash is a bad access: `Thread 1: EXC_BAD_INSTRUCTION (code=EXC_I386_INVOP, subcode=0x0)`, happening in the `set` hook.

## GetStream Environment
**GetStream Chat version:** 3.1.4
**GetStream Chat frameworks:** StreamChat
**iOS version:** 14.4
**Swift version:** 5.2
**Xcode version:** 12.4
**Device:** both simulators and real devices

## Additional context

It is happening 100% of the times, i.e. our app is completely stuck. Here is how we initialize it:

```swift
var config = ChatClientConfig(apiKeyString: apiKey ?? "")
config.baseURL = .dublin
config.shouldConnectAutomatically = false
config.shouldFlushLocalStorageOnStart = false
config.localCaching.chatChannel.lastActiveMembersLimit = 2
config.localCaching.chatChannel.lastActiveWatchersLimit = 2
config.localCaching.chatChannel.latestMessagesLimit = 10

client = MyChatClient(config: config, tokenProvider: .closure { chatClient, completion in
  // our token refresh implementation
  [...]

  self.client.connectionController().connect()
})
```

olejnjak:
This PR fixes CoreData concurrency issues found by enabling `-com.apple.CoreData.ConcurrencyDebug` launch argument for demo app and for `StreamChat` test plans.

- I kept the launch argument for demo app and for both LLC test plans as I think it is useful to know about those issues
- fix issues in demo app
- fix issues in LLC tests

---

Closes #954

### DIFF
--- a/Sources/StreamChat/Controllers/MessageController/MessageController_Tests.swift
+++ b/Sources/StreamChat/Controllers/MessageController/MessageController_Tests.swift
@@ -382,15 +382,15 @@ final class MessageController_Tests: StressTestCase {
             authorUserId: .unique
         )
         
-        var replyDTO: MessageDTO?
+        var replyModel: ChatMessage?
         try client.databaseContainer.writeSynchronously { session in
-            replyDTO = try session.saveMessage(payload: reply, for: self.cid)
+            replyModel = try session.saveMessage(payload: reply, for: self.cid).asModel()
         }
     
         // Assert `insert` entity change is received by the delegate
         AssertAsync.willBeEqual(
             delegate.didChangeReplies_changes,
-            [.insert((replyDTO?.asModel())!, index: [0, 0])]
+            [.insert(replyModel!, index: [0, 0])]
         )
     }
     
--- a/Sources/StreamChat/Database/DTOs/MessageDTO_Tests.swift
+++ b/Sources/StreamChat/Database/DTOs/MessageDTO_Tests.swift
@@ -263,18 +263,25 @@ class MessageDTO_Tests: XCTestCase {
         )
 
         let (channelDTO, messageDTO): (ChannelDTO, MessageDTO) = try await { completion in
+            var channelDTO: ChannelDTO!
+            var messageDTO: MessageDTO!
+            
             // Asynchronously save the payload to the db
             database.write { session in
                 // Create the channel first
-                let channelDTO = try! session.saveChannel(payload: channelPayload, query: nil)
+                channelDTO = try! session.saveChannel(payload: channelPayload, query: nil)
 
                 // Save the message
-                let messageDTO = try! session.saveMessage(payload: payload, for: channelId)
+                messageDTO = try! session.saveMessage(payload: payload, for: channelId)
+            } completion: { _ in
                 completion((channelDTO, messageDTO))
             }
         }
-
-        XCTAssertTrue(channelDTO.pinnedMessages.contains(messageDTO))
+        
+        XCTAssertTrue(
+            channelDTO.inContext(database.viewContext).pinnedMessages
+                .contains(messageDTO.inContext(database.viewContext))
+        )
     }
 
     func test_messagePayload_isNotPinned_removedFromPinnedMessages() throws {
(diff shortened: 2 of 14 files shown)


nuno-vieira (approved):
LGTM! ✅ Thanks, didn't realise this flag wasn't ON already, it is very useful 👍

Sources/StreamChat/Controllers/MessageController/MessageController_Tests.swift
```diff
@@ -386,12 +386,16 @@ final class MessageController_Tests: StressTestCase {
...
-            delegate.didChangeReplies_changes,
-            [.insert((replyDTO?.asModel())!, index: [0, 0])]
-        )
+        context.performAndWait {
+            // Assert `insert` entity change is received by the delegate
+            AssertAsync.willBeEqual(
+                delegate.didChangeReplies_changes,
+                [.insert((replyDTO?.asModel())!, index: [0, 0])]
+            )
+        }
```
VojtaStavik, line 398: Doing the asserts inside `performAndWait` is a little bit odd. Since we only compare the model object here, anyway, can't we just get it from the `viewContext`?

Sources/StreamChat/Database/DTOs/MessageDTO_Tests.swift
```diff
@@ -274,7 +274,11 @@ class MessageDTO_Tests: XCTestCase {
             }
         }
 
-        XCTAssertTrue(channelDTO.pinnedMessages.contains(messageDTO))
+        let context = try XCTUnwrap(channelDTO.managedObjectContext)
+        
+        context.performAndWait {
+            XCTAssertTrue(channelDTO.pinnedMessages.contains(messageDTO))
+        }
```
VojtaStavik, line 281: Same here and other cases like this...

Sources/StreamChat/Database/DatabaseContainer.swift
```diff
@@ -41,7 +41,9 @@ class DatabaseContainer: NSPersistentContainer {
         let context = newBackgroundContext()
         context.automaticallyMergesChangesFromParent = true
         context.mergePolicy = NSMergeByPropertyObjectTrumpMergePolicy
-        context.localCachingSettings = localCachingSettings
+        context.performAndWait { [localCachingSettings] in
+            context.localCachingSettings = localCachingSettings
+        }
```
VojtaStavik, line 46: I'd say just `perform` is OK here, no need to block the initialization for longer than needed.

Sources/StreamChat/Database/DatabaseContainer_Tests.swift
```diff
@@ -216,8 +220,14 @@ class DatabaseContainer_Tests: StressTestCase {
...
-        XCTAssertEqual(database.writableContext.localCachingSettings, cachingSettings)
-        XCTAssertEqual(database.backgroundReadOnlyContext.localCachingSettings, cachingSettings)
+        
+        database.writableContext.performAndWait {
+            XCTAssertEqual(database.writableContext.localCachingSettings, cachingSettings)
+        }
+        
+        database.backgroundReadOnlyContext.performAndWait {
+            XCTAssertEqual(database.backgroundReadOnlyContext.localCachingSettings, cachingSettings)
+        }
```
VojtaStavik, line 230: Here it makes sense to do it like this 👍

VojtaStavik (approved):
Looks great 👍 Just some minor changes and ergonomy tweaks.

Sources/StreamChat/Controllers/MessageController/MessageController_Tests.swift
olejnjak: That could be an option or maybe just inside `performAndWait` I can convert `replyDTO` to model and then compare it on main thread?

VojtaStavik: @olejnjak I believe both should work 👍

PR #1028, opened 2021-04-28
[CIS-811] New message list infrastructure

evsaev:
**This PR** sets up the infrastructure for the new `ChatMessageContentView`:
- updates `СhatMessageCollectionViewCell` reusing logic
- updates the `ChatMessageList/ChatThread` controllers
- replaces old message-content related UI components with the new ones 

_Note: The PR also temporarily remove `build-docs-snippets/build-apps` jobs from `Smoke Checks` workflow since there will be more breaking changes soon and the demo apps will be fixed when the message-list will be in its final shape._


SlackClone/SlackChatMessageContentView.swift
```diff
@@ -7,242 +7,27 @@ import StreamChatUI
...
-            // link preview is visible only when no other attachments available
-            options.insert(.linkPreview)
-        }
-        
-        return options
+extension ContainerStackView {
+    func changeOrdering() {
+        let subviews = self.subviews
+        removeAllArrangedSubviews()
+        addArrangedSubviews(subviews.reversed())
```
evsaev, line 31: That feels handy, how about making it part of `ContainerStackView` public API? 😄

VojtaStavik: @evsaev I'm OK with moving to SDK but I'd keep it internal for now :)

SlackClone/SlackMessageOptionsResolver.swift
```diff
@@ -0,0 +1,36 @@
...
+import UIKit
+
+final class SlackMessageOptionsResolver: ChatMessageLayoutOptionsResolver {
+    override func optionsForMessage(
+        at indexPath: IndexPath,
+        in channel: ChatChannel,
+        with messages: AnyRandomAccessCollection<ChatMessage>
+    ) -> ChatMessageLayoutOptions {
+        var options = super.optionsForMessage(at: indexPath, in: channel, with: messages)
+        options.remove([.flipped, .bubble, .metadata, .avatar, .avatarSizePadding, .authorName, .threadInfo, .reactions])
```
evsaev, line 16: This is one of the demo apps so we don't care about it ATM, however, I think this brings up an interesting topic:

With such dramatic changes in the default options, I'd rather provide yet another implementation instead of updating the `super` one. 

But that's not possible currently because of `ChatMessage+Extensions` being internal 🤔

VojtaStavik (approved):
🎉 😎

VojtaStavik:
@evsaev can you please disable the failing CI jobs for now, until we fix the issues?

PR #1029, opened 2021-04-28
Fix hiding issue when view switches container + Composer View Layout Polish

nuno-vieira:
# In this PR

- Fixes issue when a view that was already hidden would switch container, and remain with the size 0 and alpha = 0. This would cause a problem because the view would always have the size 0 and it wouldn't be possible to hide it again, because it already had alpha = 0.
- Fixes an issue on composer view that was causing the composer to layout twice
- Removes the state icon and polishes the composer layout according to Figma

With this PR, now the iMessage clone works as expected ✅

### DIFF
--- a/Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerVC.swift
+++ b/Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerVC.swift
@@ -117,7 +117,7 @@ open class _ChatMessageComposerVC<ExtraData: ExtraDataTypes>: _ViewController,
                 self.composerView.sendButton.isHidden = false
                 self.composerView.editButton.isHidden = true
                 self.composerView.messageQuoteView.isHidden = true
-                self.composerView.topContainer.isHidden = true
+                self.composerView.headerView.isHidden = true
             }
             composerView.messageInputView.setSlashCommandViews(hidden: true)
         case let .slashCommand(command):
@@ -129,7 +129,7 @@ open class _ChatMessageComposerVC<ExtraData: ExtraDataTypes>: _ViewController,
         case let .quote(messageToQuote):
             composerView.titleLabel.text = L10n.Composer.Title.reply
             Animate {
-                self.composerView.topContainer.isHidden = false
+                self.composerView.headerView.isHidden = false
                 self.composerView.messageQuoteView.isHidden = false
                 self.composerView.messageInputView.commandLabel.isHidden = true
             }
@@ -139,7 +139,7 @@ open class _ChatMessageComposerVC<ExtraData: ExtraDataTypes>: _ViewController,
             Animate {
                 self.composerView.editButton.isHidden = false
                 self.composerView.sendButton.isHidden = true
-                self.composerView.topContainer.isHidden = false
+                self.composerView.headerView.isHidden = false
                 self.composerView.messageInputView.commandLabel.isHidden = true
             }
             inputTextView.text = message.text
--- a/Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerVC.swift
+++ b/Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerVC.swift
@@ -128,9 +128,6 @@ open class _ChatMessageComposerVC<ExtraData: ExtraDataTypes>: _ViewController,
             dismissSuggestionsViewController()
         case let .quote(messageToQuote):
             composerView.titleLabel.text = L10n.Composer.Title.reply
-            let image = appearance.images.messageComposerReplyButton
-                .tinted(with: appearance.colorPalette.inactiveTint)
-            composerView.stateIcon.image = image
             Animate {
                 self.composerView.topContainer.isHidden = false
                 self.composerView.messageQuoteView.isHidden = false
@@ -139,9 +136,6 @@ open class _ChatMessageComposerVC<ExtraData: ExtraDataTypes>: _ViewController,
             composerView.messageQuoteView.content = .init(message: messageToQuote, avatarAlignment: .left)
         case let .edit(message):
             composerView.titleLabel.text = L10n.Composer.Title.edit
-            let image = appearance.images.messageComposerEditMessage
-                .tinted(with: appearance.colorPalette.inactiveTint)
-            composerView.stateIcon.image = image
             Animate {
                 self.composerView.editButton.isHidden = false
                 self.composerView.sendButton.isHidden = true
--- a/Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerVC.swift
+++ b/Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerVC.swift
@@ -97,6 +97,12 @@ open class _ChatMessageComposerVC<ExtraData: ExtraDataTypes>: _ViewController,
         setupInputView()
     }
 
+    override open func setUpLayout() {
+        super.setUpLayout()
+
+        view.embed(composerView)
+    }
+
     override open func updateContent() {
         super.updateContent()
         switch state {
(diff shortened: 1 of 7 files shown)


Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerView.swift
```diff
@@ -11,7 +11,7 @@ open class _ChatMessageComposerView<ExtraData: ExtraDataTypes>: _View, ThemeProv
     public private(set) lazy var container = ContainerStackView()
         .withoutAutoresizingMaskConstraints
 
-    public private(set) lazy var topContainer = ContainerStackView()
+    public private(set) lazy var topContainer = UIView()
         .withoutAutoresizingMaskConstraints
```
VojtaStavik, line 15: Why this change?

Sources/StreamChatUI/CommonViews/ChatInputView/ChatMessageInputView.swift
```diff
@@ -36,10 +36,9 @@ open class _ChatMessageInputView<ExtraData: ExtraDataTypes>: _View, ComponentsPr
     }
     
     override open func setUpLayout() {
-        embed(container)
+        embed(container, insets: .init(top: 1, leading: 4, bottom: 1, trailing: 0))
```
VojtaStavik, line 39: The preferred way of doing this is to pin the view to the layout margins and change the margins. It makes it easier to change the layout just by changing the margins constants rather than adding new constraints.

In fact, the `inset` parameter in `embed` should be probably removed because it doesn't have a good use case.

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerVC.swift
```diff
@@ -128,9 +128,6 @@ open class _ChatMessageComposerVC<ExtraData: ExtraDataTypes>: _ViewController,
             dismissSuggestionsViewController()
         case let .quote(messageToQuote):
             composerView.titleLabel.text = L10n.Composer.Title.reply
-            let image = appearance.images.messageComposerReplyButton
```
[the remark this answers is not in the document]
nuno-vieira: Ah yes that makes sense 👍 Will remove it

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerView.swift
nuno-vieira: image

Since there's no more left button on the top, it just easier to layout the views without the container. Or I would need to add an empty view to make the same layout. WDYT? Or is it the aim to always use the ContainerStackView? always?

Sources/StreamChatUI/CommonViews/ChatInputView/ChatMessageInputView.swift
nuno-vieira: ✅

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerVC.swift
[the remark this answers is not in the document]
nuno-vieira: ✅

Sources/StreamChatUI/CommonViews/ChatInputView/ChatMessageInputView.swift
nuno-vieira: I did not remove that functionality from the `embed` function btw, because it is being used in a lot of places, and it breaks the UI in multiple places. So I think it should be a ticket on its own. It would be nice to have the same API, but change the `embed` function to add the constraints to the layout margins, and the inset should be applied to the superview.layoutMargins instead

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerView.swift
```diff
@@ -173,8 +169,16 @@ open class _ChatMessageComposerView<ExtraData: ExtraDataTypes>: _View, ThemeProv
         centerLeftContainer.addArrangedSubview(attachmentButton)
         centerLeftContainer.addArrangedSubview(commandsButton)
         centerLeftContainer.addArrangedSubview(shrinkInputButton)
+
+        dismissButton.widthAnchor.pin(equalToConstant: 24).isActive = true
+        dismissButton.heightAnchor.pin(equalToConstant: 24).isActive = true
```
DominikBucher12, line 174: nit: `dismissButton.heightAnchor.pin(equalTo: dismissButton.heightAnchor).isActive = true`

Sources/StreamChatUI/CommonViews/ChatInputView/ChatMessageInputView.swift
```diff
@@ -36,10 +36,19 @@ open class _ChatMessageInputView<ExtraData: ExtraDataTypes>: _View, ComponentsPr
     }
     
     override open func setUpLayout() {
-        embed(container)
+        layoutMargins.left = 8
+        layoutMargins.top = 1
+        layoutMargins.right = 8
+        layoutMargins.bottom = 1
```
DominikBucher12, line 42: Maybe add comment why this is needed? :)

 I feel like from documentation of `layoutMargins`, you might be wanting to use `directionalLayoutMargins` instead... Also it says that they have the default value of 8 by default, maybe wanna give it a try? 

https://developer.apple.com/documentation/uikit/uiview/1622566-layoutmargins

DominikBucher12 (approved):
I have some nits mainly regarding the layoutMargins...  Otherwise this looks good to me :) Thanks for taking care of the Composer <3

Sources/StreamChatUI/CommonViews/ContainerStackView.swift
```diff
@@ -203,6 +203,14 @@ public class ContainerStackView: UIView {
         assert(subviews.contains(subview))
         subview.removeFromSuperview()
 
+        // Reset hiding state from the removed view. This important especially if the view
+        // is added to another container, since the view will have the size to 0 and the other
+        // container won't be able to hide the view because it thinks it is already hidden.
+        subview.alpha = 1.0
+        hideConstraintsByView[subview]?.isActive = false
```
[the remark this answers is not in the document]
VojtaStavik: @DominikBucher12 I don't think the constraint lifetime is defined by holding a reference to it 🤔 The `isActive` is what matters.

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerVC.swift
```diff
@@ -97,6 +97,12 @@ open class _ChatMessageComposerVC<ExtraData: ExtraDataTypes>: _ViewController,
         setupInputView()
     }
 
+    override open func setUpLayout() {
+        super.setUpLayout()
+
+        view.embed(composerView)
```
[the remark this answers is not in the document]
VojtaStavik: I don' think this established this as a rule but I personally don't like the `embed` helper and prefer using:
```
addSubview(view)
view.pin(to: self.layoutMarginsGuide)
```
since it's more explicit.

VojtaStavik (approved):
LGTM 👍 @DominikBucher12 has some good comments there

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerView.swift
nuno-vieira: I don't like that much because you have to type more code than just doing what I did, IMO it looks cleaner than the code above, and the benefits of that are really not worth it

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerVC.swift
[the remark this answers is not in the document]
nuno-vieira: Ah yes that is way better

Sources/StreamChatUI/CommonViews/ContainerStackView.swift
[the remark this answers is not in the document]
nuno-vieira: Yes exactly, the `isActive` is what is important here, the rest is just cleaning some memory

PR #1034, opened 2021-04-29
Fix image picker crash in production

LukasHromadnik:
# Description

In production code we had a typo in string constant that determines which media types are available for the user to pick from.


VojtaStavik (approved):
LGTM, thank you 🤦‍♂️ Feel free to also add changelog entry for this

PR #1036, opened 2021-04-29
Rename and Move Composer View Components

nuno-vieira:
# In this PR
Renames and moves all the current components of the Composer View.
Does not include UIConfig changes, that will happen in a follow-up PR.
The focus is only on moving and renaming things, most likely the components will be changed and totally refactored in the next PRs.
Replaces this PR: https://github.com/GetStream/stream-chat-swift/pull/987

Moved Views To CommonViews folder (Probably worth it to rename the folder to just Components?):


StreamChat.xcodeproj/project.pbxproj
```diff
@@ -4031,6 +4034,38 @@
...
+			isa = PBXGroup;
+			children = (
+				22C23505259A257B00DC805A /* ChatDocumentAttachmentsCollectionView.swift */,
+				22086B37259509350007F8C0 /* ChatDocumentAttachmentsCollectionViewCell.swift */,
+				22086B47259509AC0007F8C0 /* ChatDocumentAttachmentsCollectionViewLayout.swift */,
+				22086B3F259509450007F8C0 /* ChatDocumentAttachmentView.swift */,
+			);
+			path = DocumentAttachments;
+			sourceTree = "<group>";
+		};
```
nuno-vieira, line 4069: This makes me wonder if we move the attachments views of the composer to the common folder, where will the attachments view of the message list live? 🤔 CC @VojtaStavik @evsaev 

We kinda need to come up with something to distinguish them. If we have `ChatImageAttachmentView` for the composer image attachments item view, what will be the name of the attachment view in the message list?

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerView.swift
```diff
@@ -48,31 +48,22 @@ open class _ChatMessageComposerView<ExtraData: ExtraDataTypes>: _View, ThemeProv
...
-    public private(set) lazy var shrinkInputButton: UIButton = components
-        .messageComposer
-        .composerButton.init()
+    public private(set) lazy var shrinkInputButton: UIButton = UIButton()
         .withoutAutoresizingMaskConstraints
     
-    public private(set) lazy var dismissButton: UIButton = components
-        .messageComposer
-        .composerButton.init()
+    public private(set) lazy var dismissButton: UIButton = UIButton()
```
nuno-vieira: My idea is to refactor this and extract all these other buttons to their own button components as we have with `SendButton` and `ConfirmButton`, and probably have `Components.buttons` just like we have with `Components.images` and `Components.fonts`, WDYT?

Sources/StreamChatUI/CommonViews/Suggestions/ChatCommandSuggestionView/ChatCommandSuggestionCollectionViewCell.swift
```diff
@@ -0,0 +1,25 @@
...
+//
+
+import StreamChat
+import UIKit
+
+/// A view cell that displays a command.
+public typealias ChatCommandSuggestionCollectionViewCell = _ChatCommandSuggestionCollectionViewCell<NoExtraData>
+
+/// A view cell that displays a command.
+open class _ChatCommandSuggestionCollectionViewCell<ExtraData: ExtraDataTypes>: _CollectionViewCell, ComponentsProvider {
```
VojtaStavik, line 12: Does this need to be generic?...

Sources/StreamChatUI/CommonViews/CheckboxControl/ChatCheckboxControl.swift
```diff
@@ -5,7 +5,8 @@
 import StreamChat
 import UIKit
 
-open class ChatMessageComposerCheckmarkControl: _Control, AppearanceProvider {
+/// A view to check/uncheck an option.
+open class ChatCheckboxControl: _Control, AppearanceProvider {
```
VojtaStavik, line 9: I don't event think we need the `Chat` prefix here

Sources/StreamChatUI/Components.swift
```diff
@@ -37,6 +37,11 @@ public struct _Components<ExtraData: ExtraDataTypes> {
 
     /// Button used for sending a message, or any type of content.
     public var sendButton: UIButton.Type = ChatSendButton.self
+
+    /// A view to check/uncheck an option.
+    public var checkmarkControl: ChatCheckboxControl.Type =
+        ChatCheckboxControl.self
+
```
VojtaStavik, line 47: The type here can be just `UIControl`?

VojtaStavik (approved):
LGTM, I just left some random comments, didn't do a proper review since it's going to change, anyway.

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerView.swift
VojtaStavik: @nuno-vieira I don't we need `Components.buttons`. I kinda like the "keep-it-flat" rule :) It's simple and you don't have to think about it. It's also easier to see access the components via autocompletion like that.

Sources/StreamChatUI/CommonViews/Suggestions/ChatCommandSuggestionView/ChatCommandSuggestionCollectionViewCell.swift
nuno-vieira: It is because it needs the `components` to get the `ChatCommangSuggestionView`

Sources/StreamChatUI/CommonViews/CheckboxControl/ChatCheckboxControl.swift
nuno-vieira: We need to write down the pattern when to use the `Chat` prefix. As you can see below, right now, all the components have the `Chat` prefix, only `TypingView` doesn't have it. For me, or we either write the prefix in all views, or in the UI SDK, we don't use the prefix at all, because if we use it in some and other's we don't, will be confusing.
image

Sources/StreamChatUI/Components.swift
nuno-vieira: Yes you are right 👍

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerView.swift
nuno-vieira: Okay, my only concern is that this will be a flat list of maybe 50+ components. The autocompletion will be hard to find what you need. So the idea that I was having was to replicate the structure of the design system. We have images, icons, fonts, buttons, inputs, etc... But lets make it flat for now, and later if we think is useful, we can change it 👍

Sources/StreamChatUI/Components.swift
nuno-vieira: Actually, no @VojtaStavik, because the ChatCheckboxControl has a label that we need to access it, and that it does not exist on UIControl

PR #1038, opened 2021-04-30
[CIS-818] Implement `GalleryAttachmentViewInjector`

VojtaStavik:
This PR introduces a new more flexible way to inject attachment views to the content view. 

The goal was to introduce a unified way for both the built-in and user custom attachments are handled. This is just a proof of concept implementation of the gallery attachment.


evsaev (approved):
Looks good to me 👍 I like how we hook base content view's lifecycle methods. 

2 questions: 
- shouldn't we add the `messageContentView: _ChatMessageContentView<ExtraData>` to components to make it customisable for end-users? (and also return `components. messageContentView` in controllers by default)
- can we maybe have an array of injectors? 🤔That would help a lot if multiple attachments should be shown for a message

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -201,6 +201,9 @@ open class _ChatMessageListVC<ExtraData: ExtraDataTypes>:
         if indexPath.row + 1 >= collectionView.numberOfItems(inSection: 0) {
             channelController.loadPreviousMessages()
         }
+
+        // Force cell layout to prevent animation glitches
+        cell.layoutIfNeeded()
```
b-onc, line 206: Don't we need `setNeedsLayout` instead of `layoutIfNeeded`?

VojtaStavik: Right, fixed in 176d54b

PR #1039, opened 2021-04-30
[CIS-708] Implement DemoApp publishing to TestFlight

dmigach:
This PR includes a fastlane lane for publishing as well as GitHub actions workflow.
Also added APPSTORE_API_KEY secret so now GitHub actions can authorise in appstoreconnect if necessary

### DIFF
--- a/.github/workflows/DemoAppTestFlight.yml
+++ b/.github/workflows/DemoAppTestFlight.yml
@@ -1,0 +1,46 @@
+name: Test Flight Deploy DemoApp
+
+on:
+  release:
+    types: [published]
+
+  workflow_dispatch:
+  
+env:
+  HOMEBREW_NO_INSTALL_CLEANUP=1: 1
+
+jobs:
+  deploy:
+    runs-on: macos-latest
+
+    steps:
+    - name: Install Bot SSH Key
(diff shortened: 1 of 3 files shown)


.github/workflows/DemoAppTestFlight.yml
```diff
@@ -0,0 +1,46 @@
+name: Test Flight Deploy DemoApp
+
+on:
+  push:
+    branches: [ main ]
```
VojtaStavik, line 5: This should ideally happen only on release not push

VojtaStavik (approved):
LGTM 🚀  just one detail. Thank you!

.github/workflows/DemoAppTestFlight.yml
dmigach: Right, thank you for the reminder! Changed it to be triggered on release, next time we release we'll see if it works 🙂
It should since I took a trigger from here https://docs.github.com/en/actions/reference/events-that-trigger-workflows#release but who knows

DominikBucher12 (approved):
This Looks good to me 🎊

VojtaStavik:
@b-onc can you please take a look at this?

PR #1044, opened 2021-05-03
[CIS-828] Investigate reconnection strategy tests failure

DominikBucher12:
# What this PR do:
- FIxes `WebSocketClient_Tests` which failed on CI due to possibly bad configuration policy (using `default`, not `ehemeral`
# How to test it
- Check if the tests for this PR are green, if yes, this solution works

# Explanation of the fix:

The `WebSocketClient` was initialized in tests with `default` `URLSessionConfiguration`, which stores all the caches and other session-related data on the disk. It is possible that due to unknown nature of the CI machine where github actions run tests those cached data were still persistent through all tests during the Test run. What I did is basically change the configuration to `ephemeral` which doesn't cache and any of the related data to disk and should resolve the problem.  

You can read more about the `URLSessionConfiguration` in here:
https://developer.apple.com/documentation/foundation/urlsessionconfiguration/1410529-ephemeral

### DIFF
--- a/Sources/StreamChat/APIClient/APIClient_Mock.swift
+++ b/Sources/StreamChat/APIClient/APIClient_Mock.swift
@@ -70,7 +70,7 @@ class APIClientMock: APIClient {
 extension APIClientMock {
     convenience init() {
         self.init(
-            sessionConfiguration: .default,
+            sessionConfiguration: .ephemeral,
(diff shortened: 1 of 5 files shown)


VojtaStavik (approved):
Interesting! Let's see how it behaves. Thank you for exploring this 👍

evsaev (approved):
The change LGTM 👍

PR #1045, opened 2021-05-03
Use separate `channelsCount` for `ChatChannelListVC`

olejnjak:
Use separate `channelsCount` for channels in `ChatChannelListVC` so it can be updated inside `performBatchUpdates` as documented.

### DIFF
--- a/Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
+++ b/Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
@@ -58,11 +58,15 @@ open class _ChatChannelListVC<ExtraData: ExtraDataTypes>: _ViewController,
     
     /// Reuse identifier of `collectionViewCell`
     open var collectionViewCellReuseIdentifier: String { "Cell" }
+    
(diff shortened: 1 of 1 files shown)


Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
```diff
@@ -251,6 +255,8 @@ extension _ChatChannelListVC: _ChatChannelListControllerDelegate {
                         collectionView.reloadItems(at: [index])
                     }
                 }
+                
+                channelsCount = controller.channels.count
```
VojtaStavik, line 259: I think the `count` value should be obtained outside of the closure. Also, do we need to somehow handle the edge case where `controller.channel.count` < `channelsCount`?

olejnjak: > I think the `count` value should be obtained outside of the closure.

I think it should not be necessary as the closure is non-escaping so performed immediately and synchronously, but just to be sure we can do that.

> Also, do we need to somehow handle the edge case where `controller.channel.count` < `channelsCount`?

I don't see any reason why, do you?

VojtaStavik (approved):
OK, let's see if it helps 🤞

PR #1047, opened 2021-05-03
Fix invalid payload in TypingEvent causing assertionFailure

b-onc:
...in EventDataProcessorMiddleware.

CHANGELOG will be added to release branches after cherry pick.

Should we add a test where we check a payload is always valid? I don't think it's needed now, but Stream developers should be aware that we cannot treat `Any` type carelessly.


Sources/StreamChat/WebSocketClient/Events/TypingEvent.swift
```diff
@@ -20,10 +20,15 @@ public struct TypingEvent: UserSpecificEvent, ChannelSpecificEvent {
...
-    init(isTyping: Bool, cid: ChannelId, userId: UserId, payload: Any = 0) {
+    init(isTyping: Bool, cid: ChannelId, userId: UserId) {
         self.isTyping = isTyping
         self.cid = cid
         self.userId = userId
-        self.payload = payload
+        // This `payload` will be saved in `saveEvent` because of `EventDataProcessorMiddleware`
+        // and it's required, so we put a "phony" payload with just `eventType`
+        // so there's nothing actually to be saved
+        payload = EventPayload<NoExtraData>(
```
DominikBucher12, line 30: 👍

DominikBucher12 (approved):
This looks okay to me... Question is if `TypingEvent` needs to be `EventWithPayload` when we don't actually need the payload because everything we need is just `isTyping`, `userId` and `cid`... But that requires lot of work because of `UserSpecificEvent` and `ChannelSpecificEvent` and is maybe overthinking...

VojtaStavik (approved):
Nice 🎉 🚀

PR #1048, opened 2021-05-03
Re-architecture of `ChatMessageComposerView`'s child components

nuno-vieira:
## In this PR
- `ChatMessageInputView` Final Form: Now every input content related views are part of the `ChatMessageInputView`.
- Flattens the icons of the Appearance Config related to the composer view
- Creates button components that are part of Composer View
- Adds Documentation to the Composer View
- Renames some of the composer view components + General Cleanup

## Not in this PR
Currently, the new data flow is still missing. Each view will have its `Content` and so it will thin a lot the `ComposerVC` which right now is responsible to populate and know the implementation details on how to populate each of its child components.


Sources/StreamChatUI/CommonViews/CloseButton/ChatCloseButton.swift
```diff
@@ -0,0 +1,16 @@
+//
+// Copyright © 2021 Stream.io Inc. All rights reserved.
+//
+
+import StreamChat
+import UIKit
+
+/// Button for closing, dismissing or clearing information.
+open class ChatCloseButton: _Button, AppearanceProvider {
```
nuno-vieira, line 9: Something that is not clear yet for me is: Should all our views have `Chat` prefix, or only the ones that use the LLC models, example `ChatMessageInputView` ? If this is the case, then all these button components prefixes should be removed

MessengerClone/ChatClient.swift
```diff
@@ -14,7 +14,7 @@ extension ChatClient {
         var components = Components()
         
         components.navigation.channelListRouter = MessengerChatChannelListRouter.self
-        appearance.images.newChat = UIImage(systemName: "square.and.pencil")!
+        appearance.images.pencil = UIImage(systemName: "square.and.pencil")!
```
nuno-vieira, line 17: At first, I've changed the image names without realising that they can be replaced and so describing they're use is a better approach. Let me know if it is better here: https://github.com/GetStream/stream-chat-swift/pull/1048/commits/2fd3fc658ed622ec44cc9eb85af85eefb7d361d0

iMessageClone/iMessageChatMessageComposerView.swift
```diff
@@ -10,15 +10,15 @@ final class iMessageChatMessageComposerView: ChatMessageComposerView {
...
-        centerContentContainer.distribution = .natural
-        centerContentContainer.alignment = .fill
-        centerContentContainer.subviews.forEach {
-            centerContentContainer.removeArrangedSubview($0)
+        centerInputContainer.axis = .horizontal
+        centerInputContainer.distribution = .natural
+        centerInputContainer.alignment = .fill
+        centerInputContainer.subviews.forEach {
+            centerInputContainer.removeArrangedSubview($0)
         }
```
VojtaStavik, line 22: We have `removeAllArrangedSubviews()` method on `ContainerStackView` (it was you who added it 😅 😁 )

Sources/StreamChatUI/CommonViews/CloseButton/ChatCloseButton.swift
VojtaStavik: @nuno-vieira I don't think we need `Chat` prefix for all components. I think the _default_ should be no prefix. The `Chat` prefix should be used only for components where it's expected users will interact with them a lot and they have a too generic name. `User` -> `ChatUser` for example.

Sources/StreamChatUI/CommonViews/MessageInputView/ChatMessageInputView.swift
```diff
@@ -24,44 +45,69 @@ open class _ChatMessageInputView<ExtraData: ExtraDataTypes>: _View, ComponentsPr
...
         addSubview(container)
         container.pin(to: layoutMarginsGuide)
-        directionalLayoutMargins.leading = 8
-        directionalLayoutMargins.top = 1
-        directionalLayoutMargins.trailing = 8
-        directionalLayoutMargins.bottom = 1
+        directionalLayoutMargins.leading = 0
+        directionalLayoutMargins.top = 0
+        directionalLayoutMargins.trailing = 0
+        directionalLayoutMargins.bottom = 0
```
VojtaStavik, line 72:
```swift
directionalLayoutMargins = .zero
```
? :)

Sources/StreamChatUI/CommonViews/MessageInputView/ChatMessageInputView.swift
```diff
@@ -24,44 +45,69 @@ open class _ChatMessageInputView<ExtraData: ExtraDataTypes>: _View, ComponentsPr
...
 
-        container.preservesSuperviewLayoutMargins = false
-        container.alignment = .center
-        container.spacing = 4
+        container.isLayoutMarginsRelativeArrangement = true
+        container.directionalLayoutMargins.leading = 4
+        container.directionalLayoutMargins.top = 0
+        container.directionalLayoutMargins.trailing = 4
+        container.directionalLayoutMargins.bottom = 0
```
VojtaStavik, line 78:
```swift
container.directionalLayoutMargins = .init(top: 0, leading: 4, bottom: 0, trailing: 4)
```
Maybe this style is a bit better? Especially when you're changing all values there.

Sources/StreamChatUI/Appearance+Images.swift
```diff
@@ -12,21 +12,21 @@ public extension Appearance {
         public var loadingIndicator: UIImage = UIImage(named: "loading_indicator", in: .streamChatUI)!
         public var close: UIImage = UIImage(named: "close", in: .streamChatUI)!
         public var closeCircleTransparent: UIImage = UIImage(named: "close_circle_transparent", in: .streamChatUI)!
-        public var closeCircleFilled: UIImage = UIImage(named: "close_circle_filled", in: .streamChatUI)!
+        public var discardAttachment: UIImage = UIImage(named: "close_circle_filled", in: .streamChatUI)!
```
VojtaStavik: Yep, I like the context-based naming better 👍

VojtaStavik (approved):
Great to see the composer getting its final shape. I like the flattened version much better 👍 I left a couple of comments but mostly nits. As for the image names, I think the context-based naming works better for customization 👍

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerView.swift
```diff
@@ -42,34 +42,49 @@ open class _ChatMessageComposerView<ExtraData: ExtraDataTypes>: _View, ThemeProv
...
+    /// A button to dismiss the current state (quoting, editing, etc..).
+    public private(set) lazy var dismissButton: UIButton = components
+        .closeButton.init()
         .withoutAutoresizingMaskConstraints
-    
+
+    /// A label part of the header view to display the current state (quoting, editing, etc..).
     public private(set) lazy var titleLabel: UILabel = UILabel()
         .withoutAutoresizingMaskConstraints
         .withBidirectionalLanguagesSupport
```
evsaev, line 97: `.withAdjustingFontForContentSizeCategory`

MessengerClone/ChatClient.swift
skorepak: I prefer the use description rather than look description. Clients can (and will) customise them a lot

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerView.swift
```diff
@@ -5,71 +5,99 @@
...
+/// High level overview of the composer layout:
+/// ```
+/// |---------------------------------------------------------|
+/// |                       headerView                        |
+/// |---------------------------------------------------------|--|
+/// | leadingContainer | messageInputView | trailingContainer |  | = centerContainer
+/// |---------------------------------------------------------|--|
+/// |                     bottomContainer                     |
+/// |---------------------------------------------------------|
+/// ```
```
skorepak, line 19: ❤️

Sources/StreamChatUI/CommonViews/CloseButton/ChatCloseButton.swift
skorepak: I think we should treat all the components the same, we don't know what the users will or won't use. As long as they can import only a few specific views, the Stream views in the client's codebase should be easily recognisable.

skorepak (commented):
Just a few comments. My bit to the discussion :)

Sources/StreamChatUI/CommonViews/CloseButton/ChatCloseButton.swift
nuno-vieira: @VojtaStavik Jakub has a good point, if we prefix all views with `Chat`, I can see 3 advantages: 
- The views are easily recognizable as a view from the SDK
- The view names are consistent
- When naming a new view, you don't need to think about if it needs the `Chat` prefix or not

iMessageClone/iMessageChatMessageComposerView.swift
nuno-vieira: Yes this is old code, when I introduced that new method, forgot to change it here 😅 These are changes done by the name refactoring on Xcode

Sources/StreamChatUI/CommonViews/MessageInputView/ChatMessageInputView.swift
nuno-vieira: Done ✅

Sources/StreamChatUI/CommonViews/MessageInputView/ChatMessageInputView.swift
nuno-vieira: Done ✅

Sources/StreamChatUI/Appearance+Images.swift
nuno-vieira: Done ✅

Sources/StreamChatUI/ChatMessageList/MessageComposer/ChatMessageComposerView.swift
nuno-vieira: Fixed it here: https://github.com/GetStream/stream-chat-swift/pull/1048/commits/acbd51a33a87f6f5ca3f21bcdf4117542c1f0de3 ✅

iMessageClone/iMessageChatMessageComposerView.swift
```diff
@@ -10,49 +10,16 @@ final class iMessageChatMessageComposerView: ChatMessageComposerView {
...
-        ])
-
-        centerInputContainer.addArrangedSubview(newStack)
-        centerInputContainer.addArrangedSubview(sendButtonContainer)
-
-        NSLayoutConstraint.activate([
-            sendButton.widthAnchor.constraint(equalToConstant: 30),
-            sendButton.heightAnchor.constraint(equalToConstant: 35)
-        ])
+        messageInputView.inputTextContainer.addArrangedSubview(sendButton)
```
nuno-vieira: 4 lines of code to customize the iMessageClone now, instead of 40 😄 🎉

PR #1049, opened 2021-05-04
[CIS-812] Implement smart channel list clean up when /sync fails

dmigach:
When the sync endpoint fails to return missing events, we need to delete some of the local data to prevent having DB in the undefined state


Sources/StreamChat/Workers/ChannelListCleanupUpdater.swift
```diff
@@ -0,0 +1,93 @@
...
+    }
+    
+    func cleanupChannelList() {
+        database.write { _ in
+            self.allChannels.forEach { $0.syncFailedCleanUp() }
+        } completion: { error in
+            if let error = error {
+                log.error("Failed cleaning up channels data: \(error).")
+            }
+            self.updateChannels()
```
dmigach, line 44: I thought it makes sense to hold self with strong reference here so we won't end up in some half-cleaned undefined state

Sources/StreamChat/Workers/ChannelListCleanupUpdater.swift
```diff
@@ -0,0 +1,93 @@
...
+            do {
+                return try ChannelId(cid: $0)
+            } catch {
+                log.error("Failed to decode `ChannelId` from \($0).")
+                return nil
+            }
+        }
+        let channelListQuery = _ChannelListQuery<ExtraData.Channel>(
+            filter: .in(.cid, values: channelIds),
+            pageSize: allChannels.count
```
dmigach, line 61: Not sure about this

Sources/StreamChat/Workers/Background/MissingEventsPublisher.swift
```diff
@@ -86,6 +103,10 @@ class MissingEventsPublisher<ExtraData: ExtraDataTypes>: EventWorker {
                     self?.eventNotificationCenter.process(payload.eventPayloads)
                 case let .failure(error):
                     log.error("Internal error: Failed to fetch and reply missing events: \(error)")
+                    if let error = (error as? ClientError)?.underlyingError as? ErrorPayload,
+                       error.statusCode == 400 {
```
VojtaStavik, line 107: nit: It would be nice to have this logic encapsulated somehow: `if error.isSyncFailedError { ... }`

Sources/StreamChat/Workers/ChannelListCleanupUpdater.swift
```diff
@@ -0,0 +1,110 @@
+//
+// Copyright © 2021 Stream.io Inc. All rights reserved.
+//
+
+import CoreData
+
+class ChannelListCleanupUpdater<ExtraData: ExtraDataTypes>: Worker {
```
VojtaStavik, line 7: nit: please add documentation

Sources/StreamChat/Workers/ChannelListCleanupUpdater.swift
```diff
@@ -0,0 +1,110 @@
...
+        super.init(
+            database: database,
+            apiClient: apiClient
+        )
+    }
+    
+    func cleanupChannelList() {
+        database.write { _ in
+            self.allChannels.forEach { $0.syncFailedCleanUp() }
+        } completion: { error in
```
VojtaStavik, line 40: IMO the `self.allChannels` indirection is not necessary and dangerous since it can be called only from within the `write` block. I think inlining it would be better:
```
database.write { session in 
   let channels = (session as NSManagedObjectContext)?.fetch(ChannelDTO.allChannelsFetchRequest)
   channels.forEach { $0. syncFailedCleanUp() }
}
```

Sources/StreamChat/Workers/ChannelListCleanupUpdater.swift
```diff
@@ -0,0 +1,110 @@
...
+    
+    private func updateChannels() {
+        database.backgroundReadOnlyContext.perform {
+            self.queries.forEach {
+                self.channelListUpdater.update(channelListQuery: $0) { error in
+                    if let error = error {
+                        log.error("Internal error. Failed to update ChannelListQueries for the new channel: \(error)")
+                    }
+                }
+            }
```
VojtaStavik, line 58: Same here, `self.queries` is not the best way to handle this when it's possible to access it only from within the `backgroundReadOnlyContext` queue.

Sources/StreamChat/Workers/ChannelListCleanupUpdater_Tests.swift
```diff
@@ -0,0 +1,116 @@
...
+            AssertAsync.willBeTrue(channel.watchers.isEmpty)
+            AssertAsync.willBeTrue(channel.members.isEmpty)
+            AssertAsync.willBeTrue(channel.attachments.isEmpty)
+            AssertAsync.willBeTrue(channel.pinnedMessages.isEmpty)
+            AssertAsync.willBeTrue(channel.reads.isEmpty)
+            AssertAsync.willBeTrue(channel.queries.isEmpty)
+            AssertAsync.willBeNil(channel.oldestMessageAt)
+            AssertAsync.willBeNil(channel.hiddenAt)
+            AssertAsync.willBeNil(channel.truncatedAt)
+            AssertAsync.willBeFalse(channel.needsRefreshQueries)
```
VojtaStavik, line 84: Having `AssertAsync` on every line means that every line will be evaluated separately. If you do
```swift
AssertAsync { 
  Assert.willlBeTrue(channel.messages.isEmpty)
  Assert.willlBeTrue(channel.watchers.isEmpty)
  ...
}
```
all asserts will be evaluated at once -> much faster.

Sources/StreamChat/Workers/ChannelListCleanupUpdater_Tests.swift
```diff
@@ -0,0 +1,116 @@
...
+            withMessages: true,
+            withQuery: true,
+            hiddenAt: Date(timeIntervalSince1970: 56789),
+            needsRefreshQueries: false
+        )
+        
+        channelListCleanupUpdater?.cleanupChannelList()
+        
+        let channel1 = database.viewContext.channel(cid: cid1)!
+        let channel2 = database.viewContext.channel(cid: cid2)!
```
VojtaStavik, line 70: please use `XCTUnwrap` rather than `!`

Sources/StreamChat/Workers/ChannelListCleanupUpdater_Tests.swift
```diff
@@ -0,0 +1,116 @@
...
+    
+    func test_cleanupChannelList_cleansChannelsData() throws {
+        let cid1 = ChannelId.unique
+        let cid2 = ChannelId.unique
+        
+        try database.createChannel(
+            cid: cid1,
+            withMessages: true,
+            withQuery: true,
+            hiddenAt: Date(timeIntervalSince1970: 56789),
```
VojtaStavik, line 55: why not `.unique`?

Sources/StreamChat/Workers/ChannelListCleanupUpdater_Tests.swift
```diff
@@ -0,0 +1,116 @@
...
+            needsRefreshQueries: false
+        )
+        
+        channelListCleanupUpdater?.cleanupChannelList()
+        
+        let channel1 = database.viewContext.channel(cid: cid1)!
+        let channel2 = database.viewContext.channel(cid: cid2)!
+        
+        func checkChannelClearedOutProperly(channel: ChannelDTO) {
```
VojtaStavik, line 72: I'm not sure about this. If one of the asserts fails it won't be clear for which channel it fails. What if we did this:
```
extension ChannelDTO { 
  var isClearedOutProperly: Bool { 
     messages.isEmpty && members.isEmtpy ....
  }
}

...

AssertAsync { 
  Assert.willBeTrue(channel1. isClearedOutProperly)
  Assert.willBeTrue(channel2. isClearedOutProperly)
}
```

VojtaStavik (commented):
Looks good 👍 I have a couple of comments and suggestions

Sources/StreamChat/Workers/Background/MissingEventsPublisher.swift
dmigach: But it only means that in that context, so I'm not sure it makes sense to generalise it. We can do something like `has400statusCode`

Sources/StreamChat/Workers/ChannelListCleanupUpdater_Tests.swift
dmigach: Nice, was looking for something like this

Sources/StreamChat/Workers/ChannelListCleanupUpdater_Tests.swift
dmigach: Data in unit tests should be static unless it's the idea behind a stress test to run it with various random inputs

Sources/StreamChat/Utils/InternetConnection/Error+InternetNotAvailable.swift
```diff
@@ -24,4 +24,12 @@ extension Error {
...
         return error.domain == parameters.domain && error.code == parameters.errorCode
     }
+    
+    var isBackendErrorWith400StatusCode: Bool {
+        if let error = (self as? ClientError)?.underlyingError as? ErrorPayload,
+           error.statusCode == 400 {
+            return true
+        }
+        return false
+    }
```
VojtaStavik, line 34: > But it only means that in that context, so I'm not sure it makes sense to generalise it. We can do something like has400statusCode

I was actually thinking about making it local context only for `MissingEventsPublisher`. Even the fact that we're checking for `400` doesn't tell much. The difference I was aiming for was:
```swift
// The backend returns 400 if sync fails because of too many missing events
if error. isBackendErrorWith400StatusCode { ... } 

vs .

if error.isTooManyMissingEventsToSyncError { ... }
```
or something like that.


```

File: Sources/StreamChat/Workers/Background/MissingEventsPublisher.swift [hunk 12]
```diff
@@ -86,6 +103,9 @@ class MissingEventsPublisher<ExtraData: ExtraDataTypes>: EventWorker {
                     self?.eventNotificationCenter.process(payload.eventPayloads)
                 case let .failure(error):
                     log.error("Internal error: Failed to fetch and reply missing events: \(error)")
```
VojtaStavik, line 105: This is no longer a failuer. Can you please change this from `error` to `info` and also update the message?

Sources/StreamChat/Workers/ChannelListCleanupUpdater.swift
```diff
@@ -0,0 +1,108 @@
+//
+// Copyright © 2021 Stream.io Inc. All rights reserved.
+//
+
+import CoreData
+
+/// Cleans up local channels data and refetches it from backend
+class ChannelListCleanupUpdater<ExtraData: ExtraDataTypes>: Worker {
```
VojtaStavik, line 8: I'd really appreaciate if you spend more time on the documentation here + on the methods of the class 👍

VojtaStavik: Since this updater is touching `ChannelDTO` and it also reloads all existing queries, can we maybe call it something like `DatabaseCleanupUpdater`?

Sources/StreamChat/Workers/ChannelListCleanupUpdater.swift
```diff
@@ -0,0 +1,108 @@
...
+        channelListUpdater: ChannelListUpdater<ExtraData>
+    ) {
+        self.channelListUpdater = channelListUpdater
+        super.init(
+            database: database,
+            apiClient: apiClient
+        )
+    }
+    
+    func cleanupChannelList() {
```
VojtaStavik, line 38: I'm not sure we should combine resetting the `ChannelDTO` fields and the query re-fetching. What if we keep it as two separate methods?
```
func resetAllExistingChannelsData()
func refetchAllExistingChannelListQueries()
```
+ documentation of course :)

Sources/StreamChat/Workers/ChannelListCleanupUpdater.swift
```diff
@@ -0,0 +1,108 @@
...
+
+import CoreData
+
+/// Cleans up local channels data and refetches it from backend
+class ChannelListCleanupUpdater<ExtraData: ExtraDataTypes>: Worker {
+    // MARK: - Properties
+    
+    private let channelListUpdater: ChannelListUpdater<ExtraData>
+    
+    // MARK: - Init
```
VojtaStavik, line 13: nit: I don't think these things are needed for such a short files :)

Sources/StreamChat/Workers/ChannelListCleanupUpdater.swift
```diff
@@ -0,0 +1,108 @@
...
+            apiClient: apiClient
+        )
+    }
+    
+    func cleanupChannelList() {
+        database.write { session in
+            if let channels = try? (session as? NSManagedObjectContext)?
+                .fetch(ChannelDTO.allChannelsFetchRequest) {
+                channels.forEach { $0.syncFailedCleanUp() }
+            }
```
VojtaStavik, line 43: `write { }` accepts a throwing closure so there's no need for `try?` here

Sources/StreamChat/Workers/ChannelListCleanupUpdater.swift
```diff
@@ -0,0 +1,108 @@
...
+                }
+            } catch {
+                log.error("Internal error: Failed to fetch [ChannelListQueryDTO]: \(error)")
+            }
+        }
+    }
+}
+
+private extension ChannelDTO {
+    func syncFailedCleanUp() {
```
VojtaStavik, line 82: Documentation please

Sources/StreamChat/Workers/ChannelListCleanupUpdater.swift
```diff
@@ -0,0 +1,108 @@
...
+        // triggers, which leads to `Too many requests for user` backend error
+        needsRefreshQueries = false
+        currentlyTypingMembers = []
+        reads = []
+        queries = []
+    }
+}
+
+private extension ChannelListQueryDTO {
+    func asChannelListQuery<ExtraData: ChannelExtraData>() throws -> _ChannelListQuery<ExtraData> {
```
VojtaStavik, line 101: Please add documentation

Sources/StreamChat/Workers/ChannelListCleanupUpdater_Tests.swift
VojtaStavik: Yes, that's exactly the idea :)

VojtaStavik (changes requested):
I'd really appreciate seeing MUCH more documentation. Even though this is _just_ internal API it's a courtesy to our future selves to explain what is happening and why.

Sources/StreamChat/Utils/InternetConnection/Error+InternetNotAvailable.swift
dmigach: Got it, let's leave generic isBackendErrorWith400StatusCode so tests for this logic are simpler and cleaner, but add a local private proxy isTooManyMissingEventsToSyncError

VojtaStavik (approved):
LGTM 👍 Please squash and merge

PR #1050, opened 2021-05-04
Add date label to message list

olejnjak:
Add static label that changes its text based on message that is displayed below.

Currently it looks like this, I expect some changes there.

date label

When doing this I noticed that dynamic type doesn't work correctly on message list so I also fixed that.

### DIFF
--- a/Sources/StreamChatUI/ChatMessageList/ChatMessageListCollectionView.swift
+++ b/Sources/StreamChatUI/ChatMessageList/ChatMessageListCollectionView.swift
@@ -12,6 +12,113 @@ open class ChatMessageListCollectionView: UICollectionView {
     
     open var needsToScrollToMostRecentMessage = false
     open var needsToScrollToMostRecentMessageAnimated = false
+    
+    /// View used to display date of currently displayed messages
+    open lazy var scrollOverlayView: ChatMessageListScrollOverlayView = {
+        let scrollOverlayView = components.messageList.scrollOverlayView.init()
+        scrollOverlayView.isHidden = true
+        return scrollOverlayView.withoutAutoresizingMaskConstraints
+    }()
+    
+    private var isInitialized = false
+    
+    /// Usual `dataSource` method cast to `ChatMessageListCollectionViewDataSource`
+    private var chatDataSource: ChatMessageListCollectionViewDataSource? {
+        dataSource as? ChatMessageListCollectionViewDataSource
+    }
+    
+    private var contentOffsetObservation: NSKeyValueObservation?
+    
+    override open func didMoveToSuperview() {
+        super.didMoveToSuperview()
+        
+        guard !isInitialized, superview != nil else { return }
+        
+        isInitialized = true
+        
+        setUp()
+        setUpLayout()
+        setUpAppearance()
+        updateContent()
+    }
+    
+    open func setUp() {
+        // Setup `contentOffset` observation so `delegate` is free for anyone that wants to use it
+        contentOffsetObservation = observe(\.contentOffset) { cv, _ in
+            /// To display correct date we use bottom edge of `dateView` (we use `cv.layoutMargins.top` for both vertical offsets of `dateView`
+            let dateViewRefPoint = CGPoint(
+                x: cv.scrollOverlayView.center.x,
+                y: cv.scrollOverlayView.frame.maxY + cv.layoutMargins.top
+            )
+            let refPoint = cv.convert(dateViewRefPoint, to: cv)
+            // To get correct indexPath, we cannot use `cv.indexPathForItem(at:)` as it breaks collectionView layout
+            // and all cells have fix 72pt width, so we search which visible cell contains our `refPoint`
+            // and then search for its indexPath
+            /// Cell that contains our `refPoint` if any
+            let cell = cv.visibleCells.first(where: { $0.frame.contains(refPoint) })
+
+            // If we cannot find any indexPath for `cell` we try to use max visible indexPath (we have bottom to top) layout
+            guard let indexPath = cell.flatMap(cv.indexPath) ?? cv.indexPathsForVisibleItems.max() else { return }
+            
+            let overlayText = cv.chatDataSource?.collectionView(cv, scrollOverlayTextForItemAt: indexPath)
+            
+            // As cells can overlay our `dateView` we need to keep it above them
+            cv.bringSubviewToFront(cv.scrollOverlayView)
+            
+            // If we have no date we have no reason to display `dateView`
+            cv.scrollOverlayView.isHidden = (overlayText ?? "").isEmpty
+            cv.scrollOverlayView.content = overlayText
+            
+            // Apple's naming is quite weird as actually this property should rather be named `isScrolling`
+            // as it stays true when user stops dragging and scrollView is decelerating and becomes false
+            // when scrollView stops decelerating
+            //
+            // But this case doesn't cover situation when user drags scrollView to a certain `contentOffset`
+            // leaves the finger there for a while and then just lifts it, it doesn't change `contentOffset`
+            // so this handler is not called, this is handled by `scrollStateChanged`
+            // that reacts on `panGestureRecognizer` states and can handle this case properly
+            if !cv.isDragging {
+                cv.setOverlayViewAlpha(0)
+            }
+        }
+        
+        panGestureRecognizer.addTarget(self, action: #selector(scrollStateChanged))
+    }
+    
+    @objc
+    open func scrollStateChanged(_ sender: UIPanGestureRecognizer) {
+        switch sender.state {
+        case .began:
+            setOverlayViewAlpha(1)
+        case .ended, .cancelled, .failed:
+            // This case handles situation when user pans to certain `contentOffset`, leaves the finger there
+            // and then lifts it without `contentOffset` change, so `scrollView` will not decelerate, if it does,
+            // it is handled by `contentOffset` observation
+            if !isDecelerating {
+                setOverlayViewAlpha(0)
+            }
+        default: break
+        }
+    }
+    
+    open func setUpLayout() {
+        addSubview(scrollOverlayView)
+        
+        NSLayoutConstraint.activate([
+            scrollOverlayView.topAnchor.pin(equalTo: layoutMarginsGuide.topAnchor),
+            scrollOverlayView.centerXAnchor.pin(equalTo: layoutMarginsGuide.centerXAnchor),
+            scrollOverlayView.leadingAnchor.pin(greaterThanOrEqualTo: layoutMarginsGuide.leadingAnchor),
+            scrollOverlayView.trailingAnchor.pin(lessThanOrEqualTo: layoutMarginsGuide.trailingAnchor)
+        ])
+    }
+    
+    open func setUpAppearance() {
+        // Nothing to do
+    }
+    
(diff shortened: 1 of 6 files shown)


Sources/StreamChatUI/ChatMessageList/ChatMessageListCollectionView.swift
```diff
@@ -5,13 +5,112 @@
 import StreamChat
 import UIKit
 
+/// Protocol that adds delegate methods specific for `ChatMessageListCollectionView`
+public protocol ChatMessageListCollectionViewDelegate: UICollectionViewDelegate {
```
VojtaStavik, line 9: Isn't this more a data source than a delegate? 🤔

Sources/StreamChatUI/ChatMessageList/ChatMessageListCollectionView.swift
```diff
@@ -5,13 +5,112 @@
...
+    /// Get date for item at given indexPath
+    /// - Parameters:
+    ///   - collectionView: CollectionView requesting date
+    ///   - indexPath: IndexPath that should be used to get date
+    func collectionView(_ collectionView: ChatMessageListCollectionView, dateForItemAt indexPath: IndexPath) -> Date?
+}
```
VojtaStavik, line 15: What about something like this to make it more flexiable?
```swift
func collectionView(_ collectionView: ChatMessageListCollectionView, scrollOverlayStringForItemAt indexPath: IndexPath) -> String?
```

Sources/StreamChatUI/ChatMessageList/ChatMessageListCollectionView.swift
```diff
@@ -5,13 +5,112 @@
...
+
 /// The collection view that provides convenient API for dequeuing `_СhatMessageCollectionViewCell` instances
 /// with the provided content view type and layout options.
-open class ChatMessageListCollectionView: UICollectionView {
+open class ChatMessageListCollectionView: UICollectionView, Customizable, ComponentsProvider {
+    // Class does not use any ExtraData generic so we just need to make compiler happy
+    public typealias ExtraData = NoExtraData
```
VojtaStavik, line 21: Even though the compiler is happy our customers with custom extra data types won't be 😅 The `ExtraData` here must match other extra data of the client. So this should rather be:
```
open class ChatMessageListCollectionView<ExtraData: ExtraDataTypes>: UICollectionView, ... {}
```

Sources/StreamChatUI/ChatMessageList/ChatMessageListCollectionView.swift
```diff
@@ -5,13 +5,112 @@
...
     
     open var needsToScrollToMostRecentMessage = false
     open var needsToScrollToMostRecentMessageAnimated = false
+    
+    /// View used to display date of currently displayed messages
+    open lazy var dateView: ChatMessageListDateView = {
+        let dateView = components.messageList.dateView.init()
+        dateView.isHidden = true
+        return dateView.withoutAutoresizingMaskConstraints
+    }()
```
VojtaStavik, line 33: Maybe we can name this `scrollOverlayView`? This view should be visible only when the collection view is scrolling.

Sources/StreamChatUI/ChatMessageList/ChatMessageListDateView.swift
```diff
@@ -0,0 +1,55 @@
...
+import UIKit
+
+open class ChatMessageListDateView: _View, AppearanceProvider {
+    public static var dateFormatter: DateFormatter = {
+        let df = DateFormatter()
+        df.dateStyle = .short
+        df.timeStyle = .none
+        df.locale = .autoupdatingCurrent
+        return df
+    }()
```
VojtaStavik, line 14: Ideally we'd move this to the controller

VojtaStavik (changes requested):
Looks like a step in the right direction 👍  My main two comments are about the visibility of the overlay, and about making it more flexible.

Sources/StreamChatUI/ChatMessageList/ChatMessageListCollectionView.swift
olejnjak: So if I want to inject non-generic view from `ComponentsProvider` to another non-generic view, it is necessary to make both (and possibly all custom superviews) generic?

Sources/StreamChatUI/ChatMessageList/ChatMessageListCollectionView.swift
```diff
@@ -5,13 +5,109 @@
...
+        // Setup `contentOffset` observation so `delegate` is free for anyone that wants to use it
+        contentOffsetObservation = observe(\.contentOffset) { cv, _ in
+            /// To display correct date we use bottom edge of `dateView` (we use `cv.layoutMargins.top` for both vertical offsets of `dateView`
+            let dateViewRefPoint = CGPoint(x: cv.dateView.center.x, y: cv.dateView.frame.maxY + cv.layoutMargins.top)
+            let refPoint = cv.convert(dateViewRefPoint, to: cv)
+            // To get correct indexPath, we cannot use `cv.indexPathForItem(at:)` as it breaks collectionView layout
+            // and all cells have fix 72pt width, so we search which visible cell contains our `refPoint`
+            // and then search for its indexPath
+            /// Cell that containe
+            let cell = cv.visibleCells.first(where: { $0.frame.contains(refPoint) })
```
DominikBucher12, line 83: 👍

Sources/StreamChatUI/ChatMessageList/ChatMessageListCollectionView.swift
```diff
@@ -5,13 +5,109 @@
...
+    open func setUp() {
+        // Setup `contentOffset` observation so `delegate` is free for anyone that wants to use it
+        contentOffsetObservation = observe(\.contentOffset) { cv, _ in
+            /// To display correct date we use bottom edge of `dateView` (we use `cv.layoutMargins.top` for both vertical offsets of `dateView`
+            let dateViewRefPoint = CGPoint(x: cv.dateView.center.x, y: cv.dateView.frame.maxY + cv.layoutMargins.top)
+            let refPoint = cv.convert(dateViewRefPoint, to: cv)
+            // To get correct indexPath, we cannot use `cv.indexPathForItem(at:)` as it breaks collectionView layout
+            // and all cells have fix 72pt width, so we search which visible cell contains our `refPoint`
+            // and then search for its indexPath
+            /// Cell that containe
```
DominikBucher12, line 82: nit: unfinished documentation 😅

Sources/StreamChatUI/ChatMessageList/ChatMessageListCollectionView.swift
```diff
@@ -5,13 +5,109 @@
...
+        NSLayoutConstraint.activate([
+            dateView.topAnchor.pin(equalTo: layoutMarginsGuide.topAnchor),
+            dateView.centerXAnchor.pin(equalTo: layoutMarginsGuide.centerXAnchor),
+            dateView.leadingAnchor.pin(greaterThanOrEqualTo: layoutMarginsGuide.leadingAnchor),
+            dateView.trailingAnchor.pin(lessThanOrEqualTo: layoutMarginsGuide.trailingAnchor)
+        ])
+    }
+    
+    open func setUpAppearance() {
+        topContentInset = dateView.frame.height + 2 * layoutMargins.top
```
DominikBucher12: nit: DRY for line 62, is this even needed when it gets called on `layoutSubviews()` ?

Sources/StreamChatUI/ChatMessageList/ChatMessageListCollectionView.swift
```diff
@@ -99,4 +212,25 @@ open class ChatMessageListCollectionView: UICollectionView {
...
+        
+        collectionViewLayout.invalidateLayout()
+    }
+    
+    open func setOverlayViewAlpha(_ alpha: CGFloat, animated: Bool = true) {
+        let animations = { [scrollOverlayView] in
+            scrollOverlayView.alpha = alpha
+        }
+        
+        animated ? Animate(animations) : animations()
```
VojtaStavik, line 240: You can just do
```swift
Animate(isAnimated: animated) { ... }
```

DominikBucher12 (approved):
LGTM, just some minor comments :)

Sources/StreamChatUI/ChatMessageList/ChatMessageListCollectionView.swift
olejnjak: 🤦

Sources/StreamChatUI/ChatMessageList/ChatMessageListCollectionView.swift
olejnjak: Nope 🙂

Sources/StreamChatUI/ChatMessageList/ChatMessageListCollectionView.swift
```diff
@@ -5,13 +5,147 @@
...
+/// Protocol that adds delegate methods specific for `ChatMessageListCollectionView`
+public protocol ChatMessageListCollectionViewDataSource: UICollectionViewDataSource {
+    /// Get date for item at given indexPath
+    /// - Parameters:
+    ///   - collectionView: CollectionView requesting date
+    ///   - indexPath: IndexPath that should be used to get date
+    func collectionView<ExtraData: ExtraDataTypes>(
+        _ collectionView: ChatMessageListCollectionView<ExtraData>,
+        scrollOverlayTextForItemAt indexPath: IndexPath
+    ) -> String?
```
VojtaStavik, line 17: TBH having just `UICollectionView` here would probably be enough 🤔

Sources/StreamChatUI/ChatMessageList/ChatMessageListScrollOverlayView.swift
```diff
@@ -0,0 +1,48 @@
...
+// Copyright © 2021 Stream.io Inc. All rights reserved.
+//
+
+import UIKit
+
+/// View that is displayed as top overlay when message list is scrolling
+open class ChatMessageListScrollOverlayView: _View, AppearanceProvider {
+    open var text: String? {
+        didSet { updateContentIfNeeded() }
+    }
```
VojtaStavik, line 11: nit: this should ideally be called `content` (because of automatic SwiftUI wrapper export)

olejnjak:
Well everything should be addressed now, this is current state now.

current state

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -17,8 +17,16 @@ open class _ChatMessageListVC<ExtraData: ExtraDataTypes>:
...
+    ChatMessageListCollectionViewDataSource,
     GalleryContentViewDelegate {
+    open var overlayDateFormatter: DateFormatter = {
+        let df = DateFormatter()
+        df.dateStyle = .short
+        df.timeStyle = .none
+        df.locale = .autoupdatingCurrent
+        return df
+    }()
+
```
VojtaStavik, line 29: Can you please add documentation and ideally also move it somewhere else in the file? Like this it looks like this is the most important thing in the controller :D

VojtaStavik (approved):
LGTM 👍 Just some minor comments and adjustments.

PR #1053, opened 2021-05-05
[CIS-832] Set ENABLE_TESTABILITY to YES for StreamChatUI so we could archive it

dmigach:
Because recently @testable import was added to DemoApp and we cannot archive it


b-onc (approved):

VojtaStavik (approved):
OK as a temp workaround 👍

PR #1055, opened 2021-05-05
Fix swift warnings

b-onc:
With the new `async/await` that'll eventually be introduced to Swift, our own `await` stuff was causing a lot of warnings, now that's resolved, alongside other minor warnings.

### DIFF
--- a/Sources/StreamChat/APIClient/APIClient_Tests.swift
+++ b/Sources/StreamChat/APIClient/APIClient_Tests.swift
@@ -78,7 +78,7 @@ class APIClient_Tests: StressTestCase {
         let testEndpoint = Endpoint<Data>.mock()
         
         // Create a request and assert the result is failure
-        let result = try await { apiClient.request(endpoint: testEndpoint, completion: $0) }
+        let result = try waitFor { apiClient.request(endpoint: testEndpoint, completion: $0) }
         AssertResultFailure(result, testError)
     }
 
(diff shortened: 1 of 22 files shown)


Sources/StreamChat/APIClient/APIClient_Tests.swift
```diff
@@ -78,7 +78,7 @@ class APIClient_Tests: StressTestCase {
         let testEndpoint = Endpoint<Data>.mock()
         
         // Create a request and assert the result is failure
-        let result = try await { apiClient.request(endpoint: testEndpoint, completion: $0) }
+        let result = try `await` { apiClient.request(endpoint: testEndpoint, completion: $0) }
```
VojtaStavik, line 81: Shouldn't we rather rename it so something that won't cause the conflict before it will be possible to migrate? Like `_await` for example? :)

VojtaStavik (approved):

Sources/StreamChat/APIClient/APIClient_Tests.swift
b-onc: Or even like `waitFor` so it's 100% unrelated to builtin `await` :D 
```swift
let result = try waitFor { apiClient.request(...) }
```

VojtaStavik: image

Tests/Shared/CustomAssertions/AssertAsync.swift
```diff
@@ -336,7 +336,7 @@ struct AssertAsync {
     }
 }
 
-@_functionBuilder
+@resultBuilder
```
VojtaStavik, line 339: Isn't this Swift 5.4 only?

b-onc: Yep, hence it was failing in CI, so I removed it for now

PR #1056, opened 2021-05-05
Remove `Chat` prefix from views that don't need it

nuno-vieira:
# In this PR
Removes the `Chat` prefix from the views that don't need it.


skorepak (approved):

Sources/StreamChatUI/CommonViews/MessageQuoteView/MessageQuoteView.swift
```diff
@@ -16,10 +16,10 @@ struct ChatMessageQuoteAvatarAlignment: Equatable {
 }
 
 /// A view that displays a quoted message.
-public typealias ChatMessageQuoteView = _ChatMessageQuoteView<NoExtraData>
+public typealias MessageQuoteView = _MessageQuoteView<NoExtraData>
```
VojtaStavik, line 19: Is this a quote view for a "message" or a `ChatMessage`? 🤔 Maybe the latter?

Sources/StreamChatUI/CommonViews/CreateChannelButton/CreateChannelButton.swift
```diff
@@ -6,7 +6,7 @@ import StreamChat
 import UIKit
 
 /// A Button subclass that should be used for creating new channels.
-open class ChatChannelCreateNewButton: _Button, AppearanceProvider {
+open class CreateChannelButton: _Button, AppearanceProvider {
```
VojtaStavik, line 9: Same here? Is it a button for creating a new "channel" or creating a new `ChatChannel`? 🤔 Probably the former?

VojtaStavik (approved):
LGTM, just some nit questions. Not sure what are the correct answer there to be honest 🤔

Sources/StreamChatUI/CommonViews/MessageQuoteView/MessageQuoteView.swift
nuno-vieira: Then should it be `ChatMessageQuoteView` or `QuoteChatMessageView` ? This is why having `Chat` prefix in some views makes It harder to name things, eventually the views won't have consistent and will be confusing IMO 🤔

Sources/StreamChatUI/CommonViews/CreateChannelButton/CreateChannelButton.swift
nuno-vieira: In this case, it is clear that it should be `CreateChatChannelButton` right? `ChatChannelCreateButton` does not make sense while reading it

Sources/StreamChatUI/CommonViews/MessageQuoteView/MessageQuoteView.swift
nuno-vieira: Done ✅

Sources/StreamChatUI/CommonViews/CreateChannelButton/CreateChannelButton.swift
nuno-vieira: Done ✅

PR #1057, opened 2021-05-05
Reorganize StreamChatUI Folders

nuno-vieira:
# In this PR
Reorganizes some of the folders in `StreamChatUI` since they were all over the place.
- Adds a new Extensions folder, and moves all extensions from Utils to this new folder and subfolders, each subfolder is the module that the extensions belong to.
- Introduces a Customization folder.
- Renames CommonViews to CommonComponents
- Renames Utils to ComponentUtils and only the building blocks of the component belong in this folder.

Please feel free to give some feedback :) 

**Result:**


VojtaStavik (approved):
In general, I'm a fan of more flat structures. When the folder structure is too granular it takes more brainpower to figure out where to put certain files. I'm ok with whatever the team agrees on, however, the flatten the structure will be the better IMO :) 

Screen Shot 2021-05-06 at 14 56 41

PR #1061, opened 2021-05-06
Update `ChatMessageAttachment` to have non-optional payload

evsaev:
**This PR**:
- makes `ChatMessageAttachment.payload` non-optional
- adds dynamic payload lookup to `ChatMessageAttachment`

### DIFF
--- a/Sources/StreamChat/Models/Attachments/ChatMessageAttachment.swift
+++ b/Sources/StreamChat/Models/Attachments/ChatMessageAttachment.swift
@@ -11,11 +11,16 @@ public struct AttachmentUploadingState: Equatable {
     public let state: LocalAttachmentState
 }
 
+@dynamicMemberLookup
 public struct _ChatMessageAttachment<Payload> {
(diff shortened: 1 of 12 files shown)


Sources/StreamChat/Models/Attachments/ChatMessageAttachment.swift
```diff
@@ -26,14 +26,18 @@ extension _ChatMessageAttachment {
...
-        let concretePayload: Payload?
+        let concretePayload: Payload
         switch payload {
         case let payload as Payload:
             concretePayload = payload
         case let data as Data:
-            concretePayload = try? JSONDecoder.stream.decode(Payload.self, from: data)
+            guard
+                let decodedPayload = try? JSONDecoder.stream.decode(Payload.self, from: data)
+            else { return nil }
```
VojtaStavik, line 41: We shouldn't silently fail here

VojtaStavik (approved):
Nice 👍

PR #1062, opened 2021-05-06
Audit public-facing attachments API

evsaev:
**This PR**:
- renames `AttachmentEnvelope` to `ChatMessageAttachmentEnvelope
- renames `AttachmentPayload` to `MessageAttachmentPayload`
- renames `AttachmentPayloadType` to `AttachmentPayload`
- renames `AttachmentLinkPayload` to `LinkAttachmentPayload`
- renames `AttachmentGiphyPayload` to `GiphyAttachmentPayload`
- renames `AttachmentFilePayload` to `FileAttachmentPayload`
- renames `AttachmentImagePayload` to `ImageAttachmentPayload`
- rename `ChatMessageAttachment` to `AnyChatMessageAttachment`
- fixes `URL` namespace pollution with attachments stuff

_Note: docs, missing tests, `unknown` attachment type, and attachment scores will be added in the follow-up PRs_


Sources/StreamChat/APIClient/Endpoints/Payloads/MessageAttachmentPayload.swift
```diff
@@ -5,7 +5,7 @@
 import Foundation
 
 /// A type that describes attachment JSON payload.
-struct AttachmentPayload {
+struct MessageAttachmentPayload {
```
nuno-vieira, line 8: I'm only not sure about this naming, it feels odd to see the `Message` without the `Chat` prefix, since everywhere in the codebase we use `ChatMessage`. 🤔  Although I know `Message` is not related to the model per se in this case, this will confuse us and the customers. From what I understand, this is like the "Default" attachment payload, so why not keep it simple and name it `DefaultAttachmentPayload`?  WDYT?

Sources/StreamChat/Models/Message.swift
```diff
@@ -225,7 +220,7 @@ public extension _ChatMessage {
     func attachments<Payload: AttachmentPayload>(
         payloadType: Payload.Type
     ) -> [_ChatMessageAttachment<Payload>] {
-        attachments.compactMap {
+        _attachments.compactMap {
             $0.attachment(payloadType: payloadType)
         }
```
VojtaStavik: I'm not sure about this decision. If we expose it only like this it's not possible to check the total amount of attachments of the message because you will always get just the attachment of the type you specify.

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -337,7 +337,6 @@ open class _ChatMessageListVC<ExtraData: ExtraDataTypes>:
     /// Executes the provided action on the message
     open func handleTapOnAttachmentAction(
         _ action: AttachmentAction,
-        for attachment: ChatMessageAttachment,
```
VojtaStavik, line 340: maybe we should replace it with `AttachmentId`?

Sources/StreamChat/Models/Attachments/ChatMessageAttachment.swift
```diff
@@ -4,8 +4,6 @@
 
 import Foundation
 
-public typealias ChatMessageAttachment = _ChatMessageAttachment<Any>
-
```
VojtaStavik, line 8: I think we should keep this but rename it to `AnyChatMessageAttachment` to match the Swift naming style for type-erased types.

VojtaStavik (changes requested):
All LGTM but the last commit. Let's talk in Slack about details.

VojtaStavik:
Merging this for now, we will open a follow up PR soon.

Sources/StreamChat/Models/Attachments/AttachmentTypes.swift
```diff
@@ -152,6 +137,19 @@ public struct AttachmentFile: Codable, Hashable {
...
+
+        let fileType = AttachmentFileType(ext: url.pathExtension)
+        let attributes = try? FileManager.default.attributesOfItem(atPath: url.path)
+
+        self.init(
+            type: fileType,
+            size: attributes?[.size] as? Int64 ?? 0,
+            mimeType: fileType.mimeType
+        )
+    }
```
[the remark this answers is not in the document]
evsaev: Made `init` throwable in b3b5f72ccb0a50beec5fbe8935f03cf43f804139

Sources/StreamChat/Models/Message.swift
evsaev: Reverted those changes and kept attachments publicly exposed. Will change this in the scope of `attachment score` PR

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
evsaev: This was changed in the PR for `GiphyAttachmentViewInjector` but you are right 👍 I'd add `attachment: ChatMessageGiphyAttachment` param to delegate's method (now only `action` and `indexPath` are given)

evsaev: cc @DominikBucher12

VojtaStavik (approved):
Thank you! 👏

PR #1066, opened 2021-05-10
Fix container stack view when a subview is hidden

fortmarek:

When a subview of `ContainerStackView` was hidden and distribution was set to `.equal`, constraints forcing views to be the same width/height were not deactivated. This PR resolves this issue.


Sources/StreamChatUI/CommonViews/ContainerStackView.swift
```diff
@@ -437,6 +431,27 @@ public class ContainerStackView: UIView {
...
+                    lView.widthAnchor.constraint(equalTo: rView.widthAnchor)
+                )
+            } else {
+                equalConstraints.append(
+                    lView.heightAnchor.constraint(equalTo: rView.heightAnchor)
+                )
+            }
+        }
+        NSLayoutConstraint.activate(equalConstraints)
+    }
```
VojtaStavik, line 454: We already have a pattern here where we story the constraints in a dictionary so we can selectively disable/enable them if needed. Why not do it the same here? This feels a bit hacky + it does a lot of unnecessary work. We should be able to do just something similar to this:
```swift
distributionConstraintByView[subview]?.isActive = true / false

// or even...

distributionConstraintsByView[subview]?.forEach { $0.isActive = true / false }
```

VojtaStavik (changes requested):
Thanks for fixing this 🎉 I think it should be possible to do this in a bit more robust way, please see my comment.

Sources/StreamChatUI/CommonViews/ContainerStackView.swift
fortmarek: it is not that simple in this case. Let's say I have three views: A, B, C. Constraints for the same width are created between A-B and B-C. I hide subview B. Therefore, I should create a new constraint between A and C.

fortmarek: I thought recomputing them altogether would be more straightforward - or do you think we should try to be "smarter" here? Even in the case I presented, if you have a lot of views, most of them will have the same constraints before and after

VojtaStavik (approved):
Nice 🎉

nuno-vieira (approved):
LGTM! 🎉

PR #1068, opened 2021-05-10
Fix newly joined members not having ChannelReads

b-onc:
### Cases:

1. User A is watching the channel, non-member User B sends `messageRead` event
  -> We discard the event and `log.debug` it
1. User A is watching the channel, new member User B sends `messageRead` event
  -> We create a read event for User B, **this is the fix introduced in this PR**
1. User A is watching the channel, old member User B sends `messageRead` event
  -> Since User B is an old member, we received their read object when we queried the channel
1. User A is watching the channel and is not a member, User B sends a message
  -> User A doesn't have a read object until they re-query the channel (backend bug) or they join, we `log.info` the case
1. User A is watching the channel and is a new member, User B sends a message
   -> `MemberEventMiddleware` received the `memberAdded` event and re-queries the channel, creating a read object
   **-> This is the only edge-edge case**, before `MemberEventMiddleware` finishes creating the read object, all new messages cause the `log.info` log
1. User A is watching the channel and is an old member, User B sends a message
  -> Since User A is an old member, we received their read object when we queried the channel


VojtaStavik:
@b-onc can you add changelog entry for this, please?

PR #1073, opened 2021-05-13
[CIS-838] Implement typing indicator

DominikBucher12:
# What this PR do:
- Implements TypingIndicator into `ChatMessageListVC`.

# How to test it
- Open any chat on 2 different devices, write something from one or other account and expect typing indicator to show up.

Dicussion:
- Should this be implemented for `ChatThreadVC` as well?
- Code formatting not working? I aligned the code via Xcode and this seems to have a lot of changes in whitespaces, never happened before... But it's probably safe to ignore now... 

Preview of typing event: 

https://user-images.githubusercontent.com/17381941/118067602-2e21b680-b3a1-11eb-8888-145d203d4229.mov


Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -110,14 +116,24 @@ open class _ChatMessageListVC<ExtraData: ExtraDataTypes>:
...
         messageComposerBottomConstraint?.isActive = true
+        
+        if channelController.channel?.config.typingEventsEnabled ?? false {
+            view.addSubview(typingIndicatorView)
+            typingIndicatorView.heightAnchor.pin(equalToConstant: 22).isActive = true
+            typingIndicatorView.pin(anchors: [.leading, .trailing], to: view)
+            typingIndicatorView.bottomAnchor.pin(equalTo: messageComposerViewController.view.topAnchor).isActive = true
+            typingIndicatorView.isHidden = true
+            
+            collectionView.contentInset.bottom = 22
```
VojtaStavik, line 133: nit a) would be nice to have the `22` abstracted into a local variable when it's used twice
nit b) you shouldn't override the existing value but rather do something like `+= 22`

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -385,9 +401,38 @@ open class _ChatMessageListVC<ExtraData: ExtraDataTypes>:
...
+        _ channelController: _ChatChannelController<ExtraData>,
+        didChangeTypingMembers typingMembers: Set<_ChatChannelMember<ExtraData.User>>
+    ) {
+        if typingMembers.contains(where: { member in member.id == self.channelController.client.currentUserId }) {
+            let typingMembersWithoutCurrentUser = typingMembers.filter { $0.id != self.channelController.client.currentUserId }
+            showTypingIndicatorIfNeeded(typingMembers: typingMembersWithoutCurrentUser)
+        } else {
+            // Current user is not member of the room but can see that someone is actually typing
+            showTypingIndicatorIfNeeded(typingMembers: typingMembers)
+        }
```
VojtaStavik, line 415: I don't understand the logic here 🤔 Why should it matter if the current user is a member or not? In no case I want to show him as a typing user. Isn't the following enough?
```swift
let typingMembersWithoutCurrentUser = typingMembers
    .filter { $0.id != self.channelController.client.currentUserId }
showTypingIndicatorIfNeeded(typingMembers: typingMembers)
```

VojtaStavik (approved):
LGTM just some minor change to code style, thank you! 🚀 

(btw - feel free to remove the white-space changes, there's no need to include them in the commit)

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -110,14 +116,24 @@ open class _ChatMessageListVC<ExtraData: ExtraDataTypes>:
...
         addChildViewController(messageComposerViewController, targetView: view)
-
+        
         messageComposerViewController.view.topAnchor.pin(equalTo: collectionView.bottomAnchor).isActive = true
         messageComposerViewController.view.leadingAnchor.pin(equalTo: view.safeAreaLayoutGuide.leadingAnchor).isActive = true
         messageComposerViewController.view.trailingAnchor.pin(equalTo: view.safeAreaLayoutGuide.trailingAnchor).isActive = true
         messageComposerBottomConstraint = messageComposerViewController.view.bottomAnchor.pin(equalTo: view.bottomAnchor)
         messageComposerBottomConstraint?.isActive = true
+        
+        if channelController.channel?.config.typingEventsEnabled ?? false {
```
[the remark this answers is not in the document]
DominikBucher12: Done

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
DominikBucher12: Totally right, changed.

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
DominikBucher12: Yes, the following is enough, the solution above is a bit more complex than it needs to be...

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -388,6 +406,30 @@ open class _ChatMessageListVC<ExtraData: ExtraDataTypes>:
...
+            guard let member = typingMembers.first(where: { user in user.name != nil }), let name = member.name else {
+                typingIndicatorView.content = L10n.MessageList.TypingIndicator.typingUnknown
+                typingIndicatorView.isHidden = false
+                return
+            }
+            
+            typingIndicatorView.content = L10n.MessageList.TypingIndicator.users(name, typingMembers.count - 1)
+            typingIndicatorView.isHidden = false
+        }
+    }
```
nuno-vieira, line 431: IMO I think we should avoid as much as possible using "ifNeeded" methods. Although those are used on native iOS APIs, they are usually related to layout. I would say this is a bad practice for business logic because it hides the condition of when the method is executed, it is less flexible and customizable in our case. If we just do: 

```Swift
        ....
        if typingMembers.isEmpty {
            typingIndicatorView.isHidden = true
        } else {
            showTypingIndicator(typingMembers: typingMembersWithoutCurrentUser)
        }
        ....

    open func showTypingIndicator(typingMembers: Set<_ChatChannelMember<ExtraData.User>>) {
            // If we somehow cannot fetch any member name, we simply show that `Someone is typing`
            guard let member = typingMembers.first(where: { user in user.name != nil }), let name = member.name else {
                typingIndicatorView.content = L10n.MessageList.TypingIndicator.typingUnknown
                typingIndicatorView.isHidden = false
                return
            }
            
            typingIndicatorView.content = L10n.MessageList.TypingIndicator.users(name, typingMembers.count - 1)
            typingIndicatorView.isHidden = false
    }
```

With this, we can also make the `showTypingIndicator` function `open` and the customer can customize the indicator style.

nuno-vieira: And also exporting `typingIndicatorView.isHidden = true` to 
```Swift
open func hideTypingIndicator() {
   typingIndicatorView.isHidden = true
}
```
In case the client wants to customize the hiding

DominikBucher12: Wow thanks, this is actually great solution and makes it much more customizable... Thank you!

Sources/StreamChatUI/Composer/ComposerVC.swift
```diff
@@ -262,7 +262,10 @@ open class _ComposerVC<ExtraData: ExtraDataTypes>: _ViewController,
 
     override open func viewDidDisappear(_ animated: Bool) {
         super.viewDidDisappear(animated)
-
+        
+        // We need to cleanup the state when composer not to type when it is not presented.
+        channelController.sendStopTypingEvent()
```
DominikBucher12, line 267: Implementing `stopTyping` inside `EventsSender`'s `deinit` feels like unnecessary call.. But this is up to discussion since `EventsSender` is used only for sending `typingEvents`... Maybe rename it?

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -406,6 +424,41 @@ open class _ChatMessageListVC<ExtraData: ExtraDataTypes>:
         updateNavigationTitle()
     }
 
+    open func channelController(
+        _ channelController: _ChatChannelController<ExtraData>,
+        didChangeTypingMembers typingMembers: Set<_ChatChannelMember<ExtraData.User>>
+    ) {
+        let typingMembersWithoutCurrentUser = typingMembers.filter { $0.id != self.channelController.client.currentUserId }
+        
+        if typingMembers.isEmpty {
```
skorepak, line 433: Shouldn't this be `typingMembersWithoutCurrentUser`?  
If only the current user is typing, it will be in `typingMembers`, so this condition will hit the `else` branch but you will pass empty set into `showTypingIndicator`, which will use typingUnknown and show "Someone is typing" :D

DominikBucher12: yep I fixed this already, but thanks :D

Sources/StreamChatUI/Composer/ComposerVC.swift
DominikBucher12: Resolved by renaming `EventsSender` to `TypingEventsSender` and implementing on `deinit`

Sources/StreamChat/Workers/TypingEventsSender.swift
```diff
@@ -17,10 +17,18 @@ extension TimeInterval {
...
     var timer: Timer.Type = DefaultTimer.self
-        
+    /// ChannelId for channel that typing has occured in. Stored to stop typing when `TypingEventsSender` is dealocated
+    private var typingChannelId: ChannelId?
+    
     @Atomic private var currentUserTypingTimerControl: TimerControl?
     @Atomic private var currentUserLastTypingDate: Date?
     
+    deinit {
+        // We need to cleanup the state when composer not to type when it is not presented.
```
VojtaStavik, line 27: This sentence is very hard to decode 🙃

Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
```diff
@@ -10,8 +10,16 @@ public typealias ChatChannelListItemView = _ChatChannelListItemView<NoExtraData>
 
 /// An `UIView` subclass that shows summary and preview information about a given channel.
 open class _ChatChannelListItemView<ExtraData: ExtraDataTypes>: _View, ThemeProvider, SwiftUIRepresentable {
+    /// The content of this view.
+    public struct Content<ExtraData: ExtraDataTypes> {
```
VojtaStavik, line 20: Does this have to be generic? 🤔 I think you can access the `ExtraData` of the enclosing type.

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -408,6 +426,48 @@ open class _ChatMessageListVC<ExtraData: ExtraDataTypes>:
...
+    open func channelController(
+        _ channelController: _ChatChannelController<ExtraData>,
+        didChangeTypingMembers typingMembers: Set<_ChatChannelMember<ExtraData.User>>
+    ) {
+        let typingMembersWithoutCurrentUser = typingMembers
+            .sorted(
+                by: {
+                    guard let firstMemberName = $0.name, let secondMemberName = $1.name else { return false }
+                    return firstMemberName < secondMemberName
+                }
```
VojtaStavik, line 438: Why not just `.sorted { $0.id < $1.id }` ? :)

Sources/StreamChatUI/Resources/en.lproj/Localizable.stringsdict
```diff
@@ -13,11 +13,11 @@
...
 			<string>d</string>
 			<key>zero</key>
-			<string>%1$@ is typing</string>
+			<string> is typing</string>
 			<key>one</key>
-			<string>%1$@ and %2$d more is typing</string>
+			<string> and %2$d more is typing</string>
 			<key>other</key>
-			<string>%1$@ and %2$d more are typing</string>
+			<string> and %2$d more are typing</string>
```
VojtaStavik, line 20: The approach here was correct IMO (we talked about it on a call)

Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
```diff
@@ -8,14 +8,25 @@ import UIKit
 /// An `UIView` subclass that shows summary and preview information about a given channel.
 public typealias ChatChannelListItemView = _ChatChannelListItemView<NoExtraData>
 
+/// The content of this view.
+///
+/// Initializing `Content` as `_ChatChannelListItemView<NoExtraData>.Content<NoExtraData>`
+/// probably pushes the compiler too far and makes `Content` ambiguous according to `SwiftUIRepresentable.Content`
+public typealias ChatChannelListItemViewContent = _ChatChannelListItemView<NoExtraData>.Content<NoExtraData>
+
```
VojtaStavik, line 16: please see my comment about Content not having to be generic

Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
```diff
@@ -86,7 +97,7 @@ open class _ChatChannelListItemView<ExtraData: ExtraDataTypes>: _View, ThemeProv
     /// Text of `titleLabel` which contains the channel name.
     open var titleText: String? {
         if let content = content {
-            return components.channelList.channelNamer(content.channel, content.currentUserId)
+            return components.channelList.channelNamer(content.channel, content.channel.membership?.id)
```
VojtaStavik: Why?

VojtaStavik (approved):
LGTM, just some minor issues to fix. thanks 👍

Sources/StreamChat/Workers/TypingEventsSender.swift
VojtaStavik: @DominikBucher12 Also, this should be tested.

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -408,6 +426,48 @@ open class _ChatMessageListVC<ExtraData: ExtraDataTypes>:
...
+        if typingMembersWithoutCurrentUser.isEmpty {
+            hideTypingIndicator()
+        } else {
+            showTypingIndicator(typingMembers: typingMembersWithoutCurrentUser)
+        }
+    }
+    
+    /// Shows typing Indicator
+    /// Override this method to implement your custom logic and animations for showingthe typing indicator.
+    /// - Parameter typingMembers: typing members gotten from
```
nuno-vieira, line 451:
```suggestion
    /// Shows typing Indicator
    /// - Parameter typingMembers: typing members gotten from
```
No need to describe that the customer can customize this method, it is open, so the customer will understand it can override it. In every `open` method we are not describing that it is overridable so let's keep it consistent 👍

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -408,6 +426,48 @@ open class _ChatMessageListVC<ExtraData: ExtraDataTypes>:
...
+            return
+        }
+        
+        typingIndicatorView.content = L10n.MessageList.TypingIndicator.users(name, typingMembers.count - 1)
+        typingIndicatorView.isHidden = false
+    }
+    
+    /// Hides typing Indicator
+    /// Override this method to implement your custom logic and animations for hiding the typing indicator.
+    /// - Parameter typingMembers: typing members gotten from
```
nuno-vieira, line 466:
```suggestion
    /// Hides typing Indicator
    /// - Parameter typingMembers: typing members gotten from
```
Same as above

Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
DominikBucher12: Fixed.

VojtaStavik (approved):

PR #1074, opened 2021-05-13
[CIS-791] Make CDN customizable

dmigach:
A draft PR for initial assessment of implementation


Sources/StreamChat/Workers/Background/AttachmentUploader_Tests.swift
```diff
@@ -66,11 +66,11 @@ final class AttachmentUploader_Tests: StressTestCase {
...
             // Assert attachment is in `.pendingUpload` state.
             XCTAssertEqual(attachment.localState, .pendingUpload)
-
+            
             // Wait attachment uploading begins.
             AssertAsync.willBeEqual(
-                apiClient.uploadFile_endpoint.flatMap(AnyEndpoint.init),
-                AnyEndpoint(.uploadAttachment(with: attachmentId, type: attachment.attachmentType))
+                apiClient.uploadFile_attachment?.id,
+                attachment.asAnyModel()?.id
```
VojtaStavik, line 73: if both are `nil` it can produce false positive. We should probably `XCTUnwrap` the `attachment.asAnyModel()?.id` value.

Sources/StreamChat/APIClient/APIClient_Tests.swift
```diff
@@ -4,6 +4,8 @@
 
 @testable import StreamChat
 import XCTest
+@testable
+import StreamChatTestTools
```
VojtaStavik, line 8: Are you sure you have the formatter config properly in the pre-commit hook? 🤔

VojtaStavik (approved):
LGTM 👍 Just nits. Thank you!

Sources/StreamChat/APIClient/APIClient_Tests.swift
```diff
@@ -282,43 +180,28 @@ class APIClient_Tests: StressTestCase {
...
+        
+        XCTAssertTrue(cdnClient.uploadAttachmentMockFunc.called)
+        
+        let parameters = try XCTUnwrap(cdnClient.uploadAttachmentMockFunc.calls.last)
+        
+        parameters.1?(42)
+        XCTAssertEqual(progress, 42)
+        
+        parameters.2(.success(.sampleYodaImage))
+        XCTAssertEqual(try result?.get(), .sampleYodaImage)
```
b-onc, line 204: `42` and `sampleYogaImage` shouldn't be hardcoded IMHO.

Sources/StreamChat/Models/Attachments/ChatMessageAttachment_Sample.swift
```diff
@@ -0,0 +1,23 @@
...
+        uploadingState: AttachmentUploadingState? = nil
+    ) -> ChatMessageAttachment {
+        ChatMessageAttachment(
+            id: id,
+            type: type,
+            payload: payload,
+            uploadingState: uploadingState
+        )
+    }
+}
```
b-onc, line 23: Is this dummy data? We put these under `StreamChatTestTools > DummyData` and name them `dummy`

Sources/StreamChat/APIClient/DefaultCDNClient_Tests.swift
```diff
@@ -0,0 +1,223 @@
...
+        RequestRecorderURLProtocol.startTestSession(with: &sessionConfiguration)
+        MockNetworkURLProtocol.startTestSession(with: &sessionConfiguration)
+        
+        return DefaultCDNClient(
+            encoder: encoder,
+            decoder: decoder,
+            sessionConfiguration: sessionConfiguration
+        )
+    }
+}
```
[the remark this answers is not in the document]
dmigach: There is two ideas behind that:
- setUp() has an issue of shared state, ideally every test should be isolated
- it has no means of behaviour parametrisation which this newly introduced pattern helps with

As for your concern on two extra lines - Voijta actually has a pretty nice idea on how to optimise it, but even in the current state imho I don't see any serious issues with a couple of extra lines because the benefits outweigh the downsides and it's not like there are options here because shared fixture is a code smell in tests. So it's just a fundamental feature of any test - it has a setup stage

Sources/StreamChat/Models/Attachments/ChatMessageAttachment_Sample.swift
dmigach: I wanted to raise this with the team
if you think about it, it's not really a dummy, because it's a totally normal fully functioning object. So I wanted to suggest this little improvement to naming for a method that helps you to build this logic

Sources/StreamChat/APIClient/APIClient_Tests.swift
dmigach: It does trigger every time I commit 🤔 
```
Running SwiftFormat...
Swiftformat completed successfully.
```

Sources/StreamChat/Workers/Background/AttachmentUploader_Tests.swift
dmigach: Nice catch!

Sources/StreamChat/APIClient/APIClient_Tests.swift
dmigach: Fair, extracted to variables

b-onc: I don't know how to handle progress, but `result` should be like
```swift
XCTAssertEqual(try result?.get(), attachment.file)
```
or something

dmigach: Not really I'm afraid, it has nothing to do with attachment.file
Basically we just make sure that completions we pass to APIClient it then proxies to CDNClient

Sources/StreamChat/Models/Attachments/ChatMessageAttachment_Sample.swift
nuno-vieira: I would say the right technical naming for this kind of test doubles is a `stub`. A dummy is normally when you want to just pass data to conform to an interface but the values of the dummy don't really influence the test. For stubs, you pre-populate the data in order for a specific test to behave in some manner. Because this is so subjective and makes it hard to think about how to name your mocks/stubs etc... I just default everything to `.mock()`. In this case, because we are already using `dummy` everywhere, let's keep it like that for now, since it is the pattern that is currently in the project. We can create a tech debt ticket to change this tho after discussing it with the team 👍

Tests/Shared/MockFunc.swift
```diff
@@ -0,0 +1,61 @@
...
+    public func succeeds<T, Error>(_ value: T)
+        where Out == Result<T, Error> {
+        result = { _ in .success(value) }
+    }
+
+    public func fails<T, Error>(_ error: Error)
+        where Out == Result<T, Error> {
+        result = { _ in .failure(error) }
+    }
+}
```
nuno-vieira, line 61: Although this is nice, and it has the possibility of having more consistency when mocking functions, my concern is that we are introducing a lot of custom patterns to write tests. We already have `AssertAsync`, and now `MockFunc`, this has the potential for making it harder to reason about how to make a test, than helping. Especially since right now, we are not using this everywhere.

nuno-vieira (approved):
Looks really good 👌 Just added some thoughts, but none of them are blockers 👍

PR #1071, opened 2021-05-11
Audit ChatMessageComposer Attachments

skorepak:
## Description of the pull request
Audit and merge Image and Documents Attachments in the `ChatMessageComposer`

### I have
- Moved the CollectionView data source and delegate from collection view itself into separated ViewController and used this as a child view controller in the `ChatMessageComposerVC`
- Merged most of the views for Images and Documents and IMO simplified the code in exchange for two switches based on attachment type
- Renamed Documents to Files to match the name of attachment type
- Removed a few magic constants

## Links
https://stream-io.atlassian.net/browse/CIS-749

PR #1076, opened 2021-05-13
New `ComposerVC` Data Flow

nuno-vieira:
# In this PR
- Renames `ComposerViewController`-> `ComposerVC`
- Fixes close button in composer to have wrong image
- Renames `InputChatMessageView.inputTextView` -> `InputChatMessageView.textView`
- Refactors `InputChatMessageView` so it has `Content`
- Major `ComposerVC` refactor with new data flow.


Sources/StreamChatUI/CommonViews/Attachments/AttachmentPreview/ImageAttachmentPreview.swift
```diff
@@ -12,14 +12,32 @@ public struct ImageAttachmentPreview: AttachmentPreview {
...
+    public let size: Int64
+    public let localURL: URL
+
+    public init(image: UIImage, localURL: URL, name: String, size: Int64) {
+        self.image = image
+        self.name = name
+        self.size = size
+        self.localURL = localURL
+    }
+}
```
nuno-vieira, line 43: Ignore this bit, this will be changed, when https://github.com/GetStream/stream-chat-swift/pull/1071 is merged

Sources/StreamChatUI/Composer/ComposerVC.swift
```diff
@@ -0,0 +1,624 @@
...
+    }
+
+    /// The delegate of the ComposerVC that notifies composer events.
+    open weak var delegate: ComposerVCDelegate?
+
+    /// A symbol that is used to recognise when the user is mentioning a user.
+    open var mentionSymbol = "@"
+
+    /// A symbol that is used to recognise when the user is typing a command.
+    open var commandSymbol = "/"
```
VojtaStavik, line 163: 👍 👍 👍

Sources/StreamChatUI/Composer/ComposerVC.swift
```diff
@@ -0,0 +1,624 @@
...
+            return
+        }
+
+        if let (typingMention, mentionRange) = typingMention(in: composerView.inputMessageView.textView) {
+            showMentionSuggestions(for: typingMention, mentionRange: mentionRange)
+            return
+        }
+
+        dismissSuggestions()
+    }
```
VojtaStavik: So easy to understand and follow 😍 👍

Sources/StreamChatUI/Composer/ComposerVC.swift
```diff
@@ -254,25 +274,28 @@ open class _ComposerVC<ExtraData: ExtraDataTypes>: _ViewController,
...
             Animate {
-                self.composerView.confirmButton.isHidden = true
-                self.composerView.sendButton.isHidden = false
-                self.composerView.headerView.isHidden = true
+                self.composerView.confirmButton.isHidden = false
+                self.composerView.sendButton.isHidden = true
+                self.composerView.headerView.isHidden = false
             }
+        default:
+            break
```
VojtaStavik, line 298: Maybe we can add some `log.debug` here for debugging purposes? Even though it's not an error, it could be handy to know about this.

Sources/StreamChatUI/Composer/ComposerVC.swift
```diff
@@ -10,6 +10,20 @@ public protocol ComposerVCDelegate: AnyObject {
     func composerDidCreateNewMessage()
 }
 
+/// The possible composer states. An Enum is not used so it does not cause
+/// future breaking changes and is possible to extend with new cases.
+public class ComposerState: RawRepresentable, Equatable {
```
VojtaStavik, line 15: I get the idea but I don't understand why it's `class` 🤔

VojtaStavik (approved):
Many great changes! It's hard to review this in detail but in general, it looks like a giant leap forward. Let's merge it (even with the WIP commits) and finish it step by step on `main`.

Sources/StreamChatUI/Composer/ComposerVC.swift
```diff
@@ -0,0 +1,645 @@
...
+                .map(\.localURL)
+                .compactMap(AttachmentEnvelope.init)
+        case .none:
+            return []
+        }
+    }
+
+    public func setDelegate(_ delegate: ComposerVCDelegate) {
+        self.delegate = delegate
+    }
```
skorepak, line 229: What are the rules to use the `setDelegate` function or just set `var delegate` from outside?

Sources/StreamChatUI/Composer/ComposerVC.swift
```diff
@@ -0,0 +1,645 @@
...
+            handler: { [weak self] _ in
+                guard let self = self else { return }
+                self.present(self.imagePickerVC, animated: true, completion: nil)
+            }
+        )
+        let cancelAction = UIAlertAction(title: L10n.Composer.Picker.cancel, style: .cancel)
+
+        [showFilePickerAction, showImagePickerAction, cancelAction].forEach {
+            actionSheet.addAction($0)
+        }
```
skorepak, line 387: This was a computed variable before, but now it's initialised every time the user tap to add an attachment, even when the `actionSheet` is not needed anymore, please change it back to save some time, opening of the image or file picker is slow enough 😅

skorepak (approved):
One small comment, not a merge blocker.
LGTM 👍🏻

Sources/StreamChatUI/Composer/ComposerVC.swift
nuno-vieira: Yes makes sense, will log here 👍

Sources/StreamChatUI/Composer/ComposerVC.swift
nuno-vieira: It doesn't need to be class yes, forgot to remove it after doing some experiments 😄 Will change it now 👍

Sources/StreamChatUI/Composer/ComposerVC.swift
nuno-vieira: There are no rules yet I think, this function was just to keep consistency with LLC 🤔 But I guess we should opt for one way or another, in the whole codebase. CC @VojtaStavik

skorepak: Yes, please 🙌🏻

Sources/StreamChatUI/Composer/ComposerVC.swift
nuno-vieira: Why would it save time if it was a computed variable? 😄 That is just syntax sugar. A computed variable is not a lazy variable, it will always return a new `UIAlertController`. But what you are saying maybe makes sense, we probably can reuse the `UIAlertController`, but for that, we should have a lazy ref just like we have with the other view controllers we have in this component.

skorepak: A computed variable is computed at the time it's accessed, and not computed if it's not accessed.

Sources/StreamChatUI/CommonViews/InputChatMessageView/InputChatMessageView.swift
```diff
@@ -96,12 +115,35 @@ open class _InputChatMessageView<ExtraData: ExtraDataTypes>: _View, ComponentsPr
...
 
-    public func setSlashCommandViews(hidden: Bool) {
-        commandLabelView.isHidden = hidden
-        clearButton.isHidden = hidden
+    override open func updateContent() {
+        super.updateContent()
+
+        guard let content = self.content else { return }
+
+        if let quotingMessage = content.quotingMessage {
```
DominikBucher12, line 127: super nit: Why not quoted message?

DominikBucher12 (approved):
LGTM!

Sources/StreamChatUI/CommonViews/InputChatMessageView/InputChatMessageView.swift
nuno-vieira: The reason is that the message is not yet quoted, we are "quoting" or "replying" at the moment 😄 But maybe I'm wrong, naming is always subjective 😛

PR #1083, opened 2021-05-18
[CIS-844] Expose attachment counts on `ChatMessage`

evsaev:
**This PR**:
- exposes attachment counts on `ChatMessage`
- hides type-erased attachments on `ChatMessage`

### DIFF
--- a/Sources/StreamChat/Database/DTOs/MessageDTO.swift
+++ b/Sources/StreamChat/Database/DTOs/MessageDTO.swift
@@ -196,6 +196,49 @@ class MessageDTO: NSManagedObject {
         request.fetchOffset = offset
         return try! context.fetch(request)
     }
+
+    static func loadAttachmentCounts(
+        for messageId: MessageId,
+        context: NSManagedObjectContext
+    ) -> [AttachmentType: Int] {
+        enum AttachmentScoreKey: String {
+            case type
+            case count
+        }
+
+        let count = NSExpressionDescription()
+        count.name = AttachmentScoreKey.count.rawValue
+        count.expressionResultType = .integer64AttributeType
+        count.expression = NSExpression(
+            forFunction: "count:",
+            arguments: [NSExpression(forKeyPath: AttachmentScoreKey.type.rawValue)]
+        )
+
+        let request = NSFetchRequest<NSDictionary>(entityName: AttachmentDTO.entityName)
+        request.propertiesToFetch = [AttachmentScoreKey.type.rawValue, count]
+        request.propertiesToGroupBy = [AttachmentScoreKey.type.rawValue]
+        request.resultType = .dictionaryResultType
+        request.predicate = NSPredicate(
+            format: "%K.%K == %@",
+            #keyPath(AttachmentDTO.message),
+            #keyPath(MessageDTO.id),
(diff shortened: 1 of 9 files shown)


Sources/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -196,6 +196,48 @@ class MessageDTO: NSManagedObject {
...
+            format: "%K.%K == %@",
+            #keyPath(AttachmentDTO.message),
+            #keyPath(MessageDTO.id),
+            messageId
+        )
+
+        guard let results = try? context.fetch(request) else {
+            log.error("Failed to fetch attachment counts for the message with id: \(messageId)")
+            return [:]
+        }
```
VojtaStavik, line 231: Maybe it's better to capture the error and put it into `log.error`?

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -172,13 +172,15 @@ open class _ChatMessageListVC<ExtraData: ExtraDataTypes>:
     }
 
     open func attachmentViewInjectorClassForMessage(at indexPath: IndexPath) -> _AttachmentViewInjector<ExtraData>.Type? {
-        if messageForIndexPath(indexPath).imageAttachments.isEmpty == false {
+        let attachmentCounts = messageForIndexPath(indexPath).attachmentCounts
+
+        if attachmentCounts[.image, default: 0] > 0 {
             return components.galleryAttachmentInjector
-        } else if messageForIndexPath(indexPath).giphyAttachments.isEmpty == false {
+        } else if attachmentCounts[.giphy, default: 0] > 0 {
```
VojtaStavik, line 179: hmm... can we just make sure only the non-empty keys are there? Then we could do `if attachmentCounts.keys.contains(.giphy)` which is a bit nicer.

VojtaStavik (approved):
Nice 👍 Just some code ergonomy nit

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
evsaev: We can do it right away. Decided to check for the value instead of key existence to prevent errors if `attachmentCounts` start containing zeros one time

VojtaStavik: That's what I meant by "can we just make sure only the non-empty keys are there" :) Since LLC creates the value we own the rules there.

evsaev: ✅ Done. We already make sure we don't have `type: 0` counts [in here](https://github.com/GetStream/stream-chat-swift/blob/7d6b70519951cd71606066663a94667b004fb4bd/Sources/StreamChat/Database/DTOs/MessageDTO_Tests.swift#L473)

Sources/StreamChat/Database/DTOs/MessageDTO.swift
evsaev: ✅ Done

PR #1084, opened 2021-05-18
Polish UI for beta release 4.0 (vol.1)

nuno-vieira:
## UI Bug fixes
- [x] Remove text label from Giphy message content view

- [x] Fix bubble background color of link attachments view

- [x] Fix Thread Arrow View height surpassing small bubble view

## Channel config not respected fixes
- [x] Typing events
- [x] Attachments upload
- [x] Reactions
- [x] Thread Reply


Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
```diff
@@ -377,7 +377,6 @@ open class _ChatMessageContentView<ExtraData: ExtraDataTypes>: _View, ThemeProvi
             let content = content,
             content.imageAttachments.isEmpty,
             content.giphyAttachments.isEmpty,
-            content.linkAttachments.isEmpty,
             !content.linkAttachments.isEmpty {
             bubbleView?.backgroundColor = appearance.colorPalette.highlightedAccentBackground1
```
VojtaStavik, line 381: I think it's safe to remove this logic from here and rather implement changing the background color in the link attachment injector. It will also solve the edge case with multiple attachments with different types.

Sources/StreamChatUI/Composer/ComposerVC.swift
```diff
@@ -405,6 +409,10 @@ open class _ComposerVC<ExtraData: ExtraDataTypes>: _ViewController,
         Animate {
             let leadingContainerSubviews = self.composerView.leadingContainer.subviews
             for subview in leadingContainerSubviews where subview !== self.composerView.shrinkInputButton {
+                let channelConfig = self.channelController.channel?.config
+                if subview is AttachmentButton, channelConfig?.uploadsEnabled == false {
+                    continue
+                }
```
VojtaStavik, line 415: Comparing to `AttachmentButton` will fail for custom UIButton subclasses. We should probably check if it's the actual button instance for the attachment button.

Sources/StreamChatUI/ChatMessageList/Reactions/ChatMessageReactionsBubbleView.swift
```diff
@@ -35,6 +35,11 @@ open class _ChatMessageReactionsBubbleView<ExtraData: ExtraDataTypes>: _View, Th
     // MARK: - Life Cycle
 
     override open func updateContent() {
+        guard content?.channel.config.reactionsEnabled == true else {
+            isHidden = true
+            return
+        }
+
```
VojtaStavik, line 42: I believe a better place to handle this is inside `_ChatMessageLayoutOptionsResolver` and just not return the `reactions` layout option.

VojtaStavik (approved):
LGTM 👍 Feel free to merge this after fixing the comments and continue in another PR. Many of the missing things will be easily solvable by changing the logic inside `ChatMessageLayoutOptionsResolver`

Sources/StreamChatUI/Composer/ComposerVC.swift
nuno-vieira: Yes you are totally right 🤦‍♂️

Sources/StreamChatUI/ChatMessageList/Attachments/ChatMessageInteractiveAttachmentView.swift
```diff
@@ -69,16 +58,12 @@ open class _ChatMessageInteractiveAttachmentView<ExtraData: ExtraDataTypes>: _Vi
             preview.topAnchor.pin(equalTo: topAnchor),
             preview.heightAnchor.pin(equalTo: preview.widthAnchor),
             
-            titleLabel.topAnchor.pin(equalToSystemSpacingBelow: preview.bottomAnchor, multiplier: 1),
-            titleLabel.leadingAnchor.pin(equalTo: leadingAnchor),
-            titleLabel.trailingAnchor.pin(equalTo: trailingAnchor),
-            
-            separator.topAnchor.pin(equalToSystemSpacingBelow: titleLabel.bottomAnchor, multiplier: 1),
+            separator.topAnchor.pin(equalTo: preview.bottomAnchor, constant: 0),
```
evsaev, line 61:
```suggestion
            separator.topAnchor.pin(equalTo: preview.bottomAnchor),
```

Sources/StreamChatUI/ChatMessageList/Attachments/ChatMessageInteractiveAttachmentView.swift
```diff
@@ -69,16 +58,12 @@ open class _ChatMessageInteractiveAttachmentView<ExtraData: ExtraDataTypes>: _Vi
...
             separator.leadingAnchor.pin(equalTo: leadingAnchor),
             separator.trailingAnchor.pin(equalTo: trailingAnchor),
             separator.heightAnchor.pin(equalToConstant: 1),
             
-            actionsStackView.topAnchor.pin(equalTo: separator.bottomAnchor),
+            actionsStackView.topAnchor.pin(equalTo: separator.bottomAnchor, constant: 0),
```
evsaev, line 66:
```suggestion
            actionsStackView.topAnchor.pin(equalTo: separator.bottomAnchor),
```

Sources/StreamChatUI/MessageActionsPopup/ChatMessageActionsVC.swift
```diff
@@ -80,14 +80,23 @@ open class _ChatMessageActionsVC<ExtraData: ExtraDataTypes>: _ViewController, Th
...
             var actions: [ChatMessageActionItem] = [
-                inlineReplyActionItem(),
-                threadReplyActionItem(),
-                copyActionItem()
+                inlineReplyActionItem()
             ]
 
+            if channelConfig?.repliesEnabled == true {
+                actions.append(threadReplyActionItem())
+            }
```
evsaev, line 96: Shouldn't we also disable inline replies if `repliesEnabled == false`? 🤔 The docs on `repliesEnabled` say: `Enables message threads and replies. Enabled by default.`

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
nuno-vieira: Btw here I just fixed, a typo. `content.linkAttachments.isEmpty && !content.linkAttachments.isEmpty` would never work. Do you think we should move all this logic to each injector:

Sources/StreamChatUI/ChatMessageList/Reactions/ChatMessageReactionsBubbleView.swift
nuno-vieira: @VojtaStavik @evsaev It does not seem to work. Even with hardcoding the options to always not include the reactions. The reactions are always rendered. Let's discuss on slack

Sources/StreamChatUI/MessageActionsPopup/ChatMessageActionsVC.swift
nuno-vieira: I've updated the documentation

Sources/StreamChatUI/Composer/ComposerVC.swift
```diff
@@ -403,11 +407,20 @@ open class _ComposerVC<ExtraData: ExtraDataTypes>: _ViewController,
...
+            }
+            let showView = { (view: UIView) in view.isHidden = false }
+
+            self.composerView.leadingContainer.subviews
+                .filter(isNotShrinkInputButton)
+                .forEach(showView)
+
+            if self.channelController.channel?.config.uploadsEnabled == false {
+                self.composerView.attachmentButton.isHidden = true
+            }
```
VojtaStavik, line 423: There's some dark magic happening here :) I must say I don't think that making the closures named helps the readability 🤔 The "normal" style is shorter and IMO clear enough:

```swift
self.composerView.leadingContainer.subviews
    .filter { $0 !== composerView.srinkInputButton }
    .forEach { $0.isHidden = false }
```

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -303,9 +303,10 @@ open class _ChatMessageListVC<ExtraData: ExtraDataTypes>:
             cid: channelController.cid!,
             messageId: message.id
         )
-
+        
         let actionsController = _ChatMessageActionsVC<ExtraData>()
         actionsController.messageController = messageController
+        actionsController.channelController = channelController
```
VojtaStavik, line 309: We can't do this. LLC controllers can have just one delegate, we shouldn't be passing them around just like that :) Why not pass just the config?

VojtaStavik (approved):
LGTM, a couple of comments but looking good to be merged after those are fixed 👍

Sources/StreamChatUI/Composer/ComposerVC.swift
nuno-vieira: Yes I totally exaggerated on this one 😄

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
nuno-vieira: I see, I thought that it would make sense for the VC's to have controllers, and interact with them to get what they need, but putting only the config also works fine, and is actually easier 👍

DominikBucher12 (approved):
🚀

PR #1086, opened 2021-05-19
[CIS-855] Flat components structure

evsaev:
**This PR** moves to root level:
- message list components
- message composer components
- current user components
- navigation components


Sources/StreamChatUI/Components.swift
```diff
@@ -82,7 +82,54 @@ public struct _Components<ExtraData: ExtraDataTypes> {
...
+
+    /// The label that shows channel's title (the result of `ChatChannelNamer`).
+    public var channelTitleLabel: UILabel.Type = UILabel.self
+
+    /// The label that shows a subtitle of channel, typically shows if users are typing or last message,
+    /// see `typingMemberOrLastMessageString` in `ChatChannelListItemView`.
+    public var channelSubtitleLabel: UILabel.Type = UILabel.self
+
+    /// The label that shows the timestamp of the latest channel message.
+    public var channelTimestampLabel: UILabel.Type = UILabel.self
```
evsaev: Just moved it from `Component.ChannelList` but do we need to expose these ones? If we do I'd consider injection via `ChatChannelListItemView`

Sources/StreamChatUI/Components.swift
```diff
@@ -130,8 +130,69 @@ public struct _Components<ExtraData: ExtraDataTypes> {
     public var channelReadStatusView: ChatChannelReadStatusCheckmarkView.Type =
         ChatChannelReadStatusCheckmarkView.self
 
+    // MARK: - Message composer components
+
+    /// The view controller used to compose a message.
+    public var messageComposerVC: _ComposerVC<ExtraData>.Type =
+        _ComposerVC<ExtraData>.self
```
evsaev, line 204: It seems to me a good idea to have VC suffixes for controllers in `Component` as well. 
@nuno-vieira @VojtaStavik WDYT?

Sources/StreamChatUI/CommonViews/AvatarView/CurrentChatUserAvatarView.swift
```diff
@@ -29,7 +29,6 @@ open class _CurrentChatUserAvatarView<ExtraData: ExtraDataTypes>: _Control, Them
     
     /// The view that shows the current user's avatar.
     open private(set) lazy var avatarView: ChatAvatarView = components
-        .currentUser
```
evsaev, line 32: This updates the `CurrentChatUserAvatarView` to use the common `components.avatarView`. If we want the avatar in `CurrentChatUserAvatarView` be customizable separately I'll create a new dedicated item

Sources/StreamChatUI/Components.swift
```diff
@@ -198,8 +198,21 @@ public struct _Components<ExtraData: ExtraDataTypes> {
     public var currentUserAvatarView: _CurrentChatUserAvatarView<ExtraData>.Type =
         _CurrentChatUserAvatarView<ExtraData>.self
 
+    // MARK: - Navigation
+
+    /// The navigation bar.
+    public var navigationBar: ChatNavigationBar.Type = ChatNavigationBar.self
```
evsaev, line 271: This component is not used by default in UI SDK and customers have to deal with it creating the navigation controller with our navigation bar on their side (the code-snippet is taken from `Sample` app):

```swift 
let navigationController = UINavigationController(
 navigationBarClass: channelList.components.navigationBar.self,
 toolbarClass: nil
)
```
We might consider having a navigation controller subclass with the `components.navigationBar` set.

Sources/StreamChatUI/Components.swift
nuno-vieira: Yes that is the aim 👍

Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
```diff
@@ -31,54 +31,38 @@ open class _ChatChannelListItemView<ExtraData: ExtraDataTypes>: _View, ThemeProv
...
         .withAdjustingFontForContentSizeCategory
         .withBidirectionalLanguagesSupport
     
     /// The `UILabel` instance showing the time of the last sent message.
-    open private(set) lazy var timestampLabel: UILabel = components
-        .channelList
-        .itemSubviews
-        .timestampLabel
-        .init()
+    open private(set) lazy var timestampLabel: UILabel = UILabel()
```
nuno-vieira, line 46: All these generic UILabel´s should not come from `Components`. If the customer wants to customise it, simply overrides the `ChatChannelListItemView`.

Sources/StreamChatUI/Components.swift
```diff
@@ -67,27 +67,205 @@ public struct _Components<ExtraData: ExtraDataTypes> {
...
+    // MARK: - Channel list components
+
+    /// The logic to generate a name for the given channel.
+    public var channelNamer: ChatChannelNamer<ExtraData> = DefaultChatChannelNamer()
+
+    /// The collection view of the channel list.
+    public var channelListCollectionView: UICollectionView.Type = UICollectionView.self
+
+    /// The collection view layout of the channel list.
+    public var channelListLayout: UICollectionViewLayout.Type = ListCollectionViewLayout.self
```
nuno-vieira, line 161: Maybe we also don't need these to be in the components struct. If the customer wants to override the channelList, it can override there and change the UICollectionViewLayout and UICollectionView 🤔 The idea, is that really only the main shared components go into this config. Most likely, the channelListLayout won't be shared with other components.

MessengerClone/ChatClient.swift
```diff
@@ -16,7 +16,7 @@ extension ChatClient {
         components.navigation.channelListRouter = MessengerChatChannelListRouter.self
         appearance.images.pencil = UIImage(systemName: "square.and.pencil")!
         
-        components.channelList.itemView = ChatChannelListItemView.SwiftUIWrapper<MessengerChatChannelListItem>.self
+        components.channelList.channelListItemView = ChatChannelListItemView.SwiftUIWrapper<MessengerChatChannelListItem>.self
```
nuno-vieira, line 19: This is probably not compiling right ? 🤔 channelList does not exist anymore

nuno-vieira (commented):
Looks really good overall, just some small comments 👍

Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
evsaev: That's true. Already fixed 👍

MessengerClone/ChatClient.swift
evsaev: We can also put back CI actions that check the sample apps are buildable

VojtaStavik: Let's not worry about the clone app for now please.

Sources/StreamChatUI/Components.swift
```diff
@@ -82,7 +82,41 @@ public struct _Components<ExtraData: ExtraDataTypes> {
...
     public var filesAttachmentInjector: _FilesAttachmentViewInjector<ExtraData>.Type = _FilesAttachmentViewInjector<ExtraData>.self
 
-    public var channelList = ChannelList()
+    // MARK: - Channel list components
+
+    /// The logic to generate a name for the given channel.
+    public var channelNamer: ChatChannelNamer<ExtraData> = DefaultChatChannelNamer()
+
+    /// The collection view of the channel list.
+    public var channelListCollectionView: UICollectionView.Type = UICollectionView.self
```
VojtaStavik, line 158: This is not a custom type nor it's reused.

Sources/StreamChatUI/Components.swift
VojtaStavik: Yep, I like the idea of exposing this as a `UINavigationController` subclass 👍

Sources/StreamChatUI/Components.swift
```diff
@@ -67,21 +67,88 @@ public struct _Components<ExtraData: ExtraDataTypes> {
...
+
+    /// The view that shows a single reaction.
+    public var reactionItemView: _ChatMessageReactionsView<ExtraData>.ItemView.Type =
+        _ChatMessageReactionsView<ExtraData>.ItemView.self
+
+    /// The view that shows author avatar in `messageContentView`.
+    public var messageAuthorAvatarView: ChatAvatarView.Type = ChatAvatarView.self
+
+    /// The view that shows thread participant avatar in `messageContentView`.
+    public var messageThreadParticipantAvatarView: ChatAvatarView.Type = ChatAvatarView.self
```
VojtaStavik, line 116: This breaks the #1 rule of "UI Config Rules Re-Think":
> 1. The `UIConfig` inner structure should be flat and it should contain each view type just once.

VojtaStavik (approved):
LGTM, just some minor things 👍

Sources/StreamChatUI/Components.swift
evsaev: ✅ Removed

Sources/StreamChatUI/Components.swift
evsaev: ✅ Used `components.avatarView` instead

Sources/StreamChatUI/Components.swift
evsaev: ✅ Introduced in 78bb645e1ffea4adb4d80f0892ab03fc24ea50f8

Sources/StreamChatUI/Components.swift
evsaev: The layout is out custom type so it should stay probably. The collection view for the channel list is removed ✅

PR #1087, opened 2021-05-19
[CIS-856] Router improvements

VojtaStavik:
The goal of this task was to take a look at the routers and find if this is a pattern we'd like to use. I'd say, **yes**!

- I cleaned up the implementation a bit and add missing documentation
- Routers are now part of the responder chain so we can access `components` from them

Not big changes yet. This will be properly tested during the audit.


VojtaStavik:
Would be nice to merge https://github.com/GetStream/stream-chat-swift/pull/1086 first

Sources/StreamChatUI/Navigation/NavigationRouter.swift
```diff
@@ -0,0 +1,34 @@
...
+    public unowned var rootViewController: Controller
+
+    /// A convenience method to get the navigation controller of the root view controller.
+    public var rootNavigationController: UINavigationController? {
+        rootViewController.navigationController
+    }
+
+    override open var next: UIResponder? {
+        rootViewController.next
+    }
```
evsaev, line 25: booooo, that is very nice trick 👍

VojtaStavik: Not sure how it will work IRL though :)

Sources/StreamChatUI/Navigation/ChatChannelListRouter.swift
```diff
@@ -0,0 +1,41 @@
...
+open class _ChatChannelListRouter<ExtraData: ExtraDataTypes>:
+    NavigationRouter<_ChatChannelListVC<ExtraData>>,
+    ComponentsProvider
+{
+    /// Presents the view controller with the profile of the current user.
+    open func showCurrentUserProfile() {
+        debugPrint("Not implemented", #function)
+    }
+
+    /// Presents the view controller with messages for the provided cid.
```
evsaev, line 21: nit: `Presents..` wording confused me here 🤔 Can we maybe say `show` to follow the method naming patter?

evsaev (approved):
Looks awesome 🎉 Really like the changes 👍

Sources/StreamChatUI/Navigation/ChatChannelListRouter.swift
VojtaStavik: Rebased to apply changes from #1086 and fixed

Sources/StreamChatUI/Navigation/ChatChannelListRouter.swift
```diff
@@ -0,0 +1,41 @@
...
+            return
+        }
+        
+        navController.show(vc, sender: self)
+    }
+
+    /// Presents the user with the new channel creation flow.
+    open func showCreateNewChannelFlow() {
+        debugPrint("Not implemented", #function)
+    }
```
b-onc, line 40: still not sure if we should have this or not :D

b-onc (approved):
LGTM

Sources/StreamChatUI/Navigation/NavigationRouter.swift
nuno-vieira: What is the intention with this? It does not seem where are using it anywhere in the PR 🤔

nuno-vieira (approved):
LGTM ✅ Just a small comment to understand why we need the Router to confirm to the responder chain, and adding a comment in the code why it can be useful 👍

Sources/StreamChatUI/Navigation/NavigationRouter.swift
```diff
@@ -0,0 +1,34 @@
...
+/// A root class for all routes in the SDK.
+///
+/// Router objects are used to handle navigation between view controllers.
+///
+/// - Important: ⚠️ The lifetime of a router must be similar to the lifetime of its `rootViewController`. Only the
+/// root view controller is allowed to hold a strong reference to its router object.
+///
+open class NavigationRouter<Controller: UIViewController>: UIResponder {
+    /// The root `UIViewController` object of this router.
+    public unowned var rootViewController: Controller
```
nuno-vieira, line 16: Btw why do we need to know the concrete type of the rootViewController? 🤔 That just makes it dangerous that the Router can call actions on the view controller, and also limits the router in a way that it can only be launched by the provided controller, where if it was just `UIViewController`, it can be launched from everywhere, even from a custom view controller provided by the customer. 

And couldn't `NavigationRouter`just be a protocol? 🤔

Sources/StreamChatUI/Navigation/NavigationRouter.swift
VojtaStavik: It's because we want to have access to `Components` and you need to be part of the responder chain for that. I'll added a comment 👍

Sources/StreamChatUI/Navigation/NavigationRouter.swift
VojtaStavik: > Btw why do we need to know the concrete type of the rootViewController?
I was asking the same question and at the moment I don't have a good answer for that. I wasn't sure if it's a good idea to remove it or not. Let me think of it a bit more, and I'll remove it if I don't find a good reason for keeping it.

> And couldn't NavigationRouterjust be a protocol? 🤔
It could be. However, it's easier to have it like this because you want to make sure `rootViewController` is referenced weakly.

Sources/StreamChatUI/Navigation/ChatChannelListRouter.swift
VojtaStavik: Yeah, let's just log the error that this is not handled 👍

Sources/StreamChatUI/Navigation/NavigationRouter.swift
VojtaStavik: @nuno-vieira , so it is needed and it's quite handy. It allows us to reason about the type of the root VC. For example, in `showMessageList` we get the current client instance from the root VC. So let's keep it 👍

PR #1090, opened 2021-05-20
[CIS-835] Introduce channel freeze/unfreeze

b-onc:
This is actually partial channel update, but we don't have the bandwidth to figure it out completely now. It's in our backlog and will be implemented fully soon.


Sources/StreamChat/APIClient/Endpoints/ChannelEndpoints.swift
```diff
@@ -177,4 +177,14 @@ extension Endpoint {
...
+    
+    static func freezeChannel(_ freeze: Bool, cid: ChannelId) -> Endpoint<EmptyResponse> {
+        .init(
+            path: "channels/" + cid.apiPath,
+            method: .patch,
+            queryItems: nil,
+            requiresConnectionId: false,
+            body: ["set": ["frozen": freeze]]
+        )
+    }
```
VojtaStavik, line 189: This is actually a very nice and elegant workaround 👍

VojtaStavik (approved):
Looks great 👍 Thank you

PR #1091, opened 2021-05-20
Polish UI for beta release 4.0 (vol.2)

nuno-vieira:
# Description

Fixes more UI bugs and UI Logic issues for beta 4.0:
- [x]  Fix wrong color for current user messages

- [x]  Fix placeholder image when an image is missing

- [x]  Own quote message design off

- [x]  When you reply inside a thread, only the first message is sent to the thread.

https://user-images.githubusercontent.com/12814114/118994055-eccf6f00-b97d-11eb-8590-2c4969a7735d.mov


Sources/StreamChatUI/Appearance+ColorPalette.swift
```diff
@@ -31,6 +31,7 @@ public extension Appearance {
         public var background3: UIColor = .streamOverlay
         public var background4: UIColor = .streamOverlayDark
         public var background5: UIColor = .streamOverlayDarkStatic
+        public var background6: UIColor = .streamGrayWhisper
```
VojtaStavik, line 34: Screen Shot 2021-05-20 at 14 12 19

Sources/StreamChatUI/CommonViews/QuotedChatMessageView/QuotedChatMessageView.swift
```diff
@@ -83,6 +83,7 @@ open class _QuotedChatMessageView<ExtraData: ExtraDataTypes>: _View, ThemeProvid
         contentContainerView.layer.borderWidth = 1
         contentContainerView.layer.borderColor = appearance.colorPalette.border.cgColor
         contentContainerView.layer.masksToBounds = true
+        contentContainerView.backgroundColor = appearance.colorPalette.background
```
VojtaStavik, line 86: If I understand this correctly the rule is that the quoted messages has always the other color than the outer bubble. So this probably can’t be easily solved inside the view itself but should be set from the outside. it’s basically a reversed logic than for the main bubble color

VojtaStavik (approved):

Sources/StreamChatUI/CommonViews/QuotedChatMessageView/QuotedChatMessageView.swift
```diff
@@ -124,6 +124,10 @@ open class _QuotedChatMessageView<ExtraData: ExtraDataTypes>: _View, ThemeProvid
         guard let message = content?.message else { return }
         guard let avatarAlignment = content?.avatarAlignment else { return }
 
+        contentContainerView.backgroundColor = message.linkAttachments.isEmpty
+            ? appearance.colorPalette.background
+            : appearance.colorPalette.highlightedAccentBackground1
+
```
VojtaStavik, line 130: This should ideally come from the injector we don't have such a thing here 🤔 Let's leave it as is and this about a better solution later...

VojtaStavik (approved):
👍

Sources/StreamChatUI/CommonViews/QuotedChatMessageView/QuotedChatMessageView.swift
nuno-vieira: The `QuotedChatMessageView` is independent of the injectors since it can be used outside of it (Composer). For me, it makes sense that the color is defined here since the background color is impacted by the inner content of the quoted message, not by where the quoted message is (Although this completely possible to do as well, if the customer's logic is different).

PR #1093, opened 2021-05-20
Polish UI for beta release 4.0 (vol.3)

nuno-vieira:

UI Fixes for Thread related issues:

- [x] Fix long press inside thread should not show thread reply

- [x] Fix missing Avatar placeholder from Thread Arrow

- [x] Fix thread reply showing 0 when a message is show_in_channel

- [x] Fix going to an empty thread when clicking thread reply from the child

https://user-images.githubusercontent.com/12814114/119031855-75143b00-b9a3-11eb-90f2-cce14964f34c.mov

### DIFF
--- a/Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
+++ b/Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
@@ -400,7 +400,7 @@ open class _ChatMessageContentView<ExtraData: ExtraDataTypes>: _View, ThemeProvi
 
         // Thread info
         threadReplyCountButton?.setTitleColor(tintColor, for: .normal)
-        if let replyCount = content?.replyCount {
+        if let replyCount = content?.replyCount, replyCount > 0 {
             threadReplyCountButton?.setTitle(L10n.Message.Threads.count(replyCount), for: .normal)
         } else {
             threadReplyCountButton?.setTitle(L10n.Message.Threads.reply, for: .normal)
--- a/Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
+++ b/Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
@@ -406,7 +406,10 @@ open class _ChatMessageContentView<ExtraData: ExtraDataTypes>: _View, ThemeProvi
             threadReplyCountButton?.setTitle(L10n.Message.Threads.reply, for: .normal)
         }
         let latestReplyAuthorAvatar = content?.latestReplies.first?.author.imageURL
-        threadAvatarView?.imageView.loadImage(from: latestReplyAuthorAvatar)
+        threadAvatarView?.imageView.loadImage(
+            from: latestReplyAuthorAvatar,
+            placeholder: appearance.images.userAvatarPlaceholder4
+        )
 
         // Reactions view
         reactionsBubbleView?.tailDirection = content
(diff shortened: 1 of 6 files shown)


Sources/StreamChatUI/Utils/ChatMessage+Extensions.swift
```diff
@@ -27,13 +27,17 @@ extension _ChatMessage {
...
         let isThreadStart = replyCount > 0
         let isThreadReplyInChannel = showReplyInChannel
 
         return isThreadStart || isThreadReplyInChannel
     }
 
+    var isThreadChild: Bool {
+        parentMessageId != nil
+    }
+
```
VojtaStavik, line 40: I'd suggest a little bit different naming:
- `isRootOfThread` -> the root message of the thread
- `isPartOfThread` -> all messages that are a part of thread

(btw - this extension should be public and needs documentation)

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
```diff
@@ -400,7 +400,7 @@ open class _ChatMessageContentView<ExtraData: ExtraDataTypes>: _View, ThemeProvi
 
         // Thread info
         threadReplyCountButton?.setTitleColor(tintColor, for: .normal)
-        if let replyCount = content?.replyCount {
+        if let replyCount = content?.replyCount, replyCount > 0 {
```
VojtaStavik, line 403: let's confirm this in Slack

Sources/StreamChatUI/ChatMessageList/ChatMessageListRouter.swift
```diff
@@ -59,7 +59,7 @@ open class _ChatMessageListRouter<ExtraData: ExtraDataTypes>:
         threadVC.channelController = client.channelController(for: channel.cid)
         threadVC.messageController = client.messageController(
             cid: channel.cid,
-            messageId: message.id
+            messageId: message.parentMessageId ?? message.id
```
VojtaStavik, line 62: I don't think we should fallback to `message.id`. Calling this when `message.parentMessageId == nil` is a programmer error.

VojtaStavik (approved):

Sources/StreamChatUI/ChatMessageList/ChatMessageListRouter.swift
nuno-vieira: Its not actually a programmer error. This is valid logic, since if theres no parentId it means the message is the root if the thread, so we need to pass its own id. But I see that it is hidden a bit and is not clear. So the best way is to provide a comment or do this a bit differently, maybe pass the id only on the argument 🤔

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
nuno-vieira: I think it is confirmed 🤔

Sources/StreamChatUI/ChatMessageList/ChatMessageListRouter.swift
VojtaStavik: aaah, I got it now! Thanks for the explanation. A comment in the code is enough for this 👍

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
VojtaStavik: Yep, let's keep this implementation until the new designs are ready 👍

VojtaStavik (approved):

Sources/StreamChatUI/ChatMessageList/ChatMessageListRouter.swift
nuno-vieira: ✅

Sources/StreamChatUI/Utils/ChatMessage+Extensions.swift
nuno-vieira: ✅

PR #1095, opened 2021-05-21
Polish UI for beta release 4.0 (vol.4)

nuno-vieira:
Fixes threads and delete relates issues.

- [x]  Link message doesn't render properly on Threads
- [x]  Image message does not render properly on Threads
- [x]  Delete message is not working because the alert is not shown
- [x]  Deleted message has wrong text, should have "Message deleted"

### DIFF
--- a/Sources/StreamChatUI/ChatMessageList/ChatThreadVC.swift
+++ b/Sources/StreamChatUI/ChatMessageList/ChatThreadVC.swift
@@ -114,7 +114,7 @@ open class _ChatThreadVC<ExtraData: ExtraDataTypes>:
         messageComposerVC.view.translatesAutoresizingMaskIntoConstraints = false
         addChildViewController(messageComposerVC, targetView: view)
 
-        addHeaderMessage()
+        addThreadRootMessageHeader()
 
         messageComposerVC.view.topAnchor.pin(equalTo: collectionView.bottomAnchor).isActive = true
         messageComposerVC.view.leadingAnchor.pin(equalTo: view.safeAreaLayoutGuide.leadingAnchor).isActive = true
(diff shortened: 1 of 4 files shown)


Sources/StreamChatUI/ChatMessageList/ChatThreadVC.swift
```diff
@@ -428,3 +442,38 @@ open class _ChatThreadVC<ExtraData: ExtraDataTypes>:
         print(#function, indexPath)
     }
 }
+
+extension _ChatThreadVC {
```
VojtaStavik, line 446: This can't be in an extension because then it's no overridable.

Sources/StreamChatUI/ChatMessageList/ChatThreadVC.swift
```diff
@@ -428,3 +442,38 @@ open class _ChatThreadVC<ExtraData: ExtraDataTypes>:
...
+            return components.galleryAttachmentInjector
+        } else if attachmentCounts.keys.contains(.giphy) {
+            return components.giphyAttachmentInjector
+        } else if attachmentCounts.keys.contains(.file) {
+            return components.filesAttachmentInjector
+        } else if attachmentCounts.keys.contains(.linkPreview) {
+            return components.linkAttachmentInjector
+        } else {
+            return nil
+        }
```
VojtaStavik, line 464: Can we move this into a similar component to `ChatMessageLayoutOptionsResolver` and share it between the two VCs?

VojtaStavik (changes requested):

Sources/StreamChatUI/ChatMessageList/ChatThreadVC.swift
nuno-vieira: There's a lot of duplicate logic right now in both ChatThreadVC and ChatMessageList can we do this refactor in this ticket? https://stream-io.atlassian.net/secure/RapidBoard.jspa?rapidView=52&projectKey=CIS&selectedIssue=CIS-860

VojtaStavik (approved):

PR #1097, opened 2021-05-21
Add safer updates to collection view changes

nuno-vieira:
Fixes potential crashes when there are collection view changes. Checks if there are any conflicting changes, if there is, reloads the collection data, instead of performing batch updates.

### DIFF
--- a/Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
+++ b/Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
@@ -233,31 +233,70 @@ extension _ChatChannelListVC: _ChatChannelListControllerDelegate {
         _ controller: _ChatChannelListController<ExtraData>,
         didChangeChannels changes: [ListChange<_ChatChannel<ExtraData>>]
     ) {
-        let newChannelsCount = controller.channels.count
-        var movedItems: [IndexPath] = []
-        
+        var allIndexes = Set<IndexPath>()
+        var moveIndexes = Set<IndexPathMove>()
+        var insertIndexes = Set<IndexPath>()
+        var removeIndexes = Set<IndexPath>()
+        var updateIndexes = Set<IndexPath>()
+
+        // The verification of conflicts is just a temporary solution
+        // until the root cause of the conflicts has been solved.
+        var hasConflicts = false
+        let verifyConflict = { (indexPath: IndexPath) in
+            let (inserted, _) = allIndexes.insert(indexPath)
+            hasConflicts = !inserted || hasConflicts
+        }
+
+        for change in changes {
+            if hasConflicts {
+                break
+            }
+
+            switch change {
+            case let .insert(_, index):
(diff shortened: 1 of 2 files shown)


Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
```diff
@@ -233,31 +233,65 @@ extension _ChatChannelListVC: _ChatChannelListControllerDelegate {
...
+        // If doing an union of all IndexPath changes and the count is less than before
+        // It means there are duplicates, and so we have conflicting updates
+        let hasConflictingUpdates = mergedItems.count != numberOfUpdates
+
+        if hasConflictingUpdates {
+            collectionView.reloadData()
+
+            channelsCount = controller.channels.count
+            return
+        }
```
VojtaStavik, line 278: Can you move all this to a separate method/extension? Also, if the updates are conflicting updates, can we do something like:
```
log.error("⚠️ Inconsistent updates from ChatChannelListController. Please copy the following log and open an issue at https://github.com/GetStream/stream-chat-swift/issues")
+ pretty print the changes
```

VojtaStavik (approved):
This is so far the most sophisticated approach to this issue. Let's use it to gather as much info about the issue as possible.

olejnjak (approved):
🤯

nuno-vieira:
> 🤯

@olejnjak Can you please re-review again? 😄 I think now is more readable and easier to understand and also more optimized. But this is a temporary solution anyways :)

Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
```diff
@@ -233,31 +233,70 @@ extension _ChatChannelListVC: _ChatChannelListControllerDelegate {
...
+                removeIndexes.insert(index)
+            case let .update(_, index):
+                verifyConflict(index)
+                updateIndexes.insert(index)
+            }
+        }
+
+        if hasConflicts {
+            collectionView.reloadData()
+            channelsCount = controller.channels.count
```
VojtaStavik, line 274: this has to be the other way around - first update the count, then reload the view

VojtaStavik (approved):

olejnjak:
> > 🤯
> 
> @olejnjak Can you please re-review again? 😄 I think now is more readable and easier to understand and also more optimized. But this is a temporary solution anyways :)

That was more like it is insane that something like this is necessary 🤦 so more because of the cause, not the solution 🙂

olejnjak (approved):

nuno-vieira:
> > > 🤯
> > 
> > 
> > @olejnjak Can you please re-review again? 😄 I think now is more readable and easier to understand and also more optimized. But this is a temporary solution anyways :)
> 
> That was more like it is insane that something like this is necessary 🤦 so more because of the cause, not the solution 🙂

Yes it is said :/ Thank you 🙂

PR #1099, opened 2021-05-21
Remove tabs from attachments migration guide

evsaev:
**This PR** removes tabs from the attachments migration guide


VojtaStavik (approved):
Thank you!

PR #1103, opened 2021-05-21
Fix composer crash when creating new chat

nuno-vieira:
The composer would crash because when creating a new chat, we don't have a channel controller


b-onc (approved):
I did the exact same thing 2 days ago, should've PR'd :D This makes sense, we should be able to show composer without a specific controller

VojtaStavik (approved):

PR #1105, opened 2021-05-21
Add basic draft for UI customization guide

VojtaStavik:
... it's really that simple 🎉


b-onc (approved):
We should add a "color guide" so it's apparent which color is used where
:shipit:

PR #1110, opened 2021-05-25
[CIS-997] Fix over fetching previous messages

nuno-vieira:
When getting the previous messages, we would make unnecessary requests. Now we only load the previous messages if there's more to fetch.

The only thing left to do is add test cases. I will do it once the implementation approach is approved.


b-onc (commented):
I think this approach is good 👍

Sources/StreamChat/Workers/ChannelUpdater.swift
```diff
@@ -17,19 +17,27 @@ class ChannelUpdater<ExtraData: ExtraDataTypes>: Worker {
...
                 let payload = try result.get()
                 channelCreatedCallback?(payload.channel.cid)
-                self.database.write { (session) in
+                self.database.write { session in
+                    try session.saveChannel(payload: payload)
+                    completion?(.success(payload))
+                }
+                self.database.write { session in
                     try session.saveChannel(payload: payload)
                 } completion: { error in
```
evsaev, line 32: why do we save channel twice? 🤔

Sources/StreamChat/Workers/ChannelUpdater.swift
```diff
@@ -17,19 +17,27 @@ class ChannelUpdater<ExtraData: ExtraDataTypes>: Worker {
...
         apiClient.request(endpoint: .channel(query: channelQuery)) { (result) in
             do {
                 let payload = try result.get()
                 channelCreatedCallback?(payload.channel.cid)
-                self.database.write { (session) in
+                self.database.write { session in
+                    try session.saveChannel(payload: payload)
+                    completion?(.success(payload))
+                }
+                self.database.write { session in
```
evsaev, line 30: why do we save channel twice? 🤔

nuno-vieira: This seems wrong, nice catch 👍 Will fix it

nuno-vieira: @evsaev Done ✅

evsaev (approved):
Looks good 👍

PR #1111, opened 2021-05-25
[CIS-789,CIS-790] Use custom keys for image caching and Load image thumbnails correctly

skorepak:
## In this PR
- I created image CDN protocol to filter out query parameters from image URL and add thumbnail size parameters to the URL
- used the provider to fix an issue with constantly reloading images 

## Links
https://stream-io.atlassian.net/browse/CIS-789
https://stream-io.atlassian.net/browse/CIS-790

### DIFF
--- a/Sources/StreamChatUI/Components.swift
+++ b/Sources/StreamChatUI/Components.swift
@@ -216,7 +216,7 @@ public struct _Components<ExtraData: ExtraDataTypes> {
     public var messageComposerFileAttachmentView: FileAttachmentView.Type = FileAttachmentView.self
 
     /// A view that displays the image attachment.
-    public var messageComposerImageAttachmentView: ImageAttachmentView.Type = ImageAttachmentView.self
+    public var messageComposerImageAttachmentView: _ImageAttachmentView<ExtraData>.Type = _ImageAttachmentView<ExtraData>.self
 
     /// A view controller that shows suggestions of commands or mentions.
     public var suggestionsVC: _ChatSuggestionsViewController<ExtraData>.Type =
--- a/Sources/StreamChatUI/Components.swift
+++ b/Sources/StreamChatUI/Components.swift
@@ -70,6 +70,9 @@ public struct _Components<ExtraData: ExtraDataTypes> {
     
     /// The view that shows a loading indicator.
     public var loadingIndicator: ChatLoadingIndicator.Type = ChatLoadingIndicator.self
+    
+    /// Object with set of function for handling images from CDN
+    public var imageCDN: ImageCDN = StreamImageCDN()
 
     // MARK: - Message list components
 
--- a/Sources/StreamChatUI/Utils/ImageCDN.swift
+++ b/Sources/StreamChatUI/Utils/ImageCDN.swift
@@ -1,0 +1,68 @@
+//
+// Copyright © 2021 Stream.io Inc. All rights reserved.
+//
+
+import UIKit
+
+/// ImageCDN is providing set of functions to improve handling of images from CDN.
+public protocol ImageCDN {
+    /// Customised (filtered) key for image cache.
+    /// - Parameter imageURL: URL of the image that should be customised (filtered).
+    /// - Returns: String to be used as an image cache key.
+    func cachingKey(forImage url: URL) -> String
+    
+    /// Enhance image URL with size parameters to get thumbnail
+    /// - Parameters:
+    ///   - originalURL: URL of the image to get the thumbnail for.
+    ///   - preferredSize: The requested thumbnail size.
+    ///
+    /// Use view size in points for `preferredSize`, point to pixel ratio (scale) of the device is applied inside of this function.
+    func thumbnailURL(originalURL: URL, preferredSize: CGSize) -> URL
+}
+
+public struct StreamImageCDN: ImageCDN {
+    public static var streamCDNURL = "stream-io-cdn.com"
+    
+    public func cachingKey(forImage url: URL) -> String {
+        let key = url.absoluteString
+        
+        guard
+            var components = URLComponents(url: url, resolvingAgainstBaseURL: true),
+            let host = components.host,
+            host.contains(StreamImageCDN.streamCDNURL)
+        else { return key }
+        
+        // Keep these parameters in the cache key as they determine the image size.
+        let persistedParameters = ["w", "h"]
+        
+        let newParameters = components.queryItems?.filter { persistedParameters.contains($0.name) } ?? []
+        components.queryItems = newParameters.isEmpty ? nil : newParameters
+        return components.string ?? key
+    }
+    
+    public func thumbnailURL(originalURL: URL, preferredSize: CGSize) -> URL {
+        guard
+            var components = URLComponents(url: originalURL, resolvingAgainstBaseURL: true),
+            let host = components.host,
+            host.contains(StreamImageCDN.streamCDNURL)
+        else { return originalURL }
+
+        let scale = UIScreen.main.scale
+        components.queryItems = components.queryItems ?? []
+        components.queryItems?.append(contentsOf: [
+            URLQueryItem(name: "w", value: String(format: "%.0f", preferredSize.width * scale)),
+            URLQueryItem(name: "h", value: String(format: "%.0f", preferredSize.height * scale)),
+            URLQueryItem(name: "crop", value: "center"),
+            URLQueryItem(name: "resize", value: "fill"),
+            URLQueryItem(name: "ro", value: "0") // Required parameter.
+        ])
+        return components.url ?? originalURL
+    }
+}
+
+public extension CGSize {
+    /// Maximum size of avatar used in the UI.
+    ///
+    /// It's better to use single size of avatar thumbnail to utilise the cache.
+    static var avatarThumbnailSize: CGSize { CGSize(width: 40, height: 40) }
+}
--- a/Sources/StreamChatUI/Utils/ImageCDN_Tests.swift
+++ b/Sources/StreamChatUI/Utils/ImageCDN_Tests.swift
@@ -1,0 +1,102 @@
+//
+// Copyright © 2021 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+import StreamChat
+@testable import StreamChatUI
+import XCTest
+
+class ImageCDN_Tests: XCTestCase {
+    func test_cache_validStreamURL_filtered() {
+        let provider = StreamImageCDN()
+        
+        let url = URL(string: "https://wwww.stream-io-cdn.com/image.jpg?name=Luke&father=Anakin")!
+        let filteredUrl = "https://wwww.stream-io-cdn.com/image.jpg"
+        let key = provider.cachingKey(forImage: url)
+        
+        XCTAssertEqual(key, filteredUrl)
+    }
+    
+    func test_cache_validStreamUrl_withSizeParameters() {
+        let provider = StreamImageCDN()
+        
+        let url = URL(string: "https://wwww.stream-io-cdn.com/image.jpg?name=Luke&w=128&h=128&crop=center&resize=fill&ro=0")!
+        let filteredUrl = "https://wwww.stream-io-cdn.com/image.jpg?w=128&h=128"
+        let key = provider.cachingKey(forImage: url)
+        
+        XCTAssertEqual(key, filteredUrl)
+    }
+    
+    func test_cache_validStreamURL_unchanged() {
+        let provider = StreamImageCDN()
+        
+        let url = URL(string: "https://wwww.stream-io-cdn.com/image.jpg")!
+        let key = provider.cachingKey(forImage: url)
+        
+        XCTAssertEqual(key, url.absoluteString)
+    }
+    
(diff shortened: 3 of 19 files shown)


Sources/StreamChatUI/Utils/ImageCDN.swift
```diff
@@ -0,0 +1,39 @@
+//
+// Copyright © 2021 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+public var cachingKeyProvider: CachingKeyProvider = StreamCachingKeyProvider()
```
VojtaStavik, line 7: I'd love to see how far can we get if we used this inside `Compomnents`

Sources/StreamChatUI/Utils/ImageCDN.swift
```diff
@@ -0,0 +1,39 @@
...
+
+/// Caching key provider is providing custimised (filtered) key to be used for caching.
+///
+/// The URL can contain extra parameters that are changing every time. Those parameters are effectively blocking the caching.
+/// By providing customised (filtered) key, the cache will work as expected.
+public protocol CachingKeyProvider {
```
VojtaStavik, line 13: Maybe `ImageCachingKeyProvider`?

Sources/StreamChatUI/Utils/ImageCDN.swift
```diff
@@ -0,0 +1,39 @@
...
+        
+        // Stream CDN
+        if host.contains(StreamCachingKeyProvider.streamCDNURL) {
+            components.query = nil
+            return components.string ?? key
+        }
+        
+        return key
+    }
+}
```
VojtaStavik, line 39: This probably doesn't have to be `open` as it's more an implementation detail of the default CDN.

VojtaStavik (commented):
Looks like the correct approach 👍

Sources/StreamChatUI/Utils/ImageCDN.swift
skorepak: Working on it 👍🏻

Sources/StreamChatUI/Utils/ImageCDN.swift
```diff
@@ -0,0 +1,37 @@
...
+/// By providing customised (filtered) key, the cache will work as expected.
+public protocol CachingKeyProvider {
+    /// Customised (filtered) key for image cache.
+    /// - Parameter imageURL: URL of the image that should be customised (filtered).
+    /// - Returns: String to be used as an image cache key.
+    func cachingKey(forImage url: URL) -> String
+}
+
+open class StreamCachingKeyProvider: CachingKeyProvider {
+    public static var streamCDNURL = "stream-io-cdn.com"
```
DominikBucher12, line 21: nit: `public static let`?? Why is it not constant?

Sources/StreamChatUI/Utils/ImageCDN.swift
```diff
@@ -0,0 +1,39 @@
...
+    
+    public func cachingKey(forImage url: URL) -> String {
```
evsaev, line 23: I'd step out from inheritance and make the `StreamCachingKeyProvider` a struct. We have a protocol and customers will build their own implementation of `CachingKeyProvider` from scratch anyway

skorepak: You are right, good catch!

Sources/StreamChatUI/Utils/ImageCDN.swift
skorepak: Tried multiple ways but I'm failing to pull it out of components without ExtraData on `UIImageView` the only way I can see right now is to subclass 
```
class StreamImageView<ExtraData: ExtraDataTypes>: UIImageVIew {}
```
and add `StreamImageView` to components ass well

skorepak: I will open another PR with implementing the `UIImageView` subclass so we can decide if we want it or not. this PR can come trough

skorepak:
Added function to load thumbnail of preferred size and renamed protocol to ImageCDN. 
I'm wondering whether to limit the thumbnail URL just for our Image CDN. Anybody nows?

Sources/StreamChatUI/Utils/ImageCDN.swift
```diff
@@ -0,0 +1,74 @@
...
+
+public enum ImageResize: String {
+    case crop, scale, fill
+}
+
+/// Caching key provider is providing customised (filtered) key to be used for caching.
+///
+/// The URL can contain extra parameters that are changing every time. Those parameters are effectively blocking the caching.
+/// By providing customised (filtered) key, the cache will work as expected.
+public protocol ImageCDN {
```
b-onc, line 19: This works more like a URL provider than CDN, so I suggest renaming this to `ImageURLProvider` or something like it, basically removing `CDN` from the name

Sources/StreamChatUI/Utils/ImageCDN.swift
```diff
@@ -0,0 +1,74 @@
...
+    /// - Returns: String to be used as an image cache key.
+    func cachingKey(forImage url: URL) -> String
+    
+    /// Enhance image URL with size parameters to get thumbnail
+    /// - Parameters:
+    ///   - originalURL: URL of the image to gwt the thumbnail for.
+    ///   - preferedSize: The requested thumbnail size.
+    ///   - crop: Crop type to be used for the thumbnail.
+    ///   - resize: Resize type to be used for the thumbnail.
+    func thumbnailURL(originalURL: URL, preferedSize: CGSize, crop: ImageCrop, resize: ImageResize) -> URL
```
b-onc, line 31: This func is not actually being used?

Sources/StreamChatUI/Utils/ImageCDN.swift
```diff
@@ -0,0 +1,74 @@
...
+
+import UIKit
+
+public enum ImageCrop: String {
+    case top, bottom, left, right, center
+}
+
+public enum ImageResize: String {
+    case crop, scale, fill
+}
```
b-onc, line 13: These should not public since they're params for Stream CDN. Custom CDN's can have their own params

Sources/StreamChatUI/Utils/ImageCDN.swift
skorepak: The problem is that it's not providing URL, is providing URL modifiers/processors, so I was struggling with the name a little. I used the one suggested by @VojtaStavik in the task

Sources/StreamChatUI/Utils/ImageCDN.swift
skorepak: No, not yet. I’m trying to figure out how/when to use it. We don't want to use it every time, but only when the thumbnail is required. Which I don't know yet.

Sources/StreamChatUI/Utils/ImageCDN.swift
skorepak: It needs to be public because it's used in public protocol.

VojtaStavik: Hmm, do you think we need to expose this to users? I'd be ok with hardcoding this to `fill + center` 🤔

Sources/StreamChatUI/Utils/ImageCDN.swift
VojtaStavik: For sure `ImageCDN` is not the best name but I think it kinda works 🤷‍♂️ You'd need to implement these methods if you use your own CDN for images. We'll probably add more methods here as we add new features.

Sources/StreamChatUI/Components.swift
```diff
@@ -70,6 +70,9 @@ public struct _Components<ExtraData: ExtraDataTypes> {
     
     /// The view that shows a loading indicator.
     public var loadingIndicator: ChatLoadingIndicator.Type = ChatLoadingIndicator.self
+    
+    /// Object providing custom key to use for image cache.
+    public var imageCDN: ImageCDN = StreamImageCDN()
```
VojtaStavik, line 75: outdated description

Sources/StreamChatUI/Utils/ImageCDN.swift
```diff
@@ -0,0 +1,74 @@
...
+        if host.contains(StreamImageCDN.streamCDNURL) {
+            components.query = nil
+            return components.string ?? key
+        }
+        
+        return key
+    }
+    
+    public func thumbnailURL(originalURL: URL, preferedSize: CGSize, crop: ImageCrop, resize: ImageResize) -> URL {
+        guard var components = URLComponents(url: originalURL, resolvingAgainstBaseURL: true) else { return originalURL }
```
VojtaStavik, line 61: Here we also need to check that the URL is a Stream CDN URL

skorepak: Yes, already fixed that locally 👍🏻

Sources/StreamChatUI/Utils/ImageCDN.swift
skorepak: It's being used now

Sources/StreamChatUI/Utils/UIImageView+Extensions.swift
```diff
@@ -3,14 +3,18 @@
 //
 
 import Nuke
+import StreamChat
 import UIKit
 
 extension UIImageView {
     @discardableResult
-    func loadImage(
+    func loadImage<ExtraData: ExtraDataTypes>(
```
VojtaStavik, line 11: I think it's about time to add documentation to this method :)

Sources/StreamChatUI/Utils/UIImageView+Extensions.swift
```diff
@@ -3,14 +3,18 @@
...
         from url: URL?,
         placeholder: UIImage? = nil,
         resizeAutomatically: Bool = true,
+        requestThumbnail: Bool = false,
```
VojtaStavik, line 15: TBH this API now feels a bit complex. In the end, it all comes to the question - do you need the image in a specific size? If yes, you should try to get the correct URL and get it.

Maybe there's a way to simplify this?

VojtaStavik (approved):
LGTM, just a few comments to the API design 👍

Sources/StreamChatUI/Utils/UIImageView+Extensions.swift
skorepak: I was thinking about some enum with associated value like:
```
enum Resize {
    .resizeAfterDownload
    .requestThumbnail(size: CGSize?)
    .none
}
```
But you didn't like adding an extra option for ImageCrop and this would be the same. Adding extra public enum just for this. I would say it's a bit extensive, but not that complex. 

I'm open to suggestions on how to simplify this

Sources/StreamChatUI/Utils/ImageCDN_Tests.swift
```diff
@@ -0,0 +1,100 @@
...
+    func test_thumbnail_validStreamUrl_withParameters() {
+        let provider = StreamImageCDN()
+        
+        let url = URL(string: "https://wwww.stream-io-cdn.com/image.jpg?name=Luke")!
+        let thumbnailUrl =
+            URL(string: "https://wwww.stream-io-cdn.com/image.jpg?name=Luke&w=120&h=120&crop=center&resize=fill&ro=0")!
+        let processedURL = provider.thumbnailURL(
+            originalURL: url,
+            preferredSize: CGSize(width: 40, height: 40)
+        )
```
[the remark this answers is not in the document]
skorepak: That's true! thanks for noticing 👍🏻

Sources/StreamChatUI/Utils/UIImageView+Extensions.swift
```diff
@@ -79,3 +79,36 @@ private extension UIImageView {
...
+
+extension ImageProcessors {
+    public struct LateResize: ImageProcessing {
+        private var size: CGSize {
+            var size: CGSize = .zero
+            DispatchQueue.main.sync { size = sizeProvider() }
+            return size
+        }
+
+        private let sizeProvider: () -> CGSize
```
VojtaStavik, line 91: Maybe this should return `CGSize?` rather than using `.zero` as the fallback?

Sources/StreamChatUI/Utils/UIImageView+Extensions.swift
```diff
@@ -79,3 +79,36 @@ private extension UIImageView {
...
+                size: size,
+                unit: .points,
+                contentMode: .aspectFill,
+                crop: true,
+                upscale: false
+            ).process(image)
+        }
+
+        public var identifier: String {
+            "com.github.kean/nuke/lateResize"
```
VojtaStavik, line 111: Shouldn't we make this unique every time we create a processor?

skorepak: I was trying to figure out if it matters or not. Nuke docs don't say anything and the original `Resize` is adding just the parameters of the processor. If you create two `Resize` processors with the same size parameter, they will have the same identifier. Since we have size closure and getting the size is waiting synchronously for the main queue (may be expensive), I decided to remove it.

PR #1112, opened 2021-05-25
Audit `ChatMessageLayoutOptionsResolver`

evsaev:
**This PR**:
- exposes `isMessageLastInSequence` on `ChatMessageLayoutOptionsResolver`
- adds missing tests to `ChatMessageLayoutOptionsResolver` type

### DIFF
--- a/Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageLayoutOptionsResolver_Tests.swift
+++ b/Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageLayoutOptionsResolver_Tests.swift
@@ -1,0 +1,1169 @@
+//
+// Copyright © 2021 Stream.io Inc. All rights reserved.
+//
+
+@testable import StreamChat
+import StreamChatTestTools
+@testable import StreamChatUI
+import XCTest
+
+final class ChatMessageLayoutOptionsResolver_Tests: XCTestCase {
+    private var optionsResolver: ChatMessageLayoutOptionsResolver!
+
+    // MARK: - Setup
+
+    override func setUp() {
+        super.setUp()
+
+        optionsResolver = .init()
+    }
+
+    override func tearDown() {
+        super.tearDown()
+
+        optionsResolver = nil
+    }
+
+    // MARK: - Flipped
+
+    func test_optionsForMessage_whenMessageIsSentByCurrentUser_includesFlipped() {
+        // Create a message sent NOT by the current user
+        let messageSentByCurrentUser: ChatMessage = .mock(
+            id: .unique,
+            text: .unique,
+            author: .mock(id: .unique),
+            isSentByCurrentUser: true
+        )
+
+        // Calculate layout options for the message
+        let layoutOptions = optionsResolver.optionsForMessage(
+            at: .init(item: 0, section: 0),
+            in: .mock(cid: .unique),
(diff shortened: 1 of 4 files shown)


Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageLayoutOptionsResolver_Tests.swift
```diff
@@ -0,0 +1,336 @@
...
+                    && !channel.isDirectMessageChannel
+            }
+        }
+    }
+
+    func test_quotedMessage() {
+        // Iterate channel history and check `.quotedMessage` is there if conditions are met
+        assertOptionIsIncluded(.quotedMessage) { message, _ in
+            message.quotedMessage != nil && message.deletedAt == nil
+        }
```
nuno-vieira, line 113: I don't think this is the best way to test this. Tests should be as dummy as possible and should not contain any logic, and with this, we are not really testing that "quotedMessage" added the `.quotedMessage` option. For example, if there was a bug, and `isSentByCurrentUser` option

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageLayoutOptionsResolver_Tests.swift
```diff
@@ -0,0 +1,336 @@
...
+            )
+
+            // Assert the `option` is there or not depending on the requirement
+            XCTAssertEqual(
+                layoutOptions.contains(option),
+                mustBeIncludedFor(message, isStandaloneOrLastInSequence)
+            )
+        }
+    }
+}
```
nuno-vieira, line 194: Like I said above, with this helper, is not possible to check the exact message which added the option, which will not catch possible future bugs

Sources/StreamChatUI/Utils/ChatMessage+Extensions.swift
```diff
@@ -28,10 +28,7 @@ extension _ChatMessage {
 
     /// A boolean value that checks if the message is the root of a thread.
     var isRootOfThread: Bool {
-        let isThreadStart = replyCount > 0
-        let isThreadReplyInChannel = showReplyInChannel
-
-        return isThreadStart || isThreadReplyInChannel
+        replyCount > 0
```
nuno-vieira, line 31: 👍

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageLayoutOptionsResolver_Tests.swift
evsaev: Yeah, I see what you mean. Makes sense 👍 Let me have one more iteration on this

evsaev: ✅ Updated

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageLayoutOptionsResolver_Tests.swift
evsaev: ✅ Removed

nuno-vieira (approved):
LGTM

PR #1114, opened 2021-05-26
Open `ChatMessage` extensions, add missing tests

evsaev:
**This PR**:
- adds `ChatMessage.isDeleted` helper
- fixes `ChatMessage.isRootOfThread` helper
- adds missing tests for `ChatMessage` extensions
- renames fields to follow naming convention

### DIFF
--- a/Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
+++ b/Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
@@ -382,7 +382,7 @@ open class _ChatMessageContentView<ExtraData: ExtraDataTypes>: _View, ThemeProvi
         }
 
         // Metadata
-        onlyVisibleForYouContainer?.isVisible = content?.onlyVisibleForCurrentUser ?? false
+        onlyVisibleForYouContainer?.isVisible = content?.isOnlyVisibleForCurrentUser == true
 
         authorNameLabel?.isVisible = layoutOptions?.contains(.authorName) == true
         authorNameLabel?.text = content?.author.name
--- a/Tests/StreamChatUITests/ChatMessage_Tests.swift
+++ b/Tests/StreamChatUITests/ChatMessage_Tests.swift
@@ -125,9 +125,9 @@ final class ChatMessage_Tests: XCTestCase {
         }
     }
 
-    // MARK: - lastActionFailed
+    // MARK: - isLastActionFailed
 
-    func test_lastActionFailed_returnsTrue_forNonDeletedMessageWithFailedLocalState() {
+    func test_isLastActionFailed_returnsTrue_forNonDeletedMessageWithFailedLocalState() {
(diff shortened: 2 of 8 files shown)


Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
```diff
@@ -382,7 +382,7 @@ open class _ChatMessageContentView<ExtraData: ExtraDataTypes>: _View, ThemeProvi
         }
 
         // Metadata
-        onlyVisibleForYouContainer?.isVisible = content?.onlyVisibleForCurrentUser ?? false
+        onlyVisibleForYouContainer?.isVisible = content?.isOnlyVisibleForCurrentUser ?? false
```
nuno-vieira, line 385:
```suggestion
        onlyVisibleForYouContainer?.isVisible = content?.isOnlyVisibleForCurrentUser == true
```
I know it is not part of this PR, but let's try to remove the `?? false` whenever we can :)

evsaev: > whenever we can

you mean for all occurrences of `ChatMessage` fields or ...? 🤔

Tests/StreamChatUITests/ChatMessage_Tests.swift
```diff
@@ -68,4 +70,311 @@ final class ChatMessage_Tests: XCTestCase {
         
         XCTAssertEqual(message.lastActiveThreadParticipant?.name, "Second")
     }
+
+    // MARK: - isInteractionEnabled
+
+    func test_isInteractionEnabled_returnsFalse_forEphemeralMessage() {
```
nuno-vieira, line 76:
```suggestion
    func test_isInteractionEnabled_whenEphemeralMessage_returnsFalse() {
```
We should write a doc for the naming test convention. But usually, the expected behaviour comes last. We should try to follow the pattern: 

##### `test_Given_When_Then`
Given -> Name of the func/prop
When -> Condition
Then -> Expected result

This is also true for the rest of the test cases in ChatMessage_Tests

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
evsaev: ✅ Replaced the `??` with `==` for `ChatMessage` fields

nuno-vieira (commented):
Looks really good overall 👍 The only point I would suggest an improvement is the test naming convention 🙂

Tests/StreamChatUITests/ChatMessage_Tests.swift
evsaev: give me 5 min 😄

evsaev: ✅ Done

nuno-vieira (approved):
🎉

PR #1118, opened 2021-05-27
Audit `ChatMessageBubbleView`

evsaev:
**This PR**:
- audits `ChatMessageBubbleView` view
- adds `ChatMessageBubbleView` to components
- adds tests for `ChatMessageBubbleView`


Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageBubbleView_Tests.swift
```diff
@@ -0,0 +1,297 @@
...
+
+    func test_appearance() {
+        // Create a bubble
+        let bubble = ChatMessageBubbleView().withFixedSize
+
+        // Assert the bubble is rendered correctly
+        AssertSnapshot(bubble, variants: .onlyUserInterfaceStyles)
+    }
+
+    func test_appearance_whenEphemeralMessageSentByCurrentUserLastInSequence() {
```
evsaev, line 21: The snapshots are not rendered correctly when `layer.maskedCorners` is set to round only some corners, not all of them 🙁 That's the key thing the message bubble does. Here is the issue: https://github.com/pointfreeco/swift-snapshot-testing/issues/358

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageBubbleView.swift
```diff
@@ -5,22 +5,69 @@
...
+        if message.isSentByCurrentUser {
+            if message.type == .ephemeral {
+                return appearance.colorPalette.background8
+            } else {
+                return appearance.colorPalette.background6
+            }
+        } else {
+            return appearance.colorPalette.background8
+        }
+    }
```
nuno-vieira, line 57: I think that for this small view there's no advantage in creating these helpers. I think writing this logic directly on the `updateContent` function is enough and simple I guess, WDYT? And if we remove the `bubble` from the background color, it just is `backgroundColor`. If the customer wants to override the backgroundColor logic, it is just question of overriding `updateContent` in this case.

Sources/StreamChatUI/Appearance+ColorPalette.swift
```diff
@@ -33,6 +33,7 @@ public extension Appearance {
         public var background5: UIColor = .streamOverlayDarkStatic
         public var background6: UIColor = .streamGrayWhisper
         public var background7: UIColor = .streamDarkGray
+        public var background8: UIColor = .streamWhite
```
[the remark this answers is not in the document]
evsaev: > I really don't like these background numbers. They don't provide any context

Agree with you and like the naming neither. AFAIR the approach with numbers was chosen because it was not possible to find a common name for a single color used in several places and step by step this became messy 👎 

Like the idea to have meaningful names but this requires some time spent on investigation and looking for suitable names. On the one hand, I don't see nothing wrong in having a color for concrete purpose (like `messageListBackground`). On there other hand, there should be some `general purpose` for those used across the entire app. The question is what is `general purpose` and what is `component specific`.

Would be nice to dedicate some effort to `ColorPalette` but that is not that urgent since component colors are easily customisable while proper names can be added in minor releases.

*Regarding bubble case: didn't use `popoverBackground` because it was create for popover but is also used in reactions/suggestions/image gallery/file list etc but there was no such `backgroundX`.

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageBubbleView.swift
```diff
@@ -5,22 +5,69 @@
 import StreamChat
 import UIKit
 
-open class ChatMessageBubbleView: _View, AppearanceProvider {
-    open var roundedCorners: CACornerMask = .all {
+/// A view that displays a bubble around a message.
+public typealias ChatMessageBubbleView = _ChatMessageBubbleView<NoExtraData>
+
+/// A view that displays a bubble around a message.
+open class _ChatMessageBubbleView<ExtraData: ExtraDataTypes>: _View, AppearanceProvider, SwiftUIRepresentable {
```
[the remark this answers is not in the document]
evsaev: I kinda like the `bubble` terminology since it's commonly used to refer to the thing drawn around the message (no matter what shape it has) 😄 

The name I was considering is `ChatMessageBackgroundView` but it has no soul 🤷

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageBubbleView.swift
evsaev: There's no difference where we to put these `if's` but when wrapped in a property it's becomes testable. 

Another reason is that a single property is a smaller customization point then the `updateAppearance`. When overriding a property you don't need to think from where it's invoked, but only about value to return. With a property there's no way the `super` is invoked at the end other than in the beginning (when `super` is at the end of `updateContent` the customisation can be discarded)

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageBubbleView.swift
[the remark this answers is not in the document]
nuno-vieira: Yes, I totally agree with you, `Bubble` is very easily recognizable. And for that reason, it is the right choice. Bubble is fine for me then 👍

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageBubbleView.swift
nuno-vieira: The `super` here can never be used at the end since the point of overreding the `updateContent` is to extend or replace the logic based on content, so if the customer would use the `super` at the end it would just be a mistake. 

In terms of testing, in this case, it is also easily testable. What we are testing here is the backgroundColor of the bubble view. So it is just a question of changing the content, calling `updateContent` and checking the backgroundColor. 

IMO it does not make sense to create a new background color property, since we already have one, we are simply changing it based on content. But the customer might want to have the background color static, and not change based on the content.

Lets maybe discuss with the team and see what everyone thinks about this 👍

evsaev: @nuno-vieira I will put it to `updateContent`. Content based properties indeed is a new pattern and can be discussed later when we are not in a hurry (this PR is a blocker for message content view audit and needs to be merged as soon as we can)

I can pitch this when we are not in a hurry 🙂

nuno-vieira: Oh okay sorry, so yes maybe better to put everything in updateContent for now before we discuss with the team to define the final API. With this, we won't introduce breaking changes 👍

evsaev: ✅ Done

nuno-vieira (approved):
LGTM

PR #1121, opened 2021-05-27
[CIS-869] Use Swiftdocs to generate SDK Documentation (UI + LLC)

DominikBucher12:
# What this PR do:
- Adds ability to generate automatically documentation with folder structure using [swift-doc](https://github.com/SwiftDocOrg/swift-doc) tool.

# Where you can start
- Asking anything about the documentation so I can fill in the details, will create notion page on how to use this together with @b-onc so we have nice steps

# How to test this PR
You can run this code inside branch with `git checkout experiment && cd stream-chat-swift-docs && yarn run start` 

# How this works:

1. Install [swift-doc](https://github.com/SwiftDocOrg/swift-doc) 
2. Choose which documentation you want to generate (`StreamChat`/`StreamChatUI`)
3. Run `bash Scripts/generateDocumentation.sh StreamChat` or  `bash Scripts/generateDocumentation.sh StreamChatUI` to generate the documentation for the given target
4. Observe what documents are generated whether there is anything missing
5. Ask @DominikBucher12 to clarify or improve something of the workflow 
6. Document <3 

Note: This PR is huge with all the generated files, just a friendly warning you might not wan review the last 2 documentation-generation commits with github UI, rather than checkout the branch and take a look at the structure.

# Open questions and suggestions so far for working with docusaurus:
- [x] Images need to be generated into Docusaurus folder and path set to be to `static/UIComponentImages/example.png`
- [x] `swift-doc` generates the documentation with `Home.md`, `Footer.md` and `Sidebar.md` for each folder, we probably need to delete Footer and dropdown Sidebar and rename the `Home.md` to `<FolderWhichContainsHome>.md` so it provides quick overView of the whole abstract component, this should be easy with simple bash script
- [x] Docusaurus generates automatically header for the documentation as well as `swift-doc`, this is problem because we generate double headers for the component. We can either delete the header from generation again with some script or adjust this inside docusaurus...


Scripts/generateDocumentation.sh
```diff
@@ -0,0 +1,31 @@
...
+scriptDir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+cd "$scriptDir/../"
+
+# As of May 2021, swift-doc doesn't support directories, so we find all the subdirectories instead
+find "$TARGET_DIRECTORY" -type d -print > Scripts/StreamChat_folder_directories.txt
+
+# Let's run the documentation inside all subdirectories in the target and directory so we can mirror the directories structure
+while read directory; do
+  swift doc generate $directory -n StreamChatUI  -o Documentation/$directory 
+done < Scripts/StreamChat_folder_directories.txt
```
DominikBucher12, line 24: nit: can be actually /tmp 😅  will polish later

Scripts/generateDocumentation.sh
```diff
@@ -0,0 +1,31 @@
...
+    echo "Please specify target to generate docs for (StreamChat or StreamChatUI)"
+    exit 1
+fi
+
+# Move to project root directory
+scriptDir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+cd "$scriptDir/../"
+
+# As of May 2021, swift-doc doesn't support directories, so we find all the subdirectories instead
+find "$TARGET_DIRECTORY" -type d -print > Scripts/StreamChat_folder_directories.txt
```
DominikBucher12, line 19: The nitty problem with this approach is we need somehow to figure out the emissions which it creates (iterates through assets folders and Generated folder, which we don't need)

Documentation/Sources/StreamChatUI/ChatChannelList/_ChatChannelListItemView.md
```diff
@@ -170,3 +170,5 @@ override open func setUpLayout()
 ``` swift
 override open func updateContent() 
 ```
+
+![ChatChannelListItemView](Tests/StreamChatUITests/UISDKdocumentation/__Snapshots__/ChatChannelListItemView_Documentation_Tests/ChannelListItemView_documentation.default-light.png)
```
DominikBucher12, line 174: ok so as it turns out relative path doesn't work as I expected 😅

Scripts/generateDocumentation.sh
DominikBucher12: Fixed.

Scripts/addImagesToDocumentation.sh
```diff
@@ -0,0 +1,24 @@
...
+PATH_TO_SNAPSHOTS="Tests/StreamChatUITests/UISDKdocumentation/__Snapshots__"
+
+for UI_SNAPSHOT in ${PATH_TO_SNAPSHOTS}/*;do
+    STRIPPED_PATH=`basename $UI_SNAPSHOT`
+    COMPONENT_NAME=${STRIPPED_PATH%_*_*}
+    
+    DOCUMENTATION_FILE=`find Documentation/Sources/StreamChatUI -name "_$COMPONENT_NAME.md"`
+    # Let's use just light variation of the screenshot, we can support dark mode later.
+    FINAL_SNAPSHOT=`ls $UI_SNAPSHOT | grep light`
+    tail -1 "$DOCUMENTATION_FILE" | grep "$FINAL_SNAPSHOT"
```
nuno-vieira, line 16: I actually think there's no point in generating the dark version 🤔 Or does swift-docs support light/dark mode? Because the point of this is to only verify the component hierarchy, and light/dark mode doesn't influence anything on the hierarchy. So my suggestion is for generating the hierarchy snapshot only for light mode.

DominikBucher12: Hmm I was thinking that we might wanna support dark mode in our documentation, because you know how handy this comes. But there is problem with markdown and I am not quite sure how we can handle this in docusaurus, this needs some research as well... I will check if we can support dark/light mode with docusaurus and try to figure out what to do with this...

nuno-vieira: I think docussarus supports dark/Light mode. But even if it does, I think change the image to dark or light is not really the point of this I guess? We can just put a static background colour of the component hierarchy? 🤔 Lets discuss this with the team on Monday

Scripts/generateDocumentation.sh
```diff
@@ -0,0 +1,34 @@
...
+fi
+
+IGNORED_PATHS='Resources\|__Snapshots__\|Tests\|Generated'
+
+# As of May 2021, swift-doc doesn't support directories, so we find all the subdirectories instead
+find "$TARGET_DIRECTORY" -type d -print | grep -v -e $IGNORED_PATHS > /tmp/StreamChat_folder_directories.txt
+
+# Let's run the documentation inside all subdirectories in the target and directory so we can mirror the directories structure
+while read directory; do
+  swift doc generate $directory -n StreamChatUI  -o Documentation/$directory
```
DominikBucher12, line 26: `-n StreamChatUI` 😱 oh no, anyway this will be fixed in next PR for making compatible with docusaurus: https://stream-io.atlassian.net/browse/CIS-875?atlOrigin=eyJpIjoiNTM3MTQ3ZThiNjhiNDc4YWEwOWFjMTZmNDE2YmQxZWEiLCJwIjoiaiJ9 You can run this code inside branch with `git checkout experiment && cd stream-chat-swift-docs && yarn run start`

b-onc (approved):
:shipit:

PR #1124, opened 2021-05-28
Introduce uploading videos

b-onc:
This makes video functionality in line with v2, but we should introduce video previews soon.

### DIFF
--- a/Sources/StreamChatUI/Composer/ComposerVC.swift
+++ b/Sources/StreamChatUI/Composer/ComposerVC.swift
@@ -371,15 +371,6 @@ open class _ComposerVC<ExtraData: ExtraDataTypes>: _ViewController,
             )
             actionSheet.popoverPresentationController?.sourceView = sender
 
-            let showFilePickerAction = UIAlertAction(
-                title: L10n.Composer.Picker.file,
-                style: .default,
-                handler: { [weak self] _ in
-                    guard let self = self else { return }
-                    self.selectedAttachmentType = .file
-                    self.present(self.filePickerVC, animated: true, completion: nil)
-                }
-            )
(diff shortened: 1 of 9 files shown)


Sources/StreamChatUI/Composer/ComposerVC.swift
```diff
@@ -642,14 +644,22 @@ open class _ComposerVC<ExtraData: ExtraDataTypes>: _ViewController,
...
+            if let imageURL = info[.imageURL] as? URL {
+                // User picked an image
+                let attachment = try AnyAttachmentPayload(localFileURL: imageURL, attachmentType: .image)
+                content.attachments.append(attachment)
+            } else if let videoURL = info[.mediaURL] as? URL {
+                // User picked a video
+                // TODO: after video attachments (`VideoAttachmentInjector`) is properly implemented,
+                // the `attachmentType` should be `video`
+                let attachment = try AnyAttachmentPayload(localFileURL: videoURL, attachmentType: .file)
+                content.attachments.append(attachment)
```
evsaev, line 657: Hm, that's the first step, did I get it right? 🤔

The full LLC support would include:
- `.video` attachment type
- `VideoAttachmentPayload` + an alias for attachment with video payload
- `videoAttachments` exposed on the message model

Without ⬆️ changes, the videos can only be distinguished from files by checking the `FileAttachmentPayload.file`.

 Are we gonna have a follow-up PR?

b-onc: @evsaev very good points... how much effort do you foresee the full support would be?

b-onc:
This turned out to be not-so-straightforward, so I'm closing this. It's high in our backlog and it'll be properly implemented soon.

PR #1125, opened 2021-05-28
Audit `ChatMessageErrorIndicator`

evsaev:
**This PR**:
- fixes subviews arrangement issue on `ChatMessageContentView` when the content is flipped
- audits `ChatMessageErrorIndicator`


Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
```diff
@@ -351,6 +332,59 @@ open class _ChatMessageContentView<ExtraData: ExtraDataTypes>: _View, ThemeProvi
...
+                    after: errorIndicatorContainer!
+                )
+            }
+
+            constraintsToActivate += [
+                mainContainer.leadingAnchor
+                    .pin(equalTo: leadingAnchor)
+                    .almostRequired
+            ]
+        }
```
evsaev, line 378: I've just realized that the `.flipped` option was not working properly for the case when messages from the current user are aligned to the left. When the content is flipped we should add `mainContainer` subviews in reverse order. 

ATM this is done manually for `mainContainer` and `threadInfoContainer` but i'd really love if this can be done in 1 line (via some kind of API on `ContainerStackView` which mirrors the arranged subviews)

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
```diff
@@ -278,10 +282,14 @@ open class _ChatMessageContentView<ExtraData: ExtraDataTypes>: _View, ThemeProvi
             errorIndicatorView.setContentCompressionResistancePriority(.streamRequire, for: .horizontal)
             errorIndicatorView.setContentCompressionResistancePriority(.streamRequire, for: .vertical)
 
+            let errorIndicatorContainer = createErrorIndicatorContainer()
+            errorIndicatorContainer.addSubview(errorIndicatorView)
```
VojtaStavik, line 286: This creates a new local variable `errorIndicatorContainer`. You probably wanted to assign the value to `self. errorIndicatorContainer`?

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
VojtaStavik: @evsaev This was originally part of the `ContainerStackView` and I think it was @nuno-vieira who removed it. It was causing more issues than it was helpful. Doing this manually turned out to be a better option.

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
```diff
@@ -351,6 +332,59 @@ open class _ChatMessageContentView<ExtraData: ExtraDataTypes>: _View, ThemeProvi
...
+            errorIndicatorContainer,
+            bubbleThreadMetaContainer
+        ].compactMap { $0 }
+
+        if options.contains(.flipped) {
+            mainContainer.addArrangedSubviews(mainContainerSubviews.reversed())
+
+            if let errorIndicator = errorIndicatorView {
+                mainContainer.setCustomSpacing(
+                    .init(-errorIndicator.intrinsicContentSize.width / 2),
```
VojtaStavik, line 353: What's the purpose of this? Do you want the view to be centered somehow? You can just add an additional constraint for that, no need to do this kind magic of magic imo.

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
nuno-vieira: Yes this feel like it is very specific use case of the message content view, it should be on the container IMO. From the content view should be easy to implement or not? 🤔 It is just reversing the subviews and remove them and add them afterwards

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
evsaev: Yeah, this is done in `createErrorIndicatorContainer` (if the container is `nil` we create it first and then give it to the caller)

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
evsaev: The error indicator is embedded into `mainContainer` where all the built-in constraints are `.required`. So setting the custom spacing is the easiest way.

The obvious solution with error indicator at root level with `centerX = mainContainer.leading/trailing` anchor has issues:
- for messages from current user (aligned to the right and without avatar) the error indicator is partially outside the screen
- when user avatar is shown we want error indicator to be in-between the author avatar and message content I guess ([content - error - avatar]) so the messages are aligned. Having error indicator inside `mainContainer` makes it possible + also handles the case when the cell is flipped

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
evsaev: Managing custom spacing after the error indicator introduces some complexity (when reversed the custom spacing should be set after the view that precedes the error indicator).

It's easy to implement but we have this in 2 places already. If we decide to introduce `.flippedVertically` (which makes quite popular `Slack` design doable in 1 line) the same logic will be needed also for `bubbleThreadMetaContainer`.

Let's have it as it is, we have more urgent things to work on now. Put a comment to just highlight the use-case where the `reverse` API would fit nicely 🙂

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
VojtaStavik: Ah, yeah ... sorry 👍

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
VojtaStavik: All right, thank you for explaining this to me 👍

VojtaStavik (approved):

DominikBucher12 (approved):
Awesome, 🛳🇮🇹

PR #1131, opened 2021-06-01
Update SPM Dependencies

b-onc:
Nuke 9.2 had a bug that spammed the console: https://github.com/kean/Nuke/issues/416
Resolved with 9.2.1, so we can update safely.

#skip_changelog


VojtaStavik (approved):

VojtaStavik (commented):
Maybe we can update other dependencies, too?

PR #1138, opened 2021-06-02
[CIS-880] Fix a flaky test

dmigach:
It was reported a couple of times that the test is flaky, I wasn't able to make it fail, but I think the reason is that we subscribe for things after we simulate the response. Probably it worked because usually we get to subscribe before the code we test actually execute, and in some rare cases it's faster than we get to subscribe.
So let's place response simulation below subscriptions


VojtaStavik (approved):
Nice, let's see if it helps 👍

PR #1140, opened 2021-06-02
Fix invisible images in the message composer

skorepak:
Not sure why, but Nuke sometimes fail to crop the image resulting in a full black square instead of a resized image.
I removed the `crop: true` parameter (default is `false`) and it seems to work.


VojtaStavik (approved):
Please add a CHANGELOG entry

PR #1141, opened 2021-06-02
Fix bubble appearance for recent messages

evsaev:
**This PR**:
- updates previous message bubble when the new message comes
- fixes the reload animation for message cell
- removes `metadata` layout option (the metadata container will be added automatically if any of `timestamp/authorName/onlyVisibleForYouIndicator` is included)

https://user-images.githubusercontent.com/12818985/120638853-591b9980-c479-11eb-9252-2a9755deea31.mp4


Sources/StreamChatUI/ChatMessageList/ChatMessageListCollectionView.swift
```diff
@@ -199,10 +200,39 @@ open class ChatMessageListCollectionView<ExtraData: ExtraDataTypes>: UICollectio
                 }
             }
         } completion: { flag in
+            let indexPathsToReload = self.indexPathsToMessagesBeforeInsertedAfterBatchUpdate(with: changes)
+            self.reloadItems(at: indexPathsToReload)
```
evsaev, line 204: Ideally, the reload for previous message cell happens during the same batch-updates session we insert but that would require the overall message count before updates happen -> we would have to cache the message count in VCs (both messages list and thread) as we do in channel list and pass it to the CV.

To avoid this boilerplate we reload `after` the batch-update when the # of items in section is known.

evsaev: It can also be replaced with just `reloadData` as @VojtaStavik suggested. 

It actually looks nicer since we have smth wrong with attributes for appearing/disappearing items in the layout but is less performant (not that less)

VojtaStavik (approved):

PR #1147, opened 2021-06-03
[CIS-882] Implement scroll to bottom button in the Message List

skorepak:
## In this PR I have
-  Added marking the channel as read when the last collection view cell is fully shown on the screen
- Added "scroll to bottom" button visible only when there are unread messages and the last collection view cell is not fully visible
- Added temporary fix for slow-ish WS when the button disappears with a little delay after the scroll to the bottom

https://user-images.githubusercontent.com/7560049/120669910-321c9200-c490-11eb-8037-283d4e7480a2.mp4

## Links
https://stream-io.atlassian.net/browse/CIS-882


Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -534,3 +584,20 @@ open class _ChatMessageListVC<ExtraData: ExtraDataTypes>:
         return df
     }()
 }
+
+private extension UICollectionView {
+    var isLastCellVisible: Bool { isLastCellFullyVisible }
```
skorepak, line 589: We have multiple options when the cell is visible. Any part (even 1px), fully visible or something in between. I tested it and the fully looks good. It's shown on the video in the PR description.

skorepak: Other options listed for visibility, will remove them when there are no comments on this.

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -80,6 +80,12 @@ open class _ChatMessageListVC<ExtraData: ExtraDataTypes>:
         .typingIndicatorView
         .init()
         .withoutAutoresizingMaskConstraints
+    
+    /// A button to scroll the collection view to the bottom.
+    ///
+    /// Visible when there is unread message and the collection view is not at the bottom already.
+    public private(set) lazy var scrollToBottomButton: UIButton = UIButton()
+        .withoutAutoresizingMaskConstraints
```
VojtaStavik, line 88: This should ideally come from `Components`

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -159,6 +178,11 @@ open class _ChatMessageListVC<ExtraData: ExtraDataTypes>:
         collectionView.backgroundColor = appearance.colorPalette.background
         
         navigationItem.titleView = titleView
+        
+        scrollToBottomButton.setImage(appearance.images.scrollToBottomArrow, for: .normal)
+        scrollToBottomButton.backgroundColor = appearance.colorPalette.background8
+        scrollToBottomButton.layer.addShadow(color: appearance.colorPalette.shadow)
+        scrollToBottomButton.layer.cornerRadius = 20
```
VojtaStavik, line 185: Ideally we'd move this setup code into a standalone button component

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -144,6 +156,13 @@ open class _ChatMessageListVC<ExtraData: ExtraDataTypes>:
             collectionView.contentInset.bottom += typingIndicatorViewHeight
         }
         
+        view.addSubview(scrollToBottomButton)
+        scrollToBottomButton.trailingAnchor.pin(equalTo: view.layoutMarginsGuide.trailingAnchor).isActive = true
+        scrollToBottomButton.bottomAnchor.pin(equalTo: collectionView.bottomAnchor, constant: -8).isActive = true
```
VojtaStavik, line 161: You can use one of the `greaterThanOrEqualToSystemSpacing***` methods instead of hardcoding `8`

Sources/StreamChatUI/Resources/Assets.xcassets/arrow_down.imageset/arror_down.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="14" height="8" viewBox="0 0 14 8" fill="none" xmlns="http://www.w3.org/2000/svg">
+<path d="M0.305189 0.306019C-0.102381 0.714049 -0.102381 1.37559 0.305189 1.78361L6.20904 7.694C6.43734 7.9225 6.74514 8.0231 7.04334 7.9956C7.31234 7.9969 7.58174 7.8949 7.78694 7.6897L13.6942 1.78283C14.1021 1.37504 14.1021 0.713879 13.6942 0.306099C13.2864 -0.101691 12.6252 -0.101691 12.2174 0.306099L6.99634 5.5269L1.78117 0.306019C1.37359 -0.102001 0.712769 -0.102011 0.305189 0.306019Z" fill="#005EFF"/>
+</svg>
```
VojtaStavik, line 3: please use PDF instead of SVG 👍

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -252,6 +276,20 @@ open class _ChatMessageListVC<ExtraData: ExtraDataTypes>:
...
         overlayDateFormatter.string(from: channelController.messages[indexPath.item].createdAt)
     }
+
+    public func scrollViewDidScroll(_ scrollView: UIScrollView) {
+        let lastCellIsVisible = collectionView.isLastCellVisible
+        if lastCellWasVisible != lastCellIsVisible {
+            lastCellWasVisible = lastCellIsVisible
+
+            if lastCellIsVisible, channelController.channel?.isUnread == true {
+                channelController.markRead()
```
VojtaStavik, line 286: Calling `markRead` multiple times is safe. LLC won't make an API call if not needed. So feel free to just call is as much as you want.

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -268,6 +306,16 @@ open class _ChatMessageListVC<ExtraData: ExtraDataTypes>:
         collectionView.scrollToMostRecentMessage(animated: animated)
     }
     
+    open func updateScrollToBottomButton() {
+        scrollToBottomButton.isVisible
+            = channelController.channel?.isUnread == true
+            && !collectionView.isLastCellVisible
```
VojtaStavik, line 312: Can make this animated?

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
VojtaStavik: Sounds good 👍 If you think the UX feels right let's use `isLastCellFullyVisible`

VojtaStavik (changes requested):
This is a very good start. I left a couple of comments but overall looks good 👍

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -534,3 +575,13 @@ open class _ChatMessageListVC<ExtraData: ExtraDataTypes>:
...
     }()
 }
+
+private extension UICollectionView {
+    var isLastCellVisible: Bool {
+        if numberOfItems(inSection: 0) == 0 { return true }
+        let lastIndexPath = IndexPath(item: 0, section: 0)
+
+        guard let attributes = collectionViewLayout.layoutAttributesForItem(at: lastIndexPath) else { return false }
+        return bounds.contains(attributes.frame)
```
nuno-vieira, line 585:
```suggestion
        let lastIndexPath = IndexPath(item: 0, section: 0)
        return collectionView.indexPathsForVisibleItems.contains(lastIndexPath)
```

This should be enough, I've been using this 👍

skorepak: I had this option listed before but I already removed it, sorry. 
Your suggested change will be true immediately when the cell will appear on the screen, even when only partially visible, but the code I'm using will be true only when the full cell is visible, which is what I opted for here.

nuno-vieira: Ah I see... In my case, I actually just want the cell partially visible. Maybe rename this to `isLastCellFullyVisible`? 🤔  I don't know 😄

skorepak: I had this exact name with an extra getter 
```
var isLastCellVisible: Bool { isLastCellFullyVisible }
```
And removed it to siplify the code 😅
I will think about it until tomorrow 👍🏻

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -268,6 +297,17 @@ open class _ChatMessageListVC<ExtraData: ExtraDataTypes>:
         collectionView.scrollToMostRecentMessage(animated: animated)
     }
     
+    open func updateScrollToLatestMessageButton() {
```
DominikBucher12, line 300: nit 1: this `open func` doesn't have documentation which would probably help with nit 2 😝
nit 2:  `updateScrollToLatestMessageButton` sounds like it returns `UIButton` subclass... what about naming it  `setScrollToLatestMessageButton(visible: Bool)`? It's even more flexible when some customer would want to implement custom logic for hiding/showing the button, what do you think?

Sources/StreamChatUI/ChatMessageList/ScrollToLatestMessageButton.swift
```diff
@@ -0,0 +1,17 @@
...
+
+/// A Button that is used to indicate unread messages in the Message list.
+open class ScrollToLatestMessageButton: _Button, AppearanceProvider {
+    override open func setUpAppearance() {
+        super.setUpAppearance()
+        
+        setImage(appearance.images.scrollDownArrow, for: .normal)
+        backgroundColor = appearance.colorPalette.background8
+        layer.addShadow(color: appearance.colorPalette.shadow)
+        layer.cornerRadius = 20
```
DominikBucher12, line 15: Wonder if the corner radius shouldn't be set from the outside so you don't actually need to override the whole `setupAppearance` when you change the size of the `ScrollToLatestMessageButton` via constraints or produce some strange side-effects 🤔 I would probably set this property in `setupAppearance` inside `_ChatMessageListVC` and we are good to go :)

DominikBucher12 (approved):
Just few comments <3 Nice

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -144,6 +155,13 @@ open class _ChatMessageListVC<ExtraData: ExtraDataTypes>:
             collectionView.contentInset.bottom += typingIndicatorViewHeight
         }
         
+        view.addSubview(scrollToLatestMessageButton)
+        collectionView.bottomAnchor.pin(equalToSystemSpacingBelow: scrollToLatestMessageButton.bottomAnchor).isActive = true
+        scrollToLatestMessageButton.trailingAnchor.pin(equalTo: view.layoutMarginsGuide.trailingAnchor).isActive = true
+        scrollToLatestMessageButton.widthAnchor.pin(equalToConstant: 40).isActive = true
+        scrollToLatestMessageButton.heightAnchor.pin(equalToConstant: 40).isActive = true
```
VojtaStavik, line 162: nit: if the goal is to have a circular button, it's better to do:
```swift
x.widthAnchor.pin(equalTo: x.heightAnchor)
x.widthAnchor.pin(equalToConstant: 40)
```
because users will be able to change the size of the button just by overriding a single value

Sources/StreamChatUI/ChatMessageList/ScrollToLatestMessageButton.swift
VojtaStavik, line 15: This should probably be in `layoutSubivews` and should be something like `bounds.height / 2.0`?

Sources/StreamChatUI/Resources/Assets.xcassets/arrow_down.imageset/Contents.json
```diff
@@ -0,0 +1,16 @@
...
+      "idiom" : "universal"
+    }
+  ],
+  "info" : {
+    "author" : "xcode",
+    "version" : 1
+  },
+  "properties" : {
+    "preserves-vector-representation" : true,
+    "template-rendering-intent" : "template"
```
VojtaStavik, line 14: 👍

VojtaStavik (approved):
LGTM, just a couple of nit comments. Thank you 👍

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
skorepak: This function contains logic when to show/hide, I would have to implement another function to show and hide the button. And it felt too bloated with functions for such a "simple" task. but now thinking about it, when you want to change the animation or anything, it would be a separated function would be a good place to do it 👍🏻

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -80,6 +80,14 @@ open class _ChatMessageListVC<ExtraData: ExtraDataTypes>:
         .typingIndicatorView
         .init()
         .withoutAutoresizingMaskConstraints
+    
+    /// A button to scroll the collection view to the bottom.
+    ///
+    /// Visible when there is unread message and the collection view is not at the bottom already.
+    public private(set) lazy var scrollToLatestMessageButton: UIButton = components
```
evsaev, line 87:
```suggestion
    open private(set) lazy var scrollToLatestMessageButton: UIButton = components
```

skorepak: @evsaev the `typingIndicatorView` above is also `public`, should I change it to `open` as well?

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -144,6 +155,13 @@ open class _ChatMessageListVC<ExtraData: ExtraDataTypes>:
             collectionView.contentInset.bottom += typingIndicatorViewHeight
         }
         
+        view.addSubview(scrollToLatestMessageButton)
+        collectionView.bottomAnchor.pin(equalToSystemSpacingBelow: scrollToLatestMessageButton.bottomAnchor).isActive = true
```
evsaev, line 159: nit: Can we do it vice versa? 😄 i.e. pin the bottom to the collection view? At first glance it feels like a constraint for another random view is added

skorepak: I hate it but not without hardcoding a value :( 
The button needs to be below collection view, If I would swap them, the system spacing would be in the opposite direction

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -144,6 +155,13 @@ open class _ChatMessageListVC<ExtraData: ExtraDataTypes>:
...
+        scrollToLatestMessageButton.trailingAnchor.pin(equalTo: view.layoutMarginsGuide.trailingAnchor).isActive = true
+        scrollToLatestMessageButton.widthAnchor.pin(equalTo: scrollToLatestMessageButton.heightAnchor).isActive = true
+        scrollToLatestMessageButton.heightAnchor.pin(equalToConstant: 40).isActive = true
```
evsaev, line 162: nit: WDYT about having such constants in `intrinsicContentSize` ? 🤔 It can easily be customised by setting constraints on end-user side in VC / overriding the `intrinsicContentSize / overriding the `setupLayout` in component

```suggestion
        // in ScrollToLatestMessageButton
        override public var intrinsicContentSize: CGSize { .init(width: 40, height: 40) }
```

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -252,6 +270,15 @@ open class _ChatMessageListVC<ExtraData: ExtraDataTypes>:
     ) -> String? {
         overlayDateFormatter.string(from: channelController.messages[indexPath.item].createdAt)
     }
+
+    public func scrollViewDidScroll(_ scrollView: UIScrollView) {
```
evsaev, line 274:
```suggestion
    open func scrollViewDidScroll(_ scrollView: UIScrollView) {
```

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
VojtaStavik: @evsaev it's definitely an option and it requires a bigger discussion. The general approach is that the parent view decides on the size and position of its superview. The way it's done here is in sync with how we do it elsewhere.

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
VojtaStavik: if it really bothers you, you can add `pin(equalToSystemSpacingAbove:)` and just flip the views inside. Let's please do it later during the audit of this VC.

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -534,3 +583,13 @@ open class _ChatMessageListVC<ExtraData: ExtraDataTypes>:
         return df
     }()
 }
+
+private extension UICollectionView {
+    var isLastCellFullyVisible: Bool {
+        if numberOfItems(inSection: 0) == 0 { return true }
```
evsaev, line 589: Do we need this check? 🤔 The layout returns `nil` when asking for the attributes, I guess (didn't test it)

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
VojtaStavik: ah yes, all of this should be `open`

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
skorepak: It's an interesting idea.
Pros: multiple customisation points
Cons: not consistent in our codebase, might be confusing for less experienced devs

Definitely something to consider. Not at this PR but more generally within the team

evsaev (approved):
🚢

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
skorepak: The guard on line 592 return `false`, where this check returns `true`.
A better question is: How it should behave when there are no messages (not loaded yet) in the message list? 
Considering that the message list will open scrolled to the latest message (the last cell will be visible) I'm returning `true`.

PR #1149, opened 2021-06-03
Fix previous message bubble appearance on message deletion

evsaev:
**This PR** fixes the issue when the previous message bubble is not updated when the following message is deleted

https://user-images.githubusercontent.com/12818985/120696660-7a9a7680-c4b5-11eb-82c0-0de8692bdfa7.mp4


VojtaStavik (approved):
Nice catch, thank you 🙏

skorepak (commented):
LGTM 👍🏻

PR #1150, opened 2021-06-03
Fix scrolling when new message is sent + Refactor Keyboard Observer

nuno-vieira:
- Fixes list not scrolling when the current user sends a new message
- Removes "scrollToBottomIfNeeded" since it is not used anywhere, and is not useful
- Fixes when a new message is received, and the list is scrolled to the bottom, should remain scrolled to the bottom
- Content inset is now adjusted for typing indicator view when it appears.
- Refactors Keyboard Observer Implementation:
   - Uses Apple's recommended way to get the keyboard height, to support docked keyboards.
   - If the last message is hidden by the keyboard, just check if it is visible, and scroll to bottom.

### DIFF
--- a/Sources/StreamChatUI/ChatMessageList/ChatMessageListCollectionView.swift
+++ b/Sources/StreamChatUI/ChatMessageList/ChatMessageListCollectionView.swift
@@ -18,10 +18,7 @@ public protocol ChatMessageListCollectionViewDataSource: UICollectionViewDataSou
 /// with the provided content view type and layout options.
 open class ChatMessageListCollectionView<ExtraData: ExtraDataTypes>: UICollectionView, Customizable, ComponentsProvider {
     private var identifiers: Set<String> = .init()
-    
-    open var needsToScrollToMostRecentMessage = false
-    open var needsToScrollToMostRecentMessageAnimated = false
-    
+
     /// View used to display date of currently displayed messages
     open lazy var scrollOverlayView: ChatMessageListScrollOverlayView = {
         let scrollOverlayView = components.messageListScrollOverlayView.init()
@@ -199,11 +196,25 @@ open class ChatMessageListCollectionView<ExtraData: ExtraDataTypes>: UICollectio
                 }
             }
         } completion: { flag in
+
+            if shouldScrollToBottom {
+                self.scrollToMostRecentMessage()
+            }
+
             // If a new message was inserted, reload the previous message to give it chance to update
             // its appearance in case it's now part of a group.
-            self.reloadItems(
-                at: self.indexPathsToMessagesBeforeInsertedAfterBatchUpdate(with: changes)
-            )
+            let indexPaths = self.indexPathsToMessagesBeforeInsertedAfterBatchUpdate(with: changes)
+            if indexPaths.isEmpty == false {
+                self.reloadItems(
+                    at: indexPaths
+                )
+
+                // If the previous cell was updated, we should re-adjust the scrolling again.
+                if self.isLastCellFullyVisible {
+                    self.scrollToMostRecentMessage()
+                }
+            }
+
             completion?(flag)
         }
     }
@@ -231,23 +242,8 @@ open class ChatMessageListCollectionView<ExtraData: ExtraDataTypes>: UICollectio
         }
     }
     
-    /// Will scroll to most recent message on next `updateMessages` call
-    open func setNeedsScrollToMostRecentMessage(animated: Bool = true) {
-        needsToScrollToMostRecentMessage = true
-        needsToScrollToMostRecentMessageAnimated = animated
-    }
-
-    /// Force scroll to most recent message check without waiting for `updateMessages`
-    open func scrollToMostRecentMessageIfNeeded() {
-        guard needsToScrollToMostRecentMessage else { return }
-        
-        scrollToMostRecentMessage(animated: needsToScrollToMostRecentMessageAnimated)
-    }
-
     /// Scrolls to most recent message
     open func scrollToMostRecentMessage(animated: Bool = true) {
-        needsToScrollToMostRecentMessage = false
-
         // our collection is flipped, so (0; 0) item is most recent one
         scrollToItem(at: IndexPath(item: 0, section: 0), at: .bottom, animated: animated)
     }
@@ -269,5 +265,21 @@ open class ChatMessageListCollectionView<ExtraData: ExtraDataTypes>: UICollectio
         Animate(isAnimated: animated) { [scrollOverlayView] in
             scrollOverlayView.alpha = alpha
         }
+    }
+
+    /// A Boolean that returns true if the bottom cell is fully visible.
+    /// Which is also means that the collection view is fully scrolled to the boom.
+    open var isLastCellFullyVisible: Bool {
+        if numberOfItems(inSection: 0) == 0 { return true }
+        let lastIndexPath = IndexPath(item: 0, section: 0)
+
+        guard let attributes = collectionViewLayout.layoutAttributesForItem(at: lastIndexPath) else { return false }
+        return bounds.contains(attributes.frame)
+    }
+
+    /// A Boolean that returns true if the last cell is visible, but can be just partially visible.
+    open var isLastCellVisible: Bool {
+        let lastIndexPath = IndexPath(item: 0, section: 0)
+        return indexPathsForVisibleItems.contains(lastIndexPath)
     }
 }
--- a/Sources/StreamChatUI/ChatMessageList/ChatMessageListCollectionView.swift
+++ b/Sources/StreamChatUI/ChatMessageList/ChatMessageListCollectionView.swift
@@ -182,6 +182,10 @@ open class ChatMessageListCollectionView<ExtraData: ExtraDataTypes>: UICollectio
         with changes: [ListChange<_ChatMessage<ExtraData>>],
         completion: ((Bool) -> Void)? = nil
     ) {
+        // Before committing the changes, we need to check if were scrolled
+        // to the bottom, if yes, we should stay scrolled to the bottom
+        let shouldScrollToBottom = isLastCellFullyVisible
+
         performBatchUpdates {
             for change in changes {
                 switch change {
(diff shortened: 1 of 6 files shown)


Sources/StreamChatUI/ChatMessageList/ChatMessageListCollectionView.swift
```diff
@@ -182,6 +182,13 @@ open class ChatMessageListCollectionView<ExtraData: ExtraDataTypes>: UICollectio
         with changes: [ListChange<_ChatMessage<ExtraData>>],
         completion: ((Bool) -> Void)? = nil
     ) {
+        var shouldScrollToBottom = false
+
+        let bottomEdge = contentOffset.y + frame.size.height
+        if bottomEdge >= contentSize.height {
+            shouldScrollToBottom = true
+        }
+
```
VojtaStavik, line 191: Please add a comment here explaining what this magic code does :)

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -85,6 +85,9 @@ open class _ChatMessageListVC<ExtraData: ExtraDataTypes>:
     open lazy var router = components
         .messageListRouter
         .init(rootViewController: self)
+
+    /// The height of the typing indicator view
+    open var typingIndicatorViewHeight: CGFloat = 22
```
VojtaStavik, line 90: This should be `private(set)` since the value can't be changed dynamically

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -465,7 +476,14 @@ open class _ChatMessageListVC<ExtraData: ExtraDataTypes>:
     /// Hides typing Indicator
     /// - Parameter typingMembers: typing members gotten from `channelController`
     open func hideTypingIndicator() {
+        guard !typingIndicatorView.isHidden else { return }
```
VojtaStavik, line 479: `guard typingIndicatorView.isVisible else { return }` ? 🤯

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -451,6 +450,18 @@ open class _ChatMessageListVC<ExtraData: ExtraDataTypes>:
...
+        if typingIndicatorView.isHidden {
+            Animate {
+                self.collectionView.contentInset.bottom += self.typingIndicatorViewHeight
+                self.collectionView.scrollIndicatorInsets.bottom += self.typingIndicatorViewHeight
+            }
+
+            let lastIndexPath = IndexPath(item: 0, section: 0)
+            if collectionView.indexPathsForVisibleItems.contains(lastIndexPath) {
+                scrollToMostRecentMessage()
+            }
```
VojtaStavik, line 462: Is this here on purpose? Feels more like a leftover 🤔

Sources/StreamChatUI/ChatMessageList/ChatMessageListKeyboardObserver.swift
```diff
@@ -40,58 +40,28 @@ open class ChatMessageListKeyboardObserver {
...
-            options: UIView.AnimationOptions(rawValue: curve),
+            options: UIView.AnimationOptions(rawValue: keyboardAnimationCurve),
             animations: { [weak self] in
                 self?.containerView.layoutIfNeeded()
-                if needUpdateContentOffset {
-                    self?.scrollView.contentOffset = newContentOffset
+                if shouldScrollToBottom {
+                    self?.collectionView.scrollToItem(at: lastIndexPath, at: .bottom, animated: true)
                 }
             }
```
VojtaStavik, line 66: I don't think this is enough. This works OK when the last message is visible, but when you're scrolled up, the content doesn't move when it should:

https://user-images.githubusercontent.com/6388510/120754447-f70c7400-c50c-11eb-82b3-57f4d181e847.mov

Sources/StreamChatUI/ChatMessageList/ChatMessageListCollectionView.swift
skorepak: Maybe we can move the `isLastCellFullyVisible` from private extension in https://stream-io.atlassian.net/browse/CIS-882 directly to this collection view and reuse it. 
The `isLastCellFullyVisible` function handles `contentInset` well.

VojtaStavik: @skorepak great idea 💯

Sources/StreamChatUI/ChatMessageList/ChatMessageListKeyboardObserver.swift
nuno-vieira: Ah.. I see.. Didn't notice this behaviour before, had to watch other apps to understand what you mean. Will takkle this then

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
nuno-vieira: Or maybe just `if typingIndicatorView.isHidden { return }`. The `isVisible` extension is a candidate for people forgetting to use it and making the code inconsistent, I didn't even know we had this extension 😅

Sources/StreamChatUI/ChatMessageList/ChatMessageListCollectionView.swift
nuno-vieira: Yes makes sense 👌 Probably with @skorepak extension, the code will be more clear 👍

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
nuno-vieira: It's not a leftover, the goal of these lines of code are to scroll the message list to the bottom if the indicator is over the last cell. Example:

https://user-images.githubusercontent.com/12814114/120781187-b9b1e180-c520-11eb-8936-c016cd5ac552.mp4

Sources/StreamChatUI/ChatMessageList/ChatMessageListKeyboardObserver.swift
nuno-vieira: Done ✅

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
nuno-vieira: Done ✅

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
nuno-vieira: Done ✅

Sources/StreamChatUI/ChatMessageList/ChatMessageListCollectionView.swift
nuno-vieira: Done ✅

Sources/StreamChatUI/ChatMessageList/ChatMessageListKeyboardObserver.swift
```diff
@@ -40,58 +40,29 @@ open class ChatMessageListKeyboardObserver {
...
-            options: UIView.AnimationOptions(rawValue: curve),
+            options: UIView.AnimationOptions(rawValue: keyboardAnimationCurve),
             animations: { [weak self] in
                 self?.containerView.layoutIfNeeded()
-                if needUpdateContentOffset {
-                    self?.scrollView.contentOffset = newContentOffset
+
+                let isKeyboardShowing = intersectedKeyboardHeight > 0
+                if let indexPathNearestToKeyboard = indexPathNearestToKeyboard, isKeyboardShowing {
+                    self?.collectionView.scrollToItem(at: indexPathNearestToKeyboard, at: .bottom, animated: true)
```
VojtaStavik, line 65: nit: `animated` should be `false` here so the animation speed matches the keyboard

Sources/StreamChatUI/ChatMessageList/ChatMessageListKeyboardObserver.swift
```diff
@@ -40,58 +40,29 @@ open class ChatMessageListKeyboardObserver {
...
-        // message composer follows keyboard
-        composerBottomConstraint?.constant = -(containerView.bounds.height - localFrame.minY)
+        composerBottomConstraint?.constant = -intersectedKeyboardHeight
 
-        // calculate new contentOffset for message list, so bottom message still visible when keyboard appears
-        var keyboardTop = localFrame.minY
-        if keyboardTop == containerView.bounds.height {
-            keyboardTop -= containerView.safeAreaInsets.bottom
-        }
+        let indexPathNearestToKeyboard = collectionView.indexPathsForVisibleItems.sorted().first
```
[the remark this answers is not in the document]
nuno-vieira: Ahah yes, we could play with the contentOffset.y, but that requires more calculations when the collection view is already scrolled to the bottom. This solution is simpler and IMO better, since the cells don't get cut in half. (Although on other apps, this happens)

Sources/StreamChatUI/ChatMessageList/ChatMessageListKeyboardObserver.swift
nuno-vieira: yes makes sense, Thanks 👍

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -500,6 +499,18 @@ open class _ChatMessageListVC<ExtraData: ExtraDataTypes>:
...
     /// - Parameter typingMembers: typing members gotten from `channelController`
     open func showTypingIndicator(typingMembers: [_ChatChannelMember<ExtraData.User>]) {
+        if typingIndicatorView.isHidden {
+            Animate {
+                self.collectionView.contentInset.bottom += self.typingIndicatorViewHeight
+                self.collectionView.scrollIndicatorInsets.bottom += self.typingIndicatorViewHeight
+            }
+
+            let lastIndexPath = IndexPath(item: 0, section: 0)
+            if collectionView.indexPathsForVisibleItems.contains(lastIndexPath) {
```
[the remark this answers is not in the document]
nuno-vieira: In this case, I actually do not want that, but I want `isLastCellPartiallyVisible` 😄 Maybe I can create that extension as well?

PR #1155, opened 2021-06-04
Fix hardcorded view controller types

VojtaStavik:
Just a couple of forgotten hardcoded types...


b-onc (approved):
Don't we need CHANGELOG entries for these?

PR #1160, opened 2021-06-06
[CIS-885] Fix context menu for messages

DominikBucher12:
# What this PR do:
- Fixes issue where longtap on a message in messageList would not be seen because the message is too long and there is no scrollViewAnymore
# How to test it
- Go to any messagelist and type down message which has bigger height than the device, then long-tap it and expect the message to be truncated and all items from contextMenu available. 

# Known issues: 
This fixes the issue, however introduces some strange UI behavior, this is probably work for @skorepak aka UI Polisher :) 

https://user-images.githubusercontent.com/17381941/124087990-d32c4780-da52-11eb-8273-65fc759f98a9.mov

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### 🐞 Fixed 
 - `ConnectionController` fires its `controllerDidChangeConnectionStatus` method only when the connection status actually changes [#1207](https://github.com/GetStream/stream-chat-swift/issues/1207)
 - Fix cancelled ephemeral (giphy) messages and deleted messages are visible in threads [#1238](https://github.com/GetStream/stream-chat-swift/issues/1238)
-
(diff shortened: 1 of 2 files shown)


Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -392,10 +392,13 @@ open class _ChatMessageListVC<ExtraData: ExtraDataTypes>:
             return controller
         }()
 
+        messageContentView.textView?.textContainer.maximumNumberOfLines = 6
```
VojtaStavik, line 395: Rather than limiting it like this, it would be better to use the autolayout to truncate the message on its max allowed with.

VojtaStavik (commented):
Not that this is wrong but it would be better to have a more flexible solution that works with all possible screen sizes. This can be part of @skorepak's UI polishing efforts :)

skorepak (approved):
👀

skorepak:
@DominikBucher12 Do you have idea how to fix the UI? can you try to fix it? Or let me know and I'll take over and will try to fix that.

PR #1166, opened 2021-06-08
Primitive components release prep

VojtaStavik:
The first batch of components from bucket #1.


dmigach (approved):

Sources/StreamChatUI/CommonViews/InputTextView/InputTextView.swift
```diff
@@ -8,17 +8,17 @@ import UIKit
 /// A view for inputting text with placeholder support. Since it is a subclass
 /// of `UITextView`, the `UITextViewDelegate` can be used to observe text changes.
 open class InputTextView: UITextView, AppearanceProvider {
-    public lazy var placeholderLabel: UILabel = UILabel()
+    open private(set) lazy var placeholderLabel: UILabel = UILabel()
```
DominikBucher12, line 11: **I hope I won't ruin party with this nit** 😅 But this is our custom property on this class and it should probably have a little documentation as we do even with obvious properties.

DominikBucher12 (approved):
LGTM 🤘🏻

PR #1170, opened 2021-06-09
Fix app auto-reconnect on background

b-onc:
Also moves the handling logic from WSClient to ChatClient.

Basically, when we disconnected, we automatically set a timer for reconnection, and reconnected while in background.

### DIFF
--- a/Sources/StreamChat/ChatClient.swift
+++ b/Sources/StreamChat/ChatClient.swift
@@ -196,6 +196,8 @@ public class _ChatClient<ExtraData: ExtraDataTypes> {
     
     private(set) lazy var internetConnection = environment.internetConnection()
     private(set) lazy var clientUpdater = environment.clientUpdaterBuilder(self)
+    /// Used for starting and ending background tasks. Hides platform specific logic.
+    private lazy var backgroundTaskScheduler = environment.backgroundTaskSchedulerBuilder()
     
     /// The environment object containing all dependencies of this `Client` instance.
     private let environment: Environment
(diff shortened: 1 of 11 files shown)


Sources/StreamChat/ChatClient.swift
```diff
@@ -342,6 +349,42 @@ public class _ChatClient<ExtraData: ExtraDataTypes> {
...
+            // Can't initiate a background task, close the connection
+            clientUpdater.disconnect(source: .systemInitiated)
+        }
+    }
+    
+    private func handleAppDidBecomeActive() {
+        cancelBackgroundTaskIfNeeded()
+        
+        guard config.shouldConnectAutomatically else { return }
+        guard connectionStatus != .connected && connectionStatus != .connected else {
```
VojtaStavik, line 378: nit: we should maybe add one more `&& connectionStatus != .connected` to be 300% sure :trollface:

b-onc:
```swift
let confidenceChecks = (1...10).random()
for _ in confidenceChecks {
  guard connectionStatus != .connected else { return }
}
```

Sources/StreamChat/Config/ChatClientConfig.swift
```diff
@@ -66,6 +66,21 @@ public struct ChatClientConfig {
     /// Is `true` by default.
     public var shouldConnectAutomatically = true
     
+    /// If set to `true`, the `ChatClient` will try to stay connected while app is backgrounded.
+    /// If set to `false`, websocket disconnects immediately when app is backgrounded.
```
VojtaStavik, line 70: nit: Would be fine to mention that the client starts starts a background task for this

VojtaStavik (approved):
Great refactoring 👍 Please add CHANGELOG

PR #1171, opened 2021-06-09
Composer SwiftUI compatibility fixes

VojtaStavik:
The way we present suggestions was not compatible with SwiftUI. I updated the logic to use autolayout rather than manual frame calculation.

### DIFF
--- a/Sources/StreamChatUI/CommonViews/Suggestions/ChatSuggestionsViewController.swift
+++ b/Sources/StreamChatUI/CommonViews/Suggestions/ChatSuggestionsViewController.swift
@@ -52,7 +52,14 @@ open class _ChatSuggestionsViewController<ExtraData: ExtraDataTypes>: _ViewContr
     // otherwise it causes loop for height of this controller and as a result this controller height will be 0 as well.
     // Note: This value can be 1, it's just for purpose of 1 cell being visible.
     private let defaultRowHeight: CGFloat = 44
-    private var frameObserver: NSKeyValueObservation?
+
+    /// The constraints responsible for setting the height of the main view.
+    public lazy var heightConstraints: NSLayoutConstraint = {
+        let constraint = view.heightAnchor.pin(equalToConstant: 0)
+        constraint.isActive = true
+        return constraint
+    }()
+    
     private var collectionViewHeightObserver: NSKeyValueObservation?
     
     override open func setUp() {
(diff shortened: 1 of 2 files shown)


Sources/StreamChatUI/Composer/ComposerVC.swift
```diff
@@ -287,8 +287,9 @@ open class _ComposerVC<ExtraData: ExtraDataTypes>: _ViewController,
         composerView.confirmButton.isEnabled = !content.isEmpty
 
         let isAttachmentButtonHidden = !content.isEmpty || channelConfig?.uploadsEnabled == false
-        let isCommandsButtonHidden = !content.isEmpty
+        let isCommandsButtonHidden = !content.isEmpty || channelConfig?.commands?.isEmpty == true
```
DominikBucher12, line 290: This is possibly nit, but here we go :D 

This will fail if `channelConfig?.commands?` is `nil`, I just took a look and you can make `commands` property not optional since it either get parsed or is assigned with empty Array...

Sources/StreamChatUI/CommonViews/Suggestions/ChatSuggestionsViewController.swift
```diff
@@ -84,47 +91,26 @@ open class _ChatSuggestionsViewController<ExtraData: ExtraDataTypes>: _ViewContr
...
-                          newSize.height < cellHeight * self.numberOfVisibleRows
-                    else {
-                        self.view.frame.size.height = cellHeight * self.numberOfVisibleRows
-                        self.updateViewFrame()
-                        return
-                    }
-                    self.view.frame.size.height = newSize.height
-                    self.updateViewFrame()
+                    let newHeight = min(newSize.height, cellHeight * self.numberOfVisibleRows)
+                    self.heightConstraints.constant = newHeight
```
DominikBucher12, line 101: nice 👌

Sources/StreamChatUI/CommonViews/Suggestions/ChatSuggestionsViewController.swift
```diff
@@ -84,47 +91,26 @@ open class _ChatSuggestionsViewController<ExtraData: ExtraDataTypes>: _ViewContr
...
                 }
             }
         )
-        updateContent()
```
DominikBucher12, line 105: I probably do not understand deletion of this as well, I knew this had some meaning, however I should call `updateContentIfNeeded`

DominikBucher12 (approved):
Looks good to me, just 1 nit and one question about deleting the `updateContent`

Sources/StreamChatUI/CommonViews/Suggestions/ChatSuggestionsViewController.swift
VojtaStavik: updateContent is called automatically

PR #1172, opened 2021-06-09
UI components fixes

dmigach:
* Add a SwiftUI wrapper for ChatMessageList
* Add snapshot tests for `ChatMessageList`
* Fix `deleteButtonPressedForCell` and `moreButtonPressedForCell` calling nothing in `ChatChannelListVC`


Sources/StreamChatUI/ChatMessageList/ChatMessageListVC+SwiftUI.swift
```diff
@@ -0,0 +1,32 @@
...
+import SwiftUI
+
+@available(iOS 13.0, *)
+/// A `UIViewControllerRepresentable` subclass which wraps `ChatMessageListVC` and shows list of channels.
+public typealias ChatMessageListView = _ChatMessageListVC<NoExtraData>.View
+
+@available(iOS 13.0, *)
+extension _ChatMessageListVC {
+    /// A `UIViewControllerRepresentable` subclass which wraps `ChatMessageListVC` and shows list of channels.
+    public struct View: UIViewControllerRepresentable {
```
VojtaStavik: nit: `UIViewControllerRepresentable` is a protocol so it can't have subclasses :) Maybe `A SwiftUI View that wraps ...` ?

VojtaStavik (approved):

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC+SwiftUI.swift
dmigach: Thanks for the spot, fixed it both here and where I copied this from 😬

PR #1176, opened 2021-06-11
Add snapshot tests for ChatThreadVC

dmigach:
* Add snapshot tests for ChatThreadVC
* Only consider message list layout initially prepared if the width is not equal to zero


Sources/StreamChatUI/ChatMessageList/ChatMessageListCollectionViewLayout.swift
```diff
@@ -277,11 +277,13 @@ open class ChatMessageListCollectionViewLayout: UICollectionViewLayout {
         super.prepare()
 
         guard !didPerformInitialLayout else { return }
-        didPerformInitialLayout = true
 
         guard currentItems.isEmpty else { return }
         guard let cv = collectionView else { return }
         currentCollectionViewWidth = cv.bounds.width
+        
+        guard currentCollectionViewWidth != 0 else { return }
```
dmigach, line 285: For `ChatThreadVC` after the initial preparation the width is zero which lead to this:
image
I wasn't able to figure out why this happens for `ChatThreadVC` but not for `ChatMessageListVC`, but we agreed with @VojtaStavik that this fix shouldn't hurt

b-onc (approved):

PR #1175, opened 2021-06-10
UI polishing - message layout animation

skorepak:
## In this PR, I have 
- fixed position of the reactions bubble based on figma [link](https://www.figma.com/file/ekifwChR9tR7zRJg1QEzSM/Chat-Sample-App?node-id=13285%3A230)
old:
Bez názvu 2
new:
Bez názvu
- added LayoutOption dataSource to the CollectionView layout to read the layout options and make diff
- added animation for appearing a disappearing reactions on the message

Video (first is standard speed, then again with Slow animation option in the simulator):
https://user-images.githubusercontent.com/7560049/121557894-6b677b80-ca15-11eb-895e-b8b930a44d33.mp4

PR #1181, opened 2021-06-11
Fix shrink button is visible when it shouldn't be

b-onc:
If the channel type doesn't support attachments or commands, shrink button shouldn't be displayed.

The change in UIImageView+Extension was to silence the Nuke deprecation warning

Also add CHANGELOG entry for #1171

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,20 +13,22 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   - `didLongPress` -> `handleLongPress`
   - `textDidChange` -> `handleTextChange`
   If you've subclassed UI components and overridden these functions, you should rename your overrides.
-  For more information, see (#1177)[https://github.com/GetStream/stream-chat-swift/pull/1177] and [#1178](https://github.com/GetStream/stream-chat-swift/issues/1178)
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -23,10 +23,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Add `asView` helper for getting SwiftUI views from StreamChatUI UIViewControllers (#1174)[https://github.com/GetStream/stream-chat-swift/pull/1174] 
 
 ### 🔄 Changed
+- Logic for displaying suggestions (commands or mentions) were not compatible with SwiftUI, so it's changed to AutoLayout (#1171)[https://github.com/GetStream/stream-chat-swift/pull/1171]
```
VojtaStavik, line 26: the link format is broken (see preview)

VojtaStavik (approved):

CHANGELOG.md
b-onc: Someone broke the link format (possibly due to manual writing) and all of them broke due to copy/pastes..... Good catch, fixed them all

VojtaStavik: All credits to @skorepak https://github.com/GetStream/stream-chat-swift/pull/1175#discussion_r649357018

PR #1192, opened 2021-06-17
Mentions refactoring

b-onc:
1. We had some un-overridable functions in ChatChannelListVC, reported by users. First commits `open`s them
1. `mentionedUsers` was not being sent to backend, now that is fixed
1. `ComposerVC` now correctly handles mentions, and it's also possible to customize what's displayed as mention text. 

Only missing feature is editing mentions, currently not possible.


nuno-vieira (approved):
LGTM 🎉

PR #1193, opened 2021-06-17
Refactor sidebar

b-onc:
Our Sidebar is much closer to other SDKs now, and looks much better

Quickly check it out here: https://swift-docs-new-sidebar.loca.lt/chat/docs/sdk/ios/?language=swift


docusaurus/sidebars-ios.json
```diff
@@ -1,14 +1,18 @@
 {
   "docs": [
-    "introduction",
     {
-      "Getting Started": [
-        "getting-started/integration",
-        "getting-started/up-running"
-     ]
+      "Basics": [
```
DominikBucher12, line 4: 👍

nuno-vieira:
image

Can we just call it "Controllers" ? Instead of "StreamChat Controller". IMO looks better. Unless we instead of "UI Components" have "StreamChatUI Components"

b-onc:
@nuno-vieira good idea 👍

docusaurus/sidebars-ios.json
```diff
@@ -42,7 +42,7 @@
       ]
     },
     {
-      "StreamChat Controllers": [
+      "Controllers": [
```
DominikBucher12, line 45: 👍

DominikBucher12 (approved):
From my side this looks good, I like this actually better... Aand +1 point For the migration guide outside of scope of name of PR :D

PR #1195, opened 2021-06-18
[CIS-907] Add Working with the Composer Guide

nuno-vieira:
- Renames the Chat Suggestions View Controller to follow the pattern across the app
- Changes the way we were disabling the suggestions of the iMessageClone
- Adds the guide for working with the composer

## How to Test
https://stream-composer-guide.loca.lt/chat/docs/sdk/ios/guides/working-with-the-composer (In case it is offline, please ask me on Slack to put it online)


Sources/StreamChatUI/Composer/ComposerVC.swift
```diff
@@ -40,7 +40,7 @@ open class _ComposerVC<ExtraData: ExtraDataTypes>: _ViewController,
         /// The text of the input text view.
         public var text: String
         /// The state of the composer.
-        public var state: ComposerState
+        public let state: ComposerState
```
b-onc, line 43: Why do we need it as immutable?

docusaurus/docs/iOS/guides/working-with-the-composer.md
```diff
@@ -0,0 +1,221 @@
+---
+title: Working with the Composer
+---
+
+The `ComposerVC` is the component responsible to create new messages or change existing messages. The composer belongs to the `ChatVC` component that integrates the message list and the composer in one view. But if you are not using our `ChatVC` you can use the composer component in isolation with your own chat view as well. The `ComposerVC` manages all the logic and actions of the `ComposerView`, where the latter is only responsible for the composer's layout.
```
b-onc, line 5: Link to ComposerVC reference docs (and other components like ChatVC) would be nice in intro

docusaurus/docs/iOS/guides/working-with-the-composer.md
```diff
@@ -0,0 +1,221 @@
...
+        super.updateContent()
+
+        switch content.state {
+        case .new:
+            composerView.inputMessageView.textView.placeholderLabel.text = "iMessage"
+        default:
+            break
+        }
+    }
+}
```
b-onc, line 40: We should add comments to this code and explain why we're doing the stuff we're doing

docusaurus/docs/iOS/guides/working-with-the-composer.md
```diff
@@ -0,0 +1,221 @@
...
+
+// Then, replace the components:
+components.messageComposerVC = iMessageComposerVC.self
+components.messageComposerView = iMessageComposerView.self
+```
+
+**Result:**
+<img src={require("../assets/composer-ui-custom-layout.png").default} width="50%"/>
```
b-onc, line 48: A before/after table would be better

docusaurus/docs/iOS/guides/working-with-the-composer.md
```diff
@@ -0,0 +1,221 @@
...
+
+Since the Composer is built with multiple `ContainerStackView`'s with a similar API of a `UIStackView`, it is very easy to change the Composer layout it is just a question of removing views from containers and adding them back to the desired containers. In this case, we override the `setUpLayout()` lifecycle to remove the `commandsButton` all together and we move the `sendButton` from the `trailingContainer` to the `inputTextContainer`. As a last touch, we change the `attachmentButton` image by overriding the `setUpAppearance()` lifecycle.
```
b-onc, line 50: We should show the composer's layout (image) showing the container/element positions here. We mention stacks and components but we don't show where in the composer these stacks/components are located

docusaurus/docs/iOS/guides/working-with-the-composer.md
```diff
@@ -0,0 +1,221 @@
...
+
+## Customizing the Composer Actions
+All the actions triggered by the Composer buttons are wired in the `ComposerVC.setUp()` lifecycle. When adding new buttons it is possible to add new actions by overriding this function. Let's imagine an example where we have added a new custom emoji button to open an emoji picker:
+```swift
+final class CustomComposerVC: ComposerVC {
+    override func setUp() {
+        super.setUp()
+
+        composerView.emojiButton.addTarget(self, action: #selector(showEmojiPicker), for: .touchUpInside)
```
b-onc, line 59: Let's show introducing this button to the layout, too

docusaurus/docs/iOS/guides/working-with-the-composer.md
```diff
@@ -0,0 +1,221 @@
...
+```swift
+final class CustomComposerVC: ComposerVC {
+    override var mentionSymbol: String { "👤" }
+
+    override var commandSymbol: String { "⚡️" }
+}
+
+components.messageComposerVC = CustomComposerVC.self
+```
+That's it, with this in place, the suggestions will now be triggered when you type your custom symbol.
```
b-onc, line 95: Let's remove this section completely now, or move it to the bottom, it's not very relevant

docusaurus/docs/iOS/guides/working-with-the-composer.md
```diff
@@ -0,0 +1,221 @@
...
+
+components.suggestionsVC = CustomChatSuggestionsVC.self
+```
+Here we simply remove the shadow and customize the border style of the suggestions component.
+
+**Result:**
+
+| Before  | After |
+| ------------- | ------------- |
+| <img src={require("../assets/composer-suggestionsVC-before.png").default} width="100%"/> | <img src={require("../assets/composer-suggestionsVC-after.png").default} width="100%"/> |
```
b-onc, line 122: TBH it looks like the first image is blurry and second is more clear... we need a better example for this, like sharp borders or getting rid of borders

docusaurus/docs/iOS/guides/working-with-the-composer.md
```diff
@@ -0,0 +1,221 @@
...
+final class iMessageComposerVC: ComposerVC {
+    override func typingMention(in textView: UITextView) -> (String, NSRange)? {
+        // Don't show user suggestions
+        return nil
+    }
+
+    override func typingCommand(in textView: UITextView) -> String? {
+        // Don't show command suggestions
+        return nil
+    }
```
b-onc, line 166: We should explain that this should be disabled from channel type in dashboard, but it's possible like this too

docusaurus/docs/iOS/guides/working-with-the-composer.md
```diff
@@ -0,0 +1,221 @@
...
+components.messageComposerVC = iMessageComposerVC.self
+```
+
+## Adding a new Composer State
+
+The `ComposerVC` has by default three different states, `.new`, `.edit` and `.quote`, which are represented by the `ComposerState`. These states define if the composer is creating a new message, editing a new message, or quoting an existing message.
+
+| `.new`  | `.edit` | `.quote` |
+| ------------- | ------------- | ------------- |
+| <img src={require("../assets/composer-ui-state-new.png").default} width="100%"/> | <img src={require("../assets/composer-ui-state-edit.png").default} width="100%"/> | <img src={require("../assets/composer-ui-state-quote.png").default} width="100%"/> |
```
b-onc, line 178: Kudos

docusaurus/docs/iOS/guides/working-with-the-composer.md
```diff
@@ -0,0 +1,221 @@
...
+        super.updateContent()
+
+        switch content.state {
+        case .custom:
+            composerView.titleLabel.text = "Custom State"
+        default:
+            break
+        }
+    }
+}
```
b-onc, line 218: We need to come up with a better example....

b-onc (commented):
Looks very good!

Sources/StreamChatUI/Composer/ComposerVC.swift
nuno-vieira: It is explained in the documentation, we want to only change the state by creating a new `mutating func` and create the `Content` again. This is to avoid like when editing a message, we change the state to `state = .quote` but forget to change the `editingMessage` and `quotingMessage` accordingly

docusaurus/docs/iOS/guides/working-with-the-composer.md
nuno-vieira: How do we reference this? Manually? Would be nice to have a way that even if we change the structure of the docs, the refs are not broken 🤔 But currently this is manual right?

docusaurus/docs/iOS/guides/working-with-the-composer.md
nuno-vieira: Isn't this layout image to be shown only on the UI Components docs? Because AFAIK; they are populated automatically on the docs right? Should we also add them to the guides, or we can reference the ComposerVC component so they can look at the layout?

docusaurus/docs/iOS/guides/working-with-the-composer.md
nuno-vieira: You mean adding the emoji button to the ComposerView layout right? Yes makes sense 👍

docusaurus/docs/iOS/guides/working-with-the-composer.md
nuno-vieira: Yes makes sense, let's remove it, since anyone can understand this just by looking at the code 👍

docusaurus/docs/iOS/guides/working-with-the-composer.md
nuno-vieira: Oh, I just saw that the config logic is wrongly on `showCommandsFunction`, it should be here. But these are different things. You maybe still want the commands, but don't want the typing suggestions. This is showing how you disable only the typing suggestions, not the functionality of the commands, that should be with the config yes.

docusaurus/docs/iOS/guides/working-with-the-composer.md
nuno-vieira: Yes you are right 🤔 I can't think about a new state actually.. and if even is a valid use case... It is a shame because it was quite the work to document this, but I don't see a point of adding new states at all :/

docusaurus/docs/iOS/guides/working-with-the-composer.md
DominikBucher12:
```suggestion
The [`ComposerVC`](../ReferenceDocs/Sources/StreamChatUI/Composer/ComposerVC.md) is the component responsible to create new messages or change existing messages. The composer belongs to the `ChatVC` component that integrates the message list and the composer in one view. But if you are not using our `ChatVC` you can use the composer component in isolation with your own chat view as well. The `ComposerVC` manages all the logic and actions of the `ComposerView`, where the latter is only responsible for the composer's layout.
```


@nuno-vieira just take a look at other guides, the structure doesn't matter because you reference to the actual location of the file instead of some docusaurus alias... see `Working with messagelist layout` and the suggestion :)

docusaurus/docs/iOS/guides/working-with-the-composer.md
b-onc: Since the guide is explaining this stuff, replicating the overview image is necessary, to reduce the amount of jump users need to do. And sometimes, the image in the guide will be more informative (more detailed) than the one in the reference doc, as should be the case here. We mention a lot of elements but don't show them visually

docusaurus/docs/iOS/guides/working-with-the-composer.md
b-onc: Then let's remove the example or move it to the bottom

docusaurus/docs/iOS/guides/working-with-the-composer.md
```diff
@@ -0,0 +1,221 @@
...
+
+        leadingContainer.removeArrangedSubview(commandsButton)
+        trailingContainer.removeArrangedSubview(sendButton)
+        inputMessageView.inputTextContainer.addArrangedSubview(sendButton)
+    }
+    
+    override func setUpAppearance() {
+        super.setUpAppearance()
+
+        attachmentButton.setImage(
```
tbarbugli, line 22: isn't better to make this configurable on the `Appearance.images` ?

nuno-vieira: Yes it should! This is wrong, the customization shouldn't be like this, nice catch 👍

b-onc (approved):
Looks really solid!

tbarbugli (approved):

tbarbugli (approved):

PR #1196, opened 2021-06-18
Fix composer shrink input button always hidden

nuno-vieira:
The shrink input button was broken after some logic introduced to support hiding it when the commands and attachments are not enabled


skorepak (approved):

DominikBucher12 (approved):
Thank you for such a quick fix!

PR #1199, opened 2021-06-21
Nicer intro

tbarbugli:
- basics/introduction and basics/integration are now merged into basics/overview
- basics/migrating-from-3-to-4 moved to guides for now
- basics/migrating-from-1.x-and-2.x moved to guides for now
- basics/faq removed empty page (we can re-add later on)

TODOs:

* [ ] basics/getting-started (example needs to be improved)
* [ ] find a place to keep general docs removed from basics


docusaurus/docs/iOS/basics/getting-started.md
```diff
@@ -2,18 +2,17 @@
 title: Getting Started
 ---
 
-To get started with StreamChat, we suggest you register and acquire an API token [on our website](getstream.io). Please enable developer tokens for your app before moving on with the guide, since we'll use them for prototyping.
+This section provides a high level overview of the library setup, core components, and how they fit together. It's a great starting point and you can follow along in your code editor. For a complete, step-by-step guide in terms setting up a React project or instructions on creating specific files, see our [iOS Chat tutorial](/tutorials/ios-chat/).
+
+## Your First App with Stream Chat React
```
evsaev, line 7:
```suggestion
## Your First App with Stream Chat iOS
```

docusaurus/docs/iOS/basics/getting-started.md
```diff
@@ -2,18 +2,17 @@
 title: Getting Started
 ---
 
-To get started with StreamChat, we suggest you register and acquire an API token [on our website](getstream.io). Please enable developer tokens for your app before moving on with the guide, since we'll use them for prototyping.
+This section provides a high level overview of the library setup, core components, and how they fit together. It's a great starting point and you can follow along in your code editor. For a complete, step-by-step guide in terms setting up a React project or instructions on creating specific files, see our [iOS Chat tutorial](/tutorials/ios-chat/).
```
evsaev, line 5:
```suggestion
This section provides a high level overview of the library setup, core components, and how they fit together. It's a great starting point and you can follow along in your code editor. For a complete, step-by-step guide in terms setting up an iOS project or instructions on creating specific files, see our [iOS Chat tutorial](/tutorials/ios-chat/).
```

docusaurus/docs/iOS/basics/getting-started.md
```diff
@@ -2,18 +2,17 @@
...
+
+## Your First App with Stream Chat React
+
+Before starting, make sure you have installed `StreamChatUI`, as explained in the [Installation](../#installation) section.
+
+The below example is all the code you need to launch a fully functioning chat experience. The Chat and Channel components are React context providers that pass a variety of values to their children, including UI components, stateful data, and action handler functions.
```
evsaev, line 11: this should probably be adjusted for iOS 😄 smth about `Components/Appearance`, offline support etc.

Scripts/generateDocumentation.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-
+set -e
```
DominikBucher12, line 2: 👍

docusaurus/docs/iOS/customization/theming.md
```diff
@@ -1,3 +1,72 @@
 ---
 title: Theming
----
\ No newline at end of file
+---
+
+You can customize the look and feel of all UI components provided by `StreamChatUI`. The SDK allows you to change appearance of components such as colors and fonts via the `Appearance` component. Changes to appearance should be done as early as possible in your application, `SceneDelegate` and `AppDelegate` are usually the right place to do this. The SDK comes with a singleton object `Appearance.default`.
```
DominikBucher12, line 5: > The SDK comes with a singleton object `Appearance.default`. 

This is true. Just a note that customers can also create custom `Appearance` instance, then set the properties to it and then assign it `Appearance.default =  myApperance` or basically any `AppearanceProvider` implementing view so those it can have different appearance than rest of SDK.  This is very specific usecase but I think it's still worth mentioning.

docusaurus/docs/iOS/components/channel-list-item-view.md
```diff
@@ -0,0 +1,74 @@
+---
+title: ChatChannelListItemView
+---
+
+import Properties from '../reference-docs/sources/stream-chat-ui/appearance.color-palette-properties.md'
```
DominikBucher12: 🤯

docusaurus/docs/iOS/customization/components.md
```diff
@@ -46,53 +46,62 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
...
 
-### Example: custom avatar
+### `updateContent()`
+Finally, this last lifecycle is called whenever the data of the component changes. Here is where you can change the logic of the component, change how the data is displayed or formatted. In the Stream SDK all of the components have a `content` propery that represents the data of the component. The rule of thumb to use this lifecycle is that if the change you want to do depends on the data of the component, then you should use this lifecycle method, even, for example, to do layout changes that are impacted by the content.
 
-Let's say, we want to change the appearance of avatars from the rounded ones to the rectangular ones:
+In addition to this, view components expose their content with the `content` property. For instance the `MessageContent` component `content`'s property holds the `ChatMessage` object.
```
DominikBucher12, line 79: I would add some note that if you need to force update content to reflect some changes, you need to call `updateContentIfNeeded()` instead od calling `updateContent()` directly, that `updateContentIfNeeded` ensures that the view is added to superView or viewController is already loaded.

DominikBucher12 (commented):
This looks very good 🧨

docusaurus/docs/iOS/guides/working-with-custom-attachments.md
```diff
@@ -0,0 +1,352 @@
+---
+title: Working with Attachments in Composer
+---
```
nuno-vieira, line 3: Should we stop using the "Working with" pattern? I think it is very redundant that for every guide we start with this naming. For example on React Native I think it is much better and simpler:
image

VojtaStavik:
The `ChatChannelListItemView` looks great, so does the new structure. Awesome 👏

docusaurus/docs/iOS/customization/theming.md
nuno-vieira: I actually think a better name for this is `Appearance.current`. This pattern is also user in other Apple's APIs. `default` is a bit confusing, it looks that there's more appearance objects, where in fact we can only have one. Either `current` or `shared` are better options, IMO.

VojtaStavik: `Appearance` and `Components` are based on the responder chain. You can have different objects for different objects. In the future, we will have built-in presets for specific UI: like messaging (the current default), live chat, etc.

VojtaStavik: `.default` means - if you don't specify your custom appearance, we will use the default one.

DominikBucher12 (approved):
I am approving this, I really love it and in my eyes it looks cool, would wait for someone else to approve or possibly change the suggestions written in comments...

PR #1201, opened 2021-06-21
[CIS-911] Fix Docusaurus linking with slugs 

DominikBucher12:
# What this PR do:
- Adds slug parameter to all reference-docs generated files
# How to test it
- Deploy to staging and see if after refresh the website doesn't show 404


docusaurus/docs/iOS/ReferenceDocs/Sources/StreamChat/APIClient/Array.md
```diff
@@ -1,7 +1,8 @@
 ---
 id: array 
 title: Array
---- 
+slug: ReferenceDocs/Sources/StreamChat/APIClient/array
+---
```
b-onc, line 5: This should be:
```
---
id: array
title: Array
slug: reference-docs/sources/streamchat/apiclient/array
```
and make sure it works locally (local url is the same as slug)

PR #1203, opened 2021-06-21
[CIS-913] Fix layout feedback loop in ChatMessageContentView

dmigach:
I reproduced the issue @b-onc described and was able to fix it with this change.
I checked the resulting layouts before/after and they're identical.
I would really appreciate an opinion on how adequate the fix is, and if it's not, we still can figure out another solution based on the fact that this is still a root cause of the issue.

### DIFF
--- a/Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
+++ b/Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
@@ -298,7 +298,7 @@ open class _ChatMessageContentView<ExtraData: ExtraDataTypes>: _View, ThemeProvi
         // Quoted message
         if options.contains(.quotedMessage) {
             let quotedMessageView = createQuotedMessageView()
-            bubbleContentContainer.addArrangedSubview(quotedMessageView, respectsLayoutMargins: true)
+            bubbleContentContainer.addArrangedSubview(quotedMessageView)
         }
 
         // Text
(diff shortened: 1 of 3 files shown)


Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
```diff
@@ -545,7 +545,6 @@ open class _ChatMessageContentView<ExtraData: ExtraDataTypes>: _View, ThemeProvi
             textView?.adjustsFontForContentSizeCategory = true
             textView?.textContainerInset = .init(top: 0, left: 8, bottom: 0, right: 8)
             textView?.textContainer.lineFragmentPadding = 0
-            textView?.translatesAutoresizingMaskIntoConstraints = false
```
dmigach, line 548: Because it's already `.withoutAutoresizingMaskConstraints` above

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
```diff
@@ -298,7 +298,7 @@ open class _ChatMessageContentView<ExtraData: ExtraDataTypes>: _View, ThemeProvi
         // Quoted message
         if options.contains(.quotedMessage) {
             let quotedMessageView = createQuotedMessageView()
-            bubbleContentContainer.addArrangedSubview(quotedMessageView, respectsLayoutMargins: true)
```
DominikBucher12, line 301: 🤯

VojtaStavik (approved):
If the layouts are identical, I think we're good to do here 👍  Please also add a changelog entry

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
skorepak: 👍🏻

skorepak (approved):

PR #1211, opened 2021-06-23
[CIS-243] Implement silent messages

SagarSDagdu:
## What does this PR do
With this PR, sending silent messages is possible.
More about silent messages can be read here: https://getstream.io/chat/docs/react/silent_messages/?language=swift

## Changes
- Add `isSilent` flag to `MessageRequestBody`
- Make changes to public APIs in `ChannelController` and `MessageController` to accept this flag
- Propagate this flag to lower levels and make relevant changes to the tests.

### DIFF
--- a/Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
+++ b/Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
@@ -756,6 +756,7 @@ public extension _ChatChannelController {
     /// - Parameters:
     ///   - text: Text of the message.
     ///   - pinning: Pins the new message. `nil` if should not be pinned.
+    ///   - isSilent: A flag indicating whether the message is a silent message. Silent messages are special messages that don't increase the unread messages count nor mark a channel as unread.
     ///   - attachments: An array of the attachments for the message.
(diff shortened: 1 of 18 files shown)


Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
```diff
@@ -756,6 +756,7 @@ public extension _ChatChannelController {
     /// - Parameters:
     ///   - text: Text of the message.
     ///   - pinning: Pins the new message. `nil` if should not be pinned.
+    ///   - isSilent: A flag indicating whether the message is a silent message.
```
VojtaStavik, line 759: nit: Can you add the following to the doc:
```
Silent messages are special messages that don't increase the unread messages count nor mark a channel as unread. 
```

b-onc (approved):

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
SagarSDagdu: Added it 👍

nuno-vieira (approved):
LGTM!

PR #1216, opened 2021-06-24
[CIS-936] `ConnectionController` fix for duplicate events

VojtaStavik:
Connection events are now fired only when the connection actually changes. Before this PR, we sent the events periodically with the same connection status value. This was causing issues with unnecessary calls `/sync` endpoint.


DominikBucher12 (approved):
🚀

PR #1228, opened 2021-06-29
[CIS-967] Fix ContainerStackView spacing issue

skorepak:
## This PR:
- fixes issue where spacing didn't work when the subview of container stack view was hidden twice before showing. 
- Add snapshot test for this case

## Links: 
https://stream-io.atlassian.net/browse/CIS-967


VojtaStavik (approved):

Sources/StreamChatUI/CommonViews/ContainerStackView.swift
```diff
@@ -496,13 +496,15 @@ extension UILayoutPriority {
 
 extension NSLayoutConstraint {
     func setTemporaryConstant(_ value: CGFloat) {
+        guard originalConstant == nil else { return }
```
evsaev, line 499: hm, the observer shouldn't even be invoked when there's an attempt to hide already hidden subview 🤔

```swift
hidingObserversByView[subview] = subview
            .observe(\.isHidden, options: [.new]) { [weak self] (view, isHiddenChange) in
                if isHiddenChange.newValue == true {
                    self?.hideArrangedSubview(view)
                } else {
                    self?.showArrangedSubview(view)
                }
            }
```

skorepak: That's what I would expect! But the test I added is failing without this fix 😞

evsaev: hm, how about
```swift
.observe(\.isHidden, options: [.new, .old]) { [weak self] (view, isHiddenChange) in
    guard isHiddenChange.oldValue != isHiddenChange.newValue else { return }
    ...
}
```

skorepak: I was testing this and I think the logic failed sometimes. 
The `show/hideArrangedSubview` has some logic inside, which may result in calling the `hideArrangedSubview` method, but not calling the `setTemporaryConstant` method inside. Therefore I fixed it on the lowest level possible to remove any other logic involved on the higher level

evsaev (approved):
🚢

PR #1230, opened 2021-06-29
[CIS-938] Introduce `connectUser` API

dmigach:
As a part of effort on unification across platforms we're introducing new methods for initiating connection on `ChatClient`:
- `connectUser` and overloads for connecting normal users
- `connectGuestUser` and overloads
- `connectAnonymousUser`

In order to connect now it's necessary to first instantiate an instance of `ChatClient` and then call one of the available connection methods:
```
let config = ChatClientConfig(apiKey: .init(userCredentials.apiKey))
let client = ChatClient(config: config)
client.connectUser(token: token)
```
Also `tokenProvider` is now optional in `ChatClient`'s init parameters, but we still instantiate connection in init if token provider was passed for backward compatibility.

More examples on how it now works:
```swit
// Create a token
let token = try! Token(rawValue: userCredentials.token)
// Create a client
let config = ChatClientConfig(apiKey: .init(userCredentials.apiKey))
let client = ChatClient(config: config)
// Normal user case
client.connectUser(
    userInfo: .init(id: userId)
    token: token
) { error in
    if let error = error {
        print(error)
    } else {
        print("connection was successful")
    }
}

// Set a token provider for the token expiration case
client.tokenProvider = { completion in
    authService.fetchToken { result in
        completion(result)
    }
}
client.disconnect()
// A guest user
client.connectGuestUser(userInfo: .init(id: userId))  { error in
    if let error = error {
        print(error)
    } else {
        print("connection was successful")
    }
}

// An anonymous user
client.connectAnonymousUser { error in
    if let error = error {
        print(error)
    } else {
        print("connection was successful")
    }
}
```


Sources/StreamChat/ChatClient.swift
```diff
@@ -299,8 +299,10 @@ public class _ChatClient<ExtraData: ExtraDataTypes> {
         self.eventWorkerBuilders = eventWorkerBuilders
 
         currentUserId = fetchCurrentUserIdFromDatabase()
-
-        clientUpdater.reloadUserIfNeeded(completion: completion)
+        // Backward compatiability
+        if let tokenProvider = tokenProvider {
+            setConnectionInfoAndConnect(tokenProvider: tokenProvider, completion: completion)
```
dmigach, line 304: @VojtaStavik  This part is controversial, on the one hand we need it for backward compatibility, on the other hand it's now possible to both pass a tokenProvider provider, and then connect once again by calling one of the `connect` methods. It doesn't break anything, plus you can always just call `connect` multiple times, so maybe it's ok after all

Sources/StreamChat/Models/UserConnectionInfo.swift
```diff
@@ -0,0 +1,11 @@
...
+// Copyright © 2021 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+public struct UserConnectionInfo<ExtraData: ExtraDataTypes> {
+    let name: String?
+    let imageURL: URL?
+    let extraData: ExtraData.User?
+}
```
VojtaStavik, line 11: nit: Can we name this `UserInfo` and add documentation?

Sources/StreamChat/Config/ChatClientConfig.swift
```diff
@@ -64,6 +64,7 @@ public struct ChatClientConfig {
     /// but has to be initiated manually by calling `connect`.
     ///
     /// Is `true` by default.
+    @available(*, deprecated, message: "Has no effect anymore. It now always behaves like it was set to `true`.")
     public var shouldConnectAutomatically = true
```
VojtaStavik, line 68: message: "This flag has no effect anymore. The flow for setting and for connecting the user has been unified to the `connectUser` set of methods."

Sources/StreamChat/ChatClient.swift
```diff
@@ -314,6 +314,96 @@ public class _ChatClient<ExtraData: ExtraDataTypes> {
...
+        extraData: ExtraData.User = .defaultValue,
+        token: Token,
+        completion: ((Error?) -> Void)? = nil
+    ) {
+        setConnectionInfoAndConnect(
+            name: name,
+            imageURL: imageURL,
+            tokenProvider: .static(token),
+            completion: completion
+        )
```
VojtaStavik, line 330: Extra data are ignored here and in the other calls, too.

Sources/StreamChat/ChatClient.swift
```diff
@@ -314,6 +314,96 @@ public class _ChatClient<ExtraData: ExtraDataTypes> {
         completeConnectionIdWaiters(connectionId: nil)
         completeTokenWaiters(token: nil)
     }
+    
+    public func connectUser(
```
VojtaStavik, line 318: please add documentation all new public API 🙏

Sources/StreamChat/ChatClient.swift
```diff
@@ -314,6 +314,96 @@ public class _ChatClient<ExtraData: ExtraDataTypes> {
...
+    ) {
+        tokenProvider { token in
+            switch token {
+            case let .success(token):
+                setConnectionInfoAndConnect(tokenProvider: .static(token), completion: completion)
+            case let .failure(error):
+                completion?(error)
+            }
+        }
+    }
```
VojtaStavik, line 373: This one won't be needed in the end.

Sources/StreamChat/ChatClient.swift
```diff
@@ -480,6 +581,15 @@ extension ClientError {
             """
         }
     }
+    
+    public class ConnectionWasNotInitiated: ClientError {
+        override public var localizedDescription: String {
+            """
+                Before performing any other actions on chat client it's required to connect by using
```
VojtaStavik, line 588: nit: You can use `\` at the end of the line to prevent an incorrect new line when printing the message.

Sources/StreamChat/Workers/ChatClientUpdater.swift
```diff
@@ -66,8 +66,17 @@ class ChatClientUpdater<ExtraData: ExtraDataTypes> {
         try client.databaseContainer.removeAllData(force: true)
     }
 
-    func reloadUserIfNeeded(completion: ((Error?) -> Void)? = nil) {
-        client.tokenProvider.getToken(client) {
+    func reloadUserIfNeeded(
+        name: String? = nil,
+        imageURL: URL? = nil,
+        completion: ((Error?) -> Void)? = nil
+    ) {
```
VojtaStavik, line 73: extra data is missing from the signature

Sources/StreamChat/ChatClient_Tests.swift
```diff
@@ -617,6 +612,213 @@ class ChatClient_Tests: StressTestCase {
...
+    func test_reloadUserIfNeededIsNotCalled_whenClientIsInitialized_andTokenProviderIsNil() {
+        _ = ChatClient(
+            config: inMemoryStorageConfig,
+            workerBuilders: [],
+            eventWorkerBuilders: [],
+            environment: testEnv.environment
+        )
+        
+        // `reloadUserIfNeeded` is not called on `clientUpdater` hence it wasn't lazily initialised
+        XCTAssertNil(testEnv.clientUpdater)
```
VojtaStavik, line 625: this is just waiting to get randomly broken in the future :) Can we do something like `XCTAssert(testEnv.clientUpdater?.reloadUserIfNeeded_called != true)` instead?

CHANGELOG.md
```diff
@@ -5,6 +5,16 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 # Upcoming
 
+### ✅ Added
+- New methods for establishing connection for `ChatClient`:
```
VojtaStavik, line 9: `New methods for setting the current user for...`?

tbarbugli (changes requested):
please include more code examples in the PR description:

ie.

- using token provider
- connectUser using a closure
- connectUser using an object

Sources/StreamChat/ChatClient.swift
dmigach: What about the one with `connectionInfoProvider: ((Result<(UserInfo<ExtraData>, Token), Error>) -> Void) -> Void`?

dmigach:
Addressed comments, added more examples on the new APIs usage.
There is a couple of things related to new changes that needs a discussion, hopefully it will be resolved tomorrow morning

skorepak (approved):
Quite extensive PR which is hard to follow. 
I didn't find any issues
LGTM 👍🏻

Sources/StreamChat/Models/UserInfo.swift
```diff
@@ -0,0 +1,24 @@
...
+// Copyright © 2021 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+public struct UserInfo<ExtraData: ExtraDataTypes> {
+    public let id: UserId
+    public let name: String?
+    public let imageURL: URL?
+    public let extraData: ExtraData.User
```
evsaev, line 11: nit: docs

Sources/StreamChat/WebSocketClient/WebSocketConnectPayload.swift
```diff
@@ -4,24 +4,34 @@
...
+struct WebSocketConnectPayload<ExtraData: ExtraDataTypes>: Encodable {
     private enum CodingKeys: String, CodingKey {
         case userId = "user_id"
         case userDetails = "user_details"
         case serverDeterminesConnectionId = "server_determines_connection_id"
     }
     
     let userId: UserId
-    let userDetails: UserWebSocketPayload
+    let userDetails: UserWebSocketPayload<ExtraData>
```
evsaev, line 15: why do we send user id twice? 🤔It's in `WebSocketConnectPayload.userId` and in `WebSocketConnectPayload.userDetails.id`

Sources/StreamChat/Workers/ChatClientUpdater.swift
```diff
@@ -11,7 +11,10 @@ class ChatClientUpdater<ExtraData: ExtraDataTypes> {
         self.client = client
     }
 
-    func prepareEnvironment(newToken: Token) throws {
+    func prepareEnvironment(
+        userInfo: UserInfo<ExtraData>?,
+        newToken: Token
+    ) throws {
```
evsaev, line 17: shouldn't we also reconnect if `userInfo` has changed?

Sources/StreamChat/ChatClient.swift
```diff
@@ -312,6 +310,63 @@ public class _ChatClient<ExtraData: ExtraDataTypes> {
...
+            userInfo: userInfo,
+            userConnectionProvider: .guest(
+                userId: userId,
+                name: userInfo.name,
+                imageURL: userInfo.imageURL,
+                extraData: userInfo.extraData
+            ),
+            completion: completion
+        )
+    }
```
evsaev, line 353:
```suggestion
    public func connectGuestUser(
        userInfo: UserInfo<ExtraData>,
        completion: ((Error?) -> Void)? = nil
    ) {
        setConnectionInfoAndConnect(
            userInfo: userInfo,
            userConnectionProvider: .guest(
                userId: userInfo.id,
                name: userInfo.name,
                imageURL: userInfo.imageURL,
                extraData: userInfo.extraData
            ),
            completion: completion
        )
    }
```

Sources/StreamChat/Models/UserInfo.swift
dmigach: Nice spot, thank you

Sources/StreamChat/WebSocketClient/WebSocketConnectPayload.swift
dmigach: I'm not sure, just preserved this behaviour. Maybe Vojta remembers

Sources/StreamChat/Workers/ChatClientUpdater.swift
dmigach: A good question, in my understanding the answer is no because we used to live without passing user info at all, so we only care if the user or token have changed

Sources/StreamChat/ChatClient.swift
dmigach: Great catch

Sources/StreamChat/WebSocketClient/WebSocketConnectPayload.swift
evsaev: ok 👍 not a blocker for me but would be nice to ask backend team

evsaev (approved):
LGTM 👍 Let's fix the docs and🚢

tbarbugli (approved):
awesome work

PR #1236, opened 2021-06-30
[CIS-976] Scroll to bottom animation

evsaev:
**This PR**:
- fixes weird scroll to bottom animation (we go to the first message and then show the most recent one with animation)
- fixes scrolling to bottom when someone is typing if the last message was not fully visible (when the last message is huge the automatic scrolling would be unwanted)

https://user-images.githubusercontent.com/12818985/123972038-31a9e500-d9c3-11eb-8185-1aea049fa99c.mp4

### DIFF
--- a/Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
+++ b/Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
@@ -481,7 +481,7 @@ open class _ChatMessageListVC<ExtraData: ExtraDataTypes>:
                 self.listView.scrollIndicatorInsets.top += self.typingIndicatorViewHeight
             }
 
-            if listView.isLastCellVisible {
+            if listView.isLastCellFullyVisible {
                 scrollToMostRecentMessage()
             }
         }
--- a/Sources/StreamChatUI/ChatMessageList/ChatMessageListView.swift
+++ b/Sources/StreamChatUI/ChatMessageList/ChatMessageListView.swift
@@ -209,9 +209,7 @@ open class _ChatMessageListView<ExtraData: ExtraDataTypes>: UITableView, Customi
         with changes: [ListChange<_ChatMessage<ExtraData>>],
         completion: (() -> Void)? = nil
     ) {
-        // Before committing the changes, we need to check if were scrolled
-        // to the bottom, if yes, we should stay scrolled to the bottom
-        var shouldScrollToBottom = isLastCellFullyVisible
+        var shouldScrollToBottom = false
                 
         performBatchUpdates({
             changes.forEach {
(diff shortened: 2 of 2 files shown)


Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -481,7 +481,7 @@ open class _ChatMessageListVC<ExtraData: ExtraDataTypes>:
                 self.listView.scrollIndicatorInsets.top += self.typingIndicatorViewHeight
             }
 
-            if listView.isLastCellVisible {
+            if listView.isLastCellFullyVisible {
                 scrollToMostRecentMessage()
```
VojtaStavik, line 485: one maybe unrelated idea how to easily improve the overall UX is to animate the typing indicator view in/out with slide animation which visually matches the inset update animation of the table view. Slack does it like this.

Sources/StreamChatUI/ChatMessageList/ChatMessageListView.swift
```diff
@@ -211,7 +211,7 @@ open class _ChatMessageListView<ExtraData: ExtraDataTypes>: UITableView, Customi
     ) {
         // Before committing the changes, we need to check if were scrolled
         // to the bottom, if yes, we should stay scrolled to the bottom
-        var shouldScrollToBottom = isLastCellFullyVisible
+        var shouldScrollToBottom = false
```
VojtaStavik, line 220: nit: the comment now doesn't make sense

VojtaStavik (approved):

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
evsaev: Just tried it.

Slide in/out works nice if typing indicator is added to table view below the content so it becomes visible/hidden natively when inset is changed. 

But we can't keep it below the content because it's not visible if channel is scrolled up 🤔

We could have a bottom edge typing indicator constraint which we would animate together with insets but can't find correct animation options yet

Sources/StreamChatUI/ChatMessageList/ChatMessageListView.swift
evsaev: ✅ removed

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
evsaev: @VojtaStavik can we address this in a separate PR?

VojtaStavik: For sure!! It was just a random idea :) I approved the PR

PR #1238, opened 2021-06-30
[CIS-978] Fix deleted messages are visible in threads

b-onc:
We didn't include proper predicates for reply fetching

### DIFF
--- a/Sources/StreamChat/Database/DTOs/MessageDTO.swift
+++ b/Sources/StreamChat/Database/DTOs/MessageDTO.swift
@@ -90,18 +90,9 @@ class MessageDTO: NSManagedObject {
         return request
     }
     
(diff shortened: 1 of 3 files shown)


VojtaStavik (approved):
Nice catch 👍

Sources/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -133,6 +133,33 @@ class MessageDTO: NSManagedObject {
         ])
     }
     
+    /// Returns predicate with thread messages that should be shown in the thread.
+    static func threadRepliesPredicate(for messageId: MessageId) -> NSCompoundPredicate {
```
VojtaStavik, line 137: Since this logic is very similar not channel messages, should we find a way how to share this logic to prevent similar inconsistencies in the future?

PR #1246, opened 2021-07-02
[CIS-974] Messages maintain correct order of the local time is off

### LINKED ISSUE #1231: Comments and replies not showing in chronological order (IOS), see the screenshot for reference
## What did you do?
   I have done a chat between two users, both are IOS devices.

    Comments and replies should be  in chronological order
    
## What happened instead?
  Comments and replies not showing in chronological order

## GetStream Environment
**GetStream Chat version:**
**GetStream Chat frameworks:** StreamChat, StreamChatUI, StreamChatClient, StreamChatCore
**iOS version:**
**Swift version:**
**Xcode version:**
**Device:**

## Additional context
Simulator Screen Shot - iPhone 12 - 2021-06-29 at 08 22 18

VojtaStavik:
[Fixes #1231]

We didn't reset `locallyCreatedAt` value after the message was sent which was causing the wrong ordering of the messages when the local time was different from the server time.

I also added an artificial time offset for the unsent messages sorting key such that the most recent message is always at the bottom of the list.

### DIFF
--- a/Sources/StreamChat/Database/DTOs/MessageDTO.swift
+++ b/Sources/StreamChat/Database/DTOs/MessageDTO.swift
@@ -51,7 +51,7 @@ class MessageDTO: NSManagedObject {
         
         prepareDefaultSortKeyIfNeeded()
     }
-    
+
     /// Makes sure the `defaultSortingKey` value is computed and set.
     fileprivate func prepareDefaultSortKeyIfNeeded() {
         let newSortingKey = locallyCreatedAt ?? createdAt
@@ -279,6 +279,7 @@ extension NSManagedObjectContext: MessageDatabaseSession {
         showReplyInChannel: Bool,
         isSilent: Bool,
         quotedMessageId: MessageId?,
+        createdAt: Date?,
         extraData: ExtraData
     ) throws -> MessageDTO {
         guard let currentUserDTO = currentUser else {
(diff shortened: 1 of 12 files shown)


Sources/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -51,10 +51,15 @@ class MessageDTO: NSManagedObject {
         
         prepareDefaultSortKeyIfNeeded()
     }
-    
+
+    /// We use this value as the offset for the sorting key for unsent messages. Even if the local time of the device
+    /// is different from the server time, we want to be sure the newly sent message is always the last one. By applying
+    /// the same offset to all unsent message, we make sure the order of the messages is maintained.
+    static let localSortingKeyTimeOffset: TimeInterval = 24 * 60 * 60
```
tbarbugli, line 58: I don't think we should do this, in some cases the failed message will make no sense if re-ordered (in the meantime newer messages will be in the channel).

Sources/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -51,10 +51,15 @@ class MessageDTO: NSManagedObject {
...
+
     /// Makes sure the `defaultSortingKey` value is computed and set.
     fileprivate func prepareDefaultSortKeyIfNeeded() {
-        let newSortingKey = locallyCreatedAt ?? createdAt
+        let newSortingKey = locallyCreatedAt.map { $0.addingTimeInterval(Self.localSortingKeyTimeOffset) } ?? createdAt
```
tbarbugli, line 62: disclaimer: this is unrelated to your change and possible N/A

From a db perspective computing the sort key sounds is very inefficient (all rows/objects will get a computed value which will be used to sort, gets slower as you data set grows). You normally want your sort key to come from an indexed column

VojtaStavik: The value is not computed on the fly. We add an additional column that's re-computed every time the object is changed (see the `willSave` method) and the db uses this value for sorting. This is a limitation of CoreData which doesn't allow sorting by more complex expression (see also channels where we sort by `max(createdAt, lastMessageAt)`.

However, this column is not indexed atm and it should be. Fixing 👍

Sources/StreamChat/Database/DTOs/MessageDTO.swift
VojtaStavik: Good point 👍 We can either ignore this edge case (all iPhones have the time synced remotely by default - you have to explicitly turn this off and then change the time), or we can just make sure the `locallyCreatedAt` is always bigger than the last message in the channel we locally have. WDYT?

VojtaStavik (commented):
LGTM, thank you!

Issue #1234, opened 2021-06-29
Does V4 support ChatMessageAttachment ExtraData?


srgray:
## What did you do?
I'm trying to access custom properties on an `Attachment`. In V2 SDK, I used `Attachment.extraDataType` to accomplish this. Does V4 SDK have that capability? Similar to `MessageExtraData` in V4?

## GetStream Environment
**GetStream Chat version:** 4.0.0-beta.4
**GetStream Chat frameworks:** StreamChat, StreamChatUI

VojtaStavik:
Hi @srgray. v4 has much more advanced and (I hope) better support for working with custom attachments. You can find more information about this topic in the WIP guide https://github.com/GetStream/stream-chat-swift/blob/main/docusaurus/docs/iOS/guides/working-with-attachments.md .

srgray:
@VojtaStavik Thanks for the link to docs. Can you help clarify something for me?
We have a process which adds messages with attachments to Stream channels. Here is a snippet of how we do that:

```
const attachment = {
            type: mediaType,
            thumb_url: thumbnailUrl,
            cnMediaId: mediaId,
            cnComments: commentCount,
        }
        if (mediaType === 'image') attachment['image_url'] = url
        if (mediaType === 'video') attachment['video_url'] = url
        const message = {
            text: caption || '',
            user_id: userId,
            attachments: [attachment],
        }
```

What we want is a `video` or `image` attachment with 2 extra properties: `cnMediaId` and `cnComments`. After reading through those docs, I don't see how I can access those 2 extra properties with the Stream V4 SDK _unless_ I use a custom attachment type. Is that correct? We were hoping to use the standard attachment types of `video` and `image` and still be able to access those 2 extra properties.

PR #1248, opened 2021-07-02
[CIS-989] Expose extra data on built-in attachments

evsaev:
**This PR** addresses https://github.com/GetStream/stream-chat-swift/issues/1234 and allows to send in-rich built-in attachments with custom fields and parse it back.

To extend an image attachment with a custom fields declare a `Codable` type first:
```swift
struct CommentPayload: Codable {
 let comment: String
}
```

To send an image attachment with custom payload do the following:
```swift
// 1. Create custom payload.
let commentPayload = CommentPayload(comment: "Happy birthday! 🎉")
// 2. Create an envelope for built-in attachment with custom payload.
let envelope = AnyAttachmentEnvelope(localFileURL: *url*, attachmentType: .image, extraData: commentPayload)
// 3. Send the message with attachment
channelController.createNewMessage(..., attachments: [envelope])
```

When message is sent an attachment JSON would looks like this:
```swift
{
 "type": "image",
 "title": "cake.png",
 "image_url": "https://getstream.io/attachments/cake.png",
 "thumb_url": "https://getstream.io/attachments/cake_thumb.png",
 "comment": "Happy birthday! 🎉"
}
```

Custom `comment` field can now be accessed as shown below :
```swift
// 1. Get an attachment from the message
let attachment = message.imageAttachments.first

// 2. Decode attachment extra data as your payload type.
let commentPayload = attachment?.extraData(ofType: CommentPayload.self)

// 3. Work with custom fields
print(commentPayload?.comment ?? "There's no comment :(") // Happy birthday! 🎉
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### ⚠️ Breaking Changes from `4.0-beta.4`
 - The `CreateChatChannelButton` component was removed. The component acted only as a placeholder and the functionality should be always provided by the hosting app. For an example implementation see the [Demo app](https://github.com/GetStream/stream-chat-swift/blob/main/DemoApp/ChatPresenter.swift).
+- The payload of `AnyChatMessageAttachment` changed from `Any` to `Data` [#1248](https://github.com/GetStream/stream-chat-swift/pull/1248).
 
 ### ✅ Added
 - `search(query:)` function to `UserSearchController` to make a custom search with a query [#1206](https://github.com/GetStream/stream-chat-swift/issues/1206)
@@ -16,6 +17,25 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   -  `connectUser`
   -  `connectGuestUser`
   -  `connectAnonymousUser`
+- Extra data support for `.image/.video/.file` attachments [#1248](https://github.com/GetStream/stream-chat-swift/pull/1248):
+1. Declare custom extra data type conforming to `Codable`:
+```swift
+struct CommentPayload: Codable {
+    let comment: String
+}
+```
+2. Create an instance of extra data and provide it to an envelope for `.image/.video/.file` attachment:
+```swift
+let commentPayload = CommentPayload(comment: "Happy birthday! 🎉")
+let envelope = try AnyAttachmentEnvelope(localFileURL: *url*, attachmentType: .image, extraData: commentPayload)
+channelController.createNewMessage(..., attachments: [envelope])
+```
+3. Receive extra data back from attachment:
+```swift
+if let commentPayload = message.imageAttachments.first?.extraData(ofType: CommentPayload.self) {
+    print(commentPayload.comment") // Happy birthday! 🎉
+}
+```
 
 ### 🔄 Changed
 - `shouldConnectAutomatically` setting in `ChatConfig`, it now has no effect and all logic that used it now behaves like it was set to `true`.
--- a/Sources/StreamChat/Models/Attachments/ChatMessageFileAttachment.swift
+++ b/Sources/StreamChat/Models/Attachments/ChatMessageFileAttachment.swift
@@ -43,18 +43,10 @@ extension FileAttachmentPayload: Encodable {
 extension FileAttachmentPayload: Decodable {
     public init(from decoder: Decoder) throws {
         let container = try decoder.container(keyedBy: AttachmentCodingKeys.self)
-
-        guard
-            let assetURL = try container
-            .decodeIfPresent(String.self, forKey: .assetURL)?
-            .attachmentFixedURL
-        else {
-            throw ClientError.AttachmentDecoding("File attachment must contain `assetURL`")
-        }
-
+        
         self.init(
             title: try container.decodeIfPresent(String.self, forKey: .title),
-            assetURL: assetURL,
+            assetURL: try container.decode(URL.self, forKey: .assetURL),
             file: try AttachmentFile(from: decoder)
         )
     }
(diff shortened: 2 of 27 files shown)


tbarbugli:
> **This PR** allows to get access to custom fields sent with built-in attachment type.

Can you add some examples here on how this looks like? Just some Swift examples here would help (ideally with the JSON payload next to it).
 
> Decoding of built-in attachments is postponed keeping a payload as `Data` by the time it's accessed (just like for custom attachments). This allows to treat attachment of built-in type (`.image/.video/.file`) as a custom one giving access to custom fields.

Is there any reason we dont expose not-captured fields into a map? (ie. if you add a "comment" field to an attachment of type=image it would be convenient to have that exposed in imageAttachment.data["comment"])

b-onc (approved):
I'd love to see the resulting API in the PR desc, and also a CHANGELOG entry is needed for this addition

evsaev:
> Is there any reason we dont expose not-captured fields into a map? (ie. if you add a "comment" field to an attachment of type=image it would be convenient to have that exposed in imageAttachment.data["comment"])

That's a valid comment 👍 

The map we wanna expose on built-in attachment should be of `[String: Any]` type `Any` is not codable this map cannot be encoded/decoded from JSON directly. 

We solved this for attachments using `RawJSON` type which can represents any JSON and conforms to `Encodable/Decodable` protocols:

```swift
enum RawJSON: Codable, Hashable {
    case double(Double)
    case string(String)
    case integer(Int)
    case bool(Bool)
    case dictionary([String: RawJSON])
    case array([RawJSON])
    case `nil`
}

struct MessageAttachmentPayload {
    let type: AttachmentType
    /// A raw attachment payload data.
    /// It's possible to have attachments of custom type with unknown structure
    /// so we need to keep in raw data form so it will be possible to decode later.
    let payload: RawJSON
}
```

But the issue is that it's a nested enum and it has no friendly API. Exposing `customFields: RawJSON` on attachment should require some convenient helpers. Checking now how we can approach this to make the API user-friendly

Sources/StreamChat/Models/Attachments/ChatMessageImageAttachment.swift
```diff
@@ -42,19 +50,16 @@ extension ImageAttachmentPayload: Encodable {
...
-            .decodeIfPresent(String.self, forKey: .thumbURL)?
-            .attachmentFixedURL
+        guard case let .dictionary(payload) = try RawJSON(from: decoder) else {
+            throw ClientError.AttachmentDecoding("Failed to parse image payload extra data.")
+        }
 
+        let imageURL = try
+            container.decodeIfPresent(URL.self, forKey: .image) ??
+            container.decodeIfPresent(URL.self, forKey: .imageURL) ??
+            container.decode(URL.self, forKey: .assetURL)
```
evsaev, line 61: @tbarbugli are we doing the right thing when using a fallback chain for image URL and title?

evsaev:
@b-onc @tbarbugli addressed your comments and requested a review since a PR has changed a lot. Please have a look

Sources/StreamChat/Models/Attachments/ChatMessageFileAttachment.swift
```diff
@@ -22,6 +22,13 @@ public struct FileAttachmentPayload: AttachmentPayload {
     public internal(set) var assetURL: URL
     /// The file itself.
     public let file: AttachmentFile
+    
+    let _extraData: [String: RawJSON]
+    
+    /// An extra data.
+    public var extraData: [String: Any] {
+        _extraData.compactMapValues(\.value)
+    }
```
evsaev, line 31: We could make it smarter and expose generic API that would decode from `[String: RawJSON]`. 

Smth like:
```swift
public func extraData<T: Decodable>() -> T? {
 // try to decode `T` from [String: Any] 
}
```

Let me know if it worth trying

CHANGELOG.md
```diff
@@ -7,11 +7,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
 - The `CreateChatChannelButton` component was removed. The component acted only as a placeholder and the functionality should be always provided by the hosting app. For an example implementation see the [Demo app](https://github.com/GetStream/stream-chat-swift/blob/main/DemoApp/ChatPresenter.swift).
+- The payload of `AnyChatMessageAttachment` changed from `Any` to `Data` [#1248](https://github.com/GetStream/stream-chat-swift/pull/1248).
 
 ### ✅ Added
 - `search(query:)` function to `UserSearchController` to make a custom search with a query [#1206](https://github.com/GetStream/stream-chat-swift/issues/1206)
 - `queryForMentionSuggestionsSearch(typingMention:)` function to `ComposerVC`, users can override this function to customize mention search behavior [#1206](https://github.com/GetStream/stream-chat-swift/issues/1206)
-- `.contains` added to `Filter` to be able to filter for `teams` [#1206](https://github.com/GetStream/stream-chat-swift/issues/1206)
+- `.contains` added to `Filter` to be able to filter for `teams` [#1206](https://github.com/GetStream/stream-chat-swift/issues/1206) [#1206](https://github.com/GetStream/stream-chat-swift/issues/1206)
```
b-onc, line 15:
```suggestion
- `.contains` added to `Filter` to be able to filter for `teams` [#1206](https://github.com/GetStream/stream-chat-swift/issues/1206)
```

CHANGELOG.md
```diff
@@ -7,11 +7,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
+- Extra data support for `.image/.video/.file` attachments [#1248](https://github.com/GetStream/stream-chat-swift/pull/1248).
```
b-onc, line 16: We should briefly show how it's done here, too

CHANGELOG.md
evsaev: ✅ fixed

CHANGELOG.md
evsaev: ✅ explained in more detail

Issue #1165, opened 2021-06-08
Hide keyboard when tapped on message list


whiplash00:
## What did you do?
I am using StreamchatUI and _ChatMessageListVC as my view controller. Current way of hiding the keyboard is not typical of chat app (normally user expect to tap anywhere other than keyboard to hide keyboard, not do a drag up action) hence I added  the hideKeyboardWhenTappedAround() to _ChatMessageListVC.

## What did you expect to happen?
That everything will work as per normal.

## What happened instead?
The keyboard is indeed hidden when simply tapping anywhere, however the @mention is broken. The @mention user selection menu is opened but cannot be clicked on.

## GetStream Environment
GetStream Chat version: 4.0.0
GetStream Chat frameworks: StreamChatUI - 4.0.0
iOS version: 14.5.1
Swift version: 5.3.3
Xcode version: 12.5
Device: iphone 12 Pro Max

b-onc:
Hello @whiplash00 ,

Can you share your `hideKeyboardWhenTappedAround` implementation?

whiplash00:
@b-onc sorry didnt realize `hideKeyboardWhenTappedAround` is my own implementation. Here you go:

```
extension UIViewController {
    func hideKeyboardWhenTappedAround() {
        let tapGesture = UITapGestureRecognizer(target: self,
                                                action: #selector(hideKeyboard))
        view.addGestureRecognizer(tapGesture)
    }
    
    @objc func hideKeyboard() {
        view.endEditing(true)
    }
}
```

whiplash00:
@b-onc I think this should be default behaviour? Should be able to tap anywhere to hide keyboard. Just my one cent. Cheers.

b-onc:
Hello @whiplash00 ,

Thanks for the suggestion. We think this makes sense too, so we're adding it to our backlog. I'm keeping this issue open, converted to a feature request.

whiplash00:
@b-onc Thank you!

PR #1253, opened 2021-07-07
Dismiss the keyboard on tapping anywhere on the message list

SagarSDagdu:
## What does this PR do

- Whenever the user taps anywhere on the message list, the keyboard is dismissed. This PR closes https://github.com/GetStream/stream-chat-swift/issues/1165


VojtaStavik (approved):
Looks good 👍 We need the same functionality for the thread VC, too.

nuno-vieira (approved):
LGTM! ✅

PR #1254, opened 2021-07-07
[CIS-877] User typing events

skorepak:
## This PR
- replaces `currentlyTypingMebmers` with `currentlyTypingUsers`
- changes type from `ChatChannelMember` to `ChatUser`

## Links
https://stream-io.atlassian.net/browse/CIS-877


Sources/StreamChat/Models/Channel.swift
```diff
@@ -71,9 +71,9 @@ public struct _ChatChannel<ExtraData: ExtraDataTypes> {
     public var lastActiveMembers: [_ChatChannelMember<ExtraData.User>] { _lastActiveMembers }
     @CoreDataLazy private var _lastActiveMembers: [_ChatChannelMember<ExtraData.User>]
     
-    /// A list of currently typing channel members.
-    public let currentlyTypingMembers: Set<_ChatChannelMember<ExtraData.User>>
-
+    /// A list of currently typing users.
+    public let currentlyTypingUsers: Set<_ChatUser<ExtraData.User>>
```
evsaev, line 75: would be nice to have backing `@CoreDataLazy` field for this

evsaev (approved):
Looking good, thanks 👍 There's just one comment about lazy loading of typing users

Sources/StreamChat/Models/Channel.swift
skorepak: done 👍🏻 , please check

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
```diff
@@ -370,7 +370,7 @@ extension _ChatChannel {
             isFrozen: dto.isFrozen,
             lastActiveMembers: { fetchMembers() },
             membership: dto.membership.map { $0.asModel() },
-            currentlyTypingUsers: Set(typingUsers),
+            currentlyTypingUsers: { Set(typingUsers) },
```
evsaev, line 373: hm, to keep lazy loading working we should covert user DTOs to models inside the closure, I think 🤔 Otherwise faults are evaluated at the time we create a channel model, not when typing users are requested

Sources/StreamChat/Models/Channel.swift
VojtaStavik: Can you please add the `currentlyTypingMembers` variable into `Deprecations.swift` (with empty implementation) and mark it as renamed? It will help customers with the transition because they will be able to apply fixits.

CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 # Upcoming
 
+### ⚠️ Breaking Changes from `4.0-beta.5`
+- Changed ChannelDTO from  `currentlyTypingMembers: Set<MemberDTO>` to `currentlyTypingUsers: Set<UserDTO>` to show all typing users (not only channel members; eg: watching users) [#1254](https://github.com/GetStream/stream-chat-swift/pull/1254)
```
VojtaStavik, line 9: DTOs are internal types, please refer to the public API that was changed

VojtaStavik (approved):
LGTM 👍 Let's wait with merging this since we have to release a hotfix and we shouldn't include breaking changes there.

PR #1256, opened 2021-07-08
[CIS-994] Fix the issue with receiving .didBecomeActive before connecting the client for the first time

dmigach:
We spotted one additional case when it's possible for `ChatClientUpdater.connect` to be called before obtaining credentials:
https://github.com/GetStream/stream-chat-swift/issues/1218

It happens because we subscribe for notifications in `ChatClient`'s init
Two fixes to prevent this:
1) Do not call `ChatClientUpdater.connect` if one of  `connect` methods has not been called on `ChatClient`
2) Only listen to app state notifications in connected state, i.e. only subscribe when we connect and unsubscribe on disconnect

### DIFF
--- a/Sources/StreamChat/ChatClient.swift
+++ b/Sources/StreamChat/ChatClient.swift
@@ -299,11 +299,6 @@ public class _ChatClient<ExtraData: ExtraDataTypes> {
         self.eventWorkerBuilders = eventWorkerBuilders
 
         currentUserId = fetchCurrentUserIdFromDatabase()
-        
-        backgroundTaskScheduler?.startListeningForAppStateUpdates(
-            onEnteringBackground: { [weak self] in self?.handleAppDidEnterBackground() },
-            onEnteringForeground: { [weak self] in self?.handleAppDidBecomeActive() }
-        )
     }
     
     deinit {
@@ -363,6 +358,11 @@ public class _ChatClient<ExtraData: ExtraDataTypes> {
     /// are received.
     public func disconnect() {
         clientUpdater.disconnect()
+        
+        backgroundTaskScheduler?.startListeningForAppStateUpdates(
+            onEnteringBackground: {},
+            onEnteringForeground: {}
+        )
     }
 
     func fetchCurrentUserIdFromDatabase() -> UserId? {
@@ -413,6 +413,11 @@ public class _ChatClient<ExtraData: ExtraDataTypes> {
             userConnectionProvider: userConnectionProvider,
             completion: completion
         )
+        
+        backgroundTaskScheduler?.startListeningForAppStateUpdates(
+            onEnteringBackground: { [weak self] in self?.handleAppDidEnterBackground() },
+            onEnteringForeground: { [weak self] in self?.handleAppDidBecomeActive() }
+        )
     }
     
     private func handleAppDidEnterBackground() {
@@ -441,6 +446,11 @@ public class _ChatClient<ExtraData: ExtraDataTypes> {
     private func handleAppDidBecomeActive() {
         cancelBackgroundTaskIfNeeded()
 
+        guard userConnectionProvider != nil else {
+            // The client has not been connected yet during this session
+            return
+        }
+        
         guard connectionStatus != .connected && connectionStatus != .connecting else {
             // We are connected or connecting anyway
             return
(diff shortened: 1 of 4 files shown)


Sources/StreamChat/ChatClient.swift
```diff
@@ -363,6 +358,11 @@ public class _ChatClient<ExtraData: ExtraDataTypes> {
     /// are received.
     public func disconnect() {
         clientUpdater.disconnect()
+        
+        backgroundTaskScheduler?.startListeningForAppStateUpdates(
```
dmigach, line 362: Not sure if it's ok or we better add a `BackgroundTaskScheduler.stopListeningForAppStateUpdates` and incapsulate it in there

b-onc: `stopListeningForAppStateUpdates` would be a good addition

Sources/StreamChat/ChatClient.swift
```diff
@@ -441,6 +446,11 @@ public class _ChatClient<ExtraData: ExtraDataTypes> {
     private func handleAppDidBecomeActive() {
         cancelBackgroundTaskIfNeeded()
 
+        guard userConnectionProvider != nil else {
+            // The client has not been connected yet during this session
+            return
+        }
```
b-onc, line 452: Isn't this enough to fix the issue? then we don't need `stopListeningForAppUpdates` or changing the place of `startListening...`

dmigach: It is enough, but we shouldn't connect on going back active if the client was disconnected manually

b-onc: There's no `connect` without setting user info anymore, so on explicit disconnect we can clear `userConnectionProvider`

dmigach: This is true, though don't you think it adds an implicit assumption? Subscribe/unsubscribe is not 100% robust neither, but it's more explicit

b-onc: the implicit assumption is made in the `guard` statement above anyway ¯\_(ツ)_/¯ up to you, both approaches seem fine at first sight

dmigach: Yeah this is also true 🤔 So are we good with leaving both safety mechanisms in place? I'm adding `stopListeningForAppStateUpdates`, can set `userConnectionProvider ` to `nil` on disconnecting as well

b-onc: fine by me

Sources/StreamChat/ChatClient.swift
```diff
@@ -413,6 +410,11 @@ public class _ChatClient<ExtraData: ExtraDataTypes> {
             userConnectionProvider: userConnectionProvider,
             completion: completion
         )
+        
+        backgroundTaskScheduler?.startListeningForAppStateUpdates(
+            onEnteringBackground: { [weak self] in self?.handleAppDidEnterBackground() },
+            onEnteringForeground: { [weak self] in self?.handleAppDidBecomeActive() }
+        )
```
VojtaStavik, line 417: Shouldn't this happen only if the connection is successful?

Sources/StreamChat/ChatClient_Tests.swift
```diff
@@ -858,6 +862,7 @@ class ChatClient_Tests: StressTestCase {
             eventWorkerBuilders: [],
             environment: testEnv.environment
         )
+        client.connectAnonymousUser()
```
VojtaStavik, line 865: why this change?

VojtaStavik (approved):

Sources/StreamChat/ChatClient_Tests.swift
dmigach: We need to connect before we can simulate going to background. Basically all the existing tests that test the way we handle config settings and implementation details of going background/active now need to connect first

Sources/StreamChat/ChatClient.swift
dmigach: I think you're right, there is one interesting case when we go background not having connected yet, but current implementation ignores it anyway
```
    private func handleAppDidEnterBackground() {
        // We can't disconnect if we're not connected
        guard connectionStatus == .connected else { return }
    ...
```

PR #1258, opened 2021-07-08
[CIS-588] Add support for pasting images in the composer

SagarSDagdu:
### What does this PR do?

- This PR allows the user to copy images from anywhere and paste them into the composer.
- This support is limited to images currently, as the pasteboard readily provides us the copied image

https://user-images.githubusercontent.com/11807135/124904280-76a4cb80-e002-11eb-98d4-cdce8322c9d1.mp4

### DIFF
--- a/Sources/StreamChatUI/CommonViews/InputTextView/InputTextView.swift
+++ b/Sources/StreamChatUI/CommonViews/InputTextView/InputTextView.swift
@@ -2,12 +2,28 @@
 // Copyright © 2021 Stream.io Inc. All rights reserved.
 //
 
+import MobileCoreServices
 import StreamChat
 import UIKit
+
+/// The delegate of the `InputTextView` that notifies when an `UIImage` is pasted.
+public protocol InputTextViewImagePasteDelegate: AnyObject {
+    /// Notifies that an `UIImage` has been pasted into the text view
+    /// - Parameters:
+    ///   - inputTextView: The `InputTextView` in which the image was pasted
+    ///   - image: The `UIImage`
+    func inputTextView(_ inputTextView: InputTextView, didPasteImage image: UIImage)
+}
 
 /// A view for inputting text with placeholder support. Since it is a subclass
 /// of `UITextView`, the `UITextViewDelegate` can be used to observe text changes.
 open class InputTextView: UITextView, AppearanceProvider {
+    /// The delegate which gets notified when an image is pasted into the text view
+    open weak var imagePasteDelegate: InputTextViewImagePasteDelegate?
+    
+    /// Whether this text view should allow images to be pasted
+    open var allowsPastingImages: Bool = true
+    
     /// Label used as placeholder for textView when it's empty.
     open private(set) lazy var placeholderLabel: UILabel = UILabel()
         .withoutAutoresizingMaskConstraints
@@ -90,5 +106,24 @@ open class InputTextView: UITextView, AppearanceProvider {
         
     @objc open func handleTextChange() {
         placeholderLabel.isHidden = !text.isEmpty
+    }
+    
+    // MARK: - Actions on the UITextView
+    
+    override open func canPerformAction(_ action: Selector, withSender sender: Any?) -> Bool {
+        // If action is paste and the pasteboard has an image, we allow it
+        if action == #selector(paste(_:)) && allowsPastingImages && UIPasteboard.general.hasImages {
+            return true
+        }
+
+        return super.canPerformAction(action, withSender: sender)
+    }
+
+    override open func paste(_ sender: Any?) {
+        if let pasteboardImage = UIPasteboard.general.image {
+            imagePasteDelegate?.inputTextView(self, didPasteImage: pasteboardImage)
+        } else {
+            super.paste(sender)
+        }
     }
 }
--- a/Sources/StreamChatUI/CommonViews/InputTextView/InputTextView.swift
+++ b/Sources/StreamChatUI/CommonViews/InputTextView/InputTextView.swift
@@ -2,12 +2,11 @@
 // Copyright © 2021 Stream.io Inc. All rights reserved.
 //
 
-import MobileCoreServices
 import StreamChat
 import UIKit
 
-/// The delegate of the `InputTextView` that notifies when an `UIImage` is pasted.
-public protocol InputTextViewImagePasteDelegate: AnyObject {
+/// The delegate of the `InputTextView` that notifies when an attachment is pasted in the text view.
+public protocol InputTextViewClipboardAttachmentDelegate: AnyObject {
     /// Notifies that an `UIImage` has been pasted into the text view
     /// - Parameters:
     ///   - inputTextView: The `InputTextView` in which the image was pasted
@@ -18,8 +17,8 @@ public protocol InputTextViewImagePasteDelegate: AnyObject {
 /// A view for inputting text with placeholder support. Since it is a subclass
 /// of `UITextView`, the `UITextViewDelegate` can be used to observe text changes.
 open class InputTextView: UITextView, AppearanceProvider {
-    /// The delegate which gets notified when an image is pasted into the text view
-    open weak var imagePasteDelegate: InputTextViewImagePasteDelegate?
+    /// The delegate which gets notified when an attachment is pasted into the text view
+    open var clipboardAttachmentDelegate: InputTextViewClipboardAttachmentDelegate?
     
     /// Whether this text view should allow images to be pasted
     open var allowsPastingImages: Bool = true
@@ -121,7 +120,7 @@ open class InputTextView: UITextView, AppearanceProvider {
 
     override open func paste(_ sender: Any?) {
         if let pasteboardImage = UIPasteboard.general.image {
-            imagePasteDelegate?.inputTextView(self, didPasteImage: pasteboardImage)
+            clipboardAttachmentDelegate?.inputTextView(self, didPasteImage: pasteboardImage)
         } else {
             super.paste(sender)
         }
--- a/Sources/StreamChatUI/CommonViews/InputTextView/InputTextView.swift
+++ b/Sources/StreamChatUI/CommonViews/InputTextView/InputTextView.swift
@@ -18,10 +18,10 @@ public protocol InputTextViewClipboardAttachmentDelegate: AnyObject {
 /// of `UITextView`, the `UITextViewDelegate` can be used to observe text changes.
 open class InputTextView: UITextView, AppearanceProvider {
     /// The delegate which gets notified when an attachment is pasted into the text view
-    open var clipboardAttachmentDelegate: InputTextViewClipboardAttachmentDelegate?
+    open weak var clipboardAttachmentDelegate: InputTextViewClipboardAttachmentDelegate?
     
     /// Whether this text view should allow images to be pasted
-    open var allowsPastingImages: Bool = true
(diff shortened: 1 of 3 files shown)


Sources/StreamChatUI/Composer/ComposerVC.swift
```diff
@@ -785,4 +794,24 @@ open class _ComposerVC<ExtraData: ExtraDataTypes>: _ViewController,
         alert.addAction(.init(title: L10n.Alert.Actions.ok, style: .default, handler: { _ in }))
         present(alert, animated: true)
     }
+    
+    // MARK: - InputTextViewImagePasteDelegate
+    
+    open func inputTextView(_ inputTextView: InputTextView, didPasteImage image: UIImage) {
+        do {
+            let imageUrl = try image.temporaryLocalFileUrl()
```
VojtaStavik, line 802: The function is throwing and it still returns an optional URL?

Sources/StreamChatUI/Composer/ComposerVC.swift
```diff
@@ -785,4 +794,24 @@ open class _ComposerVC<ExtraData: ExtraDataTypes>: _ViewController,
...
+            let urlAndType: (URL, AttachmentType) = (imageUrl!, .image)
```
VojtaStavik, line 803: - Are you sure the `!` is safe here?
- Why do you create this tuple? It's not used in the code as a tuple anywhere.

Sources/StreamChatUI/Composer/ComposerVC.swift
SagarSDagdu: Yes. It throws in case of a write failure but returns optional in case we could not create a jpeg representation of the image.

Sources/StreamChatUI/Composer/ComposerVC.swift
```diff
@@ -785,4 +794,24 @@ open class _ComposerVC<ExtraData: ExtraDataTypes>: _ViewController,
...
+            let fileSize = try AttachmentFile(url: urlAndType.0).size
+            let maxFileSize = channelController?.client.config.maxAttachmentSize ?? 0
+            
+            if fileSize < maxFileSize {
+                let attachment = try AnyAttachmentPayload(localFileURL: urlAndType.0, attachmentType: urlAndType.1)
+                content.attachments.append(attachment)
+            } else {
+                showAttachmentExceedsMaxSizeAlert()
+            }
```
evsaev, line 812: the same attachment size validation logic is used for image picker selection. Does it makes sense to encapsulate it in some func and reuse?

Sources/StreamChatUI/Composer/ComposerVC.swift
SagarSDagdu: Fixed the force unwrap, it wasn't safe in that context. Also removed the tuple

Sources/StreamChatUI/Composer/ComposerVC.swift
SagarSDagdu: Added a common function

Sources/StreamChatUI/Composer/ComposerVC.swift
```diff
@@ -731,10 +760,7 @@ open class _ComposerVC<ExtraData: ExtraDataTypes>: _ViewController,
         _ picker: UIImagePickerController,
         didFinishPickingMediaWithInfo info: [UIImagePickerController.InfoKey: Any]
     ) {
-        var completion: (() -> Void)?
-        defer {
-            picker.dismiss(animated: true, completion: completion)
-        }
+        picker.dismiss(animated: true, completion: nil)
```
evsaev, line 763: does showing max size alert work fine without waiting for a picket to be dismissed?
```suggestion
        picker.dismiss(animated: true) {
            //do the processing after the picket is dismissed?
        }
```

SagarSDagdu: Yes, better to wait for its complete dismissal. Changed it

Sources/StreamChatUI/CommonViews/InputTextView/InputTextView.swift
```diff
@@ -2,12 +2,28 @@
 // Copyright © 2021 Stream.io Inc. All rights reserved.
 //
 
+import MobileCoreServices
 import StreamChat
 import UIKit
 
+/// The delegate of the `InputTextView` that notifies when an `UIImage` is pasted.
+public protocol InputTextViewImagePasteDelegate: AnyObject {
```
evsaev, line 10:
```suggestion
public protocol InputTextViewDelegate: AnyObject {
```

Sources/StreamChatUI/CommonViews/InputTextView/InputTextView.swift
```diff
@@ -2,12 +2,28 @@
...
+    ///   - inputTextView: The `InputTextView` in which the image was pasted
+    ///   - image: The `UIImage`
+    func inputTextView(_ inputTextView: InputTextView, didPasteImage image: UIImage)
+}
+
 /// A view for inputting text with placeholder support. Since it is a subclass
 /// of `UITextView`, the `UITextViewDelegate` can be used to observe text changes.
 open class InputTextView: UITextView, AppearanceProvider {
+    /// The delegate which gets notified when an image is pasted into the text view
+    open weak var imagePasteDelegate: InputTextViewImagePasteDelegate?
```
evsaev, line 22: hm... how about
```suggestion
    open var inputViewViewDelegate: InputTextViewImagePasteDelegate? {
     delegate as? InputTextViewDelegate
    }
```

Sources/StreamChatUI/CommonViews/InputTextView/InputTextView.swift
nuno-vieira: Why would we want this? The `InputTextView` is already a subclass of `UITextView` so we already have the `delegate: UITextViewDelegate?` out of the box.

Sources/StreamChatUI/CommonViews/InputTextView/InputTextView.swift
nuno-vieira: I think merging the delegates, will be confusing IMO. Because we will still have the `inputTextView.delegate` and then the `inputTextView.inputTextViewDelegate` too. 

Maybe can make the other delegate a bit more generic like `clipboardAttachmentDelegate` something like this.. 🤔 Where we put all possible attachments from the clipboard in one delegate

Sources/StreamChatUI/Composer/ComposerVC.swift
```diff
@@ -228,6 +229,9 @@ open class _ComposerVC<ExtraData: ExtraDataTypes>: _ViewController,
         super.setUp()
 
         composerView.inputMessageView.textView.delegate = self
+        
+        // Set the delegate for handling the pasting of UIImages in the text view
+        composerView.inputMessageView.textView.imagePasteDelegate = self
```
nuno-vieira, line 234: We probably going to handle more attachment types, like pasting videos, etc.. we should try to make the delegate a bit more generic 👍

Sources/StreamChatUI/Composer/ComposerVC.swift
```diff
@@ -703,6 +712,26 @@ open class _ComposerVC<ExtraData: ExtraDataTypes>: _ViewController,
...
+        do {
+            let fileSize = try AttachmentFile(url: url).size
+            let maxFileSize = channelController?.client.config.maxAttachmentSize ?? 0
+            
+            if fileSize < maxFileSize {
+                let attachment = try AnyAttachmentPayload(localFileURL: url, attachmentType: type)
+                content.attachments.append(attachment)
+            } else {
+                showAttachmentExceedsMaxSizeAlert()
+            }
```
nuno-vieira, line 730: This should throw an error, because then the function name is not right, it is doing 2 things.
- Adding the attachment to content
- Showing an alert if the max file size is reached

So there's a side effect that is not clear from the function name. I would recommend throwing an error, and the caller of this function should present the alert if it wants, not this function 🙂

Sources/StreamChatUI/CommonViews/InputTextView/InputTextView.swift
SagarSDagdu: I think we should keep the delegates separate as @nuno-vieira suggests. But yes, we can make the other delegate more generic, like: `InputTextViewClipboardAttachmentDelegate`. This will currently contain only one method for the image, and we may add more in the future as needed. Will this be fine?

evsaev: With a protocol inherited from the base one there's no need in additional delegate field (the casted delegate is computed and can be private) and no need to assign multiple delegates to handle all available callbacks. This is how delegates are extended in UIKit (UITableViewDelegate: ScrollViewDelegate) and I find it useful, tbh.

If customer decide to have a custom `InputTextView` with additional callback it'd would be simple:

```swift
protocol MyInputTextViewDelegate: InputTextViewDelegate {
 func callback()
}

class MyInputTextView: InputTextView {
 func smthDidHappen() {
 (delegate as? MyInputTextViewDelegate).callback()
 }
}

class MyComposerVC: ComposerVC, MyInputTextViewDelegate {
  func callback() { ... }
}
```

evsaev: it works great in message list with custom attachments but here custom delegate is an edge case. both approaches work for me so feel free to go with one you like more 👍

Sources/StreamChatUI/CommonViews/InputTextView/InputTextView.swift
```diff
@@ -5,9 +5,24 @@
...
+    ///   - inputTextView: The `InputTextView` in which the image was pasted
+    ///   - image: The `UIImage`
+    func inputTextView(_ inputTextView: InputTextView, didPasteImage image: UIImage)
+}
+
 /// A view for inputting text with placeholder support. Since it is a subclass
 /// of `UITextView`, the `UITextViewDelegate` can be used to observe text changes.
 open class InputTextView: UITextView, AppearanceProvider {
+    /// The delegate which gets notified when an attachment is pasted into the text view
+    open var clipboardAttachmentDelegate: InputTextViewClipboardAttachmentDelegate?
```
evsaev, line 21: weak

evsaev (approved):
🚢

Sources/StreamChatUI/CommonViews/InputTextView/InputTextView.swift
```diff
@@ -5,9 +5,24 @@
...
+    
+    /// Whether this text view should allow images to be pasted
+    open var allowsPastingImages: Bool = true
```
nuno-vieira, line 24: I think we should try to make our boolean properties read as assertions like the Swift Style Guide, we already have a pattern for enabling/disabled things:
```suggestion
    /// Whether this text view should allow images to be pasted
    open var isPastingImagesEnabled: Bool = true
```

nuno-vieira (commented):
Only one more small change and then 🚢 🎉

Sources/StreamChatUI/CommonViews/InputTextView/InputTextView.swift
SagarSDagdu: Changed this according to the Swift Style guide

nuno-vieira (approved):
LGTM! ✅

PR #1262, opened 2021-07-09
[CIS-999] Check row is within bounds before scroll

evsaev:
**This PR** fixes the crash on message list by checking that the row we wanna scroll to exists.


VojtaStavik (approved):

elemanhillary:
This does not resolve the crash in '4.0.0-beta.6'

PR #1263, opened 2021-07-09
[CIS-982] Crash when thread root message is not loaded

evsaev:
**This PR** fixes crash on `ChatThreadVC` when root message is not loaded

### DIFF
--- a/Sources/StreamChatUI/ChatMessageList/ChatThreadVC.swift
+++ b/Sources/StreamChatUI/ChatMessageList/ChatThreadVC.swift
@@ -214,7 +214,26 @@ open class _ChatThreadVC<ExtraData: ExtraDataTypes>:
     }
     
     public var messages: [_ChatMessage<ExtraData>] {
-        messageController.replies + [messageController.message!]
+        /*
+         Thread replies are evaluated from DTOs when converting `messageController.replies` to an array.
+         Adding thread root message into replies would require `insert/append` API on lazy map which should
+         update both source collection and a cache to not break the indexing and keep 1-1 match with evaluated
+         and non-evaluated elements.
+         
+         We have evaluated thread root message in `messageController.message` but to get keep lazy map
+         working after an insert we also need an underlaying DTO to be added to source collection and it's getting
+         hard since the information about source collection `Element` type is available only during lazy map
+         initialization and does not get stored for later use.
+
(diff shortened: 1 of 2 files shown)


Sources/StreamChatUI/ChatMessageList/ChatThreadVC.swift
```diff
@@ -378,6 +384,25 @@ open class _ChatThreadVC<ExtraData: ExtraDataTypes>:
...
+        switch change {
+        case .create(let item):
+            listChange = .insert(item, index: indexPath)
+        case .update(let item):
+            listChange = .update(item, index: indexPath)
+        case .remove(let item):
+            listChange = .remove(item, index: indexPath)
+        }
+        
+        updateMessages(with: [listChange])
```
evsaev, line 404: we were supposed to call `reloadData` but `didChangeMessage` is triggered when thread root message is updated which can happen multiple times. Reloading single cell sounds better (the index calculations are straightforward since thread root is always at the end of message list)

Sources/StreamChatUI/ChatMessageList/ChatThreadVC.swift
```diff
@@ -214,7 +214,13 @@ open class _ChatThreadVC<ExtraData: ExtraDataTypes>:
...
     
     public var messages: [_ChatMessage<ExtraData>] {
-        messageController.replies + [messageController.message!]
+        let replies = Array(messageController.replies)
+        
+        if let threadRootMessage = messageController.message {
+            return replies + [threadRootMessage]
+        } else {
+            return replies
+        }
```
evsaev, line 223: Thread replies are evaluated from DTOs when converting `messageController.replies` to an array. The solution I was thinking of is appending thread root message to `replies: LazyCachedMapCollection` but seems it's not feasible 🤔 

To not break the indexes of lazy map we would have to modify the source collection of message DTOs and a cache. It's quite straightforward with cache but the information about source collection `Element` type is available only in `init` and doesn't get stored for later use.

Don't see how we can handle this nicely on UI side (on LLC side we could tweak a replies observer to also include a thread root though)

VojtaStavik (approved):
LGTM 👍 Your comments on this PR are valuable and I think they actually should be included in the source code, rather than here ;-)

PR #1266, opened 2021-07-12
[CIS-1005] Events observing

evsaev:
**This PR** introduces `EventsController` used to observe system events.

To start listening for events via the delegate:
```swift
class ThreadVC: ChatThreadVC, EventsControllerDelegate {
    // [1] Create `EventsController` instance, make sure to keep strong reference to the controller.
    lazy var eventsController = channelController.client.eventsController()
    
    override func setUp() {
        super.setUp()
        
        // [2] Set the delegate that will be notified about all incoming events
        eventsController.delegate = self
    }

     // [3] Provide `EventsControllerDelegate` implementation
        switch event {
        case let event as MessageUpdatedEvent where event.messageId == messageController.messageId:
            // handle updated event for concrete message
        case let event as TypingEvent where event.cid == channelController.cid
            // handle typing event for concrete channel
        default:
            break
        }
    }
}
```

To start listening for events via `Combine` (available starting from `iOS13`):
```swift
class ThreadVC: ChatThreadVC {
    // [1] Create array for subscriptions
    var subscriptions: [AnyCancellable] = []
    
    override func setUp() {
        super.setUp()
        
        // [2] Create `EventsController` instance
        let eventsController = channelController.client.eventsController()

        // [3] Use event publishers to subscribe on events you're interested in
        subscriptions = [
            // Subscribe on all events
            eventsController
                .allEventsPublisher
                .sink { print($0) },
            // Subscribe on updated event for concrete message.
            eventsController
                .eventPublisher(MessageUpdatedEvent.self)
                .filter { [messageId = messageController.messageId] in $0.messageId == messageId }
                .sink { print($0) }
            // Subscribe on typing event for concrete channel.
            eventsController
                .eventPublisher(TypingEvent.self)
                .filter { [cid = channelController.cid] in $0.cid == cid }
                .sink { print($0) }
        ]
    }
}

```


evsaev (commented):
The PR showcases 2 approaches to event exposure:

**Model specific controllers + delegate based approach**
[+] great autocompletion and syntax highlight
[+] event filtering is available out-of-the-box and happens inside a model controller

**Events controller + subscriptions**
[+] all events both system and custom are handled in a single place
[+] it's possible to subscribe on events related to different models with a single controller
[+] it's easy to write a custom wrapper around `subscribe` func with custom filter
[+] less boilerplate

Sources/StreamChat/Controllers/EventsController/EventsController.swift
```diff
@@ -0,0 +1,154 @@
...
+            handler(event)
+        }
+    }
+    
+    public func send<T: CustomEvent>(
+        event: T.Type = T.self,
+        completion: @escaping (Error?) -> Void
+    ) {
+        // TODO:
+    }
```
evsaev, line 59: @tbarbugli @VojtaStavik what is the endpoint for sending custom events? Is it possible to manually initiate/send a system event?

Sources/StreamChat/Controllers/EventsController/EventsController.swift
```diff
@@ -0,0 +1,154 @@
...
+    func subscribeOnMessageDeletedEvent(
+        for messageId: MessageId,
+        handler: @escaping (MessageDeletedEvent) -> Void
+    ) {
+        subscribe(
+            filter: { $0.messageId == messageId },
+            handler: handler
+        )
+    }
+}
```
evsaev, line 100: wrappers are mainly for autocompletion

Sources/StreamChat/Controllers/EventsController/EventsController.swift
```diff
@@ -0,0 +1,43 @@
+//
+// Copyright © 2021 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+public extension _ChatClient {
+    func createEventsController() -> EventsController {
+        .init(notificationCenter: eventNotificationCenter)
+    }
```
nuno-vieira, line 10:
```suggestion
    func makeEventsController() -> EventsController {
        .init(notificationCenter: eventNotificationCenter)
    }
```

Swift Guidelines specify that for factories we should use `make` instead of `create`

Sources/StreamChat/Controllers/MessageController/MessageController.swift
```diff
@@ -466,6 +470,31 @@ private extension _ChatMessageController {
         return observer
     }
     
+    func createMessageEventsObserver() -> EventObserver? {
```
nuno-vieira, line 473:
```suggestion
    func makeMessageEventsObserver() -> EventObserver? {
```

nuno-vieira:
It would be nice to showcase also the usages of each approach 👍

Sources/StreamChat/Controllers/EventsController/EventsController.swift
evsaev: decided to handle custom events in separate PR

Sources/StreamChat/Controllers/EventsController/EventsController.swift
evsaev: renamed to `eventsController` to follow the naming convention used for other controllers

Sources/StreamChat/Controllers/MessageController/MessageController.swift
evsaev: ✅ Done

evsaev:
> It would be nice to showcase also the usages of each approach 👍

Done, please have a look @nuno-vieira 🙂

[1] exposing models specific events on dedicated controllers leads to much boilerplate + not obvious how to get access to all events (seems we still need `EventsController` for this)
[2] in `EventsController + closures` approach we have to store subscriptions -> additional field is needed
[3] in `EventsController + delegate` approach there's no need to keep subscriptions and controller is simply set as a delegate

Choosing between [2] and [3] I'd go with the latter: it follows the pattern used in other controllers (with multicast delegate and combine support) and we don't need extra field for subscriptions.

nuno-vieira:
> > It would be nice to showcase also the usages of each approach 👍
> 
> Done, please have a look @nuno-vieira 🙂
> 
> [1] exposing models specific events on dedicated controllers leads to much boilerplate + not obvious how to get access to all events (seems we still need `EventsController` for this)
> [2] in `EventsController + closures` approach we have to store subscriptions -> additional field is needed
> [3] in `EventsController + delegate` approach there's no need to keep subscriptions and controller is simply set as a delegate
> 
> Choosing between [2] and [3] I'd go with the latter: it follows the pattern used in other controllers (with multicast delegate and combine support) and we don't need extra field for subscriptions.

With the delegate approach. How do we make multiple components subscribe to the events controller in UIKit world? Do we need to create a copy of the events controller for each component and assign the delegate to each one?

b-onc (approved):
This doesn't allow observing custom events but it's a good step!

evsaev:
> With the delegate approach. How do we make multiple components subscribe to the events controller in UIKit world? Do we need to create a copy of the events controller for each component and assign the delegate to each one?

There can be as many event controllers as needed. If a component relies on event listening, a separate `EventsController` would be instantiate for the internal usage (and a component would be set as the delegate). That's the same pattern we use in model controllers so we're consistent here.

nuno-vieira (approved):
LGTM!

energo:
Hello, could provide clear example, for situation like this:
1. We have a list of channels, that represented by inheritance ChatChannelListVC
2. How the best way to subscribe on each of them to show push notification, when we received event - "New Message" ?

b-onc:
Hello @energo ,

You should create an `EventsController` and assign a delegate to it. You delegate will get called on any new event. You can check if the event is `NewMessage`. Like so:
```swift
class CustomChannelListVC: ChatChannelListVC, EventsControllerDelegate {
    
    var eventsController: EventsController!
    
    override func viewDidLoad() {
        super.viewDidLoad()
        
        eventsController = controller.client.eventsController()
        eventsController.delegate = self
    }
    
        switch event {
        case let event as MessageNewEvent:
            // do what you need with the event
            print(event.messageId)
            // you can check if the app is in BG state and show local notification here
        default:
            break
        }
    }
}
```

energo:
@b-onc thank you! 
Seems it's looks what we need.

energo:
One more question: 
I need data from "payload" which come in "MessageNewEvent",  but It seems they accessible only as internal data for sdk usage. 
How could I get user name, channel name, message text ? 
What is the convenient way to do it ?

energo:
@b-onc 
Guys, we need your assist. Could you please help us with this, cause its blocking our release.
Thanks

b-onc:
Hello @energo ,

We're looking into this. For this to not get lost in this already merged PR, can you open a new ticket in our repo?

PR #1269, opened 2021-07-12
[CIS-829] Make the visibility of deleted messages configurable

VojtaStavik:
This PR adds an option to configure the visibility of deleted messages. This applies both to the channel messages, and to threads. 

We currently expose the following options:

```swift
/// All deleted messages are always hidden.
case alwaysHidden

/// Deleted message by current user are visible, other deleted messages are hidden.
case visibleForCurrentUser

/// Deleted messages are always visible.
case alwaysVisible
```


evsaev (approved):
👏

PR #1270, opened 2021-07-13
[CIS-621] Fix quoting a quoted message bug

SagarSDagdu:
### What does this PR do
- Fixes a bug where quoting a message which already quotes some message shows the already quoted message as non-quoted

**Example**
- There's message "A"
- A user creates an inline reply (quotes) "B" that quotes message "A"
- A user creates an inline reply (quotes) "C" that quotes message "B"

Result (earlier) -> message "B" no longer has the quoted message visible.
Result (now)  -> message "B" correctly displays the quoted message.

### DIFF
--- a/Sources/StreamChat/APIClient/Endpoints/Payloads/MessagePayloads.swift
+++ b/Sources/StreamChat/APIClient/Endpoints/Payloads/MessagePayloads.swift
@@ -57,6 +57,7 @@ class MessagePayload<ExtraData: ExtraDataTypes>: Decodable {
     let parentId: String?
     let showReplyInChannel: Bool
     let quotedMessage: MessagePayload<ExtraData>?
+    let quotedMessageId: MessageId?
     let mentionedUsers: [UserPayload<ExtraData.User>]
     let threadParticipants: [UserPayload<ExtraData.User>]
     let replyCount: Int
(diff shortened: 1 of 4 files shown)


Sources/StreamChat/APIClient/Endpoints/Payloads/MessagePayloads.swift
```diff
@@ -57,6 +57,7 @@ class MessagePayload<ExtraData: ExtraDataTypes>: Decodable {
     let parentId: String?
     let showReplyInChannel: Bool
     let quotedMessage: MessagePayload<ExtraData>?
+    let quotedMessageId: MessageId?
```
VojtaStavik, line 60: This needs to be tested

Sources/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -383,7 +383,14 @@ extension NSManagedObjectContext: MessageDatabaseSession {
             dto.pinnedBy = try saveUser(payload: pinnedByUser)
         }
 
-        dto.quotedMessage = try payload.quotedMessage.flatMap { try saveMessage(payload: $0, for: cid) }
+        if let quotedMessage = payload.quotedMessage {
+            dto.quotedMessage = try saveMessage(payload: quotedMessage, for: cid)
+        } else if let quotedMessageId = payload.quotedMessageId {
+            // In case we do not have a fully formed quoted message in the payload, we check for quotedMessageId. This can happen in the case of nested quoted messages.
```
VojtaStavik, line 389: The formatter doesn't automatically wrap comments. I recommend setting the Xcode guide to 132 to make sure also the comments are wrapped at the correct column.

Screen Shot 2021-07-13 at 08 58 00

VojtaStavik (approved):
LGTM 👍

Sources/StreamChat/Database/DTOs/MessageDTO.swift
SagarSDagdu: Thanks for this tip!

Sources/StreamChat/APIClient/Endpoints/Payloads/MessagePayloads.swift
SagarSDagdu: Added tests

PR #1277, opened 2021-07-14
[CIS-1023] Fix composer buttons not shown in first render

nuno-vieira:
### Problem
If the `ChatMessageListVC` was presented without the channel list, the `commandsButton` and `attachmentsButton` would not be shown in the first render of the composer (first `updateContent()`).  Because when presenting the channel list this way the channels are not cached and so the `channelConfig` will be `nil` on the first `updateContent()`, which means that the `commandsButton` and `attachmentsButton` will be hidden. 

### Solution
The solution is to call `composerVC.updateContent() ` when the `channelController.synchronize` finishes from the `MessageListVC`. With this, the composer is updated and will get the channel config properly and show the buttons.

### Demo
I've changed the demo app to present a message list without the channel list to replicate the problem. In this video we can see that after the loading of the channel the buttons are shown. Before they were not.

https://user-images.githubusercontent.com/12814114/125656137-2db3d53a-9578-43b3-94ec-cd089aafd6c7.mp4

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### 🐞 Fixed
 - Fix deleted messages appearance [#1267](https://github.com/GetStream/stream-chat-swift/pull/1267)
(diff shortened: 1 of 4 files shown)


evsaev (approved):
LGTM 👍

PR #1282, opened 2021-07-15
Handle not supported reactions correctly

tbarbugli:
- `ChatMessageLayoutOptionsResolver` will not add the `.reactions` layout option to a message if all reactions are missing from `appareance.images.availableReactions`
- `ChatMessageReactionsView` do not render an empty reaction view if the reaction type is missing

In both cases we log a warning message


Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageLayoutOptionsResolver.swift
```diff
@@ -91,6 +102,29 @@ open class _ChatMessageLayoutOptionsResolver<ExtraData: ExtraDataTypes> {
         return options
     }
 
+    func isQuoted(_ message: _ChatMessage<ExtraData>) -> Bool {
```
evsaev, line 105: nit: we're checking if the message quotes another message, not if it's quoted by another message. We can either change the func name to `isQuotingAnotherMessage` or just keep the `if` untouched.

*the options name is `.quotedMessage` because it says whether the the cell should include a quoted message

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageLayoutOptionsResolver.swift
```diff
@@ -73,15 +84,15 @@ open class _ChatMessageLayoutOptionsResolver<ExtraData: ExtraDataTypes> {
...
+        if isQuoted(message) {
             options.insert(.quotedMessage)
         }
         if message.isRootOfThread || message.isPartOfThread {
             options.insert(.threadInfo)
             // The bubbles with thread look like continuous bubbles
             options.insert(.continuousBubble)
         }
-        if !message.reactionScores.isEmpty && channel.config.reactionsEnabled {
+        if hasReactions(channel, message, appearance) {
```
evsaev, line 95: nit: would be great to have a unit-test that checks `.reactions` option is not included if for all reactions icons are missing (`ChatMessageLayoutOptionsResolver_Tests`)

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageLayoutOptionsResolver.swift
```diff
@@ -18,16 +18,27 @@ open class _ChatMessageLayoutOptionsResolver<ExtraData: ExtraDataTypes> {
...
     }
 
+    @available(*, deprecated, message: "Use the same method with the appearance parameter instead")
+    open func optionsForMessage(
+        at indexPath: IndexPath,
+        in channel: _ChatChannel<ExtraData>,
+        with messages: AnyRandomAccessCollection<_ChatMessage<ExtraData>>
+    ) -> ChatMessageLayoutOptions {
+        optionsForMessage(at: indexPath, in: channel, with: messages, appearance: Appearance.default)
+    }
```
evsaev, line 28: we've got quite a few warnings in `ChatMessageLayoutOptionsResolver_Tests` file because of this change 😄 would be really cool if we can update the tests to access the updated new version of a func

evsaev (commented):
don't know if we need a CHANGELOG entry for this change

evsaev (approved):
Looks good, thanks! 👍

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageLayoutOptionsResolver.swift
```diff
@@ -73,15 +84,15 @@ open class _ChatMessageLayoutOptionsResolver<ExtraData: ExtraDataTypes> {
         if isLastInSequence && !message.isSentByCurrentUser && !channel.isDirectMessageChannel {
             options.insert(.authorName)
         }
-        if message.quotedMessage?.id != nil {
+        if isQuoted(message) {
             options.insert(.quotedMessage)
         }
```
nuno-vieira, line 89: like @evsaev mentioned, this should actually be: 
```suggestion
        if hasQuotedMessage(message) {
            options.insert(.quotedMessage)
        }
```
Because this checks if the current message is quoting a message. Another option is: `isQuoting(message)`.

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageLayoutOptionsResolver.swift
evsaev: in `ChatMessageLayoutOptionsResolver_Tests.swift` (where the unit-tests for options resolver live) we keep calling the func which is now deprecated -> the compiler warns us about it just like it does for customers 🙂 The fix would be to invoke the new func with `Appearance` param added

Issue #1283, opened 2021-07-15
ChatMessageListView.updateMessages(with:completion:) Crash


osawhoop:
## What did you do?
I just created an instance of a channel and I got the crash, second time I opened the channel it did not crash

## What did you expect to happen?
Not crash and show the messages of the channel

## What happened instead?
It crashes

## GetStream Environment
**GetStream Chat version:**4.0.0-beta.5
**GetStream Chat frameworks:** StreamChatUI
**iOS version:**14.6
**Swift version:**5
**Xcode version:**12.5.1
**Device:**iPhone 12 mini

## Additional context
Attaching crash log

dmigach:
Hello @osawhoop!
Is it possible to [symbolicate](https://developer.apple.com/documentation/xcode/adding-identifiable-symbol-names-to-a-crash-report#Symbolicate-the-Crash-Report-in-Xcode) the crash report?

osawhoop:
I hope this one works, thanks!
[StreamChatTest.crash.zip](https://github.com/GetStream/stream-chat-swift/files/6830787/StreamChatTest.crash.zip)

dmigach:
Hello @osawhoop! The [PR](https://github.com/GetStream/stream-chat-swift/pull/1286/files) addressing the issue was merged, we're preparing a release that's going to include this change

PR #1286, opened 2021-07-16
[CIS-1029] Fix message list crashing on updates

dmigach:
https://github.com/GetStream/stream-chat-swift/issues/1283
@evsaev suggested that we experience the same thing we have for channel list, hence here goes the same fix

### DIFF
--- a/Sources/StreamChatUI/ChatMessageList/ChatMessageListView.swift
+++ b/Sources/StreamChatUI/ChatMessageList/ChatMessageListView.swift
@@ -175,6 +175,54 @@ open class _ChatMessageListView<ExtraData: ExtraDataTypes>: UITableView, Customi
         completion: (() -> Void)? = nil
     ) {
         var shouldScrollToBottom = false
+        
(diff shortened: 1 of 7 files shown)


Sources/StreamChatUI/ChatMessageList/ChatMessageListView.swift
```diff
@@ -175,6 +175,54 @@ open class _ChatMessageListView<ExtraData: ExtraDataTypes>: UITableView, Customi
...
+            reloadData()
+            
+            logConflicts(
+                moves: moveIndexes,
+                inserts: insertIndexes,
+                updates: updateIndexes,
+                removes: removeIndexes
+            )
+            return
+        }
```
nuno-vieira, line 225: Maybe it is better to extract this to an internal component and use it in both places? 🤔

evsaev (approved):
Looks good 👍

nuno-vieira (approved):
LGTM

Issue #1279, opened 2021-07-15
"MemberRole string value `user` doesn\'t match any of the known roles", underlyingError: nil))


BarragaoA:
Hello

## What did you do?
I'm fetching channel list :

> let query = ChannelListQuery(filter: Filter.containMembers(userIds: [userId]), sort: [Sorting(key: .lastMessageAt)], pageSize: .max, messagesLimit: .max)
>         let channelList = ChatClient.shared.channelListController(query: query)
>         channelList.synchronize { error in
>             if error != nil {
>                 completion(.failure(error!))
>             } else{
>                 let unreadChannels = channelList.channels.filter { channel in
>                     channel.isUnread
>                 }
>                 completion(.success(unreadChannels))
>             }
>         }

## What did you expect to happen?
Get channel list

## What happened instead?
Everything was working well on tuesday but since this morning I got the following error :
`[ERROR] [com.apple.NSURLSession-delegate] [RequestDecoder.swift:71] [decodeRequestResponse(data:response:error:)] > dataCorrupted(Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "channels", intValue: nil), _JSONKey(stringValue: "Index 0", intValue: 0), CodingKeys(stringValue: "membership", intValue: nil), CodingKeys(stringValue: "role", intValue: nil)], debugDescription: "MemberRole string value `user` doesn\'t match any of the known roles", underlyingError: nil))`
Did something change ?

## GetStream Environment
**GetStream Chat version:** 3.1.11
**GetStream Chat frameworks:** StreamChat, StreamChatUI
**iOS version:** 14.4.2
**Swift version:** 5
**Xcode version:** 12.5.1
**Device:** iPhone Xr

## Additional context
Thanks for your help

BarragaoA:
Actually I understand that I'm receiving a bad memberRole value. Why do I receive 'user' as MemberRole value ?

whiplash00:
I suddenly got this same issue since early this week. This bug seems to stop new channels from showing up in the directory (using development token). I have to make everyone admin role in order for things to continue to work. Please fix it asap.

tbarbugli:
@BarragaoA @whiplash00 there was an API regression that returned "user" instead of "member" for the membership role. These two are aliases but unfortunately the iOS SDK is rather strict about this. A patch for this was rolled out yesterday and does not require updating anything on the SDK. 

I am going to close this one, please re-open if you can still reproduce this issue.

PR #1288, opened 2021-07-19
[CIS-1028] Allow any value for `MemberRole`

SagarSDagdu:
### What does this PR do
- This PR allows the backend to send any value for the MemberRole(`role`) in the payload, thus making the SDK a bit more resilient to potential problems like [this](https://github.com/GetStream/stream-chat-swift/issues/1279)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 # Upcoming
 
+### 🐞 Fixed
+- Fix an issue where member role sent from backend was not recognized by the SDK [#1288](https://github.com/GetStream/stream-chat-swift/pull/1288)
+
 ### 🔄 Changed
 
 # [4.0.0-beta.7](https://github.com/GetStream/stream-chat-swift/releases/tag/4.0.0-beta.7)
--- a/Sources/StreamChat/Models/Member.swift
+++ b/Sources/StreamChat/Models/Member.swift
@@ -109,21 +109,34 @@ public class _ChatChannelMember<ExtraData: UserExtraData>: _ChatUser<ExtraData> 
     }
 }
 
-/// An enum describing possible roles of a member in a channel.
-public enum MemberRole: String, Codable, Hashable {
+/// A  `struct` describing roles of a member in a channel.
+/// There are some predefined types but any type can be introduced and sent by the backend.
+public struct MemberRole: RawRepresentable, Codable, Hashable, ExpressibleByStringLiteral {
+    public let rawValue: String
+    
+    public init(rawValue: String) {
+        self.rawValue = rawValue
+    }
+    
+    public init(stringLiteral value: String) {
+        self.init(rawValue: value)
+    }
+}
+
+public extension MemberRole {
     /// This is the default role assigned to any member.
-    case member
-    
+    static let member = Self(rawValue: "member")
+
     /// Allows the member to perform moderation, e.g. ban users, add/remove users, etc.
-    case moderator
-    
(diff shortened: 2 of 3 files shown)


Sources/StreamChat/Models/Member.swift
```diff
@@ -136,10 +149,7 @@ public enum MemberRole: String, Codable, Hashable {
         case "owner":
             self = .owner
```
DominikBucher12, line 150: This is probably nit. IMHO either keep all the default values in the initial switch or let there be only the asignment in the `default` branch...

Sources/StreamChat/Models/Member_Tests.swift
```diff
@@ -34,5 +34,9 @@ class MemberRole_Tests: XCTestCase {
 
         json = try embed("owner")
```
DominikBucher12, line 35: testing nit 😅  We do test owner but not reset of the default offered roles...

DominikBucher12 (changes requested):
You can ignore the nits comments :) But CHANGELOG is missing in here :)

Sources/StreamChat/Models/Member.swift
```diff
@@ -136,10 +149,7 @@ public enum MemberRole: String, Codable, Hashable {
...
         default:
-            throw DecodingError.dataCorruptedError(
-                in: container,
-                debugDescription: "MemberRole string value `\(value)` doesn't match any of the known roles"
-            )
+            self = MemberRole(rawValue: value)
```
nuno-vieira, line 152: Maybe this is actually only what we need? 🤔 We don't need a switch case. 

If we need to compare roles, we would do `user.memberRole == .moderator`, this will always compare the raw value. So there is no need for the switch case. The static props is just used to be easier to compare with known roles, or to create easily roles .

nuno-vieira (changes requested):

Sources/StreamChat/Models/Member.swift
SagarSDagdu: I think we need the switch case as we map two values i.e. "channel_moderator" and "moderator" to the role `moderator`?

nuno-vieira: Why do we have those two types tho ? 🤔 are they really the same role?

nuno-vieira (approved):
LGTM! ✅

Sources/StreamChat/Models/Member.swift
SagarSDagdu: Yes, we do need to map those two to the same role

Sources/StreamChat/Models/Member.swift
SagarSDagdu: All the default values are in the initial switch, and only the assignment is in the `default` branch

CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 # Upcoming
 
+### 🐞 Fixed
+- Fix an issue where member role sent from backend was not recognized by the SDK [#1279](https://github.com/GetStream/stream-chat-swift/issues/1279)
```
evsaev, line 9: if PR fixes an issue should we reference the PR or the issue from the CHANGELOG? 🤔
cc @b-onc

b-onc: doesn't matter that much, but I'd prefer
PR refers to issue (fixes #xxxx)
changelog refers to PR

SagarSDagdu: Makes sense, I'll make this change 👍

evsaev (approved):
LGTM!

DominikBucher12 (approved):
🎉

Issue #1274, opened 2021-07-14
Crash on `ChannelListUpdater.swift:20`


Lascorbe:
## What did you do?
Not sure if just opening the app or opening the chat view, it's a Firebug crash report.

## What did you expect to happen?
Not to crash

## What happened instead?
Crashed

## GetStream Environment
**GetStream Chat version:** 4.0.0-beta.6 (SwiftPM pointing to commit `959df0b0c4ab0f2446fabcdc7ba083e44ae40e5e`)
**GetStream Chat frameworks:** StreamChat, StreamChatUI, StreamChatClient, StreamChatCore
**iOS version:** 14.4.2
**Swift version:** Swift 5.4
**Xcode version:** Xcode 12.5.1
**Device:** iPhone 11

## Additional context
![image](https://user-images.githubusercontent.com/1515520/125623600-9e5e18c6-e387-44b5-b0ff-37e8f344e97d.png)
![image](https://user-images.githubusercontent.com/1515520/125623730-20c146de-a921-4718-8fbe-fd5f83cb31a7.png)

```
Crashed: com.apple.NSURLSession-delegate
0  libsystem_kernel.dylib         0x1b5285414 __pthread_kill + 8
1  libsystem_pthread.dylib        0x1d2de0b50 pthread_kill + 272
2  libsystem_c.dylib              0x190763b74 abort + 104
3  libswiftCore.dylib             0x18b29df20 swift_vasprintf(char**, char const*, char*) + 58
4  libswiftCore.dylib             0x18b29e078 swift::swift_abortRetainUnowned(void const*) + 36
5  libswiftCore.dylib             0x18b2ed844 swift_unknownObjectUnownedTakeStrong + 74
6  Chat                           0x1054fd748 closure #1 in ChannelListUpdater.update(channelListQuery:completion:) + 20 (ChannelListUpdater.swift:20)
7  Chat                           0x1054fe0b8 partial apply for thunk for @escaping @callee_guaranteed (@guaranteed Result<ChannelListPayload<A>, Error>) -> () + 4317651128 (<compiler-generated>:4317651128)
8  Chat                           0x10537e6e8 closure #1 in closure #1 in APIClient.request<A>(endpoint:completion:) + 68 (APIClient.swift:68)
9  Chat                           0x10537ee80 partial apply for closure #1 in closure #1 in APIClient.request<A>(endpoint:completion:) + 4316081792 (<compiler-generated>:4316081792)
[...]

PR #1289, opened 2021-07-19
[CIS-1030] Fix crash in `ChannelListUpdater`

evsaev:
**This PR** fixes crash in `ChannelListUpdater` (#1274)

The `ChannelListUpdater` is used by `DatabaseCleanupUpdater` background worker to update channel list queries when reconnection happens (the ownership chain is `ChatClient -> MissingEventsPublisher -> DatabaseCleanupUpdater -> ChannelListUpdater`). 

We expect all background workers to deallocate together with `ChatClient` but this was not the case for `ChannelListUpdater`. Channel updater kept itself alive during `/channels` API call and if `ChatClient` was deallocated before the response arrival, access to `unowned let database` was causing a crash since the was no strong refs to the db container.

The change lets the `ChannelListUpdater` to deallocate together with `ChatClient` no matter in-progress API calls.


nuno-vieira (approved):
LGTM!

tbarbugli (approved):

Sources/StreamChat/Workers/ChannelListUpdater.swift
```diff
@@ -14,10 +14,13 @@ class ChannelListUpdater<ExtraData: ExtraDataTypes>: Worker {
     ///
     func update(channelListQuery: _ChannelListQuery<ExtraData.Channel>, completion: ((Error?) -> Void)? = nil) {
         apiClient
-            .request(endpoint: .channels(query: channelListQuery)) { (result: Result<ChannelListPayload<ExtraData>, Error>) in
+            .request(endpoint: .channels(query: channelListQuery)) { [weak self] (result: Result<
+                ChannelListPayload<ExtraData>,
+                Error
+            >) in
```
b-onc, line 20: this is very weird formatting... is it the formatter's work?

b-onc (approved):

dmigach (approved):

Sources/StreamChat/Workers/ChannelListUpdater.swift
evsaev: yep, it's a formatter 🙂

PR #1291, opened 2021-07-20
[CIS-1035] Introduce `urlRequest(forImage)` to ImageCDN

b-onc:
This will allow users to inject their HTTP headers to image loading requests

Would look like:
```swift
class CustomImageCDN: StreamImageCDN {
    override func urlRequest(forImage url: URL) -> URLRequest {
        var request = URLRequest(url: url)
        request.setValue("Custom!", forHTTPHeaderField: "X-Custom-Header")
        return request
    }
}
```
and inject the class somewhere...
```swift
Components.default.imageCDN = CustomImageCDN()
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 # Upcoming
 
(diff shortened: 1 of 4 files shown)


nuno-vieira (approved):
LGTM

PR #1293, opened 2021-07-20
[CIS-1038] Fix composer allowing whitespace only messages to be sent

b-onc:
Composer send button should be disabled when the message is only whitespace. Sending whitespace only message breaks our UI.

Also, backend trims the whitespace when the message is published. If we send the message with leading/trailing whitespace, the UI first shows the message, but when backend sends the actual message, UI jumps. So it's better to send trimmed message to backend, to prevent UI jumps.


nuno-vieira (approved):
LGTM!

PR #1294, opened 2021-07-20
[CIS-990] Extract Message List Header Component

nuno-vieira:
Extracts the Message List Header to a component to be easier to customize and make the `MessageListVC` smaller and with fewer responsibilities. 

- Minor improvements to the SuggestionsVC
- Fixes Message List Header displaying Online Status for the current user instead of the other one
- Extracts the header of the Message List to `ChatChannelHeaderView`:
 This component uses `TitleContainerView` to render the title and subtitle views of the header, and all the data comes from the `channelController`. So, from now on the header of the message list is responsible to subscribe to channel events and update itself, instead of `ChatMessageListVC`.
- Rename`Components.navigationTitleView` -> `Components.titleContainerView`
- Add `Components.channelHeaderView`

### Breaking Changes
The previous `Components.navigationTitleView` is now `Components.titleContainerView`. This component is used by both `ChatChannelHeaderView` and `ChatThreadHeaderView` which are new components and represent the Header of the `ChatMessageListVC` and `ChatThreadVC`, respectively.

Components Hierachy: 

- **ChatChannelHeaderView**:
     - TitleContainerView
- **ChatThreadHeaderView**:
     - TitleContainerView


Sources/StreamChatUI/ChatMessageList/ChatChannelHeaderView.swift
```diff
@@ -0,0 +1,137 @@
...
+            }
+        }
+
+        return L10n.Message.Title.group(channel.memberCount, channel.watcherCount)
+    }
+
+    /// Create a timer if the channel is direct between two people. Will call update content
+    /// repeatedly every minute to update the online status of the members.
+    open func makeTimer() {
+        if timer == nil, channelController?.channel?.isDirectMessageChannel == true {
```
tbarbugli, line 87: I suggest we simplify this and always have the ticker running

Sources/StreamChatUI/ChatMessageList/ChatChannelHeaderView.swift
```diff
@@ -0,0 +1,137 @@
...
+            timer = Timer.scheduledTimer(withTimeInterval: 60, repeats: true) { [weak self] _ in
```
tbarbugli, line 88: perhaps make the withTimeInterval configurable (via subclassing) and do not create the timer when set to 0

CHANGELOG.md
```diff
@@ -5,6 +5,15 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 # Upcoming
 
+### ⚠️ Breaking Changes from `4.0-beta.8`
+Renamed `Components.navigationTitleView` -> `Components.titleContainerView`
```
tbarbugli, line 9: when everything else on this PR is good to go I suggest adding instructions on how to upgrade from previous versions (if still applicable)

Sources/StreamChatUI/Components.swift
```diff
@@ -10,8 +10,8 @@ public typealias Components = _Components<NoExtraData>
 
 /// An object containing types of UI Components that are used through the UI SDK.
 public struct _Components<ExtraData: ExtraDataTypes> {
-    /// The view used as a navigation bar title view for some view controllers.
-    public var navigationTitleView: TitleContainerView.Type = TitleContainerView.self
+    /// A view that displays a title label and subtitle in a container stack view.
+    public var titleContainerView: TitleContainerView.Type = TitleContainerView.self
```
tbarbugli, line 14: we should have two vars so that you can pick different titles for thread and channel screens 

what about:

```swift
class TitleContainerView: _View, ... {}

class ChatChannelHeaderView:  TitleContainerView {}

class ChatThreadHeaderView:  TitleContainerView {}


    public var chatChannelHeaderView: ChatChannelHeaderView.Type = ChatChannelHeaderView.self

    public var chatThreadHeaderView: ChatThreadHeaderView.Type = ChatThreadHeaderView.self


```

Sources/StreamChatUI/ChatMessageList/ChatChannelHeaderView.swift
nuno-vieira: Good idea 👍

CHANGELOG.md
nuno-vieira: On the PR Description?

Sources/StreamChatUI/ChatMessageList/ChatChannelHeaderView.swift
nuno-vieira: ✅

Sources/StreamChatUI/ChatMessageList/ChatChannelHeaderView.swift
```diff
@@ -0,0 +1,144 @@
...
+
+        embed(titleContainerView)
+    }
+
+    override open func updateContent() {
+        super.updateContent()
+
+        titleContainerView.content = (titleText, subtitleText)
+
+        makeTimer()
```
evsaev, line 53: hm, why we create a timer in `updateContent` and not in `setUp`?

Sources/StreamChatUI/ChatMessageList/ChatChannelHeaderView.swift
```diff
@@ -0,0 +1,144 @@
...
+    /// Controller for observing data changes within the channel.
+    open var channelController: _ChatChannelController<ExtraData>?
+
+    /// The user id of the current logged in user.
+    open var currentUserId: UserId? {
+        channelController?.client.currentUserId
+    }
+
+    /// Timer used to update the online status of member in the channel.
+    open var timer: Timer?
```
evsaev, line 25:
```suggestion
    open var timer: Timer? {
     didSet { oldValue?.invalidate() }
    }
```

Sources/StreamChatUI/ChatMessageList/ChatChannelHeaderView.swift
```diff
@@ -0,0 +1,144 @@
...
+    _View,
+    ThemeProvider,
+    _ChatChannelControllerDelegate {
+    /// Controller for observing data changes within the channel.
+    open var channelController: _ChatChannelController<ExtraData>?
+
+    /// The user id of the current logged in user.
+    open var currentUserId: UserId? {
+        channelController?.client.currentUserId
+    }
```
evsaev, line 22: nit: I wound't publicly expose current user id on a UI component, tbh. Clients can always get it from a chat client/controller

Sources/StreamChatUI/ChatMessageList/ChatChannelHeaderView.swift
```diff
@@ -0,0 +1,144 @@
...
+        // but this can be overridden by subclassing this component.
+    }
+
+    open func channelController(
+        _ channelController: _ChatChannelController<ExtraData>,
+        didReceiveMemberEvent: MemberEvent
+    ) {
+        // By default the header view is not interested in member events
+        // but this can be overridden by subclassing this component.
+    }
```
evsaev, line 131: we don't have `updateContentIfNeeded` here because for these changes we get `.update` change, right? 🤔

Sources/StreamChatUI/ChatMessageList/ChatChannelHeaderView.swift
```diff
@@ -0,0 +1,144 @@
...
+
+    // MARK: - ChatChannelControllerDelegate Implementation
+
+    open func channelController(
+        _ channelController: _ChatChannelController<ExtraData>,
+        didUpdateChannel channel: EntityChange<_ChatChannel<ExtraData>>
+    ) {
+        switch channel {
+        case .update:
+            updateContentIfNeeded()
```
evsaev, line 111:
```suggestion
        switch channel {
        case .update, .create:
            updateContentIfNeeded()
```

Sources/StreamChatUI/ChatMessageList/ChatChannelHeaderView.swift
nuno-vieira: Yes, you are right, nice catch! This was because before the timer was dependent on `is.directMessageChannel` and that could change after the `updateContent`, right now it doesn't make sense to be in `updateContent` and indeed it can go to setUp, will change 👍

Sources/StreamChatUI/ChatMessageList/ChatChannelHeaderView.swift
nuno-vieira: This timer is only created one time, so I don't think this is needed 🤔 Unless the customer change this 🤔

Sources/StreamChatUI/ChatMessageList/ChatChannelHeaderView.swift
```diff
@@ -0,0 +1,144 @@
...
+
+    /// A view that displays a title label and subtitle in a container stack view.
+    open private(set) lazy var titleContainerView: TitleContainerView = components
+        .titleContainerView.init()
+        .withoutAutoresizingMaskConstraints
+
+    override open func setUp() {
+        super.setUp()
+
+        channelController?.setDelegate(self)
```
evsaev, line 39: nit: I think it's safer to set the delegate in `channelController.didSet` to fix the case when the header keeps being a delegate of an initial channel controller

Sources/StreamChatUI/ChatMessageList/ChatChannelHeaderView.swift
nuno-vieira: This is stored property, it will always get it from the `client`, maybe we can change this to `public` instead of `open`. But this makes the code cleaner when using the current user id, also makes it easier to test the component by overriding `currentUserId` on tests

Sources/StreamChatUI/ChatMessageList/ChatChannelHeaderView.swift
nuno-vieira: Yes 👍

Sources/StreamChatUI/ChatMessageList/ChatChannelHeaderView.swift
nuno-vieira: Why do you think "create" is needed ? 🤔 What is the use case?

Sources/StreamChatUI/ChatMessageList/ChatChannelHeaderView.swift
nuno-vieira: Done ✅

Sources/StreamChatUI/ChatMessageList/ChatChannelHeaderView.swift
nuno-vieira: ✅

Sources/StreamChatUI/ChatMessageList/ChatChannelHeaderView.swift
nuno-vieira: ✅

Sources/StreamChatUI/ChatMessageList/ChatChannelHeaderView.swift
evsaev: if we expose `currentUserId` on one component customer may start asking why it's missing in other UI components which we're not willing/planning to do 😄

evsaev: if it's a useful helper/is needed in tests we could make it `internal` but exposing identifiers on UI components feels like the wrong thing to me

Sources/StreamChatUI/ChatMessageList/ChatChannelHeaderView.swift
evsaev: the use case is showing a message list when channel hasn't yet been created (an optimistic update). In that case, we would get `create` change first (most likely followed by several `update` but I wouldn't rely on that)

nuno-vieira: ✅

Sources/StreamChatUI/ChatMessageList/ChatChannelHeaderView.swift
nuno-vieira: I don't think it is big deal, to be honest, even the customer can add this property themselves. Also will help them for testing. If the customers have custom logic in the components, they will need to test the logic, and this will be very useful for them. This property is just a helper to get the current user from the `client`. I don't see any potential danger with this approach

Sources/StreamChatUI/ChatMessageList/ChatChannelHeaderView.swift
evsaev: but a single timer being active in the moment is what we want, right? 😄 With `makeTime` moved to `setUp` it became less fragile to errors on the client side but a different time can still be assigned. At the same time, `private(set)` also works fine

evsaev (approved):
Looks good 👍

Issue #1295, opened 2021-07-20
Delete message action causes crash 


srgray:
## What did you do?
Set the new `ChatClientConfig.deletedMessagesVisibility` = `.alwaysHidden`
Delete a message in a chat channel

## What did you expect to happen?
Message to be deleted and not shown in UI

## What happened instead?
App crash:
Swift/ContiguousArrayBuffer.swift:580: Fatal error: Index out of range
`LazyCachedMapCollection.swift` line 39

## GetStream Environment
**GetStream Chat version:** 4.0.0-beta.7
**GetStream Chat frameworks:** StreamChat, StreamChatUI

b-onc:
Hello @srgray ,

I can't reproduce the issue. Is the issue consistently reproducible for you? Can you share the app id (not API key, App id) and channel id this happens?

tbarbugli:
@srgray if possible a stacktrace would be super useful as well to quickly resolve this

srgray:
Yes, I am able to reproduce this very consistently.
AppID: `82241`
ChannelID: `livestream:71c386ae-2788-46d4-8545-172de370dfcd` (but I have reproduced this on several channels)
Stack Trace:
```
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = Fatal error: Index out of range
    frame #0: 0x00007fff2f36ab50 libswiftCore.dylib`_swift_runtime_on_report
    frame #1: 0x00007fff2f3e4e8c libswiftCore.dylib`_swift_stdlib_reportFatalErrorInFile + 204
    frame #2: 0x00007fff2f07d8ee libswiftCore.dylib`closure #1 (Swift.UnsafeBufferPointer<Swift.UInt8>) -> () in closure #1 (Swift.UnsafeBufferPointer<Swift.UInt8>) -> () in closure #1 (Swift.UnsafeBufferPointer<Swift.UInt8>) -> () in Swift._assertionFailure(_: Swift.StaticString, _: Swift.StaticString, file: Swift.StaticString, line: Swift.UInt, flags: Swift.UInt32) -> Swift.Never + 478
[...]

PR #1298, opened 2021-07-21
[CIS-1043] Fix delete message action causing a crash

SagarSDagdu:
### What does this PR do
- This PR fixes [this](https://github.com/GetStream/stream-chat-swift/issues/1295) issue where a crash would occur when deleting a message from a channel when the deleted messages visibility is set to `alwaysHidden`

### DIFF
--- a/Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
+++ b/Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
@@ -251,7 +251,14 @@ open class _ChatMessageListVC<ExtraData: ExtraDataTypes>:
         _ overlay: ChatMessageListScrollOverlayView,
         textForItemAt indexPath: IndexPath
     ) -> String? {
-        DateFormatter
+        // When a message from a channel is deleted,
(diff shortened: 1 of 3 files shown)


nuno-vieira (approved):
LGTM

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -251,7 +251,14 @@ open class _ChatMessageListVC<ExtraData: ExtraDataTypes>:
         _ overlay: ChatMessageListScrollOverlayView,
         textForItemAt indexPath: IndexPath
     ) -> String? {
-        DateFormatter
+        // When a message from a channel is deleted,
+        // and the visibility of deleted messages is set to `alwaysHidden`,
+        // the messages list won't contain the message and hence it would crash
+        guard channelController.messages.indices.contains(indexPath.item) else {
```
tbarbugli, line 257: @SagarSDagdu why not

```swift
if channelController.messages.count <= indexPath.item {
    return nil
}
```

PR #1299, opened 2021-07-21
[CIS-42] Implement markdown support

dmigach:
High level idea is the following:
* We include source files for https://github.com/JohnSundell/Ink to render markdown into HTML. We don't want to display this HTML in a web-view so instead we...
* Turn HTML into an attributed string using
```swift
NSMutableAttributedString(
            data: data,
            options: [
                .documentType: NSAttributedString.DocumentType.html,
                .characterEncoding: String.Encoding.utf8.rawValue
            ],
            documentAttributes: nil
        )
```
* This attributed string is not what we want because we have our own fonts, colors etc. So we need an additional pass over an attributed string to preserve traits we parsed from markdown but apply our styling. We need to repeat it when fonts change dynamically.

Some examples:
Simulator Screen Shot - iPhone 12 - 2021-07-21 at 14 50 31
Screenshot 2021-07-21 at 15 21 51

Additionally:

During your review please ignore the first commit that simply adds files from Ink, if you do so than the PR becomes more than manageable
Tests for Markdown -> HTML parser will be added later if we decide to proceed with this PR
There are performance concerns caused by the amount of updates we perform on Message List and especially on Channel List. They are to be addressed separately
There is a known issue with monospace trait, for some reason it ignores font size attribute


Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
```diff
@@ -409,6 +409,12 @@ open class _ChatMessageContentView<ExtraData: ExtraDataTypes>: _View, ThemeProvi
 
         NSLayoutConstraint.activate(constraintsToActivate)
     }
+    
+    override open func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
```
dmigach, line 413: Because otherwise the view doesn't properly respond to dynamic font changes.
Also there are a couple of snapshot tests for system messages that were recorded incorrectly for small fonts because this was abscent, re-recorded them in this PR

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
```diff
@@ -438,12 +444,23 @@ open class _ChatMessageContentView<ExtraData: ExtraDataTypes>: _View, ThemeProvi
...
             textColor = appearance.colorPalette.textLowEmphasis
+        } else {
+            textFont = appearance.fonts.body
         }
         
         textView?.textColor = textColor
         textView?.font = textFont
-        textView?.text = content?.textContent
         
+        if let text = content?.textContent {
```
tbarbugli, line 454: code-duplication problem here

Sources/StreamChatUI/Appearance.swift
```diff
@@ -21,6 +21,9 @@ public struct Appearance {
     /// By providing different object or changing individal images, you can change the look of the views.
     public var images = Images()
     
+    /// Setting this property to true enables support for most common markdown features for messages and last message previews
+    var isMarkdownEnabled = false
```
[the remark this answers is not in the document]
tbarbugli: Android handles it like this:

```kotlin
ChatUI.markdown = ChatMarkdown { textView: TextView, text: String ->
    // Parse markdown from text and apply formatting here
    textView.text = applyMarkdown(text)
}
```

tbarbugli (commented):
I think Android encapsulate this nicely by exposing a configurable `ChatMarkdown` (allows you to swap and disable easily)

Performance-wise it would be great to see some kind of check to avoid the parsing for messages that do not contain any markup (most messages don't have markdown)

Good if you add some information in the PR description:

- what do we support out of the box from Markdown syntax
- how you configure this at the SDK level
- what kind of performance cost we are adding 
- anything you are not 100% sure about (if applicable)

PR #1303, opened 2021-07-22
[CIS-558] MessageComposer - Limit the maximum height of composer's textview and add scrolling

SagarSDagdu:
### What does this PR do
- This PR enables us to limit the maximum height of the composer's textView before it enables scrolling behaviour. By this, we enhance the User Experience by not letting the textView grow to full screen. It fixes [this](https://github.com/GetStream/stream-chat-swift/issues/1335) reported issue.

Simulator Screen Recording - iPhone 12 Pro - 2021-07-22 at 13 51 40

### DIFF
--- a/Sources/StreamChatUI/CommonViews/InputTextView/InputTextView.swift
+++ b/Sources/StreamChatUI/CommonViews/InputTextView/InputTextView.swift
@@ -34,6 +34,22 @@ open class InputTextView: UITextView, AppearanceProvider {
         }
     }
     
+    /// The minimum height of the text view.
+    /// When there is no content in the text view OR the height of the content is less than this value,
+    /// the text view will be of this height
+    open var minimumHeight: CGFloat {
+        38.0
+    }
+    
+    /// The constraint responsible for setting the height of the text view.
+    open var heightConstraint: NSLayoutConstraint?
+    
+    /// The maximum height of the text view.
+    /// When the content in the text view is greater than this height, scrolling will be enabled and the text view's height will be restricted to this value
+    open var maximumHeight: CGFloat {
+        120.0
+    }
+    
     override open var attributedText: NSAttributedString! {
         didSet {
             textDidChangeProgrammatically()
@@ -82,7 +98,9 @@ open class InputTextView: UITextView, AppearanceProvider {
         )
         placeholderLabel.pin(anchors: [.centerY], to: self)
         
-        isScrollEnabled = false
+        heightConstraint = heightAnchor.constraint(equalToConstant: minimumHeight)
+        heightConstraint?.isActive = true
+        isScrollEnabled = true
     }
 
     /// Sets the given text in the current caret position.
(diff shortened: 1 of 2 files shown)


Sources/StreamChatUI/CommonViews/InputTextView/InputTextView.swift
```diff
@@ -34,6 +34,21 @@ open class InputTextView: UITextView, AppearanceProvider {
         }
     }
     
+    /// The minimum height of the text view.
+    /// When there is no content in the text view OR the height of the content is less than this value,
+    /// the text view will be of this height
+    open var minimumHeight: CGFloat = 38.0
```
DominikBucher12, line 40: I posted into our channel message that we should probably think more of using stored properties vs read-only properties and overriding them, IMHO this is exactly the case where we want to use get-only property so it's easier to override when someone subclasses `InputTextView`.

DominikBucher12:
```suggestion
   open var minimumHeight: CGFloat {
       return 38.0
   }
```

Sources/StreamChatUI/CommonViews/InputTextView/InputTextView.swift
```diff
@@ -34,6 +34,21 @@ open class InputTextView: UITextView, AppearanceProvider {
...
+    
+    /// The maximum number of lines before the text view starts scrolling. When the content in the text view causes
+    /// the text view to grow upto these number of lines, scrolling is enabled.
+    open var maximumNumberOfLinesBeforeScroll: Int = 5
```
DominikBucher12, line 44: Same comment as `minimumHeight` :)

Sources/StreamChatUI/CommonViews/InputTextView/InputTextView.swift
```diff
@@ -82,7 +97,9 @@ open class InputTextView: UITextView, AppearanceProvider {
         )
         placeholderLabel.pin(anchors: [.centerY], to: self)
         
-        isScrollEnabled = false
+        heightConstraint = heightAnchor.constraint(equalToConstant: minimumHeight)
+        heightConstraint?.isActive = true
+        isScrollEnabled = true
```
DominikBucher12, line 102: Why setting `scrollEnabled` to true? Can you maybe add comment? From this context of PR I don't get it 😅 Also why this is not set on `setTextViewHeight` according to height? Or TextView handles this on it's own?

Sources/StreamChatUI/CommonViews/InputTextView/InputTextView.swift
SagarSDagdu: 💯  true. I'll change this

Sources/StreamChatUI/CommonViews/InputTextView/InputTextView.swift
SagarSDagdu: The textView handles this internally, depending on its height and the content's height. Hence, we do not need to modify it depending on the conditions

Sources/StreamChatUI/CommonViews/InputTextView/InputTextView.swift
```diff
@@ -105,6 +123,29 @@ open class InputTextView: UITextView, AppearanceProvider {
...
+
+        heightConstraint?.constant = heightToSet
+        layoutIfNeeded()
+
+        // This is due to bug in UITextView where the scroll sometimes disables
+        // when a very long text is pasted in it.
+        // Doing this ensures that it doesn't happen
+        // Reference: https://stackoverflow.com/a/33194525/3825788
+        isScrollEnabled = false
+        isScrollEnabled = true
```
DominikBucher12, line 148: Wow 😅 This appears to be working pretty good❗️ Which is awesome.

DominikBucher12 (approved):
Love it! Very nice :)

dmigach (approved):

PR #1305, opened 2021-07-22
[CIS-1040] Add a backoff mechanism for expired token reobtaining

dmigach:
When we try to reobtain token and the new one is expired too, we need to back off so there is no endless loop. 
I'm not sure whether we should utilise approach similar to `WebSocketClientReconnectionStrategy` or one retry is enough.
Because Marcelo said we just need to do this:
```
after receiving a new token SDKs should check if  claims.exp < now() and avoid  any API call if expired
```


Sources/StreamChat/ChatClient.swift
```diff
@@ -593,7 +593,16 @@ extension _ChatClient: ConnectionStateDelegate {
                     clientUpdater.reloadUserIfNeeded(
                         userConnectionProvider: .closure { _, completion in
                             tokenProvider() { result in
-                                completion(result)
+                                switch result {
+                                case let .success(token):
+                                    if token.isExpired {
+                                        completion(.failure(ClientError.ExpiredToken()))
```
b-onc, line 599: There's no backoff and retry mechanism, right?

dmigach: This is a backoff mechanism, because we're inside the code that reobtains expired tokens

b-onc: Can you add a test to check if reconnection is scheduled when we try to connect with an expired token?

dmigach: There was such a test already, and I added another one that checks for the new case

tbarbugli (commented):
I think it makes more sense to apply an exp. backoff  retry when Chat API returns a token expired error (ie. keep track of consecutive errors and schedule the next try using that)

token.isExpired uses the device clock, this might create problems

dmigach:
> I think it makes more sense to apply an exp. backoff retry when Chat API returns a token expired error (ie. keep track of consecutive errors and schedule the next try using that)
> 
> token.isExpired uses the device clock, this might create problems

A newly issued token that expires before we get a response with it and check its expiration is a pretty weird case, but this answers the question from the description, let's do it this way 👌

Sources/StreamChat/ChatClient.swift
b-onc: I can't find the test where we check if there's a reconnection scheduled after an expired token error

dmigach: The name's `test_client_webSocketIsDisconnected_becauseTokenExpired_callsReloadUserIfNeeded`

PR #1309, opened 2021-07-22
[CIS-430] [CIS-968] Custom events support

evsaev:
**This PR**:
- introduces `ChannelEventsController` which handles only events related to the concrete channel
- adds support for custom events

To prepare for sending/receiving custom events declare a payload type:
```swift
extension EventType {
   static let newIdea: Self = "new_idea"
}

struct IdeaEventPayload: CustomEventPayload {
    static let eventType = .newIdea

    let idea: String
}
```

To work with custom events `ChannelEventsController` is needed:
```swift
// Create an instance using the chat client
let channelEventsController = chatClient.channelEventsController(for: *cid*)

// Create an instance using the channel controller. None events are observed until the channel is created.
let channelEventsController = channelController.eventsController() 
```

To send a custom event:
```swift
// Create custom event payload
let idea = IdeaEventPayload(idea: "💡")

// Send custom event with the payload to the channel using `ChannelEventsController `
channelEventsController.sendEvent(idea) { error in 
  ...
}

// The JSON that is sent to the backend:
{
  "event" : {
    "type" : "new_idea",
    "idea" : "💡"
  }
}
```

To receive a custom event:
```swift
// Assign a delegate to events controller
channelEventsController.delegate = delegate

// Implement `EventsControllerDelegate` protocol
    // Check if an event is a custom one
    guard let customEvent = event as? UnknownEvent else { return }
    
    // Get `IdeaEventPayload` and handle it
    if let ideaPayload = customEvent.payload(ofType: IdeaEventPayload.self) {
        print("\(customEvent.userId) has shared an idea: \(ideaPayload.idea)")
    }
}
```


Sources/StreamChat/APIClient/Endpoints/Payloads/UnknownEvent.swift
```diff
@@ -0,0 +1,47 @@
+//
+// Copyright © 2021 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+public protocol CustomEventPayload: Codable, Hashable {
+    static var eventType: String { get }
```
evsaev, line 8: we could make the `EventType` to be a string representable other than an `enum` so customers can declare their own types 🤔 But maybe it makes sense to keep `EventType` for system events only. WDYT?

Sources/StreamChat/APIClient/Endpoints/Payloads/UnknownEvent.swift
```diff
@@ -0,0 +1,47 @@
...
+}
+
+public struct UnknownEvent: Event {
```
evsaev, line 11: The naming is taken from android SDK

Sources/StreamChat/Controllers/EventsController/ChannelEventsController.swift
```diff
@@ -0,0 +1,83 @@
...
+    /// - Returns: A new instance of `ChannelEventsController`.
+    func eventsController() throws -> ChannelEventsController {
+        guard let cid = cid else { throw ClientError.ChannelNotCreatedYet() }
+        
+        return .init(
+            cid: cid,
+            eventSender: .init(database: client.databaseContainer, apiClient: client.apiClient),
+            notificationCenter: client.eventNotificationCenter
+        )
+    }
```
evsaev, line 36: we have a discussion with @tbarbugli that it'd be nice to make this func not throwing 👍 Looking how we can approach this

evsaev: resolved by switching to cid provider closure

Sources/StreamChat/APIClient/Endpoints/Payloads/UnknownEvent.swift
evsaev: made the `EventType` string representable

tbarbugli (approved):

PR #1314, opened 2021-07-23
[CIS-1059] [CIS-860] Refactor `ChatMessageListVC` and `ChatThreadVC`

nuno-vieira:
This PR aims to decouple the `ChatMessageListVC` and make it configurable so that it can be used by the channel view and the thread view. For this to be possible the `ChatMessageListVC` has now a `ChatMessageListVCDataSource` and `ChatMessageListVCDelegate`.

 A new `ChatChannelVC`is introduced that represents the old `ChatMessageListVC`, that it was responsible to display the messages from a channel. The `ChatThreadVC` remains the same and it is responsible for displaying the replies in a thread, but now instead of duplicating the implementation from the channel, both use the `ChatMessageListVC` and configure it for their needs.

### New components hierarchy:
- `ChatChannelVC`:
    - `ChatChannelHeaderView`
    - `ChatMessageListVC`
    - `ComposerVC`
   
- `ChatThreadVC`:
    - `ChatThreadHeaderView`
    - `ChatMessageListVC`
    - `ComposerVC`

Both `ChatChannelVC` and `ChatThreadVC` implement the `ChatMessageListVCDataSource` and `ChatMessageListVCDelegate`.

### Breaking Changes
- Introduces `ChatChannelVC`. If you were subclassing `ChatMessageListVC` in your app, now that component is `ChatChannelVC`, which contains `ChatMessageListVC`. The last one is only responsible to render the message list. If your changes were only message list related, then it should be fine and you should not have breaking changes.
- Replaces `ChatMessageActionsVC.Delegate` with `ChatMessageActionsVCDelegate`
- Renames`ChatChannelListRouter.showMessageList()` -> `showChannel()`
- Removal of `ComposerVCDelegate`
- Replaces `ChatMessageListKeyboardObserver` with `ComposerKeyboardHandler`.

### Fixes
- This ticket also improves the efficiency when loading more messages, since the request was called multiple times before.


DemoApp/CreateChatViewController.swift
```diff
@@ -27,7 +27,7 @@ class CreateChatViewController: UIViewController {
                 super.createNewMessage(text: text)
                 
                 // Present the new chat and controller
-                let vc = ChatMessageListVC()
+                let vc = ChatChannelVC()
```
kojiba, line 30: I would recommend to start removing VC and any other abbreviation from names, but this looks like out of scope for this PR and should be discussed later

>https://github.com/raywenderlich/swift-style-guide#naming
>generally avoiding abbreviations

Sample/LoginViewController.swift
```diff
@@ -143,7 +143,7 @@ extension LoginViewController {
 }
 
 final class MyChatChannelListRouter: ChatChannelListRouter {
-    override func showMessageList(for cid: ChannelId) {
+    override func showChannel(for cid: ChannelId) {
```
kojiba, line 146: regarding naming: it should be consistent across all the places, 
its just a note to double-check
if its a message list, it should be always: MessageListModel, MessageListViewController, MessageListView, etc

@petyosi could you please check how this name across platforms, so we could be consistent

Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
```diff
@@ -0,0 +1,215 @@
+//
+// Copyright © 2021 Stream.io Inc. All rights reserved.
+//
+
+import StreamChat
+import UIKit
+
+/// Controller responsible for displaying the channel messages.
+@available(iOSApplicationExtension, unavailable)
+open class ChatChannelVC: _ViewController, ThemeProvider {
```
kojiba, line 10: Is there any reason we are going for `_ViewController` naming instead of private/public `ViewController` or public suffixed `StreamViewController`?

When user will have multiple ViewController namings, they will resolve correctly due they are placed in separate modules, and you always can specify by module name: StreamChat.ViewController and OtherModule.ViewController and make any typealias for it.

kojiba: https://github.com/raywenderlich/swift-style-guide#class-prefixes

Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
```diff
@@ -0,0 +1,215 @@
...
+
+    /// Component responsible for setting the correct offset when keyboard frame is changed.
+    open lazy var keyboardHandler: KeyboardHandler = ComposerKeyboardHandler(
+        composerParentVC: self,
+        composerBottomConstraint: messageComposerBottomConstraint
+    )
+
+    open lazy var messageListVC: ChatMessageListVC = components
+        .messageListVC
+        .init()
```
kojiba, line 29: short initialisers styling item for TBD

Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
kojiba, line 10: I would recommend put ThemeProvider conformance to `open extension` to make it more readable.
https://github.com/raywenderlich/swift-style-guide#protocol-conformance

Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
```diff
@@ -0,0 +1,215 @@
...
+            self?.messageComposerVC.updateContent()
+        }
+    }
+
+    override open func setUpLayout() {
+        super.setUpLayout()
+
+        view.backgroundColor = appearance.colorPalette.background
+
+        messageListVC.view.translatesAutoresizingMaskIntoConstraints = false
```
kojiba, line 69: this is repeating some N times across the code, `translatesAutoresizingMaskIntoConstraints = false` 
lets put this code into the private UIView extension and make it more sugar
```Swift
private extension UIView {
   func useAutoLayout() {
       translatesAutoresizingMaskIntoConstraints = false
   }
}
```
then it will be `view.useAutoLayout()` instead of `view.translatesAutoresizingMaskIntoConstraints = false`

Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
```diff
@@ -0,0 +1,215 @@
...
+        messageComposerVC.view.translatesAutoresizingMaskIntoConstraints = false
+        addChildViewController(messageComposerVC, targetView: view)
+        messageComposerVC.view.pin(anchors: [.leading, .trailing], to: view)
+        messageComposerVC.view.topAnchor.pin(equalTo: messageListVC.view.bottomAnchor).isActive = true
+        messageComposerBottomConstraint = messageComposerVC.view.bottomAnchor.pin(equalTo: view.bottomAnchor)
+        messageComposerBottomConstraint?.isActive = true
+
+        NSLayoutConstraint.activate([
+            channelAvatarView.widthAnchor.pin(equalTo: channelAvatarView.heightAnchor),
+            channelAvatarView.heightAnchor.pin(equalToConstant: 32)
```
kojiba, line 82: magic number 32 should be moved into 
```Swift
private enum Dimensions {
    static let channelAvatarHeight: CGFloat = 32
}
```

and used as 
`channelAvatarView.heightAnchor.pin(equalToConstant: Dimensions.channelAvatarHeight)`

https://github.com/raywenderlich/swift-style-guide#constants

Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
```diff
@@ -0,0 +1,215 @@
...
+            in: channel,
+            with: AnyRandomAccessCollection(channelController.messages),
+            appearance: appearance
+        )
+    }
+}
+
+extension ChatChannelVC: ChatMessageListVCDelegate {
+    open func chatMessageListVC(
+        _ vc: ChatMessageListVC,
```
kojiba, line 142: if function has 2-3 arguments there is no need to declare them each on a new line
https://github.com/raywenderlich/swift-style-guide#function-declarations

>Keep short function declarations on one line including the opening brace:

Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
```diff
@@ -0,0 +1,215 @@
...
+            // Hide the badge immediately. Temporary solution until CIS-881 is implemented.
+            messageListVC.scrollToLatestMessageButton.content = .noUnread
+        }
+    }
+}
+
+extension ChatChannelVC: ChatChannelControllerDelegate {
+    open func channelController(
+        _ channelController: ChatChannelController,
+        didUpdateMessages changes: [ListChange<ChatMessage>]
```
kojiba, line 186: >Keep short function declarations on one line including the opening brace

Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
```diff
@@ -0,0 +1,215 @@
...
+    open func channelController(
+        _ channelController: ChatChannelController,
+        didUpdateChannel channel: EntityChange<ChatChannel>
+    ) {
+        let channelUnreadCount = channelController.channel?.unreadCount ?? .noUnread
+        messageListVC.scrollToLatestMessageButton.content = channelUnreadCount
+    }
+
+    open func channelController(
+        _ channelController: ChatChannelController,
```
kojiba, line 200: >Keep short function declarations on one line including the opening brace

Sources/StreamChatUI/ChatChannel/ChatChannelVC+SwiftUI.swift
```diff
@@ -0,0 +1,22 @@
+//
+// Copyright © 2021 Stream.io Inc. All rights reserved.
+//
+
+import StreamChat
+import SwiftUI
+
+@available(iOSApplicationExtension, unavailable)
+extension ChatChannelVC: SwiftUIRepresentable {
+    public var content: (
```
kojiba, line 10: I would recommend also make a single line for two-argument declarations/function/etc

Sources/StreamChatUI/ChatMessageList/ChatMessageListVCDataSource.swift
```diff
@@ -0,0 +1,34 @@
...
+    /// Asks the data source to return the number of messages in the message list.
+    /// - Parameter vc: The message list requesting the number of messages.
+    func numberOfMessages(in vc: ChatMessageListVC) -> Int
+
+    /// Asks the data source for the message in a particular location of the message list.
+    /// - Parameters:
+    ///   - vc: The message list requesting the message.
+    ///   - indexPath: An index path locating the row in the message list.
+    func chatMessageListVC(
+        _ vc: ChatMessageListVC,
```
kojiba, line 22: >double-arguments can be on same lane

Sources/StreamChatUI/ChatMessageList/ChatMessageListVCDataSource.swift
```diff
@@ -0,0 +1,34 @@
...
+        messageAt indexPath: IndexPath
+    ) -> ChatMessage?
+
+    /// Asks the data source for the message layout options in a particular location of the message list.
+    /// - Parameters:
+    ///   - vc: The message list requesting the layout options.
+    ///   - indexPath: An index path locating the row in the message list.
+    func chatMessageListVC(
```
kojiba, line 30: >double-arguments can be on same lane

Sources/StreamChatUI/ChatMessageList/ChatMessageListVCDelegate.swift
```diff
@@ -0,0 +1,38 @@
...
+import UIKit
+
+/// The object that acts as the delegate of the message list.
+public protocol ChatMessageListVCDelegate: AnyObject {
+    /// Tells the delegate the message list is about to draw a message for a particular row.
+    /// - Parameters:
+    ///   - vc: The message list informing the delegate of this event.
+    ///   - indexPath: An index path locating the row in the message list.
+    func chatMessageListVC(
+        _ vc: ChatMessageListVC,
```
kojiba, line 15: >double-arguments can be on same lane

Sources/StreamChatUI/ChatMessageList/ChatMessageListVCDelegate.swift
```diff
@@ -0,0 +1,38 @@
...
+        willDisplayMessageAt indexPath: IndexPath
+    )
+
+    /// Tells the delegate when the user scrolls the content view within the receiver.
+    /// - Parameters:
+    ///   - vc: The message list informing the delegate of this event.
+    ///   - scrollView: The scroll view that belongs to the message list.
+    func chatMessageListVC(
+        _ vc: ChatMessageListVC,
```
kojiba, line 24: >double-arguments can be on same lane

Sources/StreamChatUI/ChatThread/ChatThreadVC+SwiftUI.swift
```diff
@@ -0,0 +1,22 @@
...
+// Copyright © 2021 Stream.io Inc. All rights reserved.
+//
+
+import StreamChat
+import SwiftUI
+
+@available(iOSApplicationExtension, unavailable)
+extension ChatThreadVC: SwiftUIRepresentable {
+    public var content: (
+        channelController: ChatChannelController,
```
kojiba, line 11: >double-arguments can be on same lane

Sources/StreamChatUI/Composer/ComposerKeyboardHandler.swift
```diff
@@ -0,0 +1,70 @@
...
+open class ComposerKeyboardHandler: KeyboardHandler {
+    public weak var composerParentVC: UIViewController?
+    public weak var composerBottomConstraint: NSLayoutConstraint?
+
+    /// The component for handling keyboard events and adjust the composer.
+    /// - Parameters:
+    ///   - composerParentVC: The parent view controller of the composer.
+    ///   - composerBottomConstraint: The bottom constraint of the composer.
+    public init(
+        composerParentVC: UIViewController,
```
kojiba, line 26: >double-arguments can be on same lane

Sources/StreamChatUI/ChatThread/ChatThreadVC.swift
```diff
@@ -0,0 +1,227 @@
...
+        _ vc: ChatMessageListVC,
+        scrollViewDidScroll scrollView: UIScrollView
+    ) {
+        // No-op
+    }
+}
+
+extension ChatThreadVC: ChatMessageControllerDelegate {
+    open func messageController(
+        _ controller: ChatMessageController,
```
kojiba, line 203: >double-arguments can be on same lane

Sources/StreamChatUI/ChatThread/ChatThreadVC.swift
```diff
@@ -0,0 +1,227 @@
...
+            listChange = .update(item, index: indexPath)
+        case let .remove(item):
+            listChange = .remove(item, index: indexPath)
+        }
+
+        messageListVC.updateMessages(with: [listChange])
+    }
+
+    open func messageController(
+        _ controller: ChatMessageController,
```
kojiba, line 222: >double-arguments can be on same lane

Sources/StreamChatUI/ChatThread/ChatThreadVC.swift
```diff
@@ -0,0 +1,227 @@
...
+            dismiss(animated: true) { [weak self] in
+                self?.messageComposerVC.content.quoteMessage(message)
+            }
+        default:
+            return
+        }
+    }
+
+    open func chatMessageListVC(
+        _ vc: ChatMessageListVC,
```
kojiba, line 194: >double-arguments can be on same lane

Sources/StreamChatUI/ChatThread/ChatThreadVC.swift
```diff
@@ -0,0 +1,227 @@
...
+        scrollViewDidScroll scrollView: UIScrollView
+    ) {
+        // No-op
```
kojiba, line 197: do we have a marking policy for such comments? I would assume `Note: Do nothing cause its a UI requirement` will looks better, or feel free to propose something in this approach 

>https://github.com/raywenderlich/swift-style-guide#comments

Sources/StreamChatUI/ChatThread/ChatThreadVC.swift
```diff
@@ -0,0 +1,227 @@
...
+    open lazy var keyboardHandler: KeyboardHandler = ComposerKeyboardHandler(
+        composerParentVC: self,
+        composerBottomConstraint: messageComposerBottomConstraint
+    )
+
+    /// User search controller passed directly to the composer
+    open lazy var userSuggestionSearchController: ChatUserSearchController =
+        channelController.client.userSearchController()
+
+    open lazy var messageListVC: ChatMessageListVC = components
```
kojiba, line 31: in 3 line above there is usage of:
>    /// Component responsible for setting the correct offset when keyboard frame is changed
    open lazy var keyboardHandler: KeyboardHandler = ComposerKeyboardHandler(
        composerParentVC: self,
        composerBottomConstraint: messageComposerBottomConstraint
    )
    
could we think about removing short `.init` statements, they are decreasing readability, cause you won't see exact type

kojiba (commented):
Some comments to be addressed

Sources/StreamChatUI/ChatThread/ChatThreadVC.swift
dmigach: There is no exact type because customers can set their own 🙂

DemoApp/CreateChatViewController.swift
nuno-vieira: The reason we have `VC` is that we had a lot of big names in the project, and so the team decided we would suffix the view controllers with `VC`. Anyways, we then further improved the naming, and they are not that big anymore, so we can revisit this later since changing it now will cause a lot of breaking changes to the customers (more than we are already causing here). But I also prefer not having abbreviations. But we can do this now if the team agrees @dmigach @tbarbugli @SagarSDagdu @b-onc

Sample/LoginViewController.swift
nuno-vieira: The `MessageList` is the component to render a list of messages, just like the name describes it. But the message list can be rendered in a Channel or in a Thread. In this case, we are showing a channel (with a message list).

Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
nuno-vieira: I remember the reason was that the customers should not be using `_ViewController` or at least should be careful when using it. I'm not sure maybe @b-onc can clarify. I would also prefer naming it just `ViewController` or `StreamViewController`. Either way, I believe this is out of scope of the PR, and all these overall improvements to the SDK can be discussed later 👍

nuno-vieira: Theme Provider doesn't require implementations. so the extension would be `extensions ChatChannelVC: ThemeProvider {}` which is way less clear.

Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
nuno-vieira: We already have a helper for this. `withoutAutoresizingMaskConstraints`. In most of the codebase, we are using the helper, in some parts we are not, it depends on where it makes sense.

Sample/LoginViewController.swift
kojiba: Do we have a difference between Channel/Tread/MessageList somewhere documented?

Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
kojiba: Also as idea: we could put a `Note:` comments for empty extension

Sources/StreamChatUI/ChatThread/ChatThreadVC.swift
kojiba: we could try to directly state that this is a injected Type we are using like 
`components.messageComposerVC.Type()` or something with full-type-name constructor

Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
nuno-vieira: The team is well whereof the swift style guide btw 🙂 We try to follow when it makes sense, but we don't follow it religiously. 

Either way, in this case, the style guide doesn't mention that a function with 2-3 arguments should be in one line. It mentions "short" function declarations. A function can have 1 argument only hand be a long declaration. In this case, all function names in the delegate are pretty long, and so it makes sense to be in each line.

Sources/StreamChatUI/ChatThread/ChatThreadVC.swift
nuno-vieira: Will prove a better comment 👍

Sources/StreamChatUI/ChatThread/ChatThreadVC.swift
nuno-vieira: The exact type is on the property of the variable (`messageListVC: ChatMessageListVC`), and it is actually a rule of the SDK when declaring properties on the UI components, you need to explicitly declare the type. But the customer can override this, or set a subclass of the component, so we need to use `init()`. `components.messageComposerVC.init()` vs `components.messageComposerVC.Type()` is a minor detailed IMO, but we can discuss later in the team if you feel it is worth it 👍

Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
kojiba: Okay than this could be used, also ideas how to make a DRY-ied code
https://en.wikipedia.org/wiki/Don%27t_repeat_yourself

```Swift
[firstView, secondView, thirdView].forEach {
   $0. withoutAutoresizingMaskConstraints
}
```

This will decrease amount of LOCs -> decrease cost of support / amount of possible bugs

kojiba: also upper solution with array also can have a nice sugared extension with specified `where` type for Array generic 
and in result will be
```Swift
[firstView, secondView, thirdView].withoutAutoresizingMaskConstraints()
```

Sources/StreamChatUI/ChatThread/ChatThreadVC.swift
kojiba: lets put on TBD list, readability always worth the discussion :)

Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
kojiba: okay, lets TBD on maxLineLenght and enforce it with a swiftlint config

Sample/LoginViewController.swift
nuno-vieira: Not sure, but it seems pretty obvious. You open a channel, you have a list of messages, you open a thread, you have a list of messages. This is the aim of this PR, is to reuse the `messageList` component between the Channel and Thread, before it was duplicated code

Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
nuno-vieira: We already have that in place through **swiftformat** pre-commit hook in `hooks/pre-commit.sh`. The rules are in `.swiftformat` file and the max length is 132 `--maxwidth 132`

Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
nuno-vieira: Done ✅

Sources/StreamChatUI/ChatThread/ChatThreadVC.swift
nuno-vieira: ✅

Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
nuno-vieira: We do have DRY coding (hence the need for this PR), but right now we want to focus mainly on architectural problems. Don't think we should focus on small details like this. Also +LOCs != Cost of Support. Code should be clear and readable, and normally that means +LOCs.

kojiba: Making copy-paste code = have problems in near future, please DRY PR as much as possible. 
We should focus on every detail we have cause we making open-source product which should show good approach to customers/other our code readers.

nuno-vieira: You are asking to change `messageComposerVC.view.translatesAutoresizingMaskIntoConstraints = false` -> `messageComposerVC.view.withoutAutoresizingMaskConstraints()` . It is replacing one line, with another one line. Also currently `withoutAutoresizingMaskCOnstraints` returns the view itself, so I will need to do `_ = messageComposerVC.view.withoutAutoresizingMaskConstraints` which will look even worse. (And no I can't use @discardableResult because it is a property). So in order to fix this, I will need to convert `withoutAutoresizingMaskConstraints` into a function instead which will cause a lot of changes in an already huge PR. 

Btw, DRY is not always a good thing: [Why DRY principle is bad advice](https://rotemtam.com/2020/05/18/the-dry-principle-is-bad-advice/).

kojiba: I am asking to do generic extension for array where object is UIView and make a 
>[firstView, secondView, thirdView].withoutAutoresizingMaskConstraints()

in current case it will be

```
messageListVC.view.translatesAutoresizingMaskIntoConstraints = false
messageComposerVC.view.translatesAutoresizingMaskIntoConstraints = false
```
VS

`[messageListVC.view, messageComposerVC.view].withoutAutoresizingMaskConstraints()`

and continue to use this array extension in future
for now it 2 line vs 1 line, in future it will N lines vs 1 line

nuno-vieira: `withoutAutoresizingMaskConstraints` is a property, not a function. As I said, changing to a function will do a lot of changes to the codebase and increasing an already 1000+ lines PR to even more. Feel free to create a ticket and address this issue 🙂

kojiba: Okay, lets put this to backlog then, 
I will show after, this will be a change for ~10 LOCs

kojiba: https://stream-io.atlassian.net/browse/CIS-1133

kojiba (approved):
overal LGTM

Sources/StreamChatUI/ChatThread/ChatThreadVC.swift
```diff
@@ -0,0 +1,228 @@
...
+    override open func viewDidDisappear(_ animated: Bool) {
+        super.viewDidDisappear(animated)
+
+        resignFirstResponder()
+
+        keyboardHandler.stop()
+    }
+}
+
+extension ChatThreadVC: ChatMessageListVCDataSource {
```
evsaev, line 106: extension -> class body

Sources/StreamChatUI/ChatThread/ChatThreadVC.swift
```diff
@@ -0,0 +1,228 @@
...
+                appearance: appearance
+            )
+
+        layoutOptions.remove(.threadInfo)
+
+        return layoutOptions
+    }
+}
+
+extension ChatThreadVC: ChatMessageListVCDelegate {
```
evsaev, line 164: extension -> class body

Sources/StreamChatUI/ChatThread/ChatThreadVC.swift
```diff
@@ -0,0 +1,228 @@
...
+    open func chatMessageListVC(
+        _ vc: ChatMessageListVC,
+        scrollViewDidScroll scrollView: UIScrollView
+    ) {
+        // No-op. By default this component is not interest in scrollView events,
+        // but you as customer can override this function and provide an implementation.
+    }
+}
+
+extension ChatThreadVC: ChatMessageControllerDelegate {
```
evsaev, line 202: extension -> class body

evsaev (dismissed):
Great job 💪 Let's move delegate/datasource methods from extensions and we are good to 🚢

Sources/StreamChatUI/ChatMessageList/ChatMessageListView.swift
```diff
@@ -147,14 +147,13 @@ open class ChatMessageListView: UITableView, Customizable, ComponentsProvider {
         let lastMessageIndexPath = IndexPath(row: 0, section: 0)
         let prevMessageIndexPath = IndexPath(row: 1, section: 0)
 
-        if
-            rectForRow(at: prevMessageIndexPath).minY < contentOffset.y,
-            rowsRange.contains(prevMessageIndexPath.row) {
-            scrollToRow(at: prevMessageIndexPath, at: .top, animated: false)
+        if rectForRow(at: prevMessageIndexPath).minY < contentOffset.y,
+           rowsRange.contains(prevMessageIndexPath.row) {
+            scrollToRow(at: prevMessageIndexPath, at: .bottom, animated: animated)
```
evsaev, line 152: The cell is inverted over Y axis so we should scroll to `.top`. It's confusing, I now 😄

Sources/StreamChatUI/ChatMessageList/ChatMessageListView.swift
```diff
@@ -147,14 +147,13 @@ open class ChatMessageListView: UITableView, Customizable, ComponentsProvider {
...
         }
         
         if rowsRange.contains(lastMessageIndexPath.row) {
-            scrollToRow(at: lastMessageIndexPath, at: .top, animated: animated)
+            scrollToRow(at: lastMessageIndexPath, at: .bottom, animated: animated)
```
evsaev, line 156: the same here as https://github.com/GetStream/stream-chat-swift/pull/1314/files#r695162879

Sources/StreamChatUI/ChatMessageList/ChatMessageListView.swift
nuno-vieira: With `.top` the scroll animation was a bit weird actually, or maybe was the lack of the animation 🤔 But either way, it doesn't seem to have a difference 🤔

kojiba (approved):
LGTM

Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
```diff
@@ -0,0 +1,220 @@
...
+        guard let channel = channelController.channel else { return [] }
+
+        return components.messageLayoutOptionsResolver.optionsForMessage(
+            at: indexPath,
+            in: channel,
+            with: AnyRandomAccessCollection(channelController.messages),
+            appearance: appearance
+        )
+    }
+}
```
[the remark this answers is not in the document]
kojiba: this is incorrect, they can be ovveridable, please just mark all extensions `public` as:
```Swift
public extension ChatChannelVC: ChatMessageListVCDataSource {
...
}
```

and please do not revert back @nuno-vieira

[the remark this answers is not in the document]
nuno-vieira: @kojiba Nope. @evsaev is correct. You need to mark all open func's inside the extension with `@objc` for this to work.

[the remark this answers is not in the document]
SagarSDagdu: @nuno-vieira Yes you are right, OR, the protocol can be `@objc` for this to work

evsaev (approved):
Great job! 🎉

CHANGELOG.md
```diff
@@ -3,6 +3,27 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 # Upcoming
 
+### ⚠️ Breaking Changes from `4.0-beta.11`
+- `ChatMessage.threadParticipants` is now an Array instead of Set [#1398](https://github.com/GetStream/stream-chat-swift/pull/1398)
+- Introduces `ChatChannelVC` and removes responsibilties of `ChatMessageListVC`. The latter now is only responsible to render the message list layout, the data is provided by `ChatChannelVC` or `ChatThreadVC`. [#1314](https://github.com/GetStream/stream-chat-swift/pull/1314)
+- Replaces `ChatMessageActionsVC.Delegate` with `ChatMessageActionsVCDelegate` [#1314](https://github.com/GetStream/stream-chat-swift/pull/1314)
+- Renames`ChatChannelListRouter.showMessageList()` -> `showChannel()` [#1314](https://github.com/GetStream/stream-chat-swift/pull/1314)
```
evsaev, line 10:
```suggestion
- Renames `ChatChannelListRouter.showMessageList()` -> `showChannel()` [#1314](https://github.com/GetStream/stream-chat-swift/pull/1314)
```

CHANGELOG.md
```diff
@@ -3,6 +3,27 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
+    - `ChatChannelHeaderView`
+    - `ChatMessageListVC`
+    - `ComposerVC`
+   
+- `ChatThreadVC`:
+    - `ChatThreadHeaderView`
+    - `ChatMessageListVC`
+    - `ComposerVC`
+
+A new `ChatChannelVC`is introduced that represents the old `ChatMessageListVC`, that it was responsible to display the messages from a channel. The `ChatThreadVC` remains the same and it is responsible for displaying the replies in a thread, but now instead of duplicating the implementation from the channel, both use the `ChatMessageListVC` and configure it for their needs. For this to be possible the `ChatMessageListVC` has now a `ChatMessageListVCDataSource` and `ChatMessageListVCDelegate`. Both `ChatChannelVC` and `ChatThreadVC` implement the `ChatMessageListVCDataSource` and `ChatMessageListVCDelegate`.
```
evsaev, line 25:
```suggestion
A new `ChatChannelVC` is introduced that represents the old `ChatMessageListVC`, that it was responsible to display the messages from a channel. The `ChatThreadVC` remains the same and it is responsible for displaying the replies in a thread, but now instead of duplicating the implementation from the channel, both use the `ChatMessageListVC` and configure it for their needs. For this to be possible the `ChatMessageListVC` has now a `ChatMessageListVCDataSource` and `ChatMessageListVCDelegate`. Both `ChatChannelVC` and `ChatThreadVC` implement the `ChatMessageListVCDataSource` and `ChatMessageListVCDelegate`.
```

CHANGELOG.md
```diff
@@ -3,6 +3,27 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
 
+### ⚠️ Breaking Changes from `4.0-beta.11`
+- `ChatMessage.threadParticipants` is now an Array instead of Set [#1398](https://github.com/GetStream/stream-chat-swift/pull/1398)
+- Introduces `ChatChannelVC` and removes responsibilities of `ChatMessageListVC`. The latter now is only responsible to render the message list layout, the data is provided by `ChatChannelVC` or `ChatThreadVC`. [#1314](https://github.com/GetStream/stream-chat-swift/pull/1314)
+- Replaces `ChatMessageActionsVC.Delegate` with `ChatMessageActionsVCDelegate` [#1314](https://github.com/GetStream/stream-chat-swift/pull/1314)
+- Renames `ChatChannelListRouter.showMessageList()` -> `showChannel()` [#1314](https://github.com/GetStream/stream-chat-swift/pull/1314)
+- Removal of `ComposerVCDelegate` [#1314](https://github.com/GetStream/stream-chat-swift/pull/1314)
+- Replaces `ChatMessageListKeyboardObserver` with `ComposerKeyboardHandler` [#1314](https://github.com/GetStream/stream-chat-swift/pull/1314)
+
+#### Understading `ChatChannelVC` vs `ChatTheadVC` vs `ChatMessageListVC`
```
nuno-vieira, line 14:
```suggestion
#### Understanding `ChatChannelVC` vs `ChatTheadVC` vs `ChatMessageListVC`
```

tbarbugli (dismissed):

PR #1318, opened 2021-07-26
[CIS-1042] Fix deleted message's appearance on channel list

SagarSDagdu:
### What does this PR do
- This PR fixes the appearance of a message in the channel list in the case where the channel's last message is a deleted one. Instead of showing the actual text of the message, we show the "Message Deleted" string.


DominikBucher12 (approved):
💪

PR #1320, opened 2021-07-27
[CIS-1037] Fix misplaced reaction bubbles

DominikBucher12:
# What this PR do:
- Fixes misplaced reaction bubbles on short messages caused by pinning the center X anchor to mainContainer instead of `bubbleView`/`bubbleContentContainer`. The width of `mainContainer` obviously expanded when there were threads or timestamp and it was longer than the bubble message (if user wrote short message...) 
# How to test it
- Go to an chat which contains short messages and scroll through the chat to find out if the reactions are pinned right
# Where you can start
-  `ChatMessageContentView.swift ` 

Before: 

Before

After: 
After


SagarSDagdu (approved):
LGTM 👍

b-onc (approved):

PR #1321, opened 2021-07-27
[CIS-1049] Add the `Combine` guide for the SDK

SagarSDagdu:
### This PR adds a guide for using the `Combine` components in the SDK with relevant examples
- For each controller delegate in our SDK, we also provide relevant publishers for clients that want to use `Combine` to observe the data. They can be referred to in the `Controller+Combine.swift` files
- Using these publishers, they can observe changes to the data and update their views accordingly

### We have documented this in the following way:
- Have a separate `Combine` section in the docs
- Have a separate page for each of the controllers
- In each of those pages, list down all the publishers provided by the controller, add a code snippet to use the publisher AND create some sample `UIKit` views to consume the data from these publishers.

**NOTE**: We do not cover `SwiftUI` examples here as for `SwiftUI`, we can use the `ObservableObject` versions of the controllers instead of using the publishers directly(This is the preferred approach)
This will be covered in the `SwiftUI` guide.

### DIFF
--- a/docusaurus/docs/iOS/combine/combine-overview.md
+++ b/docusaurus/docs/iOS/combine/combine-overview.md
@@ -1,0 +1,29 @@
+---
+title: Overview
+---
+
+The `StreamChat` framework ships with controllers that are fully compatible with `Combine`.
+You can use these combine-compatible controllers to observe the changes to the controllers, and build your own views to consume the data.
+
+Each controller exposes publishers based on the kind of data they control and allow you to observe changes. You can find examples on how to build your own views using these publishers here.
+
+## Using the `Combine` extensions of Controllers
+
+When using combine extensions of the controllers, you need to call the `synchronize` method on your controller instance before starting to observe the publishers.
+This is done to avoid the side-effects from initialization and to make sure that the controller is ready to be used.
+
+i.e. a channel controller needs to watch a channel before its publisher can emit changes.
+
+It's a best practice to pass the synchronize a completion block to handle error cases. However, error handling is optional, and you can use synchronize as a fire and forget method.
+
+```swift
+    /// ...
+    controller.synchronize { error in
+        /// something did not work with the controller setup
+        if error != nil {
+            log.assertionFailure(error!)
+            return
+        }
+    }
+    /// ...
+```
--- a/docusaurus/docs/iOS/combine/combine-overview.md
+++ b/docusaurus/docs/iOS/combine/combine-overview.md
@@ -2,8 +2,8 @@
 title: Overview
 ---
 
-The `StreamChat` framework ships with controllers that are fully compatible with `Combine`.
-You can use these combine-compatible controllers to observe the changes to the controllers, and build your own views to consume the data.
+The `StreamChat` framework ships with `Combine` publishers out of the box.
+`Combine` publishers are exposed by `StreamChat` controller classes, these publishers allow you to observe data and to build your own views.
 
 Each controller exposes publishers based on the kind of data they control and allow you to observe changes. You can find examples on how to build your own views using these publishers here.
 
--- a/docusaurus/sidebars-ios.json
+++ b/docusaurus/sidebars-ios.json
@@ -23,7 +23,7 @@
       ]
     },
     {
-      "Controllers with Combine" : [
+      "Combine" : [
         "combine/combine-overview",
         "combine/connection",
         "combine/channels",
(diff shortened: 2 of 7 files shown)


tbarbugli (changes requested):
I think we can make this more useful by saying something like this:

- You can build custom components and views on top of the SDK with Combine
- The SDK ships with controller classes that allow you to retrieve and observe data
- Example 1: a simple SwiftUI view using Combine (typing events?)
- Example 2: a UIKit view using Combine (connection status?)
- List all the observables with a quick description and code snippet that shows how to get the publisher

I am not 100% sure if this is needed for Combine publishers but delegates require you to call `synchronize` on the controller to get started.

SagarSDagdu:
@tbarbugli Thanks for the detailed review. 
Based on your review, I thought it makes sense to introduce a separate section for using the `Combine` extensions of our controllers
It will cover the exact pages you have covered in the "Controller and Delegates" section, but all using combine.

I have added the channels page for now, which
- lists down all the publishers in the channel controllers along with their description
- has some examples of `UIKit` views with combine

I plan to do the same for other controllers too. Let me know your thoughts

docusaurus/sidebars-ios.json
```diff
@@ -22,6 +22,15 @@
         "controllers/messages"
       ]
     },
+    {
+      "Controllers with Combine" : [
```
tbarbugli, line 26:
```suggestion
      "Combine" : [
```

docusaurus/docs/iOS/combine/combine-overview.md
```diff
@@ -0,0 +1,29 @@
+---
+title: Overview
+---
+
+The `StreamChat` framework ships with controllers that are fully compatible with `Combine`.
```
tbarbugli, line 5:
```suggestion
The `StreamChat` framework ships with `Combine` publishers out of the box. 
```

docusaurus/docs/iOS/combine/combine-overview.md
```diff
@@ -0,0 +1,29 @@
...
+You can use these combine-compatible controllers to observe the changes to the controllers, and build your own views to consume the data.
```
tbarbugli, line 6:
```suggestion
`Combine` publishers are exposed by `StreamChat` controller classes, these publishers allow you to observe data and to build your own views.
```

docusaurus/docs/iOS/combine/combine-overview.md
```diff
@@ -0,0 +1,29 @@
...
+
+Each controller exposes publishers based on the kind of data they control and allow you to observe changes. You can find examples on how to build your own views using these publishers here.
+
+## Using the `Combine` extensions of Controllers
+
+When using combine extensions of the controllers, you need to call the `synchronize` method on your controller instance before starting to observe the publishers.
```
tbarbugli, line 12: Are you sure that this is needed for `Combine` publishers as well?

docusaurus/sidebars-ios.json
```diff
@@ -22,6 +22,15 @@
...
+        "combine/combine-overview",
+        "combine/connection",
+        "combine/channels",
+        "combine/users",
+        "combine/messages"
```
tbarbugli, line 31: we need one for events as well

tbarbugli: (we allow you to subscribe to events with combine publishers)

tbarbugli (changes requested):

docusaurus/sidebars-ios.json
SagarSDagdu: Done

docusaurus/docs/iOS/combine/combine-overview.md
SagarSDagdu: Done

docusaurus/docs/iOS/combine/combine-overview.md
SagarSDagdu: Done

docusaurus/sidebars-ios.json
SagarSDagdu: Thanks for this input, added one for events as well!

docusaurus/docs/iOS/combine/combine-overview.md
SagarSDagdu: `Combine` publishers internally use the delegates, and I confirmed that `synchronize` is needed when using them too.

tbarbugli (approved):

PR #1323, opened 2021-07-27
[CIS-1071] Fix member removed from Channel still present in MemberListController

b-onc:
This happened because we forgot to update the queries of Member when it's removed. A similar bug may be present in MemberAdded and MemberUpdated events. For those events, we need to re-fetch the queries, since Client doesn't have a way to apply filters.


dmigach (approved):

DominikBucher12 (approved):
LGTM!

tbarbugli:
@dmigach @DominikBucher12 why are you approving a PR with broken CI?

PR #1324, opened 2021-07-27
CIS-1019 Controller docs

tbarbugli:
Adds docs for controllers and delegates

image


docusaurus/docs/iOS/controllers/channels.md
```diff
@@ -0,0 +1,204 @@
+---
+title: Channel Controllers
+---
+
+## ChannelListController
+
+`ChatChannelListController` allows you to observe a list of chat channels based on the provided query and to paginate channels.
```
nikomancy, line 7: Link to more information about pagination.

docusaurus/docs/iOS/controllers/channels.md
```diff
@@ -0,0 +1,204 @@
...
+) {}
+
+func channelController(_ channelController: ChatChannelController, didReceiveMemberEvent: MemberEvent) {}
+```
+
+### Example: Typing Indicator
+
+Let's build a simple UIView that shows the current unread count for the current user.
+
+```swift
```
nikomancy, line 51: This is a very large block of code. Consider breaking it into smaller blocks with prefacing content explaining the "Why" behind each of the steps or adding some comments above the two classes explaining what each does at a high level.

docusaurus/docs/iOS/controllers/channels.md
```diff
@@ -0,0 +1,204 @@
...
+    _ controller: ChatChannelListController,
+    didChangeChannels changes: [ListChange<ChatChannel>]
+)
+```
+
+## ChannelController
+
+`ChatChannelController` allows you to observe and mutate data for one channel.
+
+### Channel Delegate
```
nikomancy, line 24: Is the `ChatChannelControllerDelegate` protocol something we have documented in the API reference for iOS? If yes, we should link to it from this section.

docusaurus/docs/iOS/controllers/channels.md
```diff
@@ -0,0 +1,204 @@
...
+## ChannelListController
+
+`ChatChannelListController` allows you to observe a list of chat channels based on the provided query and to paginate channels.
+
+### ChannelList Delegate
+
+Classes that conform to the `ChatChannelListControllerDelegate` protocol will receive changes to the queried list of channels.
+
+```swift
+func controller(
```
nikomancy, line 14: Consider changing `controller` to channelListController

Ignore this if there's a Swift convention underlying this just being called `controller`

docusaurus/docs/iOS/controllers/channels.md
```diff
@@ -0,0 +1,204 @@
...
+    didUpdateMessages changes: [ListChange<ChatMessage>]
+) {}
+
+func channelController(
+    _ channelController: ChatChannelController,
+    didChangeTypingUsers typingUsers: Set<ChatUser>
+) {}
+
+func channelController(_ channelController: ChatChannelController, didReceiveMemberEvent: MemberEvent) {}
+```
```
nikomancy, line 45: These functions seem like they do not require each other to function. If they do not, we should break this code block up into separate units with some prefacing content about what they do to improve scan-ability.

docusaurus/docs/iOS/controllers/channels.md
```diff
@@ -0,0 +1,204 @@
...
+        controller.synchronize {error in
+            if error != nil {
+                log.assertionFailure(error!)
+            }
+        }
+
+        navigationItem.titleView = membersSearchResultsView
+    }
+}
+```
```
nikomancy, line 189: Consider breaking this block into smaller blocks with prefacing content explaining the "Why" behind each of the steps or adding some comments above the two classes explaining what each does at a high level.

docusaurus/docs/iOS/controllers/controllers-overview.md
```diff
@@ -2,75 +2,27 @@
...
-#### `ChatMessageController`
-Used for operations related to a message, like deleting the message, adding reacting, and observing its changes.
+Each controller exposes API functionality and supports delegation, controllers their delegates are documented here based on the kind of data they control and allow you to observe. You can find examples on how to build your own view as well.
 
-*Typical usage:* Message detail screen, quick actions on a message ("Add Reaction"), ...
+## Using Controllers
 
-&nbsp;
+Controllers expose functionality via event subscriptions and API calls. In most cases you need to call the `synchronize` method on your controller instance. This is done to 
+avoid side-effects from initialization and to make sure that the controller is ready to be used.
```
nikomancy, line 13: - [ ] Add a code block generic example of calling synchronize with a controller.

nikomancy: Writing down things as you go is dangerous and sometime you tell somebody to do something they already did. Everybody learn from my lesson. 🤦

docusaurus/docs/iOS/controllers/controllers-overview.md
```diff
@@ -2,75 +2,27 @@
...
 
-#### `ChatUserSearchController`
-Used for performing a full-text search in available users by their name and id.
+ie. a channel controller needs to watch a channel before it's delegate can receive calls.
 
-*Typical usage:* User mention suggestion, add to channel suggestion, search users screen, ...
+The `synchronize` can be used as a fire and forget method but it is recommended to pass a completion block to handle error cases as well.
```
nikomancy, line 16: Let's rewrite this to lead with the recommendation.

```md
It's a best practice to pass `synchronize` a completion block to handle error cases but this is optional. Error handling is optional and you can use `synchronize` as a fire and forget method if you want to.
```

docusaurus/docs/iOS/controllers/channels.md
tbarbugli: this is a self-explanatory way to explain the delegate protocol

docusaurus/docs/iOS/controllers/channels.md
tbarbugli: see below (protocol are similar interfaces)

docusaurus/docs/iOS/controllers/channels.md
tbarbugli: good point, I dont want to document controllers here since it's mainly an LLC thing but adding something to pagination makes sense (not a code example though, that's way too much code ;))

nikomancy (approved):
Content looks solid.

PR #1328, opened 2021-07-28
[CIS-1073] Change ContainerStackView assertions

b-onc:
UIContainerView has these operations as no-op and so should we


DominikBucher12 (approved):
🛳🇮🇹

PR #1330, opened 2021-07-28
[CIS-1062] Support custom localization

DominikBucher12:
# What this PR do:
- Adds support for overriding default english (and possibly other) localizations in Apps using StreamChat.  Also adds documentation on how to use custom localization.
# How to test it
- Run the Demo Application in Czech language and check if the static texts are translated to Czech language.
# Where you can start
-  `Appearance.default.localizationProvider`, `SwiftGen` config.

# How this works:

`Appearance.default.localizationProvider` is simple closure, which defines how what format of localized strings will be used. Once the localizationProvider is set, it is used inside internal enum `L10n`, which will use the format for getting localized strings.


docusaurus/docs/iOS/customization/localization.md
```diff
@@ -0,0 +1,138 @@
...
+```
+3. Now everything that has left is to implement your  `strings` and `stringsdict`  files for different languages or even customize the default strings for your own.
+
+:::caution
+Because of internal implementation of `L10n` inside StreamChatSDK, you should assign the custom `localizationProvider` only to `Appearance.default` instance. Setting the provider to custom appearance instance would cause custom localization not to won't work.
+:::
+
+## Resources
+
+To ease your work, you can just copy the following content of files and add it to your current Localization:
```
tbarbugli, line 32: - nice to have: generate the english version of all strings using code-generation
- we should not include the list of strings here, instead we should point to the files in Github

docusaurus/docs/iOS/customization/localization.md
```diff
@@ -0,0 +1,138 @@
+---
+title: Localization
```
tbarbugli, line 2: Other SDKs use `Translation and i18n` as the title

docusaurus/docs/iOS/customization/localization.md
```diff
@@ -0,0 +1,138 @@
...
+---
+
+StreamChat offers pre-defined localization for text indicating across the whole UISDK. Currently, StreamChat supports only English language, but if you want to provide your custom strings, it's very easy to do.
```
tbarbugli, line 5: > The SDK allows you to translate all strings used in UI components. To do this you only need to provide a `strings` and `stringsdict` file for each language that you want to translate.

docusaurus/docs/iOS/customization/localization.md
```diff
@@ -0,0 +1,138 @@
...
+Appearance.default.localizationProvider = { key, table in
+    // For most cases, you will probably use Bundle.main, unless your strings are in different Bundle.
+    // Note: `table` is equivalent to filename of the `Localization.strings` file without file extension. 
+    Bundle.main.localizedString(forKey: key, value: nil, table: table)
+}
+```
+3. Now everything that has left is to implement your  `strings` and `stringsdict`  files for different languages or even customize the default strings for your own.
+
+:::caution
+Because of internal implementation of `L10n` inside StreamChatSDK, you should assign the custom `localizationProvider` only to `Appearance.default` instance. Setting the provider to custom appearance instance would cause custom localization not to won't work.
```
tbarbugli, line 27: I dont understand what this means

docusaurus/docs/iOS/customization/localization.md
```diff
@@ -0,0 +1,138 @@
...
+## Adding Custom Localization
+
+As said earlier, StreamChatUI uses `strings` and `stringsdict` files to handle localize the framework.
+`Appearance` structure contains closure `localizationProvider`, which is responsible for deriving `localizedString` from exact bundle. 
+By default, the bundle is  `StreamChatUI` framework bundle, but you can customize it to anything you need in 3 steps:
+
+1. First thing you need to do is to copy `StreamChatUI` localization keys into your `strings` file (Or just copy the default `Localizable.strings` and `Localizable.stringsdict` files to your project from our repository).
+2. Once you have those files in your application, you need to set the `localizationProvider` to provide your `AppBundle` instead of `StreamChatUI` frameworks like this: 
+```swift
+Appearance.default.localizationProvider = { key, table in
```
tbarbugli, line 18: why don't we do this by default?

docusaurus/docs/iOS/customization/localization.md
```diff
@@ -0,0 +1,138 @@
...
+    // For most cases, you will probably use Bundle.main, unless your strings are in different Bundle.
+    // Note: `table` is equivalent to filename of the `Localization.strings` file without file extension. 
+    Bundle.main.localizedString(forKey: key, value: nil, table: table)
+}
+```
+3. Now everything that has left is to implement your  `strings` and `stringsdict`  files for different languages or even customize the default strings for your own.
```
tbarbugli, line 24: - how should you name strings so that are linked to a specific language?
- how do you get the app to use a different language other than english?

tbarbugli (changes requested):
The docs are not very clear, see comments (easier if you look at React/React Native to get an idea about this)

Sources/StreamChatUI/Generated/L10n.swift
```diff
@@ -166,8 +166,8 @@ internal enum L10n {
 
 extension L10n {
   private static func tr(_ table: String, _ key: String, _ args: CVarArg...) -> String {
-    let format = BundleToken.bundle.localizedString(forKey: key, value: nil, table: table)
-    return String(format: format, locale: Locale.current, arguments: args)
+     let format = Appearance.default.localizationProvider(key, table)
```
b-onc, line 169: We should add a TODO here, this using `Appearance.default` will prohibit using Appearance injection

docusaurus/docs/iOS/customization/localization.md
DominikBucher12: Rephrased a bit. Maybe ask specific if something isn't clear.

docusaurus/docs/iOS/customization/localization.md
DominikBucher12: I pointed out to files in repo

b-onc (approved):
Code changes LGTM

docusaurus/docs/iOS/customization/localization.md
tbarbugli: what I mean is that by default we do this:

- Lookup string on Bundle.main.localizedString
- If nothing is found lookup the string from the SDK

this way adding translations to the app bundle will require no other change and you only need to touch `Appearance.default.localizationProvider` if you do something different/less common

b-onc: @tbarbugli that's simply not possible. SDK provides it's own bundle and app provides a totally different bundle. we can't reach app bundle from SDK

docusaurus/docs/iOS/customization/localization.md
```diff
@@ -0,0 +1,50 @@
+---
+title: Translation and i10n
+---
+
+The SDK offers you in-house localization for all strings contained in UI components. For our internal localization, we provide `strings` file for each language that we translate. For advanced plural localization cases (typing indicator showing how many people are typing or how many responses are in thread), we use `stringsdict` with `NSStringPluralRuleType`. 
+
+## The Change of language on the device.
```
nikomancy, line 7: Change to "Changing the device's language"

docusaurus/docs/iOS/customization/localization.md
```diff
@@ -0,0 +1,50 @@
...
+
+Language for the StreamChat is changed together language of the device or simulator. See [how to change device language](https://support.apple.com/en-us/HT204031).
```
nikomancy, line 9: Not sure how common / relevant this is, but what language does Stream Chat default to if the device is set to a language that we do not support?

nikomancy: > Language for the StreamChat is changed together language of the device or simulator.

Change to: 

Stream Chat tries to use the same language as the device it is running on. If the device is set to a language that Stream Chat does not support, then Stream Chat will default to `${behavior the library takes when the device uses an unsupported language}`.

docusaurus/docs/iOS/customization/localization.md
```diff
@@ -0,0 +1,50 @@
...
+There are however some workarounds how you can enforce the application to use only specific languages or let the users choose.
+
+## Adding new language which is not present in StreamChat
+
+StreamChatUI uses `strings` and `stringsdict` files to handle localize the framework.
+`Appearance` structure contains closure `localizationProvider`, which is responsible for deriving `localizedString` from exact bundle. 
+By default, this bundle is the  `StreamChatUI` framework bundle, but you can customize it to anything you need in 3 steps:
+
+1. If you don't have yet your `strings` or `stringsdict` files, you need to create those new files in your project/target, preferably `Localizable.strings` and `Localizable.stringsdict`.
```
nikomancy, line 18: Change to:

1. If you don't have `strings` or `stringsdict` files in your project, add those new files to `Localizable.strings` and `Localizable.stringsdict`.

docusaurus/docs/iOS/customization/localization.md
```diff
@@ -0,0 +1,50 @@
...
+2. When you have those files present, you need to [add new language to thole project](https://developer.apple.com/forums/thread/97252?answerId=295557022#295557022).
```
nikomancy, line 19: At a minimum we'll want to change this line to: 

2. Next, [add new language to the project](https://developer.apple.com/forums/thread/97252?answerId=295557022#295557022).

nikomancy: Also, it's definitely not a best practice to link to a forum post. If there's official Apple docs for this, we can link to those. If not, we should write out the steps here.

docusaurus/docs/iOS/customization/localization.md
```diff
@@ -0,0 +1,50 @@
...
+3. You need to do is to copy available `StreamChatUI` localization keys into your `strings` and `stringsdict` files.
```
nikomancy, line 20: Change to:

3. Copy the `StreamChatUI` localization keys into your `strings` and `stringsdict` files.

docusaurus/docs/iOS/customization/localization.md
```diff
@@ -0,0 +1,50 @@
...
+4. Once you have those files in your application, you need to set the `localizationProvider` to provide your `AppBundle` instead of `StreamChatUI` frameworks like this:
```
nikomancy, line 21: Change to:

Set the `localizationProvider` to provide your `AppBundle` instead of `StreamChatUI` frameworks:

docusaurus/docs/iOS/customization/localization.md
```diff
@@ -0,0 +1,50 @@
...
+`Appearance` structure contains closure `localizationProvider`, which is responsible for deriving `localizedString` from exact bundle. 
+By default, this bundle is the  `StreamChatUI` framework bundle, but you can customize it to anything you need in 3 steps:
+
+1. If you don't have yet your `strings` or `stringsdict` files, you need to create those new files in your project/target, preferably `Localizable.strings` and `Localizable.stringsdict`.
+2. When you have those files present, you need to [add new language to thole project](https://developer.apple.com/forums/thread/97252?answerId=295557022#295557022).
+3. You need to do is to copy available `StreamChatUI` localization keys into your `strings` and `stringsdict` files.
+4. Once you have those files in your application, you need to set the `localizationProvider` to provide your `AppBundle` instead of `StreamChatUI` frameworks like this: 
+```swift
+Appearance.default.localizationProvider = { key, table in
+    // For most cases, you will probably use Bundle.main, unless your strings are in different Bundle.
```
nikomancy, line 24: Change to:

// A best practice is to use `Bundle.main` unless your strings are in a different bundle.

docusaurus/docs/iOS/customization/localization.md
```diff
@@ -0,0 +1,50 @@
...
+    // Note: `table` is equivalent to filename of the `Localization.strings` file without file extension. 
+    // the name for `strings` and `stringsdict` should be same!
+    Bundle.main.localizedString(forKey: key, value: nil, table: table)
+}
+```
+5. Now everything that has left is to implement your  `strings` and `stringsdict`  files for different languages or even customize the default strings for your own.
```
nikomancy, line 30: Change to:

Now, you're ready to implement your `strings` and `stringsdict` files for different languages.

nikomancy: Removing the step from the preceding <ol> because this is more of a debrief on what to do after completing the previous steps.

docusaurus/docs/iOS/customization/localization.md
```diff
@@ -0,0 +1,50 @@
...
+
+This step is the same for overriding existing languages, because there is no user-friendly way to override just some keys from the `strings` file as the Localization is getting loaded as whole file.
```
nikomancy, line 32: Do we have the steps for overriding existing languages recorded in the docs? If yes, link to that page to make this easier for users to navigate.

docusaurus/docs/iOS/customization/localization.md
```diff
@@ -0,0 +1,50 @@
+---
+title: Translation and i10n
```
nikomancy, line 2: Change the title to: Adding Localization

docusaurus/docs/iOS/customization/localization.md
DominikBucher12: Changed.

docusaurus/docs/iOS/customization/localization.md
DominikBucher12: Changed.

docusaurus/docs/iOS/customization/localization.md
DominikBucher12: Changed.

docusaurus/docs/iOS/customization/localization.md
DominikBucher12: Thank you for this, yes this makes more sense :)

docusaurus/docs/iOS/customization/localization.md
DominikBucher12: Done.

docusaurus/docs/iOS/customization/localization.md
DominikBucher12: Yes thanks. Fixed by pointing out to official documentation.

docusaurus/docs/iOS/customization/localization.md
DominikBucher12: Done.

docusaurus/docs/iOS/customization/localization.md
DominikBucher12: Done.

docusaurus/docs/iOS/customization/localization.md
DominikBucher12: Thanks, changed. :)

docusaurus/docs/iOS/customization/localization.md
DominikBucher12: Nope, because it's the same, in the cocoa world, you cannot simply override to localization keys only without doing some swizzling system calls and hacks.

docusaurus/docs/iOS/customization/localization.md
```diff
@@ -0,0 +1,51 @@
...
+title: Adding Localization
+---
+
+The SDK offers you in-house localization for all strings contained in UI components. For our internal localization, we provide `strings` file for each language that we translate. For advanced plural localization cases (typing indicator showing how many people are typing or how many responses are in thread), we use `stringsdict` with `NSStringPluralRuleType`. 
+
+## Changing the device's language
+
+Stream Chat tries to use the same language as the device it is running on. If the device is set to a language that Stream Chat does not support, then Stream Chat will default to English.
+See [how to change device language](https://support.apple.com/en-us/HT204031).
+There are however some workarounds how you can enforce the application to use only specific languages or let the users choose.
```
nikomancy, line 11: If we're calling out workflows for forcing the app to only use a specific language, we should link to them or add them to this page.

DominikBucher12: According to internal discussion, deleted the sentence. Thank you :)

nikomancy:
The docs are good to go.

nikomancy (approved):
Approving this for docs.

PR #1338, opened 2021-08-03
[CIS-1025] Improve the way of customisation list ordering for message list

dmigach:
So now it's possible to pass a desired value before the controller is initiated so it's reflected right away. 
Because at the moment if you set this value, you need to reload your UI manually.

After this change desired behaviour can be achieved for instance by overriding `showMessageList` on the `ChatChannelListRouter`:

```
    open func showMessageList(for cid: ChannelId) {
        let vc = components.messageListVC.init()
        vc.channelController = rootViewController.controller.client.channelController(
            for: cid,
            // ** The new part **
            listOrdering: .bottomToTop
        )

        guard let navController = rootNavigationController else {
            log.error("Can't push chat detail, no navigation controller available")
            return
        }
        
        navController.show(vc, sender: self)
    }
```


b-onc (approved):
CHANGELOG please :)

tbarbugli (approved):

PR #1343, opened 2021-08-04
[CIS-1081] Expose models on Events

b-onc:
Missing:
- Changelog
- Tests

The way this works, we put a helper object to the event, and we put closures to helper object to fetch the data from DB. The closure is ran lazily when users access the fields (or never ran)

### DIFF
--- a/DemoApp/ChatPresenter.swift
+++ b/DemoApp/ChatPresenter.swift
@@ -176,22 +176,48 @@ class DemoChatChannelListRouter: ChatChannelListRouter {
     }
 }
 
-class DemoChannelListVC: ChatChannelListVC {
+class DemoChannelListVC: ChatChannelListVC, EventsControllerDelegate {
     /// The `UIButton` instance used for navigating to new channel screen creation,
     lazy var createChannelButton: UIButton = {
         let button = UIButton()
         button.setImage(UIImage(named: "pencil")!, for: .normal)
         return button
     }()
+    
+    var eventsController: EventsController!
 
     override func viewDidLoad() {
         super.viewDidLoad()
 
         navigationItem.rightBarButtonItem = UIBarButtonItem(customView: createChannelButton)
         createChannelButton.addTarget(self, action: #selector(didTapCreateNewChannel), for: .touchUpInside)
+        
+        eventsController = controller.client.eventsController()
+        eventsController.delegate = self
     }
 
     @objc open func didTapCreateNewChannel(_ sender: Any) {
         (router as! DemoChatChannelListRouter).showCreateNewChannelFlow()
+    }
+    
+    func eventsController(_ controller: EventsController, didReceiveEvent event: Event) {
+        switch event {
+        case let event as MessageNewEvent:
+            if let message = event.message(), let channel = event.channel() {
+                // Show local notification for the event
+                let content = UNMutableNotificationContent()
+                content.title = "\(message.author.name ?? message.author.id) @ \(channel.name ?? channel.cid.id)"
+                content.body = message.text
+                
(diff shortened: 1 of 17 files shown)


Sources/StreamChat/WebSocketClient/Events/Event.swift
```diff
@@ -2,16 +2,94 @@
...
+        (self as? EventWithPayload)?.savedData?.unreadCount
+    }
+    
+    public func message() -> ChatMessage? {
+        if let eventWithPayload = self as? EventWithPayload {
+            if let savedData = eventWithPayload.savedData {
+                if let message = savedData.message {
+                    return message
+                } else {
+                    print("### no message")
```
b-onc, line 45: I forgot these during debugging 😓

Sources/StreamChat/Workers/EventNotificationCenter.swift
```diff
@@ -34,6 +34,46 @@ class EventNotificationCenter: NotificationCenter {
...
+            )
+            event = channelSpecificEvent
+        case var currentUserEvent as CurrentUserEvent:
+            currentUserEvent.savedData = SavedEventData(
+                currentUser: { [weak database] in database?.viewContext.currentUser?.asModel() },
+                underlyingContext: database.viewContext
+            )
+            event = currentUserEvent
+        default:
+            break
```
tbarbugli, line 74: we need to handle custom events as well

Sources/StreamChatUI/ChatMessageList/ChatChannelHeaderView.swift
```diff
@@ -127,7 +127,7 @@ open class ChatChannelHeaderView:
 
     open func channelController(
         _ channelController: ChatChannelController,
-        didReceiveMemberEvent: MemberEvent
+        didReceiveMemberEvent: Event
```
tbarbugli, line 130: why this change?

DemoApp/ChatPresenter.swift
```diff
@@ -176,22 +176,48 @@ class DemoChatChannelListRouter: ChatChannelListRouter {
...
     }
 
     @objc open func didTapCreateNewChannel(_ sender: Any) {
         (router as! DemoChatChannelListRouter).showCreateNewChannelFlow()
     }
+    
+    func eventsController(_ controller: EventsController, didReceiveEvent event: Event) {
+        switch event {
+        case let event as MessageNewEvent:
+            if let message = event.message(), let channel = event.channel() {
```
dmigach, line 206: The whole approach totally makes sense to me, the only thing I'd maybe change is turn `message()`, `channel()` etc on `Event` into computed properties so when you use it the syntax is more natural, unless you think there is some value in explicitly showing that there is computation under the hood

Sources/StreamChatUI/ChatMessageList/ChatChannelHeaderView.swift
b-onc: `MemberEvent` needs to have `savedData`, which is defined in `EventWithPayload`. `EventWithPayload` is `internal`, so we either expose `EventWithPayload` (bad idea) or we change these params `MemberEvent` -> `Event` so we don't need to expose payloads. The good part is, this change is basically free for the user, since the user needs to cast `MemberEvent` to a specific event (`MemberAddedEvent` or `MemberRemovedEvent`) anyway

DemoApp/ChatPresenter.swift
b-onc: if those are computed properties instead of functions, we have to implement them for all conformers of `Event` type. when they're functions, we can have a default implementation and all conformers will get the default implementation

PR #1344, opened 2021-08-05
WIP - [CIS-652] Implement compound avatar for non DM channels

SagarSDagdu:
### What does this PR do

This PR implements the compound avatar feature for non-DM channels. 
When a channel consists of more than 1 member other than the user, we show an avatar that combines the other users' avatars.

Screenshot 2021-08-05 at 6 50 12 PM


Sources/StreamChatUI/CommonViews/AvatarView/ChatChannelAvatarView.swift
```diff
@@ -33,28 +33,204 @@ open class ChatChannelAvatarView: _View, ThemeProvider, SwiftUIRepresentable {
...
+            let combinedImage = getCombinedAvatarImage(from: images) ?? appearance.images.userAvatarPlaceholder2
+            
+            presenceAvatarView.avatarView.imageView.loadImage(
+                from: nil,
+                placeholder: combinedImage,
+                preferredSize: .avatarThumbnailSize,
+                components: components
+            )
+            return
+        }
```
nuno-vieira, line 116: It isn't a good sign we have `SystemEnvironment.isTests` everywhere. Ideally, we should replace the ´ImageLoader` implementation in the tests, with a mocked one. The problem here is that `loadImage` is an extension on `UIImageView`, but I would prefer that we have an `ImageLoader` component that is responsible to load all the images and handle all of that. This loader should be a dependency of each component that needs to load images. So it is nice that you introduced `ImageLoader`, but I would not make it a singleton, and would make it a dependency of each component that needs it so that it is easily testable + easier to change implementations. And so, I would also move the `UIImageView.loadImage` extension to `ImageLoader`.

Sources/StreamChatUI/CommonViews/AvatarView/ChatChannelAvatarView.swift
```diff
@@ -33,28 +33,204 @@ open class ChatChannelAvatarView: _View, ThemeProvider, SwiftUIRepresentable {
...
+                from: nil,
+                placeholder: appearance.images.userAvatarPlaceholder3,
+                preferredSize: .avatarThumbnailSize,
+                components: components
+            )
+            return
+        } else if urls.count > 4 {
+            // We show a combination of at max 4 images combined
+            urls = Array(urls[0...3])
+        }
```
nuno-vieira, line 101: It is probably a good idea to breakdown `loadAvatar()` into more functions and have something like:
- `loadAvatar()`:
     -`loadChannelAvatar()`
     -`loadDirectChannelAvatar()`
     -`loadMergedAvatars()`
     
So that if the customer only wants to change the logic of DM avatars, it doesn't need to override all the `loadAvatar` function 👍

Sources/StreamChatUI/CommonViews/AvatarView/ChatChannelAvatarView.swift
```diff
@@ -33,28 +33,204 @@ open class ChatChannelAvatarView: _View, ThemeProvider, SwiftUIRepresentable {
...
-            placeholder: appearance.images.userAvatarPlaceholder4,
-            preferredSize: .avatarThumbnailSize,
-            components: components
-        )
-        presenceAvatarView.isOnlineIndicatorVisible = isOnlineIndicatorVisible
+        }
+        
+        group.notify(queue: DispatchQueue.main) {
+            completion(avatars, channelId)
+        }
```
nuno-vieira, line 162: This should be part of `ImageLoader` component. To have something like `ImageLoader.loadAllImages(urls:, placeholders:, completion:)`.

Sources/StreamChatUI/CommonViews/AvatarView/ChatChannelAvatarView.swift
```diff
@@ -33,28 +33,204 @@ open class ChatChannelAvatarView: _View, ThemeProvider, SwiftUIRepresentable {
...
+            let rightImage = UIImageCombiner.combine(
+                images: [
+                    images[1],
+                    images[2]
+                ],
+                orientation: .vertical
+            )?
+                .scaled(to: .avatarThumbnailSize)
+                .crop(to: halfContainerSize) ?? appearance.images.userAvatarPlaceholder3.crop(to: halfContainerSize)
+            combinedImage = UIImageCombiner.combine(
```
nuno-vieira, line 196: Like mentioned before, we should not use singletons here, it is better that the `ImageLoader` and `ImageCombiner` are dependencies of `ChatChannelAvatarView` so that they can be easily changed.

Sources/StreamChatUI/CommonViews/AvatarView/ChatChannelAvatarView.swift
```diff
@@ -33,28 +33,204 @@ open class ChatChannelAvatarView: _View, ThemeProvider, SwiftUIRepresentable {
...
-        )
-        presenceAvatarView.isOnlineIndicatorVisible = isOnlineIndicatorVisible
+        }
+        
+        group.notify(queue: DispatchQueue.main) {
+            completion(avatars, channelId)
+        }
+    }
+    
+    private func getCombinedAvatarImage(from avatars: [UIImage]) -> UIImage? {
```
nuno-vieira, line 165: This method should also be `open`

nuno-vieira (changes requested):
Good first implementation 👍 Left some comments to make it more customizable and also to maybe improve problems that were already present.

Sources/StreamChatUI/CommonViews/AvatarView/ChatChannelAvatarView.swift
```diff
@@ -33,28 +33,204 @@ open class ChatChannelAvatarView: _View, ThemeProvider, SwiftUIRepresentable {
...
+                preferredSize: .avatarThumbnailSize,
+                components: components
+            )
+            return
+        }
+        
+        // If the channel is a direct message channel, load the other user's avatar
+        // and also set the online indicator visible
+        if channel.isDirectMessageChannel {
+            let otherMember = lastActiveMembers.first
```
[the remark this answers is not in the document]
SagarSDagdu: If you look above this statement, the filtering of current user is already applied when creating the `lastActiveMembers` array

Sources/StreamChatUI/Components.swift
```diff
@@ -72,6 +72,13 @@ public struct Components {
     
     /// Object with set of function for handling images from CDN
     public var imageCDN: ImageCDN = StreamImageCDN()
+    
+    /// Object responsible for loading images from a given URL
+    public var imageLoader: ImageLoading = DefaultImageLoader()
```
nuno-vieira, line 77:
```suggestion
    /// Object responsible for loading images.
    public var imageLoader: ImageLoading = DefaultImageLoader()
```

Sources/StreamChatUI/Utils/ImageLoader.swift
```diff
@@ -0,0 +1,115 @@
...
+    ///   - url: The URL of the image
+    ///   - imageCDN: The CDN to be used for downloading the image
+    ///   - resize: Whether the image should be resized
+    ///   - preferredSize: The preferred size of the image to be downloaded
+    ///   - completion: Completion that gets called when the download is finished
+    @discardableResult
+    func loadImage(
+        from url: URL,
+        imageCDN: ImageCDN,
+        resize: Bool,
```
nuno-vieira, line 21: Let's add `ImageCDN` as a dependency of `DefaultImageLoader` and make the `Components.imageLoader` lazy var

Sources/StreamChatUI/Utils/UIImage+Extensions.swift
```diff
@@ -37,3 +38,41 @@ extension UIImage {
         return photoURL
     }
 }
+
+// MARK: - UIImage extension related to size
+
+extension UIImage {
+    func scaled(to targetSize: CGSize) -> UIImage {
+        // Determine the scale factor that preserves aspect ratio
```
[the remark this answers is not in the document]
SagarSDagdu: Nuke's resize function doesn't work for this, I checked it

Sources/StreamChatUI/CommonViews/AvatarView/ChatChannelAvatarView.swift
SagarSDagdu: As discussed, loading images for an array of URLs doesn't seem to be a responsibility of the loader, and it also creates a problem with the parameters even if we try to do so, hence skipping this and keeping the method in the avatarview itself

Sources/StreamChatUI/CommonViews/AvatarView/ChatChannelAvatarView.swift
SagarSDagdu: My thought process was, the customer would have just a single point of customization i.e overriding 
`open func loadAvatar(for channel: ChatChannel)`
if they want a custom implementation. Creating different methods for different cases would be a pain in terms of understanding that which is called in which conditions IMO
In the single func, they can easily make decisions based on the `channel` instance we are passing. WDYT?

nuno-vieira: I think that pain will be for the customer needing to override and re-implement all of the `loadAvatar()` function from scratch if he is only interested in changing one little thing on for example Direct messages. We have this pattern in other parts of the code, and I think for now it is for the best. Unless we remove the early returns, then the customer can still duplicate some logic and override only for one scenario. But I think the right way to do it is to split it up into different responsibilities since it makes it easier to read, and easier to customize.

Sources/StreamChatUI/Utils/ImageMerger.swift
```diff
@@ -0,0 +1,91 @@
...
+        images: [UIImage],
+        orientation: ImageMergeOrientation
+    ) -> UIImage? {
+        orientation == .horizontal ? combineSideToSide(images: images) : combineTopToBottom(images: images)
+    }
+    
+    /// Combines images in top to bottom fashion
+    /// - Parameter images: The images
+    /// - Returns: The combined image
+    open func combineTopToBottom(images: [UIImage]) -> UIImage? {
```
nuno-vieira, line 46:
```suggestion
    open func mergeTopToBottom(images: [UIImage]) -> UIImage? {
```

Sources/StreamChatUI/Utils/ImageMerger.swift
```diff
@@ -0,0 +1,91 @@
...
+        let finalImage = UIGraphicsGetImageFromCurrentImageContext()
+        UIGraphicsEndImageContext()
+
+        return finalImage
+    }
+    
+    /// Combines images in top to side to side order (left -> right)
+    /// - Parameter images: The images
+    /// - Returns: The combined image
+    open func combineSideToSide(images: [UIImage]) -> UIImage? {
```
nuno-vieira, line 70:
```suggestion
    open func mergeSideToSide(images: [UIImage]) -> UIImage? {
```

Sources/StreamChatUI/CommonViews/AvatarView/ChatChannelAvatarView.swift
SagarSDagdu: Got your point 👍🏻  implemented separated methods accordingly

Sources/StreamChatUI/Utils/ImageLoader.swift
```diff
@@ -0,0 +1,74 @@
...
+import Nuke
+import UIKit
+
+/// ImageLoading is providing set of functions for downloading of images from URLs.
+public protocol ImageLoading: AnyObject {
+    /// Load an image from the given URL
+    /// - Parameters:
+    ///   - url: The URL of the image
+    ///   - imageCDN: The CDN to be used for downloading the image
+    ///   - resize: Whether the image should be resized
```
nuno-vieira, line 14:
```suggestion
    ///   - url: The URL of the image
    ///   - resize: Whether the image should be resized
```

Sources/StreamChatUI/Utils/ImageLoader.swift
```diff
@@ -0,0 +1,74 @@
...
+    ///   - preferredSize: The preferred size of the image to be downloaded
+    ///   - completion: Completion that gets called when the download is finished
+    @discardableResult
+    func loadImage(
+        from url: URL,
+        resize: Bool,
+        preferredSize: CGSize?,
+        completion: @escaping ((_ result: Result<UIImage, Error>) -> Void)
+    ) -> ImageTask?
```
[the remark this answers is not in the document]
nuno-vieira: To replace `ImageTask` with our own implementation. I guess for now a `protocol Cancellable { func cancel() }` is enough. and then we make `ImageTask: Cancellable`

Sources/StreamChatUI/Utils/ImageLoader.swift
nuno-vieira: Please remove the "imageCDN" comment since it doesn't exist anymore

Sources/StreamChatUI/Utils/ImageLoader.swift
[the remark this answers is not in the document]
SagarSDagdu: Excellent advice, I've removed the dependency. Thanks!

nuno-vieira (approved):
LGTM 🎉

Sources/StreamChatUI/Utils/UIImage+Extensions.swift
```diff
@@ -37,3 +38,41 @@ extension UIImage {
         return photoURL
     }
 }
+
+// MARK: - UIImage extension related to size
+
+extension UIImage {
```
tbarbugli, line 44: IMO we should not do this an extension of UIImage but instead expose some public functions/methods that you can use in your application.

use-case: I want to change user avatar, ideally I can copy paste our component and re-use our utilities

Sources/StreamChatUI/Utils/ImageLoading/DefaultImageLoader.swift
```diff
@@ -0,0 +1,58 @@
...
+//
+
+import Nuke
+import UIKit
+
+extension ImageTask: Cancellable {}
+
+/// The class which is resposible for loading images from URLs.
+/// Internally uses `Nuke`'s shared object of `ImagePipeline` to load the image.
+open class DefaultImageLoader: ImageLoading {
```
tbarbugli, line 12: I rather call `DefaultImageLoader` `NukeImageLoader`

Sources/StreamChatUI/Utils/ImageLoading/DefaultImageLoader.swift
```diff
@@ -0,0 +1,58 @@
...
+    /// The CDN to be used
+    open var imageCDN: ImageCDN
+    
+    public init(imageCDN: ImageCDN) {
```
tbarbugli, line 16: weird coupling here

tbarbugli (commented):
I am not 100% about how CDN and ImageLoading are currently coupled

ie. say that you use our CDN and want to render an image coming from a different domain, in this case `imageCDN.thumbnailURL` does not really make lot of sense

Sources/StreamChatUI/Utils/UIImage+Extensions.swift
nuno-vieira: Yes that was the plan, just like mentioned here: https://github.com/GetStream/stream-chat-swift/pull/1344#discussion_r685411028

According to @SagarSDagdu this a big refactor and big effort to change that.

Sources/StreamChatUI/Utils/ImageLoading/DefaultImageLoader.swift
nuno-vieira: Agree 👍

Sources/StreamChatUI/Utils/ImageLoading/DefaultImageLoader.swift
SagarSDagdu: Agreed, image loading shouldn't have any knowledge about CDN. Changed the code to remove this coupling

Sources/StreamChatUI/Utils/UIImage+Extensions.swift
SagarSDagdu: Agreed. I've introduced `StreamImageProcessor` and got rid of the extension.

Sources/StreamChatUI/Utils/UIImage+Extensions.swift
```diff
@@ -2,6 +2,7 @@
 // Copyright © 2021 Stream.io Inc. All rights reserved.
 //
 
+import Nuke
```
nuno-vieira, line 5:
```suggestion

```

Sources/StreamChatUI/Utils/StreamImageProcessor.swift
```diff
@@ -0,0 +1,52 @@
+//
+// Copyright © 2021 Stream.io Inc. All rights reserved.
+//
+
+import Nuke
+import UIKit
+
+/// This class provides resizing operations for `UIImage`
+open class StreamImageProcessor {
```
nuno-vieira, line 9: I think we should have the same pattern as the other image components:
- ImageProcessing
- NukeImageProcessing (Not StreamImageProcessing, since we are using Nuke here)

Sources/StreamChatUI/Utils/ImageLoading/NukeImageLoader.swift
```diff
@@ -0,0 +1,42 @@
...
+
+extension ImageTask: Cancellable {}
+
+/// The class which is resposible for loading images from URLs.
+/// Internally uses `Nuke`'s shared object of `ImagePipeline` to load the image.
+open class NukeImageLoader: ImageLoading {
+    @discardableResult
+    open func loadImage(
+        using urlRequest: URLRequest,
+        cachingKey: String,
```
nuno-vieira, line 16:
```suggestion
        cachingKey: String?,
```
cachingKey should be optional, but even better would be to have an `ImageLoadingOptions` just like `ImageRequestOptions`. Otherwise, whenever we add more options, the arguments of the func will increase a lot

nuno-vieira (commented):
Just some more comments, but after that it should be LGTM 👍

Sources/StreamChatUI/Utils/ImageLoading/NukeImageLoader.swift
SagarSDagdu: Good point. I have made the `cachingKey` optional for now. As it is a single option currently, not introducing `ImageLoadingOptions`. We can introduce this later if needed

Sources/StreamChatUI/Utils/StreamImageProcessor.swift
SagarSDagdu: As the functionalities provided by this class are very generic(i.e. cropping, scaling), didn't feel the need to create a protocol for this. However, I have renamed the class to `NukeImageProcessor` to better suit other namings. Let me know your thoughts about not having a protocol.

Sources/StreamChatUI/Utils/ImageLoading/NukeImageLoader.swift
nuno-vieira: Introducing it later will cause a breaking change. And by introducing it now, we will avoid breaking the API in the future, since the `ImageLoadingOptions` should be a struct with all of the props with default arguments. Currently `cachingKey` is optional, but doesn't have a default value. So ideally we want `struct ImageLoadingOptions` and with all props with default arguments, in this case: `ImageLoadingOptions(cachingKey: String? = nil)`

Sources/StreamChatUI/Utils/StreamImageProcessor.swift
nuno-vieira: I think for consistency and also for testability it is better we still introduce the protocol here 👍

Sources/StreamChatUI/CommonViews/AvatarView/ChatChannelAvatarView.swift
kojiba: yes, agree this completely breaks SOLID and makes code tight-coupled

Sources/StreamChatUI/CommonViews/AvatarView/ChatChannelAvatarView.swift
```diff
@@ -16,6 +16,23 @@ open class ChatChannelAvatarView: _View, ThemeProvider, SwiftUIRepresentable {
     open var content: (channel: ChatChannel?, currentUserId: UserId?) {
         didSet { updateContentIfNeeded() }
     }
+    
+    /// Object responsible for loading images.
+    open lazy var imageLoader: ImageLoading = {
```
kojiba, line 21: is there any reason to make all properties lazy? 
can we simplify this code to: 

```Swift
open var imageLoader: ImageLoading = NukeImageLoader()
open var imageMerger: ImageMerging = DefaultImageMerger()
open var imageProcessor: NukeImageProcessor = NukeImageProcessor()
```
I would also assume that we want to do the protocols injections for all dependencies for making FIRST tests, but this can be TBD

Sources/StreamChatUI/CommonViews/AvatarView/ChatChannelAvatarView.swift
```diff
@@ -30,31 +47,293 @@ open class ChatChannelAvatarView: _View, ThemeProvider, SwiftUIRepresentable {
...
-                .sorted { $0.memberCreatedAt < $1.memberCreatedAt }
-                .first(where: { $0.id != content.currentUserId })
-
-            return (firstOtherMember?.imageURL, firstOtherMember?.isOnline ?? false)
-        }()
-
+        loadAvatar(for: channel)
+    }
+    
+    open func loadAvatar(for channel: ChatChannel) {
```
kojiba, line 57: question just to double-check, we don't show any loading state, we showing some hardcoded image instead?

Sources/StreamChatUI/CommonViews/AvatarView/ChatChannelAvatarView.swift
```diff
@@ -30,31 +47,293 @@ open class ChatChannelAvatarView: _View, ThemeProvider, SwiftUIRepresentable {
...
+        // If the channel has an avatar set, load that avatar
```
kojiba, line 58: // If the channel has an avatar set, load that avatar
// Use the appropriate method to load avatar based on channel type

The comments are redundant on my opinion, this code is good and readable, so we can remove them but this is up to you.

Sources/StreamChatUI/CommonViews/AvatarView/ChatChannelAvatarView.swift
```diff
@@ -30,31 +47,293 @@ open class ChatChannelAvatarView: _View, ThemeProvider, SwiftUIRepresentable {
...
             preferredSize: .avatarThumbnailSize,
             components: components
         )
+    }
+    
+    /// Loads avatar for a directMessageChannel
+    /// - Parameter channel: The channel
+    open func loadDirectMessageChannelAvatar(channel: ChatChannel) {
+        let lastActiveMembers = channel.lastActiveMembers
+            .sorted { $0.memberCreatedAt < $1.memberCreatedAt }
```
kojiba, line 87: In case when we will have 100500 participants this one can impact on execution speed. 
The complexity of sorting is big, means time required will also be big enough (O(n log n)). 
I see couple of options here:
1. make all counting in background
2. add field to data entity with `lastActiveMemebers` with array of 3-4 participants, but this will possible require backend and internal DB changes

@tbarbugli can you advice here please?

Sources/StreamChatUI/CommonViews/AvatarView/ChatChannelAvatarView.swift
```diff
@@ -30,31 +47,293 @@ open class ChatChannelAvatarView: _View, ThemeProvider, SwiftUIRepresentable {
...
             placeholder: appearance.images.userAvatarPlaceholder4,
             preferredSize: .avatarThumbnailSize,
             components: components
         )
+    }
+    
+    /// Loads avatar for a directMessageChannel
+    /// - Parameter channel: The channel
+    open func loadDirectMessageChannelAvatar(channel: ChatChannel) {
+        let lastActiveMembers = channel.lastActiveMembers
```
kojiba, line 86: question to double-check: maximum amount of lastActiveMembers 
(if its up to 10 its fine, if its 100+ need improve this code)

sorting of large array can impact on performance, 
if needed please add the counting in background here

Sources/StreamChatUI/CommonViews/AvatarView/ChatChannelAvatarView.swift
```diff
@@ -30,31 +47,293 @@ open class ChatChannelAvatarView: _View, ThemeProvider, SwiftUIRepresentable {
...
         presenceAvatarView.isOnlineIndicatorVisible = isOnlineIndicatorVisible
     }
+    
+    /// Loads an avatar which is merged (tiled) version of the first four active members of the channel
+    /// - Parameter channel: The channel
+    open func loadMergedAvatars(channel: ChatChannel) {
+        // The channel is a non-DM channel, hide the online indicator
+        presenceAvatarView.isOnlineIndicatorVisible = false
+        
+        let lastActiveMembers = channel.lastActiveMembers
```
kojiba, line 120: this line is the copypase of line 86, please make a separate function for it and DRY it out

https://uk.wikipedia.org/wiki/Don%27t_repeat_yourself

Sources/StreamChatUI/CommonViews/AvatarView/ChatChannelAvatarView.swift
```diff
@@ -30,31 +47,293 @@ open class ChatChannelAvatarView: _View, ThemeProvider, SwiftUIRepresentable {
...
+            .sorted { $0.memberCreatedAt < $1.memberCreatedAt }
+            .filter { $0.id != content.currentUserId }
+        
+        // If there are no members other than the current user in the channel, load a placeholder
+        if lastActiveMembers.isEmpty {
+            presenceAvatarView.avatarView.imageView.loadImage(
```
kojiba, line 126: this piece of code also can be made an private extension for ImageView

```
   presenceAvatarView.avatarView.imageView.loadImage(
                from: nil,
                placeholder: appearance.images.userAvatarPlaceholder4,
                preferredSize: .avatarThumbnailSize,
                components: components
            )
```

There is a 3 places where only name of placeholder changes,  .avatarThumbnailSize and other parameters stay the same, so this can be DRY-out

Sources/StreamChatUI/CommonViews/AvatarView/ChatChannelAvatarView.swift
```diff
@@ -30,31 +47,293 @@ open class ChatChannelAvatarView: _View, ThemeProvider, SwiftUIRepresentable {
...
+            )
+            return
+        }
+        
+        // We show a combination of at max 4 images combined
+        urls = Array(urls.prefix(4))
+        
+        guard !SystemEnvironment.isTests else {
+            // When running tests, we load the images synchronously
+            let images = urls.map { UIImage(data: try! Data(contentsOf: $0!))! }
```
kojiba, line 152: force unwraps should be removed please use `try?` and guard-if-let chain instead of force unwrapping

Sources/StreamChatUI/CommonViews/AvatarView/ChatChannelAvatarView.swift
```diff
@@ -30,31 +47,293 @@ open class ChatChannelAvatarView: _View, ThemeProvider, SwiftUIRepresentable {
...
+                preferredSize: .avatarThumbnailSize,
+                components: components
+            )
+            return
+        }
+        
+        // We show a combination of at max 4 images combined
+        urls = Array(urls.prefix(4))
+        
+        guard !SystemEnvironment.isTests else {
```
kojiba, line 150: test code should never be part of production code, this breaks SOLID, makes code tightly coupled and unreliable tests, 
if needed lets setup a call and discuss approach how to do that correctly

Sources/StreamChatUI/CommonViews/AvatarView/ChatChannelAvatarView.swift
```diff
@@ -30,31 +47,293 @@ open class ChatChannelAvatarView: _View, ThemeProvider, SwiftUIRepresentable {
...
+        var placeholderImages = [
+            appearance.images.userAvatarPlaceholder1,
+            appearance.images.userAvatarPlaceholder2,
+            appearance.images.userAvatarPlaceholder3,
+            appearance.images.userAvatarPlaceholder4
+        ]
+        
+        var avatarUrls: [URL] = []
+        
+        for url in urls.prefix(4) {
```
kojiba, line 201: .prefix(4) - magic number

can be moved to private enum
```Swift
private enum Configuration {
   static let httpPrefixSymbolsCount = 4
}
```

and use as `Configuration.httpPrefixSymbolsCount`

Sources/StreamChatUI/CommonViews/AvatarView/ChatChannelAvatarView.swift
```diff
@@ -30,31 +47,293 @@ open class ChatChannelAvatarView: _View, ThemeProvider, SwiftUIRepresentable {
...
+            group.notify(queue: .main) {
+                completion(images, channelId)
+            }
+        }
+    }
+    
+    /// Creates a merged avatar from the given images
+    /// - Parameter avatars: The individual avatars
+    /// - Returns: The merged avatar
+    open func createMergedAvatar(from avatars: [UIImage]) -> UIImage? {
```
kojiba, line 246: for me this name is more like `createMergedAvatarOrPlaceholder`, but please think on a better naming

cause for caller its unobvious that in case of !avatars.isEmpty it will return placeholder

Sources/StreamChatUI/Utils/ImageLoading/ImageLoading.swift
```diff
@@ -0,0 +1,25 @@
+//
+// Copyright © 2021 Stream.io Inc. All rights reserved.
+//
+
+import UIKit
+
+/// Cancellable provides a set of functions that enable cancelling a task
+public protocol Cancellable {
```
kojiba, line 8: should be moved in separate file `Cancellable.swift`

kojiba: cause its very generic, and can be used out of ImageLoading proto

Sources/StreamChatUI/Utils/ImageLoading/NukeImageLoader.swift
```diff
@@ -0,0 +1,42 @@
+//
+// Copyright © 2021 Stream.io Inc. All rights reserved.
+//
+
+import Nuke
+import UIKit
+
+extension ImageTask: Cancellable {}
```
kojiba, line 8: this is internal by default, please check do we need to do this `public extension`?

Sources/StreamChatUI/Utils/ImageLoading/NukeImageLoader.swift
```diff
@@ -0,0 +1,42 @@
...
+/// The class which is resposible for loading images from URLs.
+/// Internally uses `Nuke`'s shared object of `ImagePipeline` to load the image.
+open class NukeImageLoader: ImageLoading {
+    @discardableResult
+    open func loadImage(
+        using urlRequest: URLRequest,
+        cachingKey: String?,
+        completion: @escaping ((Result<UIImage, Error>) -> Void)
+    ) -> Cancellable? {
+        guard !SystemEnvironment.isTests else {
```
kojiba, line 19: test code should be removed from production code

Sources/StreamChatUI/Utils/ImageLoading/NukeImageLoader.swift
```diff
@@ -0,0 +1,42 @@
...
+            // When running tests, we load the images synchronously
+            let image = UIImage(data: try! Data(contentsOf: urlRequest.url!))!
```
kojiba, line 21: force unwraps should exchanged with if-let/guard/try?/or do-catch

Sources/StreamChatUI/Utils/ImageLoading/NukeImageLoader.swift
```diff
@@ -0,0 +1,42 @@
...
+            completion(.success(image))
+            return nil
+        }
+
+        let request = ImageRequest(
+            urlRequest: urlRequest,
+            options: ImageRequestOptions(filteredURL: cachingKey)
+        )
+        
+        let imageTask = ImagePipeline.shared.loadImage(with: request) { result in
```
kojiba, line 31: `ImagePipeline.shared.loadImage(with: request)`
this is tightly coupling code, breaking D from SOLID, and makes test unreliable
should be exchanged with dependency injection of protocol usage

Sources/StreamChatUI/Utils/ImageMerger.swift
```diff
@@ -0,0 +1,91 @@
...
+// Copyright © 2021 Stream.io Inc. All rights reserved.
+//
+
+import UIKit
+
+/// The orientation to be used for merge the images
+public enum ImageMergeOrientation {
+    /// Merge the given images in horizontal orientation.
+    /// The width of the resulting images will be the addition of the widths of all the images,
+    /// whereas the height of the resulting image will be equal to the max of heights in the images
```
kojiba, line 11: this are the good comments example :)

Sources/StreamChatUI/Utils/ImageMerger.swift
```diff
@@ -0,0 +1,91 @@
...
+    ///   - orientation: The orientation to be used for combining the images
+    /// - Returns: A combined image
+    func merge(
+        images: [UIImage],
+        orientation: ImageMergeOrientation
+    ) -> UIImage?
+}
+
+open class DefaultImageMerger: ImageMerging {
+    // Initializer required for subclasses
```
kojiba, line 33: we can double-mark this with `Note`: or `NOTE:`

@nuno-vieira your opinion here?

Sources/StreamChatUI/Utils/ImageMerger.swift
```diff
@@ -0,0 +1,91 @@
...
+    open func mergeTopToBottom(images: [UIImage]) -> UIImage? {
+        var dimensions = CGSize(width: 0.0, height: 0.0)
+        for image in images {
+            dimensions.width = max(dimensions.width, image.size.width)
+            dimensions.height += max(dimensions.height, image.size.height)
+        }
+
+        UIGraphicsBeginImageContext(dimensions)
+
+        var lastY = CGFloat(0.0)
```
kojiba, line 55: should be just `var lastY: CGFloat = 0`

Sources/StreamChatUI/Utils/ImageMerger.swift
```diff
@@ -0,0 +1,91 @@
...
+        var dimensions = CGSize.zero
+        
+        for image in images {
+            dimensions.width += image.size.width
+            dimensions.height = max(dimensions.height, image.size.height)
+        }
+
+        UIGraphicsBeginImageContext(dimensions)
+
+        var lastX = CGFloat(0.0)
```
kojiba, line 80: should be just `var lastY: CGFloat = 0

Sources/StreamChatUI/Utils/ImageMerger.swift
```diff
@@ -0,0 +1,91 @@
...
+    }
+    
+    /// Merges images in top to bottom fashion
+    /// - Parameter images: The images
+    /// - Returns: The merged image
+    open func mergeTopToBottom(images: [UIImage]) -> UIImage? {
+        var dimensions = CGSize(width: 0.0, height: 0.0)
+        for image in images {
+            dimensions.width = max(dimensions.width, image.size.width)
+            dimensions.height += max(dimensions.height, image.size.height)
```
kojiba, line 50: why we dont just making `dimensions.height +=  image.size.height`?

Sources/StreamChatUI/Utils/NukeImageProcessor.swift
```diff
@@ -0,0 +1,52 @@
...
+
+/// This class provides resizing operations for `UIImage`. It internally uses `Nuke` porcessors to implement operations on images.
+open class NukeImageProcessor {
+    /// Crop the image to a given size. The image is center-cropped
+    /// - Parameters:
+    ///   - image: The image to crop
+    ///   - size: The size to which the image needs to be cropped
+    /// - Returns: The cropped image
+    open func crop(image: UIImage, to size: CGSize) -> UIImage? {
+        let imageProccessor = ImageProcessors.Resize(size: size, crop: true)
```
kojiba, line 16: `ImageProcessors.Resize(size: size, crop: true)`
should be dependency-injected with protocol usage

Sources/StreamChatUI/Utils/NukeImageProcessor.swift
```diff
@@ -0,0 +1,52 @@
...
+        let scaleFactor = min(widthRatio, heightRatio)
+        
+        // Compute the new image size that preserves aspect ratio
+        let scaledImageSize = CGSize(
+            width: image.size.width * scaleFactor,
+            height: image.size.height * scaleFactor
+        )
+        
+        // Draw and return the resized UIImage
+        let renderer = UIGraphicsImageRenderer(
```
kojiba, line 39: can be oneliner

kojiba (commented):
Couple of questions/items to be addressed

Sources/StreamChatUI/CommonViews/AvatarView/ChatChannelAvatarView.swift
nuno-vieira: What makes you think sorting an array with 100+ is going to cause performance issues? Especially in a language performant like swift? Maybe 100k+ yes, not 100 😅

kojiba: We displaying avatars in table view, i.e. each cell want to perform this operation, this can cause UI freezes on older devices, while we should always to be focused on reaching 60-120 FPS of rendering. 
Should be tested, but to be secured we can just move this to background.

SagarSDagdu: According to current configuration in`ChatClientConfig.lastActiveMembersLimit`, the value is just `5`. This can of course be changed by the client, but I do not think so it will go above 100, hence I ignored the sorting cost

Sources/StreamChatUI/CommonViews/AvatarView/ChatChannelAvatarView.swift
SagarSDagdu: This code was inside Tests code, i.e. `SystemEnvironment.isTests` hence the force unwrapping. Let's decide on the placement of the test code then.. because Snapshot tests need the images to be loaded synchronously in order to capture a snapshot

Sources/StreamChatUI/CommonViews/AvatarView/ChatChannelAvatarView.swift
nuno-vieira: Even for each cell, this will perform in nanoseconds, don't think it will ever block the main thread ever

Sources/StreamChatUI/Utils/ImageMerger.swift
SagarSDagdu: Nice catch. Will change this

Sources/StreamChatUI/Utils/ImageLoading/ImageLoading.swift
SagarSDagdu: Makes sense, moved it to a separate file

Sources/StreamChatUI/Utils/ImageLoading/NukeImageLoader.swift
SagarSDagdu: We do not need this to be public

Sources/StreamChatUI/Utils/NukeImageProcessor.swift
SagarSDagdu: Done

Sources/StreamChatUI/CommonViews/AvatarView/ChatChannelAvatarView.swift
```diff
@@ -16,6 +16,23 @@ open class ChatChannelAvatarView: _View, ThemeProvider, SwiftUIRepresentable {
     open var content: (channel: ChatChannel?, currentUserId: UserId?) {
         didSet { updateContentIfNeeded() }
     }
+    
+    /// Object responsible for loading images.
+    open var imageLoader: ImageLoading {
```
kojiba, line 21: jfyi: with current syntax this will be get-only variable instead of being mutating one

SagarSDagdu: The client can override this property to provide a different implementation if needed, hence the syntax.

Sources/StreamChatUI/CommonViews/AvatarView/ChatChannelAvatarView.swift
selanthiraiyan: 1. I hope the number `4` here represents the maximum number of images that we merge to create a new avatar. I don't understand why it should be named as `httpPrefixSymbolsCount`.

1. Wouldn't a private let be enough? 

```swift
private let maxNoOfImagesToBeMergedToGenerateAvatar = 4
```
Thanks!

kojiba: 1. thats why its called 'magic number', cause for reader is very hard to guess why this constant is placed this place :)
2. yes, private let or static let inside private enum is fine, we could keep it without enum as long its only one variable

selanthiraiyan: Sure, I understand the magic number part. I was to trying to understand why you named it as `httpPrefixSymbolsCount`.

SagarSDagdu: ✅

Sources/StreamChatUI/CommonViews/AvatarView/ChatChannelAvatarView.swift
tbarbugli: we should expose a configuration for `ImageLoading

Sources/StreamChatUI/Components.swift
```diff
@@ -73,6 +73,10 @@ public struct Components {
     /// Object with set of function for handling images from CDN
     public var imageCDN: ImageCDN = StreamImageCDN()
 
+    /// Object responsible for providing functionality of merging images.
+    /// Used when creating compound avatars from channel members individual avatars
+    public var imageMerger: ImageMerging = DefaultImageMerger()
```
tbarbugli, line 78: lets not expose this here

Sources/StreamChatUI/CommonViews/AvatarView/ChatChannelAvatarView.swift
tbarbugli: this is annoying if you want to make your own avatar view and dont have your own image loader

Sources/StreamChatUI/CommonViews/AvatarView/ChatChannelAvatarView.swift
```diff
@@ -30,31 +50,296 @@ open class ChatChannelAvatarView: _View, ThemeProvider, SwiftUIRepresentable {
...
         )
+    }
+    
+    /// Loads avatar for a directMessageChannel
+    /// - Parameter channel: The channel
+    open func loadDirectMessageChannelAvatar(channel: ChatChannel) {
+        let lastActiveMembers = lastActiveMembers()
+        
+        // If there are no members other than the current user in the channel, load a placeholder
+        if lastActiveMembers.isEmpty {
```
tbarbugli, line 92: AFAICT using guard let is more idiomatic here

```swift
if guard let otherMember = lastActiveMembers.first else {
    return
}
```

tbarbugli: it also simplifies a bit the rest of the code

Sources/StreamChatUI/ChatMessageList/Attachments/ChatMessageImageGallery+ImagePreview.swift
```diff
@@ -87,15 +87,15 @@ extension ChatMessageGalleryView {
...
-                .loadImage(
-                    from: attachment?.payload.imagePreviewURL,
-                    resize: false,
-                    components: components
-                ) { [weak self] _ in
+            imageTask = components.imageLoader.loadImage(
+                into: imageView,
+                url: attachment?.payload.imagePreviewURL,
+                imageCDN: components.imageCDN,
+                completion: { [weak self] _ in
```
evsaev, line 94: Hm, is it guaranteed the completion is called on main thread? 🤔

SagarSDagdu: Yes, Nuke guarantees that the completion gets called on the main thread. In case the customer provides own loader, they will need to ensure this

Sources/StreamChatUI/ChatMessageList/Attachments/ChatMessageLinkPreviewView.swift
```diff
@@ -116,7 +116,11 @@ open class ChatMessageLinkPreviewView: _Control, ThemeProvider {
 
         authorLabel.textColor = tintColor
 
-        imagePreview.loadImage(from: payload?.previewURL, components: components)
+        components.imageLoader.loadImage(
+            into: imagePreview,
+            url: payload?.previewURL,
+            imageCDN: components.imageCDN
+        )
```
evsaev, line 123: Hm, we loose resizing here, don't we?
```suggestion
        components.imageLoader.loadImage(
            into: imagePreview,
            url: payload?.previewURL,
            imageCDN: components.imageCDN,
            resize: true
        )
```

Sources/StreamChatUI/CommonViews/Attachments/AttachmentViews/ImageAttachmentComposerPreview.swift
```diff
@@ -41,7 +41,10 @@ open class ImageAttachmentComposerPreview: _View, ThemeProvider {
     
     override open func updateContent() {
         super.updateContent()
-        
-        imageView.loadImage(from: content, components: components)
+        components.imageLoader.loadImage(
```
evsaev, line 44: it seems we have the same issue here https://github.com/GetStream/stream-chat-swift/pull/1344/files#r694799931

Sources/StreamChatUI/ChatMessageList/Attachments/ChatMessageLinkPreviewView.swift
SagarSDagdu: Changed the default behavior to the implementation we had previously 👍

Sources/StreamChatUI/CommonViews/AvatarView/ChatChannelAvatarView.swift
SagarSDagdu: Removed the tests check, and also removed the `UIImageView` extension.

evsaev (approved):
LGTM! 🚢

PR #1347, opened 2021-08-09
[CIS-1087] Fix data races in message list's message count

dmigach:
My idea was the following:
From the reports on the issue it seemed like the issue is in messages count going out of sync with view's state, so I though of caching message's count and only updating it when we need it

And then I noticed that for channel list there actually already was such a mechanism and it seems like channel list does not experience this kind of issues

The only difference of the suggested fix from the one on the channel list is that message list's structure is a little bit more complicated because it has a VC and a view, but essentially it's the same

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### 🔄 Changed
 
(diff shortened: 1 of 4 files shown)


nuno-vieira (approved):
LGTM!

nuno-vieira (approved):
LGTM

PR #1348, opened 2021-08-09
[CIS-1091] Introduce `ChannelListSortingKey`s

b-onc:
Also, some functions in `ChatMessageReactionsVC` needed to be `open`

### DIFF
--- a/Sources/StreamChat/Query/Sorting/ChannelListSortingKey.swift
+++ b/Sources/StreamChat/Query/Sorting/ChannelListSortingKey.swift
@@ -19,6 +19,11 @@ public enum ChannelListSortingKey: String, SortingKey {
     /// Sort channels by `cid`.
     /// **Note**: This sorting option can extend your response waiting time if used as primary one.
     case cid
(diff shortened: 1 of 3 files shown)


Sources/StreamChat/Query/Sorting/ChannelListSortingKey.swift
```diff
@@ -19,6 +19,11 @@ public enum ChannelListSortingKey: String, SortingKey {
     /// Sort channels by `cid`.
     /// **Note**: This sorting option can extend your response waiting time if used as primary one.
     case cid
+    /// Sort channels by unread state. When using this sorting key, every unread channel weighs the same,
+    /// so they're sorted by `updatedAt`
```
tbarbugli, line 23: you should clarify that channels with unread messages will come first ;)

b-onc: They'll come first only if you use `ascending: true`, they can come last if you use `ascending: false`. Sorting keys do not contain sorting order

dmigach (approved):

PR #1352, opened 2021-08-10
Fix not selecting composer suggestions

nuno-vieira:
This bug was introduced after moving the suggestions to be a child view controller of the composer. This creates a problem because the suggestions are outside of the Composer's bounds and so it doesn't get touch events. The `showSuggestions()` implementation was reverted to add the childViewController into the parent of the composer.


b-onc (approved):
Thank you! Please create a ticket and put the ticket number in the PR title

PR #1353, opened 2021-08-10
[CIS-1094] Fix ChannelListController.synchronize not calling completion

b-onc:
...when connected in offline mode

This happened because the EventObserver didn't get `connected` callback when we're already connected


tbarbugli (approved):
LGTM

dmigach (approved):

PR #1357, opened 2021-08-10
[CIS-1093] Search users inside channel for mentions

tbarbugli:
The default behaviour is now fixed to perform user search on channel watchers and members. It is possible to perform user search and suggest any user from the app by changing `Components.mentionAllAppUsers` to true.


Sources/StreamChatUI/CommonViews/Suggestions/ChatSuggestionsVC.swift
```diff
@@ -189,6 +189,9 @@ open class ChatMessageComposerSuggestionsCommandDataSource: NSObject, UICollecti
 open class ChatMessageComposerSuggestionsMentionDataSource: NSObject,
     UICollectionViewDataSource,
     ChatUserSearchControllerDelegate {
+    /// internal cache for users
+    private(set) var usersCache: [ChatUser]
+
```
nuno-vieira, line 194: I was expecting this to be part of `ChatUserSearchController`. I can see that there is a comment on `ChatUserSearchController` saying: **Note: Currently, no local data will be searched, only remote data will be queried.**

Which means that the goal is to add support for it there. And it makes sense, since the search could be implemented outside of the composer suggestions, so it feels like this should not be done in the composer, but in the search implementation. But not sure how that could be done inside the user search controller, maybe @b-onc has a better understanding of this?

Sources/StreamChatUI/CommonViews/Suggestions/ChatSuggestionsVC.swift
```diff
@@ -233,7 +239,7 @@ open class ChatMessageComposerSuggestionsMentionDataSource: NSObject,
     }
 
     public func collectionView(_ collectionView: UICollectionView, numberOfItemsInSection section: Int) -> Int {
-        searchController.users.count
+        usersCache.count
     }
```
nuno-vieira, line 243: Like I said above, the data should still come from the search controller. The search controller will decide if the data comes from remote or local

Sources/StreamChatUI/Components.swift
```diff
@@ -264,6 +264,9 @@ public struct Components {
     /// A view controller that shows suggestions of commands or mentions.
     public var suggestionsVC: ChatSuggestionsVC.Type = ChatSuggestionsVC.self
 
+    /// When true the suggestionsVC will search users from the entire application instead of limit search to the current channel.
+    public var mentionAllAppUsers: Bool = false
+
```
nuno-vieira, line 269: This looks like it should be a config in the `ChatUserSearchController`.

Sources/StreamChatUI/Composer/ComposerVC.swift
```diff
@@ -175,6 +176,9 @@ open class ComposerVC: _ViewController,
         channelConfig?.uploadsEnabled == true
     }
 
+    /// When enabled mentions search users across the entire app instead of searching
+    open private(set) lazy var mentionAllAppUsers: Bool = components.mentionAllAppUsers
+
```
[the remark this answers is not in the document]
nuno-vieira: Update: This should live in `ChatUserSearchController`

Sources/StreamChatUI/Composer/ComposerVC.swift
```diff
@@ -627,20 +631,41 @@ open class ComposerVC: _ViewController,
...
+            userSearchController.search(
+                query: queryForMentionSuggestionsSearch(typingMention: typingMention)
+            )
+        } else {
+            usersCache = searchUsers(
+                channel.watchers.map { $0 } + channel.cachedMembers.map { $0 },
+                by: typingMention,
+                excludingId: currentUserId
+            )
+        }
```
nuno-vieira, line 653: All this should be an implementation detail of `ChatUserSearchController`. Ideally, the composer doesn't need to know where the data is coming from and how, it should just send a query to the search controller, and get users.

Sources/StreamChatUI/Composer/ComposerVC.swift
```diff
@@ -627,20 +631,41 @@ open class ComposerVC: _ViewController,
...
         suggestionsVC.didSelectItemAt = { [weak self] userIndex in
             guard let self = self else { return }
+            guard dataSource.usersCache.count >= userIndex else {
+                return
+            }
 
             let textView = self.composerView.inputMessageView.textView
-            let user = self.userSearchController.users[userIndex]
-
+            let user = dataSource.usersCache[userIndex]
```
nuno-vieira, line 668: What is the data comes from remote? This would not work right? Since it assumes we are always using the usersCache?

Sources/StreamChatUI/Composer/ComposerVC.swift
```diff
@@ -794,3 +819,77 @@ open class ComposerVC: _ViewController,
...
+                    let ins = mat[i][j - 1] + 1 // insertion
+                    let sub = mat[i - 1][j - 1] + 1 // substitution
+                    mat[i][j] = min(min(del, ins), sub)
+                }
+            }
+        }
+
+        return mat[sCount][oCount]
+    }
+}
```
nuno-vieira, line 895: I think all these should live in the `UserSearchController`. And since the levenshtein algorithm can be used in other types of searches, it is worth creating the extension in a separate file, currently all our string extensions are in "Utils/String+Extensions.swift", and the test of the algorithm should also live in "Utils/String+Extensions_Tests.swift"

Sources/StreamChatUI/Composer/ComposerVC_Tests.swift
```diff
@@ -0,0 +1,80 @@
...
+import XCTest
+
+class ComposerVC_Tests: XCTestCase {
+    func testLevenshtein() throws {
+        XCTAssertEqual("".levenshtein(""), "".levenshtein(""))
+        XCTAssertEqual("".levenshtein(""), 0)
+        XCTAssertEqual("a".levenshtein(""), 1)
+        XCTAssertEqual("".levenshtein("a"), 1)
+        XCTAssertEqual("tommaso".levenshtein("ToMmAsO"), 4)
+    }
```
nuno-vieira, line 16: Like mentioned above, this test should be part of the string extensions tests

Sources/StreamChatUI/Composer/ComposerVC_Tests.swift
```diff
@@ -0,0 +1,80 @@
...
+        XCTAssertEqual(searchUsers(users, by: "Дмитрии"), [dmitriy])
+        XCTAssertEqual(searchUsers(users, by: "Дмитри"), [dmitriy])
+    }
+
+    func testMatchHappyFrench() throws {
+        let user = ChatUser.mock(id: "fra", name: "françois")
+
+        XCTAssertEqual(searchUsers([user], by: "françois"), [user])
+        XCTAssertEqual(searchUsers([user], by: "franc"), [user])
+    }
```
nuno-vieira, line 79: All these tests should be part of the UserSearchController. As for the testing name conventions, we should do:
`test_given_when_then`, where given is the function name. So for this case:
`test_searchUsers_whenHappyFrench_findsUser`.. something like this

nuno-vieira (changes requested):

Sources/StreamChatUI/CommonViews/Suggestions/ChatSuggestionsVC.swift
```diff
@@ -189,6 +189,9 @@ open class ChatMessageComposerSuggestionsCommandDataSource: NSObject, UICollecti
 open class ChatMessageComposerSuggestionsMentionDataSource: NSObject,
     UICollectionViewDataSource,
     ChatUserSearchControllerDelegate {
+    /// internal cache for users
+    private(set) var usersCache: [ChatUser]
```
nuno-vieira, line 193: Small detail: I think naming this just `users` is enough?
```suggestion
    private(set) var users: [ChatUser]
```

Sources/StreamChatUI/Composer/ComposerVC.swift
```diff
@@ -627,20 +631,41 @@ open class ComposerVC: _ViewController,
...
+                by: typingMention,
+                excludingId: currentUserId
+            )
+        }
+
         let dataSource = ChatMessageComposerSuggestionsMentionDataSource(
             collectionView: suggestionsVC.collectionView,
-            searchController: userSearchController
+            searchController: userSearchController,
+            usersCache: usersCache
```
nuno-vieira, line 658: Since for now, we are not going to introduce a `MemberSearchController`, for me, it would make sense that this logic leaves inside the `ChatMessageComposerSuggestionsMentionDataSource`. But since this is pretty old code, and IMO it should be refactored, we can leave it like this for now. But it does not make sense that we need to call "search" outside of the data source. Ideally, the data source would have a search() function that would abstract how it is finding the members/users... Sot it would user the `userSearchController` or the `channel.watchers`, so Ideally the data source would have a dependency of `searchController´ and `channel`. Anyways, not important for this PR I guess.

Sources/StreamChatUI/Composer/ComposerVC.swift
```diff
@@ -794,3 +819,36 @@ open class ComposerVC: _ViewController,
...
+    return uniqueUsers.values.map { $0 }.sorted {
+        distance($0) < distance($1)
+    }
+}
+
+extension String {
+    subscript(index: Int) -> Character {
+        self[self.index(startIndex, offsetBy: index)]
+    }
+}
```
nuno-vieira, line 854: It is not really common to use global functions in swift for this kind of stuff. Better to put this `searchUsers()` as a function of the `ComposerVC`, and maybe even put it ´open` in case the customer wants to change the algorithm for some reason.

nuno-vieira (commented):
Re-review after the discussion on slack. Just some small comments, I guess only the moving of "searchUsers" function is a required change. Anything else LGTM, for now

Sources/StreamChatUI/Composer/ComposerVC.swift
```diff
@@ -794,3 +819,36 @@ open class ComposerVC: _ViewController,
         }
     }
 }
+
+/// searchUsers does an autocomplete search on a list of ChatUser and returns users with `id` or `name` containing the search string
+/// results are returned sorted by their edit distance from the searched string
+/// distance is calculated using the levenshtein algorithm
+/// both search and name strings are normalized (lowercased and by replacing diacritics)
+func searchUsers(_ users: [ChatUser], by searchInput: String, excludingId: String? = nil) -> [ChatUser] {
```
tbarbugli, line 827: making this an open func means a breaking change (I am 100% sure we will change this in a few weeks to better support this feature)

nuno-vieira: Ah I see, good point 👍

nuno-vieira (approved):
LGTM as temp solution 👍

PR #1359, opened 2021-08-11
[CIS-1078] Remove print statements

petyosi:
Two stale debug statements

### DIFF
--- a/Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
+++ b/Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
@@ -563,7 +563,6 @@ open class ChatMessageListVC:
     
     open func messageContentViewDidTapOnQuotedMessage(_ indexPath: IndexPath?) {
         guard let indexPath = indexPath else { return log.error("IndexPath is not available") }
-        print(#function, indexPath)
     }
     
     open func numberOfSections(in tableView: UITableView) -> Int {
--- a/Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
+++ b/Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
@@ -563,6 +563,10 @@ open class ChatMessageListVC:
     
     open func messageContentViewDidTapOnQuotedMessage(_ indexPath: IndexPath?) {
         guard let indexPath = indexPath else { return log.error("IndexPath is not available") }
+        log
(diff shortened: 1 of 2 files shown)


Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -563,7 +563,6 @@ open class ChatMessageListVC:
     
     open func messageContentViewDidTapOnQuotedMessage(_ indexPath: IndexPath?) {
         guard let indexPath = indexPath else { return log.error("IndexPath is not available") }
-        print(#function, indexPath)
```
kojiba, line 566: Should this be exchanged with other logger?
like log.info or log.debug

Sources/StreamChatUI/ChatMessageList/ChatThreadVC.swift
```diff
@@ -456,7 +456,6 @@ open class ChatThreadVC:
 
     open func messageContentViewDidTapOnQuotedMessage(_ indexPath: IndexPath?) {
         guard let indexPath = indexPath else { return log.error("IndexPath is not available") }
-        print(#function, indexPath)
```
kojiba, line 459: Should this be exchanged with other logger?
like log.info or log.debug

kojiba (commented):
LGTM, but couple of questions

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
SagarSDagdu: We are not printing anything useful so `log` is not needed IMO. 

But in that case, we may need to delete the above statement where we get indexPath as it will generate an unused warning. Let's replace the print with `log.info()` 👍🏻

tbarbugli: I think it would be useful to log something like: "You tapped on a quoted message, you can implement this method if you want to do something about it"

petyosi: This looks more like a debugging leftover than reasonable logging. Might be wrong, though

kojiba (approved):
LGTM

SagarSDagdu (approved):
LGTM 🚀

tbarbugli (approved):

PR #1361, opened 2021-08-11
[CIS-1101] Fix jumps when presenting message popup actions in a modal

nuno-vieira:
Fixes a UI jump glitch when opening the message actions popup when the message list is presented in a modal.

**Before**
before
**After**
after


b-onc (approved):
This is tested with the SwiftUI test app from H, right?
CHANGELOG please

PR #1363, opened 2021-08-11
Fix ImagePreview nested type for objc bridging header

nuno-vieira:
Add support for `ChatMessageGalleryView.ImagePreview` to be used from Obj-c


b-onc (approved):
CHANGELOG please

PR #1365, opened 2021-08-12
[CIS-1105] Take a more defensive approach for user.role

tbarbugli:
Replace force-unwrap with a custom case for user roles.


b-onc (dismissed):
CHANGELOG please

nuno-vieira (changes requested):
Better to replicate the MemberRole's implementation: https://github.com/GetStream/stream-chat-swift/pull/1288

Sources/StreamChat/Models/User.swift
```diff
@@ -93,16 +93,49 @@ extension ChatUser: Hashable {
...
     
-    /// A user that connected using guest user authentication.
-    case guest
+    public init(stringLiteral value: String) {
+        self.init(rawValue: value)
+    }
     
+    static func custom(_ rawValue: String) -> UserRole {
+        .init(rawValue: rawValue)
+    }
```
b-onc, line 109: Why do we need this?

tbarbugli: no need indeed :)

nuno-vieira (approved):
LGTM ✅

PR #1368, opened 2021-08-12
CIS-1110 Fix time ago formatting

tbarbugli:
Adds a utility formatter to represent a date in the past as a string, I used a homegrown approach to support iOS < 13

### Before
IMG_3DDD4C3D11A8-1

## After
IMG_5E22E5BA9C0D-1

### DIFF
--- a/DemoApp/en.lproj/Localizable.strings
+++ b/DemoApp/en.lproj/Localizable.strings
@@ -51,12 +51,22 @@
 "message.sending.attachment-uploading-failed" = "UPLOADING FAILED";
 
 "message.title.online" = "Online";
-"message.title.see-minutes-ago" = "Seen %@ ago";
 "message.title.offline" = "Offline";
 "message.title.group" = "%d members, %d online";
 
 "current-selection" = "%d of %d";
 
 "attachment.max-size-exceeded" = "Attachment size exceed the limit.";
-
+"dates.time-ago-seconds-plural" = "%d seconds ago";
+"dates.time-ago-seconds-singular" = "just one second ago";
+"dates.time-ago-minutes-singular" = "one minute ago";
+"dates.time-ago-minutes-plural" = "%d minutes ago";
+"dates.time-ago-hours-singular" = "one hour ago";
+"dates.time-ago-hours-plural" = "%d hours ago";
+"dates.time-ago-days-singular" = "one day ago";
+"dates.time-ago-days-plural" = "%d days ago";
+"dates.time-ago-weeks-singular" = "one week ago";
+"dates.time-ago-weeks-plural" = "%d weeks ago";
+"dates.time-ago-months-singular" = "one month ago";
+"dates.time-ago-months-plural" = "%d months ago";
 // MARK: -  DemoApp Strings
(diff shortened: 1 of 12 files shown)


Sources/StreamChatUI/Utils/DateUtils.swift
```diff
@@ -0,0 +1,53 @@
...
+            let minuteAgo = calendar.date(byAdding: .minute, value: -1, to: now),
+            let hourAgo = calendar.date(byAdding: .hour, value: -1, to: now),
+            let dayAgo = calendar.date(byAdding: .day, value: -1, to: now),
+            let weekAgo = calendar.date(byAdding: .day, value: -7, to: now),
+            let monthAgo = calendar.date(byAdding: .day, value: -31, to: now)
+        else { return nil }
+        
+        if minuteAgo < date {
+            let diff = calendar.dateComponents([.second], from: date, to: now).second ?? 0
+            return diff > 1 ? L10n.Dates.timeAgoSecondsPlural(diff) : L10n.Dates.timeAgoSecondsSingular
```
dmigach, line 27: Do you think it makes sense to make it custom to support localisation if localization won't work properly anyways?
For Russian (and many others):
1 день (singular)
2...4 дня (plural)
5...20 дней (another plural)
21,31,41... день (singular again, whoops)

Sources/StreamChatUI/Utils/DateUtils.swift
```diff
@@ -0,0 +1,53 @@
+//
+// Copyright © 2021 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+class DateUtils {
```
b-onc, line 7: We should make this component subclassable and injectable (via Components) so users can do whatever they want

DemoApp/en.lproj/Localizable.strings
```diff
@@ -51,12 +51,22 @@
...
 "message.title.online" = "Online";
-"message.title.see-minutes-ago" = "Seen %@ ago";
 "message.title.offline" = "Offline";
 "message.title.group" = "%d members, %d online";
 
 "current-selection" = "%d of %d";
 
 "attachment.max-size-exceeded" = "Attachment size exceed the limit.";
-
+"dates.time-ago-seconds-plural" = "%d seconds ago";
```
b-onc, line 60: I think we should keep "Seen" word for these... it's weird to see just "x months ago" without the verb "Seen"

Sources/StreamChatUI/Utils/DateUtils.swift
tbarbugli: ahah nice, I had no idea Russian had this :) The current implementation makes sense, I'd say we make this swappable via Appearance object and/or subclassing

Sources/StreamChatUI/Gallery/GalleryVC.swift
```diff
@@ -46,6 +46,9 @@ open class GalleryVC:
         return videos + images
     }
     
+    /// Returns the date formater function used to represent when the user was last seen online
+    open var lastSeenDateFormatter: (Date) -> String? { DateUtils.timeAgo }
```
b-onc, line 50: Instead of this, why don't we inject this via `Components`? we do the same injection mechanism for 
- localizationProvider
- channelNamer
- messageLayoutOptionsResolver

tbarbugli: Components is growing like crazy, I think for small things like this is better not to use it (also in this case we'd need to have one for the gallery and one for the channel header)

b-onc (approved):

PR #1373, opened 2021-08-13
[CIS-1111] Copy messages to avoid out of sync table and collection crashes

tbarbugli:
Problem: `dataSource?.tableView(self, cellForRowAt: indexPath)` is used inside the code that update the message list. 

This does not work because it accesses `controller.messages` which is on a different state than the `tableView`.

The solution: keep a copy of `controller.messages` that is consistent with the state of the table view.

### DIFF
--- a/Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
+++ b/Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
@@ -323,7 +323,7 @@ open class ChatMessageListVC:
     open func updateMessages(with changes: [ListChange<ChatMessage>], completion: (() -> Void)? = nil) {
         listView.updateMessages(
             with: changes,
-            onMessagesCountUpdate: updateMessageCache,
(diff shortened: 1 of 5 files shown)


Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -323,13 +323,13 @@ open class ChatMessageListVC:
     open func updateMessages(with changes: [ListChange<ChatMessage>], completion: (() -> Void)? = nil) {
         listView.updateMessages(
             with: changes,
-            onMessagesCountUpdate: updateNumberOfMessages,
+            onMessagesCountUpdate: updateMessageCache,
```
b-onc, line 326:
```suggestion
            onMessagesUpdate: updateMessageCache,
```

b-onc (approved):
one renaming and LGTM

dmigach (dismissed):

PR #1375, opened 2021-08-13
CIS-1113 Fix inconsistent/wrong RawJSON number type handling

tbarbugli:
The initial implementation of RawJSON had `double` and `integer` types. This made decoding incorrect because integer values can be Int and Double and the decoder would always decode them as Int.

This PR removes `.integer` from RawJSON, adds a new case `.number` and keeps `.double` around as an "alias" of `.number`


nuno-vieira (approved):
LGTM!

PR #1379, opened 2021-08-16
[CIS-1115] Safer DTO -> Model

tbarbugli:


Sources/StreamChat/Database/DTOs/ChannelDTO.swift
```diff
@@ -289,29 +289,26 @@ extension ChannelDTO {
...
-            extraData = try JSONDecoder.default.decode([String: RawJSON].self, from: dto.extraData)
-        } catch {
-            log.error(
-                "Failed to decode extra data for Channel with cid: <\(dto.cid)>, using default value instead. "
-                    + "Error: \(error)"
-            )
-            extraData = [:]
+
+        if dto.isDeleted {
+            log.warning("Cannot create an instance of ChatMessageReaction, the DTO was marked as deleted")
```
nuno-vieira, line 301:
```suggestion
            log.warning("Cannot create an instance of ChatChannel, the DTO was marked as deleted")
```

Sources/StreamChat/Workers/ChannelListUpdater.swift
```diff
@@ -25,10 +25,10 @@ class ChannelListUpdater: Worker {
...
                 case let .success(channelListPayload):
                     self?.database.write { session in
-                        
-                        if trumpExistingChannels {
-                            try session.deleteChannels(query: channelListQuery)
-                        }
+                        // TODO: this is not working well, we need a better approach to remove channels that are not in-sync
+//                        if trumpExistingChannels {
+//                            try session.deleteChannels(query: channelListQuery)
+//                        }
```
nuno-vieira, line 31: Doesn't the safer DTO's resolve this? Or the crash still happens?

nuno-vieira (approved):
LGTM overall, just some small comments. I agree we should have a more defensive approach when converting the DTO's to the models 👍

Sources/StreamChat/Workers/ChannelListUpdater.swift
tbarbugli: it does not crash but it removes channels from the database that should not be removed (in my tests that breaks pagination) @dmigach FYI

Sources/StreamChat/Controllers/ListDatabaseObserver.swift
```diff
@@ -112,7 +112,7 @@ class ListDatabaseObserver<Item, DTO: NSManagedObject> {
     /// Used for observing the changes in the DB.
     private(set) var frc: NSFetchedResultsController<DTO>!
     
-    let itemCreator: (DTO) -> Item
+    let itemCreator: (DTO) -> Item?
```
nuno-vieira, line 115:
```suggestion
    let itemCreator: (DTO) -> Item
```
I don't think we need this to be optional. I would switch back to how it was

Sources/StreamChat/Controllers/ListDatabaseObserver.swift
```diff
@@ -285,7 +284,12 @@ class ListChangeAggregator<DTO: NSManagedObject, Item>: NSObject, NSFetchedResul
...
-        guard let dto = anObject as? DTO, let item = itemCreator(dto) else {
+        guard let dto = anObject as? DTO else {
+            log.warning("Skipping the update from DB because anObject can't be converted to DTO.")
+            return
+        }
+
+        guard let item = itemCreator(dto) else {
             log.warning("Skipping the update from DB because the DTO can't be converted to the model object.")
             return
         }
```
nuno-vieira, line 295: We don't want this. We want the `item` to exist, if a relationship is not found for example in `ChannelDTO` it doesn't mean the channel can't be created, it only mean it will have some properties `nil`

Sources/StreamChat/Controllers/MemberController/MemberController.swift
```diff
@@ -195,7 +195,7 @@ extension ChatChannelMemberController {
         var memberObserverBuilder: (
             _ context: NSManagedObjectContext,
             _ fetchRequest: NSFetchRequest<MemberDTO>,
-            _ itemCreator: @escaping (MemberDTO) -> ChatChannelMember,
+            _ itemCreator: @escaping (MemberDTO) -> ChatChannelMember?,
```
nuno-vieira, line 198:
```suggestion
            _ itemCreator: @escaping (MemberDTO) -> ChatChannelMember,
```

Sources/StreamChat/Controllers/MemberListController/MemberListController.swift
```diff
@@ -160,7 +160,7 @@ extension ChatChannelMemberListController {
         var memberListObserverBuilder: (
             _ context: NSManagedObjectContext,
             _ fetchRequest: NSFetchRequest<MemberDTO>,
-            _ itemCreator: @escaping (MemberDTO) -> ChatChannelMember,
+            _ itemCreator: @escaping (MemberDTO) -> ChatChannelMember?,
```
nuno-vieira, line 163:
```suggestion
            _ itemCreator: @escaping (MemberDTO) -> ChatChannelMember,
```

Sources/StreamChat/Controllers/MessageController/MessageController.swift
```diff
@@ -420,7 +420,7 @@ extension ChatMessageController {
         var messageObserverBuilder: (
             _ context: NSManagedObjectContext,
             _ fetchRequest: NSFetchRequest<MessageDTO>,
-            _ itemCreator: @escaping (MessageDTO) -> ChatMessage,
+            _ itemCreator: @escaping (MessageDTO) -> ChatMessage?,
```
nuno-vieira, line 423:
```suggestion
            _ itemCreator: @escaping (MessageDTO) -> ChatMessage,
```

Sources/StreamChat/Controllers/SearchControllers/UserSearchController.swift
```diff
@@ -222,7 +222,7 @@ extension ChatUserSearchController {
         var createUserListDatabaseObserver: (
             _ context: NSManagedObjectContext,
             _ fetchRequest: NSFetchRequest<UserDTO>,
-            _ itemCreator: @escaping (UserDTO) -> ChatUser
+            _ itemCreator: @escaping (UserDTO) -> ChatUser?
```
nuno-vieira, line 225:
```suggestion
            _ itemCreator: @escaping (UserDTO) -> ChatUser
```

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
```diff
@@ -289,29 +289,26 @@ extension ChannelDTO {
...
-            log.error(
-                "Failed to decode extra data for Channel with cid: <\(dto.cid)>, using default value instead. "
-                    + "Error: \(error)"
-            )
-            extraData = [:]
+
+        if dto.isDeleted {
+            log.warning("Cannot create an instance of ChatChannel, the DTO was marked as deleted")
+            return nil
+        }
```
nuno-vieira, line 303: I don't think we need this 
```suggestion

```

Sources/StreamChat/Database/DTOs/MemberModelDTO.swift
```diff
@@ -130,11 +130,20 @@ extension NSManagedObjectContext {
...
+    func asModel() -> ChatChannelMember? { .create(fromDTO: self) }
 }
 
 extension ChatChannelMember {
-    fileprivate static func create(fromDTO dto: MemberDTO) -> ChatChannelMember {
+    fileprivate static func create(fromDTO dto: MemberDTO) -> ChatChannelMember? {
+        if dto.isDeleted {
+            log.warning("Cannot create an instance of MemberDTO id: \(dto.id), the DTO was marked as deleted")
+            return nil
+        }
```
nuno-vieira, line 141:
```suggestion

```

Sources/StreamChat/Database/DTOs/MemberModelDTO.swift
```diff
@@ -130,11 +130,20 @@ extension NSManagedObjectContext {
...
+
+        guard let user = dto.user.asModel() else {
+            return nil
+        }
```
nuno-vieira, line 145: Some here, we can have a `ChatChannelMember` without a `user`. Ideally `dto.user` should be optional

Sources/StreamChat/Database/DTOs/MessageReactionDTO.swift
```diff
@@ -121,7 +121,16 @@ extension NSManagedObjectContext {
 
 extension MessageReactionDTO {
     /// Snapshots the current state of `MessageReactionDTO` and returns an immutable model object from it.
-    func asModel() -> ChatMessageReaction {
+    func asModel() -> ChatMessageReaction? {
+        if isDeleted {
+            log.warning("Cannot create an instance of ChatMessageReaction, the DTO was marked as deleted")
+            return nil
+        }
```
nuno-vieira, line 128:
```suggestion
```

Sources/StreamChat/Database/DTOs/UserDTO.swift
```diff
@@ -192,7 +191,12 @@ extension UserDTO {
 }
 
 extension ChatUser {
-    fileprivate static func create(fromDTO dto: UserDTO) -> ChatUser {
+    fileprivate static func create(fromDTO dto: UserDTO) -> ChatUser? {
+        if dto.isDeleted {
+            log.warning("Cannot create an instance of ChatUser id: \(dto.id), the DTO was marked as deleted")
+            return nil
+        }
```
nuno-vieira, line 198:
```suggestion

```

Sources/StreamChat/Workers/ChannelListUpdater.swift
nuno-vieira: We can uncomment this. The reason why this was now not behaving correctly it is because of this line: https://github.com/GetStream/stream-chat-swift/pull/1379/files#diff-a233c9d8c6d1ea11dc5c3dd6e22e1db3cc35e3667a3a088e23aa9a2671483d03R292. The `delete` event is triggered and we should still call `deleteRows` but because the DTO fails to be created, we don't call it.

nuno-vieira (changes requested):
The real solution should be to make all `one-to-one` relationships to be nullable. It should not be by making a parent nullable because it can't create a relationship. Example:

Bad:
- `ChannelDTO` has `createdBy: User`
- If `ChannelDTO.createdBy.asModel == nil` then `ChannelDTO is nil`

Good:
- `ChannelDTO` has `createdBy: User?`
- If `ChannelDTO.createdBy.asModel == nil` then `ChannelDTO exists without `createdBy`

The `ChannelDTO.createdBy == nil` will only happen in a delete event, so it won't cause any issues or invalid state.

Sources/StreamChat/Controllers/MessageController/MessageController.swift
```diff
@@ -458,7 +458,7 @@ private extension ChatMessageController {
                     sortAscending: sortAscending,
                     deletedMessagesVisibility: deletedMessageVisibility
                 ),
-                itemCreator: { $0.asModel() as ChatMessage }
+                itemCreator: { $0.asModel() as ChatMessage? }
```
kojiba, line 461: why not `as? ChatMessage`?

PR #1380, opened 2021-08-16
[CIS-1120] Fix making multiple requests when loading more messages

nuno-vieira:
When fetching more messages (pagination) in the message list, the request was done multiple times since it was triggered when each of the last 5 messages appeared, instead of being triggered only once, when the last 5th messages appeared.


nuno-vieira:
Closing in favor of https://github.com/GetStream/stream-chat-swift/pull/1381

PR #1381, opened 2021-08-17
[CIS-1120] Fix pagination on all lists

tbarbugli:
Ensures that Thread, MessageList and ChannelList

- Do not perform more than 1 API call in parallel to fetch prev/nex page
- Start fetching results at the right time 
- Stop paginating results when there is no more to fetch

### DIFF
--- a/Sources/StreamChat/Utils/Atomic_Tests.swift
+++ b/Sources/StreamChat/Utils/Atomic_Tests.swift
@@ -8,7 +8,8 @@ import XCTest
 class Atomic_Tests: StressTestCase {
     @Atomic var stringAtomicValue: String?
     @Atomic var intAtomicValue: Int = 0
-    
+    @Atomic var boolAtomicValue: Bool = false
+
     func test_Atomic_asPropertyWrapper() {
         stringAtomicValue = nil
         XCTAssertEqual(stringAtomicValue, nil)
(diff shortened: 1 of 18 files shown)


Sources/StreamChat/Workers/ChannelListUpdater.swift
```diff
@@ -25,24 +25,24 @@ class ChannelListUpdater: Worker {
...
                 case let .success(channelListPayload):
                     self?.database.write { session in
-                        
-                        if trumpExistingChannels {
-                            try session.deleteChannels(query: channelListQuery)
-                        }
+                        // TODO: this is not working well, we need a better approach to remove channels that are not in-sync
+//                        if trumpExistingChannels {
+//                            try session.deleteChannels(query: channelListQuery)
+//                        }
```
nuno-vieira, line 31: Do we want this go in in this PR?

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -244,11 +246,24 @@ open class ChatMessageListVC:
...
+            return
+        }
+
+        if _loadingPreviousMessages.compareAndSwap(old: false, new: true) {
+            channelController.loadPreviousMessages(completion: { [weak self] _ in
+                guard let self = self else {
+                    return
+                }
+                self.loadingPreviousMessages = false
+            })
```
nuno-vieira, line 263: We don't need `loadingPreviousMessages` to be atomic. `willDisplay` function is always called on the main thread which is a serial thread, so there won't be any concurrent issues in this case because we can guarantee this logic is always called in the same thread. 

One problem I can see with this approach is that if the network is bad, or the user is scrolling too fast, there's a chance we might skip one loading request, since you will only be able to load previous messages one at the time. There's no reason for having only one loading request per pagination. If we do this, the last loading request will always impact the next one if the network conditions are not the best.

Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
```diff
@@ -80,6 +82,28 @@ open class ChatChannelListVC: _ViewController,
         userAvatarView.addTarget(self, action: #selector(didTapOnCurrentUserAvatar), for: .touchUpInside)
     }
     
+    private func loadMoreChannels() {
+        if _loadingPreviousMessages.compareAndSwap(old: false, new: true) {
```
nuno-vieira, line 86: No need for CAS here like mentioned above

Sources/StreamChat/Utils/Atomic_Tests.swift
```diff
@@ -23,6 +24,33 @@ class Atomic_Tests: StressTestCase {
...
+                        swaps += 1
+                    }
+                }
+            }
+        }
+        
+        /// sleep for 1ms to avoid AssertAsync to quit too early (rare but possible)
+        usleep(1000)
+        AssertAsync.willBeEqual(swaps, 1)
+    }
```
nuno-vieira, line 52: Since we don't need CAS for our specific problem, I would remove this, since we might not need it at all in the future

nuno-vieira (changes requested):
Although I think the solution could be simpler and more efficient, the only thing I believe we should for sure change is the usage of `compareAndSwap` since it is not doing anything in this case. Other than that, LGTM 👍

nuno-vieira (dismissed):

PR #1390, opened 2021-08-19
[CIS-1135] Fix user mention suggestions not showing all members

nuno-vieira:
The members that had a missing `lastActiveAt` weren't being fetched. So the sort descriptors should be `lastActiveAt` and the default so that the members and watchers are ordered by last active first, and then the default descriptor, in case the DTO doesn't have `lastActiveAt`

### DIFF
--- a/Sources/StreamChatUI/Composer/ComposerVC.swift
+++ b/Sources/StreamChatUI/Composer/ComposerVC.swift
@@ -185,6 +185,9 @@ open class ComposerVC: _ViewController,
     /// A controller that manages the channel that the composer is creating content for.
     open var channelController: ChatChannelController?
 
+    /// A controller that manages the members of the channel.
+    open var memberListController: ChatChannelMemberListController?
+
     /// The channel config. If it's a new channel, an empty config should be created. (Not yet supported right now)
     public var channelConfig: ChannelConfig? {
         channelController?.channel?.config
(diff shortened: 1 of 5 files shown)


Sources/StreamChatUI/Composer/ComposerVC.swift
```diff
@@ -646,7 +655,7 @@ open class ComposerVC: _ViewController,
             )
         } else {
             usersCache = searchUsers(
-                channel.watchers.map { $0 } + channel.cachedMembers.map { $0 },
+                channel.watchers.map { $0 } + (memberListController?.members.compactMap { $0 } ?? []),
```
kojiba, line 658: `memberListController?.members.compactMap`

won't this impact performance as well? 
iterating on long arrays is N complexity

Sources/StreamChatUI/Composer/ComposerVC.swift
```diff
@@ -244,6 +247,12 @@ open class ComposerVC: _ViewController,
     override open func setUp() {
         super.setUp()
 
+        let client = channelController?.client
+        if let cid = channelController?.cid {
+            memberListController = client?.memberListController(query: .init(cid: cid))
+            memberListController?.synchronize()
```
kojiba, line 253: hope this also doing the job not on a main queue

nuno-vieira: Everything on LLC is not doing work on the main thread ofc

Sources/StreamChatUI/Composer/ComposerVC.swift
nuno-vieira: No, it won't, it's maxed at 100 members. Either way, we should not prematurely optimize for performance, but for readability instead.

tbarbugli (approved):

PR #1391, opened 2021-08-20
[CIS-1117] Change unowned to weak self in blocks

kojiba:
[-] Added replacement of unowned self everywhere in blocks
    in production code to guard let self
[-] Added replacement of unowned self everywhere in blocks
    in test code
[-] Improving tests to be compilable

Test plan:
- Open freshly installed app
- Select c3po user
- Open chat list
- Open chat
- Send message
- View photos
- Open chat with unread count 
- Create new chat with multiple users
All Pass

React on message - does no add reaction instantly

### DIFF
--- a/Sample/Samples/UserUpdateViewController.swift
+++ b/Sample/Samples/UserUpdateViewController.swift
@@ -71,7 +71,12 @@ class UserUpdateViewController: UITableViewController {
             title: cell?.textLabel?.text ?? "",
             initialValue: cell?.detailTextLabel?.text ?? ""
         )
-        inputVC.onChange = { [unowned self] newValue in
+        inputVC.onChange = { [weak self] newValue in
+            guard let self = self else {
+                log.warning("Callback called while self is nil")
+                return
+            }
+
             switch keyPath {
             case \ChatUser.name:
                 self.userName = newValue
--- a/Sources/StreamChat/Workers/Background/ConnectionRecoveryUpdater.swift
+++ b/Sources/StreamChat/Workers/Background/ConnectionRecoveryUpdater.swift
@@ -81,12 +81,17 @@ class ConnectionRecoveryUpdater: EventWorker {
         connectionObserver = EventObserver(
             notificationCenter: eventNotificationCenter,
             transform: { $0 as? ConnectionStatusUpdated },
-            callback: { [unowned self] in
+            callback: { [weak self] in
+                guard let self = self else {
+                    log.warning("Callback called while self is nil")
+                    return
+                }
+
                 switch $0.webSocketConnectionState {
                 case .connecting:
                     self.obtainLastSyncDate()
                 case .connected:
-                    fetchAndReplayMissingEvents()
+                    self.fetchAndReplayMissingEvents()
                 default:
                     break
                 }
--- a/Sources/StreamChat/APIClient/APIClient.swift
+++ b/Sources/StreamChat/APIClient/APIClient.swift
@@ -41,7 +41,7 @@ class APIClient {
     ///   - endpoint: The `Endpoint` used to create the network request.
     ///   - completion: Called when the networking request is finished.
     func request<Response: Decodable>(endpoint: Endpoint<Response>, completion: @escaping (Result<Response, Error>) -> Void) {
-        encoder.encodeRequest(for: endpoint) { [unowned self] (requestResult) in
+        encoder.encodeRequest(for: endpoint) { [weak self] (requestResult) in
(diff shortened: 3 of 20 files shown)


kojiba:
Updating PR with tests using weak and runnable

evsaev:
Looking good 👍 The only concern is that a warning might not be noticed/ignored by a customer.
Using `log.assertionFailure` fits better here: the app will crash in `Debug` so no chance for a bug to be ignored and at the same time, there won't any crashes causes by SDK in prod

evsaev (commented):
Looking good 👍 The only concern is that a warning might not be noticed/ignored by a customer. Using log.assertionFailure fits better here: the app will crash in Debug so no chance for a bug to be ignored and at the same time, there won't any crashes causes by SDK in prod

Sources/StreamChat/Workers/Background/ConnectionRecoveryUpdater.swift
```diff
@@ -81,12 +81,18 @@ class ConnectionRecoveryUpdater: EventWorker {
         connectionObserver = EventObserver(
             notificationCenter: eventNotificationCenter,
             transform: { $0 as? ConnectionStatusUpdated },
-            callback: { [unowned self] in
+            callback: { [weak self] in
+
+                guard let self = self else {
+                    log.warning("Callback called while self is nil")
```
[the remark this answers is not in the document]
kojiba: yes, it automatically includes #function parameter inside

kojiba:
> Looking good 👍 The only concern is that a warning might not be noticed/can be ignored by a customer. Using `log.assertionFailure` fits better here: the app will crash in Debug so no chance for a bug to be ignored and at the same time, there won't any crashes inside SDK on prod

if this situation should never be executed in theory, then I could replace with log.assertionFailure, 
otherwise I would prefer to keep this warning/error

evsaev:
> > Looking good 👍 The only concern is that a warning might not be noticed/can be ignored by a customer. Using `log.assertionFailure` fits better here: the app will crash in Debug so no chance for a bug to be ignored and at the same time, there won't any crashes inside SDK on prod
> 
> if this situation should never be executed in theory

@kojiba you mean it impossible for a customer to skip/forget to notify us about the warning in a console? 🤔

Sample/Samples/UserUpdateViewController.swift
```diff
@@ -71,7 +71,13 @@ class UserUpdateViewController: UITableViewController {
             title: cell?.textLabel?.text ?? "",
             initialValue: cell?.detailTextLabel?.text ?? ""
         )
-        inputVC.onChange = { [unowned self] newValue in
+        inputVC.onChange = { [weak self] newValue in
+
+            guard let self = self else {
+                log.warning("Callback called while self is nil")
+                return nil
+            }
```
nuno-vieira, line 79: I wouldn't even log. @kojiba @evsaev What is the value of logging here? There is none. I would remove all logs from `guard let self = self ...`.  Because in this happens, it most likely because the original `self` doesn't exist anymore, because it was deallocated, and so it doesn't matter anymore that the closure is completed.

evsaev: in some cases we must have `self` alive to do some job but strong ref is a no-go because of retain cycle. In this case, we don't want silently fail but crash/log to know there's an issue and we have wrong understanding of object lifetime

evsaev: Illiya has updated cases were `unowned` was used -> we were expecting `self` to be alive for some job. If it's not, there's an issue

nuno-vieira: We will never know the objects lifetime because we don't control them fully, since an app can use our SDK in a way we don't expect. Either way, if there is one job that we need to make sure it is done, let's log on those ones **only** and not blindly log everywhere, otherwise, the logs will be a mess 🙂

kojiba: yes, agree that in couple of cases we assume that it should be alive, so better to keep logs

tbarbugli (commented):
can we stick with unowned in tests? easy to miss a warning from tests. Rest looks OK to me

kojiba:
Yes, I will separate `unowned-weak test` changes into a new branch and we could decide do we want to carry it in or no

kojiba:
Separated tests to CIS-1117-Unowned-Refactor-Test branch

tbarbugli (approved):

PR #1392, opened 2021-08-20
[CIS-1136] Make it possible to encrypt/decrypt attachments

dmigach:
This change makes it possible to subclass our implementation of StreamCDNClient to implement their own
Also we make attachments mutable so it's possible to decrypt encrypted attachments
A customer gets names and avatarURLs for users from their own backend for security reasons, so they need to be able to set them.
Also this PR disables a flaky test


SagarSDagdu (dismissed):
LGTM!

Sources/StreamChat/Config/ChatClientConfig.swift
```diff
@@ -91,15 +91,20 @@ public struct ChatClientConfig {
     /// - Parameter apiKey: The API key of the chat app the `ChatClient` connects to.
     ///
     
-    /// Allows to inject a custom API client for uploading attachments, if not specified `StreamCDNClient` is used
-    public var customCDNClient: CDNClient?
+    public enum CDNClientConfig {
```
evsaev, line 94: Would be great to have some docs here 👍

Sources/StreamChat/Config/ChatClientConfig.swift
```diff
@@ -91,15 +91,20 @@ public struct ChatClientConfig {
...
     /// Returns max possible attachment size in bytes.
     /// The value is taken from custom `maxAttachmentSize` type custom `CDNClient` type.
     /// The default value is 20 MiB.
     public var maxAttachmentSize: Int64 {
-        if let customCDNClient = customCDNClient {
-            return type(of: customCDNClient).maxAttachmentSize
+        if case let .custom(client) = cdnClientConfig {
+            return type(of: client).maxAttachmentSize
         } else {
             return StreamCDNClient.maxAttachmentSize
```
evsaev, line 109: Shouldn't we take a type from `cdnClientConfig` here? 🤔

Sources/StreamChat/Models/Attachments/ChatMessageAudioAttachment.swift
```diff
@@ -17,13 +17,13 @@ public struct AudioAttachmentPayload: AttachmentPayload {
...
+    public var title: String?
     /// A link to the audio.
-    public internal(set) var audioURL: URL
+    public var audioURL: URL
     /// The audio itself.
-    public let file: AttachmentFile
+    public var file: AttachmentFile
     /// An extra data.
-    let extraData: [String: RawJSON]?
+    var extraData: [String: RawJSON]?
```
evsaev, line 26:
```suggestion
    public var extraData: [String: RawJSON]?
```

Sources/StreamChat/Models/Attachments/ChatMessageFileAttachment.swift
```diff
@@ -17,13 +17,13 @@ public struct FileAttachmentPayload: AttachmentPayload {
...
+    public var title: String?
     /// A link to the file.
-    public internal(set) var assetURL: URL
+    public var assetURL: URL
     /// The file itself.
-    public let file: AttachmentFile
+    public var file: AttachmentFile
     /// An extra data.
-    let extraData: [String: RawJSON]?
+    var extraData: [String: RawJSON]?
```
evsaev, line 26:
```suggestion
    public var extraData: [String: RawJSON]?
```

Sources/StreamChat/Models/Attachments/ChatMessageVideoAttachment.swift
```diff
@@ -17,13 +17,13 @@ public struct VideoAttachmentPayload: AttachmentPayload {
...
+    public var title: String?
     /// A link to the video.
-    public internal(set) var videoURL: URL
+    public var videoURL: URL
     /// The video itself.
-    public let file: AttachmentFile
+    public var file: AttachmentFile
     /// An extra data.
-    let extraData: [String: RawJSON]?
+    var extraData: [String: RawJSON]?
```
evsaev, line 26:
```suggestion
    public var extraData: [String: RawJSON]?
```

Sources/StreamChat/Config/ChatClientConfig.swift
evsaev: would be great to also have a test for it 👍

evsaev (commented):
Looks good, thanks! Let's address comments and we are good to 🚢

evsaev (commented):
Let's also have a changelog entry for this 🙏

dmigach:
A customer who requested these change is going to check if it helps them, if it does, we're going to finalise it. In the meantime I'll start addressing the issues 👌

Sources/StreamChat/Workers/ChannelUpdater_Tests.swift
```diff
@@ -425,7 +425,8 @@ class ChannelUpdater_Tests: StressTestCase {
         XCTAssertEqual(apiClient.request_endpoint, AnyEndpoint(referenceEndpoint))
     }
 
-    func test_hideChannel_successfulResponse_isPropagatedToCompletion() throws {
+    // TODO: Fix and enable
+    func _test_hideChannel_successfulResponse_isPropagatedToCompletion() throws {
```
nuno-vieira, line 429: Please remove this, it seems I was able to at least make it more stable on https://github.com/GetStream/stream-chat-swift/pull/1314

nuno-vieira:
> A customer who requested these change is going to check if it helps them, if it does, we're going to finalise it. In the meantime I'll start addressing the issues 👌

Any news about this one?

kojiba:
This decrease code readability cause when reader sees `var` he assumes its changing somewhere. 
Maybe there is options how to do this extendable but in more smart way, instead of just making everything public and var?

kojiba:
For example we could introduce a new delegate for ChannelController with will have a method 
`process(attachment: AttachmentInputType) -> AttachmentOutputType`
for input we can give a struct, and for output we could require just some proto to display data

Sources/StreamChat/Config/ChatClientConfig.swift
dmigach: It seems like we might want to drop this change in the end if the customer will be able to achieve e2e without it

Sources/StreamChat/Config/ChatClientConfig.swift
```diff
@@ -171,3 +176,22 @@ public struct APIKey: Equatable {
...
+            return client
+        }
+        set {
+            if let customClient = newValue {
+                cdnClientConfig = .custom(client: customClient)
+            } else {
+                cdnClientConfig = .default(type: StreamCDNClient.self)
+            }
+        }
+    }
```
nuno-vieira, line 196: No need to add deprecations, V4 will include breaking changes, so we can break the API for now 👍 Anyways, all deprecations should live in the `Deprecations.swift` file.

nuno-vieira (changes requested):

dmigach:
> > A customer who requested these change is going to check if it helps them, if it does, we're going to finalise it. In the meantime I'll start addressing the issues 👌
> 
> Any news about this one?

The customer was happy with the suggested solution, and they won't need to deal with CDN in the end. They went for implementing it so we might expect additional request in the nearest future

tbarbugli (approved):

tbarbugli (approved):

PR #1394, opened 2021-08-20
[CIS-1059] Improve re-create deleted channels issue

kojiba:
Test plan:
- Create chat between user and user2
- post message №1
- delete chat 
- create chat between user and user2
- post message №2
- only last message is visible, which is correct


nuno-vieira:
https://github.com/GetStream/stream-chat-swift/pull/1314/commits/d53fb012bac365342f7a69eccca1f63136b699b2

Issue #1377, opened 2021-08-15
StreamChat is not finishing background tasks - app terminated.


astas79:
When app with integrated ChatUI goes to background - IOS terminates the app with following message:

[BackgroundTask] Background task still not ended after expiration handlers were called: <_UIBackgroundTaskInfo: 0x2833cf1c0>: taskID = 96, taskName = Called by Test, from $s10StreamChat26IOSBackgroundTaskSchedulerC05beginD017expirationHandlerSbyycSg_tF, creationTime = 159368 (elapsed = 27). This app will likely be terminated by the system. Call UIApplication.endBackgroundTask(_:) to avoid this.

## What did you expect to happen?
App stays in the background in suspended state, instead of being terminated.

## What happened instead?
App gets terminated.

## GetStream Environment
beta.11
StreamChatUI
**iOS 14.7

dmigach:
Hello @astas79 ! 
Thanks for reporting this, I'm going to reproduce it and will let you know once I know more about the issue.

dmigach:
I spotted a minor inconsistency in logic related to your issue, but it's not clear if this could cause it. I wasn't able to reproduce the issue, so you can either try providing more details on how to reproduce this, or wait until the next release and check out if the issue still persists for you

PR #1395, opened 2021-08-20
[CIS-1137] Remove redundant background task cancelation

dmigach:
I investigated [this](https://github.com/GetStream/stream-chat-swift/issues/1377) issue and spotted this inconsistency
We don't need to pass it inside expiration handler because this is called in `BackgroundTaskScheduler` after the expiration handler. It's unclear if this caused any issues but it makes sense to fix it anyway

### DIFF
--- a/Sources/StreamChat/ChatClient_Tests.swift
+++ b/Sources/StreamChat/ChatClient_Tests.swift
@@ -925,9 +925,6 @@ class ChatClient_Tests: StressTestCase {
         
         // Assert that `disconnect` is called
         XCTAssertEqual(testEnv.clientUpdater?.disconnect_called, true)
-        
-        // Assert that background task is cancelled
-        XCTAssertEqual(testEnv.backgroundTaskScheduler?.endBackgroundTask_called, true)
     }
     
     func test_backgroundTaskCancelled_whenAppIsForegrounded() {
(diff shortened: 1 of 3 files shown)


SagarSDagdu (approved):

tbarbugli (commented):
can you explain why `cancelBackgroundTaskIfNeeded` is not needed? the PR description is not very clear

Sources/StreamChat/ChatClient_Tests.swift
```diff
@@ -893,43 +893,6 @@ class ChatClient_Tests: StressTestCase {
...
-        XCTAssertEqual(testEnv.backgroundTaskScheduler?.beginBackgroundTask_called, true)
-        
-        // Simulate background task execution finish
-        testEnv.backgroundTaskScheduler?.beginBackgroundTask_expirationHandler?()
-        
-        // Assert that `disconnect` is called
-        XCTAssertEqual(testEnv.clientUpdater?.disconnect_called, true)
-        
-        // Assert that background task is cancelled
-        XCTAssertEqual(testEnv.backgroundTaskScheduler?.endBackgroundTask_called, true)
```
evsaev, line 930: hm, shouldn't we only get rid of last 2 lines (where we check that background task is cancelled)? 🤔

evsaev (dismissed):
LGTM 👍

Sources/StreamChat/ChatClient_Tests.swift
dmigach: Yes, you're right, it was accidental

evsaev (approved):
🚢

nuno-vieira:
@dmigach Changelog please 🙏

PR #1401, opened 2021-08-25
Simplify the way `Components.default` is stored

evsaev:
Initially, `Components` were generic over extra data so we couldn't have stored type property. 
It's not generic anymore so we can simplify it.


nuno-vieira (approved):
LGTM

PR #1412, opened 2021-08-30
Improve Push Support

tbarbugli:
- Add a notification service extension to demo app
- Added `NotificationService` an helper class to handle push notifications 
- Added support for App Groups (needed if you want to share core data between app and service extension)
- Added `setToken` to `ChatClient` to use APIs without creating a connection (needed for the service ext.)
- `addDevice` will not call the API if the device is already registered 
- Push docs


docusaurus/docs/iOS/guides/push-notifications.md
```diff
@@ -2,135 +2,382 @@
...
             }
         }
-    return true
 }
 ```
-### Adding devices
 
-In case if the device was successfully registered for receiving push notifications, your app will receive an `application(_:didRegisterForRemoteNotificationsWithDeviceToken)` callback. At this point you should send an obtained device token to StreamChat's backend to finish the device's registration for push notifications.
+:::note
+Push notifications can be tricky to setup correctly, make sure to check for errors and settings on the [Dashboard](https://getstream.io/dashboard/)
```
nikomancy, line 63: Add links to our content on Push Logging and 

- Push FAQ : https://getstream.io/chat/docs/ios-swift/push_-_common_issues_and_faq/?language=swift
- https://getstream.io/chat/docs/ios-swift/push_logs/?language=swift

DemoAppPush/NotificationService.swift
```diff
@@ -0,0 +1,115 @@
...
+        }
+
+        var config = ChatClientConfig(apiKey: .init(apiKeyString))
+        config.isLocalStorageEnabled = true
+        config.applicationGroupIdentifier = applicationGroupIdentifier
+
+        let client = ChatClient(config: config)
+        client.setToken(token: Token(stringLiteral: userCredentials.token))
+
+        let chatHandler = ChatRemoteNotificationHandler(client: client, content: content)
```
b-onc, line 85: IMHO it'd be better if this entity was able to handle any content, so it wouldn't accept `content` in `init` but it'd accept `content` in `handleNotification` as `handleNotification(content:)`. Then we'd expose the default handler from `ChatClient` as `client.remoteNotificationHandler`

Sources/StreamChat/APIClient/ChatRemoteNotificationHandler.swift
```diff
@@ -0,0 +1,199 @@
...
+            session.currentUser?.lastReceivedEventDate = lastReceivedEventDate
+            completion()
+        }
+    }
+
+    private func shouldSyncWithAPI(since: Date, now: Date = .init()) -> Bool {
+        now.timeIntervalSince(since) > syncCooldown
+    }
+
+    private func syncChannel(completion: @escaping () -> Void) {
```
b-onc, line 93:
```suggestion
    private func syncChannels(completion: @escaping () -> Void) {
```

Sources/StreamChat/APIClient/ChatRemoteNotificationHandler.swift
```diff
@@ -0,0 +1,199 @@
...
+        database = client.databaseContainer
+    }
+
+    private func obtainLastSyncDate(completion: @escaping (Date?) -> Void) {
+        database.backgroundReadOnlyContext.perform {
+            completion(self.database.backgroundReadOnlyContext.currentUser?.lastReceivedEventDate)
+        }
+    }
+
+    private func bumpLastSyncData(lastReceivedEventDate: Date, completion: @escaping () -> Void) {
```
b-onc, line 82:
```suggestion
    private func bumpLastSyncDate(lastReceivedEventDate: Date, completion: @escaping () -> Void) {
```

DemoAppPush/NotificationService.swift
```diff
@@ -0,0 +1,115 @@
...
+            }
+            return
+        }
+        if let attachment = message.imageAttachments.first {
+            addAttachments(url: attachment.imageURL, content: content) {
+                completion($0)
+            }
+            return
+        }
+    }
```
nuno-vieira, line 61: This also seems like helpers functions for the `UNNotificationServiceExtension` right? Can we move them to the bottom then? And I guess make them private? Are they used outside of the NotificationService? 

Another option is to extract these functions into a different component, it would allow testing `NotificationService` easier since these helper functions would be mockable. Maybe introducing a `NotificationContentService`.

Sources/StreamChat/APIClient/ChatRemoteNotificationHandler.swift
```diff
@@ -0,0 +1,199 @@
...
+                completion()
+                return
+            }
+
+            let endpoint: Endpoint<MissingEventsPayload> = .missingEvents(
+                since: lastSyncAt,
+                cids: cids
+            )
+
+            self.client.apiClient.request(endpoint: endpoint) {
```
b-onc, line 125: We're doing an API request inside `backgroundReadOnlyContext`'s queue... We call the completion block of `obtainLastSyncDate` in this queue and it continues as so. We should switch queues.

Sources/StreamChat/APIClient/ChatRemoteNotificationHandler.swift
```diff
@@ -0,0 +1,199 @@
...
+    private func obtainLastSyncDate(completion: @escaping (Date?) -> Void) {
+        database.backgroundReadOnlyContext.perform {
+            completion(self.database.backgroundReadOnlyContext.currentUser?.lastReceivedEventDate)
+        }
+    }
+
+    private func bumpLastSyncData(lastReceivedEventDate: Date, completion: @escaping () -> Void) {
+        database.write { session in
+            session.currentUser?.lastReceivedEventDate = lastReceivedEventDate
+            completion()
```
b-onc, line 85: there's a `completion` parameter of `database.write`, the `completion` should be called there

Sources/StreamChat/APIClient/ChatRemoteNotificationHandler.swift
```diff
@@ -0,0 +1,199 @@
...
+
+    public init(client: ChatClient, content: UNNotificationContent) {
+        self.client = client
+        self.content = content
+        database = client.databaseContainer
+    }
+
+    private func obtainLastSyncDate(completion: @escaping (Date?) -> Void) {
+        database.backgroundReadOnlyContext.perform {
+            completion(self.database.backgroundReadOnlyContext.currentUser?.lastReceivedEventDate)
```
b-onc, line 78: this completion shouldn't be called inside `perform`, see comment below

Sources/StreamChat/APIClient/ChatRemoteNotificationHandler.swift
```diff
@@ -0,0 +1,199 @@
...
+            messageId = nil
+        }
+    }
+}
+
+public class ChatRemoteNotificationHandler {
+    var client: ChatClient
+    var content: UNNotificationContent
+    let chatCategoryIdentifiers: Set<String> = ["stream.chat", "MESSAGE_NEW"]
+    let syncCooldown: TimeInterval = 6.0
```
b-onc, line 67: Where does 6 come from? Can we comment it?

Sources/StreamChat/APIClient/ChatRemoteNotificationHandler.swift
```diff
@@ -0,0 +1,199 @@
...
+            let request = ChannelDTO.allChannelsFetchRequest
+            request.fetchLimit = 1000
+            request.propertiesToFetch = ["cid"]
+
+            guard let results = try? self.database.backgroundReadOnlyContext.fetch(request) else {
+                completion()
+                return
+            }
+
+            let cids = results.map(\.cid).compactMap { try? ChannelId(cid: $0) }
```
b-onc, line 114: There's 2 `map`s here, it'd be more efficient to do
```swift
results.compactMap { try? ChannelId(cid: $0.cid) }
```

nuno-vieira (commented):
Overall really looks good ✅  just added minor comments

DemoAppPush/NotificationService.swift
```diff
@@ -0,0 +1,115 @@
...
+            default:
+                content.title = "You received an update to one conversation"
+                contentHandler(content)
+            }
+        }
+        
+        if !chatNotification {
+            /// this was not a notification from Stream Chat
+            /// perform any other transformation to the notification if needed
+        }
```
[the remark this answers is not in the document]
tbarbugli: its a placeholder in case you receive remote notifications from other services

DemoAppPush/NotificationService.swift
tbarbugli: this is part of the demo, I did not want to expand the scope of the SDK and add more helpers. This serves as a decent example / starting point

Sources/StreamChat/APIClient/ChatRemoteNotificationHandler.swift
tbarbugli: sorry not sure what you mean

tbarbugli: ideally we did everything sync since this code is meant to run in a service extension. any idea?

tbarbugli: @b-onc ping on this, afaict this does not matter given the environment the code runs (ext). let me know :)

PR #1415, opened 2021-09-01
Remove Group By from DTO

tbarbugli:
Rewrite `attachmentCounts` to not use Core Data (previous approach with GROUP BY did not work on iOS 12 with in memory storage)

### DIFF
--- a/Sources/StreamChat/Models/ChatMessage.swift
+++ b/Sources/StreamChat/Models/ChatMessage.swift
@@ -94,10 +94,12 @@ public struct ChatMessage {
     @CoreDataLazy internal var _attachments: [AnyChatMessageAttachment]
 
     /// The overall attachment count by attachment type.
-    public var attachmentCounts: [AttachmentType: Int] { _attachmentCounts }
+    public var attachmentCounts: [AttachmentType: Int] {
+        _attachments.reduce(into: [:]) { counts, attachment in
+            counts[attachment.type] = (counts[attachment.type] ?? 0) + 1
+        }
+    }
 
-    @CoreDataLazy internal var _attachmentCounts: [AttachmentType: Int]
-        
     /// A list of latest 25 replies to this message.
     ///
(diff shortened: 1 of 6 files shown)


Sources/StreamChatUI/ChatMessageList/Attachments/AttachmentViewCatalog.swift
```diff
@@ -14,8 +14,7 @@ open class AttachmentViewCatalog {
         message: ChatMessage,
         components: Components
     ) -> AttachmentViewInjector.Type? {
-        // Do not show attachments for deleted messages
-        guard !message.isDeleted else { return nil }
+        if message.isDeleted { return nil }
```
evsaev, line 17: nit: in general, I wouldn't include changes that are not directly connected to the topic. In this case it's fine but i it would complicate review process for large PRs.. 

Regarding the change, when you see `guard` you expect to see early return OR throwing an error while `ifs` do not trigger `early return` expectation and are good for checks inside a function body.

Sources/StreamChat/Models/ChatMessage.swift
```diff
@@ -168,7 +170,7 @@ public struct ChatMessage {
         currentUserReactions: @escaping () -> Set<ChatMessageReaction>,
         isSentByCurrentUser: Bool,
         pinDetails: MessagePinDetails?,
-        attachmentCounts: @escaping () -> [AttachmentType: Int],
+        attachmentCounts: [AttachmentType: Int],
```
evsaev, line 173: it seems `attachmentCounts` can be excluded from `init` params since we transformed it to computed property 👍

Sources/StreamChatUI/ChatMessageList/Attachments/AttachmentViewCatalog.swift
tbarbugli: - guard not deleted else -> exit
- if deleted -> exit

second form looks much more readable

PR #1416, opened 2021-09-01
[CIS 1155] Add flag to disable mentions + fix disable commands

nuno-vieira:
- Adds a new flag `isMentionsEnabled` to make it easier to disable the user mentions in the `ComposerVC`
- Fixes an issue where the `isCommandsEnabled` didn't disable the typing commands, it was only hiding the commands button.


evsaev (approved):
LGTM 👍

CHANGELOG.md
```diff
@@ -31,6 +31,7 @@ A new `ChatChannelVC` is introduced that represents the old `ChatMessageListVC`,
 ### ✅ Added
 - For non-DM channels, the avatar is now shown as a combination of the avatars of the last active members of the channel [#1344](https://github.com/GetStream/stream-chat-swift/pull/1344)
 - New DateFormatter methods `rfc3339Date` and `rfc3339DateString` [#1403](https://github.com/GetStream/stream-chat-swift/pull/1403)
+- Add a new `isMentionsEnabled` flag to make it easier to disable the user mentions in the ComposerVC [#1416](https://github.com/GetStream/stream-chat-swift/pull/1416)
```
b-onc, line 34: We should name the flag `shouldShowMentions` for SDK parity

Sources/StreamChatUI/Composer/ComposerVC.swift
```diff
@@ -163,6 +163,11 @@ open class ComposerVC: _ViewController,
         channelConfig?.commands.isEmpty == false
     }
 
+    /// A Boolean value indicating whether the user mentions are enabled.
+    open var isMentionsEnabled: Bool {
+        true
+    }
```
b-onc, line 169: What if users just want to assign this, in their MessageList subclass? IMHO better to have this get/set and not overridable

b-onc (approved):
2 small things, LGTM

CHANGELOG.md
nuno-vieira: We already have `ComposerVC.isCommandsEnabled` and `ComposerVC.isAttachmentsEnabled` so it makes sense for this one to have the same pattern. Also, for feature flags it is way more common to use the `is...Enabled` pattern than `should...`.

Sources/StreamChatUI/Composer/ComposerVC.swift
nuno-vieira: As mentioned above the `ComposerVC.isCommandsEnabled` and `ComposerVC.isAttachmentsEnabled` are also overridable properties, so it makes sense for this one to be as well. These two are really useful to be computed properties because by default they read from the feature flags on the channel config. So, it is way to override that behaviour. Either way, for consistency at least on the `ComposerVC` component we should keep it like this.

PR #1417, opened 2021-09-01
[CIS-1145] Unit Tests For Filter Decoding

kojiba:
[-] Replaced deprecated throws of NSError
    with correct DecodingErrors
[-] Improved Warnings in JSONDecoder and Encoder Tests
[-] Moved reusable test logic to FilterTestExtensions_Tests
[-] Renamed TestScope to FilterTestScope
[-] Updated Changelog according to changes
[-] Updated Package.swift with new tests ignored

Test plan: 
- run all unit tests succeed
- quickly run through demo app, see everything work as usual


CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ A new `ChatChannelVC` is introduced that represents the old `ChatMessageListVC`,
 
 ### 🔄 Changed
 - Nuke dependency was updated to v10 [#1405](https://github.com/GetStream/stream-chat-swift/pull/1405)
+- Instead of deprecated `NSError()` on a filter decoding now correct DecodingError thrown [#1417](https://github.com/GetStream/stream-chat-swift/pull/1417/)
```
tbarbugli, line 30: we dont have to update changelog for something like this

Package.swift
```diff
@@ -191,7 +191,9 @@ var streamChatSourcesExcluded: [String] { [
     "Workers/ChannelUpdater_Mock.swift",
     "Workers/EventNotificationCenter_Tests.swift",
     "Workers/ChatClientUpdater_Tests.swift",
+    "Query/FilterTestExtensions_Tests.swift",
```
tbarbugli, line 194: just to be sure, you did not edit this manually, right?

kojiba: Sure, runned script from Scripts

Sources/StreamChat/Query/FilterDecoding_Tests.swift
```diff
@@ -0,0 +1,221 @@
+//
+// Copyright © 2021 Stream.io Inc. All rights reserved.
+//
+
+@testable import StreamChat
+import XCTest
+
+class FilterDecoding_Tests: XCTestCase {
```
tbarbugli, line 8: please make sure to test that encoding and decoding are symmetrical so that we dont end up with encoded data failing at decoding

it must not be possible to encode data that cannot be decoded

Sources/StreamChat/Query/FilterDecoding_Tests.swift
```diff
@@ -0,0 +1,221 @@
...
+    }
+
+    func testFilterRightSideDecodingThrowExceptionWithoutOperationKey() {
+        let filterJson = #"{ "test_key": { "eq":"test_value_1" } }"#
+
+        checkDecodingFilterThrowException(json: filterJson)
+    }
+
+    func testFilterRightSideDecodingThrowExceptionOnNonExistingOperation() {
+        let filterJson = #"{ "test_key": { "$myspecialOperation":"test_value_1" } }"#
```
tbarbugli, line 35: Let's be 100% sure that it's impossible for the filter to end up encoded this way (ie. if we have something like .customOperator("myspecialOperation") we are in trouble)

tbarbugli (commented):
see my comments

Package.swift
b-onc: You shouldn't run this script manually. Please run `./bootstrap.sh`, it'll install this script as a commit hook, so it runs automatically.

Sources/StreamChat/Query/FilterDecoding_Tests.swift
kojiba: Yes, for sure, but for decoder, we should test this corner case to be thrown exception and not crash

Sources/StreamChat/Query/Filter.swift
```diff
@@ -368,10 +372,22 @@ private struct FilterRightSide: Decodable {
...
+            throw DecodingError.dataCorruptedError(
+                forKey: container.allKeys.last ?? ArbitraryKey(""),
+                in: container,
+                debugDescription: "FilterRightSide keys count should be only 1"
+            )
+        }
         
         let key = container.allKeys.first!
-        guard key.stringValue.hasPrefix("$") else { throw NSError() }
+        guard key.stringValue.hasPrefix("$") else {
```
evsaev, line 384:
```suggestion
        guard let operator = container.allKeys.first?.stringValue, operator.hasPrefix("$") else { ... }
        self.operator = operator
```

evsaev (approved):
👏 🚢

nuno-vieira (approved):
LGTM ✅

PR #1419, opened 2021-09-02
[CIS-1152] Fix connection recovery

evsaev:
This PR changes the way we sync when connection is back.

When internet connection is back it's not enough to call `/sync` but we should also refetch existed channel queries because there can be channels user was invited to/became a member of while being offline. Since we don't know anything about these new channels, they can be inserted in any position of a locally existed channel list, we do a refetch. We also have to refetch a query to start `watching` it to get channel events.

**Channel list**

Initially, connection recovery logic lived in `ConnectionRecoveryUpdater` but it was not correct: the `ChannelListQueryDTO` we get from DB does not contain `sort` option so if it was different from default one we did sync a different query. The only place where we have correct `ChannelListQuery` is `ChannelListController` so now this object is responsible for syncing the query it manages. 

Here is how channel list query sync happens:
1. connection comes back
2. we refetch the 1st page of a query (first 20 channels) and we start watching them because of [.watch] option included
4. we unlink all locally existed channels from a local query (so only 1st recently fetched page of channels is displayed)
5. we unlink all messages from the channels we didn't get with the first page because we are no longer watching them so messages of local channels can be outdated

**Active channel**

While for channel list we can refetch the query, we should carefully handle the channel that is being read when connection comes back. For active channels we call `/sync` with a single `cid` to get events that we've missed. Since we ask for single channel events, chances to exceed 1k event limit are low. To track active channels `isBeingRead` flag is introduced which is set to `true` when `ChannelController` is synchronised and set back to `false` when `ChannelController` is deallocated. 

If `/sync` is successful we start `watching` the channel. If `/sync` fails we reset the channel history (by unlinking all messages from this channel) and refetch the first page of messages.

### DIFF
--- a/Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift
+++ b/Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift
@@ -42,12 +42,6 @@ public class ChatChannelListController: DataController, DelegateCallable, DataSt
             client.apiClient
         )
     
-    private var connectionObserver: EventObserver?
-    private let requestedChannelsLimit = 25
-
-    /// A Boolean value that returns wether pagination is finished
-    public private(set) var hasLoadedAllPreviousChannels: Bool = false
-
     /// A type-erased delegate.
     var multicastDelegate: MulticastDelegate<AnyChannelListControllerDelegate> = .init() {
         didSet {
@@ -105,55 +99,13 @@ public class ChatChannelListController: DataController, DelegateCallable, DataSt
     
     override public func synchronize(_ completion: ((_ error: Error?) -> Void)? = nil) {
         startChannelListObserverIfNeeded()
-        setupEventObserversIfNeeded(completion: completion)
-    }
-    
-    private func setupEventObserversIfNeeded(completion: ((_ error: Error?) -> Void)? = nil) {
-        guard !client.config.isLocalStorageEnabled else {
-            return updateChannelList(trumpExistingChannels: false, completion)
-        }
-        
-        updateChannelList(trumpExistingChannels: channels.count > requestedChannelsLimit) { [weak self] error in
-            completion?(error)
-            
-            guard let self = self else { return }
-            self.connectionObserver = nil
-            // We can't setup event observers in connectionless mode
-            guard let webSocketClient = self.client.webSocketClient else { return }
-            let center = webSocketClient.eventNotificationCenter
-            // We setup a `Connected` Event observer so every time we're connected,
-            // we refresh the channel list
-            self.connectionObserver = EventObserver(
-                notificationCenter: center,
-                transform: { $0 as? ConnectionStatusUpdated },
-                callback: { [unowned self] in
-                    switch $0.webSocketConnectionState {
-                    case .connected:
-                        self.updateChannelList(trumpExistingChannels: self.channels.count > self.requestedChannelsLimit)
-                    default:
-                        break
-                    }
-                }
-            )
-        }
-    }
-    
-    private func updateChannelList(
-        trumpExistingChannels: Bool,
-        _ completion: ((_ error: Error?) -> Void)? = nil
-    ) {
-        worker.update(
-            channelListQuery: query,
-            trumpExistingChannels: trumpExistingChannels
-        ) { result in
-            switch result {
-            case .success:
-                self.state = .remoteDataFetched
-                self.callback { completion?(nil) }
-            case let .failure(error):
+        queryChannels(offset: 0) { error in
+            if let error = error {
                 self.state = .remoteDataFetchFailed(ClientError(with: error))
-                self.callback { completion?(error) }
+            } else {
+                self.state = .remoteDataFetched
             }
+            self.callback { completion?(error) }
         }
     }
     
@@ -196,22 +148,8 @@ public class ChatChannelListController: DataController, DelegateCallable, DataSt
         limit: Int? = nil,
         completion: ((Error?) -> Void)? = nil
     ) {
-        if hasLoadedAllPreviousChannels {
-            completion?(nil)
-            return
-        }
-
-        let limit = limit ?? requestedChannelsLimit
-        var updatedQuery = query
-        updatedQuery.pagination = Pagination(pageSize: limit, offset: channels.count)
-        worker.update(channelListQuery: updatedQuery) { result in
-            switch result {
-            case let .success(payload):
-                self.hasLoadedAllPreviousChannels = payload.channels.count < limit
-                self.callback { completion?(nil) }
-            case let .failure(error):
-                self.callback { completion?(error) }
-            }
+        queryChannels(limit: limit) { error in
+            self.callback { completion?(error) }
         }
     }
 
@@ -223,6 +161,30 @@ public class ChatChannelListController: DataController, DelegateCallable, DataSt
         worker.markAllRead { error in
             self.callback {
                 completion?(error)
+            }
+        }
+    }
+    
+    private func queryChannels(
+        offset: Int? = nil,
+        limit: Int? = nil,
+        completion: @escaping (Error?) -> Void
+    ) {
+        var updatedQuery = query
+        updatedQuery.pagination = Pagination(
+            pageSize: limit ?? query.pagination.pageSize,
+            offset: offset ?? channels.count
+        )
+        
+        worker.update(
+            channelListQuery: updatedQuery,
+            trumpExistingChannels: updatedQuery.pagination.offset == 0
+        ) {
+            switch $0 {
+            case .success:
+                completion(nil)
+            case let .failure(error):
+                completion(error)
             }
         }
     }
--- a/Sources/StreamChat/Database/DTOs/ChannelDTO.swift
+++ b/Sources/StreamChat/Database/DTOs/ChannelDTO.swift
@@ -251,7 +251,7 @@ extension ChannelDTO {
         let sortDescriptors = query.sort.compactMap { $0.key.sortDescriptor(isAscending: $0.isAscending) }
         request.sortDescriptors = sortDescriptors.isEmpty ? [ChannelListSortingKey.defaultSortDescriptor] : sortDescriptors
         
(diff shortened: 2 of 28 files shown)


Sources/StreamChat/Database/DTOs/ChannelDTO.swift
```diff
@@ -19,6 +23,8 @@ class ChannelDTO: NSManagedObject {
     @NSManaged var defaultSortingAt: Date
     @NSManaged var updatedAt: Date
     @NSManaged var lastMessageAt: Date?
+    
+    @NSManaged var isBeingRead: Bool
```
evsaev, line 27: I wonder if we can avoid introducing a local flag but instead look for current user in `channel.watchers`?

cc @b-onc @tbarbugli

Sources/StreamChat/Query/Pagination.swift
```diff
@@ -49,28 +49,37 @@ public struct MessagesPagination: Encodable, Equatable {
     let pageSize: Int?
     /// Parameter for pagination.
     let parameter: PaginationParameter?
+    /// An offset.
+    let offset: Int?
```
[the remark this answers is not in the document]
b-onc: It's supported, we made the decision to use `greaterThan/lessThan messageId` instead of `offset` to keep the API simpler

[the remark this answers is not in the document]
evsaev: I see 👍 We need `offset` to refetch the first page of channel messages. `greaterThan/lessThan` and `offset` seem to be mutually exclusive so I'll address it.

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
evsaev: I've just realised there's another solution. What if we keep track of all created `channelList/channel` controllers in `ChatClient`? 🤔 then, we don't need to set/reset any local flag on `ChannelDTO` but can get channels that are being read simply by iterating over `[Weak<ChannelController>].map(\.cid)`

evsaev: this way, the sync can stay in `ConnectionRecoveryUpdater`

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
```diff
@@ -7,7 +7,11 @@ import Foundation
 
 @objc(ChannelDTO)
 class ChannelDTO: NSManagedObject {
-    @NSManaged var cid: String
+    @NSManaged private var cid: String
+    var channelId: ChannelId {
+        try! .init(cid: cid)
```
kojiba, line 12: I would recommend to use try? with default variable like 

`ChannelId(cid: cid) ?? .unknown`  or `.invalid`

Sources/StreamChat/Workers/ChannelListUpdater.swift
```diff
@@ -6,6 +6,21 @@ import CoreData
 
 /// Makes a channels query call to the backend and updates the local storage with the results.
 class ChannelListUpdater: Worker {
+    /// Makes a channels query call to the backend and callas completion with results.
```
nuno-vieira: calls*

Sources/StreamChat/Workers/Background/NewChannelQueryUpdater.swift
```diff
@@ -60,58 +50,82 @@ final class NewChannelQueryUpdater: Worker {
...
+                self?.save(channel, andLinkTo: query)
+            case let .failure(error):
+                log.error("Failed to check if query should include new channel: \(error)")
             }
+        }
+    }
+    
+    private func fetchExistedQueries(_ completion: @escaping ([ChannelListQuery]) -> Void) {
+        let context = database.backgroundReadOnlyContext
+        context.perform {
```
nuno-vieira, line 111: What do you mean by "existed"?

Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift
```diff
@@ -42,12 +42,6 @@ public class ChatChannelListController: DataController, DelegateCallable, DataSt
             client.apiClient
         )
     
-    private var connectionObserver: EventObserver?
-    private let requestedChannelsLimit = 25
-
-    /// A Boolean value that returns wether pagination is finished
-    public private(set) var hasLoadedAllPreviousChannels: Bool = false
```
nuno-vieira, line 49: Why did we remove this? We still need this right? How do we know the pagination has finished?

Sources/StreamChat/Workers/Background/ConnectionRecoveryUpdater.swift
```diff
@@ -101,81 +101,141 @@ class ConnectionRecoveryUpdater: EventWorker {
...
-
-    private func sync(completion: @escaping () -> Void) {
-        guard let lastSyncedAt = lastSyncedAt else { return }
+    
+    private func fetchFirstPage(
+        of queries: [ChannelListQuery],
+        completion: @escaping ([String: Result<ChannelListPayload, Error>]) -> Void
+    ) {
+        let group = DispatchGroup()
+        let semaphore = DispatchSemaphore(value: 1)
```
nuno-vieira, line 124: Why do we need both a group and a semaphore? The group should be enough 🤔

evsaev: A group is for tracking when all tasks are done, a `semaphore` makes sure we don't write `results` dict from multiple threads at the same time 👍

nuno-vieira: Ah I see, makes sense 👍 Thanks!

Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift
evsaev: Yep, good question 👍 

The use-case is: a user who loaded all the channels (100) goes offline and when connection comes back we refetch the first page (0..<20) of the channel query but user scrolls down the next page (20..<40) won't be loaded because of  `hasLoadedAllPreviousChannels == true`

That's not a final solution but a temp fix 🙂 I plan to reset the flag if we see `.remove` change reported by database observer but haven't yet addressed this

Sources/StreamChat/Workers/Background/NewChannelQueryUpdater.swift
evsaev: i meant `fetchLocalQueries`. will do the rename 👍

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
b-onc, line 12: Do we really need this? ChannelId is our decoded value, IMHO it's better to keep it the way it was (using `cid` and `channelId.rawValue`)

Sources/StreamChat/Workers/Background/DatabaseCleanupUpdater.swift
```diff
@@ -31,48 +31,91 @@ class DatabaseCleanupUpdater: Worker {
...
+            self?.fetchFirstPage(of: localQueries) { remoteQueries in
+                self?.database.write { session in
+                    for (queryHash, queryResult) in remoteQueries {
+                        switch queryResult {
+                        case let .success(firstPage):
+                            log.debug("Did load first page of channels for \(queryHash)")
+                                                        
+                            // Load local query
+                            guard let queryDTO = session.channelListQuery(queryHash: queryHash) else {
+                                assertionFailure("Channel list query no longer exists: \(queryHash)")
```
b-onc, line 47:
```suggestion
                                log.error("Channel list query no longer exists: \(queryHash)")
```

Sources/StreamChat/Workers/Background/DatabaseCleanupUpdater.swift
```diff
@@ -31,48 +31,91 @@ class DatabaseCleanupUpdater: Worker {
...
+        let group = DispatchGroup()
+        let semaphore = DispatchSemaphore(value: 1)
+        
+        var results: [String: Result<ChannelListPayload, Error>] = [:]
+        for query in queries {
+            group.enter()
+            
+            log.debug("Will fetch first page of channels for \(query)")
+            channelListUpdater.fetch(query) {
+                semaphore.wait()
```
b-onc, line 105: Instead of using a semaphore, we can create a temp serial queue and use it to access:
```
let queue = DispatchQueue("dict-access")
....
queue.async { results[query.queryHash] = queryHash }
....
```

Sources/StreamChat/Workers/Background/ConnectionRecoveryUpdater.swift
```diff
@@ -95,87 +95,61 @@ class ConnectionRecoveryUpdater: EventWorker {
...
                 
+                self?.eventNotificationCenter.process(payload.eventPayloads) {
+                    completion(nil)
+                }
             case let .failure(error):
-                log.info(
-                    """
-                    Backend couldn't handle replaying missing events - there was too many (>1000) events to replay. \
-                    Cleaning local channels data and refetching it from scratch
-                    """
```
b-onc, line 148: So we don't have this mechanism anymore? on error, we just keep the channel data around?

b-onc (commented):
First round of questions

Sources/StreamChat/Workers/Background/DatabaseCleanupUpdater.swift
kojiba: for some async operations we could carry in QueuedExecutor 
from my draft PR on Members Pagination

kojiba: for syncronization of dictionary, I would also recommend @b-onc  approach, 
cause serial queues is general mechanism recommended by Apple to handle concurrency syncronizations

Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift
nuno-vieira: I think we should try to fix this on this PR because then we will have a regression :/

PR #1420, opened 2021-09-02
[CIS-1158] Fix filters with wrong Date encoding strategy

nuno-vieira:
The filters didn't have the correct Date encoding strategy because Filters use`AnyEncodable` heavily and the implementation of `AnyEncodable` was not preserving the encoding strategies. To know more about why this was happening you can read this swift forum thread: https://forums.swift.org/t/how-to-encode-objects-of-unknown-type


evsaev (approved):
🚢

dmigach (approved):

PR #1421, opened 2021-09-02
[CIS-1159] Provide different Objc name for `InputTextView`

nuno-vieira:
A customer that is using MessageKit with Stream and is using Objc could not compile the app since MessageKit also contains an `InputTextView`. To fix this, we can provide a different name for Objc only.


evsaev (approved):
🚢

PR #1422, opened 2021-09-03
CIS-1160 Better autocomplete for mentions

tbarbugli:
- Support `options.minimumRequiredCharacters = 0` so that suggestions are presented after the @
- Improve search result ordering, matches with same distance from search are not sorted consistently (by id)

### DIFF
--- a/Sources/StreamChatUI/Composer/ComposerVC.swift
+++ b/Sources/StreamChatUI/Composer/ComposerVC.swift
@@ -823,15 +823,11 @@ func searchUsers(_ users: [ChatUser], by searchInput: String, excludingId: Strin
     let matchingUsers = users.filter { $0.id != excludingId }
         .filter { searchInput == "" || $0.id.contains(searchInput) || (normalize($0.name ?? "").contains(searchInput)) }
 
-    let uniqueUsers = matchingUsers.reduce(into: [String: ChatUser]()) {
-        $0[$1.id] = $1
-    }
-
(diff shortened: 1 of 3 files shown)


Sources/StreamChatUI/Composer/ComposerVC.swift
```diff
@@ -821,7 +821,7 @@ func searchUsers(_ users: [ChatUser], by searchInput: String, excludingId: Strin
     let searchInput = normalize(searchInput)
 
     let matchingUsers = users.filter { $0.id != excludingId }
-        .filter { $0.id.contains(searchInput) || (normalize($0.name ?? "").contains(searchInput)) }
+        .filter { searchInput == "" || $0.id.contains(searchInput) || (normalize($0.name ?? "").contains(searchInput)) }
 
     let uniqueUsers = matchingUsers.reduce(into: [String: ChatUser]()) {
```
nuno-vieira, line 826: How not `Set(matchingUsers)` ? And then `Array(uniqueUsers).sorted...`, The reduce seems a bit unnecessary I guess?

nuno-vieira (approved):
LGTM ✅  , just a little nit comment, not a blocker

Sources/StreamChatUI/Composer/ComposerVC.swift
tbarbugli: good point, ChatUser is indeed hashable :)

nuno-vieira (approved):

PR #1425, opened 2021-09-03
Make `ChatMessageReactionData.init` public

nuno-vieira:
Makes the init public, which is required by a client.


kojiba (approved):
LGTM

PR #1430, opened 2021-09-06
[CIS-1165] Fix last event date being overriden by `HealthCheckEvent`

evsaev:
This PR changes the way we update local data on reconnect. Instead of tracking `lastReceivedEventDate` and bumping it each time new event is observed we use `lastSyncedAt` date which is bumped depending on `/sync` and queries refetch results.

**Case 1**
If `/sync` succeeds and returns events, channel list queries are re-fetched considering all locally existed channels as synced.
If re-fetch succeeds, `lastSyncedAt` is set to most recent event timestamp (received from `/sync`).
If re-fetch fails, `lastSyncedAt` stays the same.

**Case 2**
If `/sync` succeeds and returns zero events, channel list queries are re-fetched considering all locally existed channels as synced.
If re-fetch succeeds, `lastSyncedAt` is set to current date.
If re-fetch fails, `lastSyncedAt` stays the same.

**Case 3**
If `/sync` fails with error, channel list queries are re-fetched considering all locally existed channels as out of sync.
If re-fetch succeeds, `lastSyncedAt` is set to current date otherwise.
If re-fetch fails, `lastSyncedAt` stays the same.

*** When session begins, we receive all events from previous session + those we've missed (this is needed to avoid event gaps because of web-socket connection not being stable). We should save events to db to know which events we've never seen and should be saved/published.

### DIFF
--- a/Sources/StreamChat/APIClient/ChatRemoteNotificationHandler.swift
+++ b/Sources/StreamChat/APIClient/ChatRemoteNotificationHandler.swift
@@ -84,16 +84,15 @@ public class ChatRemoteNotificationHandler {
     }
 
     private func obtainLastSyncDate(completion: @escaping (Date?) -> Void) {
-        var lastReceivedEventDate: Date?
-        database.viewContext.performAndWait {
-            lastReceivedEventDate = self.database.viewContext.currentUser?.lastReceivedEventDate
+        let context = database.viewContext
+        context.perform {
+            completion(context.currentUser?.lastSyncedAt)
         }
-        completion(lastReceivedEventDate)
     }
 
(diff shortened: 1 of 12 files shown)


b-onc:
Why don't we fetch the `lastReceivedEventDate` and make the `sync` call immediately after state becomes `connected`?

evsaev:
> Why don't we fetch the `lastReceivedEventDate` and make the `sync` call immediately after state becomes `connected`?

State becomes `connected` when `HealtchCheckEvent` is saved to db. At this moment `lastReceivedEventDate` is already overridden by `HealtchCheckEvent.createdAt`

PR #1431, opened 2021-09-06
CIS-1048 reactions docs

tbarbugli:


docusaurus/docs/iOS/components/reactions.md
```diff
@@ -0,0 +1,82 @@
...
+
+let customReactions: [MessageReactionType: ChatMessageReactionAppearanceType] = [
+    .init(stringLiteral: "fire"): ChatMessageReactionAppearance(
+        smallIcon: reactionFireSmall,
+        largeIcon: reactionFireBig
+    ),
+    .init(stringLiteral: "wave"): ChatMessageReactionAppearance(
+        smallIcon: reactionWaveSmall,
+        largeIcon: reactionWaveBig
+    )
```
nuno-vieira, line 42:
```suggestion
    "fire": ChatMessageReactionAppearance(
        smallIcon: reactionFireSmall,
        largeIcon: reactionFireBig
    ),
    "wave": ChatMessageReactionAppearance(
        smallIcon: reactionWaveSmall,
        largeIcon: reactionWaveBig
    )
```
Whenever you see you are using "stringLiteral" as an argument in an initialiser, that is a bad sign. This initialiser comes from `ExpressibleByStringLiteral`, the goal of this is too initialise a type from just a string, but you need to help the compiler understand what type you want. In this case you have `customReactions: [MessageReactionType: ...]` so you only need the string to initialise it. 

Another example:
**This is wrong:**
`let token = Token(stringLiteral: "....")`
**This is good:**
`let token: Token = "..."`

To sum up, we should not use the string literal initialiser ourselves, this is for the compiler to use it. In case the value comes from a property or something else, then it is better to use for example in this case the rawValue initialiser:
`let token = Token(rawValue: remoteConfig.token)`

docusaurus/docs/iOS/components/reactions.md
```diff
@@ -0,0 +1,82 @@
...
+
+### ChatMessageReactionsView.ItemView
+
+This component renders the single message reaction.
+
+```swift
+class CustomChatMessageReactionsItemView: ChatMessageReactionsView.ItemView {}
+
+Components.default.reactionItemView = CustomChatMessageReactionsItemView.self
+```
```
nuno-vieira, line 82: It would be nice to put concrete examples here with some screenshots. Right now this document doesn't have any images, and normally that makes it harder to read and also visualise what we are doing.

nuno-vieira (changes requested):
Overall I really like the structure of the docs, but it feels it needs some concrete examples 🤔

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
```diff
@@ -112,7 +112,7 @@ open class ChatMessageContentView: _View, ThemeProvider {
 
     /// Shows the bubble around message reactions.
     /// Exists if `layout(options: MessageLayoutOptions)` was invoked with the options containing `.reactions`.
-    public private(set) var reactionsBubbleView: ChatReactionsBubbleView?
+    public private(set) var reactionsBubbleView: ChatReactionBubbleBaseView?
```
nuno-vieira, line 115: This is a breaking change right? 🤔

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatReactionsBubbleView.swift
```diff
@@ -5,12 +5,14 @@
...
-    private let tailHeight: CGFloat = 6
-
+open class ChatReactionBubbleBaseView: _View, AppearanceProvider {
     open var tailDirection: ChatThreadArrowView.Direction? {
         didSet { updateContentIfNeeded() }
     }
+}
+
+open class ChatReactionsBubbleView: ChatReactionBubbleBaseView {
+    private let tailHeight: CGFloat = 6
```
nuno-vieira, line 15: Why is the tailHeight private? And not public or even `open` ?

Sources/StreamChatUI/ChatMessageList/Reactions/ChatMessageReactionsView.swift
```diff
@@ -5,11 +5,36 @@
 import StreamChat
 import UIKit
 
+open class ChatReactionPickerReactionsView: ChatMessageReactionsView {
+    override public var reactionItemView: ChatMessageReactionItemView.Type {
+        components.reactionPickerReactionItemView
+    }
+}
+
 open class ChatMessageReactionsView: _View, ThemeProvider {
```
nuno-vieira, line 14: Should we also change the name of the file to `ChatReactionPickerReactionsView.swift`?

Sources/StreamChatUI/ChatMessageList/Reactions/ChatMessageReactionsBubbleView.swift
```diff
@@ -5,6 +5,8 @@
 import StreamChat
 import UIKit
 
+public typealias ReactionPickerBubbleView = ChatMessageReactionsBubbleView
+
 open class ChatMessageReactionsBubbleView: _View, ThemeProvider {
```
nuno-vieira, line 10: Same here, better to update the file name with the new name

Sources/StreamChatUI/ChatMessageList/Reactions/ChatMessageReactionsBubbleView+Default.swift
```diff
@@ -5,7 +5,7 @@
 import StreamChat
 import UIKit
 
-open class ChatMessageDefaultReactionsBubbleView: ChatMessageReactionsBubbleView {
+open class ChatMessageDefaultReactionsBubbleView: ReactionPickerBubbleView {
```
nuno-vieira, line 8: I know this was like this before, but it doesn't make a lot of sense the current file name, it should just be:
`ChatMessageDefaultReactionsBubbleView.swift`

nuno-vieira (commented):
Looks really  good overall, the naming is now much better 👍 Just added some questions that I have

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
tbarbugli: it should not be `ChatReactionsBubbleView ` is a subclass of `ChatReactionBubbleBaseView`

Sources/StreamChatUI/ChatMessageList/Reactions/ChatReactionPickerBubbleView+Default.swift
```diff
@@ -5,7 +5,7 @@
 import StreamChat
 import UIKit
 
-open class ChatMessageDefaultReactionsBubbleView: ChatMessageReactionsBubbleView {
+open class ChatMessageDefaultReactionsBubbleView: ChatReactionPickerBubbleView {
```
nuno-vieira, line 8: The name of this file is still weird 🤔 The name of the file should be the name of the class. And when using `+` the right side should contain the function or class name being extended. 

In this case the name of the file should be `ChatMessageDefaultReactionsBubbleView.swift`, currently is
`ChatReactionPickerBubbleView+Default.swift`. I don't see any `PickerBubbleView`, also the `+Default` should be there as well

nuno-vieira (approved):
Just a small nit comment, other then that LGTM ✅

nuno-vieira:
Also, probably we should add a changelog explaining the new names?

nuno-vieira (approved):
LGTM ✅

PR #1433, opened 2021-09-07
CIS-1166 Improve logging

tbarbugli:
- Set sample app log level to .warning
- Log error responses from API with .warning level]
- Skip useless warning message when connecting

### DIFF
--- a/DemoApp/DemoAppCoordinator.swift
+++ b/DemoApp/DemoAppCoordinator.swift
@@ -16,10 +16,6 @@ final class DemoAppCoordinator: NSObject, UNUserNotificationCenterDelegate {
     let connectionDelegate: BannerShowingConnectionDelegate
 
     init(navigationController: UINavigationController) {
-        // Since log is first touched in `BannerShowingConnectionDelegate`,
-        // we need to set log level here
-        LogConfig.level = .warning
-        
         self.navigationController = navigationController
         connectionDelegate = BannerShowingConnectionDelegate(
             showUnder: navigationController.navigationBar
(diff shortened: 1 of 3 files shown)


Sources/StreamChat/Workers/ChatClientUpdater.swift
```diff
@@ -159,7 +159,7 @@ class ChatClientUpdater {
             return
         }
 
-        guard client.connectionId != nil else {
+        if client.connectionId == nil && source == .userInitiated {
```
dmigach, line 162: I'm wondering if going past this expression for system initiated calls and calling `client.completeConnectionIdWaiters(connectionId: nil)` might cause any unpleasant side-effects, but probably it's all right

dmigach (approved):

Sources/StreamChat/Workers/ChatClientUpdater.swift
tbarbugli: you are 100% right, fixing this

DemoApp/DemoAppCoordinator.swift
```diff
@@ -72,6 +72,7 @@ final class DemoAppCoordinator: NSObject, UNUserNotificationCenterDelegate {
         // Create a token
         let token = try! Token(rawValue: userCredentials.token)
         
+        LogConfig.level = .warning
```
b-onc, line 75: This is set in line 21

b-onc (commented):
Redundant LogConfig.level setting in DemoApp
OOT, LGTM

PR #1434, opened 2021-09-08
[CIS-1162] Members pagination 

kojiba:
[-] Added BlockingExecutor with logic to return error if
    called tries to call operation when another is
    in progress
[-] Integrated BlockingExecutor to ChannelController
    and MembersController
[-] Added tests for BlockingExecutor
[-] Added atomic tests wich ten concurrent operations
[-] Updated ChannelListUpdated to return the payload
[-] Updated corresponding test and mocks
[-] Updated Package.swift

### DIFF
--- a/Sources/StreamChat/Controllers/MemberListController/MemberListController.swift
+++ b/Sources/StreamChat/Controllers/MemberListController/MemberListController.swift
@@ -21,6 +21,9 @@ extension ChatClient {
 public class ChatChannelMemberListController: DataController, DelegateCallable, DataStoreProvider {
     /// The query specifying sorting and filtering for the list of channel members.
     @Atomic public private(set) var query: ChannelMemberListQuery
+
+    /// A Boolean value that returns whether the previous members have all been loaded or not.
+    public private(set) var hasLoadedAllPreviousMembers: Bool = false
     
     /// The `ChatClient` instance this controller belongs to.
     public let client: ChatClient
@@ -53,6 +56,10 @@ public class ChatChannelMemberListController: DataController, DelegateCallable, 
     /// and expose the published values by mapping them to a read-only `AnyPublisher` type.
     @available(iOS 13, *)
     lazy var basePublishers: BasePublishers = .init(controller: self)
+
+    /// Executor to control loading of messages, only one loading in progress
+    private let membersLoadingExecutor: BlockingExecutor =
+        .init(executorTitle: "ChatChannelMemberListController.MembersLoading")
     
     private let environment: Environment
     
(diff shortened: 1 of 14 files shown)


Sources/StreamChat/Utils/QueuedExecutor.swift
```diff
@@ -0,0 +1,47 @@
...
+
+import Foundation
+
+private struct QueuedExecutorIdentifier {
+    static var identifier = 0
+}
+
+class QueuedExecutor {
+    typealias Completion = (Error?) -> Void
+    typealias ExecutorClosure = (_: @escaping Completion) -> Void
```
nuno-vieira, line 13: I don't get why we need this. The ticket clearly says the implementation should be the same as the channel and message pagination. I would recommend taking a look at `ChannelController` and use the same strategy to do pagination so that the code is consistent across the codebase.

Sources/StreamChat/Controllers/MemberListController/MemberListController.swift
```diff
@@ -140,13 +142,37 @@ public extension ChatChannelMemberListController {
...
+                    self.query = updatedQuery
+                    executorCompletion(nil)
+                    return
+                }
+                
+                // Perform Loading
+                self.memberListUpdater.load(updatedQuery) { [weak self] error in
+                    self?.query = updatedQuery
+                    executorCompletion(error)
+                }
```
nuno-vieira, line 168: You will need the payload response of the `memberListUpdater` to check if there are more members or not. If the fetched payload.members.count is below the provided limit, it means there are no more members to fetch. Like mentioned above, please see the `ChannelController` implementation.

nuno-vieira (changes requested):

Sources/StreamChat/Utils/QueuedExecutor.swift
kojiba: I will check for sure, 

This approach handles multiple requests on same range and also concurrent requests for different ranges, 
with tests this should be more clear

kojiba: In this case all the requests are put to serial queue which is executing them 1-by-1 and never double-loads any ranges, which can be caused by incorrect UI implementation.

tbarbugli (commented):
BlockingExecutor seems over simplified (returning an error when one request is inflight is not very user-friendly)

kojiba:
> BlockingExecutor seems over simplified (returning an error when one request is inflight is not very user-friendly)

It is reusing logic from ChannelController request handling, firstly I've done a QueuedLoader but @nuno-vieira noted that its not following current codebase approach. 
Lets firstly make logic reused and then if needed we could add an improvement for this in future (cause improvement will take a lot of work due to current `@Atomic` usage instead of QueueSerial.async), and to not make another syncronization-threading-queueing bugs in current codebase.

PR #1435, opened 2021-09-09
Fix Channel List / Channels showing no data after reconnection

b-onc:
Discussion in slack

### DIFF
--- a/Sources/StreamChat/ChatClient.swift
+++ b/Sources/StreamChat/ChatClient.swift
@@ -229,16 +229,7 @@ public class ChatClient {
             ]
             
             // All production event workers
-            eventWorkerBuilders = [
-                {
-                    ConnectionRecoveryUpdater(
-                        database: $0,
-                        eventNotificationCenter: $1,
-                        apiClient: $2,
-                        useSyncEndpoint: config.isLocalStorageEnabled
-                    )
-                }
-            ]
+            eventWorkerBuilders = []
         } else {
             workerBuilders = []
             eventWorkerBuilders = []
--- a/Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift
+++ b/Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift
@@ -129,7 +129,7 @@ public class ChatChannelListController: DataController, DelegateCallable, DataSt
                 callback: { [unowned self] in
                     switch $0.webSocketConnectionState {
                     case .connected:
-                        self.updateChannelList(trumpExistingChannels: self.channels.count > self.requestedChannelsLimit)
+                        self.updateChannelList(trumpExistingChannels: true)
                     default:
                         break
                     }
@@ -149,6 +149,7 @@ public class ChatChannelListController: DataController, DelegateCallable, DataSt
             switch result {
             case .success:
                 self.state = .remoteDataFetched
+                self.hasLoadedAllPreviousChannels = false
                 self.callback { completion?(nil) }
             case let .failure(error):
                 self.state = .remoteDataFetchFailed(ClientError(with: error))
--- a/Sources/StreamChat/Database/DatabaseSession.swift
+++ b/Sources/StreamChat/Database/DatabaseSession.swift
@@ -162,6 +162,8 @@ protocol ChannelDatabaseSession {
         query: ChannelListQuery?
     ) throws -> ChannelDTO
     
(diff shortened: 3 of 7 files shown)


Sources/StreamChat/ChatClient.swift
```diff
@@ -229,16 +229,7 @@ public class ChatClient {
             ]
             
             // All production event workers
-            eventWorkerBuilders = [
```
tbarbugli, line 232: @b-onc better if you comment this instead deleting + add a /// patch comment

Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift
```diff
@@ -129,7 +129,7 @@ public class ChatChannelListController: DataController, DelegateCallable, DataSt
                 callback: { [unowned self] in
                     switch $0.webSocketConnectionState {
                     case .connected:
-                        self.updateChannelList(trumpExistingChannels: self.channels.count > self.requestedChannelsLimit)
+                        self.updateChannelList(trumpExistingChannels: true)
```
tbarbugli, line 132: AFAICT this is not a needed change, I would leave it alone

Sources/StreamChat/Database/DatabaseSession.swift
```diff
@@ -162,6 +162,8 @@ protocol ChannelDatabaseSession {
         query: ChannelListQuery?
     ) throws -> ChannelDTO
     
+    @discardableResult func saveQuery(query: ChannelListQuery) -> ChannelListQueryDTO
```
tbarbugli, line 165: out of scope

Sources/StreamChat/Workers/ChannelListUpdater.swift
```diff
@@ -27,7 +27,8 @@ class ChannelListUpdater: Worker {
                     self?.database.write { session in
                         
                         if trumpExistingChannels {
-                            try session.deleteChannels(query: channelListQuery)
+                            let queryDTO = session.saveQuery(query: channelListQuery)
```
tbarbugli, line 30: out of scope

Sources/StreamChat/Workers/Background/DatabaseCleanupUpdater.swift
```diff
@@ -72,7 +72,7 @@ class DatabaseCleanupUpdater: Worker {
     }
 }
 
-private extension ChannelDTO {
+extension ChannelDTO {
```
tbarbugli, line 75: out of scope

Sources/StreamChat/Database/DatabaseSession_Mock.swift
```diff
@@ -165,6 +165,10 @@ extension DatabaseSessionMock {
         underlyingSession.loadChannelReads(for: userId)
     }
     
+    func saveQuery(query: ChannelListQuery) -> ChannelListQueryDTO {
```
tbarbugli, line 168: out of scope

tbarbugli (changes requested):
lets limit the changes to the least we needed

Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift
evsaev: Hm, I would have it 🤔 

When we reconnect, we should invalidate history of local channels because there can be changes we are not aware of.

Use case:
1. load 1st page (25 channels)
2. go offline
3. user becomes a member of a new channel that goes to 1st page (forcing locally existed channel X out from 1st page)
4. smth changes in channel X outside the 1st page of messages
5. go back online and do not trump local data

The history of channel X will be out of sync and user won't see the updates

Sources/StreamChat/Workers/Background/DatabaseCleanupUpdater.swift
evsaev: Extension contains `ChannelDTO .resetLocalData` used in channel list updater to trump a query -> we need this change 🙂

Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift
evsaev: I would trump on `synchronise/reconnect` no matter the storage setup. With no `/sync` and enabled local storage, queries will go out of sync very soon

PR #1438, opened 2021-09-10
[CIS-1157] Introduce linking delegate callbacks

b-onc:
These new delegate funcs will be called for any inserted/updated channels which is not linked to the controller's query. Users will have a way to link new/updated channels to the query, without the need for `NewChannelQueryUpdater`

### DIFF
--- a/Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift
+++ b/Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift
@@ -77,6 +77,7 @@ public class ChatChannelListController: DataController, DelegateCallable, DataSt
                 }
                 $0.controller(self, didChangeChannels: changes)
             }
+            self?.handleLinkedChannels(changes)
         }
 
         observer.onWillChange = { [weak self] in
@@ -90,6 +91,20 @@ public class ChatChannelListController: DataController, DelegateCallable, DataSt
             }
         }
 
+        return observer
+    }()
+    
+    lazy var updatedChannelObserver: ListDatabaseObserver<ChatChannel, ChannelDTO> = {
+        let observer = self.environment.createChannelListDatabaseObserver(
+            client.databaseContainer.viewContext,
+            ChannelDTO.channelsFetchRequest(notLinkedTo: query),
+            { $0.asModel() }
+        )
+        
+        observer.onChange = { [weak self] changes in
+            self?.handleUnlinkedChannels(changes)
+        }
+        
         return observer
     }()
     
@@ -181,6 +196,7 @@ public class ChatChannelListController: DataController, DelegateCallable, DataSt
         guard state == .initialized else { return }
         do {
             try channelListObserver.startObserving()
+            try updatedChannelObserver.startObserving()
             state = .localDataFetched
         } catch {
             state = .localDataFetchFailed(ClientError(with: error))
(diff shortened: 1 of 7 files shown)


Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift
```diff
@@ -93,6 +93,29 @@ public class ChatChannelListController: DataController, DelegateCallable, DataSt
...
     
+    lazy var updatedChannelObserver: ListDatabaseObserver<ChatChannel, ChannelDTO> = {
+        let request = NSFetchRequest<ChannelDTO>(entityName: ChannelDTO.entityName)
+        request.sortDescriptors = [ChannelListSortingKey.defaultSortDescriptor]
+        // Channels which are not linked to this query
+        request.predicate = NSCompoundPredicate(
+            notPredicateWithSubpredicate: NSPredicate(
+                format: "ANY queries.filterHash == %@", query.filter.filterHash
+            )
+        )
```
evsaev, line 104: Would be nice to have this predicate declared in `ChannelDTO` next to other ones 👍

Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift
```diff
@@ -93,6 +93,29 @@ public class ChatChannelListController: DataController, DelegateCallable, DataSt
...
+        
+        let observer = self.environment.createChannelListDatabaseObserver(
+            client.databaseContainer.viewContext,
+            request,
+            { $0.asModel() }
+        )
+        
+        observer.onChange = { [unowned self] changes in
+            handleUpdatedChannels(changes)
+        }
```
evsaev, line 114:
```suggestion
        observer.onChange = { [weak self] changes in
            self?.handleUpdatedChannels(changes)
        }
```

Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift
```diff
@@ -196,6 +220,33 @@ public class ChatChannelListController: DataController, DelegateCallable, DataSt
...
+    private func handleUpdatedChannels(_ changes: [ListChange<ChatChannel>]) {
+        let channels = changes.compactMap { change -> ChatChannel? in
+            switch change {
+            case let .insert(channel, _):
+                return (delegate?.controller(self, shouldLinkNewChannel: channel) ?? false) ? channel : nil
+            case let .update(channel, _):
+                return (delegate?.controller(self, shouldLinkUpdatedChannel: channel) ?? false) ? channel : nil
+            default: return nil
+            }
+        }
```
evsaev, line 233: Should we wrap this with `delegateCallback` to make sure the delegate is invoked on callback queue?

Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift
```diff
@@ -286,6 +337,26 @@ public protocol ChatChannelListControllerDelegate: DataControllerStateDelegate {
...
+    
+    /// The controller asks the delegate if the newly updated `ChatChannel` should be linked to this Controller's query.
+    /// Defaults to `false`
+    /// - Parameters:
+    ///   - controller: The controller,
+    ///   - shouldLinkUpdatedChannel: The newly updated `ChatChannel` instance. This instance is not linked to the controller's query.
+    func controller(
+        _ controller: ChatChannelListController,
+        shouldLinkUpdatedChannel: ChatChannel
+    ) -> Bool
```
evsaev, line 359: I like it 👍 What if updated channel no longer matches the query, though? 🤔 Do we need:
```swift
func controller( _ controller: ChatChannelListController, shouldUnlinkUpdatedChannel: ChatChannel) -> Bool
```

evsaev (changes requested):

Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift
evsaev: One extra came to my mind: what if we give a `completion: (Bool) -> Void` to these methods? 🙂 Then we'll also cover use-cases when it's not possible to make a decision synchronously

Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift
b-onc: IMHO it's not needed, the same reason `completion` is not needed for this func: users should keep the processing to a minimum, we don't want them to make an API call for this, it'd just make stuff more complex

Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift
b-onc: users should keep the processing to a minimum, we don't want them to make an API call for this, it'd just make stuff more complex

`shouldUnlinkUpdatedChannel` is not needed for now

evsaev (dismissed):
👏 🚢

evsaev (approved):
🚢

PR #1440, opened 2021-09-10
[CIS-1152] Fix connection recovery

evsaev:
**This PR** fixes the way we sync data when connection comes back.

Connection recovery logic involves 2 steps:
1. calling `/sync` to get the events we've missed
2. re-fetching channel list queries to:
- get new channels
- pick up the latest ordering
- start watching the channels

**Step 1 (/sync)**

In `ConnectionRecoveryUpdater` when `connected` event is observed we call `/sync` endpoint with all locally existed channels and `lastSyncAt` date persisted in database. If `/sync` succeeds we apply the events we've received which makes locally existed channels up-to-date. If `/sync` fails, all locally existed channels are out-of-sync and we should reset it.

**Step 2 (re-fetching queries)**
When we refetch a first page of channels (with `.watch` option) from backend for the given `query`, we unlink all locally existed channels from a query and link only channels we received from backend (the first page). This guarantees the local channel list query matches the remote one. For local channels that were **synced** (on step 1) and returned from API we preserve the message history. If local channel was not synced OR not returned from API, the message history is cleared.

At the same time, we should gracefully handle channels that are being read (`Channel.isBeingRead` is set to `true` when channel is opened, and reset to `false` when it is closed). If a channel that is being read is synced and returned from API, we keep it's local history and make additional query with [`.watch, .presense`] options to start getting all the events. If a channel that is being read was not synced, we query is but then reset the local history (since it can be outdated) and save only message we received from API (the first page of messages).

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -51,6 +51,7 @@ var streamChatSourcesExcluded: [String] { [
     "Database/DatabaseSession_Mock.swift",
     "Database/DatabaseSession_Tests.swift",
     "Database/DTOs/UserDTO_Tests.swift",
+    "Database/DTOs/ChannelListQueryDTO_Tests.swift",
     "Database/DTOs/ChannelMuteDTO_Tests.swift",
     "Database/DTOs/ChannelDTO_Tests.swift",
     "Database/DTOs/CurrentUserDTO_Tests.swift",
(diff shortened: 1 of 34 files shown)


evsaev:
Closing this since it's not gonna be merged. The decision on who should be responsible for syncing (channel list/channel controller or some background worker) and what should be the way of tracking active components (either persist in the db or store in-memory within the current session) have not been made yet. We'll keep working on this topic

PR #1446, opened 2021-09-13
[CIS-1167] Implement token refreshing for HTTP requests

dmigach:
Currently we only refresh expired token when we get an error over WebSocket connection which is suboptimal because backend doesn't send an event for token expiration. So we should react to this error when we get it as a response for an HTTP request

There are three main things to consider, all covered with a corresponding test:

When request fail, we add them to a queue to retry them later with a new token.
Then we try refreshing the token, if it's successful - retry all the failed requests
If refreshing failed or hanging - timeout all the requests with failures 

Two main cases to test:
1) Token expires, you send a couple of requests (like creating new messages). They hang for a bit and then retried once the token is refreshed
2) Token expires, token refreshing fails, requests should fail after the timeout


Sources/StreamChat/APIClient/APIClient.swift
```diff
@@ -15,7 +20,19 @@ class APIClient {
     /// `APIClient` uses this object to decode the results of network requests.
     let decoder: RequestDecoder
     
+    /// Used for reobtating tokens when they expire and API client receives token expiration error
+    let tokenRefresher: (ClientError, @escaping () -> Void) -> Void
+    
     let cdnClient: CDNClient
+
+    /// Used for syncrhonizing access to requestsQueue
```
tbarbugli, line 28: typo

Sources/StreamChat/APIClient/APIClient.swift
```diff
@@ -15,7 +20,19 @@ class APIClient {
     /// `APIClient` uses this object to decode the results of network requests.
     let decoder: RequestDecoder
     
+    /// Used for reobtating tokens when they expire and API client receives token expiration error
```
tbarbugli, line 23: typo

Sources/StreamChat/APIClient/APIClient.swift
```diff
@@ -15,7 +20,19 @@ class APIClient {
...
     let cdnClient: CDNClient
+
+    /// Used for syncrhonizing access to requestsQueue
+    private let requestsAccessQueue = DispatchQueue(label: "io.getstream.requests")
+    
+    /// Stores request failed with token expired error for retrying them later
+    private var requestsQueue = [RequestsQueueItem]()
+    
+    /// Shows whether the token is being refreshed at the moment
+    private var isRefreshingToken: Bool = false
```
tbarbugli, line 35: unless access `isRefreshingToken` is serialized, I suggest using @Atomic to avoid surprises

Sources/StreamChat/APIClient/APIClient.swift
```diff
@@ -90,6 +128,46 @@ class APIClient {
...
+                self.requestsAccessQueue.asyncAfter(deadline: .now() + timeout) {
+                    self.requestsQueue.forEach { $0.failureAction() }
+                    self.requestsQueue = []
+                }
+            }
+            
+            self.requestsQueue.append(item)
+            
+            // Only initiate token refreshing once
+            guard !self.isRefreshingToken else { return }
```
tbarbugli, line 161: see previous comment about concurrent usage of self and `isRefreshingToken `. If needed the Atomic `propertyWrapper` can be really useful here.

```swift

/// ...
self.requestsQueue.append(item)

guard self._ isRefreshingToken.compareAndSwap(old: false, new: true) else {
   return
}

self.requestsAccessQueue.asyncAfter(deadline: .now() + timeout) {
   self.requestsQueue.forEach { $0.failureAction() }
   self.requestsQueue = []
}

self.tokenRefresher(...)
```

Sources/StreamChat/APIClient/APIClient.swift
```diff
@@ -90,6 +128,46 @@ class APIClient {
...
+            
+            self.isRefreshingToken = true
+            self.tokenRefresher(ClientError.ExpiredToken()) {
+                self.requestsQueue.forEach { $0.requestAction() }
```
tbarbugli, line 165: Following feedback is debatable since it only applies if the token expires quickly after one is obtained:

Order here is not 100% correct, you probably want to first swap `isRefreshingToken` to false and then consume the queue. 

It also sounds safer to consume the queue this way:

- while requestsQueue is not empty
   - check `isRefreshingToken` or exit
   - pop requestsQueue 
   - exec

Sources/StreamChat/APIClient/APIClient.swift
```diff
@@ -15,7 +20,19 @@ class APIClient {
...
+    /// Used for reobtating tokens when they expire and API client receives token expiration error
+    let tokenRefresher: (ClientError, @escaping () -> Void) -> Void
+    
     let cdnClient: CDNClient
+
+    /// Used for syncrhonizing access to requestsQueue
+    private let requestsAccessQueue = DispatchQueue(label: "io.getstream.requests")
+    
+    /// Stores request failed with token expired error for retrying them later
+    private var requestsQueue = [RequestsQueueItem]()
```
tbarbugli, line 32: not sure if we reuse the same instance of APIClient if you login/logout (connectUser/disconnect), if we do we need to make sure that state is reset

Sources/StreamChat/APIClient/APIClient.swift
```diff
@@ -90,6 +128,46 @@ class APIClient {
...
         )
     }
+    
+    /// Queues a failed request for executing later or failing after a timeout
+    func handleTokenExpirationError<Response: Decodable>(
+        urlRequest: URLRequest,
+        timeout: TimeInterval,
+        completion: @escaping (Result<Response, Error>) -> Void
+    ) {
+        requestsAccessQueue.async {
```
tbarbugli, line 138: do we need [weak self] here? what if the APIClient gets `deinit` while some stuff is queued?

dmigach: Imho we should keep everything alive until we complete all the requests, because we didn't show errors for instance for messages sent, so a user might've left and then these messages will never actually reach their destination. Of course keeping references to the client won't prevent a user from killing the app, but it's still better than nothing

Sources/StreamChat/APIClient/APIClient.swift
```diff
@@ -15,8 +20,26 @@ class APIClient {
...
+    private var requestsQueue = [RequestsQueueItem]()
     
+    /// Shows whether the token is being refreshed at the moment
+    @Atomic private var isRefreshingToken: Bool = false
+    
+    /// How many times refreshing the token failed consecutively
+    @Atomic private var tokenRefreshConsecutiveFailures: Int = 0
+
+    /// How many times can the token refresh fail before giving up with an error
+    let maxTokenRefreshAttempts = 10
```
nuno-vieira, line 41: Isn't this too much? 🤔 AFAIK on iOS the default for this kind of thing is 5.

nuno-vieira (approved):
LGTM

Sources/StreamChat/APIClient/APIClient.swift
```diff
@@ -27,20 +50,38 @@ class APIClient {
         sessionConfiguration: URLSessionConfiguration,
         requestEncoder: RequestEncoder,
         requestDecoder: RequestDecoder,
-        CDNClient: CDNClient
+        CDNClient: CDNClient,
+        tokenRefresher: @escaping (ClientError, @escaping () -> Void) -> Void
```
nuno-vieira, line 54: This closure is a bit hard to understand, wouldn't be easier to introduce a new class like `TokenRefresher` that would handle everything about the refreshing? The state, making refresh requests, etc... ? With the refreshing logic all inside the `request()` function, it makes it harder to understand, especially with all the queue's handling. With a `TokenRefresher` that would be all inside it

PR #1457, opened 2021-09-15
[CIS-1180] Fix currentDevice is always nil

### LINKED ISSUE #1448: `currentDevice` is always `nil`
## What did you do?

I am adding a device as follows:
```swift
chatClient.currentUserController().addDevice(token: deviceToken) { (error) in
   // ... handle error
    print("[Chat] Added device with token: \(deviceToken)") // Gets called - no issues.
}
```

When attempting to access the current device (with the intention of removing it):

```swift
// synchronize devices
self?.chatClient.currentUserController().synchronizeDevices { [weak self] error in
    // ... handle error 
    
let currentDevice = self?.chatClient.currentUserController().currentUser?.currentDevice // GetStream bug? Always nil.
    print("[Chat] Current device: \(currentDevice)") // nil
    let devices = self?.chatClient.currentUserController().currentUser?.devices
    print("[Chat] Devices: \(devices ?? [])") // array of all the devices added
}
```

it is always `nil`.

`currentDevice` to be non-nil and have a reference to the current device. Instead, it's `nil` even though all added devices (on this single device) appear in the `devices` array.

## What happened instead?

`currentDevice` is `nil`. There is no device to pass to `removeDevice(id:)`.

## Documentation Mismatch

Besides this bug above, the GetStream docs explain removing the current user's device differently.

Under [Push Notifcations / Devices](https://getstream.io/chat/docs/ios-swift/push_devices/?language=swift) the example shows using the currentDevice (which seems reasonable):

```swift
let currentUserController = chatClient.currentUserController() 
 
guard let device = currentUserController.currentUser?.currentDevice else { 
    // No device to remove 
} 
 
currentUserController.removeDevice(id: device.id) { error in 
    if let error == error { 
        print("Error removing the device", error) 
    } 
}
```

Under [iOS Chat UX / Notifications](https://getstream.io/chat/docs/ios-swift/ios_push_notifications/) the example shows using the last element in the `devices` array (which doesn't make sense – unless there is some hidden ordering?):

```swift
let deviceId = chatClient.currentUserController().currentUser!.devices.last!.id 
 
chatClient.currentUserController().removeDevice(id: deviceId) 
// or 
chatClient.currentUserController().removeDevice(id: deviceId) { error in 
    if let error = error { 
        // handle error 
        print(error) 
    } 
}
```

## GetStream Environment
**GetStream Chat version:** 4.0.0
**GetStream Chat frameworks:** StreamChat, StreamChatUI, StreamChatClient, StreamChatCore
**iOS version:** 14
**Swift version:** 5.4.2
**Xcode version:** 13 beta 5
**Device:**

## Additional context

Help would be appreciated, thanks!

b-onc:
Fixes #1448


Sources/StreamChat/Controllers/CurrentUserController/CurrentUserController.swift
```diff
@@ -193,30 +193,26 @@ public extension CurrentChatUserController {
...
     
     /// Registers a device to the current user.
-    /// `setUser` must be called before calling this.
+    /// `connectUser` must be called before calling this.
     /// - Parameters:
     ///   - token: Device token, obtained via `didRegisterForRemoteNotificationsWithDeviceToken` function in `AppDelegate`.
+    ///   - asCurrentDevice: Specifies if the device should be saved as `currentDevice` for the user. Defaults to `true`.
     ///   - completion: Called when device is successfully registered, or with error.
-    func addDevice(token: Data, completion: ((Error?) -> Void)? = nil) {
+    func addDevice(token: Data, asCurrentDevice: Bool = true, completion: ((Error?) -> Void)? = nil) {
```
tbarbugli, line 201: @b-onc whats the usecase for `asCurrentDevice: false`?

b-onc: As a provider, I keep the user's device tokens in my own backend, and when they install an updated client (which includes StreamChat SDK) to one of their devices I want to enable push notifications for all of their devices. So I add all of their devices from a single client

tbarbugli (commented):
I rather take a better look at how we handle devices, looks way too complicated that it needs to be (ie. we should not have a DTO for devices)

nuno-vieira (dismissed):
LGTM

nuno-vieira (approved):

tbarbugli (approved):

PR #1458, opened 2021-09-15
[CIS-1177] Fix typing events always disabled when channel opened without cache from Channel List

nuno-vieira:
When opening a channel without coming from a channel list, the typing events would be disabled even if the config had them enabled. The reason is that the `ChatMessageListVC.setUp` function is called before the `ChatChannelVC` finishing the synchronization, so at that time the channel config, won't be available. 

This fixes the issue by always adding the typing indicator view as a subview in the `setUp` function. And only uses the config when showing the typing indicator, if it is enabled, it sets `isHidden = true` , if not, it remains hidden.


evsaev (approved):
👍

Issue #1332, opened 2021-07-29
Get payload data from MessageNewEvent for local push notifications 


energo:
In continuation of the previous [discussion](https://github.com/GetStream/stream-chat-swift/pull/1266), in the way to show notifications when the app in the background we need to use an event listener

```
class CustomChannelListVC: ChatChannelListVC, EventsControllerDelegate {
    
    var eventsController: EventsController!
    
    override func viewDidLoad() {
        super.viewDidLoad()
        
        eventsController = controller.client.eventsController()
        eventsController.delegate = self
    }
    
    func eventsController(_ controller: EventsController, didReceiveEvent event: Event) {
        switch event {
        case let event as MessageNewEvent:
            // do what you need with the event
            print(event.messageId)
            // you can check if the app is in BG state and show local notification here
        default:
            break
        }
    }
}
```

We already have all data that we needed to show a local notification in "payload" property at "MessageNewEvent" object.
For title "name", "chat name" and for body "message text" 
But seems its accessibly only for SDK uses, and not in public API.

<img width="786" alt="Screenshot 2021-07-28 at 15 33 35" src="https://user-images.githubusercontent.com/1804123/127530365-8ad675f6-e79c-404c-a881-79387691bf09.png">

kcollignon:
@b-onc I seem to be struggling with this as well. Have you all been able to take a look at this yet?

tbarbugli:
@kcollignon ideally this event had a `message` property, team is looking into this and I think we'll be able to address it this week

evsaev:
Hey @kcollignon @energo 👋 

The PR that adds models to events will be merged soon 👍

Thanks for your patience!

energo:
> Hey @kcollignon @energo 👋
> 
> The PR that adds models to events will be merged soon 👍
> 
> Thanks for your patience!

Great news, thank you @evsaev

PR #1459, opened 2021-09-16
[CIS-1081] Expose models on events

evsaev:
**This PR** updates all public facing events to expose chat models in a non-breaking way.

For each existed `***Event` type an `***EventDTO` type is introduced (e.g. `UserUpdatedEventDTO/UserUpdatedEvent` ). When an event comes via web-socket it is parsed as `EventPayload`, then depending on `type` it is transformed to the `***EventDTO`. When event has been processed and database is updated accordingly, the last middleware in a chain (`EventConverterMiddleware`) performs `EventDTO  -> Event` transformation and the final event which chat models exposed gets published.

Before:
```swift
public struct MessageUpdatedEvent: MessageSpecificEvent {
    public let userId: UserId
    public let cid: ChannelId
    public let updatedAt: Date
}
```

After:
```swift
/// Triggered when a message is updated.
public struct MessageUpdatedEvent: Event {
    /// The use who updated the message.
    public let user: ChatUser
    /// The channel identifier the message is sent to.
    public let channel: Channel
    /// The updated message.
    public let message: ChatMessage
    /// The event timestamp.
    public let createdAt: Date
}
```

How it can be used to show a local notification when a new message is observed:
```swift
class MyChannelListVC: ChatChannelListVC, EventsControllerDelegate {
    lazy var eventsController: EventsController = controller.client.eventsController()

    override func viewDidLoad() {
        super.viewDidLoad()

        eventsController.delegate = self
    }
    
        guard let event = event as? MessageNewEvent else { return }
        
        let message = event.message
        
        let content = UNMutableNotificationContent()
        content.title = "\(message.author.name ?? message.author.id) @ \(event.channel.name ?? event.channel.cid.id)"
        content.body = message.text
        
        let request = UNNotificationRequest(
            identifier: message.id,
            content: content,
            trigger: nil
        )
        
        UNUserNotificationCenter.current().add(request) { error in
            if let error = error {
                log.error("Error when showing local notification: \(error)")
            }
        }
    }
```

Public-facing changes:
- All existed events are updated to exposed models instead of model ids
- `***Id` properties are deprecated
- Docs are added to all event types/fields
- `HasUnreadCount` protocol is introduces so its convenient to catch all events with unread counts

Internal changes:
- For each existed event type an `***EventDTO` is introduced
- Introduced `EventConverterMiddleware` which performs `DTO -> Event` transformation
- Internal `UserSpecific/UserSpecificEvent/MessageSpecificEvent/ReactionEvent` protocols are gone (`ChannelSpecificEvent` has stayed since it is used by `ChannelEventsController`)
- `EventWithPayload` now exposes payload as `EventPayload` instead of `Any`
- `EventWithPayload` is renamed to `EventDTO`


Sources/StreamChat/WebSocketClient/Events/ChannelEvents.swift
```diff
@@ -4,60 +4,173 @@
...
+            createdAt: event.createdAt
+        )
     }
 }
 
-public struct ChannelDeletedEvent: ChannelSpecificEvent {
-    public let cid: ChannelId
+public struct ChannelDeletedEvent: Event {
+    public let channel: ChatChannel
+    public let user: ChatUser?
```
evsaev, line 41: If it's a user who deleted a channel, how about `deletedBy` naming? 🙂

Sources/StreamChat/WebSocketClient/Events/ChannelEvents.swift
```diff
@@ -4,60 +4,173 @@
...
+        )
     }
 }
 
-public struct ChannelTruncatedEvent: ChannelSpecificEvent {
-    public let cid: ChannelId
-    let payload: Any
+public struct ChannelTruncatedEvent: Event {
+    public let channel: ChatChannel
+    public let user: ChatUser
```
evsaev, line 77: `truncatedBy: ChatUser`? 🙂

Sources/StreamChat/WebSocketClient/Events/MemberEvents.swift
```diff
@@ -4,41 +4,112 @@
 
 import Foundation
 
-public struct MemberAddedEvent: MemberEvent, ChannelSpecificEvent, EventWithPayload {
-    public let memberUserId: UserId
+public struct MemberAddedEvent: MemberEvent {
     public let cid: ChannelId
+    public let user: ChatUser
```
evsaev, line 9: Maybe `addedBy: ChatUser`?

Sources/StreamChat/WebSocketClient/Events/MessageEvents.swift
```diff
@@ -4,78 +4,168 @@
 
 import Foundation
 
-public struct MessageNewEvent: MessageSpecificEvent {
-    public let userId: UserId
-    public let cid: ChannelId
-    public let messageId: MessageId
+public struct MessageNewEvent: Event {
```
evsaev, line 7: For some reason, there's not such event on `Android`, they have only `NotificationMessageNewEvent`. Should we do the same?

Sources/StreamChat/WebSocketClient/Events/MessageEvents.swift
```diff
@@ -4,78 +4,168 @@
...
     }
 }
 
 /// `ChannelReadEvent`, this event tells that User has mark read all messages in channel.
 public typealias ChannelReadEvent = MessageReadEvent
 
 /// `ChannelReadEvent`, this event tells that User has mark read all messages in channel.
-public struct MessageReadEvent: UserSpecificEvent, ChannelSpecificEvent {
-    public let userId: UserId
+public struct MessageReadEvent: Event {
```
evsaev, line 137: There's no `messageId` in the event so having names vice versa would make more sense: `ChannelReadEvent` - a real type and `MessageReadEvent` as a second name

Sources/StreamChat/WebSocketClient/Events/MemberEvents.swift
```diff
@@ -4,41 +4,112 @@
...
 }
 
-public struct MemberUpdatedEvent: MemberEvent, ChannelSpecificEvent, EventWithPayload {
-    public let memberUserId: UserId
+public struct MemberUpdatedEvent: MemberEvent, HasCid, HasUser, HasMember {
     public let cid: ChannelId
+    public let user: ChatUser
+    public let member: ChatChannelMember
+    public let createdAt: Date
+}
```
nuno-vieira, line 51: Another question is.. do we really need these protocols? Why just simply create `MemberUpdatedEvent` with user, member and cid, without the protocols? It looks to me that the protocols are not needed at all.

Sources/StreamChat/WebSocketClient/Events/MemberEvents.swift
```diff
@@ -4,41 +4,112 @@
...
+        }
+    }
     
-    let payload: Any
-    
-    init(from response: EventPayload) throws {
-        memberUserId = try response.value(at: \.memberContainer?.member?.user.id)
-        cid = try response.value(at: \.cid)
-        payload = response
+    init?(event: DTO, session: DatabaseSession) {
```
nuno-vieira, line 70:
```suggestion
    init?(eventDTO: DTO, session: DatabaseSession) {
```

nuno-vieira (commented):
Added some suggestions to the current approach. 

But I also have a proposal. Looking into how the other SDKs are doing it here: https://getstream.io/chat/docs/react/event_listening/, it looks like we are very different from the rest of the SDKs. Since it is not possible to change right now the EventsController delegate approach to a more subscription-based approach, could we maybe make the `EventsController` a generic one? Like `EventsController<T: Event>` ? The question here is, what if you want to subscribe to multiple events right? This would require creating multiple events controllers? 🤔 

Another idea: 
What if instead of having one delegate function from `EventsController`, to delegate function per event. In this case, we would avoid all the casting that we are doing here, right?

tbarbugli (commented):
I would be great to get some explanation about what you are doing here. 

ie

Add a DTO type for each event so that we have XXXXEvent and XXXXEventDTO.

- XXXXEvent is crated by ... and is responsible for ...
- When [Y happens] the SDK ... and .... XXXXEventDTO ...

A new middleware is added and .... then ....

tbarbugli (commented):
Can you write a working solution (which uses this PR's code) for what was reported here? https://github.com/GetStream/stream-chat-swift/issues/1332

Sources/StreamChat/Deprecations.swift
```diff
@@ -5,3 +5,58 @@
...
+    var memberUserId: UserId { user.id } // invalid?
+}
+
+public extension ChannelDeletedEvent {
+    var deletedAt: Date { createdAt }
+}
+
+public extension ChannelHiddenEvent {
+    var hiddenAt: Date { createdAt }
+}
```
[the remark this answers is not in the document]
evsaev: yeah, definitely 👍 didn't add since it's a draft

Sources/StreamChat/WebSocketClient/Events/MemberEvents.swift
evsaev: The reason of using `Has***` is mainly to stay aligned with [Android](https://github.com/GetStream/stream-chat-android/blob/880cfccf9dec7f1f088a1380e888cd6dee5a7536/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/events/ChatEvent.kt). 

The use-case on customer side I was thinking of is catching all the events related to a specific model:
```swift
extension DemoChannelListVC: EventsControllerDelegate {
        guard let userEvent = event as? HasUser else { return }
        
        let user = userEvent.user
        ...
    }
}
```

Without a bounding protocol they would have to check for concrete event types which not that nice 🤔 On the other hand, we indeed can remove these protocols/make them `internal` to reduce public API changes so there're pros in both approaches.

@nuno-vieira @b-onc @adamrushy @tbarbugli WDYT guys?

evsaev:
> Can you write a working solution (which uses this PR's code) for what was reported here? #1332

Would it work if I rebase https://github.com/GetStream/stream-chat-swift/pull/1439 on top of this branch and update it accordingly to show-case the complexity being reduced? 🤔 Local notifications stuff included into this PR since feels not the best idea since using models for local notifications is just one of the use-cases and it's purely about `DemoApp`..

evsaev:
> Added some suggestions to the current approach.
> 
> But I also have a proposal. Looking into how the other SDKs are doing it here: https://getstream.io/chat/docs/react/event_listening/, it looks like we are very different from the rest of the SDKs. Since it is not possible to change right now the EventsController delegate approach to a more subscription-based approach, could we maybe make the `EventsController` a generic one? Like `EventsController<T: Event>` ? The question here is, what if you want to subscribe to multiple events right? This would require creating multiple events controllers? 🤔
> 
> Another idea:
> What if instead of having one delegate function from `EventsController`, to have a delegate function per event. In this case, we would avoid all the casting that we are doing here, right?

>The question here is, what if you want to subscribe to multiple events right? This would require creating multiple events controllers? 🤔

Yes, exactly! 🙂 

It's kinda annoying to have a controller per event + `EventsController<T>` would be a breaking change (unless we keep both `EventController` and `NewEventsController<T>` till next major release). 

Suggestion to have a func per event sounds good to me 👍 @tbarbugli WDYT? Should we have a ticket in JIRA for this?

Sources/StreamChat/WebSocketClient/Events/MemberEvents.swift
nuno-vieira: What if we could use here something like @dynamicMemberLookup? 🤔 

Since an event can have multiple models, and it is pretty dynamic, it sounds like a good use case for @dynamicMemberLookup, what do you guys think? 

The idea is that an Event can have multiple models, so we would store the models in a dictionary or something, and then the customer could access it with `event.user` or `event.channel` it was like you were accessing a dictionary

Sources/StreamChat/WebSocketClient/Events/ChannelEvents.swift
nuno-vieira: Yes I think `deletedBy`makes more sense IMO

nuno-vieira:
> > Added some suggestions to the current approach.
> > But I also have a proposal. Looking into how the other SDKs are doing it here: https://getstream.io/chat/docs/react/event_listening/, it looks like we are very different from the rest of the SDKs. Since it is not possible to change right now the EventsController delegate approach to a more subscription-based approach, could we maybe make the `EventsController` a generic one? Like `EventsController<T: Event>` ? The question here is, what if you want to subscribe to multiple events right? This would require creating multiple events controllers? 🤔
> > Another idea:
> > What if instead of having one delegate function from `EventsController`, to have a delegate function per event. In this case, we would avoid all the casting that we are doing here, right?
> 
> > The question here is, what if you want to subscribe to multiple events right? This would require creating multiple events controllers? 🤔
> 
> Yes, exactly! 🙂
> 
> It's kinda annoying to have a controller per event + `EventsController<T>` would be a breaking change (unless we keep both `EventController` and `NewEventsController<T>` till next major release).
> 
> Suggestion to have a func per event sounds good to me 👍 @tbarbugli WDYT? Should we have a ticket in JIRA for this?

Couldn't we do this as part of this PR ? 🤔 I think that making the customer to cast every event is not a good dev experience. A function per event seems a much better approach, where each function would be an optional requirement ofc. CC @tbarbugli

Sources/StreamChat/Deprecations.swift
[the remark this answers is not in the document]
evsaev: ✅

Sources/StreamChat/WebSocketClient/Events/MemberEvents.swift
evsaev: Discussed internally, using `dynamicMemberLookup` would require breaking changes so will keep that option in mind for the next major.

Sources/StreamChat/WebSocketClient/Events/MemberEvents.swift
evsaev: Made it a little bit differently: an `EventDTO` type as `toDomainEvent` func so the naming no longer makes sense since the DTO is accessed as `self`.

evsaev:
Opening a PR not to block a review process. Tomorrow will add `HasUnreadCount` and rebase PR with local notifications to show-case the usage of exposed models.

Sources/StreamChat/WebSocketClient/Events/Event.swift
```diff
@@ -15,7 +15,7 @@ protocol EventWithPayload: Event {
 }
 
 /// A protocol for any `ChannelEvent` where it has a  `channel` payload.
-protocol ChannelSpecificEvent: EventWithPayload {
+protocol ChannelSpecificEvent: Event {
```
evsaev, line 29: The only place where it is used is `ChatChannelController` and it seems too much to have a separate type and make multiple events implement it for a single internal use-case. Would it be better to just list all the events that have `cid` instead, I don't know. 🤔

Sources/StreamChat/WebSocketClient/Events/UserEvents.swift
```diff
@@ -4,17 +4,34 @@
...
-public struct UserPresenceChangedEvent: UserSpecificEvent {
-    public let userId: UserId
-    public let createdAt: Date?
+/// Triggered when user status changes (eg. online, offline, away, etc.)
+public struct UserPresenceChangedEvent: Event {
+    /// The user the status changed for
+    public let user: ChatUser
     
+    /// The event timestamp
+    public let createdAt: Date?
```
evsaev, line 13: The `createdAt` must exist but it's exposed as optional so we have to keep it as it is until the next major

Sources/StreamChat/WebSocketClient/Events/UserEvents.swift
```diff
@@ -34,17 +34,34 @@ struct UserPresenceChangedEventDTO: EventWithPayload {
...
-public struct UserUpdatedEvent: UserSpecificEvent {
-    public let userId: UserId
-    public let createdAt: Date?
+/// Triggered when user is updated
+public struct UserUpdatedEvent: Event {
+    /// The updated user
+    public let user: ChatUser
     
+    /// The event timestamp
+    public let createdAt: Date?
```
evsaev, line 43: The same here: `createdAt` it exposed as optional so we have to keep it.

Sources/StreamChat/WebSocketClient/Events/UserEvents.swift
```diff
@@ -194,17 +194,38 @@ public struct UserGloballyUnbannedEvent: UserSpecificEvent {
...
+    /// The channel identifer user is unbanned at.
     public let cid: ChannelId
-    public let userId: UserId
-    public let createdAt: Date?
     
+    /// The unbanned user.
+    public let user: ChatUser
+    
+    /// The event timestamp
+    public let createdAt: Date?
```
evsaev, line 242: Exposed as optional, keeping as is not to have a breaking change.

Sources/StreamChat/WebSocketClient/Events/UserEvents.swift
```diff
@@ -116,16 +116,34 @@ struct UserWatchingEventDTO: EventWithPayload {
 
 // MARK: - User Ban
 
-public struct UserGloballyBannedEvent: UserSpecificEvent {
-    var userId: UserId
-    var createdAt: Date?
-    var payload: Any
+/// Triggered when user is banned not in a specific channel but globally.
+public struct UserGloballyBannedEvent: Event {
```
evsaev, line 120: Not used in SDK, currently. When `user.banned` comes without `cid` - that's a `global` ban.

Sources/StreamChat/WebSocketClient/Events/UserEvents.swift
```diff
@@ -200,16 +200,34 @@ struct UserBannedEventDTO: EventWithPayload {
     }
 }
 
-public struct UserGloballyUnbannedEvent: UserSpecificEvent {
-    var userId: UserId
-    var createdAt: Date?
-    var payload: Any
+/// Triggered when user is removed from global ban.
+public struct UserGloballyUnbannedEvent: Event {
```
evsaev, line 204: Not used in SDK, currently. When `user.unbanned` comes without `cid` - the user was unbanned  `globally`.

Sources/StreamChat/WebSocketClient/Events/MessageEvents.swift
```diff
@@ -106,21 +106,49 @@ struct MessageUpdatedEventDTO: EventWithPayload {
     }
 }
 
-public struct MessageDeletedEvent: MessageSpecificEvent {
-    public let userId: UserId
+/// Triggered when a new message is deleted.
+public struct MessageDeletedEvent: ChannelSpecificEvent {
+    /// The user who deleted the message.
+    public let user: ChatUser
```
evsaev, line 120: AFAIK `message.deleted` can have no user if it was server-side deletion. But because `userId` is currently exposed as non-optional so we have 2 options:
1. always try to parse a user and fail for server-side triggered event -> it won't be posted
2. make `user` optional and fallback to empty identifier: `var userId: UserId { user?.id ?? "" }`

The 2nd seems better to me. WDYT?

Sources/StreamChat/WebSocketClient/Events/MessageEvents.swift
```diff
@@ -4,78 +4,212 @@
...
+    
+    /// The message that was sent.
+    public let message: ChatMessage
+    
+    /// The channel identifier the message was sent to.
+    public var cid: ChannelId { channel.cid }
+    
+    /// The channel a message was sent to.
+    public let channel: ChatChannel
```
evsaev, line 19: That is disputable change 🙂 On the one hand, `channel` does not come with `message.new` event payload. But to start getting `message.new` events, a user has to start watching a channel and the REST API response returns the channel which we save to the database. So at the moment the event arrives, the channel must already exist locally

*On Android, only `cid` is exposed

evsaev: The same applies for `message.updated/message.deleted/message.read`.This change is a separate commit so it's easy to roll it back.

Sources/StreamChat/Deprecations.swift
```diff
@@ -5,3 +5,206 @@
 import Foundation
 
 /// - NOTE: Deprecations of the next major release.
+
+public extension UserPresenceChangedEvent {
+    @available(*, deprecated, message: "Use user.id")
+    var userId: UserId { user.id }
+}
```
nuno-vieira, line 12: Should we also mention that `user` is now accessible?

Sources/StreamChat/Deprecations.swift
```diff
@@ -5,3 +5,206 @@
...
+public extension TypingEvent {
+    @available(*, deprecated, message: "Use user.id")
+    var userId: UserId { user.id }
+}
+
+@available(*, deprecated, renamed: "NotificationInviteRejectedEvent")
+public typealias NotificationInviteRejected = NotificationInviteRejectedEvent
+
+@available(*, deprecated, renamed: "NotificationInviteAcceptedEvent")
+public typealias NotificationInviteAccepted = NotificationInviteAcceptedEvent
```
nuno-vieira, line 210: We need better deprecation message. We need to mention that the whole model is now available

Sources/StreamChat/WebSocketClient/EventMiddlewares/EventConverterMiddleware.swift
```diff
@@ -0,0 +1,16 @@
+//
+// Copyright © 2021 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+/// The middleware responsible for converting event DTOs to events.
+struct EventConverterMiddleware: EventMiddleware {
```
nuno-vieira, line 8: It is probably a better and more clear name 👍 
```suggestion
/// The middleware responsible for converting event DTOs to events.
struct EventDTOConverterMiddleware: EventMiddleware {
```

Sources/StreamChat/WebSocketClient/EventMiddlewares/EventConverterMiddleware_Tests.swift
```diff
@@ -0,0 +1,71 @@
...
+            createdAt: .unique,
+            payload: [:]
+        )
+        
+        // Feed event to middleware
+        let result = middleware.handle(event: event, session: database.viewContext)
+        
+        // Assert
+        XCTAssertEqual(result as! UnknownEvent, event)
+    }
```
nuno-vieira, line 70: Are these tests even running? 🤔 The func shouldn't start with `test `prefix?

Sources/StreamChat/WebSocketClient/Events/ChannelEvents.swift
```diff
@@ -4,60 +4,210 @@
...
 public struct ChannelUpdatedEvent: ChannelSpecificEvent {
-    public let cid: ChannelId
+    /// The identifier of updated channel.
+    public var cid: ChannelId { channel.cid }
+    
+    /// The updated channel.
+    public let channel: ChatChannel
+    
+    /// The user who updated the channel.
+    public let user: ChatUser?
```
nuno-vieira, line 16: I think in some events it makes sense to name `updatedBy`  and `deletedBY` like you mentioned before, it provides better context 🤔

Sources/StreamChat/WebSocketClient/Events/ChannelEvents.swift
```diff
@@ -4,60 +4,210 @@
...
-    public let cid: ChannelId
-    let payload: Any
+    /// The identifier of deleted channel.
+    public var cid: ChannelId { channel.cid }
+    
+    /// The truncated channel.
+    public let channel: ChatChannel
+    
+    /// The user who truncated a channel.
+    public let user: ChatUser?
```
nuno-vieira, line 105: truncatedBy?

Sources/StreamChat/WebSocketClient/Events/MemberEvents.swift
nuno-vieira: I agree with addedBy

Sources/StreamChat/WebSocketClient/Events/MemberEvents.swift
```diff
@@ -4,41 +4,128 @@
...
+        )
+    }
 }
 
-public struct MemberUpdatedEvent: MemberEvent, ChannelSpecificEvent, EventWithPayload {
-    public let memberUserId: UserId
+/// Triggered when a channel member is updated.
+public struct MemberUpdatedEvent: MemberEvent, ChannelSpecificEvent {
+    /// The user who updated a member.
+    public let user: ChatUser
```
nuno-vieira, line 55: updatedBy?

Sources/StreamChat/WebSocketClient/Events/MemberEvents.swift
```diff
@@ -4,41 +4,128 @@
...
+        )
+    }
 }
 
-public struct MemberRemovedEvent: MemberEvent, ChannelSpecificEvent, EventWithPayload {
-    public var memberUserId: UserId
+/// Triggered when a member is removed from a channel.
+public struct MemberRemovedEvent: MemberEvent, ChannelSpecificEvent {
+    /// The user who stopped being a member.
+    public let user: ChatUser
```
nuno-vieira, line 100: removedUser? 🤔

Sources/StreamChat/WebSocketClient/Events/MessageEvents.swift
nuno-vieira: What if it is like a really old message that was updated? 🤔 We might not have the channel in the DB in that case right?

nuno-vieira (approved):
Overall LGTM 👌 Great Job! Just added some minor comments 🙂

Sources/StreamChat/WebSocketClient/Events/ChannelEvents.swift
tbarbugli: let's not rename fields, this makes REST docs less + makes it easy to be inconsistent between SDKs

Sources/StreamChat/WebSocketClient/Events/MemberEvents.swift
tbarbugli: same comment on the rename above

Sources/StreamChat/WebSocketClient/Events/ChannelEvents.swift
nuno-vieira: Ah I see... yes that also makes sense 👍 If other SDKs are like this, then let's keep it that way. Will resolve my comments 👍

Sources/StreamChat/WebSocketClient/Events/MessageEvents.swift
evsaev: From what I know, we do not get `message.new/message.updated/message.deleted` unless user starts watching a channel by the API call which returns the channel payload in the response (which we save to db).

Sources/StreamChat/WebSocketClient/EventMiddlewares/EventConverterMiddleware_Tests.swift
evsaev: 🤦

evsaev: good catch, let me fix this 💣

Sources/StreamChat/WebSocketClient/EventMiddlewares/MemberEventMiddleware.swift
```diff
@@ -34,38 +40,25 @@ struct MemberEventMiddleware: EventMiddleware {
...
         }
 
         return event
     }
 }
+
+private extension ChannelDTO {
+    func markNeedsRefreshQueries() {
+        needsRefreshQueries = true
+        queries.removeAll()
```
evsaev, line 62: @b-onc this no longer makes sense, right? 🤔 Since we switched to `ChannelListController.shouldListUpdatedChannel`, we need to trigger a channel update

Sources/StreamChat/WebSocketClient/Events/MessageEvents.swift
evsaev: @nuno-vieira the issue I can imagine is related to race conditions. When we start watching a channel and `message.new/message.updated/message.deleted` come before we received the REST API response with a channel, there's no channel in db -> `eventDTO -> event` conversion fails -> event is not published.

Is it OK not to publish the event in this case? From customer perspective, unless `watch` completion is called the watching hasn't yet started so they do not expect to be notified about the channel events so it seems fine.

cc @tbarbugli @b-onc @adamrushy

Sources/StreamChat/WebSocketClient/Events/MessageEvents.swift
evsaev: ✅ Made the `user` optional

Sources/StreamChat/Deprecations.swift
evsaev: ✅ Updated deprecation messages

Sources/StreamChat/WebSocketClient/EventMiddlewares/EventConverterMiddleware_Tests.swift
evsaev: ✅ Fixed

Sources/StreamChat/WebSocketClient/EventMiddlewares/EventConverterMiddleware.swift
evsaev: ✅ Renamed

Sources/StreamChat/WebSocketClient/EventMiddlewares/MemberEventMiddleware.swift
evsaev: ✅ Update the middleware to trigger channel update

PR #1460, opened 2021-09-17
[CIS-1157] Rename ChannelListController linking delegate funcs

b-onc:
And disable `NewChannelQueryUpdater` since delegate funcs take care of this now


evsaev (approved):
LGTM

Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift
```diff
@@ -356,33 +357,27 @@ public protocol ChatChannelListControllerDelegate: DataControllerStateDelegate {
...
-    
-    /// The controller asks the delegate if the newly updated `ChatChannel` should be unlinked from this Controller's query.
-    /// Defaults to `false`
-    /// - Parameters:
-    ///   - controller: The controller,
-    ///   - shouldUnlinkUpdatedChannel: The newly updated `ChatChannel` instance. This instance is linked to the controller's query.
-    func controller(
-        _ controller: ChatChannelListController,
-        shouldUnlinkUpdatedChannel channel: ChatChannel
+        shouldListUpdatedChannel channel: ChatChannel
```
nuno-vieira, line 380: If these are breaking changes, shouldn't we deprecate them first? We are already in production 4.0.

b-onc: This wasn't released, only on main

nuno-vieira: Hum okay 🤔 Then if we fix the Channel List tests, it should be LGTM

tbarbugli (approved):

Issue #1447, opened 2021-09-13
`unreadCountPublisher` publishes only an initial value (and always zero).


cjmconie:
## What did you do?

Subscribe to `currentUserController`'s `unreadCountPublisher` and receive only initial values of 0. 

To reproduce:

Inside `DemoAppCoordinator` update `presentChat(userCredentials, channelID)` implementation of `connectUser` as follows:
```swift
        ChatClient.shared.connectUser(
            userInfo: .init(id: userCredentials.id, name: userCredentials.name, imageURL: userCredentials.avatarURL),
            token: token
        ) { error in
            if let error = error {
                log.error("connecting the user failed \(error)")
                return
            }
            self.setupRemoteNotifications()
            
            // ----------TEST-----------
            // print out the non-publisher values for comparison
            print("[TEST] unreadCount: \(ChatClient.shared.currentUserController().unreadCount)")
            
            // synchronise (as suggested here: https://github.com/GetStream/stream-chat-swift/issues/1224#issuecomment-876388294)
            ChatClient.shared.currentUserController().synchronize { [weak self] error in
                guard let self = self else { return }
                if let error = error {
                    print("[TEST] synchronize error: \(error)")
                }
                
                // This does not work because the data controller is in state `.initialized` and so `startObservingIfNeeded` returns early?
                // Seems related to: https://github.com/GetStream/stream-chat-swift/issues/1224
                ChatClient.shared.currentUserController().unreadCountPublisher.sink { completion in
                    print("[TEST] unreadCountPublisher completion: \(completion)")
                } receiveValue: { unreadCount in
                    print("[TEST] unreadCountPublisher value: \(unreadCount)")
                }.store(in: &self.cancelBag)
            }
            // ----------TEST-----------
        }
```

## What did you expect to happen?
The `unreadCountPublisher` publisher to publish appropriate values. (Same applies to `currentUserChangePublisher`)

## What happened instead?
The publishers fire with initial values of 0. The non-publisher version `ChatClient.shared.currentUserController().unreadCount` outputs the correct values.

## GetStream Environment
**GetStream Chat version:** 4.0.0
**GetStream Chat frameworks:** StreamChat, StreamChatUI, StreamChatClient, StreamChatCore
**iOS version:** 14.0
**Swift version:** 5.4.2
**Xcode version:** 13 beta 5
**Device:** Any

tbarbugli:
[...]

PR #1462, opened 2021-09-19
CIS-1172 Do not make a weak ref to self, too easy to get confused

tbarbugli:
Removes the `weak self` to avoid issues like these: https://github.com/GetStream/stream-chat-swift/issues/1447


evsaev (approved):
LGTM 👍

evsaev:
Would be really nice to have a unit test + `CHANGELOG` entry for the change made👍

PR #1467, opened 2021-09-22
Local notifications support (showcase for CIS-1081)

evsaev:
**This PR** is a showcase for #1459. It updates channel list to listen to `MessageNewEvent` event and show local notification with data from exposed models (`event.message` & `event.channel`).

### DIFF
--- a/DemoApp/ChatPresenter.swift
+++ b/DemoApp/ChatPresenter.swift
@@ -195,22 +195,48 @@ class DemoChatChannelListRouter: ChatChannelListRouter {
     }
 }
 
-class DemoChannelListVC: ChatChannelListVC {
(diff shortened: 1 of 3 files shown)


Sources/StreamChat/WebSocketClient/Events/Event.swift
```diff
@@ -30,6 +30,12 @@ protocol ChannelSpecificEvent: Event {
     var cid: ChannelId { get }
 }
 
+/// A bounding protocol for all events that have unread counts.
+public protocol HasUnreadCount: Event {
```
adamrushy, line 34: The name "HasUnreadCount" feels slightly odd as a protocol, usually, this is on a boolean object.

PR #1468, opened 2021-09-22
[CIS-1181] Implement uploadFile API

b-onc:
It'll allow uploading a file to CDN without having to upload an attachment to a message

### DIFF
--- a/Sources/StreamChat/Models/Attachments/ChatMessageFileAttachment.swift
+++ b/Sources/StreamChat/Models/Attachments/ChatMessageFileAttachment.swift
@@ -33,6 +33,16 @@ public struct FileAttachmentPayload: AttachmentPayload {
             .flatMap { try? JSONEncoder.stream.encode($0) }
             .flatMap { try? JSONDecoder.stream.decode(T.self, from: $0) }
     }
+    
+    /// Creates `FileAttachmentPayload` instance.
+    ///
+    /// Use this initializer if the attachment is already uploaded and you have the remote URLs.
+    public init(title: String?, assetRemoteURL: URL, file: AttachmentFile, extraData: [String: RawJSON]?) {
+        self.title = title
+        assetURL = assetRemoteURL
+        self.file = file
+        self.extraData = extraData
+    }
 }
 
 extension FileAttachmentPayload: Hashable {}
@@ -58,7 +68,7 @@ extension FileAttachmentPayload: Decodable {
         
         self.init(
             title: try container.decodeIfPresent(String.self, forKey: .title),
-            assetURL: try container.decode(URL.self, forKey: .assetURL),
+            assetRemoteURL: try container.decode(URL.self, forKey: .assetURL),
             file: try AttachmentFile(from: decoder),
             extraData: try Self.decodeExtraData(from: decoder)
         )
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 # Upcoming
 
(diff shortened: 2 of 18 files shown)


Sources/StreamChat/Models/Attachments/ChatMessageFileAttachment.swift
```diff
@@ -33,6 +33,13 @@ public struct FileAttachmentPayload: AttachmentPayload {
             .flatMap { try? JSONEncoder.stream.encode($0) }
             .flatMap { try? JSONDecoder.stream.decode(T.self, from: $0) }
     }
+    
+    public init(title: String?, assetURL: URL, file: AttachmentFile, extraData: [String: RawJSON]?) {
```
evsaev, line 37: Opening an `init` looks safe but it actually opens the door for a miss-use 🥲 When `***AttachmentPayload` is instantiated manually there's a chance it is passed to `AnyAttachmentPayload.init` designed for  **custom non-uploadable** attachments:
```swift
init<Payload: AttachmentPayload>(payload: Payload) {
        self.init(
            type: Payload.type,
            payload: payload,
            localFileURL: nil
        )
    }
```

instead of one designed for **built-in uploadable attachments**:
```swift
init(
        localFileURL: URL,
        attachmentType: AttachmentType,
        extraData: Encodable? = nil
    ) { ... }
```

Attachment without `localURL` are not picked up for uploading and will be sent to the backend as is (with local URLs) 🥲

tbarbugli (commented):
Can you include code examples about this? It is not 100% obvious why this is not going via the CDN class

Sources/StreamChat/Models/Attachments/ChatMessageFileAttachment.swift
b-onc: IMHO that's safe, developers don't need to create built-in payload instances unless they handle the uploads themselves.
```swift
override func addAttachmentToContent(from url: URL, type: AttachmentType) throws {
        if pickingEncryptedMedia {
            pickingEncryptedMedia = false
            // Encrypt the attachment
            // Upload the attachment
            channelController?.uploadFile(type: type, localFileURL: url, completion: { result in
                do {
                    let uploadedURL = try result.get()
                    // Append already uploaded attachment
                    let attachment = AnyAttachmentPayload(
                        payload: ImageAttachmentPayload(
                            title: nil,
                            imageURL: uploadedURL,
                            extraData: nil
                        )
                    )
                    DispatchQueue.main.async {
                        self.content.attachments.append(attachment)
                    }
                } catch {
                    // Handle error
                    log.error("Failed uploading attachment: \(error)")
                }
            })
        } else {
            try super.addAttachmentToContent(from: url, type: type)
        }
    }
    ```

**Comment by b-onc** [comment]
@tbarbugli It's actually going through the CDN class, just via `ChannelController -> APIClient -> CDN`, we don't want to expose CDN by itself

File: Sources/StreamChat/Models/Attachments/ChatMessageFileAttachment.swift (hunk 1 above)
Comment (evsaev) [reply Sources/StreamChat/Models/Attachments/ChatMessageFileAttachment.swift:37]: They do not need but the API allows it, that's what I was saying 🙂

That's the correct way:
```swift
let attachment = try AnyAttachmentPayload(localFileURL: url, attachmentType: .image)
content.attachments.append(attachment)
```

That's the 💣 
```swift
let attachment = AnyAttachmentPayload(
    payload: ImageAttachmentPayload(
         title: nil, 
         imageURL: url, 
         extraData: nil
     )
)
content.attachments.append(attachment)
```

evsaev (dismissed):
Looks good 👍

tbarbugli (approved):

PR #1472, opened 2021-09-22
Fix unread count logic

tbarbugli:
Local unread count logic was not the same as the one used by server-side


adamrushy (approved):
Looks great 👍🏼

PR #1476, opened 2021-09-23
[CIS-1187] Fix channel list showing `currentUser.mutedChannels` initially

evsaev:
This **PR** fixes the bug when all updated/inserted channels are added to the channel list before the underlaying query is synced. The fix updates `ChannelListController` to handle unlinked channels only when the `controller.synchronize` is successfully completed.

### DIFF
--- a/Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift
+++ b/Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift
@@ -214,6 +214,11 @@ public class ChatChannelListController: DataController, DelegateCallable, DataSt
     }
     
     private func handleUnlinkedChannels(_ changes: [ListChange<ChatChannel>]) {
+        guard state == .remoteDataFetched else {
+            log.debug("Ignoring inserted/updated unlinked channels due to query \(query) not being synced.")
+            return
+        }
+        
         let channels = changes.compactMap { change -> ChatChannel? in
             switch change {
             case let .insert(channel, _):
@@ -241,13 +246,17 @@ public class ChatChannelListController: DataController, DelegateCallable, DataSt
     private func link(channels: [ChatChannel]) {
         guard !channels.isEmpty else { return }
         client.databaseContainer.write { session in
+            guard let queryDTO = session.channelListQuery(filterHash: self.query.filter.filterHash) else {
+                log.debug("Channel list query has not yet created \(self.query)")
+                return
+            }
+            
             for channel in channels {
                 guard let channelDTO = session.channel(cid: channel.cid) else {
                     log.error("Channel \(channel.cid) cannot be found in database.")
                     continue
                 }
-                let query = session.saveQuery(query: self.query)
-                query.channels.insert(channelDTO)
+                queryDTO.channels.insert(channelDTO)
             }
         }
     }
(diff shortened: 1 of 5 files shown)


Sources/StreamChat/Database/DatabaseSession.swift
```diff
@@ -167,6 +167,10 @@ protocol ChannelDatabaseSession {
         query: ChannelListQuery?
     ) throws -> ChannelDTO
     
+    /// Loads channel list query with the given filter hash from the database.
+    /// - Parameter filterHash: The filter hash.
+    func channelListQuery(filterHash: String) -> ChannelListQueryDTO?
+
```
nuno-vieira, line 173: Why was this added? 🤔 I don't see it being used

Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift
```diff
@@ -241,27 +241,35 @@ public class ChatChannelListController: DataController, DelegateCallable, DataSt
...
+            }
+            
             for channel in channels {
                 guard let channelDTO = session.channel(cid: channel.cid) else {
                     log.error("Channel \(channel.cid) cannot be found in database.")
                     continue
                 }
-                let query = session.saveQuery(query: self.query)
-                query.channels.remove(channelDTO)
+                queryDTO.channels.insert(channelDTO)
```
nuno-vieira, line 272: Could we somehow add tests to make sure there are no regressions in the future?

b-onc: Used in the `guard` statements

b-onc (dismissed):
Interesting find! This can use a test case, and CHANGELOG since it's released

Sources/StreamChat/Database/DatabaseSession.swift
evsaev: Used in `ChannelListController.link/unlink` funcs 🙂

Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift
evsaev: ✅ Added tests

evsaev:
> Interesting find! This can use a test case, and CHANGELOG since it's released

✅ Added tests and CHANGELOG entry

Sources/StreamChat/Controllers/ChannelListController/ChannelListController_Tests.swift
```diff
@@ -255,42 +255,28 @@ class ChannelListController_Tests: XCTestCase {
...
-            }
-            
-            func controller(_ controller: ChatChannelListController, shouldListUpdatedChannel channel: ChatChannel) -> Bool {
-                shouldListUpdatedChannel(channel)
-            }
-        }
-        
-        // Simulate `synchronize` call
-        controller.synchronize()
+    func test_newChannel_callsListHook_whenSynchronized() throws {
```
b-onc, line 258: We should make sure this hook is not called before `sync` is called

evsaev: There's a `test_unlinkedChannels_doNotTriggerHooks_whenNotSynchronized`. 

And we can handle updates of linked channel even when a query is not yet synced. It is covered by `test_linkedChannels_doesTriggerUnlinkHook_whenNotSynchronized`

b-onc (approved):

nuno-vieira (approved):
LGTM ✅

tbarbugli (changes requested):
muted channels are not showed anymore but "mute channel" action from the demo app is not working (channel list, slide, mute channel)

PR #1480, opened 2021-09-27
[CIS-1195] Dynamic height for the composer attachment previews

nuno-vieira:
Previously the size of the scroll view containing the attachment previews on the composer had a fixed height. The height should be dynamic so that the size is determined by the size of the attachment previews.


adamrushy (approved):
Looks good 👍🏼
Do we need to update the changelog?

b-onc (dismissed):

b-onc:
CHANGELOG please

PR #1481, opened 2021-09-27
New folder structure to contain Example apps.

adamrushy:
This PR will move all the sample apps inside an `Example` folder. This will help us understand and make more clearly which sample projects to use.

Over time, it will be easier to maintain the projects we want to keep. I have also fixed a small deprecation within the MessengerClone project.


nuno-vieira (dismissed):
LGTM! ✅

nuno-vieira (commented):
It seems we need to update the `Scripts/run-linter.sh` script, since the Sample folder is now not on the root anymore. Let me know if you need any help with this 👍

nuno-vieira (approved):
LGTM! ✅

PR #1485, opened 2021-09-27
Fix composer horizontal attachment after adding vertical attachment

nuno-vieira:
This fixes a UI glitch when adding a horizontal attachment preview in the composer after adding a vertical attachment.

Steps to reproduce:
1. Add a vertical attachment (file)
2. Remove vertical attachment (file)
3. Add a horizontal attachment (image or video)

This PR also refactors the way horizontal and vertical attachments are layout in a manner that it will be easier to support multiple attachments at once in the future.


adamrushy (approved):
Looks good 👍

Issue #1487, opened 2021-09-28
Chat client APIs randomly fail due to JSON decoding


basememara:
## What did you do?
Called `chatClient.channelListController(...)`

## What did you expect to happen?
Data retrieved from the chat client with no errors.

## What happened instead?
Randomly no data is returned from the chat client and get the following error in the console:

```
An error occured while fetching the channel list | Error: dataCorrupted(Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "channels", intValue: nil), _JSONKey(stringValue: "Index 0", intValue: 0), CodingKeys(stringValue: "messages", intValue: nil), _JSONKey(stringValue: "Index 0", intValue: 0), MessagePayloadsCodingKeys(stringValue: "user", intValue: nil), UserPayloadsCodingKeys(stringValue: "last_active", intValue: nil)], debugDescription: "Invalid date: 2021-09-28T08:17:18.030196604Z", underlyingError: nil)) | Context: ["app_state": "active"]
```

## GetStream Environment
**GetStream Chat version:** 4.0.2
**GetStream Chat frameworks:** StreamChat
**iOS version:** 15
**Swift version:** 5.5
**Xcode version:** 13
**Device:** simulator

## Additional context
See PR #1486 for further details and fix.

tbarbugli:
fixed on main branch (thank you!)

PR #1486, opened 2021-09-28
Fix date formatter for decoder thread-safety

basememara:
# Submit a pull request

## CLA

- [x] I have signed the [Stream CLA](https://docs.google.com/forms/d/e/1FAIpQLScFKsKkAJI7mhCr7K9rEIOpqIDThrWxuvxnwUq2XkHyG154vQ/viewform) (required).
- [x] The code changes follow best practices
- [x] Code changes are tested (add some information if not applicable)

## Description of the pull request
The global static date formatter, `JSONDecoder.default` and `JSONDecoder.stream`, are reused and the `dateFormatter` property is updated from multiple threads. This causes the JSON decoding to randomly fail:

```
An error occured while fetching the channel list | Error: dataCorrupted(Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "channels", intValue: nil), _JSONKey(stringValue: "Index 0", intValue: 0), CodingKeys(stringValue: "messages", intValue: nil), _JSONKey(stringValue: "Index 0", intValue: 0), MessagePayloadsCodingKeys(stringValue: "user", intValue: nil), UserPayloadsCodingKeys(stringValue: "last_active", intValue: nil)], debugDescription: "Invalid date: 2021-09-28T08:17:18.030196604Z", underlyingError: nil)) | Context: ["app_state": "active"]
```

This fix separates each `DateFormatter` as its own static instance so the mutable properties aren't changed on the fly. See the included test fail without this fix and after.

PR #1488, opened 2021-09-28
Do not share DateFormatter

tbarbugli:
Patches https://github.com/GetStream/stream-chat-swift/issues/1487

### DIFF
--- a/Sources/StreamChat/Utils/Codable+Extensions.swift
+++ b/Sources/StreamChat/Utils/Codable+Extensions.swift
@@ -139,12 +139,16 @@ extension DateFormatter {
             return gmtDateFormatter.string(from: date)
         }
 
-        private static let gmtDateFormatter: DateFormatter = {
+        private static var locale: Locale = Locale(identifier: "en_US_POSIX")
+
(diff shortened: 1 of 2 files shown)


Sources/StreamChat/Utils/Codable+Extensions.swift
```diff
@@ -139,12 +139,16 @@ extension DateFormatter {
...
+
+        private static var gmtDateFormatter: DateFormatter {
             let formatter = DateFormatter()
-            formatter.locale = Locale(identifier: "en_US_POSIX")
-            formatter.timeZone = TimeZone(secondsFromGMT: 0)
+            formatter.locale = locale
+            formatter.timeZone = timeZone
             return formatter
-        }()
+        }
```
nuno-vieira, line 151: This will create a new DateFormatter every time. It is recommended by Apple that we don't do this since it is expensive to create a DateFormatter, especially if we are creating it every cell in a table view for example. 

The https://github.com/GetStream/stream-chat-swift/pull/1486 solution seems pretty solid to me

nuno-vieira (changes requested):

PR #1491, opened 2021-09-29
Fix fastlane testflight lane

b-onc:
There were a number of issues:
1. Dmitriy's cert was not valid anymore, we needed to manually get rid of it in ios-certificates repo and create new ones.
2. match_me had dmitriy's email hardcoded, not necessary
3. DemoAppPush needed profiles generated by match
4. DemoAppPush needed to link to StreamChat but not embed it, since DemoApp already embeds it
5. DemoAppPush needs version synced with DemoApp

#skip_changelog


adamrushy (approved):
We did it, yay! 🎉

PR #1495, opened 2021-09-30
Introduce Carthage compatibility

b-onc:
Let's see what CI says


nuno-vieira:
@b-onc Don't forget to update the README.md that we now support Carthage 🙂

b-onc:
@nuno-vieira README doesn't include installation instructions so i updated `overview.md`

docusaurus/docs/iOS/basics/overview.md
```diff
@@ -108,8 +108,18 @@ The above command will generate the **MyProject.xcworkspace** file automatically
...
+github "GetStream/stream-chat-swift" ~> 4.0
+```
+
+and run:
+
+```bash
+carthage update --use-xcframeworks --platform iOS
+```
+
+go into your project, in the General settings tab, in the Frameworks, Libraries, and Embedded Content section, drag and drop each XCFramework you use from the Carthage/Build folder on disk.
```
nuno-vieira, line 125: Maybe worth mentioning that if you only want the LLC, you don't need to link the StreamChatUI

b-onc: "...drag and drop each XCFramework *you use*..." is enough IMHO. If a developer is only using LLC, they'd know not to include UI framework

nuno-vieira (approved):
LGTM

PR #1500, opened 2021-10-01
Fix Firebase distribution workflow

b-onc:
...and switch to 1st party "cancel concurrent workflows" solution

#skip_changelog


.github/workflows/distribute-test-build.yml
```diff
@@ -6,6 +6,10 @@ on:
       - main
   workflow_dispatch:
 
+concurrency: 
+  group: ${{ github.head_ref }}
+  cancel-in-progress: true
```
nuno-vieira, line 11: Finally 🎉

adamrushy (approved):
Great 👍

.github/workflows/smoke-checks.yml
```diff
@@ -10,6 +10,10 @@ on:
   release:
     types:
       - created
+
+concurrency: 
+  group: ${{ github.head_ref }}
+  cancel-in-progress: true
```
nuno-vieira, line 16: I know we are not using them right now, but maybe let's add this to the full checks as well? 🤔

b-onc: Full checks are not run on pushes to main, this is not required for it. same applies to testflight

PR #1501, opened 2021-10-01
Make `messageComposerBottomConstraint` public

nuno-vieira:
Some customers require this property to be public and it doesn't make sense to make it private anyways.


b-onc (approved):
CHANGELOG

PR #1509, opened 2021-10-05
Take video and image cell types from components

evsaev:
**This PR** fixes the issue with `Image/VideoAttachmentGalleryCell` not taken from component.


adamrushy (approved):
Looks good 👍

PR #1510, opened 2021-10-05
Make it possible to hook a video asset before loading

evsaev:


Sources/StreamChatUI/Utils/VideoPreviewLoader.swift
```diff
@@ -6,21 +6,32 @@ import AVKit
...
+
+public extension VideoLoading {
+    func videoAsset(at url: URL) -> AVURLAsset {
+        .init(url: url)
+    }
 }
 
-/// The `VideoPreviewLoader` implemenation used by default.
-final class DefaultVideoPreviewLoader: VideoPreviewLoader {
+/// The `VideoLoading` implemenation used by default.
```
adamrushy, line 29: Sorry for the nitpick but 'implementation' is the correct spelling here 👍

evsaev: ✅ Fixed

Sources/StreamChatUI/Components.swift
```diff
@@ -78,9 +78,9 @@ public struct Components {
     
     /// Object responsible for providing resizing operations for `UIImage`
     public var imageProcessor: ImageProcessor = NukeImageProcessor()
-
-    /// The object that loads previews for video attachments.
-    public var videoPreviewLoader: VideoPreviewLoader = DefaultVideoPreviewLoader()
+    
+    /// The object that loads previews for video attachments and resloves assets for a video player.
```
nuno-vieira, line 82: I think this is more succinct and enough 👍 
```suggestion
    /// The object responsible for loading video attachments.
```

Sources/StreamChatUI/Utils/VideoLoading.swift
```diff
@@ -6,21 +6,32 @@ import AVKit
...
+
+public extension VideoLoading {
+    func videoAsset(at url: URL) -> AVURLAsset {
+        .init(url: url)
+    }
 }
 
-/// The `VideoPreviewLoader` implemenation used by default.
-final class DefaultVideoPreviewLoader: VideoPreviewLoader {
+/// The `VideoLoading` implementation used by default.
```
nuno-vieira, line 29:
```suggestion
/// The dafault `VideoLoading` implementation.
```

Sources/StreamChatUI/Utils/VideoLoading.swift
```diff
@@ -6,21 +6,32 @@ import AVKit
...
+open class StreamVideoLoader: VideoLoading {
     private let cache: Cache<URL, UIImage>
     
-    init(countLimit: Int = 50) {
-        cache = .init(countLimit: countLimit)
+    public init(cachedVideoPreviewsCountLimit: Int = 50) {
```
nuno-vieira, line 33: I think the required init should be `public init(cache: Cache<URL, UIImage> = .init(countLimit: 50)`. This makes it more testable and clear what is the dependencies of `StreamVideoLoader`.

Has a convenience init we can have the `public init(cachedVideoPreviewsCountLimit: Int = 50)` still 👍

nuno-vieira (commented):
Overall LGTM, just some small comments 🙂

Sources/StreamChatUI/Components.swift
evsaev: ✅

Sources/StreamChatUI/Utils/VideoLoading.swift
evsaev: ✅

Sources/StreamChatUI/Utils/VideoLoading.swift
evsaev: I agree with both args 👍 But this would require opening `Cache` type, right? 🤔 And since `Cache` is very basic it'd also be nice to hide it behind a protocol and let customers provide their own caching. But because of `Cache` being generic it will involve some extra work. And also increase the public-facing APIs with smth not directly related to chats. Not sure if it's the right move.

nuno-vieira: Ah I see, nevermind then, at least for now, we can leave it like this 👍

nuno-vieira (approved):
LGTM! ✅

PR #1517, opened 2021-10-06
Fix Stream's Playground

nuno-vieira:
This fixes the Stream playground that was not compiling. 
The Playground is very useful to quickly test LLC or even UI Components.

image


adamrushy (approved):
Looks good 👍

PR #1519, opened 2021-10-06
Fix `Components` being not retrievable

evsaev:
**This PR** fixes:
- broken responder chain for custom `Components` which leads to invalid UI/non-UI component being used
- `ChatChannelVC_Tests.test_onlyEmojiMessageAppearance` test failing consistently on the CI

### ChatMessageCell

The responder chain fails on **ChatMessageCell** affecting all views down the hierarchy. The root cause is that we instantiate the cell + message content view in `tableView(_ tableView: UITableView, cellForRowAt)` but at this point the cell has no superview yet so this part of the chain is ignored: **ChatMessageListView -> ChatMessageListVC -> ChatChannelVC/ChatThreadVC**

### ChatChannelVC/ChatThreadVC

Before **ChatMessageListVC/ComposerVC** are added as children, the `child.view` is access to turn OFF auto-resizing mask which loads the view and runs **setUp/setupLayout** where we lazily instantiate all the subviews but because a future child has no parent yet invalid components are used.


nuno-vieira:
It seems that some components still have problems 🤔

evsaev:
> It seems that some components still have problems 🤔

`ChatMessageContentView` was updated to trigger the layout when the it's added as a subview. Two tests were red because the content view was not added to the view hierarchy. It's fixed now so let's see if CI is green🤞

nuno-vieira (approved):
LGTM

nuno-vieira:
@evsaev I still find it weird, why this happens. Customers are not complaining they can't customize these components. Also, we have these components customized in our sample apps. Why is this happening only on the tests?

PR #1523, opened 2021-10-06
Fix `imageURL` encoding during `connectUser`

b-onc:
In turn, user images were not correctly displayed


nuno-vieira (changes requested):
Please add a video that shows this working properly, since it doesn't seem it fixed.

Issue #1475, opened 2021-09-23
Fatal error MessageDTO.swift 


BarragaoA:
## What did you do?
Hello, i'm registered to EventsControllerDelegate in order to send local push when a new message happens.

## What happened instead?
Do not know why and how to reproduce the issue but sometimes when we are launching the app and triggering an event a crash occurs. I can not find a way to systematically reproduce it.

Thanks for your help.

## GetStream Environment
**GetStream Chat version:** 4.0.1
**GetStream Chat frameworks:** StreamChat, StreamChatUI
**iOS version:**14.7.1
**Xcode version:** Xcode 13.0
**Device:** iPhone Xr

## Additional context
Crashlytics infos :

> EXC_BREAKPOINT 0x0000000192b6dacc:

> 'StreamChat/Sources/StreamChat/Database/DTOs/MessageDTO.swift:446: Fatal error: Should never happen, a channel should have been fetched.'

> Crashed: NSManagedObjectContext 0x2800862f0
> 0  libswiftCore.dylib             0x192b6dacc _assertionFailure(_:_:file:line:flags:) + 1532
> 1  libswiftCore.dylib             0x192b6db40 assertionFailure(_:file:line:) + 104
> 2  StreamChat                     0x106a99bdc $s10StreamChat6LoggerC16assertionFailure_12functionName04fileG010lineNumberyypyXK_s12StaticStringVAISutF + 188
> 3  StreamChat                     0x106adaa10 $sSo22NSManagedObjectContextC10StreamChatE11saveMessage7payload3forAC0G3DTOCAC0G7PayloadC_AC9ChannelIdVSgtKF + 6328
> 4  StreamChat                     0x106adc510 $sSo22NSManagedObjectContextC10StreamChat22MessageDatabaseSessionA2cDP04saveF07payload3forAC0F3DTOCAC0F7PayloadC_AC9ChannelIdVSgtKFTW + 24
> 5  StreamChat                     0x106a3d5d4 $s10StreamChat15DatabaseSessionPAAE9saveEvent7payloadyAA0F7PayloadV_tKF + 2336
> 6  StreamChat                     0x106a52df0 $s10StreamChat28EventDataProcessorMiddlewareV6handle5event7sessionAA0C0_pSgAaG_p_AA15DatabaseSession_ptF + 888
[...]

PR #1548, opened 2021-10-14
CIS-1205 Improve event to DB sync for events with a message

tbarbugli:
### 🎯 Goal

Reaction events contain the message object and the `cid` but not the full channel. 

When the SDK receives a reaction event before the channel is stored in the database a log assertion crashes the app when in dev mode.

### 🛠 Implementation

This PR handles this case by not calling DTO's `saveMessage` when the channel is not present.

### 🧪 Testing

Stress test demo app (already done): spam chat with messages and reactions while opening the app to trigger this case

### ☑️ Checklist


tbarbugli:
This PR resolves the assert log from #1475

PR #1582, opened 2021-10-26
[CIS-1183] Add view to show all reactions of a message

nuno-vieira:
### 🔗 Issue Link
CIS-1183

### 🎯 Goal
Add a view to show all reactions of a message and the author of that reaction, when tapping the reactions bubble view of a message.

### 🛠 Implementation
- Adds a new `MessageController.reactions` property that gets all the reactions of the message.
- Adds a new ` MessageController.loadReactions()` function to get reactions from the Backend given a  limit and offset
- Adds a new `MessageController.loadNextReactions()` to paginate the loading of all the reactions. When the user is scrolling all the reactions, it fetches more reactions until all reactions have been loaded.
- Adds a new UI Component `ChatMessageReactionAuthorsVC` to render all the reactions of the message along with its authors.
- Changes the `ChatMessagePopupVC` to support rendering the reaction authors.
- Adds new reaction documentation and also restructures the reactions docs.

### 🧪 Testing
1. Open a Channel
2. Tap on the reactions view of a message
3. A popup view controller should be presented with all the reactions.

### 🎨 Changes
**Video displaying all the edge cases of the `ChatMessageReactionAuthorsVC`:**

https://user-images.githubusercontent.com/12814114/138963810-9f1808c4-4e67-41f5-afcd-7c11720c26f8.mp4

### ☑️ Checklist

- [x] New code is covered by UI Tests


docusaurus/docs/iOS/components/reactions.md
```diff
@@ -45,110 +34,178 @@ let customReactions: [MessageReactionType: ChatMessageReactionAppearanceType] =
...
-| ![Default](../assets/default-message-reaction.png)  | ![Custom](../assets/custom-message-reaction.png)  |
+Here is the result for all the customizations done to the reaction picker view:
+
+| Default Picker | Custom Picker |
+| ------------- | ------------- |
+| ![Default Picker](../assets/message-reactions-tap.png)  | ![Custom Picker](../assets/message-reactions-picker-custom.png)  |
+
+## Reaction Authors
+
+TODO
```
nuno-vieira, line 211: Only this part of the documentation re-write that is missing. After that the PR is 100% complete

Sources/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -23,18 +23,24 @@ class MessageDTO: NSManagedObject {
...
-    @NSManaged var reactions: Set<MessageReactionDTO>
     @NSManaged var attachments: Set<AttachmentDTO>
     @NSManaged var quotedMessage: MessageDTO?
     @NSManaged var searches: Set<MessageSearchQueryDTO>
 
+    // We separate reactions from own reactions so that when paginating reactions,
+    // the own reactions don't break the pagination logic. This is why ownReactions
+    // are stored in a different property and as JSON Data.
+    @NSManaged var reactions: Set<MessageReactionDTO>
+    @NSManaged var ownReactions: Data
```
evsaev, line 42: Ok, it seems I understand why we have to store our own reactions separately 👍 **All** own reactions are included in the message payload but it's a subset of all reactions and they can be spread around the list. And to not mess up the list of reactions locally and follow the backend's pagination we store them in 2 different fields.

But why did we store own reactions as data and not as a relation? Is it some issue with the fetch request predicate?

Sources/StreamChat/Models/MessageReaction.swift
```diff
@@ -19,9 +19,25 @@ public struct ChatMessageReaction: Hashable {
...
+    /// Custom data
+    public let extraData: [String: RawJSON]
+
+    public init(
+        type: MessageReactionType,
+        score: Int,
+        createdAt: Date,
+        updatedAt: Date,
+        author: ChatUser,
+        extraData: [String: RawJSON]
```
evsaev, line 34: Hm, why is it `public`? 🤔

Sources/StreamChatUI/ChatMessageList/Reactions/ChatMessageReactionAuthorsVC/ChatMessageReactionAuthorsFlowLayout.swift
```diff
@@ -0,0 +1,61 @@
+//
+// Copyright © 2021 Stream.io Inc. All rights reserved.
+//
+
+import UIKit
+
+/// The default `ChatMessageReactionAuthorsVC` collection view flow layout.
+/// It manipulates the cell attributes so that each row is centred.
+class ChatMessageReactionAuthorsFlowLayout: UICollectionViewFlowLayout {
```
evsaev, line 9: shall we open it and add to `Components`?

Sources/StreamChatUI/Components.swift
```diff
@@ -90,28 +90,35 @@ public struct Components {
...
-    /// The View Controller by default used to display long-press menu of the message.
+    /// The view controller that is presented when long-pressing a message.
     public var messagePopupVC: ChatMessagePopupVC.Type = ChatMessagePopupVC.self
 
-    /// The View Controller used for showing detail of a file message attachment.
+    /// A view controller that renders the reaction and it's author avatar for all the reactions of a message.
+    public var reactionAuthorsVC: ChatMessageReactionAuthorsVC.Type = ChatMessageReactionAuthorsVC.self
+
+    /// A view cell that displays an individual reaction author of a message.
+    public var reactionAuthorsCell: ChatMessageReactionAuthorViewCell.Type = ChatMessageReactionAuthorViewCell.self
```
evsaev, line 116:
```suggestion
    public var reactionAuthorCell: ChatMessageReactionAuthorViewCell.Type = ChatMessageReactionAuthorViewCell.self
```

Sources/StreamChatUI/Navigation/ChatMessageListRouter.swift
```diff
@@ -55,6 +54,52 @@ open class ChatMessageListRouter:
...
     }
 
+    /// Shows the detail pop-up for the selected message with all the message reactions presented.
+    ///
+    /// - Parameters:
+    ///   - messageContentView: The selected message content view.
+    ///   - client: The current `ChatClient` instance.
+    open func showReactionsPopUp(
+        messageContentView: ChatMessageContentView,
+        client: ChatClient
```
evsaev, line 64: nit: the `ChatMessageListRouter` was initially used on both channel and thread controllers so it was not possible to make it generic over a single VC type so we went with `NavigationRouter<UIViewController>`. But now with the separated message list it can be `NavigationRouter<ChatMessageListVC>` so there'll be no need to explicitly pass the client 👍

evsaev (approved):
Well done 👍

Sources/StreamChat/APIClient/Endpoints/Payloads/MessagePayloads.swift
```diff
@@ -75,6 +75,7 @@ class MessagePayload: Decodable {
     let latestReactions: [MessageReactionPayload]
     let ownReactions: [MessageReactionPayload]
     let reactionScores: [MessageReactionType: Int]
+    let reactionCounts: [MessageReactionType: Int]
```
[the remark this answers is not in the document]
tbarbugli: image

Sources/StreamChat/Database/DTOs/MessageDTO.swift
tbarbugli: this is actually quite tricky :)

- with this approach we store reactions as core data objects *and* as JSON encoded data (easy to get things wrong)
- stashing latest reactions and own reactions like we do on `main` breaks pagination and probably other stuff

**Alternative Idea**

Store `ownReactions` as a list of reaction ids (Set<String>), this way we can avoid adding another serialization method (JSON) to the mix. Message.ownReactions is enriched at read time (get all reaction where ID IN ?) and the opposite is done at write time (Set<String> makes this trivial) 

Using a separate relationship also work but it is definetely more work and adds no real benefits

Sources/StreamChat/Workers/MessageUpdater.swift
```diff
@@ -198,6 +198,35 @@ class MessageUpdater: Worker {
...
+        let endpoint: Endpoint<MessageReactionsPayload> = .loadReactions(
+            messageId: messageId,
+            pagination: pagination
+        )
+
+        apiClient.request(endpoint: endpoint) { result in
+            switch result {
+            case let .success(payload):
+                self.database.write({ session in
+                    try payload.reactions.forEach { try session.saveReaction(payload: $0) }
```
tbarbugli, line 217: do we actually want to store reactions coming from pagination in the database? sounds like an easy way to create bugs + it does not add a lot of value

tbarbugli: what I mean is, paginating reactions is a very rare use-case and its totally fine to hit the API every time (while slower the advantage is that you avoid potential sync bugs)

tbarbugli (commented):
left some comments in regards to the data model

Sources/StreamChat/APIClient/Endpoints/Payloads/MessagePayloads.swift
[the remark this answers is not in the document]
evsaev: Ok, so scores will be `["clap": 210]` and `["clap": 21]` for counts, right?

Sources/StreamChat/Workers/MessageUpdater.swift
evsaev: With offline storage turned off we actually store everything in memory 🙂 The advantage of having reactions saved to DB is that when another message controller is instantiated it will expose the same list of reactions as the one used to paginate them initially.

We had a similar discussion about message search, right? Following existing patterns has both pros and cons and persisting some info indeed doesn't bring much value. But on the other hand, keeping it consistent before we come up with the new pattern that will apply to all knows cases kinda makes sense.

Sources/StreamChatUI/MessageActionsPopup/ChatMessagePopupVC.swift
```diff
@@ -5,13 +5,14 @@
 import StreamChat
 import UIKit
 
-/// `_ChatMessagePopupVC` is shown when user long-presses a message.
+/// `ChatMessagePopupVC` is shown when user long-presses a message.
 /// By default, it has a blurred background, reactions, and actions which are shown for a given message
 /// and with which user can interact.
 open class ChatMessagePopupVC: _ViewController, ComponentsProvider {
```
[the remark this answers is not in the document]
nuno-vieira: Yes I think it makes sense, we will have time to think about it when we plan a v5 👍

Sources/StreamChat/APIClient/Endpoints/Payloads/MessagePayloads.swift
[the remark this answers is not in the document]
nuno-vieira: @evsaev Like Tomasso is showing there, the reaction scores are different than reaction counts. Reaction scores can be used for a clap system, where a "clap" reaction can have multiple scores.

Sources/StreamChat/Database/DTOs/MessageDTO.swift
nuno-vieira: @evsaev  If we store the own reactions normally in the DB, then the controller won't work because the fetch request gets all the reactions from the DB, it doesn't know which one belongs to own_reactions or not

nuno-vieira: @tbarbugli The only issue I see right now with this approach is that if we change the Reactions structure, the DB migration won't affect the stored JSON... but I guess this won't happen any time soon.

On the other hand, changing the DB approach right now is a bit risky, without planning things properly.

Sources/StreamChat/Models/MessageReaction.swift
nuno-vieira: We need this for testing. Also, the customer might need it as well. I didn't saw much value on creating an internal init just for testing

Sources/StreamChat/Workers/MessageUpdater.swift
nuno-vieira: Yes, I agree there's no value in saving these to the DB. This is the pattern followed across the codebase, so that is why it is done like this. Changing this without properly planning the Offline Refactoring it is a bit risky

Sources/StreamChatUI/MessageActionsPopup/ChatMessagePopupVC.swift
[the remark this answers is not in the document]
nuno-vieira: Removed all "_" from all the comments we had 👍

Sources/StreamChatUI/ChatMessageList/Reactions/ChatMessageReactionAuthorsVC/ChatMessageReactionAuthorsFlowLayout.swift
nuno-vieira: Actually, I made it open but didn't add it to components. It doesn't bring any value. If the customer wants to override the flow layout they can by subclassing the reaction authors VC. And if they change the authors vc to a table view for example, the flow layout is not even needed, so I think it is not necessary to add it to the components, and as far as I can see, we don't have the other flow layout on the components config

Sources/StreamChatUI/Navigation/ChatMessageListRouter.swift
nuno-vieira: Hum, but this might limit the customer right? 🤔 Do we want to do that? @evsaev And this might even be a breaking change, if a customer is using this router without the message list some how 🤔

Sources/StreamChat/Database/DTOs/MessageDTO.swift
evsaev: Yes, that is possible 👍 Let's keep a `UIViewController` as it is now

Sources/StreamChat/Controllers/MessageController/MessageController.swift
```diff
@@ -306,6 +328,68 @@ public extension ChatMessageController {
...
+            return
+        }
+
+        loadReactions(limit: limit, offset: reactions.count) { result in
+            switch result {
+            case let .success(reactions):
+                let currentReactions = Set(self.reactions)
+                for reaction in reactions where !currentReactions.contains(reaction) {
+                    self.reactions.append(reaction)
+                }
```
evsaev, line 354: Hm, if we append new reactions like that the delegate will fire multiple times, right? 🤔 Is it on purpose?

nuno-vieira: Good catch 🤔 Let me check this

Sources/StreamChat/Controllers/MessageController/MessageController.swift
```diff
@@ -306,6 +328,68 @@ public extension ChatMessageController {
...
+        messageUpdater.loadReactions(
+            cid: cid,
+            messageId: messageId,
+            pagination: Pagination(pageSize: limit, offset: offset)
+        ) { result in
+            switch result {
+            case let .success(reactions):
+                self.callback { completion(.success(reactions)) }
+            case let .failure(error):
+                self.callback { completion(.failure(error)) }
```
evsaev, line 389: I wouldn't hide switching to the callback queue in this method but rather do it explicitly where needed. 

The thing is we don't have a reliable way to check on which queue we currently are we have the issue with nested `self.callback { ... }` statements: if the callback queue is `background` and `serial` we get a deadlock:
```swift
extension Controller {
    /// A helper function to ensure the callback is performed on the callback queue.
    func callback(_ action: @escaping () -> Void) {
        if Thread.current.isMainThread {
            if callbackQueue == .main {
                // We perform the callback synchronously
                action()
            } else {
                // We dispatch from the main queue, we must perform
                // the callback must be performed asynchronously
                callbackQueue.async {
                    action()
                }
            }
        } else {
            // Dispatching from a background queue, the callback can be performed synchronously
            callbackQueue.sync { <- here we get a deadlock
                action()
            }
        }
    }
}
```

Getting back to the code, we switch to callback queue in `loadReactions` and then again in `reactions.didSet` when accessing the delegate so for with serial bg callback queue it will 💥

Sources/StreamChat/Database/DTOs/MessageReactionDTO.swift
```diff
@@ -28,6 +28,15 @@ final class MessageReactionDTO: NSManagedObject {
...
 
 extension MessageReactionDTO {
+    static func reactionsFetchRequest(
+        for messageId: MessageId
+    ) -> NSFetchRequest<MessageReactionDTO> {
+        let request = NSFetchRequest<MessageReactionDTO>(entityName: MessageReactionDTO.entityName)
+        request.predicate = NSPredicate(format: "message.id == %@", messageId)
+        request.sortDescriptors = [NSSortDescriptor(keyPath: \MessageReactionDTO.updatedAt, ascending: false)]
+        return request
+    }
```
evsaev, line 38: Do we still need this?

Sources/StreamChat/Models/MessageReaction.swift
```diff
@@ -19,9 +19,25 @@ public struct ChatMessageReaction: Hashable {
...
+    ) {
+        self.type = type
+        self.score = score
+        self.createdAt = createdAt
+        self.updatedAt = updatedAt
+        self.author = author
+        self.extraData = extraData
+    }
```
evsaev, line 42: don't we get exactly the same init auto-generated? 🤔

Sources/StreamChatUI/ChatMessageList/Reactions/ChatMessageDefaultReactionsBubbleView.swift
```diff
@@ -5,6 +5,9 @@
 import StreamChat
 import UIKit
 
+public typealias DefaultChatReactionPickerBubbleView = ChatMessageDefaultReactionsBubbleView
+
+/// This component will be renamed to `DefaultChatReactionPickerBubbleView` in future major versions.
 open class ChatMessageDefaultReactionsBubbleView: ChatReactionPickerBubbleView {
```
evsaev, line 11:
```suggestion
@available(*, deprecated, renamed: "DefaultChatReactionPickerBubbleView") // would be also nice to move it to Deprecations.swift
public typealias ChatMessageDefaultReactionsBubbleView = DefaultChatReactionPickerBubbleView

open class DefaultChatReactionPickerBubbleView: ChatReactionPickerBubbleView {
```

Sources/StreamChatUI/ChatMessageList/Reactions/ChatMessageReactionAuthorsVC/ChatMessageReactionAuthorViewCell.swift
```diff
@@ -0,0 +1,146 @@
...
+        reactionLeadingConstraint = reactionBubbleView.leftAnchor.pin(equalTo: authorAvatarView.centerXAnchor)
+
+        reactionTrailingConstraint?.isActive = false
+        reactionLeadingConstraint?.isActive = false
+    }
+
+    override open func updateContent() {
+        super.updateContent()
+
+        guard let content = self.content else { return }
```
evsaev, line 116: I don't think it's a good pattern, tbh. If `nil` is assigned I would expect the view to display nothing but not showing the previous content. It's especially useful when the cell is being prepared for reuse

Sources/StreamChatUI/ChatMessageList/Reactions/ChatMessageReactionAuthorsVC/ChatMessageReactionAuthorsVC.swift
```diff
@@ -0,0 +1,185 @@
...
+        CGSize(width: 64, height: 110)
+    }
+
+    /// By default it is 10. It will prefetch the reactions before showing the last 10th reaction.
+    open var prefetchThreshold: Int {
+        10
+    }
+
+    /// A Boolean indicating whether the reactions are currently loading.
+    public var isLoadingReactions: Bool = false
```
evsaev, line 45: nit: 
```suggestion
    public private(set) var isLoadingReactions: Bool = false
```

Sources/StreamChatUI/ChatMessageList/Reactions/ChatMessageReactionAuthorsVC/ChatMessageReactionAuthorsVC.swift
```diff
@@ -0,0 +1,185 @@
...
+        _ collectionView: UICollectionView,
+        layout collectionViewLayout: UICollectionViewLayout,
+        sizeForItemAt indexPath: IndexPath
+    ) -> CGSize {
+        reactionAuthorCellSize
+    }
+
+    // MARK: - ChatMessageControllerDelegate
+
+    public func messageController(
```
evsaev, line 161:
```suggestion
    open func messageController(
```

Sources/StreamChatUI/ChatMessageList/Reactions/ChatReactionPickerBubbleView.swift
```diff
@@ -7,6 +7,7 @@ import UIKit
 
 public typealias ChatReactionPickerBubbleView = ChatMessageReactionsBubbleView
 
+/// This component will be renamed to `ChatReactionPickerBubbleView` in future major versions.
 open class ChatMessageReactionsBubbleView: _View, ThemeProvider {
```
evsaev, line 11: the same trick with typealias applies 👍

Sources/StreamChatUI/Components.swift
```diff
@@ -272,7 +279,7 @@ public struct Components {
 
     // MARK: - Thread components
 
-    /// The View Controller used to display the detail of a message thread.
+    /// The view vontroller used to display the detail of a message thread.
```
evsaev, line 282:
```suggestion
    /// The view controller used to display the detail of a message thread.
```

evsaev (changes requested):
Look very good 👍 Left some comments, mostly minor ones

Sources/StreamChat/Controllers/MessageController/MessageController.swift
nuno-vieira: Fixed ✅ Thank you!

Sources/StreamChatUI/ChatMessageList/Reactions/ChatMessageDefaultReactionsBubbleView.swift
nuno-vieira: We can't actually do this, the deprecated message will surface to the typelalias as well, which is stupid 😓

Sources/StreamChatUI/ChatMessageList/Reactions/ChatMessageReactionAuthorsVC/ChatMessageReactionAuthorsVC.swift
nuno-vieira: I'm actually going to put this private, because this might be a temporary API until we do this in the Core SDK

evsaev (approved):
LGTM 👍

evsaev (approved):
🚢

PR #1590, opened 2021-11-01
Always show flag action

b-onc:
### 🎯 Goal

`flag` action was not implemented by backend and it's due to be removed soon. We should not show it to customers since it causes an error, and we should always show "Flag Message" action until it's implemented in Permissions v2. Currently, Flag Message action is always enabled.

For more info, see slack: https://getstream.slack.com/archives/CE5N802GP/p1635503609044000

### 🛠 Implementation

- Snapshot tests are updated accordingly
- `flagsEnabled` flag is removed
- flag command is discarded when parsing

### ☑️ Checklist


adamrushy (dismissed):
Looks good 👍🏼

Sources/StreamChat/APIClient/Endpoints/Payloads/ChannelListPayload.swift
```diff
@@ -265,9 +265,6 @@ public class ChannelConfig: Codable {
     public let createdAt: Date
     /// A channel updated date.
     public let updatedAt: Date
-    
-    /// Determines if users are able to flag messages. Enabled by default.
-    public var flagsEnabled: Bool { commands.map(\.name).contains("flag") }
```
evsaev, line 270: Can we put it back but with a deprecation warning saying it always returns `false` and will be removed soon? Just to avoid a breaking change

CHANGELOG.md
```diff
@@ -3,6 +3,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 # Upcoming
 
+### 🐞 Fixed
+- `flag` command is no longer visible on Composer [#1589](https://github.com/GetStream/stream-chat-swift/issues/1589)
```
evsaev, line 7:
```suggestion
- `flag` command is no longer visible on Composer [#1590](https://github.com/GetStream/stream-chat-swift/issues/1590)
```

evsaev (dismissed):
Let's fix the issue # and we're good to 🚢

PR #1591, opened 2021-11-01
[CIS-1264] Handle shadow bans

b-onc:
### 🔗 Issue Link

CIS-1264

### 🎯 Goal

Shadow ban means the banner user's messages will have `shadowed: true` flag in them and it's the Client's responsibility to not render those messages.
This flag will only be visible to users other than the shadow banned user so they have no way to realize they're shadow banned - ie, a shadow banned user won't see `shadowed: true` for their own messages.

### 🛠 Implementation

iOS SDK did not parse the `shadowed` flag on message, let alone use it. This PR handles it the same as `deletedMessagesVisibility`, by giving the user a choice on the matter. The user can either exclude these messages from the controller's fetch request (therefore not showing them) or include the messages (and render them differently, by providing a `ChatMessageContentView` subclass)

### 🧪 Testing

1. Shadow ban a user
2. Send a message from that user
3. Observe the same channel from the DemoApp

### 🎨 Changes

Screen Shot 2021-11-01 at 17 48 02

### ☑️ Checklist


DemoApp/CustomMessageContentView.swift
```diff
@@ -10,6 +10,11 @@ class CustomMessageContentView: ChatMessageContentView {
     override open func updateContent() {
         super.updateContent()
         
+        if content?.isShadowed == true {
+            textView?.textColor = appearance.colorPalette.textLowEmphasis
+            textView?.text = "This message is from a shadow banned user"
```
evsaev, line 15: nit: would be nice to get this localized 👍

evsaev (dismissed):
LGTM 👍 Just one nit

DemoApp/CustomMessageContentView.swift
b-onc: It's DemoApp only

nuno-vieira (dismissed):
LGTM ✅

evsaev (approved):
LGTM

PR #1592, opened 2021-11-02
CIS-1067 CIS-1315 Optimistic reactions

tbarbugli:
Adds optimistic UI for message reactions.

The goals here is to update the reaction UI immediately and fire the API call behind the scenes. This PR also adds support for retrying API calls based on some policy (backoff and max retries).

**Optimistic Reactions**

Reactions are stored on Core Data immediately, WS events are responsible for marking a reaction as done. There is no use case for manually retrying to send a reaction, the SDK makes 3 attempts to send a reaction. If the 3rd attempt fails, the UI is updated and the reaction goes away.

Deleting reactions is also done with an optimistic approach.

**Important Note 1:**

- API calls that work on the same reaction (ie. add/remove) are not performed in same order, this could be solve by introducing a queue but the complexity / added value trade-off is not clear here. When this happens is possible for API calls to be executed in a different order but the UI and Core Data will be consistent to that.

**API retries:**

```swift
apiClient.request(endpoint: endpoint, retryOptions: retryOptions) { result in }
```
This new method will retry failed calls to the endpoint:

- When the API request fails in a permanent way (ie. 401 HTTP error) there are no retries
- Requests wait in between retries (this is configurable via `retryOptions`)
- Up to `retryOptions.MaxRetries` are performed
- Connectivity errors are not counted towards the maxRetries

Changes: 

- Reactions are added/removed first to the database and then sent to the API
- MessageReactionDTO. loadOrCreate populates all required data for a reaction 
- MessageReactionDTO `createdAt` and `updatedAt` fields are not optional Swift types (matches with CoreData schema)
- Adds a `localState` field to the MessageReactionDTO on CoreData
- Fix tests
- Implement a retry mechanism for reactions


Sources/StreamChat/Workers/MessageUpdater.swift
```diff
@@ -277,8 +319,36 @@ class MessageUpdater: Worker {
...
+            self.apiClient.request(endpoint: .deleteReaction(type, messageId: messageId)) { result in
+                guard let error = result.error else {
+                    return
+                }
+
+                /// update the localState depending on the kind of API error
+                self.database.write { _ in
+                    reaction.localState = ClientError.isEphemeral(error: error) ? .pendingDelete : .deletingFailed
+                }
+            }
```
nuno-vieira, line 351: The `database.write` has a completion block to notify when the database writes finishes:
```swift
database.write { session
    try session.removeReaction(from: messageId, type: type)
} completion: { error in
   if let error = error {
       ....
       return
   }
   ...
}
```

Either way, I don't think there's any issue with the current approach 👍

Sources/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -512,6 +512,77 @@ extension NSManagedObjectContext: MessageDatabaseSession {
...
+        score: Int,
+        extraData: [String: RawJSON]
+    ) throws -> MessageReactionDTO? {
+        guard let currentUserDTO = currentUser else {
+            throw ClientError.CurrentUserDoesNotExist()
+        }
+
+        guard let message = MessageDTO.load(id: messageId, context: self) else {
+            return nil
+        }
```
nuno-vieira, line 535: `throws -> MessageReactionDTO?` is a bit redundant, it should be `throws -> MessageReactionDTO` and in this case, it should return an Error as well, something like `ClientError.MessageDoesNotExit()`

Sources/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -512,6 +512,77 @@ extension NSManagedObjectContext: MessageDatabaseSession {
...
+
+        guard let reaction = MessageReactionDTO
+            .load(userId: currentUserDTO.user.id, messageId: messageId, type: type, context: self) else {
+            return nil
+        }
+
+        guard let message = MessageDTO.load(id: messageId, context: self) else {
+            return nil
+        }
```
nuno-vieira, line 573: Same redundancy as above, we should throw errors here

Sources/StreamChat/Workers/MessageUpdater.swift
```diff
@@ -255,15 +255,57 @@ class MessageUpdater: Worker {
...
+                            return
+                        }
+                        reaction.localState = .sendingFailed
+                    }
+                    return
+                }
+
+                self.database.write { _ in
+                    reaction.localState = .pendingSend
+                }
```
nuno-vieira, line 307: Will we have a retry button to retry the reaction? Does that UI exist? 🤔 I thought that, if the reaction failed, for whatever reason, it would simply remove it. For example, Twitter and Instagram etc do this, if there is no network, or the request fails, the reaction is simply removed.

Sources/StreamChat/APIClient/RequestDecoder.swift
```diff
@@ -45,7 +45,7 @@ struct DefaultRequestDecoder: RequestDecoder {
         
         log.debug("URL request response: \(httpResponse), data:\n\(data.debugPrettyPrintedJSON))", subsystems: .httpRequests)
         
-        guard httpResponse.statusCode < 400 else {
+        guard httpResponse.statusCode < 300 else {
```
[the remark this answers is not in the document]
tbarbugli: because only 2xx response codes are OK (previous code handled HTTP redirects as OK). FTR 3xx responses are not used intentionally by the API servers

Sources/StreamChat/APIClient/APIClient.swift
```diff
@@ -28,6 +35,9 @@ class APIClient {
     /// Used for synchronizing access to requestsQueue
     private let requestsAccessQueue = DispatchQueue(label: "io.getstream.requests")
     
+    /// Used for request retires
```
nuno-vieira, line 38:
```suggestion
    /// Used for request retries
```

Sources/StreamChat/APIClient/APIClient.swift
```diff
@@ -28,6 +35,9 @@ class APIClient {
...
+    private let requestsRetriesQueue = DispatchQueue(label: "io.getstream.requests-retries")
```
nuno-vieira, line 39:
```suggestion
    private let requestsRetriesQueue = DispatchQueue(label: "io.getstream.request-retries")
```

Sources/StreamChat/APIClient/APIClient.swift
```diff
@@ -63,6 +73,64 @@ class APIClient {
...
+    ///   - timeout: The timeout in seconds for the API request.
+    ///   - retryOptions: The `RetryOptions` used to retry the request.
+    ///   - completion: Called when the networking request is finished.
+    func request<Response: Decodable>(
+        endpoint: Endpoint<Response>,
+        timeout: TimeInterval = 60,
+        retryOptions: RetryOptions,
+        completion: @escaping (Result<Response, Error>) -> Void
+    ) {
+        print("debugging: request \(endpoint.method) \(endpoint.path)")
```
nuno-vieira, line 90: Make sure to not leave this here. And use the logger 👍

Sources/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -512,6 +512,82 @@ extension NSManagedObjectContext: MessageDatabaseSession {
...
+
+        let result = try MessageReactionDTO.loadOrCreate(messageId: messageId, type: type, user: currentUserDTO.user, context: self)
+
+        // make sure we update the reactionScores for the message in a way that works for new or updated reactions
+        let scoreDiff = Int64(score) - result.dto.score
+        message
+            .reactionScores[type.rawValue] = max(0, message.reactionScores[type.rawValue] ?? Int(result.dto.score) + Int(scoreDiff))
```
nuno-vieira, line 542: We can make this a little bit more readable: 
```suggestion
        let scoreDiff = Int64(score) - result.dto.score
        let newScore = max(0, message.reactionScores[type.rawValue] ?? Int(result.dto.score) + Int(scoreDiff))
        message.reactionScores[type.rawValue] = newScore
```

Sources/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -512,6 +512,82 @@ extension NSManagedObjectContext: MessageDatabaseSession {
...
+
+        result.dto.score = Int64(score)
+        result.dto.extraData = try JSONEncoder.default.encode(extraData)
+
+        if message.reactions.filter({ MessageReactionDTO.createId(dto: $0) == MessageReactionDTO.createId(dto: result.dto) })
+            .isEmpty {
+            message.reactions.insert(result.dto)
+        }
```
nuno-vieira, line 550: What is this doing? 🤔 Are we checking if the reaction already exists? `message.reactions` is already a `Set`, if you insert a duplicate, there's no problem.

Sources/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -512,6 +512,82 @@ extension NSManagedObjectContext: MessageDatabaseSession {
...
+        guard version == nil || version == reaction.version else {
+            return nil
+        }
+
+        guard let message = MessageDTO.load(id: messageId, context: self) else {
+            return nil
+        }
+
+        message.reactions = message.reactions
+            .filter { MessageReactionDTO.createId(dto: reaction) != MessageReactionDTO.createId(dto: $0) }
```
nuno-vieira, line 581: Same here? We don't need to check if it exists?

Sources/StreamChat/APIClient/APIClient.swift
```diff
@@ -4,6 +4,13 @@
 
 import Foundation
 
+public struct RetryOptions {
```
nuno-vieira, line 7: We don't need this to be public?
```suggestion
struct RetryOptions {
```

Sources/StreamChat/Workers/MessageUpdater.swift
```diff
@@ -255,15 +255,48 @@ class MessageUpdater: Worker {
...
+            } catch {
+                log.warning("Failed to add the reaction to the database: \(error)")
+                completion?(error)
+            }
+
+            if let reaction = reaction {
+                reaction.localState = .sending
+                reaction.version = version
+            }
+        } completion: { _ in
```
nuno-vieira, line 287:
```suggestion
        } completion: { error in
```

We need to check the error in the completion block. This is what will make sure it was successfully written to the DB or not.

Sources/StreamChat/Workers/MessageUpdater.swift
```diff
@@ -255,15 +255,48 @@ class MessageUpdater: Worker {
...
+                reaction = try session.addReaction(
+                    to: messageId,
+                    type: type,
+                    score: score,
+                    extraData: extraData
+                )
+                completion?(nil)
+            } catch {
+                log.warning("Failed to add the reaction to the database: \(error)")
+                completion?(error)
```
nuno-vieira, line 280: On the other hand, this is not really true. Right now we are return `nil` on the `addReaction()` function but we should not. We should throw always an error. If this fails, the database write will also fail. So it is better to handle the DB error on the completion block

Sources/StreamChat/Database/DTOs/MessageReactionDTO.swift
```diff
@@ -8,23 +8,34 @@ import Foundation
...
+    @NSManaged var createdAt: Date?
+    @NSManaged var updatedAt: Date?
     @NSManaged var extraData: Data
     
     @NSManaged var message: MessageDTO
     @NSManaged var user: UserDTO
-    
+
+    // internal field needed to sync data optimistically
+    @NSManaged var version: String?
```
nuno-vieira, line 24: Can you be more descriptive on why we need this version prop? Didn't understand yet why need it 🤔

tbarbugli: this is avoid one easy out-of-sync scenario with the retry error handler, when adding a reaction fails the reaction is only updated unless it was changed in the meantime

Sources/StreamChat/Workers/MessageUpdater.swift
tbarbugli: tbh I am not sure, when this happens (writing the reaction to the DB for optimistic reasons) it still makes sense to make the API call and let that settle the whole situation

Sources/StreamChat/APIClient/APIClient.swift
tbarbugli: pretty sure this will be used in other places as well and that customers will want to be able to provide their own retry policy (ie. how many times we retry to send a message before giving up)

Sources/StreamChat/Workers/MessageUpdater.swift
tbarbugli: IMO its a better user experience if we retry behind the scenes, this way you can react to something while offline

Sources/StreamChat/Database/DTOs/MessageDTO.swift
tbarbugli: debatable point, I think its better to rely on the API call to settle the score here

nuno-vieira: This is not an API Call tho, it is a DB Write, it just makes the code harder to handle. Not only do you need to catch the error, but also handle a `nil`. This is not very "swifty"

Sources/StreamChat/Workers/MessageUpdater.swift
nuno-vieira: Hum, how will that work tho? The only way to make it work offline is to save the request in the DB and use a queue. Currently, we have something +/- like this with the `MessageSender` which is our background workers. It is very unlikely that if you are offline for some reason, that in the time it takes to do the retries, you will be able to make the request. So at most this retry mechanism will be useful for connection stability issues, not offline

Sources/StreamChat/APIClient/APIClient.swift
nuno-vieira: In that case, we need to create the `public init` or that won't be exposed

Sources/StreamChat/APIClient/APIClient.swift
```diff
@@ -4,6 +4,16 @@
...
 import Foundation
 
+/// The retry policy to use by the API retry mechanism
+public struct RetryOptions {
+    /// How many attempts should be performed at most
+    let maxRetries: Int = 3
+    /// How much time there should be in between attempts, this function is usually dependant on the amount of attempts already performed
+    let backoff: (Int) -> Double = {
+        min(max(Double($0) * Double($0), 0.5), 6.0)
+    }
```
nuno-vieira, line 14:
```suggestion
    public let maxRetries: Int = 3
    /// How much time there should be in between attempts, this function is usually dependant on the amount of attempts already performed
    public let backoff: (Int) -> Double = {
        min(max(Double($0) * Double($0), 0.5), 6.0)
    }
```

Sources/StreamChat/APIClient/APIClient.swift
```diff
@@ -4,6 +4,16 @@
 
 import Foundation
 
+/// The retry policy to use by the API retry mechanism
+public struct RetryOptions {
+    /// How many attempts should be performed at most
+    let maxRetries: Int = 3
+    /// How much time there should be in between attempts, this function is usually dependant on the amount of attempts already performed
+    let backoff: (Int) -> Double = {
```
nuno-vieira, line 12: This helps better understand what is the `Int` 
```suggestion
    let backoff: (maxRetries: Int) -> Double = {
```

Sources/StreamChat/APIClient/APIClient.swift
evsaev: I would also not expose it unless we have a strong reason/a customer asking for it. Currently, there's no way atm for a customer to customize that but it does increase the public-facing surface and which makes us less flexible in terms of internal design

Sources/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -550,6 +550,82 @@ extension NSManagedObjectContext: MessageDatabaseSession {
...
+        let scoreDiff = Int64(score) - result.dto.score
+        let newScore = max(0, message.reactionScores[type.rawValue] ?? Int(result.dto.score) + Int(scoreDiff))
+        message.reactionScores[type.rawValue] = newScore
+
+        result.dto.score = Int64(score)
+        result.dto.extraData = try JSONEncoder.default.encode(extraData)
+
+        if message.reactions.filter({ MessageReactionDTO.createId(dto: $0) == MessageReactionDTO.createId(dto: result.dto) })
+            .isEmpty {
+            message.reactions.insert(result.dto)
```
evsaev, line 587: Seems that adding a reaction to `message.reactions` is enough. This `Message ->> Reactions` relation is a set so the attempt to add an already existing item will take no effect and can be done without a prior check.
```suggestion
message.reactions.insert(result.dto)
```

Sources/StreamChat/Database/DTOs/MessageReactionDTO.swift
evsaev: Also didn't get this one, sorry 😄 Could you explain again what is the scenario we're trying to prevent having the `version`?

Sources/StreamChat/Database/DTOs/MessageReactionDTO.swift
```diff
@@ -48,69 +67,77 @@ extension MessageReactionDTO {
...
     }
     
     static func loadOrCreate(
-        userId: String,
         messageId: MessageId,
         type: MessageReactionType,
+        user: UserDTO,
         context: NSManagedObjectContext
-    ) -> MessageReactionDTO {
+    ) throws -> (dto: MessageReactionDTO, created: Bool) {
```
evsaev, line 98: we check `created` value somewhere, right? 🤔

Sources/StreamChat/Workers/MessageUpdater.swift
```diff
@@ -255,15 +257,47 @@ class MessageUpdater: Worker {
...
+            } catch {
+                log.warning("Failed to add the reaction to the database: \(error)")
+            }
+
+            if let reaction = reaction {
+                reaction.localState = .sending
+                reaction.version = version
+            }
+        } completion: { error in
+            self.apiClient.request(endpoint: endpoint, retryOptions: self.retryOptions) { result in
```
evsaev, line 288: If we've got the error from DB should we forward it to the caller and skip further steps? 🤔

Sources/StreamChat/Workers/MessageUpdater.swift
```diff
@@ -277,8 +311,35 @@ class MessageUpdater: Worker {
         messageId: MessageId,
         completion: ((Error?) -> Void)? = nil
     ) {
-        apiClient.request(endpoint: .deleteReaction(type, messageId: messageId)) {
-            completion?($0.error)
+        var reaction: MessageReactionDTO?
+
+        database.write { session in
```
evsaev, line 316: `NSManagedObjects` are not thread-safe. If we need access to the reaction DTO after the API call we need to open another db session calling `database.write` 👍

Sources/StreamChat/Workers/MessageUpdater.swift
```diff
@@ -277,8 +311,35 @@ class MessageUpdater: Worker {
...
+                reaction = try session.removeReaction(from: messageId, type: type, on: nil)
+            } catch {
+                log.warning("Failed to remove the reaction from to the database: \(error)")
+            }
+
+            guard let reaction = reaction else {
+                return
+            }
+            reaction.localState = .pendingDelete
+        } completion: { error in
```
evsaev, line 327: this db error should probably also be respected 🤔 e.g. if it's non-nil we should trigger the completion and stop

Sources/StreamChat/APIClient/APIClient.swift
```diff
@@ -63,6 +76,63 @@ class APIClient {
...
+    func request<Response: Decodable>(
+        endpoint: Endpoint<Response>,
+        timeout: TimeInterval = 60,
+        retryOptions: RetryOptions,
+        completion: @escaping (Result<Response, Error>) -> Void
+    ) {
+        requestsRetriesQueue.async { [weak self] in
+            self?.request(attempt: 0, endpoint: endpoint, timeout: timeout, retryOptions: retryOptions, completion: completion)
+        }
+    }
```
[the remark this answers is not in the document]
tbarbugli: reactions are not as important as messages, manually retry to send a reaction is not really a thing.

FTR what we do for messages is also not ideal, most apps retry the message automatically (and with different policies). AFAIK we do not support this yet

Sources/StreamChat/Database/DTOs/MessageReactionDTO.swift
tbarbugli: this is used to perform optimistic locking style updates (general idea is linked [here](https://en.wikipedia.org/wiki/Optimistic_concurrency_control))

example:

- insert reaction Z with version X
- [ other things happen and perhaps alter reaction Z, in that case they will erase / overwrite the version field ]
- update reaction Z if version is X

tbarbugli (commented):
meh, not really a review. I just wanted to leave replies :)

Sources/StreamChat/Workers/MessageUpdater.swift
tbarbugli: this is assuming that the database is more important than the API call and if db errors for some reason user input should fail. I think we're better off giving up on the whole optimistic UI part in case of db error and fallback to what we usually do (perform an API call and rely on event handling to update the message)

Sources/StreamChat/Workers/MessageUpdater.swift
tbarbugli: this is on purpose (DB fails to do the optimistic bit? ignore and move with API call)

Sources/StreamChat/Workers/MessageUpdater.swift
tbarbugli: retries are performed forever if the request fails due to connectivity problems.

Sources/StreamChat/APIClient/APIClient.swift
tbarbugli: @nuno-vieira 
> In that case, we need to create the public init or that won't be exposed

Are you sure about this?

nuno-vieira: Yes. This is easy to test, just try to create the struct on the demo app. It will say the init is not accessible

nuno-vieira (approved):

Sources/StreamChat/WebSocketClient/EventMiddlewares/MessageReactionsMiddleware.swift
```diff
@@ -1,50 +0,0 @@
-//
-// Copyright © 2021 Stream.io Inc. All rights reserved.
-//
-
-import CoreData
-import Foundation
-
-/// The middleware listens for `EventWithReactionPayload` events and updates `MessageReactionDTO` accordingly.
-struct MessageReactionsMiddleware: EventMiddleware {
-    func handle(event: Event, session: DatabaseSession) -> Event? {
```
nuno-vieira, line 10: Are we sure we don't need this anymore? At least with my QA it is working fine, but why did we had this in the first place then? 🤔

nuno-vieira (commented):
Overall looks good, just left some small comments trying to understand the recent changes. 
I also noticed that the Reactions List View actually doesn't filter out the available reactions, so the UI is glicthed, like you can see here: 


Something weird is that the LoL image is off as well, probably a bug on the asset in this case. Let me know if you want to fix this as part of this PR, or I can create a ticket to fix this, since this is actually something that was missed from my PR.

Sources/StreamChat/Controllers/MessageController/MessageController_Tests.swift
```diff
@@ -1183,18 +1182,20 @@ final class MessageController_Tests: XCTestCase {
...
+        )
 
         let expectedLatestReactions = mockedReactions
-            .sorted(by: { $0.updatedAt > $1.updatedAt })[0..<10]
+            .sorted(by: { $0.updatedAt > $1.updatedAt })
 
         controller.startObserversIfNeeded()
         
-        XCTAssertEqual(controller.reactions.count, 10)
+        XCTAssertEqual(controller.reactions.count, 20)
```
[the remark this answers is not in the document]
nuno-vieira: I see... you are providing all `mockedReactions` as the latest reactions. For this test to make sense again, we need to create a message, save multiple reactions, ideally 10+, then set some latestReactions, and verify that `controller.reactions.count == latestReactions.count`

Sources/StreamChat/WebSocketClient/EventMiddlewares/MessageReactionsMiddleware.swift
tbarbugli: this is not needed anymore, reactions are written to the database by `EventDataProcessorMiddleware` when a message is part of the event

Sources/StreamChat/Controllers/MessageController/MessageController_Tests.swift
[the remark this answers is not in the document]
tbarbugli: if we really want to have the 10 limit, I suggest we do this in the view layer. I wonder if this actually causes any problem tbh, if not I rather leave it as it is

tbarbugli:
> Overall looks good, just left some small comments trying to understand the recent changes. I also noticed that the Reactions List View actually doesn't filter out the available reactions, so the UI is glitched: 
> 
> And somehow the blue LoL icon is buggy as well. Let me know if you want to fix this as part of this PR, or I can create a ticket to fix this since this is actually something that was missed from my PR.

Better if we patch this in a separate PR, can you take care of that?

nuno-vieira (approved):
LGTM! ✅

PR #1596, opened 2021-11-03
[CIS-1207, CIS-1232] Fix crash and swapping the long-pressed message on `ChatMessagePopupVC`

evsaev:
### 🔗 Issue Link

CIS-1207, CIS-1232

### 🎯 Goal

- Fix crash when `ChatMessagePopupVC` is dismissed when the message is outside the visible area of the message list
- Fix behavior when a long-pressed message is swapped with a newly received message when they have the same style

### 🛠 Implementation

Stop passing `ChatMessageContentView` across the message list and pop-up view hierarchies and pass the copy instead. 

Therefore, the following subviews are removed from `MessageActionsTransitionController`:
- messageContentViewFrame
- messageContentViewActivateConstraints
- messageContentViewDeactivateConstraints
- messageContentView
- messageContentViewSuperview
- mainContainerTopAnchor

### 🧪 Testing

Manual testing:
- Launch `DemoApp` on 2 devices
- Sign in as `Luke` and `Leia`
- Open the channel where both users are members
- Send text message as `Luke`, observe the message received by `Leia`
- Long press the message by `Leia`
- Send another text message as `Luke`

**Expected result:** 
The message long-pressed by `Leia` is not swapped with the new message

### 🎨 Changes

The message inside the pop-up is not swapped:

https://user-images.githubusercontent.com/12818985/140047965-18b7957a-0803-43f9-86be-f1bf540531ec.mov

When the pop-up is shown giffy restarts but plays smoothly because the actual view is used while the transition and not the snapshot:

https://user-images.githubusercontent.com/12818985/140048297-5a3d7c1f-18f1-40b6-8263-7874992be1c2.mp4

### ☑️ Checklist

### DIFF
--- a/Sources/StreamChatUI/MessageActionsPopup/MessageActionsTransitionController.swift
+++ b/Sources/StreamChatUI/MessageActionsPopup/MessageActionsTransitionController.swift
@@ -6,25 +6,50 @@ import Foundation
 import StreamChat
 import UIKit
 
-/// Transitions controller for `ChatMessagePopupVC`.
-open class MessageActionsTransitionController: NSObject, UIViewControllerTransitioningDelegate,
+/// The controller that handles `ChatMessageListVC <-> ChatMessagePopUp` transition.
+open class ChatMessageActionsTransitionController: NSObject, UIViewControllerTransitioningDelegate,
     UIViewControllerAnimatedTransitioning {
     /// Indicates if the transition is for presenting or dismissing.
     open var isPresenting: Bool = false
-    /// `messageContentView`'s initial frame.
-    open var messageContentViewFrame: CGRect = .zero
-    /// `messageContentView`'s constraints to be activated after dismissal.
-    open var messageContentViewActivateConstraints: [NSLayoutConstraint] = []
-    /// Constraints to be deactivated after dismissal.
-    open var messageContentViewDeactivateConstraints: [NSLayoutConstraint] = []
-    /// `messageContentView` instance that is animated.
-    open weak var messageContentView: ChatMessageContentView?
-    /// `messageContentView`'s initial superview.
-    open weak var messageContentViewSuperview: UIView?
-    /// Top anchor for main container.
-    open var mainContainerTopAnchor: NSLayoutConstraint?
+    
     /// Feedback generator.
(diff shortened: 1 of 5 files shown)


b-onc (approved):

Sources/StreamChatUI/MessageActionsPopup/MessageActionsTransitionController.swift
```diff
@@ -12,20 +12,51 @@ open class MessageActionsTransitionController: NSObject, UIViewControllerTransit
...
+        deprecated,
+        message: "This field is no longer used because the message content view is not moved across view heirarchies anymore but get copied."
+    )
     open weak var messageContentViewSuperview: UIView?
     /// Top anchor for main container.
+    @available(
+        *,
+        deprecated,
+        message: "This field is no longer used because the message content view is not moved across view heirarchies anymore but get copied."
+    )
```
nuno-vieira, line 49: I don't think it is valid that we these properties as deprecated, because they are actually not working anymore. When something is deprecated, it means it still works, but it won't work in the future

nuno-vieira (changes requested):

Sources/StreamChatUI/MessageActionsPopup/MessageActionsTransitionController.swift
b-onc: that's actually a very good point. @evsaev what do you say:
1. we implement a new `MessageActionsTransitionController`
2. if the user assigned their subclass to `Components`, we use it, otherwise we use the new, improved version

nuno-vieira (dismissed):
LGTM! 🎉 Nice job 👍

Sources/StreamChatUI/MessageActionsPopup/MessageActionsTransitionController.swift
evsaev: @b-onc the `MessageActionsTransitionController` is not part of `Components` yet and the reason for this is the transition being quite complex. I assume none of the customers has this part customized so removing those fields is actually OK. Not that high price for the 2 issues we've addressed.

evsaev: ✅ Added transition controller to `Components`

b-onc (approved):

PR #1597, opened 2021-11-04
[CIS-1247] Adding new factory method to create ChatChannelListVC

adamrushy:
### 🔗 Issue Link
https://stream-io.atlassian.net/browse/CIS-1247

### 🎯 Goal

Introduce a new factory method to instantiate a ChatChannelListVC so that we're not relying on ViewController lifecycle. This will prevent a crash if `super` is called before instantiating. 

### 🛠 Implementation

New factory method that will take a UIStoryboard and/or UIStoryboardId if these aren't resolved then we return a newly created ChatChannelListVC with the controller that is passed in.

### 🧪 Testing

The implementation will look something like this:

`let controller = ChatClient.shared.channelListController(query: .init(filter: .containMembers(userIds: [userCredentials.id])))`
`let chatList = ChatChannelListVC.make(with: controller, storyboard: nil, storyboardId: nil)`

### ☑️ Checklist

### DIFF
--- a/Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
+++ b/Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
@@ -61,7 +61,7 @@ open class ChatChannelListVC: _ViewController,
     /// Create a new ChatChannelListViewController
     /// - Parameters:
     ///   - controller: Your created ChatChannelListController with required query
-    ///   - storyboard: The storyboard to instanciate your ViewController from
+    ///   - storyboard: The storyboard to instantiate your ViewController from
     ///   - storyboardId: The storyboardId that is set in your Storyboard reference
     /// - Returns: A newly created ChatChannelListViewController
     public static func make(
--- a/Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
+++ b/Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
@@ -58,12 +58,12 @@ open class ChatChannelListVC: _ViewController,
     /// Used for mapping `ListChanges` to sets of `IndexPath` and verifying possible conflicts
     private let collectionUpdatesMapper = CollectionUpdatesMapper()
     
-    /// Create a new ChatChannelListViewController
+    /// Create a new `ChatChannelListViewController`
     /// - Parameters:
-    ///   - controller: Your created ChatChannelListController with required query
-    ///   - storyboard: The storyboard to instantiate your ViewController from
-    ///   - storyboardId: The storyboardId that is set in your Storyboard reference
-    /// - Returns: A newly created ChatChannelListViewController
+    ///   - controller: Your created `ChatChannelListController` with required query
+    ///   - storyboard: The storyboard to instantiate your `ViewController` from
+    ///   - storyboardId: The `storyboardId` that is set in your `UIStoryboard` reference
+    /// - Returns: A newly created `ChatChannelListViewController`
     public static func make(
         with controller: ChatChannelListController,
         storyboard: UIStoryboard? = nil,
--- a/Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
+++ b/Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
@@ -64,8 +64,8 @@ open class ChatChannelListVC: _ViewController,
     ///   - storyboard: The storyboard to instanciate your ViewController from
     ///   - storyboardId: The storyboardId that is set in your Storyboard reference
     /// - Returns: A newly created ChatChannelListViewController
-    public static func createChannelVC(
-        _ controller: ChatChannelListController,
+    public static func make(
+        with controller: ChatChannelListController,
         storyboard: UIStoryboard? = nil,
         storyboardId: String? = nil
     ) -> ChatChannelListVC {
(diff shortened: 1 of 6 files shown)


adamrushy:
Note: I need to add tests

Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
```diff
@@ -57,6 +57,30 @@ open class ChatChannelListVC: _ViewController,
...
+    
+    /// Create a new ChatChannelListViewController
+    /// - Parameters:
+    ///   - controller: Your created ChatChannelListController with required query
+    ///   - storyboard: The storyboard to instanciate your ViewController from
+    ///   - storyboardId: The storyboardId that is set in your Storyboard reference
+    /// - Returns: A newly created ChatChannelListViewController
+    public static func createChannelVC(
+        _ controller: ChatChannelListController,
+        storyboard: UIStoryboard?,
```
martinmitrevski, line 69: Maybe add a default nil value here as well?

adamrushy: Yep good spot 👍🏼

Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
```diff
@@ -57,6 +57,30 @@ open class ChatChannelListVC: _ViewController,
...
+        storyboardId: String? = nil
+    ) -> ChatChannelListVC {
+        var chatChannelListVC: ChatChannelListVC!
+        
+        if let storyboardId = storyboardId {
+            let storyboard = storyboard
+            chatChannelListVC = storyboard?.instantiateViewController(withIdentifier: storyboardId) as? ChatChannelListVC
```
martinmitrevski, line 76: If the loading fails, this will crash - we should have a safer mechanism here.

martinmitrevski (commented):
Looks good 👍 Just some small remarks about the storyboard handling.

Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
```diff
@@ -66,14 +66,13 @@ open class ChatChannelListVC: _ViewController,
...
+        storyboard: UIStoryboard? = nil,
         storyboardId: String? = nil
     ) -> ChatChannelListVC {
         var chatChannelListVC: ChatChannelListVC!
         
-        if let storyboardId = storyboardId {
-            let storyboard = storyboard
-            chatChannelListVC = storyboard?.instantiateViewController(withIdentifier: storyboardId) as? ChatChannelListVC
+        if let storyboardId = storyboardId, let storyboard = storyboard {
+            chatChannelListVC = storyboard.instantiateViewController(withIdentifier: storyboardId) as? ChatChannelListVC
```
martinmitrevski, line 75: We still have an unhandled case here. If the chatChannelListVC is nil, it will crash in the setup lines below, since it's implicitly unwrapped above. Question is should we allow it to crash like that, or throw a fatal error with explanation, since most likely that would be a developer's error?

nuno-vieira (changes requested):
Overall LGTM 👍 Will request changes especially because the current `createChannelVC` is not accurate, since we have already a `channelVC` which is not `channelListVC` 👍

Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
```diff
@@ -57,6 +57,39 @@ open class ChatChannelListVC: _ViewController,
...
     /// Used for mapping `ListChanges` to sets of `IndexPath` and verifying possible conflicts
     private let collectionUpdatesMapper = CollectionUpdatesMapper()
+    
+    /// Create a new ChatChannelListViewController
+    /// - Parameters:
+    ///   - controller: Your created ChatChannelListController with required query
+    ///   - storyboard: The storyboard to instanciate your ViewController from
+    ///   - storyboardId: The storyboardId that is set in your Storyboard reference
+    /// - Returns: A newly created ChatChannelListViewController
+    public static func createChannelVC(
```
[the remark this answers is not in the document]
adamrushy: I think this is absolutely right, I will `make` these changes ;]

Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
```diff
@@ -57,6 +57,39 @@ open class ChatChannelListVC: _ViewController,
 
     /// Used for mapping `ListChanges` to sets of `IndexPath` and verifying possible conflicts
     private let collectionUpdatesMapper = CollectionUpdatesMapper()
+    
+    /// Create a new ChatChannelListViewController
```
martinmitrevski, line 61: I think the types in the method docs should be between ``, e.q. `ChatChannelListViewController`.

martinmitrevski (commented):
Looks good 👍 Small remark about the docs, and we still need some tests.

nuno-vieira (dismissed):
LGTM! ✅

Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
adamrushy: Yep great spot @martinmitrevski

martinmitrevski (approved):
LGTM 👍

PR #1605, opened 2021-11-10
[CIS-1262, CIS-1293] Fix message list jumps and cell not resizing on edit

nuno-vieira:
### 🔗 Issue Link
CIS-1262
CIS-1293

### 🎯 Goal
- Fix message list scrolling jumps when a new message is received.
- Fix message cell size not updated when editing a message with bigger/smaller content.
- **Bonus:**
     - Fix not executing the `completion` of `updateMessages()` which was not being called ever.
     - Removes the list changes conflicts log.

### 🛠 Implementation
- Scrolling jumps fix: Whenever the message list is reloaded, it re-calculates the message cell's height, causing a scrolling jump. When using `performBatchUpdates` this is not an issue. Use `performBatchUpdates` only for inserts. We also need to reload the previous cell to make sure the timestamp is removed if needed.
- Message cell resize fix: When a message is edited, previously we were just updating the message cell content and not reloading the cell. Now, if the text of the message is different, we also reload the cell, to re-calculate the cell size.

### 🧪 Testing
- Scrolling jumps fix: 
    1. Scroll up in the message list
    2. Send a message from another device
    3. Verify that the message list doesn't move
- Message cell resize fix:
   1. Edit a message with more lines of text than it had before
   2. Verify the cell changes its size.

### 🎨 Changes
| Message cell resize fix |  Scrolling jumps fix |
| --- | --- |
| <video src="https://user-images.githubusercontent.com/12814114/141162863-ef1803a2-8da8-4645-9fa6-2eecb120cf61.mp4"> | <video src="https://user-images.githubusercontent.com/12814114/141162941-9f54c514-537e-418f-aa48-e21787fdef8d.mp4"> |

### ☑️ Checklist

### DIFF
--- a/Sources/StreamChatUI/ChatMessageList/ChatMessageListView.swift
+++ b/Sources/StreamChatUI/ChatMessageList/ChatMessageListView.swift
@@ -172,19 +172,21 @@ open class ChatMessageListView: UITableView, Customizable, ComponentsProvider {
         with changes: [ListChange<ChatMessage>],
         completion: (() -> Void)? = nil
     ) {
+        defer {
+            completion?()
+        }
+
         guard let _ = collectionUpdatesMapper.mapToSetsOfIndexPaths(
             changes: changes,
             onConflict: {
                 reloadData()
             }
         ) else {
-            completion?()
             return
         }
 
         if changes.count > 1 {
             reloadData()
-            completion?()
             return
         }
 
@@ -199,7 +201,6 @@ open class ChatMessageListView: UITableView, Customizable, ComponentsProvider {
                         // +1 instead of -1 because the message list is inverted
                         let previousIndex = IndexPath(row: index.row + 1, section: index.section)
                         self.reloadRows(at: [previousIndex], with: .none)
-                        completion?()
                     }
                 }
 
@@ -211,15 +212,12 @@ open class ChatMessageListView: UITableView, Customizable, ComponentsProvider {
 
             case let .move(_, fromIndex: fromIndex, toIndex: toIndex):
                 self.moveRow(at: fromIndex, to: toIndex)
-                completion?()
 
             case let .update(_, index: index):
                 self.reloadRows(at: [index], with: .automatic)
-                completion?()
 
             case .remove:
                 self.reloadData()
-                completion?()
             }
         }
     }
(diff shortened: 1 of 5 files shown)


Sources/StreamChatUI/ChatMessageList/ChatMessageListView.swift
```diff
@@ -188,8 +191,16 @@ open class ChatMessageListView: UITableView, Customizable, ComponentsProvider {
...
             case let .insert(message, index: index):
                 UIView.performWithoutAnimation {
-                    self.reloadData()
+                    self.performBatchUpdates {
+                        self.insertRows(at: [index], with: .none)
+                    } completion: { _ in
+                        // Update previous row to remove timestamp if needed
+                        // +1 instead of -1 because the message list is inverted
+                        let previousIndex = IndexPath(row: index.row + 1, section: index.section)
+                        self.reloadRows(at: [previousIndex], with: .none)
```
evsaev, line 200: Is OK for the case when the inserted message is the only message in the channel ie. when the previous cell we're reloading does not exist?

Sources/StreamChatUI/ChatMessageList/ChatMessageListView.swift
```diff
@@ -228,11 +239,13 @@ open class ChatMessageListView: UITableView, Customizable, ComponentsProvider {
                 cellBeforeUpdateReuseIdentifier == cellAfterUpdateReuseIdentifier,
                 cellBeforeUpdateMessage?.id == cellAfterUpdateMessage?.id,
                 cellBeforeUpdateMessage?.type == cellAfterUpdateMessage?.type,
-                cellBeforeUpdateMessage?.deletedAt == cellAfterUpdateMessage?.deletedAt {
+                cellBeforeUpdateMessage?.deletedAt == cellAfterUpdateMessage?.deletedAt,
+                cellBeforeUpdateMessage?.text == cellAfterUpdateMessage?.text {
```
evsaev, line 243: Would be nice to add these scenarios to the pre-release QA doc to make sure we don't have regressions related to cell reloading

evsaev (changes requested):
Let's check the case with a single message and if it's OK I'll approve 👍

Sources/StreamChatUI/ChatMessageList/ChatMessageListView.swift
nuno-vieira: Nice catch, again 👏

evsaev (approved):
🚢

PR #1621, opened 2021-11-12
CIS-1313 Remove TeamDTO entity and replace with String and [String]

tbarbugli:
This PR replaces TeamDTO entity with String and [String]. There is no need to store teams in a separate entity:

- There is no additional data attached to a team besides its name
- Enumerating teams is not needed
- Storing data in two tables/entities is more expensive and complex than storing in one column/attribute


Sources/StreamChat/Database/DatabaseContainer_Tests.swift
```diff
@@ -23,34 +23,6 @@ class DatabaseContainer_Tests: XCTestCase {
...
-        }, completion: $0) }
-        
-        // Assert the completion was called with an error
-        //
-        // XCTAssertNotNil should be used but it seems to touch various properties of `errorCompletion`
-        // which results in touching `TeamDTO` stored in error's `userInfo`
-        // which is managed in local context inside write function and that makes CoreData concurrency unhappy
-        XCTAssert(errorCompletion != nil)
-    }
-
```
evsaev, line 53: Hmm, this one needs to be updated with another entity and stay, doesn't it? It checks the `write` triggers completion and propagates the error if there's one.

evsaev (changes requested):
Let's put the test back and we're good to 🚢

Sources/StreamChat/Database/StreamChatModel.xcdatamodeld/StreamChatModel.xcdatamodel/contents
```diff
@@ -253,6 +240,7 @@
         <attribute name="isOnline" attributeType="Boolean" usesScalarValueType="YES"/>
         <attribute name="lastActivityAt" optional="YES" attributeType="Date" usesScalarValueType="NO"/>
         <attribute name="name" optional="YES" attributeType="String"/>
+        <attribute name="teams" optional="YES" attributeType="Transformable" valueTransformerName="NSSecureUnarchiveFromData"/>
```
evsaev, line 243: Let's also specify the custom type just to be explicit about the data we store

Screenshot 2021-11-12 at 13 17 04

tbarbugli: good catch

evsaev (approved):
LGTM

PR #1622, opened 2021-11-12
[CIS-1316] Take sorting into account in `ChatChannelListQuery` hash

evsaev:
### 🔗 Issue Link

CIS-1316

### 🎯 Goal

Filter and sorting is what defines a query. This PR fixes the issue when queries with the same filter but different sorting were sharing the same database entity.

### 🛠 Implementation

Sort options are now taken into account when calculating a query hash.

### 🧪 Testing

N/A

### 🎨 Changes

N/A

### ☑️ Checklist

### DIFF
--- a/Sources/StreamChat/Query/ChannelListQuery.swift
+++ b/Sources/StreamChat/Query/ChannelListQuery.swift
@@ -135,3 +135,17 @@ public struct ChannelListQuery: Encodable {
         try pagination.encode(to: encoder)
     }
 }
+
+extension ChannelListQuery {
+    var queryHash: String {
+        let filterHash = filter.filterHash
+        
+        let sortingHash = sort
+            .map(\.description)
+            .joined(separator: ",")
+        
+        return [filterHash, sortingHash]
+            .filter { !$0.isEmpty }
+            .joined(separator: " | ")
+    }
+}
--- a/Sources/StreamChat/Query/ChannelListQuery_Tests.swift
+++ b/Sources/StreamChat/Query/ChannelListQuery_Tests.swift
@@ -3,6 +3,7 @@
 //
 
 @testable import StreamChat
+import StreamChatTestTools
 import XCTest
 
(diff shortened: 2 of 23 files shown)


Sources/StreamChat/Query/ChannelListQuery_Tests.swift
```diff
@@ -78,4 +79,75 @@ final class ChannelListFilterScope_Tests: XCTestCase {
...
+        let query: ChannelListQuery = .mock()
+        
+        // Declare a query with another messages limit
+        let modifiedQuery: ChannelListQuery = .mock(messagesLimit: query.messagesLimit + 10)
+        
+        // Assert hashes match
+        XCTAssertEqual(query.queryHash, modifiedQuery.queryHash)
+    }
+    
+    func test_queryHash_doesNotTakeWatchOptionsIntoAccount() {
```
tbarbugli, line 143: https://getstream.io/chat/docs/javascript/query_channels/?language=javascript#query-options

Sources/StreamChat/Query/ChannelListQuery.swift
```diff
@@ -135,3 +135,17 @@ public struct ChannelListQuery: Encodable {
         try pagination.encode(to: encoder)
     }
 }
+
+extension ChannelListQuery {
+    var queryHash: String {
```
polqf, line 140: Is there any reason why we are not using Hashable for that? Do we still want to keep it human readable? If that's the case, is the queryHash format documented somewhere?

evsaev: Hashable uses a random seed while we need it to be stable across app launches so we can determine if there is a query in the database already for the given combination of filter + sorting params. We decided that we should do real hashing so it's a one-way operation.

I'll get back to this topic when reconnection part is done and update the PR so uses some hashing algo instead of taking string representation as a query hash.

bielikb:
This is stale for couple of weeks already. Im closing it for now.

PR #1627, opened 2021-11-15
[CIS-1328] Dismiss suggestions popup when tapping outside

nuno-vieira:
### 🔗 Issue Link
CIS-1328

### 🎯 Goal
When someone taps the message list, it should dismiss the suggestions popup.

### 🛠 Implementation
I've tried multiple approaches, the current one (closure based) was the most simple and the one that didn't cause any breaking changes. 

Other approaches:
- Adding a gesture recognition to both `ChatChannelVC` and `ChatThreadVC` would not work because there is already a gesture recognizer to the message list, and these would steal the events from the first gesture recognition.
- A new event could be added to the `ChatMessageListVCDelegate` but that would be a breaking change unless we added a default empty implementation, but it is a bit overkill to add a new event to the delegate to handle such a simple use case.
- Moving the message list view gesture recognizer a layer above to ChatChannel and ChatThread VC's. This would look a little bit odd since would still need to keep the `handleTap` inside the `messageListVC` and forward the event to that so that we would not break the API.

**UPDATE: The delegate approach was chosen to make the SDK consistent.** 

### 🧪 Testing
- Open the commands suggestions
- Tap the message list
- Verify the commands is dismissed

### 🎨 Changes

https://user-images.githubusercontent.com/12814114/141822402-f1535f17-7aa9-4864-a29a-08ef8e429609.mp4

### ☑️ Checklist


Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -80,6 +80,9 @@ open class ChatMessageListVC:
 
         return !listView.isLastCellFullyVisible && isMoreContentThanOnePage
     }
+
+    /// A callback closure that is called when the message list is tapped
+    open var didTapOnMessageListView: ((UITapGestureRecognizer) -> Void)?
```
evsaev: hmm, why not a new delegate method? 🤔

evsaev (approved):
Looks good 🔥 Just one nit regarding closure vs the delegate that we already use for callbacks

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
evsaev: ok, found the answer in the description 😄

evsaev: I would personally go with a new delegate method + empty implementation since we already handle all the actions from the message list this way. But leaving it's up to you, other things look solid 👍

nuno-vieira: I've mentioned this in the PR description. It is kinda overkill to add a new delegate event for this and will add maintenance costs. Because we will need to add an empty implementation to this new delegate entry so that we don't break the API. IMO It is not worth it, even though the delegate would be more consistent. Not sure on this one. I feel that the delegates are not that great in SDK, because they easily break the API, unless we provide default implementation, which most of the cases it causes more problems.

evsaev (approved):
🚢

PR #1630, opened 2021-11-17
[CIS-1329] Get rid of `Filter.explicitHash`

evsaev:
### 🔗 Issue Link

CIS-1329

### 🎯 Goal

The goal of this PR is to remove the explicit hash from the `Filter` type and the extra complexity/indirections that it's causing. 

It was initially introduced for background workers that try to link the newly inserted entity to local queries by sending the API request with `.and([originalFilter, .equals("id", to: entity.id)])` filter. If the entity was returned from the APIs the query should be updated with a new entity and filter so if `explicitHash` was not set another query is created in the database because the new filter hash is different from the hash of the original filter.

Now we are moving towards removing background workers for all the queries and moving the decision to the controller level which is more flexible, reduces the # of API calls from the client but still allows to sort/filter by custom fields.

### 🛠 Implementation

- Existed background workers are updated to manually link the new entity to the original local query
- `ChatUserSearchController` is updated to store search results in memory without `UserListQuery` persited in database

### 🧪 Testing

To test changes applied to `ChatUserSearchController`:
- Run `DemoApp`
- Sign in as any user, wait for channel list to be loaded
- Tap `+` icon to open new channel flow
- Play with the search field and search results pagination

To test changes in `DatabaseCleanupUpdater`:
- Run `DemoApp`
- Sign in as any user
- Turn OFF the Internet
- Send a message from another user
- Turn ON the Internet
- See that new message appears in channel/channel preview

### 🎨 Changes

⚠️ This is pointing to an integration branch. No changes will be made to `main` until this feature is complete

### ☑️ Checklist

### DIFF
--- a/DemoApp/CreateChatViewController.swift
+++ b/DemoApp/CreateChatViewController.swift
@@ -65,8 +65,8 @@ class CreateChatViewController: UIViewController {
     
     var searchController: ChatUserSearchController!
     
-    var users: LazyCachedMapCollection<ChatUser> {
-        searchController.users
+    var users: [ChatUser] {
+        searchController.userArray
     }
     
     var selectedUserIds: Set<String> {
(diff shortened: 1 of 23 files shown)


b-onc (approved):
`fetch` family of functions make sense

Issue #1629, opened 2021-11-17
When Pod Updated Version 4.1.0 to 4.5.0  errors found .


vineethnkforios:
## What did you do?
Updated Streamchat Version 4.1.0 to 4.5.0

## What did you expect to happen?
No errors

## What happened instead?
2 errors found 
Error No1:
Class Name : MessageDTO.swift
Function Name: func saveMessage(payload: MessagePayload, for cid: ChannelId?) throws -> MessageDTO? {

Error No2:
Class Name : DatabaseSession.swift
Function Name: func saveEvent(payload: EventPayload) throws {

See the screenshots
![Screen Shot 2021-11-17 at 1 29 18 AM](https://user-images.githubusercontent.com/78809766/142149703-f9743c4a-efc2-4060-8b95-8ed152f64ff6.png)
![Screen Shot 2021-11-17 at 1 30 00 AM](https://user-images.githubusercontent.com/78809766/142149708-9e791e4d-9807-4d6c-b5c7-7e83aef98978.png)

Modified the code like below and seems ok for now. Can we proceed like this ?
***************************************
MessageDTO.swift
***************************************.  *************************************** 
func saveMessage(payload: MessagePayload, for cid: ChannelId?) throws -> MessageDTO? {
        guard payload.channel != nil || cid != nil else {
            throw ClientError.MessagePayloadSavingFailure("""
            Either `payload.channel` or `cid` must be provided to sucessfuly save the message payload.
            - `payload.channel` value: \(String(describing: payload.channel))
            - `cid` value: \(String(describing: cid))
            """)
        }

        if let cid = cid, let payloadCid = payload.channel?.cid {
            log.assert(cid == payloadCid, "`cid` provided is different from the `payload.channel.cid`.")
        }

        var channelDTO: ChannelDTO?

        if let channelPayload = payload.channel {
            channelDTO = try saveChannel(payload: channelPayload, query: nil)
        } else if let cid = cid {
            channelDTO = ChannelDTO.load(cid: cid, context: self)
        } else {
            log.assertionFailure("Should never happen because either `cid` or `payload.channel` should be present.")
            return nil
        }
        /*
        guard let channelDTO = channelDTO else {
            log.assertionFailure("Should never happen, a channel should have been fetched.")
            return nil
        }*/
        
        return try saveMessage(payload: payload, channelDTO: channelDTO!)
    }
***************************************.  ***************************************. 

DatabaseSession.swift

[...]

PR #1632, opened 2021-11-18
Add Xcode 12.5.1 + Fix XCode compatibility issues

b-onc:
Testing CI

### DIFF
--- a/.github/workflows/smoke-checks.yml
+++ b/.github/workflows/smoke-checks.yml
@@ -88,6 +88,49 @@ jobs:
         MATRIX_CONTEXT: ${{ toJson(matrix) }}
       if: ${{ github.event_name == 'push' && failure() }}
 
+  build-and-test-xcode12-debug:
+    name: Test LLC (Xcode 12, Debug)
+    runs-on: macos-11
+    steps:
+    - uses: actions/checkout@v1
+    - name: Set build image var
+      run: echo "ImageVersion=$ImageVersion"  >> $GITHUB_ENV
+    - name: Cache RubyGems
+      uses: actions/cache@v2
+      id: rubygem-cache
+      with:
+        path: vendor/bundle
+        key: ${{ runner.os }}-${{ env.ImageVersion }}-gem-${{ hashFiles('**/Gemfile.lock') }}
+        restore-keys: ${{ runner.os }}-${{ env.ImageVersion }}-gem-
+    - name: Cache Mint
+      uses: actions/cache@v2
+      id: mint-cache
+      with:
+        path: /usr/local/lib/mint
+        key: ${{ runner.os }}-mint-${{ hashFiles('./Mintfile') }}
+        restore-keys: ${{ runner.os }}-mint-
+    - uses: ./.github/actions/bootstrap
+    - name: Run LLC Tests (Debug)
+      env:
+        DEVELOPER_DIR: "/Applications/Xcode_12.5.1.app/Contents/Developer"
+      run: bundle exec fastlane test device:"iPhone 12"
+    - uses: codecov/codecov-action@v1
+      with:
+        token: ${{ secrets.CODECOV_TOKEN }}
+        flags: llc-tests
+        fail_ci_if_error: true # if codecov upload fails, should fail ci
+        verbose: true
+    - uses: 8398a7/action-slack@v3
+      with:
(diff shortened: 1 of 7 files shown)


.github/workflows/smoke-checks.yml
```diff
@@ -88,6 +88,49 @@ jobs:
...
+      id: mint-cache
+      with:
+        path: /usr/local/lib/mint
+        key: ${{ runner.os }}-mint-${{ hashFiles('./Mintfile') }}
+        restore-keys: ${{ runner.os }}-mint-
+    - uses: ./.github/actions/bootstrap
+    - name: Run LLC Tests (Debug)
+      env:
+        DEVELOPER_DIR: "/Applications/Xcode_12.5.1.app/Contents/Developer"
+      run: bundle exec fastlane test device:"iPhone 12"
```
nuno-vieira, line 116: Can we simply just build the project for now? I don't think it is really worth it, at least for now, to run all the tests on Xcode 12.

nuno-vieira: This will also increase the complexity and will require changes on Codecov. So think, for now, building Xcode 12 is enough to catch these errors.

Sources/StreamChat/Workers/MessageUpdater_Tests.swift
```diff
@@ -988,9 +988,7 @@ final class MessageUpdater_Tests: XCTestCase {
         // wait for the db call to be done
         wait(for: [dbCall], timeout: 0.1)
 
-        let reaction = database.viewContext.reaction(messageId: messageId, userId: userId, type: reactionType)
-
-        guard let reaction = reaction else {
+        guard let reaction = database.viewContext.reaction(messageId: messageId, userId: userId, type: reactionType) else {
```
bielikb, line 991: This is just a nitpick.

In cases like these, for readability/maintenance reasons we tend to break params on the newline.

```suggestion
        guard let reaction = database.viewContext.reaction(messageId: messageId, 
                                                                                               userId: userId, 
                                                                                               type: reactionType) else {
```

b-onc: We set our formatter line length to 132 and it does this automatically, the line doesn't exceed the limit

bielikb (dismissed):

.github/workflows/smoke-checks.yml
```diff
@@ -88,6 +88,49 @@ jobs:
         MATRIX_CONTEXT: ${{ toJson(matrix) }}
       if: ${{ github.event_name == 'push' && failure() }}
 
+  build-and-test-xcode12-debug:
+    name: Test LLC (Xcode 12, Debug)
```
nuno-vieira, line 92: We are going to add 15min+ to our CI for nothing. Tests should not change based on Xcode. We should only build the project not run the full test suite.

nuno-vieira (changes requested):

.github/workflows/smoke-checks.yml
b-onc: @tbarbugli handed the PR over to me, and we agreed that you're right. I'm updating the PR so we only build and not test the project.

tbarbugli:
This PR fixes #1629

.github/workflows/smoke-checks.yml
```diff
@@ -113,7 +113,7 @@ jobs:
     - name: Run LLC Tests (Debug)
       env:
         DEVELOPER_DIR: "/Applications/Xcode_12.5.1.app/Contents/Developer"
-      run: bundle exec fastlane test device:"iPhone 12"
+      run: bundle exec fastlane test device:"iPhone 12" build_for_testing:true
```
nuno-vieira, line 116: This will only run the LLC right? We need to run this for the UI SDK as well. 
Or we add another lane with `bundle exec fastlane test_ui device:"iPhone 12" build_for_testing:true`
Or we just build the demo app

.github/workflows/smoke-checks.yml
nuno-vieira: @b-onc Can you please update the job name?

nuno-vieira: In case we are doing 2 lanes, we can name "Build LLC (Xcode 12)" and "Build UI (Xcode 12)"

PR #1642, opened 2021-11-23
Added identifier for tracking the usage of the different SDKs

martinmitrevski:
### 🔗 Issue Link
[JIRA](https://stream-io.atlassian.net/browse/SWUI-33)

### 🎯 Goal

We need to track the usage of the different SDKs (low-level client, UIKit and SwiftUI), via the custom HTTP header X-Stream-Client.

### 🛠 Implementation

In the implementation, we're relying on whether the appropriate SDK can be imported, using the `canImport` macro.

The currently provided values are "swiftui", "uikit" and "swift". The other option would be to keep "swift" for "uikit", for backward compatibility, and use something like "llc" for the low-level client.

### 🧪 Testing

Test the DemoApp in this repo, by putting a breakpoint after setting the `httpAdditionalHeaders` in `ChatClient`, after line  186.

Test the DemoAppSwiftUI app in https://github.com/GetStream/stream-chat-swiftui. Note that you will need to setup the Swift package dependency to this branch (feature/tracking-sdks).

### 🎨 Changes

Not really visible changes.

### ☑️ Checklist

### DIFF
--- a/Sources/StreamChat/ChatClient.swift
+++ b/Sources/StreamChat/ChatClient.swift
@@ -189,8 +189,19 @@ public class ChatClient {
     
     /// Stream-specific request headers.
     private let sessionHeaders: [String: String] = [
-        "X-Stream-Client": "stream-chat-swift-client-v\(SystemEnvironment.version)"
+        "X-Stream-Client": "stream-chat-\(sdkIdentifier)-client-v\(SystemEnvironment.version)"
     ]
+    
+    /// Identifies which SDK is being used.
+    private static var sdkIdentifier: String {
+        #if canImport(StreamChatSwiftUI)
+        return "swiftui"
+        #elseif canImport(StreamChatUI)
+        return "uikit"
+        #else
+        return "swift"
+        #endif
+    }
     
     /// The current connection id
     @Atomic var connectionId: String?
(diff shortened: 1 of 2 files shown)


Sources/StreamChat/ChatClient.swift
```diff
@@ -189,9 +189,20 @@ public class ChatClient {
...
+    /// Identifies which SDK is being used.
+    private static var sdkIdentifier: String {
+        #if canImport(StreamChatSwiftUI)
+        return "swiftui"
+        #elseif canImport(StreamChatUI)
+        return "uikit"
+        #else
+        return "swift"
+        #endif
+    }
```
nuno-vieira, line 204: Tricky question: What if a customer uses both `StreamChatSwiftUI` and `StreamChatUI` ? 🤔 Is that even possible?  Is this something we want to support? Or SwiftUI SDK will have totally feature parity with the UIKit SDK?

martinmitrevski: The plan is to have feature parity with the UIKit SDK. In general, it will not make sense for both of them to be used in the same app, unless some gradual migration is done. But in those cases, the "swiftui" will be returned, which is somehow correct.

nuno-vieira (dismissed):
LGTM! ✅

Sources/StreamChat/ChatClient.swift
tbarbugli: if you use both than we should still track swiftui

bielikb (approved):

PR #1646, opened 2021-11-24
[CIS-1273] Ensure optional Core Data attributes map to optional Swift types

jeroenleenarts:
### 🔗 Issue Link
[CIS-1273](https://stream-io.atlassian.net/browse/CIS-1273)

### 🎯 Goal

A number of differences between the Core Data model and the ManagedObjects were present. This pull request removes those issues. When applicable a sensible default is chosen for what Core Data describes as primitive types.

### 🛠 Implementation

Several model definitions were changed, several managed object classes were changed, several sensible default values were chosen.
### 🧪 Testing

Run through the app. No crashes should occur.

### 🎨 Changes

_Add relevant screenshots or videos showcasing the changes._

### ☑️ Checklist

### DIFF
--- a/Sources/StreamChat/Database/DTOs/AttachmentDTO.swift
+++ b/Sources/StreamChat/Database/DTOs/AttachmentDTO.swift
@@ -15,9 +15,9 @@ class AttachmentDTO: NSManagedObject {
     }
 
     /// An attachment type.
-    @NSManaged private var type: String
+    @NSManaged private var type: String?
     var attachmentType: AttachmentType {
-        get { .init(rawValue: type) }
+        get { AttachmentType(rawValue: type ?? AttachmentType.unknown.rawValue) }
         set { type = newValue.rawValue }
     }
 
@@ -25,7 +25,9 @@ class AttachmentDTO: NSManagedObject {
     @NSManaged private var localStateRaw: String
     @NSManaged private var localProgress: Double
     var localState: LocalAttachmentState? {
-        get { LocalAttachmentState(rawValue: localStateRaw, progress: localProgress) }
+        get {
+            LocalAttachmentState(rawValue: localStateRaw, progress: localProgress)
+        }
         set {
             localStateRaw = newValue?.rawValue ?? ""
             localProgress = newValue?.progress ?? 0
@@ -221,6 +223,8 @@ extension AttachmentDTO {
 extension LocalAttachmentState {
     var rawValue: String {
         switch self {
+        case .unknown:
+            return ""
         case .pendingUpload:
             return "pendingUpload"
         case .uploading:
@@ -251,8 +255,10 @@ extension LocalAttachmentState {
             self = .uploadingFailed
         case LocalAttachmentState.uploaded.rawValue:
             self = .uploaded
+        case LocalAttachmentState.unknown.rawValue:
+            self = .unknown
         default:
-            return nil
+            self = .unknown
         }
     }
 }
--- a/Sources/StreamChat/Database/DTOs/MessageDTO.swift
+++ b/Sources/StreamChat/Database/DTOs/MessageDTO.swift
@@ -20,7 +20,7 @@ class MessageDTO: NSManagedObject {
     @NSManaged var parentMessageId: MessageId?
     @NSManaged var showReplyInChannel: Bool
     @NSManaged var replyCount: Int32
-    @NSManaged var extraData: Data
+    @NSManaged var extraData: Data?
     @NSManaged var isSilent: Bool
     @NSManaged var isShadowed: Bool
     @NSManaged var reactionScores: [String: Int]
@@ -280,6 +280,8 @@ class MessageDTO: NSManagedObject {
         
         let new = NSEntityDescription.insertNewObject(forEntityName: entityName, into: context) as! Self
         new.id = id
+        new.latestReactions = []
+        new.ownReactions = []
         return new
     }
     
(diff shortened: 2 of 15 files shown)


jeroenleenarts:
@tbarbugli I looked into automatically comparing the model definitions in the core data model against the managed object definitions. I was unsuccessful. The one thing I kept running into is the lack of information on the managed objects at runtime. I have no way to request the nullability of properties defined on te class level.

 Getting the same info from Core Data is easy.

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
```diff
@@ -10,7 +10,7 @@ class ChannelDTO: NSManagedObject {
     @NSManaged var cid: String
     @NSManaged var name: String?
     @NSManaged var imageURL: URL?
-    @NSManaged var typeRawValue: String
+    @NSManaged var typeRawValue: String?
```
tbarbugli, line 13: maybe better if we do the way around? (saving a channel without its type does not make a lot of sense)

Sources/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -36,7 +36,7 @@ class MessageDTO: NSManagedObject {
     @NSManaged var replies: Set<MessageDTO>
     @NSManaged var flaggedBy: CurrentUserDTO?
     @NSManaged var attachments: Set<AttachmentDTO>
-    @NSManaged var quotedMessage: MessageDTO?
+    @NSManaged var quotedMessage: MessageDTO? // TODO: Jeroen: this thing has an Id postfix in the model file
```
tbarbugli, line 39: what does it mean?

Sources/StreamChat/Database/DTOs/MessageReactionDTO.swift
```diff
@@ -7,15 +7,15 @@ import Foundation
 
 @objc(MessageReactionDTO)
 final class MessageReactionDTO: NSManagedObject {
-    @NSManaged fileprivate var id: String
+    @NSManaged fileprivate var id: String?
```
tbarbugli, line 10: perhaps the way around is better? (id not optional in core data)

tbarbugli (commented):
Looks pretty solid, I just left a couple comments

Sources/StreamChat/Database/DTOs/AttachmentDTO.swift
```diff
@@ -15,17 +15,22 @@ class AttachmentDTO: NSManagedObject {
...
-    @NSManaged private var localStateRaw: String
+    @NSManaged private var localStateRaw: String?
     @NSManaged private var localProgress: Double
     var localState: LocalAttachmentState? {
-        get { LocalAttachmentState(rawValue: localStateRaw, progress: localProgress) }
+        get {
+            if let localStateRaw = localStateRaw {
+                return LocalAttachmentState(rawValue: localStateRaw, progress: localProgress)
+            }
+            return nil
```
bielikb, line 32: A nitpick with guard
```suggestion
            guard let localStateRaw = localStateRaw else {
                return nil
            }
            return nil LocalAttachmentState(rawValue: localStateRaw, progress: localProgress)
```

There's also shorter, one-liner version that imho goes against the readability & is quite confusing due to combination of ternary operator + force unwrap, so I don't recommend going that far :) 

```
localStateRaw != nil ? LocalAttachmentState(rawValue: localStateRaw!, progress: localProgress) : nil
```

jeroenleenarts: I have not touched this bit of code. But I would do it like this:
```
          return localStateRaw.flatMap { localStateRaw in
                return LocalAttachmentState(rawValue: localStateRaw, progress: localProgress)
            }
```

bielikb (approved):

bielikb (dismissed):

Sources/StreamChat/Database/DTOs/AttachmentDTO.swift
```diff
@@ -15,17 +15,19 @@ class AttachmentDTO: NSManagedObject {
...
-        get { .init(rawValue: type) }
+        get { AttachmentType(rawValue: type ?? AttachmentType.unknown.rawValue) }
         set { type = newValue.rawValue }
     }
 
     /// An attachment local state.
-    @NSManaged private var localStateRaw: String
+    @NSManaged private var localStateRaw: String?
     @NSManaged private var localProgress: Double
     var localState: LocalAttachmentState? {
```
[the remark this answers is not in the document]
jeroenleenarts: Not at the moment, in fact the setter applies an empty string, which has no mapping in the LocalAttachmentState enum. So this could be the unknown marker.

bielikb (dismissed):

Sources/StreamChat/Database/DTOs/AttachmentDTO.swift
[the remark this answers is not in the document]
polqf: Should we add that .unknown case then?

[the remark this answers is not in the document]
jeroenleenarts: Pushed a change on your feedback.

Sources/StreamChat/Database/DTOs/AttachmentDTO.swift
```diff
@@ -22,11 +22,11 @@ class AttachmentDTO: NSManagedObject {
     }
 
     /// An attachment local state.
-    @NSManaged private var localStateRaw: String?
+    @NSManaged private var localStateRaw: String
     @NSManaged private var localProgress: Double
     var localState: LocalAttachmentState? {
         get {
-            localStateRaw.flatMap { LocalAttachmentState(rawValue: $0, progress: localProgress) }
+            LocalAttachmentState(rawValue: localStateRaw, progress: localProgress)
```
polqf, line 29: What about `LocalAttachmentState(rawValue: localStateRaw, progress: localProgress) ?? .unknown` so we don't expose an optional?

bielikb (approved):

PR #1651, opened 2021-11-26
[CIS-1390] Refactor `ChatMessageLayoutOptions` with more flexible and safer approach

nuno-vieira:
### 🔗 Issue Link
[CIS-1390](https://stream-io.atlassian.net/browse/CIS-1390)

### 🎯 Goal
Make the `ChatMessageLayoutOptions` easier to customize by using a `Set` instead of `OptionSet`.  

### 🛠 Implementation
Introduce a `ChatMessageLayoutOption` which will be each option, and add a typealias `ChatMessageLayoutOptions = Set<ChatMessageLayoutOption>` for backwards-compability.

### 🧪 Testing
Smoke testing should be done

#### How to add a new custom message layout option:
**Before**
 ```swift
 extension ChatMessageLayoutOptions {
   static let customOption = Self(rawValue: 1 << 14)
 }
 ```
**After**
 ```swift
 extension ChatMessageLayoutOption {
   static let customOption: Self = "customOption"
 }
 ```

### ☑️ Checklist


Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageLayoutOptions.swift
```diff
@@ -4,68 +4,95 @@
...
-    static let centered = Self(rawValue: 1 << 13)
+    static let centered: Self = "centered"
 }
 
-extension ChatMessageLayoutOptions: CustomStringConvertible {
+extension ChatMessageLayoutOption: CustomStringConvertible {
     /// Returns all options the current option set consists of separated by `-` character.
     public var description: String {
         Self.singleOptions
-            .compactMap { contains($0) ? $0.optionName : nil }
```
nuno-vieira, line 68: @evsaev Didn't quite understand this code, what was this doing before? 🤔 Why was the contains needed?

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageLayoutOptions.swift
```diff
@@ -4,68 +4,95 @@
...
+
+    mutating func insert(_ options: ChatMessageLayoutOptions) {
+        options.forEach { self.insert($0) }
+    }
+}
+
 /// Describes the layout of base message content view.
-public struct ChatMessageLayoutOptions: OptionSet, Hashable {
-    public let rawValue: Int
+public struct ChatMessageLayoutOption: RawRepresentable, Hashable, Equatable, ExpressibleByStringLiteral {
```
evsaev, line 48: nit: conforming to `Hashable` should be enough
```suggestion
public struct ChatMessageLayoutOption: RawRepresentable, Hashable, ExpressibleByStringLiteral {
```

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageLayoutOptions.swift
```diff
@@ -4,68 +4,95 @@
...
+    static let centered: Self = "centered"
 }
 
-extension ChatMessageLayoutOptions: CustomStringConvertible {
+extension ChatMessageLayoutOption: CustomStringConvertible {
     /// Returns all options the current option set consists of separated by `-` character.
     public var description: String {
         Self.singleOptions
-            .compactMap { contains($0) ? $0.optionName : nil }
+            .compactMap(\.optionName)
```
evsaev: We were checking if self (option set) contains a single option and if it does we were taking the name of this concrete option. Now it should be:
```swift
Array($0).sorted().compactMap(\.optionName)
```

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageLayoutOptions.swift
```diff
@@ -4,120 +4,114 @@
...
+        .timestamp,
+        .authorName,
+        .text,
+        .quotedMessage,
+        .threadInfo,
+        .errorIndicator,
+        .reactions,
+        .onlyVisibleForYouIndicator,
+        .centered
+    ]
```
nuno-vieira, line 44: @evsaev I did not understand why we need these? 🤔

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageLayoutOptions.swift
```diff
@@ -4,120 +4,114 @@
...
-        case .reactions:
-            return "reactions"
-        case .onlyVisibleForYouIndicator:
-            return "onlyVisibleForYouIndicator"
-        case .centered:
-            return "centered"
-        default:
-            return nil
-        }
-    }
```
nuno-vieira, line 122: It looks there were not public before, so we can remove them?

Sources/StreamChatUI/MessageActionsPopup/MessageActionsTransitionController.swift
```diff
@@ -89,7 +89,7 @@ open class ChatMessageActionsTransitionController: NSObject, UIViewControllerTra
         let messageViewFrame = selectedMessageContentViewFrame ?? .zero
         let messageViewType = type(of: originalMessageContentView)
         let messageAttachmentInjectorType = originalMessageContentView.attachmentViewInjector.map { type(of: $0) }
-        let messageLayoutOptions = originalMessageContentView.layoutOptions?.subtracting(.reactions) ?? []
+        let messageLayoutOptions = originalMessageContentView.layoutOptions?.subtracting([.reactions]) ?? []
```
evsaev, line 92: shall we introduce `subtracting` with a single item on `Set`?  🤔 if customers do that on their end it will stop compiling but the same can happen with all other set APIs

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageLayoutOptions.swift
```diff
@@ -4,68 +4,95 @@
...
+    mutating func insert(_ options: ChatMessageLayoutOptions) {
+        options.forEach { self.insert($0) }
+    }
+}
+
 /// Describes the layout of base message content view.
-public struct ChatMessageLayoutOptions: OptionSet, Hashable {
-    public let rawValue: Int
+public struct ChatMessageLayoutOption: RawRepresentable, Hashable, Equatable, ExpressibleByStringLiteral {
+    public typealias StringLiteralType = String
```
evsaev, line 49: nit: we can get rid of it since string literal type is resolved by `init(stringLiteral: String)`

evsaev (dismissed):
Great job 👏

Sources/StreamChatUI/MessageActionsPopup/MessageActionsTransitionController.swift
nuno-vieira: Yes, I spotted this one... It seems to be an overkill to do that IMO, we can add all these but I think it is an Overkill and most likely it is not being used... Hopefully 🤔 Otherwise we need to add this to `subtracting` `instersection` and `insert` right?

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageLayoutOptions.swift
polqf: +1. What does singleOptions mean? is it meant to be like the `allCases` in an enum?

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageLayoutOptions.swift
```diff
@@ -4,120 +4,114 @@
 
 import Foundation
 
+public typealias ChatMessageLayoutOptions = Set<ChatMessageLayoutOption>
+
+public extension ChatMessageLayoutOptions {
+    var rawValue: String {
+        map(\.rawValue).joined(separator: "-")
```
polqf, line 11: What is this useful for?

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageLayoutOptions.swift
```diff
@@ -4,120 +4,114 @@
...
+        .reactions,
+        .onlyVisibleForYouIndicator,
+        .centered
+    ]
+}
+
 /// Describes the layout of base message content view.
-public struct ChatMessageLayoutOptions: OptionSet, Hashable {
-    public let rawValue: Int
+public struct ChatMessageLayoutOption: RawRepresentable, Hashable, Equatable, ExpressibleByStringLiteral {
```
polqf, line 48: ExpressibleByStringLiteral can confuse our clients. Maybe this should be internal

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageLayoutOptions.swift
```diff
@@ -4,120 +4,114 @@
...
 
-    public init(rawValue: Int) {
+    public let rawValue: String
+
+    public init(rawValue: String) {
         self.rawValue = rawValue
     }
+
+    public init(stringLiteral value: String) {
+        rawValue = value
```
polqf, line 58: IMO there should be one convenience init, and you call it from all the other inits

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageLayoutOptions.swift
```diff
@@ -4,120 +4,114 @@
...
+    }
+
+    @available(*, deprecated, message: "Use the string raw value initialiser.")
+    public init(rawValue: Int) {
+        self.rawValue = "\(rawValue)"
+    }
+
+    var optionName: String? {
```
polqf, line 66: Why is this optional? rawValue is not optional

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageLayoutOptions.swift
```diff
@@ -4,120 +4,114 @@
...
 /// Describes the layout of base message content view.
-public struct ChatMessageLayoutOptions: OptionSet, Hashable {
-    public let rawValue: Int
+public struct ChatMessageLayoutOption: RawRepresentable, Hashable, Equatable, ExpressibleByStringLiteral {
+    public typealias StringLiteralType = String
 
-    public init(rawValue: Int) {
+    public let rawValue: String
+
+    public init(rawValue: String) {
```
polqf, line 53: What about 
```suggestion
    public init(_ rawValue: String) {
```
?

I imagine `ChatMessageLayoutOption("NewOption")` being better than `ChatMessageLayoutOption(rawValue: "NewOption")`

nuno-vieira: We can't change this, this initialiser is part of `RawRepresentable`

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageLayoutOptions.swift
nuno-vieira: The previous API had this optional. It was just to not add a breaking change. But I noticed now that his was not exposed before, so I will actually remove it since we don't need it 👍

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageLayoutOptions.swift
nuno-vieira: I think it will make it easier for them to create a new option.

Instead of:
```swift
extension ChatMessageLayoutOption {
    static let likes = ChatMessageLayoutOption(rawValue: "likes")
}
```
They can write:
```swift
extension ChatMessageLayoutOption {
    static let likes: Self = "likes"
}
```

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageLayoutOptions.swift
nuno-vieira: This is used for the `reuseIdentifier` of the message cell. Depending on the options, it will have a different identifier

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageLayoutOptions.swift
nuno-vieira: Why sorted? 🤔 This will mean that the ID for the reuseIdentifier will be different right? :/

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageLayoutOptions.swift
nuno-vieira: I've added documentation ✅

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageLayoutOptions.swift
nuno-vieira: Solved ✅

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageLayoutOptions.swift
nuno-vieira: Removed the `optionName` ✅

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageLayoutOptions.swift
```diff
@@ -4,120 +4,93 @@
...
+        self.init(rawValue: value)
+    }
+
+    // Probably this is not needed. It will be a breaking change anyway to the current customer
+    // since the customer was doing initialising from `ChatMessageLayoutOptions`.
+    //
+//    @available(*, deprecated, message: "Use the string raw value initialiser.")
+//    public init(rawValue: Int) {
+//        self.rawValue = "\(rawValue)"
+//    }
```
nuno-vieira, line 47: I'm planning on removing this. Because this will be a breaking API anyways :/ Since customers were initialising on `ChatMessageLayoutOptions` not in `ChatMessageLayoutOption`

Sources/StreamChatUI/MessageActionsPopup/MessageActionsTransitionController.swift
nuno-vieira: I've added all these backwards compatibility APIs to the Deprecations. I can also add "substracting" and "intersection" to those, WDYT?

nuno-vieira: @evsaev Done ✅

Sources/StreamChatUI/Deprecations.swift
```diff
@@ -165,3 +165,56 @@ extension Components {
...
+        self = intersection([option])
+    }
+
+    @available(*, deprecated, message: """
+        use `contains(_ member: ChatMessageLayoutOption` instead. And make sure the custom option is being extended in
+        `ChatMessageLayoutOption` and not in `ChatMessageLayoutOptions`.
+    """)
+    public func contains(_ options: ChatMessageLayoutOptions) -> Bool {
+        let intersected = intersection(options)
+        return intersected.count == options.count
```
b-onc, line 209: let set1 = Set(arrayLiteral: 1,2,3)
let set2 = Set(arrayLiteral: 2,3)

set2.isSubset(of: set1) // true
set1.isSuperset(of: set2) // true

nuno-vieira: This is a deprecated method, which most likely nobody is using or will ever use, so I don't mind the current implementation even the one you provided is more correct. But I can change it if you think it is valuable

nuno-vieira: I've updated it anyways 👍

Sources/StreamChatUI/Deprecations.swift
```diff
@@ -165,3 +166,55 @@ extension Components {
...
+/// typealias `typealias = Set<ChatMessageLayoutOption>` and provided an API like the `OptionSet` so we
+/// don't break the public API.
+
+extension ChatMessageLayoutOptions {
+    @available(*, deprecated, message: "use `id` instead.")
+    var rawValue: String {
+        id
+    }
+
+    @available(*, deprecated, message: "use `remove(_ member: ChatMessageLayoutOption)` instead.")
```
bielikb, line 184: @nuno-vieira Could we provide fix-its instead of deprecation messages?

Would something like the following work?
```suggestion
    @available(*, deprecated, renamed: "remove(_ member: ChatMessageLayoutOption)")
```

bielikb: Actually renamed is not needed in this case, since you provided the `typealias` that will route the calls towards the correct implementation of each respective func ;)

Screenshot 2021-12-03 at 11 39 35

bielikb:
Well done, @nuno-vieira!

bielikb (approved):

PR #1654, opened 2021-11-30
[CIS-1363] Move event batching outside notification center

evsaev:
### 🔗 Issue Link

CIS-1319, CIS-1363

### 🎯 Goal

To move event batching logic outside notification center. When the connection is being recovered and missing events are received from `/sync` endpoint we don't wanna batch these events but process them immidiately and catch the completion so we know when we can move to the next step.

### 🛠 Implementation

Batching is encapsulated in `EventBatcher` type and moved outside the notification center to the web-socket client where the batching should have been done at.

### 🧪 Testing

Play with the `DemoApp` to see that events are still delivered to the UI.
Unit tests are updated to cover the changes.

### 🎨 Changes

N/A

### ☑️ Checklist


Sources/StreamChat/Utils/EventBatcher.swift
```diff
@@ -0,0 +1,84 @@
...
+    init(period: TimeInterval, handler: @escaping (Batch) -> Void)
+    
+    /// Adds the item to the current batch of events. If it's the first event also schedules batch processing
+    /// that will happen when `period` has passed.
+    ///
+    /// - Parameter event: The event to add to the current batch.
+    func append(_ event: Event)
+    
+    /// Ignores `period` and passes the current batch of events to handler as soon as possible.
+    func processImmidiately()
```
b-onc, line 24:
```suggestion
    func processImmediately()
```

Sources/StreamChat/Utils/EventBatcher_Tests.swift
```diff
@@ -0,0 +1,95 @@
...
+        // Assert current batch is empty
+        XCTAssertEqual(batcher.currentBatch, [])
+    }
+}
+
+private extension Batch_Tests {
+    func wait(_ time: TimeInterval) {
+        let start = Date()
+        AssertAsync.willBeTrue(Date().timeIntervalSince(start) >= time)
+    }
```
b-onc, line 88: We might as well use `sleep`?

Sources/StreamChat/Utils/EventBatcher_Tests.swift
```diff
@@ -0,0 +1,95 @@
...
+        
+        // Wait a bit less then period
+        wait(0.05)
+        // Assert current batch contains expected values
+        XCTAssertEqual(batcher.currentBatch, [event1, event2])
+        // Assert handler is not called yet
+        XCTAssertEqual(handlerCalls, [])
+        
+        // Wait another bit so batch period has passed
+        wait(0.2)
```
b-onc, line 41: So many `wait`s in these tests makes me thing the component is not very testable.. How about if we inject a Timer and use it instead of `queue.asyncAfter`?

Sources/StreamChat/WebSocketClient/Events/EventPayload.swift
```diff
@@ -123,3 +123,19 @@ extension EventPayload {
         return value
     }
 }
+
+extension Array where Element == EventPayload {
+    /// Decodes events from event payloads. If decoding of some event fails the error is logged without interruption the chain.
```
b-onc, line 128:
```suggestion
    /// Decodes events from event payloads. If decoding of some event fails the error is logged without interrupting the chain.
```

Sources/StreamChat/WebSocketClient/WebSocketClient.swift
```diff
@@ -8,8 +8,13 @@ class WebSocketClient {
     /// The notification center `WebSocketClient` uses to send notifications about incoming events.
     let eventNotificationCenter: EventNotificationCenter
     
+    /// The batch of events received via the web-socket that wait to be processed.
+    private(set) lazy var eventsBatch = environment.eventBatcherBuilder { [weak self] events in
```
b-onc, line 12:
```suggestion
    private(set) lazy var eventsBatcher = environment.eventBatcherBuilder { [weak self] events in
```

Sources/StreamChat/Workers/EventNotificationCenter.swift
```diff
@@ -33,61 +29,27 @@ class EventNotificationCenter: NotificationCenter {
...
-                self._pendingEvents {
-                    eventsToProcess = $0
-                    $0.removeAll()
+        database.write({ session in
+            eventsToPost = events.compactMap {
+                self.middlewares.process(event: $0, session: session)
+            }
+        }, completion: { _ in
+            self.eventPostingQueue.sync {
+                if post {
```
b-onc, line 41: Why don't we move this `if` check above, to reduce unnecessary `queue.sync` calls?

Sources/StreamChat/Workers/EventNotificationCenter.swift
```diff
@@ -33,61 +29,27 @@ class EventNotificationCenter: NotificationCenter {
         middlewares.append(middleware)
     }
 
-    func process(_ event: Event) {
-        var shouldScheduleProcessing = false
-        _pendingEvents {
-            shouldScheduleProcessing = $0.isEmpty
-            $0.append(event)
-        }
+    func process(_ events: [Event], post: Bool = true, completion: (() -> Void)? = nil) {
```
b-onc, line 32:
```suggestion
    func process(_ events: [Event], postNotification: Bool = true, completion: (() -> Void)? = nil) {
```

the naming can be a little bit clearer

Sources/StreamChat/Utils/EventBatcher.swift
```diff
@@ -0,0 +1,84 @@
...
+}
+
+extension Batcher: EventBatcher where Item == Event {}
+
+final class Batcher<Item> {
```
b-onc, line 29: Do we really need the protocol, and the generic class?

Sources/StreamChat/Utils/EventBatcher.swift
```diff
@@ -0,0 +1,84 @@
...
+    /// The batching period. If the item is added sonner then `period` has passed after the first item they will get into the same batch.
+    private let period: TimeInterval
+    /// The closure which processes the batch.
+    private let handler: ([Item]) -> Void
+    /// The serial queue where item appends and batch processing is happening on.
+    private let queue = DispatchQueue(label: "co.getStream.Batch.\(Item.self)")
```
b-onc, line 35:
```suggestion
    private let queue = DispatchQueue(label: "io.getstream.Batch.\(Item.self)")
```

let's be consistent on these things (because we weren't before)

Sources/StreamChat/Utils/EventBatcher.swift
evsaev: ✅ Fixed

Sources/StreamChat/WebSocketClient/Events/EventPayload.swift
evsaev: ✅ Fixed

Sources/StreamChat/WebSocketClient/WebSocketClient.swift
evsaev: ✅ Fixed

Sources/StreamChat/Workers/EventNotificationCenter.swift
evsaev: ✅ Fixed. Renamed to `postNotifications` in main method and `postNotification` for helper that takes a single event.

Sources/StreamChat/Utils/EventBatcher.swift
evsaev: Good question 🙂 I was thinking like that: we need a protocol to be able to inject a mock that passes events through without batching. But events it's not the only thing we might wanna batch so would be nice if it can be reused for any other item type

Sources/StreamChat/Utils/EventBatcher.swift
evsaev: ✅ Fixed

Sources/StreamChat/Workers/EventNotificationCenter.swift
evsaev: Good stuff, fixed that 👍

Sources/StreamChat/Utils/EventBatcher_Tests.swift
evsaev: Yep, that was my initial idea 👍 If we inject time we still need a serial queue to prevent issues with appending elements in a multithreaded environment. And with both queue + time + timer the implementation gets messy in favor of tests. Multiple `wait` statements can lead to test timeout but we haven't seen these issues for quite some time and if we will that would be a matter of decreasing the batch period used.

b-onc: @evsaev You can make sure a Timer fires it's blocks in a given queue, check DispatchSourceTimer here: https://developer.apple.com/documentation/dispatch/dispatchsource/2300106-maketimersource

Sources/StreamChat/Utils/EventBatcher_Tests.swift
evsaev: Yep, sleep is also an option. It's just that asserts we use are backed by `XCTWaiter + XCTestExpectation` which is the expected way to wait for some stuff to happen and also consistent with other tests. Adding sleep can cause a new anti-pattern to occur which seems to me we'd like to avoid

adamrushy (approved):
All looks good to me, I checked out the branch and all running the Demo App looks good 🤩

b-onc (approved):
LGTM 🔥

Sources/StreamChat/Utils/EventBatcher.swift
```diff
@@ -0,0 +1,79 @@
...
+                timeInterval: self.period,
+                queue: self.queue,
+                onFire: self.process
+            )
+        }
+    }
+    
+    func processImmediately() {
+        timerType.schedule(timeInterval: 0, queue: queue) {
+            self.process()
```
bielikb, line 69: @evsaev  Wont' this cause a memory leak?

evsaev: @bielikb this makes sure the batcher will not be deallocated leaving the last batch unprocessed

bielikb: Per slack convo:

"Yes, that is by design, there’s no permanent retain cycle but the temporary one. Meaning that if no strong refs to batcher exist outside the object, it will be kept alive by the scheduled closure", Pavel Evsaev, 2021

bielikb (approved):
Looks good to me. nice work!

Issue #1647, opened 2021-11-25
After seen all the messages , red dot is not disappearing from channel.


iosiappmemes:
Hi Team , I have found one bug while using StreamChat 4.0 SDK , 
I got a message from a friend , on the channel list it shows red dot on new message.  when I clicked on that channel and view all the messages and I came back from the messages screen , the red dot not disappear from the channel.

Here is the Screen shot for ref-
![Simulator Screen Shot - iPhone 13 Pro - 2021-11-25 at 13 10 51](https://user-images.githubusercontent.com/54673061/143399666-3eaa29e4-c801-43bb-85bd-4484967d5eef.png)

adamrushy:
Hey, @iosiappmemes 

Thanks so much for sending this through to us, to help me investigate further can you provide the following:

- Steps to reproduce
- Environment details (iOS version, Xcode version etc...)
- Sample code

If you can reproduce this in a sample project or share with us your current implementation so I can see what's happening. I have tried reproducing in our Demo App but it's working as expected. 

Thank you,
Adam

iosiappmemes:
@adamrushy , this is our client based live product so I can't share the project , and I don't have any sample project. 
Just let me know one thing, I am using default ui for chat , is there anything I need to call manually to make channel read all the messages ? any Boolean value or any api I need to call ?

iosiappmemes:
IOS Version :- 15.0
Xcode Version: - 13.0

iosiappmemes:
Steps to reproduce , 
Initiate user on two device , suppose names are A,B. 
A sent a message to B , when B goes back to channel list from messages controller , it is showing red dot, but B has already seen that message.
Channel should be refresh automatically and red dot should be disappear.

b-onc:
@iosiappmemes can you share a video?

vladpre92:
@iosiappmemes You should update to the last version. #1569 is the fix related to this problem and was added to version 4.2.0.

iosiappmemes:
Okay Thanks.

iosiappmemes:
Hi @vladpre92 , I reviewed your code and I have found the issue in your code .
Actually there are two issue , because of them red dot is not disappearing.
Even I have installed latest SDK as you mentioned 4.2.0, still there are issues available.

issue one :-  This is the code where thread is returning from without marking channel read.
Try to debug by break points.
guard channel?.latestMessages.first?.author.id != userId else {
            callback {
                completion?(nil)
            }
            return
        }
[...]

PR #1656, opened 2021-11-30
[CIS-1369] Improve stability of marking channel read

nuno-vieira:
### 🔗 Issue Link
CIS-1369
https://github.com/GetStream/stream-chat-swift/issues/1647

### 🎯 Goal
Improve the stability of marking the channel read for edge cases.

BONUS: Updates SPM packages. 

### 🛠 Implementation
- Removes the verification of the last message to be from the same user. In rare scenarios, this might not be the case. 
- Make an extra call on `viewDidAppear`, since sometimes `scrollViewDidScroll` event is not called when opening the view controller.

Note: No tests were added/removed because this logic was not even being tested.

### 🧪 Testing
1. Downgrade to 4.1.0 or 4.0.0.
2. Reproduce the bug that #1569 fixes in 4.2.0 with the specification that after receiving the message from the other user (that makes unreadCount = 1), you send a message. This makes a channel with unreadCount = 1 and last message sent by current user. 
3. Kill the app or go back to channel list but without entering the unread channel again.
4. Update StreamChat to 4.2.0+.
5. You'll have a channel that will not be marked as read until the other user sends a message, bypassing `currentUserId != message.author.id` (that was introduced with 4.2.0). 

### 🎨 Changes
N/A

### ☑️ Checklist

### DIFF
--- a/Integration/SPM/SwiftPackageManager.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/Integration/SPM/SwiftPackageManager.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -6,8 +6,8 @@
         "repositoryURL": "https://github.com/kean/Nuke.git",
         "state": {
           "branch": null,
-          "revision": "472a03c15f39592007d94f25abe554350bc8aa2b",
-          "version": "10.5.1"
+          "revision": "961597d324949c7003c77d602bfc4d5519b47e81",
+          "version": "9.2.4"
         }
       },
       {
@@ -24,8 +24,8 @@
         "repositoryURL": "https://github.com/kirualex/SwiftyGif.git",
         "state": {
           "branch": null,
-          "revision": "652b344111a363a53af602b9eef0b52ab1146b42",
-          "version": "5.4.1"
+          "revision": "0cd770feed5807d9fd4fa8df5ad74b56b814b0c8",
+          "version": "5.4.0"
         }
       }
     ]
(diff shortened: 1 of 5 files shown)


Integration/SPM/SwiftPackageManager.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -6,8 +6,8 @@
         "repositoryURL": "https://github.com/kean/Nuke.git",
         "state": {
           "branch": null,
-          "revision": "961597d324949c7003c77d602bfc4d5519b47e81",
-          "version": "9.2.4"
+          "revision": "472a03c15f39592007d94f25abe554350bc8aa2b",
+          "version": "10.5.1"
```
polqf, line 10: Shouldn't we discard this change? We have not have enough time to check if there's any issue that this can cause.

polqf (commented):
Looks good to me based on what I can grasp

Integration/SPM/SwiftPackageManager.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
bielikb: @polqf I was thinking the same. We can simply discard that change on Nuno's PR as it's non-essential for the work alone.

polqf (approved):

bielikb (approved):

PR #1672, opened 2021-12-07
Fix error messages being rendered incorrectly and interactive

b-onc:
### 🎯 Goal

Error type messages, which are returned from backend when an invalid command is send (like `/invalidcommand`), should be rendered correctly

### 🛠 Implementation

Before, these messages were rendered as normal messages, and were interactive (you can reply and react to them). They should be like system messages (designs state so) and non-interactive

### 🧪 Testing

Send an invalid command (like `/gif`) to a channel

### 🎨 Changes

Before | After
--- | ---
Screen Shot 2021-12-07 at 17 51 29 | Screen Shot 2021-12-07 at 17 52 05

### ☑️ Checklist


bielikb (approved):

evsaev (approved):
Well done 👏

PR #1673, opened 2021-12-07
[CIS-1364] Encapsulate reconnection logic

evsaev:
### 🔗 Issue Link

CIS-1275, CIS-1364

⚠️ Targets `origin/CIS-1275-offline-support` branch

### 🎯 Goal

Currently the logic related to automatic web-socket reconnection is spread around `WebSocketClient/ChatClient`.  The goal is to encapsulate this logic in a single place.

### 🛠 Implementation

Move to the connection recovery handler all logic related to:
- App `foreground <-> background` transitions
- the Internet `OFF <-> ON` transitions
- Reconnect retries when because web socket connection being flacky

### 🧪 Testing

1. Set `debug` log level
2. Run `DemoApp`
3. Filter output to show only logs from `ConnectionRecoveryHandler`
4. Run scenarios listed below
5. Validate SDK's behavior

### The chat client SHOULD reconnect automatically when:

Case 1:
- connect as any user
- internet -> OFF (see web-socket disconnects)
- internet -> ON (see web-socket reconnects)

Case 2:
- connect as any user
- app -> background (see web-socket disconnects)
- app -> foreground (see web-socket reconnects)

Case 3:
- connect as any user
- app -> background (see web-socket stays connected)
- wait 1-2 mins (see background task terminates, web-socket disconnects)
- app -> foreground (see web-socket reconnects)

Case 4:
- connect as any user (`staysActiveInBackroung == false`)
- app -> background (see web-socket disconnects)
- internet -> OFF (see disconnection not triggered for 2nd time)
- app -> foreground (see connection not triggered as there's no Internet)
- internet -> ON (see web-socket reconnects)

### The chat client SHOULD NOT reconnect automatically when:

Case 1:
- do not connect
- internet -> OFF (see disconnect is not triggered)
- internet -> ON (see connect is not triggered)

Case 2:
- do not connect
- app -> background (see disconnect is not triggered)
- app -> foreground (see connect is not triggered)

Case 3:
- connect as any user
- manually disconnect
- internet -> OFF (see disconnect is not triggered)
- internet -> ON (see connect is not triggered)

Case 4:
- connect as any user
- manually disconnect (see web-socket disconnects)
- app -> background (see disconnect is not triggered)
- app -> foreground (see connect is not triggered)

Case 4:
- connect as any user
- simulate client error coming via web-socket

### 🎨 Changes

N/A

### ☑️ Checklist


evsaev:
I found one more piece of reconnection logic inside `WebSocketClient` that reconnects automatically with increasing time intervals if the connection was lost. Sounds like smth that should also be moved to `ConnectionRecoveryHandler`, right? 🤔

Sources/StreamChat/APIClient/APIClient.swift
```diff
@@ -31,7 +31,7 @@ class APIClient {
     let decoder: RequestDecoder
     
     /// Used for reobtaining tokens when they expire and API client receives token expiration error
-    let tokenRefresher: (ClientError, @escaping () -> Void) -> Void
+    let tokenRefresher: (@escaping () -> Void) -> Void
```
evsaev, line 34: I would rather have a protocol for that so the API is self-explanatory. All token-related stuff from `APIClient` e.g. `isRefreshingToken/tokenRefreshConsecutiveFailures/maxTokenRefreshAttempts` should also be moved there IMHO.

evsaev: btw,`tokenRefreshConsecutiveFailures/maxTokenRefreshAttempts` is actually a retry strategy which is already encapsulated 🙂

Sources/StreamChat/Utils/InternetConnection/InternetConnection.swift
```diff
@@ -6,7 +6,11 @@ import Foundation
 import Network
 
 extension Notification.Name {
-    static let internetConnectionStatusDidChange = Notification.Name("io.getstream.StreamChat.internetConnectionStatus")
+    /// Posted when any the Internet connection update is detected (including quality updates).
+    static let internetConnectionStatusDidChange = Self("io.getstream.StreamChat.internetConnectionStatus")
+    
+    /// Posted only when the Internet connection availability is changed (excluding quality updates).
+    static let internetConnectionAvailabilityDidChange = Self("io.getstream.StreamChat.internetConnectionAvailability")
```
evsaev, line 13: In connection recovery handler we are not interested in connection quality updates (`.available(.great) -> .available(.constrained)`) but only in availability changes (`.available` -> `.unavailable`)

There was also an option to cache the previous Internet connection state in the handler but extending the `InternetConnection` API in a way it fits such needs as well felt reasonable 🤷

I would move away from the notification center approach and go with the cancellable subscriptions model but that's the refactoring the team needs to agree upon first

Sources/StreamChat/WebSocketClient/ConnectionStatus.swift
```diff
@@ -32,17 +32,20 @@ extension ConnectionStatus {
...
             self = .connected
             
         case .disconnecting:
             self = .disconnecting
+            
+        case let .disconnected(source):
+            let isWaitingForReconnect = webSocketConnectionState.isAutomaticReconnectionEnabled || source.serverError?
+                .isInvalidTokenError == true
+            
+            self = isWaitingForReconnect ? .connecting : .disconnected(error: source.serverError)
```
evsaev, line 48: That is not nice at all but didn't find any better way 😕 The thing is we don't have token refreshing logic encapsulated so there's no easy way to move it from `ChatClient` to `ConnectionRecoveryHandler`. That's why in case of an expired token `ConnectionRecoveryHandler ` does nothing (since `isAutomaticReconnectionEnabled` returns false) and just waits the chat client to refetch the token and initiate reconnection by itself.

Sources/StreamChat/ChatClient.swift
```diff
@@ -637,8 +555,8 @@ extension ChatClient: ConnectionStateDelegate {
             connectionId = id
         case let .disconnected(source):
             if let error = source.serverError,
-               error.isTokenExpiredError {
-                refreshToken(error: error, completion: nil)
+               error.isInvalidTokenError {
+                refreshToken(completion: nil)
```
evsaev, line 559: That is not nice 😕 

The thing is we don't have token refreshing logic encapsulated so there's no easy way to move it from `ChatClient` to `ConnectionRecoveryHandler`. That's why in case of expired token `ConnectionRecoveryHandler ` does nothing (since `isAutomaticReconnectionEnabled` returns false) and just waits for the chat client to refetch the token and initiate reconnection by itself.

Sources/StreamChat/ChatClient.swift
```diff
@@ -79,12 +79,14 @@ public class ChatClient {
...
-            { [weak self] error, completion in
+            { [weak self] completion in
                 guard let self = self else {
                     completion()
                     return
                 }
-                self.refreshToken(error: error, completion: { _ in completion() })
+                self.refreshToken(
+                    completion: { _ in completion() }
+                )
```
bielikb, line 89: @evsaev can't we just call `self.refreshToken(completion)` here? Do those 2 completions differ in their definition?

Sources/StreamChat/ChatClient.swift
```diff
@@ -379,8 +356,16 @@ public class ChatClient {
     func createBackgroundWorkers() {
         backgroundWorkers = workerBuilders.map { builder in
             builder(self.databaseContainer, self.apiClient)
-        } + eventWorkerBuilders.map { builder in
-            builder(self.databaseContainer, self.eventNotificationCenter, self.apiClient)
+        }
+        
+        if let webSocketClient = webSocketClient {
```
bielikb, line 361: `if let` usually decreases the clarity/adds complexity when reading the code. 

Could we use `guard let` instead?

Sources/StreamChat/WebSocketClient/RetryStrategy.swift
```diff
@@ -0,0 +1,65 @@
+//
+// Copyright © 2021 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+/// The type encapsulating the logic of computing delays for the failed actions that needs to be retried.
+protocol RetryStrategy {
+    /// Rerturns the # of consecutively failed retries.
+    var consecutiveFailures: Int { get }
```
bielikb, line 10:
```suggestion
    var consecutiveFailuresCount: Int { get }
```

Could we attach count/attempts suffix here for better readability? Thanks

Sources/StreamChat/WebSocketClient/RetryStrategy_Tests.swift
```diff
@@ -0,0 +1,119 @@
+//
+// Copyright © 2021 Stream.io Inc. All rights reserved.
+//
+
+@testable import StreamChat
+import XCTest
+
+class RetryStrategy_Tests: XCTestCase {
```
bielikb, line 8:
```suggestion
final class RetryStrategy_Tests: XCTestCase {
```

Sources/StreamChat/WebSocketClient/RetryStrategy_Tests.swift
```diff
@@ -0,0 +1,119 @@
...
+    var strategy: DefaultRetryStrategy!
+    
+    override func setUp() {
+        super.setUp()
+        strategy = DefaultRetryStrategy()
```
bielikb, line 13: Lets clear this in the tearDown to avoid any hanging references in the memory.

[Read more here](https://qualitycoding.org/xctestcase-teardown/)

Sources/StreamChat/ChatClient.swift
evsaev: Yep, thanks for the question! There was the reason to use `if`, actually 🙂 I prefer to use `guard` when further execution should not happen because of some condition not being met. In this case absence of `webSocketClient` does not mean we should return from `createBackgroundWorkers` because there can be other workers that do not rely on the web-socket but initialized after `connectionRecoveryHandler`

Sources/StreamChat/WebSocketClient/RetryStrategy.swift
evsaev: ✅ Updated

Sources/StreamChat/WebSocketClient/RetryStrategy_Tests.swift
evsaev: ✅ Updated

Sources/StreamChat/WebSocketClient/RetryStrategy_Tests.swift
evsaev: ✅ Updated

bielikb (approved):

PR #1681, opened 2021-12-10
Introduce `LogConfig.destinationTypes`

b-onc:
### 🎯 Goal

Our Logger is flexible and one can add new destinations easily, using `LogConfig.destinations`. But there's a limitation: adding a new destination using existing LogConfig parameters is hard. Consider the scenario, where you create a new destination:
```swift
class CoolLogDestination: BaseLogDestination {
    override func write(message: String) {
        // CoolLoggingSystem.log(message)
    }
}
```
and you want to use this destination. You do:
```swift
LogConfig.destinations = [CoolLogDestination()]
```
then the new destination gets initialized with default params of `BaseLogDestination`, which is not what we want, we want it to get initialized with `LogConfig` parameters, which we can customize.

The other way is to initialize it by passing all params:
```swift
LogConfig.destinations = [CoolLogDestination(identifier: "", showLevel: false, showDate: true.......)]
```
which is just too many params, and we already have defaults in `LogConfig`

### 🛠 Implementation

I introduce a new property in `LogConfig`, which is `destinationTypes`. This works like our `Components`: you assign types instead of objects, and LogConfig initializes the types by using its own parameters.

The default params in `BaseLogDestination.init` are removed since they can cause confusion.

```swift
LogConfig.destinationTypes = [CoolLogDestination.self]
```
then you're sure `LogConfig` params will be used.

`LogConfig.destinations` are not deprecated so users can use it too if they desire.

### ☑️ Checklist

### DIFF
--- a/Sources/StreamChat/Utils/Logger/Logger.swift
+++ b/Sources/StreamChat/Utils/Logger/Logger.swift
@@ -121,25 +121,37 @@ public enum LogConfig {
         }
     }
     
+    /// Destination types this logger will use.
+    ///
+    /// Logger will initialize the destinations with its own parameters. If you want full control on the parameters, use `destinations` directly,
+    /// where you can pass parameters to destination initializers yourself.
+    public static var destinationTypes: [LogDestination.Type] = [ConsoleLogDestination.self] {
+        didSet {
+            invalidateLogger()
+        }
+    }
+    
     /// Destinations for the default logger. Please see `LogDestination`.
     /// Defaults to only `ConsoleLogDestination`, which only prints the messages.
     ///
     /// - Important: Other options in `ChatClientConfig.Logging` will not take affect if this is changed.
     public static var destinations: [LogDestination] = {
-        let consoleLogDestination = ConsoleLogDestination(
-            level: level,
-            subsystems: subsystems,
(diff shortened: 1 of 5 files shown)


Sources/StreamChat/Utils/Logger/Logger.swift
```diff
@@ -121,25 +121,37 @@ public enum LogConfig {
         }
     }
     
+    /// Destination types this logger will use.
+    ///
+    /// Logger will initialize the destinations with its own parameters. If you want full control on the parameters, use `destinations` directly,
+    /// where you can pass parameters to destination initializers yourself.
+    public static var destinationTypes: [LogDestination.Type] = [ConsoleLogDestination.self] {
+        didSet {
+            invalidateLogger()
```
bielikb, line 130: shouldn't this be called in `willSet` closure instead?

bielikb (approved):

Sources/StreamChat/Utils/Logger/Logger.swift
bielikb: Based on our discussion outside of GitHub -> `invalidateLogger` will recreate the Logger with new destination type(s).

PR #1686, opened 2021-12-13
[CIS-1290] Add Static Message List Date Separators

nuno-vieira:
### 🔗 Issue Link
CIS-1290

### 🎯 Goal
Add support for static message list separators and make it easy to configure the date separators in the message list.

### 🛠 Implementation
- Refactors the current `ChatMessageListScrollOverlayView` by adding a composable `ChatMessageListDateSeparatorView` so that the date separator view can be composable between the overlay and the static separator.
- Makes the date separators configurable through `Components.default.messageListVC.isDateSeparatorEnabled` and `Components.default.messageListVC.isDateOverlayEnabled`
- Adds the `ChatMessageListDateSeparatorView` to the `ChatMessageCell` to display the grouped messages.

**Other possible approaches that failed:**
- **Sections:** Using table view sections was not possible because of the inverted table view.
- **Separators as part of the data source:** This was too complex to do and it would cause some minor breaking changes.
- **Add the separator view to the `ChatMessageContentView` instead of `ChatMessageCell`:** This would cause more issues because this view is used for the Actions popup view, which means that if we show an actions popup view with the date separator view, the date separator would also be shown on the popup view. So this was going to add extra complexity.

From all the options above, the current one is the simplest one without requiring major refactors.

### 🎨 Changes

https://user-images.githubusercontent.com/12814114/145842354-974cb5c9-3e61-4694-9143-7ad2ac3bd546.mp4

### 🧪 Testing

Go to a message list and scroll to the top and see that the separators are shown. This configuration is enabled on the demo app, but in order to not break other customers, the config is disabled by default. 

### ☑️ Checklist


DemoApp/DemoAppCoordinator.swift
```diff
@@ -99,6 +99,8 @@ final class DemoAppCoordinator: NSObject, UNUserNotificationCenterDelegate {
         Components.default.channelListRouter = DemoChatChannelListRouter.self
         Components.default.channelVC = CustomChannelVC.self
         Components.default.messageContentView = CustomMessageContentView.self
+        Components.default.messageListVC.isDateSeparatorEnabled = true
+        Components.default.messageListVC.isDateOverlayEnabled = true
```
b-onc, line 103: IMHO these shouldn't be in `Components`, it's not like the pattern we've been using. A property on the actual `MessageListVC` would suffice?

nuno-vieira: We also have a `Components.default.mentionAllAppUsers` which is also a flag. With this approach, at least the flag is set only in the context of the message list. And ofc putting multiple flags directly on the component's config will not scale very well. I did it in this way because we have been having feedback that requiring to subclassing just to turn on/off a flag is not very dev friendly.

nuno-vieira: One limitation of the current approach is that the config will be always the same for Channels and Threads, you can't provide a different config for each of them since the properties are static... So for that reason, it might make sense to switch to instance property like we are doing in other components, even tho its hard to change :/ We can also add a property directly to `Components` and also overridable by instance. With this, we get the best of both worlds. The idea here is to make it easier to provide these turn on/off these configs without subclassing, since multiple customers have provided feedback about this

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageCell.swift
```diff
@@ -27,13 +35,23 @@ public final class ChatMessageCell: _TableViewCell {
...
         
-        contentView.addSubview(messageContentView!)
-        
-        messageContentView!.pin(
+        containerStackView.pin(
             anchors: [.leading, .top, .trailing, .bottom],
             to: contentView
         )
+
+        messageContentView!.pin(
```
polqf, line 51: Can we do optional chaining here, please? 😄

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageCell.swift
```diff
@@ -27,13 +35,23 @@ public final class ChatMessageCell: _TableViewCell {
     
     override public func setUpLayout() {
         super.setUpLayout()
+
+        containerStackView.axis = .vertical
+        containerStackView.alignment = .center
+        containerStackView.spacing = 8
+        containerStackView.addArrangedSubview(dateSeparatorView)
+        containerStackView.addArrangedSubview(messageContentView!)
```
polqf, line 43: Any way we can avoid the force unwrap?

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -310,6 +318,18 @@ open class ChatMessageListVC:
 
         cell.messageContentView?.delegate = self
         cell.messageContentView?.content = message
+        cell.dateSeparatorView.isHidden = true
+
+        let previousIndexPath = IndexPath(row: indexPath.row + 1, section: indexPath.section)
+        if let previousMessage = dataSource?.chatMessageListVC(self, messageAt: previousIndexPath),
+           let currentMessage = message,
+           Self.isDateSeparatorEnabled {
```
polqf, line 326: Self.isDateSeparatorEnabled should be the first check, so that if it is false, you don't access the data source

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -310,6 +318,18 @@ open class ChatMessageListVC:
 
         cell.messageContentView?.delegate = self
         cell.messageContentView?.content = message
+        cell.dateSeparatorView.isHidden = true
+
+        let previousIndexPath = IndexPath(row: indexPath.row + 1, section: indexPath.section)
+        if let previousMessage = dataSource?.chatMessageListVC(self, messageAt: previousIndexPath),
```
polqf, line 324: If there is no previous message, you should be showing the dataSeparator. Shouldn't you?

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
nuno-vieira: Yes that is true 👍 I did It in this way because it looks cleaner 😃 And I thought that the efficiency here in this case might not be a big deal. But I'll change it anyway 👍

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageCell.swift
nuno-vieira: Yes sure, I just moved code in this case, but I also prefer to change it 👍

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
nuno-vieira: I thought I shouldn't, but actually it makes sense to do it. Especially because our overlay disappears. Nice catch! Will change this 👍

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageCell.swift
nuno-vieira: Actually, I won't change this. The force unwrapping, in this case, is important to make sure we don't forget to set the message content view. So in order for this to crash, it is a developer mistake and it should be caught as early as possible.

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
nuno-vieira: ✅

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
nuno-vieira: ✅

evsaev (approved):
🚢

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageCell.swift
```diff
@@ -27,13 +37,23 @@ public final class ChatMessageCell: _TableViewCell {
     
     override public func setUpLayout() {
         super.setUpLayout()
+
+        containerStackView.axis = .vertical
+        containerStackView.alignment = .center
```
evsaev, line 42: do you think we should set `messageContentView.width == contentView.width` constraint to make sure it fills all the space?

evsaev: Sorry, found the place where we set leading/trailing constraints 👍 I would actually go with width constraint as it's a more explicit way of doing this. But both ways work so leaving it's up to you 🙂

evsaev (dismissed):
LGTM

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageCell.swift
nuno-vieira: The width constraint might have issues with RTL Languages? I think I prefer the current approach, it is probably safer 👍

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageCell.swift
polqf: I tend to disagree with the "it is a developer mistake" statement. We can make mistakes, and we can even miss them in a release, using force unwraps does not provide safety to our libraries.

DemoApp/DemoAppCoordinator.swift
nuno-vieira: @b-onc I've updated the approach to be more flexible. Both approaches are now supported.

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageCell.swift
nuno-vieira: @polqf Updated ✅

polqf (dismissed):

adamrushy (approved):
LGTM 🚀

adamrushy (approved):

PR #1687, opened 2021-12-13
Fix redundant API call in ChannelListController

b-onc:
### 🎯 Goal

Once, we had a trumping channel logic, which was a workaround for outdated data. It was problematic so we disabled it (by commenting out the code) but we didn't remove the whole fix, so the SDK is making 1 extra API call because of this.

### 🛠 Implementation

Tthe connection observer and the extra API call is removed from `ChannelListController`

Additionally, I found some broken tests when I ran the tests locally, so I fixed them in a separate commit. I don't know how CI was green on #1642 , perhaps CI can't import the other SDKs but local builds can?

### 🧪 Testing

- Launch the app with a proxy
- Observe the API calls.
There should be 2 `queryChannels` call, 1 is the expected one, the other will be resolved by offline storage PRs.

### ☑️ Checklist


Sources/StreamChat/Workers/ChannelListUpdater.swift
```diff
@@ -25,10 +24,6 @@ class ChannelListUpdater: Worker {
                 switch result {
                 case let .success(channelListPayload):
                     self?.database.write { session in
-//                        if trumpExistingChannels {
-//                            try session.deleteChannels(query: channelListQuery)
-//                        }
-
```
bielikb, line 31: Can this chunk be removed?

Sources/StreamChat/ChatClient_Tests.swift
```diff
@@ -1340,10 +1349,20 @@ extension ChatClient_Tests {
...
         
+        var expectedIdentifier: String {
+            #if canImport(StreamChatSwiftUI)
+            "swiftui"
+            #elseif canImport(StreamChatUI)
+            "uikit"
+            #else
+            "swift"
+            #endif
+        }
```
bielikb, line 1360: Are we explicitly hardcoding the behaviour within the test target? I thought the intention of these tests was to verify the correct value is propagated into the header field.

If we need to keep this, let's turn this into a computed var of this test case to avoid code duplication.

Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift
```diff
@@ -129,51 +128,14 @@ public class ChatChannelListController: DataController, DelegateCallable, DataSt
     
     override public func synchronize(_ completion: ((_ error: Error?) -> Void)? = nil) {
         startChannelListObserverIfNeeded()
-        setupEventObserversIfNeeded(completion: completion)
-    }
```
bielikb, line 133: Is this logic covered by tests? It would be great to capture scenario like:
Given the list controller is created & synchronize is called, only 1 API call is made.

b-onc: Yes, in `ChannelListController_Tests.test_synchronize_callsChannelQueryUpdater`

Sources/StreamChat/ChatClient_Tests.swift
evsaev: References to `StreamChatUI/SwiftUI` look a bit weird in test target for LLC. tbh. Isn't it the matter of clearing the cache by removing derived data?

b-onc: Not really. Even if it did, we can't afford to clear cache every time we want to run tests. Either we remove this piece of tests or we hardcode the behavior, or convert the check so it `startsWith("stream-swift")`

evsaev: Ok, I see your point. Then let's address the comment from Boris about code duplication 👍

bielikb (approved):
LGTM!

Sources/StreamChat/ChatClient_Tests.swift
bielikb: Ill keep track of this change. We will need to think the strategy through

PR #1691, opened 2021-12-15
Update references to use develop instead of main whenever needed

polqf:
### 🔗 Issue Link
We're moving to `develop` to be one step closer to Git-Flow 🚀 

### ☑️ Checklist

### DIFF
--- a/.github/workflows/macos-checks.yml
+++ b/.github/workflows/macos-checks.yml
@@ -46,7 +46,7 @@ jobs:
         key: ${{ runner.os }}-mint-${{ hashFiles('./Mintfile') }}
         restore-keys: ${{ runner.os }}-mint-
     - name: Run bootstrap.sh
-      run: ./bootstrap.sh
+      run: ./Scripts/bootstrap.sh
     - name: Run Tests (Release)
       env:
         MATCH_PASSWORD: ${{ secrets.MATCH_PASSWORD }}
@@ -86,7 +86,7 @@ jobs:
         key: ${{ runner.os }}-mint-${{ hashFiles('./Mintfile') }}
         restore-keys: ${{ runner.os }}-mint-
     - name: Run bootstrap.sh
-      run: ./bootstrap.sh
+      run: ./Scripts/bootstrap.sh
     - name: Run Tests (Release)
       env:
         MATCH_PASSWORD: ${{ secrets.MATCH_PASSWORD }}
--- a/.github/workflows/macos-checks.yml
+++ b/.github/workflows/macos-checks.yml
@@ -4,9 +4,11 @@ on:
   push:
     branches:
       - main
+      - develop
   pull_request:
     branches:
       - main
+      - develop
   release:
     types:
       - created
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -31,9 +31,6 @@ end
 
 desc "Start a new release"
 lane :release do |options|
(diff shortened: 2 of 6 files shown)


.github/workflows/macos-checks.yml
```diff
@@ -44,7 +46,7 @@ jobs:
         key: ${{ runner.os }}-mint-${{ hashFiles('./Mintfile') }}
         restore-keys: ${{ runner.os }}-mint-
     - name: Run bootstrap.sh
-      run: ./bootstrap.sh
+      run: ./Scripts/bootstrap.sh
```
polqf, line 49: We did not catch that before because this workflow was disabled.
Because I updated the yml, this was automatically enabled again, and I spot this issue.

It is disabled again now

fastlane/Fastfile
```diff
@@ -31,8 +31,8 @@ end
 
 desc "Start a new release"
 lane :release do |options|
-  # Ensure we are on the main branch
-  ensure_git_branch(branch: 'main')
+  # Ensure we are on the develop branch
+  ensure_git_branch(branch: 'develop')
```
nuno-vieira, line 35: Do we have a hotfix lane? 🤔 What if we want to make a release from a hotfix branch? Or this was not yet done in the script? Maybe we could also provide a "no_ensure_branch" like we did with the git_clean 🤔 

This is just a question, I'll approve it since this is out of scope of this PR 👍

nuno-vieira (dismissed):
LGTM

fastlane/Fastfile
polqf: Nice one! I think we should maybe check if it is develop OR main

fastlane/Fastfile
```diff
@@ -31,8 +31,8 @@ end
 
 desc "Start a new release"
 lane :release do |options|
-  # Ensure we are on the main branch
-  ensure_git_branch(branch: 'main')
+  # Ensure we are on `develop` or `main`
+  ensure_git_branch(branch: '^main$|^develop$')
```
nuno-vieira, line 35: I'm not sure this would work for hotfixes tho 🤔 Usually when we create a hotfix it is going to be from a tag, and not from the "main" branch. So I'm not sure if this "ensure_git_branch" is even worth being here 🤔 Actually probably an "ensure_is_not_main" is actually more useful since I don't see a case where we want to release from main. And this will avoid we make a mistake because of muscle memory 😄 (Since we have been releasing from main until now)

polqf: Oh, you are right... Let me remove this, because it might not be a good check to have

nuno-vieira (approved):
LGTM! 🚀

fastlane/Fastfile
nuno-vieira: @polqf I actually spotted one more thing 🍎 https://github.com/GetStream/stream-chat-swift/blob/53d453f732db50e9b7598e70a32c82d8a9852674/fastlane/Fastfile#L101

Do we still need that "base: main"? 🤔

polqf: Yes, because releases are merged to main

nuno-vieira: Ahhh yes! Sorry totally misread that. Though the PR was being created from "main" xD

nuno-vieira (approved):

nuno-vieira (approved):

PR #1692, opened 2021-12-15
[CIS-1378] Safely access indexes of collections

polqf:
### 🔗 Issue Link
https://stream-io.atlassian.net/browse/CIS-1378

### 🎯 Goal

Because of the nature of the core of our chat message list, we sometimes have crashes because we are accessing indexes of messages that are not present anymore. The purpose of this PR is not to fix the core nature of this edge cases, but instead make sure that we don't fail when this happens.
It also introduces a custom assertion, together with a log whenever this happens.

### 🛠 Implementation

A new extension over Collections will allow safe subscripts by checking whether the index is within the bounds of the collection. If it is not there, it would return nil, instead of crashing.

### ☑️ Checklist

### DIFF
--- a/Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
+++ b/Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
@@ -142,7 +142,7 @@ open class ChatChannelVC:
 
     open func chatMessageListVC(_ vc: ChatMessageListVC, messageAt indexPath: IndexPath) -> ChatMessage? {
         guard indexPath.item < channelController.messages.count else { return nil }
-        return channelController.messages[indexPath.item]
+        return channelController.messages[safe: indexPath.item]
     }
 
     open func chatMessageListVC(
--- a/Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
+++ b/Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
@@ -141,7 +141,7 @@ open class ChatChannelVC:
     }
 
     open func chatMessageListVC(_ vc: ChatMessageListVC, messageAt indexPath: IndexPath) -> ChatMessage? {
-        guard indexPath.item < channelController.messages.count else { return nil }
+        channelController.messages.assertIndexIsPresent(indexPath.item)
         return channelController.messages[safe: indexPath.item]
     }
 
--- a/Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
+++ b/Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
@@ -168,12 +168,11 @@ open class ChatChannelListVC: _ViewController,
             withReuseIdentifier: collectionViewCellReuseIdentifier,
             for: indexPath
         ) as! ChatChannelListCollectionViewCell
-    
+
+        guard let channel = getChannel(at: indexPath) else { return UICollectionViewCell() }
+
         cell.components = components
-        cell.itemView.content = .init(
-            channel: controller.channels[indexPath.row],
-            currentUserId: controller.client.currentUserId
-        )
+        cell.itemView.content = .init(channel: channel, currentUserId: controller.client.currentUserId)
 
         cell.swipeableView.delegate = self
         cell.swipeableView.indexPath = { [weak cell, weak self] in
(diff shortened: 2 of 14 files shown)


Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
```diff
@@ -142,7 +142,7 @@ open class ChatChannelVC:
 
     open func chatMessageListVC(_ vc: ChatMessageListVC, messageAt indexPath: IndexPath) -> ChatMessage? {
         guard indexPath.item < channelController.messages.count else { return nil }
-        return channelController.messages[indexPath.item]
+        return channelController.messages[safe: indexPath.item]
```
nuno-vieira, line 145: You can remove the line above since in this case, we were actually protecting against an invalid index 👍

nuno-vieira (dismissed):
Just add a minor comment, other than that LGTM ✅

Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
```diff
@@ -168,12 +168,11 @@ open class ChatChannelListVC: _ViewController,
             withReuseIdentifier: collectionViewCellReuseIdentifier,
             for: indexPath
         ) as! ChatChannelListCollectionViewCell
-    
+
+        guard let channel = getChannel(at: indexPath) else { return UICollectionViewCell() }
```
bielikb, line 172: Returning empty cell like this will crash the app too (due to unregistered cell identifier).
Could you log this case?
Thanks

Sources/StreamChatUI/Gallery/GalleryVC.swift
```diff
@@ -341,8 +341,10 @@ open class GalleryVC:
             withReuseIdentifier: reuseIdentifier,
             for: indexPath
         ) as! GalleryCollectionViewCell
-        
-        cell.content = items[indexPath.item]
+
+        guard let item = getItem(at: indexPath) else { return UICollectionViewCell() }
```
bielikb, line 345: Lets assert this place, too.

Sources/StreamChatUI/Utils/Array+SafeSubscript.swift
```diff
@@ -0,0 +1,37 @@
...
+//
+
+import Foundation
+import StreamChat
+
+extension Collection {
+
+    /// Returns the element at the specified index if it is within bounds, otherwise nil.
+    subscript (safe index: Index) -> Element? {
+        return indices.contains(index) ? self[index] : nil
```
bielikb, line 16: no need for returns in single line expressions

```suggestion
        indices.contains(index) ? self[index] : nil
```

Sources/StreamChatUI/Utils/Array+SafeSubscript.swift
```diff
@@ -0,0 +1,37 @@
...
+//  Copyright © 2021 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+import StreamChat
+
+extension Collection {
+
+    /// Returns the element at the specified index if it is within bounds, otherwise nil.
+    subscript (safe index: Index) -> Element? {
```
bielikb, line 15: @polqf could you pls provide simple unit test coverage (happy/nonhapoy path) for this piece of logic? Thanks!

Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
polqf: I wrapped the logging & assertion logic in getChannel

Sources/StreamChatUI/Gallery/GalleryVC.swift
polqf: I wrapped the logging & assertion logic in getItem

Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
bielikb: @polqf furtherly, the `return` can be removed from the single line expression.

bielikb (approved):

Sources/StreamChatUI/Utils/Array+SafeSubscript_Tests.swift
```diff
@@ -0,0 +1,26 @@
...
+        let collection = (0..<3).map { "index\($0)" }
+        XCTAssertEqual(collection[safe: 0], "index0")
+        XCTAssertEqual(collection[safe: 1], "index1")
+        XCTAssertEqual(collection[safe: 2], "index2")
+    }
+
+    func test_safeSubscript_returnsNil_whenIndexNotPresent() {
+        let collection = (0..<3).map { "index\($0)" }
+        XCTAssertNil(collection[safe: 3])
+        XCTAssertNil(collection[safe: 84])
```
adamrushy, line 24: 😍

adamrushy (approved):
Looks great ✅  do we want a CHANGELOG entry for this?

PR #1696, opened 2021-12-16
Expose `QueryOptions` and `start/stopWatching`

b-onc:
### 🎯 Goal

Allowing users to explicitly start/stop watching a channel, and make Channel queries without `watch`ing.

### 🛠 Implementation

The functionality were there but hidden, now exposed.

### 🧪 Testing

Already tested in `ChannelController_Tests`

### ☑️ Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,8 @@ _December 15, 2021_
 - Added `LogConfig.destinationTypes` for ease of adding new destinations to logger [#1681](https://github.com/GetStream/stream-chat-swift/issues/1681)
 - Expose container embedding top & bottom containers by `ChatChannelListItemView` [#1670](https://github.com/GetStream/stream-chat-swift/issues/1670)
 - Add Static Message List Date Separators [#1686](https://github.com/GetStream/stream-chat-swift/issues/1686) (You can read this [doc](https://getstream.io/chat/docs/sdk/ios/uikit/components/message/#date-separators) to understand how to configure this feature)
+- `ChannelQuery.options` and `ChannelListQuery.options` are now public and mutable [#1696](https://github.com/GetStream/stream-chat-swift/issues/1696)
+- `ChannelController.startWatching` and `stopWatching` are now `public`. You can explicitly stop watching a channel [#1696](https://github.com/GetStream/stream-chat-swift/issues/1696)
 
 # [4.5.2](https://github.com/GetStream/stream-chat-swift/releases/tag/4.5.2)
 _December 10, 2021_
--- a/Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
+++ b/Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
@@ -1135,11 +1135,9 @@ public extension ChatChannelController {
     ///
     /// Please check [documentation](https://getstream.io/chat/docs/android/watch_channel/?language=swift) for more information.
     ///
-    /// We keep these functions internal since we're not sure how we should interface this behavior.
-    /// If you have suggestions, please open a ticket or send us an email at support@getstream.io
     ///
     /// - Parameter completion: Called when the API call is finished. Called with `Error` if the remote update fails.
-    internal func startWatching(completion: ((Error?) -> Void)? = nil) {
+    func startWatching(completion: ((Error?) -> Void)? = nil) {
         /// Perform action only if channel is already created on backend side and have a valid `cid`.
         guard let cid = cid, isChannelAlreadyCreated else {
             channelModificationFailed(completion)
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### 🔄 Changed
 - Change `ChatMessageLayoutOptions` to a `Set` instead of an `OptionSet` for a more flexible and safer customization [#1651](https://github.com/GetStream/stream-chat-swift/issues/1651)
 - There is a new `ChatMessageListDateSeparatorView` component that should be used instead of the `ChatMessageListScrollOverlayView` if the goal is customize the styling of the date separator. Read [here](https://getstream.io/chat/docs/sdk/ios/uikit/components/message/#date-separators) for more details.
+- `ChannelQuery.options` and `ChannelListQuery.options` are now public and mutable [#1696](https://github.com/GetStream/stream-chat-swift/issues/1696)
+- `ChannelController.startWatching` and `stopWatching` are now `public`. You can explicitly stop watching a channel [#1696](https://github.com/GetStream/stream-chat-swift/issues/1696)
```
evsaev, line 13: nit: feels more like an addition than the change

Sources/StreamChat/Query/QueryOptions.swift
```diff
@@ -5,32 +5,32 @@
 import Foundation
 
 /// Query options.
-struct QueryOptions: OptionSet, Encodable {
+public struct QueryOptions: OptionSet, Encodable {
```
evsaev, line 8: @b-onc why are we making it public? what would be the use case?

b-onc: There are more than 1 case, integrators currently can't pass their own query options, which is something allowed by the API.
1. Integrators can't query with `presence` option, which is something required for getting presence events.
2. Integrators can't query without `watch` option, for cases when they only want the data and not WS events.

evsaev: @bahadir ok, didn’t know it’s allowed to pass custom query options 🙂 the thing is with the current shape of offline support allowing to query channels without the `watch` option/allowing to stop watching a channel is dangerous. The local state will go out of sync much more often which will cause another spike of conflicting updates on both channel/message list

b-onc: @evsaev that's exactly why we didn't made them `public` firstly but now customers require it

bielikb: @b-onc Could we align this to the same approach that we applied with `layout options`? We basically get rid of the `OptionSet`. Since we're turning the `struct` public, we should make sure the way how you can extend the options is consistent.

b-onc: You can't send custom options, options are well-defined in the API: https://getstream.io/chat/docs/rest/#channels-getorcreatechannel_type_id
`watch`,`state` and `presence` is allowed

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
```diff
@@ -1135,11 +1135,9 @@ public extension ChatChannelController {
     ///
     /// Please check [documentation](https://getstream.io/chat/docs/android/watch_channel/?language=swift) for more information.
     ///
-    /// We keep these functions internal since we're not sure how we should interface this behavior.
-    /// If you have suggestions, please open a ticket or send us an email at support@getstream.io
     ///
     /// - Parameter completion: Called when the API call is finished. Called with `Error` if the remote update fails.
-    internal func startWatching(completion: ((Error?) -> Void)? = nil) {
+    func startWatching(completion: ((Error?) -> Void)? = nil) {
```
polqf, line 1140: Shouldn't these be public?

Sources/StreamChat/Query/QueryOptions.swift
evsaev: In this case `QueryOptions.init` should stay internal, right?

b-onc: correct, good catch!

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
b-onc: the extension where these funcs are defined is already public

Sources/StreamChat/Query/QueryOptions.swift
b-onc: @evsaev oh, it's public because of the `RawRepresentable` conformance... Alright, we should get rid of `OptionSet` as we did with `LayoutOptions`

b-onc: @evsaev @bielikb I've worked on this but i think resulting API is not better than what we have now. Apple also exposes inits for their OptionSet s but that doesn't mean developers should use them. even if they did, it won't work.
in our case, only allowed values will be encoded to queries, the other will be discarded. so if one used the QueryOptions.init(rawValue: 10000) it would just be a no-op, as it should be.
hiding that init requires a lot more "workarounds" while having the OptionSet API, I think we should not touch it. it's fundamentally different than LayoutOptions in which, no customization is allowed, QueryOptions is just communication API between backend, not between components inside the SDK

evsaev (dismissed):
🚢

Sources/StreamChat/Query/QueryOptions.swift
bielikb: Fair enough @b-onc !

bielikb (approved):

PR #1697, opened 2021-12-16
Fix reload rows out of bounds

tbarbugli:
This PR fixes an out of bound problem where we would reload a previous row that is not present in the message list

You could reproduce this on a channel with 26 messages 🙈


polqf (approved):

nuno-vieira (approved):
LGTM! 🚀

adamrushy (approved):
Awesome

PR #1703, opened 2021-12-17
[CIS-1427] Fix quoted messages not updated after edit

nuno-vieira:
### 🔗 Issue Link
CIS-1427

### 🎯 Goal
Fix quoted messages not getting updated if that quoted message was updated.

### 🛠 Implementation
Whenever a cell is updated on the message list view, we check if the updated message is currently from a quoted message in the visible cells, if it is we updated those cells.

We only need to handle the visible cells since when scrolling the message list, the cells that are not visible will be updated automatically.

### 🎨 Changes

https://user-images.githubusercontent.com/12814114/146598964-1e841626-40ec-4cf2-90f6-9c49c517e1d0.mp4

### 🧪 Testing
- Create a message A
- Create message B quoting message A
- Edit Message A
- Quoted message A in message B should be updated

### ☑️ Checklist


Sources/StreamChatUI/ChatMessageList/ChatMessageListView.swift
```diff
@@ -259,6 +262,28 @@ open class ChatMessageListView: UITableView, Customizable, ComponentsProvider {
         }
     }
 
+    private func getVisibleQuotedMessageUpdates(forUpdatedMessageAt indexPath: IndexPath) -> [IndexPath]? {
+        let visibleCells = getVisibleCells()
+        guard let visibleUpdatedMessage = visibleCells[indexPath]?.messageContentView?.content else {
```
bielikb, line 267: Please use safe index :)

nuno-vieira: No need for safe index, its  actually a dictionary 😋

evsaev (commented):
@nuno-vieira why don't we handle this on the database level?

Sources/StreamChatUI/ChatMessageList/ChatMessageListView.swift
```diff
@@ -259,6 +262,28 @@ open class ChatMessageListView: UITableView, Customizable, ComponentsProvider {
...
+            return nil
+        }
+
+        var quotedMessageUpdates: [IndexPath] = []
+
+        for (index, messageCell) in visibleCells {
+            let quotedMessageId = messageCell.messageContentView?.content?.quotedMessage?.id
+            let visibleUpdatedMessageId = visibleUpdatedMessage.id
```
b-onc, line 275: You don't need to get `id` everytime, you can change the `guard` so it gets `visibleUpdatedMessageId` from the get go

Sources/StreamChatUI/ChatMessageList/ChatMessageListView.swift
```diff
@@ -213,6 +213,9 @@ open class ChatMessageListView: UITableView, Customizable, ComponentsProvider {
 
             case let .update(_, index: index):
                 self.reloadRows(at: [index], with: .automatic)
+                if let quotedMessageUpdates = getVisibleQuotedMessageUpdates(forUpdatedMessageAt: index) {
+                    self.reloadRows(at: quotedMessageUpdates, with: .automatic)
+                }
```
b-onc, line 218:
```swift
if let quotedMessageUpdates = ... {
  self.reloadRows(at: [index] + quotedMessageUpdates, with: ...)
} else {
  self.reloadRows(at: [index]....)
}
```
or something similar to reduce the `reloadRows` calls to 1

Sources/StreamChatUI/ChatMessageList/ChatMessageListView.swift
```diff
@@ -259,6 +262,28 @@ open class ChatMessageListView: UITableView, Customizable, ComponentsProvider {
         }
     }
 
+    private func getVisibleQuotedMessageUpdates(forUpdatedMessageAt indexPath: IndexPath) -> [IndexPath]? {
```
b-onc, line 265: Docs on what this function does would be nice, it's not straightforward from the name

nuno-vieira:
> @nuno-vieira why don't we handle this on the database level?

This would mean that we need to observe the quotedMessageId right? I think the current approach is more simple and more efficient since the db approach will trigger an update for all quoted of messages, and we are only interested in the visible ones

evsaev:
> > @nuno-vieira why don't we handle this on the database level?
> 
> This would mean that we need to observe the quotedMessageId right? I think the current approach is more simple and more efficient since the db approach will trigger an update for all quoted messages, and we are only interested in the visible ones

Yeah, whenever we get an event with a message payload we can check if this message has replies saved locally (`replies: Set<MessageDTO>`) and trigger the update for those where `showReplyInChannel == true`. I see your point about reloading only those replies that are currently visible but catching this change on the data source level feels safer and also much easier to implement and test

b-onc:
@evsaev and it'd be easier to turn off? it'd be nice if this was a toggle -> integrators were able to keep the quoted messages in their original state. after all, UX-wise, a person is not quoting the "message", they're quoting what the message "says" at that time

evsaev:
> @evsaev and it'd be easier to turn off? it'd be nice if this was a toggle -> integrators were able to keep the quoted messages in their original state. after all, UX-wise, a person is not quoting the "message", they're quoting what the message "says" at that time

@b-onc i would say that should be handled via a separate toggle on the dashboard which disables editing of quoted message on API level. The only way we can keep the quoted message in its original state is to ignore events that change it but this way local data goes out of sync with remote which is dangerous

Sources/StreamChatUI/ChatMessageList/ChatMessageListView.swift
```diff
@@ -259,6 +262,28 @@ open class ChatMessageListView: UITableView, Customizable, ComponentsProvider {
...
+            if quotedMessageId == visibleUpdatedMessageId {
+                quotedMessageUpdates.append(index)
+            }
+        }
+
+        guard !quotedMessageUpdates.isEmpty else { return nil }
+
+        return quotedMessageUpdates
+    }
+
```
polqf, line 286: The more logic we add on the views, the more I see the need of view models... 👀

polqf (dismissed):

Sources/StreamChat/Database/DatabaseSession.swift
```diff
@@ -400,11 +400,17 @@ extension DatabaseSession {
...
             payload: messagePayload,
             channelDTO: channelDTO,
             syncOwnReactions: false
         )
+
+        if payload.eventType == .messageUpdated {
+            savedMessage.quotedBy.forEach { message in
+                message.updatedAt = savedMessage.updatedAt
+            }
+        }
```
nuno-vieira, line 413: @evsaev @b-onc This is the only solution right now I've found to do this on DB Level. The issue here is that the update is triggered ~1s after, and this causes the table view to jump twice (because of the nature of reloading rows in table view)

evsaev: @nuno the trick with `id` should work, actually, we do a similar thing with channel reads 🤔

```
override func willSave() {
        super.willSave()
        
        // When the read is updated, we need to propagate this change up to holding channel
        if hasPersistentChangedValues, !channel.hasChanges {
            // this will not change object, but mark it as dirty, triggering updates
            channel.cid = channel.cid
        }
    }
```

evsaev: > update is triggered ~1s
Would you record a video? Are you saying the parent message update is propagated immediately while quote updates are delayed?

nuno-vieira: https://user-images.githubusercontent.com/12814114/146810049-19b6dac6-6374-43db-b8f0-68e6b4afadba.mp4

@evsaev ☝️

evsaev: Ok, I see 😕 But customers are waiting for the fix, and handling it on DB level is definitely the right way. So even though it feels broken, we'd better jump than keep quote replies outdated. But it also means glitches on the message list become more visible/easy -> the sooner we give love to message list the better

evsaev (approved):
LGTM 👍 

Customers are waiting for the fix, and handling it on DB level is definitely the right way. So even though it feels broken, we'd better jump than keep quote replies outdated. But it also means glitches on the message list become more visible/easy to reproduce -> the sooner we give love to the message list the better.

evsaev (commented):
Sorry, noticed that a unit test and a changelog entry are missing. Let's add those and I'll approve 👍

bielikb (approved):
Works well!

PR #1706, opened 2021-12-18
[CIS-1367] Remove unnecessary API calls from `ChannelVC` and `ChannelListVC `

tbarbugli:
### 🔗 Issue Link

CIS-1367

### 🎯 Goal

Fix sending multiple unnecessary pagination requests when `ChannelListVC` and `ChannelVC` are shown.
This PR makes sure that only 1 API call is made to load both `ChannelListVC` and `ChannelVC`

### 🎨 Changes

Before this PR `ChannelListVC` was calling the Query Channels endpoint 3 times!

**ChannelListVC**
- Pagination was incorrectly firing the request to load another page 
- The `make` method was calling the `setUp` method (which calls `synchronize`)

**ChannelVC**
- Pagination was incorrectly firing the request to load another page 

### ☑️ Checklist

### DIFF
--- a/Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
+++ b/Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
@@ -17,7 +17,7 @@ open class ChatChannelListVC: _ViewController,
     /// The `ChatChannelListController` instance that provides channels data.
     public var controller: ChatChannelListController!
 
-    private var loadingPreviousMessages: Bool = false
+    private var isPaginatingChannels: Bool = false
 
     open private(set) lazy var loadingIndicator: UIActivityIndicatorView = {
         if #available(iOS 13.0, *) {
@@ -85,8 +85,7 @@ open class ChatChannelListVC: _ViewController,
         
         // Set the Controller on the ViewController
         chatChannelListVC.controller = controller
-        chatChannelListVC.setUp()
-        
+
         // Return the newly created ChatChannelListVC
         return chatChannelListVC
     }
@@ -121,6 +120,10 @@ open class ChatChannelListVC: _ViewController,
         forItemAt indexPath: IndexPath
     ) {
         if controller.state != .remoteDataFetched {
+            return
+        }
+
+        guard collectionView.isTrackingOrDecelerating else {
             return
         }
 
@@ -221,14 +224,14 @@ open class ChatChannelListVC: _ViewController,
     }
 
     open func loadMoreChannels() {
-        guard !loadingPreviousMessages else {
+        guard !isPaginatingChannels else {
             return
         }
-        loadingPreviousMessages = true
+        isPaginatingChannels = true
 
-        controller.loadNextChannels(completion: { [weak self] _ in
-            self?.loadingPreviousMessages = false
-        })
+        controller.loadNextChannels { [weak self] _ in
+            self?.isPaginatingChannels = false
+        }
     }
 
     open func swipeableViewWillShowActionViews(for indexPath: IndexPath) {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 # Upcoming
 
-### 🔄 Changed
+### 🐞 Fixed
+- Fix multiple pagination requests being fired from `ChatChannelVC` and `ChatChannelListVC` [#1706](https://github.com/GetStream/stream-chat-swift/issues/1706)
 
 # [4.6.0](https://github.com/GetStream/stream-chat-swift/releases/tag/4.6.0)
 _December 20, 2021_
--- a/Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
+++ b/Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
@@ -64,7 +64,7 @@ open class ChatChannelVC:
         messageListVC.listView.isLastCellFullyVisible
     }
 
-    private var loadingPreviousMessages: Bool = false
+    private var isLoadingPreviousMessages: Bool = false
 
     override open func setUp() {
         super.setUp()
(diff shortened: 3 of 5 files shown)


Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
```diff
@@ -17,7 +17,7 @@ open class ChatChannelListVC: _ViewController,
     /// The `ChatChannelListController` instance that provides channels data.
     public var controller: ChatChannelListController!
 
-    private var loadingPreviousMessages: Bool = false
+    private var paginatingChannels: Bool = false
```
evsaev, line 20: nit: it's preffered to name boolean variables starting with `is`
```suggestion
    private var isLoadingMoreChannels: Bool = false
```

evsaev: the same applies to channel vc: `loadingPreviousMessages -> isLoadingPreviousMessages`

Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
```diff
@@ -85,8 +85,7 @@ open class ChatChannelListVC: _ViewController,
         
         // Set the Controller on the ViewController
         chatChannelListVC.controller = controller
-        chatChannelListVC.setUp()
```
evsaev, line 88: This seems legit, the `setUp` get invoked automatically as part of the `ChatChannelListVC's` view lifecycle 🤔

@adamrushy can we make this change or there's a reason to have it?

evsaev (approved):
Looks good, just some small comments regarding naming 👍 Let's wait for the green light from Adam and we will be good to 🚢 Do we need a changelog entry for this, btw?

Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
adamrushy: Yeah, I need to check this... I specifically added because there was a bug with delegates not being assigned if I remember rightly 🤔

Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
```diff
@@ -124,6 +123,10 @@ open class ChatChannelListVC: _ViewController,
             return
         }
 
+        guard collectionView.isTracking || collectionView.isDecelerating else {
```
polqf, line 126: Seems like this same check is duplicated in ChatChannelVC, what about creating an extension over UICollectionView?

tbarbugli: one VC uses tableview and the other collectionview

polqf: Both are UIScrollView under the hood

bielikb:
@tbarbugli please point this PR against the release candidate `release/4.6.0` + reflect this fix in the `Changelog`.

Examples/SlackClone/SlackMessageOptionsResolver.swift
```diff
@@ -14,7 +14,9 @@ final class SlackMessageOptionsResolver: ChatMessageLayoutOptionsResolver {
         appearance: Appearance
     ) -> ChatMessageLayoutOptions {
         var options = super.optionsForMessage(at: indexPath, in: channel, with: messages, appearance: appearance)
-        options.remove([.flipped, .bubble, .timestamp, .avatar, .avatarSizePadding, .authorName, .threadInfo, .reactions])
+
+        [.flipped, .bubble, .timestamp, .avatar, .avatarSizePadding, .authorName, .threadInfo, .reactions]
+            .forEach { options.remove($0) }
```
nuno-vieira, line 19: Why this change? Keep in mind, if you are getting a Warning, this was a mistake and it is fixed on 4.6.0 branch. Please revert this change

Examples/SlackClone/SlackMessageOptionsResolver.swift
```diff
@@ -27,7 +29,7 @@ final class SlackMessageOptionsResolver: ChatMessageLayoutOptionsResolver {
         }()
 
         if isFirstInGroup {
-            options.insert([.avatar, .timestamp, .authorName])
+            [.avatar, .timestamp, .authorName].forEach { options.insert($0) }
```
nuno-vieira, line 32: Same here

nuno-vieira (changes requested):
Requesting change just to make sure we revert that code. Other than that LGTM 👍

Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
adamrushy: Okay, I tested this and is all working as expected so we can proceed. I also checked by using the Storyboard implementation and it's all working as expected

`let chatList = DemoChannelListVC.make(with: controller, storyboard: UIStoryboard(name: "Main", bundle: .main), storyboardId: "ChannelList")`

adamrushy (commented):
Tested across the `make` function and removing the setUp call, all is working as expected and I also checked the Storyboard implementation and is all good. Just a conflict and a few other questions to answer then good to go 👍🏼

Examples/SlackClone/SlackMessageOptionsResolver.swift
tbarbugli: I was getting a warning about that

nuno-vieira (approved):
LGTM! 🚀

Examples/SlackClone/SlackMessageOptionsResolver.swift
evsaev: ✅ Discarded this change

Examples/SlackClone/SlackMessageOptionsResolver.swift
evsaev: ✅ Discarded this change

PR #1707, opened 2021-12-19
CIS-1339 Exclude deleted replies correctly

tbarbugli:
Change Fetch Request to exclude deleted thread replies from showing up in the main channel. When a reply message is deleted, its type changes from "reply" into "deleted", we can still use the `parent_message_id` field to determine if the. message is a reply or not.


evsaev (approved):
Nice catch 👍 Do we need to mention the fix in the changelog?

PR #1713, opened 2021-12-20
Fixing CI scripts

adamrushy:
### 🎯 Goal

The goal is to automate our full release pipeline as per the previous tickets. We discovered some issues on the latest release and this should fix those.

### 🛠 Implementation

We have extracted the logic for tweaking the XCFrameworks and we'll look at doing this after this release.

### ☑️ Checklist


.github/workflows/publish-release.yml
```diff
@@ -13,8 +13,34 @@ jobs:
...
+          restore-keys: ${{ runner.os }}-mint-
+      - name: Extract version from branch name (for release branches)
         if: startsWith(github.event.pull_request.head.ref, 'release/')
-              BRANCH_NAME="${{ github.event.pull_request.head.ref }}"
-              VERSION=${BRANCH_NAME#release/}
-        run: bundle exec fastlane publish_release version:$VERSION
+        run: |
+          BRANCH_NAME="${{ github.event.pull_request.head.ref }}"
+          VERSION=${BRANCH_NAME#release/}
+          echo "RELEASE_VERSION=$VERSION" >> $GITHUB_ENV
```
adamrushy, line 42: This is the fix here, setting as a Github Env variable means we can use it further down.

fastlane/Fastfile
```diff
@@ -110,6 +113,33 @@ lane :publish_release do |options|
...
+  # update_frameworks
+
+  # Create Github Release
+  github_release = set_github_release(
+                     repository_name: "GetStream/stream-chat-swift",
+                     api_token: ENV["GITHUB_TOKEN"],
+                     name: version,
+                     tag_name: version,
+                     description: changes
+                     # upload_assets: ["Products/StreamChat.zip", "Products/StreamChatUI.zip", "Products/StreamChat-All.zip"]
```
adamrushy, line 126: These will be added manually _for now_

fastlane/Fastfile
```diff
@@ -110,6 +113,33 @@ lane :publish_release do |options|
...
+  # Set tag
+  sh("git tag #{version}")
+
+  # Push tag to remote
+  push_to_git_remote(tags: true)
+
+  UI.success("Github release was created, please visit #{github_release["url"]} to see it")
+end
+
+desc "Update XCFrameworks and submit to the SPM repository"
```
adamrushy, line 141: This is extracted logic to deal with the frameworks as part of the release. Github Actions struggles to deal with multiple repo's in the same action, I am likely to make this a new workflow.

fastlane/Fastfile
```diff
@@ -31,6 +31,9 @@ end
 
 desc "Start a new release"
 lane :release do |options|
+  # Ensure we are on the main branch
+  ensure_git_branch(branch: 'main')
```
polqf, line 35: Why is this back? A release will start from develop now, in most of the cases.

fastlane/Fastfile
polqf: Could we add a comment in `update_frameworks` so that we remember that we should be updating the release there?

PR #1716, opened 2021-12-20
[CIS-1431] Make only necessary API calls in threads

tbarbugli:
When you load a thread we perform 4 API calls:

- getMessage by ID (for the parent message)
- getReplies
- userSearch
- queryChannel

There are a number of issues here that this PR resolves:

- The getReplies is called even when the parent messages has `reply_count` == 0
- The userSearch call is call for no good reason (probably a leftover from previous version)
- The queryChannel API call comes from calling `channelController.synchronize` that seems to be also not necessary 
- In most cases `ChatThreadVC` already gets a `ChatMessage` 

End result (based on demo app):

- 0 API calls when you open a thread for the first time
- 1 API call when you open a thread that has > 0 replies

### DIFF
--- a/Sources/StreamChatUI/ChatThread/ChatThreadVC.swift
+++ b/Sources/StreamChatUI/ChatThread/ChatThreadVC.swift
@@ -50,7 +50,7 @@ open class ChatThreadVC:
 
     public var messageComposerBottomConstraint: NSLayoutConstraint?
 
-    @Atomic private var loadingPreviousMessages: Bool = false
+    private var isLoadingPreviousMessages: Bool = false
 
     override open func setUp() {
         super.setUp()
@@ -66,20 +66,25 @@ open class ChatThreadVC:
         }
 
         messageController.delegate = self
-        let updateMessageControllerMessage: () -> Void = { [messageController, messageComposerVC] in
+
+        let completeSetup: (ChatMessage?) -> Void = { [messageController, messageComposerVC] message in
             if messageComposerVC.content.threadMessage == nil,
                let message = messageController?.message {
                 messageComposerVC.content.threadMessage = message
             }
-        }
-        channelController.synchronize { [messageController] _ in
-            messageController?.synchronize { _ in
-                updateMessageControllerMessage()
-                messageController?.loadPreviousReplies()
+            if let message = message, message.replyCount > 0 {
+                self.loadPreviousMessages()
             }
         }
 
-        userSuggestionSearchController.search(term: nil)
+        if let message = messageController.message {
+            completeSetup(message)
+            return
+        }
+
+        messageController.synchronize { [weak self] _ in
+            completeSetup(self?.messageController.message)
+        }
     }
 
     override open func setUpLayout() {
@@ -180,6 +185,17 @@ open class ChatThreadVC:
         return layoutOptions
     }
 
+    open func loadPreviousMessages() {
+        guard !isLoadingPreviousMessages else {
+            return
+        }
+        isLoadingPreviousMessages = true
+
+        messageController.loadPreviousReplies { [weak self] _ in
+            self?.isLoadingPreviousMessages = false
+        }
+    }
+
     // MARK: - ChatMessageListVCDelegate
 
     open func chatMessageListVC(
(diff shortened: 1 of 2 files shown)


Sources/StreamChatUI/ChatThread/ChatThreadVC.swift
```diff
@@ -66,20 +66,25 @@ open class ChatThreadVC:
         }
 
         messageController.delegate = self
-        let updateMessageControllerMessage: () -> Void = { [messageController, messageComposerVC] in
+
+        let completeSetup: (ChatMessage?) -> Void = { [messageController, messageComposerVC] message in
```
nuno-vieira, line 70:
```suggestion
        let completeSetUp: (ChatMessage?) -> Void = { [messageController, messageComposerVC] message in
```

Better to have the same camel case of `setUp` so that is easier to understand what the variable means.

Sources/StreamChatUI/ChatThread/ChatThreadVC.swift
```diff
@@ -66,20 +66,25 @@ open class ChatThreadVC:
...
             if messageComposerVC.content.threadMessage == nil,
                let message = messageController?.message {
                 messageComposerVC.content.threadMessage = message
             }
-        }
-        channelController.synchronize { [messageController] _ in
```
nuno-vieira, line 75: AFAIK the channelController.synchronize was needed when opening a Thread from a notification, where the call to the channel was not done. We need to verify if that this use case still works.

tbarbugli: good point, we should handle this in our demo app (atm we only navigate to the channel and not to the thread)

image

tbarbugli: this PR is opening a Pandora box :) 

- our demo app does not open threads when tapping on a message reply
- our docs should have an example `userNotificationCenter:didReceive` that handles both regular messages and replies

tbarbugli: @nuno-vieira navigating to a thread probably requires adding the channel to the navigation stack as well (otherwise you go back to 🦖)

evsaev: @nuno-vieira @tbarbugli regarding notification about thread reply in the channel that does not exist locally: what if we handle this on the message controller level? 

When `synchronize` is invoked it'll check if there's a channel in the database and if there's no - it'll fetch & save it before fetching a message from the API.

Having a message in the database without the underlying channel makes no sense, right?

nuno-vieira: I think Pavel's solution makes sense 👍

Sources/StreamChatUI/ChatThread/ChatThreadVC.swift
b-onc: done

Sources/StreamChatUI/ChatThread/ChatThreadVC.swift
b-onc: `getMessage`, which is called when `MessageController.synchronize` is invoked, already returns `channel` data alongside `message` data. We don't need to call `ChannelController.synchronize`

nuno-vieira: @b-onc Do you test opening a thread from a notification? Just want to make sure we don't have a regression here.

evsaev (approved):

Sources/StreamChatUI/ChatThread/ChatThreadVC.swift
evsaev: @nuno-vieira handling push notification is up to DemoApp. From an SDK perspective we are good, I think.

nuno-vieira: Yes, but the customer will open a thread and the thread needs to correctly load the messages, otherwise, there is a bug on our side. And this bug existed before, so I just wanted to make sure this change did not make a regression regarding that issue.

PR #1718, opened 2021-12-21
[CIS-1432] Extract date separator logic for better composability

nuno-vieira:
### 🔗 Issue Link
CIS-1432

### 🎯 Goal
Refactor the date separator logic so that is easier to understand and customize.

### 🛠 Implementation
Extract the date separator between 2 responsibilities and make it more composible.

### 🎨 Changes
N/A

### 🧪 Testing
N/A

### ☑️ Checklist


Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -303,6 +303,31 @@ open class ChatMessageListVC:
...
+        }
+        
+        // Only show the separator if the previous message has a different day.
+        let isDifferentDay = !Calendar.current.isDate(
+            message.createdAt,
+            equalTo: previousMessage.createdAt,
+            toGranularity: .day
+        )
+        return isDifferentDay
+    }
```
evsaev, line 329: The change looks nice 👌 Just thinking if it won't be the issue when switching to sectioned table view? 🤔

nuno-vieira: Good point 🤔 I think it should not, this function should be able to be moved to the section side since it only depends on the current message and indexPath. I can also make it internal for the time being, WDYT?

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -303,6 +303,31 @@ open class ChatMessageListVC:
         typingIndicatorView.isHidden = true
     }
 
+    /// Check if the current message being displayed should show the date separator.
+    /// - Parameters:
+    ///   - message: The message being displayed.
+    ///   - indexPath: The indexPath of the message.
+    /// - Returns: A Boolean value depending if it should show the date separator or not.
+    open func shouldShowDateSeparator(forMessage message: ChatMessage, at indexPath: IndexPath) -> Bool {
```
nuno-vieira, line 311:
```suggestion
    func shouldShowDateSeparator(forMessage message: ChatMessage, at indexPath: IndexPath) -> Bool {
```

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -325,24 +350,10 @@ open class ChatMessageListVC:
...
-        // TODO: simplify this and make it simpler to customize the logic (ie. extract this into a open `showDateSeparatorView` method)
-        let previousIndexPath = IndexPath(row: indexPath.row + 1, section: indexPath.section)
-        if let previousMessage = dataSource?.chatMessageListVC(self, messageAt: previousIndexPath) {
-            let previousDay = DateFormatter.messageListDateOverlay.string(from: previousMessage.createdAt)
-            cell.dateSeparatorView.isHidden = previousDay == currentDay
-        } else {
-            // If previous message doesn't exist show the separator as well
-            cell.dateSeparatorView.isHidden = false
+        
+        if let currentMessage = message {
```
evsaev, line 354: Hm, just noticed that we're working with the optional message. Why don't we do an early return if it's `nil`?

nuno-vieira: Yes, I noticed that as well. But didn't want to change it since I don't know if this is actually on purpose. 🤔 But AFAIK, there's no use case to not returning a message right? The UI would not render anything?

nuno-vieira: @evsaev Fixed it ✅

evsaev (approved):
LGTM 👍

evsaev (approved):

PR #1723, opened 2021-12-22
[CIS-1263, CIS-1376] Fix message list jumping when message updated

nuno-vieira:
### 🔗 Issue Link
None

### 🎯 Goal
Fix message list jumping when editing a message.

### 🛠 Implementation
Remove the `estimatedRowHeight` of the table view. 

### 🎨 Changes
| Before | After |
| ------------- | ------------- |
| <video src="https://user-images.githubusercontent.com/12814114/147157888-b3cbfa2c-4f5f-4ee6-9a25-522fe8025689.mp4"/> | <video src="https://user-images.githubusercontent.com/12814114/147157934-d5ab58b2-b718-423a-90d7-15ecfceddc52.mp4"/> |

### 🧪 Testing
Check the videos above :)

### ☑️ Checklist

### DIFF
--- a/Sources/StreamChatUI/ChatMessageList/ChatMessageListView.swift
+++ b/Sources/StreamChatUI/ChatMessageList/ChatMessageListView.swift
@@ -46,7 +46,6 @@ open class ChatMessageListView: UITableView, Customizable, ComponentsProvider {
     open func setUp() {
         keyboardDismissMode = .onDrag
         rowHeight = UITableView.automaticDimension
-        estimatedRowHeight = 150
         separatorStyle = .none
         transform = .mirrorY
     }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Fix multiple pagination requests being fired from `ChatChannelVC` and `ChatChannelListVC` [#1706](https://github.com/GetStream/stream-chat-swift/issues/1706)
 - Fix rendering unavailable reactions on `ChatMessageReactionAuthorsVC` [#1719](https://github.com/GetStream/stream-chat-swift/issues/1719)
 - Fix quoted messages not updated after edit [#1703](https://github.com/GetStream/stream-chat-swift/pull/1703)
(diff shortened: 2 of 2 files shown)


Sources/StreamChatUI/ChatMessageList/ChatMessageListView.swift
```diff
@@ -46,7 +46,6 @@ open class ChatMessageListView: UITableView, Customizable, ComponentsProvider {
     open func setUp() {
         keyboardDismissMode = .onDrag
         rowHeight = UITableView.automaticDimension
-        estimatedRowHeight = 150
```
polqf, line 49: Have you tried in different iOS versions? I remember this having different effect depending on the UIKit version.

Maybe it is safer to only remove it for certain versions that we managed to check?

nuno-vieira: I've tested with iOS 12 now, and it is working fine 🤔 I also can't find any info about that. Maybe that was in really old versions?

bielikb: @nuno-vieira could we still leverage `estimatedRowHeight` but with some common (awaited row height) -> eg cell with text message (1 liner)?

As per docs:
```
The default value is automaticDimension, which means that the table view selects an estimated height to use on your behalf. Setting the value to 0 disables estimated heights, which causes the table view to request the actual height for each cell. If your table uses self-sizing cells, the value of this property must not be 0.
When using height estimates, the table view actively manages the contentOffset and contentSize properties inherited from its scroll view. Do not attempt to read or modify those properties directly.
```

evsaev: I remember us removing the estimated height 🤔 For short text messages, it was OK but sending/removing large messages (e.g. with image attachments/long text) when content size changes significantly were causing jumps to be more noticeable.

I will give it some testing to see if this is still relevant

nuno-vieira: @bielikb As per the docs, removing the estimatedRowHeight, the estimation is now done automatically by Apple, which it looks it is better than our estimation.

PR #1732, opened 2021-12-27
[CIS-1455] Pinned messages pagination support

evsaev:
### 🔗 Issue Link

CIS-1455

### 🎯 Goal

1. Make it possible to paginate messages pinned in a channel
2. Make it possible to change the pinning date

It is worth mentioning that saving pinned messages to the database causes gaps and breaks pagination in the main channel. The PR contains changes allowing to create `ChatMessage` model directly from `MessagePayload` without saving it to the database first.

### 🛠 Implementation

- Add endpoint to fetch messages pinned in a channel
- Expose `loadPinnedMessages` API on channel controller
- Make it possible to change pinning date

### 🧪 Testing

Changes are covered by unit-tests.

Manual testing:
1. Checkout `origin/pinned-messages-pagination` branch
2. Launch `DemoApp`
3. Sign in as `C-3PO`
4. Open the channel, pin/unpin messages
5. Navigate to `PinnedMessagesVC` and see pagination is correct

### ☑️ Checklist


Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
```diff
@@ -1273,6 +1273,52 @@ public extension ChatChannelController {
...
+        
+        updater.loadPinnedMessages(in: cid, query: query) {
+            switch $0 {
+            case let .success(payload):
+                var messages: [ChatMessage] = []
+                
+                self.client.databaseContainer.write({ session in
+                    messages = payload
+                        .messages
+                        .map { .init(payload: $0, session: session) }
```
nuno-vieira, line 1309: Why don't we save the messages to the DB first, and then map to the model like we usually do? 🤔

Sources/StreamChat/Models/MessagePinning.swift
```diff
@@ -8,14 +8,12 @@ import Foundation
...
-    init(expirationDate: Date?) {
-        self.expirationDate = expirationDate
-    }
+    
+    /// The pinning date. Equals to current date by default.
+    public var pinnedAt = Date()
 
     /// Pins a message with infinite expiration.
-    public static let noExpiration = MessagePinning(expirationDate: nil)
+    public static var noExpiration: Self { .init(expirationDate: nil) }
```
nuno-vieira, line 16: Why the change to `var`? 🤔

Sources/StreamChat/Models/MessagePinning.swift
```diff
@@ -8,14 +8,12 @@ import Foundation
...
     /// The expiration date of the pinning. Infinite expiration in case it is `nil`.
     public let expirationDate: Date?
-
-    /// Private initialiser. The static initialisers should be used for better syntax and ease of use.
-    init(expirationDate: Date?) {
-        self.expirationDate = expirationDate
-    }
+    
+    /// The pinning date. Equals to current date by default.
+    public var pinnedAt = Date()
```
nuno-vieira, line 13: I don't think this is a good idea. The `pinnedAt` should be separate from the `MessagePinning` information. What if you want to set an expiration date prior to the actual action of pinning the message? The `pinnedAt` should be created like it was before when the action is performed. Setting the expiration date can be done separately from the actual action.

nuno-vieira (changes requested):
Did not finish the review yet, but already have some changes to request, so I'll submit the review in parts, since the PR is quite big 👍

Sources/StreamChat/Models/ChatMessage.swift
```diff
@@ -379,3 +379,110 @@ public enum LocalReactionState: String {
     /// Deleting of the reaction failed and cannot be fulfilled
     case deletingFailed
 }
+
+extension ChatMessage {
+    init(payload: MessagePayload, session: DatabaseSession) {
+        let currentUser = session.currentUser
```
nuno-vieira, line 385: I'm not sure if this is necessary 🤔 I remember when doing the `MessageController.loadReactions()` I was able to avoid mapping the Payload to the model directly. Can't we have the same strategy? Maybe I'm missing something

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
```diff
@@ -1273,6 +1273,52 @@ public extension ChatChannelController {
...
+    ///   - pageSize: The # of pinned messages to load. Equals to `25` by default.
+    ///   - sorting: The sorting options. By default, results are sorted descending by `pinned_at` field.
+    ///   - pagination: The pagination parameter. If `nil` is provided, most recently pinned messages are fetched.
+    ///   - completion: The completion to be called on **callbackQueue** when request is completed.
+    func loadPinnedMessages(
+        pageSize: Int = .messagesPageSize,
+        sorting: [Sorting<PinnedMessagesSortingKey>] = [],
+        pagination: PinnedMessagesPagination? = nil,
+        completion: @escaping (Result<[ChatMessage], Error>) -> Void
+    ) {
```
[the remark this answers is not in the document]
evsaev: Yeah, good you've asked 👍 

Regarding the `offset`, I was talking to the backend team and got the answer that it is inefficient and we should be avoiding this pagination approach.

That's a valid point that we don't persist search results ATM. The thing is we are timeboxed with this feature while having the most general stateless API allows us to add `pinnedMessages: [ChatMessage]` + `loadNext` + delegate callback on top of that in the future

Package.swift
```diff
@@ -189,6 +189,7 @@ var streamChatSourcesExcluded: [String] { [
     "Query/ChannelMemberListQuery_Tests.swift",
     "Query/FilterDecoding_Tests.swift",
     "Query/Pinned Messages/PinnedMessagesPagination_Tests.swift",
+    "Query/Pinned Messages/PinnedMessagesSortingKey_Tests.swift",
```
nuno-vieira, line 192: Same here, please avoid spaces in folders

Sources/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -565,9 +565,9 @@ extension NSManagedObjectContext: MessageDatabaseSession {
         guard let currentUserDTO = currentUser else {
             throw ClientError.CurrentUserDoesNotExist()
         }
-        let pinnedDate = Date()
+        
         message.pinned = true
-        message.pinnedAt = pinnedDate
+        message.pinnedAt = pinning.pinnedAt
```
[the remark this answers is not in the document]
evsaev: That was the most bloodless way to pass the timestamp 🙂 `MessagePinning` sounds like smth that can be extended with the timestamp. Another approach would involve changing controller/updater/database session + places where the session is invoked.

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
evsaev: The same reason why message search results should not be persisted: when saved to the database they are linked to the channel which leads to gaps in the local channel history and breaks the pagination (if we have `0..<20` messages locally and get `100th` pinned message, we will show `0..<20, 99` in channel and the next page we will be loading is `100..<120`)

Sources/StreamChat/Models/MessagePinning.swift
evsaev: Because of added `pinnedAt` field 🙂 When we get a pinning instance we want the timestamp to be equal to the current date, not the date computed when static `noExpiration` was accessed for the first time

Sources/StreamChat/Models/MessagePinning.swift
evsaev: Yeah, I was thinking about the following way:
```swift
var pinning = MessagePinning.noExpiration
pinning.pinnedAt = Date().addingTimeInterval(-10)
messageController.pinMessage(pinning) {}
```

But that's an edge case when someone wants to use the pinning date different from `now`.

Sources/StreamChat/APIClient/Endpoints/Payloads/MessagePayloads.swift
```diff
@@ -284,6 +288,11 @@ struct MessageRepliesPayload: Decodable {
     let messages: [MessagePayload]
 }
 
+/// An object describing pinned messages JSON payload.
+struct PinnedMessagesPayload: Decodable {
```
polqf, line 292: Why do we have duplicated structs with different names?

Sources/StreamChat/Models/ChatMessage.swift
evsaev: Explained in here: https://github.com/GetStream/stream-chat-swift/pull/1732#discussion_r775987706

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
polqf: Isn't storing them in the DB an issue then? Are we allowing gaps with this implementation?

Sources/StreamChat/APIClient/Endpoints/Payloads/MessagePayloads.swift
evsaev: Yep, we could reuse `MessageRepliesPayload` but having this name in terms of pinned messages would look weird. The option was to rename `MessageRepliesPayload -> MessageListPayload` but this would cause changes in other places.

While writing the reply have got the idea 💡. We can rename `MessageRepliesPayload -> MessageListPayload` and leverage type aliases to have meaningful names. WDYT @polqf?

evsaev:
> Isn't storing them in the DB an issue then? Are we allowing gaps with this implementation?

We don't save messages to the database. The session is passed to model initializer so we can populate fields (`isSentByCurrentUser`) that are not included into the payload.

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
evsaev: We don't save messages to the database. The session is passed to model initializer so we can populate fields (isSentByCurrentUser) that are not included into the payload.

Package.swift
```diff
@@ -186,6 +188,8 @@ var streamChatSourcesExcluded: [String] { [
     "Query/FilterTestExtensions_Tests.swift",
     "Query/ChannelMemberListQuery_Tests.swift",
     "Query/FilterDecoding_Tests.swift",
+    "Query/Pinned Messages/PinnedMessagesPagination_Tests.swift",
+    "Query/Pinned Messages/PinnedMessagesQuery_Tests.swift",
```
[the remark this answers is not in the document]
evsaev: ✅ Renamed

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
```diff
@@ -1273,6 +1273,52 @@ public extension ChatChannelController {
             }
         }
     }
+    
+    /// Loads the given number of pinned messages based on pagination parameter in the current channel.
+    ///
+    /// - Parameters:
+    ///   - pageSize: The # of pinned messages to load. Equals to `25` by default.
```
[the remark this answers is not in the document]
evsaev: ✅ Fixed

Sources/StreamChat/APIClient/Endpoints/Payloads/MessagePayloads.swift
evsaev: ✅ Added type aliases

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
nuno-vieira: I think we already have something to fix that issue:
```swift
// Some pinned messages might be in the local database, but should not be fetched
// if they do not belong to the regular channel query.
let ignoreOlderMessagesPredicate = NSCompoundPredicate(orPredicateWithSubpredicates: [
    .init(format: "channel.oldestMessageAt == nil"),
    .init(format: "createdAt >= channel.oldestMessageAt")
])
```

Sources/StreamChat/Models/MessagePinning.swift
nuno-vieira: Why `Date().addingTimeInterval(-10)` ? 

```swift
let pinning = MessagePinning.noExpiration
messageController.pinMessage(pinning)
```

☝️ This should be the API. The `pinnedAt = Date()` should be done automatically inside the `messageController.pinMessage(pinning)` since it is the actual action.

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
nuno-vieira: I remember I fixed this issue long time ago 🙂 It's not the cleanest solution, but it works

Sources/StreamChat/Models/MessagePinning.swift
nuno-vieira: The goal of `MessagePinning` was to only set the information of the expiration date, not the `pinnedAt`. IMO we are mixing responsibilities here that should not be mixed.

Sources/StreamChat/Models/MessagePinning.swift
evsaev: All right, let's move it to `pinMessage` as a second param 👍

evsaev: ✅ Done

Sources/StreamChat/Models/MessagePinning.swift
evsaev: With `pinnedAt: Date` as a parameter of `pinMessage` this discussion is no longer relevant.

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
evsaev: okay, I was not aware we have this 🤔 It means we're safe to save messages from search/pinned ones to database actually 🙂 cc @tbarbugli

Sources/StreamChat/Database/DTOs/MessageDTO.swift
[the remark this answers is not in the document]
evsaev: `pinnedAt` is no longer part of MessagePinning type 👍

nuno-vieira (dismissed):
LGTM! ✅

nuno-vieira:
@evsaev Where is this: "Navigate to PinnedMessagesVC and see pagination is correct"? I don't see any `PinnedMessageVC`

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
tbarbugli: @nuno-vieira is right, we can do this later on and keep the current solution to avoid gaps

PR #1733, opened 2021-12-27
Add Split View Support on iPad for Demo App

nuno-vieira:
### 🔗 Issue Link
N/A

### 🎯 Goal
Add Split View support for iPads on the Demo App. This is very important for QAing different scenarios which are not possible with the normal iOS flow. It can also serve as an example of how customers can integrate this in their apps.

### 🛠 Implementation
Adds `UISplitViewController` support in the Demo App.

### 🎨 Changes
simulator_screenshot_402CC23C-7B6D-4D55-B543-E5C3C8CB8A8F

### 🧪 Testing
- Run the App with an iPad
- Should show channels on the Left

- Run the App with an iPhone
- Should not show the split view

### ☑️ Checklist


DemoApp/DemoAppCoordinator.swift
```diff
@@ -114,7 +114,7 @@ final class DemoAppCoordinator: NSObject, UNUserNotificationCenterDelegate {
         // Channels with the current user
         let controller = ChatClient.shared
             .channelListController(query: .init(filter: .containMembers(userIds: [userCredentials.id])))
-        let chatList = DemoChannelListVC.make(with: controller)
+        let chatList = DemoChannelListVC.make(with: controller) as! DemoChannelListVC
```
adamrushy, line 117: We should safely unwrap, I know it's the Demo App but still (shows best practice)

adamrushy (changes requested):
Just one comment, otherwise ✅ 🚀

DemoApp/DemoAppCoordinator.swift
nuno-vieira: Actually, this made me spot an issue. The `make` function should actually return the type based on `Self` so that when subclassing, we don't need to cast it 🙂 

Fixed ✅

tbarbugli (changes requested):
It would be very useful to highlight the current channel from the Channel List

ie. from React
image

nuno-vieira:
> It would be very useful to highlight the current channel from the Channel List
> 
> ie. from React image

Yes good point 👍 Will update it

nuno-vieira:
> It would be very useful to highlight the current channel from the Channel List
> 
> ie. from React image

This was actually much hard to do than I thought because our channel list is constantly reporting changes, which resets the `isSelected` state. I was able to manage to make it work and have acceptable behaviour.

adamrushy (approved):
Looks good to me 🔥

PR #1736, opened 2021-12-28
[CIS-1227] [CIS-1430] Channel.membership correct handling

b-onc:
### 🔗 Issue Link

CIS-1227
CIS-1430

### 🎯 Goal

CIS-1227 : 
`Channel.membership` was not parsed / removed in all cases. Some WS events were not handled correctly.
CIS-1430 :
`MemberListController` didn't report `member.added` and `notification.added_to_channel` events correctly.

### 🛠 Implementation

`MemberEventsMiddleware` now correctly handles all events.
We cannot add members to a `MemberListQuery` if the query has a filter, since we can't verify if the member matches the filter.

### 🧪 Testing

Unit tests are implemented.

### ☑️ Checklist

### DIFF
--- a/Sources/StreamChat/WebSocketClient/EventMiddlewares/MemberEventMiddleware.swift
+++ b/Sources/StreamChat/WebSocketClient/EventMiddlewares/MemberEventMiddleware.swift
@@ -28,33 +28,59 @@ struct MemberEventMiddleware: EventMiddleware {
                 
                 // We remove the member from the channel
                 channel.members.remove(member)
+                if let membership = channel.membership, membership.user.id == event.user.id {
+                    channel.membership = nil
+                }
                 
                 // If there are any MemberListQueries observing this channel,
                 // we need to update them too
                 member.queries.removeAll()
                 
             case let event as NotificationAddedToChannelEventDTO:
-                _ = try session.saveChannel(payload: event.channel, query: nil)
+                let channel = try session.saveChannel(payload: event.channel, query: nil)
+                let member = try session.saveMember(payload: event.member, channelId: event.channel.cid)
+                channel.membership = member
                 
             case let event as NotificationRemovedFromChannelEventDTO:
                 guard let channel = session.channel(cid: event.cid) else {
                     // No need to throw ChannelNotFound error here
-                    log.debug("Channel \(event.cid) not found for MemberRemovedEventDTO")
+                    log.debug("Channel \(event.cid) not found for NotificationRemovedFromChannelEventDTO")
                     break
                 }
                 
                 guard let member = channel.members.first(where: { $0.user.id == event.member.user.id }) else {
                     // No need to throw MemberNotFound error here
-                    log.debug("Member \(event.member.user.id) not found for MemberRemovedEventDTO")
+                    log.debug("Member \(event.member.user.id) not found for NotificationRemovedFromChannelEventDTO")
                     break
                 }
                 
                 // We remove the member from the channel
                 channel.members.remove(member)
+                if let membership = channel.membership, membership.user.id == event.member.user.id {
+                    channel.membership = nil
+                }
                 
                 // If there are any MemberListQueries observing this channel,
                 // we need to update them too
                 member.queries.removeAll()
+                
+            case let event as NotificationInviteAcceptedEventDTO:
+                let channel = try session.saveChannel(payload: event.channel, query: nil)
+                let member = try session.saveMember(payload: event.member, channelId: event.channel.cid)
+                channel.membership = member
+                
+            case let event as NotificationInviteRejectedEventDTO:
+                let channel = try session.saveChannel(payload: event.channel, query: nil)
+                let member = try session.saveMember(payload: event.member, channelId: event.channel.cid)
+                channel.membership = member
+                
+            case let event as NotificationInvitedEventDTO:
+                guard let channel = session.channel(cid: event.cid) else {
+                    // No need to throw ChannelNotFound error here
+                    break
+                }
+                let member = try session.saveMember(payload: event.member, channelId: event.cid)
+                channel.membership = member
             default:
                 break
             }
--- a/Sources/StreamChat/WebSocketClient/EventMiddlewares/MemberEventMiddleware.swift
+++ b/Sources/StreamChat/WebSocketClient/EventMiddlewares/MemberEventMiddleware.swift
@@ -13,7 +13,10 @@ struct MemberEventMiddleware: EventMiddleware {
                 try session.saveMember(payload: event.member, channelId: event.cid)
                 
             case let event as MemberAddedEventDTO:
-                try session.saveMember(payload: event.member, channelId: event.cid)
+                if let channel = session.channel(cid: event.cid) {
+                    let member = try session.saveMember(payload: event.member, channelId: event.cid)
+                    insertMemberToMemberListQueries(channel, member)
+                }
 
             case let event as MemberRemovedEventDTO:
                 guard let channel = session.channel(cid: event.cid) else {
@@ -41,6 +44,8 @@ struct MemberEventMiddleware: EventMiddleware {
                 let member = try session.saveMember(payload: event.member, channelId: event.channel.cid)
                 channel.membership = member
                 
+                insertMemberToMemberListQueries(channel, member)
+                
             case let event as NotificationRemovedFromChannelEventDTO:
                 guard let channel = session.channel(cid: event.cid) else {
                     // No need to throw ChannelNotFound error here
@@ -56,9 +61,8 @@ struct MemberEventMiddleware: EventMiddleware {
                 
                 // We remove the member from the channel
                 channel.members.remove(member)
-                if let membership = channel.membership, membership.user.id == event.member.user.id {
-                    channel.membership = nil
-                }
+                // We reset membership since we're no longer a member
+                channel.membership = nil
                 
                 // If there are any MemberListQueries observing this channel,
                 // we need to update them too
@@ -81,6 +85,9 @@ struct MemberEventMiddleware: EventMiddleware {
                 }
                 let member = try session.saveMember(payload: event.member, channelId: event.cid)
                 channel.membership = member
+                
+                insertMemberToMemberListQueries(channel, member)
+                
             default:
                 break
             }
@@ -89,5 +96,17 @@ struct MemberEventMiddleware: EventMiddleware {
         }
 
         return event
+    }
+    
+    private func insertMemberToMemberListQueries(_ channel: ChannelDTO, _ member: MemberDTO) {
+        // If there are any `MemberListQuery`s observing this Channel
+        // without any filters (so the query observes all members)
+        // the new Member should be linked to them too
+        // so `MemberListController` works as expected
+        // To make it work with queries with filters, we need to mirror `ChannelListController` logic
+        // `shouldListUpdatedChannel` and such
+        channel.memberListQueries.filter { $0.filterJSONData == nil }.forEach {
+            $0.members.insert(member)
+        }
     }
 }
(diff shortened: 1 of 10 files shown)


Sources/StreamChat/WebSocketClient/EventMiddlewares/MemberEventMiddleware.swift
```diff
@@ -13,7 +13,19 @@ struct MemberEventMiddleware: EventMiddleware {
...
-                try session.saveMember(payload: event.member, channelId: event.cid)
+                let member = try session.saveMember(payload: event.member, channelId: event.cid)
+                
+                // If there are any `MemberListQuery`s observing this Channel
+                // without any filters (so the query observes all members)
+                // the new Member should be linked to them too
+                // so `MemberListController` works as expected
+                // To make it work with queries with filters, we need to mirror `ChannelListController` logic
+                // `shouldListUpdatedChannel` and such
+                if let channel = session.channel(cid: event.cid) {
```
evsaev, line 24: Should we even save a member if the channel does not exist locally?

Sources/StreamChat/WebSocketClient/EventMiddlewares/MemberEventMiddleware.swift
```diff
@@ -28,33 +40,80 @@ struct MemberEventMiddleware: EventMiddleware {
...
+                
+                // If there are any `MemberListQuery`s observing this Channel
+                // without any filters (so the query observes all members)
+                // the new Member should be linked to them too
+                // so `MemberListController` works as expected
+                // To make it work with queries with filters, we need to mirror `ChannelListController` logic
+                // `shouldListUpdatedChannel` and such
+                channel.memberListQueries.filter { $0.filterJSONData == nil }.forEach {
+                    $0.members.insert(member)
+                }
```
evsaev, line 64: Would it be possible to encapsulate this in some private func to avoid code duplication? 🙏

Sources/StreamChat/WebSocketClient/EventMiddlewares/MemberEventMiddleware.swift
```diff
@@ -28,33 +40,80 @@ struct MemberEventMiddleware: EventMiddleware {
                 
                 // We remove the member from the channel
                 channel.members.remove(member)
+                if let membership = channel.membership, membership.user.id == event.user.id {
+                    channel.membership = nil
+                }
```
evsaev, line 45: Hm, why we need this part? Won't we receive `notification.removed_from_channel` event?

Sources/StreamChat/WebSocketClient/Events/NotificationEvents.swift
```diff
@@ -217,6 +227,7 @@ public struct NotificationRemovedFromChannelEvent: ChannelSpecificEvent {
 class NotificationRemovedFromChannelEventDTO: EventDTO {
     let cid: ChannelId
     let user: UserPayload
+    // This `member` field is equal to the `membership` field in channel query
     let member: MemberPayload
```
evsaev, line 231: What if we rename `member -> membership` to be explicit that it's about the current user?

Sources/StreamChat/WebSocketClient/EventMiddlewares/MemberEventMiddleware.swift
b-onc: Channel data will be saved in another middleware

Sources/StreamChat/WebSocketClient/EventMiddlewares/MemberEventMiddleware.swift
b-onc: yes, but just to be on the safe side, to safeguard against possible backend bugs

Sources/StreamChat/WebSocketClient/Events/NotificationEvents.swift
b-onc: then we would have different naming with our docs and other SDKs

Sources/StreamChat/WebSocketClient/EventMiddlewares/MemberEventMiddleware.swift
evsaev: Yeah, but the event holds only `cid` 🙂 If we get this event without having this channel locally we will save a member. 

My suggestion is to `guard let channel = ... else { break }` before `let member = ...; channel.members.insert(member)`

Sources/StreamChat/WebSocketClient/Events/NotificationEvents.swift
evsaev: Notification events are always about the current user so to me it's fine the member data has different naming from the rest of the member events. But I'm fine with both so leaving it up to you 🙂

b-onc: I think naming parity is important so i'll leave it as is

Sources/StreamChat/WebSocketClient/EventMiddlewares/MemberEventMiddleware.swift
evsaev: Hmm, we might actually need that 🤔 If the current user is watching a channel and it gets removed from the member list, will we get `notification.removed_from_channel` or `member.removed`?

Sources/StreamChat/WebSocketClient/EventMiddlewares/MemberEventMiddleware.swift
b-onc: on another thought, you can have a `MemberListController` without having the channel data saved?

Sources/StreamChat/WebSocketClient/EventMiddlewares/MemberEventMiddleware.swift
b-onc: the first one

Sources/StreamChat/WebSocketClient/EventMiddlewares/MemberEventMiddleware.swift
```diff
@@ -28,33 +32,64 @@ struct MemberEventMiddleware: EventMiddleware {
...
-                    log.debug("Member \(event.member.user.id) not found for MemberRemovedEventDTO")
+                    log.debug("Member \(event.member.user.id) not found for NotificationRemovedFromChannelEventDTO")
                     break
                 }
                 
                 // We remove the member from the channel
                 channel.members.remove(member)
+                if let membership = channel.membership, membership.user.id == event.member.user.id {
+                    channel.membership = nil
+                }
```
evsaev, line 67: I would always reset the membership 🙂 Especially when `membership.user.id != event.member.user.id`. If the user id does not match the membership it means we didn't properly clean up the database from the previous user and in this case we'd better recover the local date to be valid for the current user

```suggestion
                channel.membership = nil
```

Sources/StreamChat/WebSocketClient/EventMiddlewares/MemberEventMiddleware.swift
evsaev: If we are getting events it means the current user is connected so, yeah, it will a bug on the backend if the wrong event/no event is fired. I would stick to the approach when `channel.membership` is modified only by `notification.*` events

Sources/StreamChat/WebSocketClient/EventMiddlewares/MemberEventMiddleware.swift
evsaev: Hm, good point 👍 The member indeed can be useful for `MemberListController` but on the other hand, it can break the pagination if the member delivered with the event is outside the fetched range

evsaev: since we are moving towards the state when search results are not persisted checking the channel existence sounds a bit safer

b-onc: alright, fair point

Sources/StreamChat/WebSocketClient/EventMiddlewares/MemberEventMiddleware.swift
evsaev: Discussed internally, decided to have the safeguard in place.

Sources/StreamChat/WebSocketClient/EventMiddlewares/MemberEventMiddleware_Tests.swift
```diff
@@ -104,6 +104,54 @@ final class MemberEventMiddleware_Tests: XCTestCase {
...
+        
+        // Load the channel
+        var channel: ChatChannel? {
+            database.viewContext.channel(cid: cid)?.asModel()
+        }
+        
+        // Load the MemberListQueryDTO
+        var memberListQueryDTO: ChannelMemberListQueryDTO? {
+            database.viewContext.channelMemberListQuery(queryHash: memberListQuery.queryHash)
+        }
```
evsaev, line 142: nit:
```suggestion
        let channel = try XCTUnwrap(
         database.viewContext.channel(cid: cid)?.asModel()
         )
        
        // Load the MemberListQueryDTO
        let memberListQueryDTO = try XCTUnwrap(
         database.viewContext.channelMemberListQuery(queryHash: memberListQuery.queryHash)
        )
```

evsaev: hmm, are we using the channel at all? 🤔

Sources/StreamChat/WebSocketClient/EventMiddlewares/MemberEventMiddleware_Tests.swift
```diff
@@ -336,9 +463,211 @@ final class MemberEventMiddleware_Tests: XCTestCase {
...
+        // Create channel and MemberListQuery in the database.
+        try database.writeSynchronously { session in
+            try session.saveChannel(payload: channelPayload)
+            try session.saveMember(payload: existingMember, channelId: cid, query: memberListQuery)
+        }
+        
+        // Load the channel
+        var channel: ChatChannel? {
+            database.viewContext.channel(cid: cid)?.asModel()
+        }
```
evsaev, line 554: we don't access this variable, do we?

Tests/StreamChatTests/MockEnpointResponses/Events/Notification/NotificationAddedToChannel+MissingFields.json
```diff
@@ -1,4 +1,29 @@
...
+      "role" : "member",
+      "channel_role" : "channel_member",
+      "user_id" : "general_grievous",
+      "updated_at" : "2021-12-21T15:51:42.125776Z",
+      "user" : {
+          "id" : "general_grievous",
+          "banned" : false,
+          "extraData" : {
+              
+          },
```
evsaev, line 15:
```suggestion
```

Tests/StreamChatTests/MockEnpointResponses/Events/Notification/NotificationAddedToChannel.json
```diff
@@ -1,4 +1,29 @@
...
+      "role" : "member",
+      "channel_role" : "channel_member",
+      "user_id" : "general_grievous",
+      "updated_at" : "2021-12-21T15:51:42.125776Z",
+      "user" : {
+          "id" : "general_grievous",
+          "banned" : false,
+          "extraData" : {
+              
+          },
```
evsaev, line 15:
```suggestion
          
```

evsaev (approved):
Great to see it fixed 💪 Thanks!

Sources/StreamChat/WebSocketClient/EventMiddlewares/MemberEventMiddleware_Tests.swift
b-onc: `channel` was a copy/paste leftover :D we can't use `let memberListQuery` since we need it to be loaded again when we call it again

Sources/StreamChat/WebSocketClient/EventMiddlewares/MemberEventMiddleware_Tests.swift
b-onc: correct, copy/paste leftover, thank you

PR #1737, opened 2021-12-28
CIS-1457 Handle notification.channel_deleted event

tbarbugli:
Handle `notification.channel_deleted` event so that we update the Channel List correctly


Sources/StreamChat/WebSocketClient/Events/NotificationEvents.swift
```diff
@@ -416,3 +416,35 @@ class NotificationInviteRejectedEventDTO: EventDTO {
         )
     }
 }
+
+/// Triggered when the current user rejects an invite to a channel.
+public struct NotificationChannelDeletedEvent: ChannelSpecificEvent {
+    let cid: ChannelId
+    let channel: ChatChannel
+    let createdAt: Date
```
polqf, line 424: Is this createdAt property the date in which the channel got deleted? if that's the case, can we document it?

Sources/StreamChat/WebSocketClient/Events/NotificationEvents.swift
```diff
@@ -416,3 +416,35 @@ class NotificationInviteRejectedEventDTO: EventDTO {
...
+    let payload: EventPayload
+}
+
+class NotificationChannelDeletedEventDTO: EventDTO {
+    let cid: ChannelId
+    let channel: ChannelDetailPayload
+    let createdAt: Date
```
polqf, line 431: Same here

nuno-vieira (approved):
LGTM! ✅

Sources/StreamChat/WebSocketClient/Events/NotificationEvents.swift
```diff
@@ -416,3 +416,38 @@ class NotificationInviteRejectedEventDTO: EventDTO {
...
+/// Triggered when a channel is deleted, this event is delivered to all channel members
+public struct NotificationChannelDeletedEvent: ChannelSpecificEvent {
+    /// The cid of the deleted channel
+    let cid: ChannelId
+
+    /// The channel that was deleted
+    let channel: ChatChannel
+
+    /// The event timestamp.
+    let createdAt: Date
```
nuno-vieira, line 429: Shouldn't these properties be public?

nuno-vieira (changes requested):
Requesting changes since I believe we are missing `public` on some props

Issue #1739, opened 2021-12-29
ChatMessageCell dateSeparatorView property should be public


vladpre92:
## What did you do?
Tried to override ChatMessageCell dateSeparatorView content format.

## What did you expect to happen?
To be able to do it.

## What happened instead?
I can't.

## GetStream Environment
**GetStream Chat version:** 4.7.0
**GetStream Chat frameworks:** StreamChat, StreamChatUI
**iOS version:** Any
**Swift version:** 5
**Xcode version:** 13.1
**Device:** Any

## Additional context
Hi guys!

Tried to override the date format for date overlay view for a message list view using `func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell` delegate method from `ChatMessageListVC` but I couldn't because of the internal access of this view (`dateSeparatorView` from `ChatMessageCell`).

I think we should be allowed to do this. Or...is there another approach to do this?

Thank you!

nuno-vieira:
Hi @vladpre92! Indeed the `dateSeparatorView` is meant to be internal for now. Anyway, one way should be able to change this is by changing the `DateFormatter.messageListDateOverlay` but unfortunately the formatted is a `static let` and it should be a `static var` instead. We can open a quick PR to change this. Would this be enough for you? Or by changing only the formatter is not enough for you? If it is not, we can consider opening the `dateSeparatorView`.

Thank you again!

vladpre92:
Thank you for your prompt response!
Either way is ok for us. 
Thank you!

bielikb:
Hi @vladpre92,

we're going to add possibility to configure your formatters in one of the following releases.

We will notify you, once this is done.

Best,
Boris

nuno-vieira:
Hi @vladpre92,

Thank you for your patience!

Our latest release 4.10.0 now provides a way to format the timestamps in a more convenient way. Make sure to check our [Data Formatting](https://getstream.io/chat/docs/sdk/ios/uikit/data-formatting/) article to learn more.

In your particular case, if you change the `Appereance.formatters.messageListSeparator`with your custom one, it will change the style on both the ribbon between each message group and also the top overlay ribbon. Let us know if you have any questions on how to use the formatters. If you do, please re-open this issue 🙂

PR #1742, opened 2021-12-29
[CIS-1459] Make Data Formatters Configurable

nuno-vieira:
### 🔗 Issue Link
CIS-1459
https://github.com/GetStream/stream-chat-swift/issues/1739

### 🎯 Goal
Provide a way to change the data formatting in the SDK.

### 🛠 Implementation
Expose all formatters in the `Appearance` configuration through the `Appearance.default.formatters`. Each formatter is represented by a Protocol so that the customer can provide their own implementation or change the default ones.

### 🎨 Changes
N/A

### 🧪 Testing

Example on how to change a default formatter:
```swift
let defaultTimestampFormatter = DefaultMessageTimestampFormatter()
defaultTimestampFormatter.dateFormatter.timeStyle = .medium
Appearance.default.formatters.messageTimestamp = defaultTimestampFormatter
```

Example on how to provide a custom formatter:
```swift
class CustomUserLastActivityFormatter: UserLastActivityFormatter {
    public lazy var dateFormatter: DateFormatter = {
        let formatter = DateFormatter()
        formatter.dateStyle = .short
        formatter.timeStyle = .short
        formatter.doesRelativeDateFormatting = true
        formatter.locale = .autoupdatingCurrent
        return formatter
    }()

    func format(_ date: Date) -> String? {
        "last active: " + dateFormatter.string(from: date)
    }
}
...
Appearance.default.formatters.userLastActivity = CustomUserLastActivityFormatter()
```

### ☑️ Checklist


CHANGELOG.md
```diff
@@ -3,6 +3,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 # Upcoming
 
+### ✅ Added
+- Make Date Formatters mutable to allow customization [#1742](https://github.com/GetStream/stream-chat-swift/pull/1742)
```
tbarbugli, line 7: Make date formatters configurable sounds better ;)

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
```diff
@@ -57,7 +57,7 @@ open class ChatMessageContentView: _View, ThemeProvider {
     }
 
     /// The date formatter of the `timestampLabel`
-    public lazy var dateFormatter: DateFormatter = .makeDefault()
+    public lazy var dateFormatter: DateFormatter = .messageTimestamp
```
tbarbugli, line 60: doc here does not make a lot of sense

tbarbugli (changes requested):
We should add something to the SDK docs about this

polqf (approved):

nuno-vieira:
> We should add something to the SDK docs about this

Will add the docs once we define the final API 👍

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
nuno-vieira: I've updated them a little bit now, but it actually makes sense, this is the formatted used by the `timestampLabel` in this case. Were you thinking of a better doc?

Sources/StreamChatUI/Utils/DateFormatter+Extensions.swift
```diff
@@ -5,33 +5,41 @@
...
     
-    /// Formatter that is used to format date for scrolling overlay that should display
-    /// day when message below was sent
-    public static let messageListDateOverlay: DateFormatter = {
-        let df = DateFormatter()
-        df.setLocalizedDateFormatFromTemplate("MMMdd")
-        df.locale = .autoupdatingCurrent
-        return df
+    /// A formatter that converts the message date to textual representation and used as a top overlay in the message list.
+    public static var messageListDateOverlay: DateFormatter = {
```
evsaev, line 18: The approach we'd like to follow is:
```swift
struct Appearance {
  var dateFormatters = DateFormatters()

  struct DateFormatters {
     var formatter1 = …
     var formatter2 = …
  }
}
```

The **messageListDateOverlay** is the only publicly exposed static formatter (which we have to maintain) that is in conflict with the `Appearance`, right? 🤔 What if we make it forwarding to `Appearance.default`:
```swift
extension DateFormatter {
    @available(*, deprecated, message: "Please use `Appearance.default.dateFormatters.formatter1` instead")
    public static var messageListDateOverlay: DateFormatter {
      get { Appearance.default.dateFormatters.formatter1 }
      set { Appearance.default.dateFormatters.formatter1 = newValue }
    }
}
```

It's fine formatters are not static thanks to `Appearance.default` used in 99% of cases. A shared instance get instantiated with all its props just once when first accessed. And it's still possible to customize the formatter for the specific view-hierarchy subtree.
 
@nuno-vieira  WDYT?

evsaev: we can also make dateFormatters and all underlying pros `lazy`

nuno-vieira: We also have Number formatters. I will change the API and try to change to `Appearance.default.formatters` 👍

nuno-vieira: ✅

bielikb (approved):

nuno-vieira:
Converted to draft, since there will be some team discussions on the API

docusaurus/docs/iOS/uikit/data-formatting.md
```diff
@@ -0,0 +1,57 @@
+---
+title: Data Formatting
+---
+
+import ThemingNote from '../common-content/theming-note.md'
+
+You can customize how data is formatted across all the UI components provided by `StreamChatUI`. The SDK allows you to change the data formatting through the `Appearance` configuration where you can change, for example, how message dates, video durations, and times are formatted in each component. Each formatter can be accessed through the `Appearance.formatters` property. You can replace a formatter with a custom one since each formatter is represented by a `Protocol`, but you can also do simple changes by changing the default formatters provided by the SDK.
```
bielikb, line 7: This paragraph could be rephrased. eg

```suggestion
You can customize how data is formatted across all the UI components provided by `StreamChatUI`. The SDK allows you to change the data formatting through the `Appearance.formatters` configuration.

Following data can be formatted:
* channel name
* message timestamps
* video durations & times

For simple customizations, use the default formatters provided by the SDK. Furtherly you can leverage your own formatter by adhering it to the respective data formatter `protocol`. 
```

docusaurus/docs/iOS/uikit/data-formatting.md
```diff
@@ -0,0 +1,57 @@
...
+        return formatter
+    }()
+
+    func format(_ date: Date) -> String? {
+        "last active: " + dateFormatter.string(from: date)
+    }
+}
+```
+
+Then, set the custom formatter in the `Appearance` config:
```
bielikb, line 47:
```suggestion
Finally, set the custom formatter in the `Appearance` config:
```

bielikb (commented):
Great job @nuno-vieira. I left two doc suggestions

Sources/StreamChatUI/Appearance+Formatters/MessageDateSeparatorFormatter.swift
```diff
@@ -0,0 +1,28 @@
...
+/// A formatter that converts the message date separator used in the message list to textual representation.
+/// This formatter is used to display the message date between each group of messages
+/// and the top date overlay in the message list.
+public protocol MessageDateSeparatorFormatter {
+    func format(_ date: Date) -> String
+}
+
+/// The default message date separator formatter.
+open class DefaultMessageDateSeparatorFormatter: MessageDateSeparatorFormatter {
+    public lazy var dateFormatter: DateFormatter = {
```
polqf, line 16: We need to be careful with that. `lazy` vars are not thread safe.

polqf: I'm seeing that this was already the case 🤔. We might need to review in the future lazy implementations that are accessed globally.

polqf (approved):

Sources/StreamChatUI/Appearance+Formatters/MessageDateSeparatorFormatter.swift
nuno-vieira: That makes sense 👍 I can easily change this to `var`, will do it now 👍 DateFromatter's are lazily created either way only when a function is called (like `string(from:)`), so it is fine to have them has normal vars in this case.

docusaurus/docs/iOS/uikit/data-formatting.md
nuno-vieira: @bielikb I've simplified this introduction, let me know if you like it 🙂

Sources/StreamChatUI/Appearance+Formatters/MessageDateSeparatorFormatter.swift
nuno-vieira: @polqf Done ✅

bielikb (approved):

PR #1745, opened 2021-12-30
[CIS-1456] Add hard delete messages support

nuno-vieira:
### 🔗 Issue Link
CIS-1456

### 🎯 Goal
- Add support for hard delete messages in the LLC SDK.
- Add Demo App Config UI

### 🛠 Implementation
- Adds a new `hard` option in the `messageController.deleteMessage()` to make the request with `hard: true`
- Adds a new  `isHardDeleted` to the `MessageDTO`. This is useful to filter out hard deleted messages from the Core Data queries. Also, when a hard delete message fails, the `isHardDeleted` is set back to `false` which will be possible to retry the hard delete. If the hard delete succeeds only then we remove it from the local DB.
- Adds a new `isHardDelete` to the events payload so that whenever a message is saved, if it came from a deleted event and the `isHardDelete == true` we remove it from the local database.

### 🎨 Changes

https://user-images.githubusercontent.com/12814114/148569077-3da4667f-fe9c-479c-9b85-344fdc106374.mp4

### 🧪 Testing

**Test Case 1:** **(Hard delete your own message)**
 - Turn on hard delete messages in Demo App Configuration
 - Send a message
 - Hard delete a message
 - Message should disappear
 - Navigate back to the channel list
 - The channel last message should be changed as well

**Test Case 2:** **(Hard deleted message from another user)**
 - Turn on hard delete messages in Demo App Configuration
 - Hard delete a message from a different device
 - Message should disappear
 - Navigate back to the channel list
 - The channel last message should be changed as well

**Test Case 3:** **(Hard deleted message while offline)**
- Turn on hard delete messages in Demo App Configuration
- Turn off Network
- Hard Delete a Message while offline
- Message is deleted from the UI
- Message should re-appear after the request failed
- Turn on Network
- Hard Delete a Message
- Message is deleted from the UI
 
 **Test Case 4:** **(Test the scenarios with deleted visibility == `.alwaysVisible`)**
 - Turn on hard delete messages in Demo App Configuration
 - Change message delete visibility to `.alwaysVisible`
 - Repeat **Test Case 1**, **Test Case 2** and **Test Case 3**
 
 Test Case 4 is important to test to make sure that even with `config.deletedMessagesVisibility = .alwaysVisible` the hard deleted messages still fully disappear.

### ☑️ Checklist


Sources/StreamChat/Workers/MessageUpdater.swift
```diff
@@ -66,14 +69,26 @@ class MessageUpdater: Worker {
...
+
+                        let deletedMessage = try session.saveMessage(
+                            payload: response.message,
+                            for: ChannelId(cid: cid),
+                            syncOwnReactions: false
+                        )
+                        deletedMessage?.localMessageState = nil
+
+                        if hard, let message = deletedMessage {
+                            session.delete(message: message)
```
tbarbugli, line 87: what's the point of hard deleting from Core Data and marking messages as hard deleted?

tbarbugli: keep in mind that session.delete can cause crashes (we hold references to messages in a couple places)

Sources/StreamChatUI/MessageActionsPopup/ChatMessageActionsVC.swift
```diff
@@ -132,7 +132,7 @@ open class ChatMessageActionsVC: _ViewController, ThemeProvider {
                 self.alertsRouter.showMessageDeletionConfirmationAlert { confirmed in
                     guard confirmed else { return }
 
-                    self.messageController.deleteMessage { _ in
+                    self.messageController.deleteMessage() { _ in
```
tbarbugli, line 135: probably a good idea to add `hard` to the `messageController.deleteMessage` method

Sources/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -16,6 +16,7 @@ class MessageDTO: NSManagedObject {
     @NSManaged var createdAt: Date
     @NSManaged var updatedAt: Date
     @NSManaged var deletedAt: Date?
+    @NSManaged var isHardDeleted: Bool
```
tbarbugli, line 19: isHardDeleted is marked as optional in the schema

Sources/StreamChat/Database/StreamChatModel.xcdatamodeld/StreamChatModel.xcdatamodel/contents
```diff
@@ -163,6 +163,7 @@
         <attribute name="deletedAt" optional="YES" attributeType="Date" usesScalarValueType="NO"/>
         <attribute name="extraData" optional="YES" attributeType="Binary"/>
         <attribute name="id" attributeType="String"/>
+        <attribute name="isHardDeleted" optional="YES" attributeType="Boolean" defaultValueString="NO" usesScalarValueType="YES"/>
```
tbarbugli, line 166: maybe better to use the same definition we have for `isShadowed`?

Sources/StreamChat/Workers/MessageUpdater.swift
nuno-vieira: The reason the messages are also hard deleted from CoreData is that it is easier to manage. We also need the `isHardDelete` property to handle failed scenarios, to be able to retry the hard delete. We could just have `isHardDelete` and not actually delete it from Core Data but this would mean we needed to expose this property from the DTO to the model as well since we make the UI Logic as part of the Models. We have properties checking for `deletedAt != nil` and this will require changing this logic everywhere, adding unnecessary complexity. Deleting directly from CoreData makes it much easier to manage.

Sources/StreamChat/Database/DTOs/MessageDTO.swift
nuno-vieira: It is not optional, it has a default value of `false`

Sources/StreamChat/Database/StreamChatModel.xcdatamodeld/StreamChatModel.xcdatamodel/contents
nuno-vieira: Oh... just saw it says `optiona = YES` 🤔 Will change this 👍

Sources/StreamChat/Database/DatabaseSession.swift
```diff
@@ -399,13 +399,18 @@ extension DatabaseSession {
...
         let savedMessage = try saveMessage(
             payload: messagePayload,
             channelDTO: channelDTO,
             syncOwnReactions: false
         )
 
+        if payload.eventType == .messageDeleted && payload.hardDelete {
+            delete(message: savedMessage)
+            return
+        }
```
evsaev, line 412: Does it work fine for quote/thread replies when the parent message is deleted?

evsaev (approved):
Codewise looks good 👍 

I was a bit concerned about having 3 fields (`type, deletedAt: Date?, isHardDeleted: Bool`) where the last 2 do not make sense unless the message is deleted. On the other hand, refactoring this part will move us away from the JSON structure and complicate the JSON -> DTO -> Model flow, so I'm fine to keep it as it is.

Sources/StreamChat/Database/DatabaseSession.swift
nuno-vieira: What do you mean? 🤔 If the parent message is deleted, or if a reply is deleted, what happens?

DemoApp/AppConfigViewController/AppConfigViewController.swift
```diff
@@ -0,0 +1,274 @@
+//
+// Copyright © 2022 Stream.io Inc. All rights reserved.
+//
+
+import StreamChat
+import UIKit
+
+/// The Demo App Configuration.
+struct DemoAppConfig {
```
polqf, line 9: While this is really nice, I feel like it should have been in a separate PR. There's no need to split this one, but let's try do take it into account for the future 🙏

polqf (approved):

DemoApp/AppConfigViewController/AppConfigViewController.swift
nuno-vieira: Yes, I agree, but usually what happens is that these "optional" PRs usually do not get much attention if they are not absolutely necessary. And I wanted to make sure this goes in since without it, this PR wouldn't be possible to QA 👍

DemoApp/AppConfigViewController/AppConfigViewController.swift
```diff
@@ -0,0 +1,274 @@
...
+    /// A Boolean value to define if an additional hard delete message action will be added.
+    var isHardDeleteEnabled: Bool
+    /// A Boolean value to define if Atlantis will be started to proxy HTTP and WebSocket calls.
+    var isAtlantisEnabled: Bool
+}
+
+class AppConfig {
```
bielikb, line 16: mark classes as `final` by default :)

DemoApp/DemoAppCoordinator.swift
```diff
@@ -305,3 +308,45 @@ class HiddenChannelListVC: ChatChannelListVC {
...
+                    self.messageController.deleteMessage(hard: true) { _ in
+                        self.delegate?.chatMessageActionsVCDidFinish(self)
+                    }
+                }
+            },
+            appearance: appearance
+        )
+    }
+
+    public struct HardDeleteActionItem: ChatMessageActionItem {
```
bielikb, line 338: Does this need to be marked as public?

DemoApp/AppConfigViewController/SwitchButton.swift
```diff
@@ -0,0 +1,23 @@
...
+//
+
+import UIKit
+
+class SwitchButton: UISwitch {
+    var didChangeValue: ((Bool) -> Void)?
+
+    override init(frame: CGRect) {
+        super.init(frame: frame)
+        addTarget(self, action: #selector(didChangeValue(sender:)), for: .valueChanged)
```
bielikb, line 12: Won't this cause a memory leak?

bielikb (approved):
I did run QA + reviewed the code. The PR looks good, I left 1 question. the comment related to `final` classes actually goes beyond the scope of this PR. Ill address this in a separate piece of work.

DemoApp/AppConfigViewController/SwitchButton.swift
nuno-vieira: This in particular is not causing memory leaks, on the other hand, forgot to add `weak` on closures using this SwicthButton. Fixed that, run the memory graph and we don't have anymore memory leaks 👍 ✅

Issue #1541, opened 2021-10-12
photo would randomly become very low resolution


whiplash00:
## What did you do?
Send numerous photos; a small number of them will stay blurry and low resolution.

Sometimes, force-killing and reopening the app will turn other previously normal resolution photos back to low resolution.

## What did you expect to happen?
The photos should be of normal resolution.

## What happened instead?
Random photo stays low resolution.

## GetStream Environment
GetStream Chat version: 4.0.4
GetStream Chat frameworks: StreamChat, StreamChatUI, StreamChatClient, StreamChatCore
iOS version: 15.0
Swift version: 5.5
Xcode version: 12..5.1
Device: iPhone 12 pro max
![IMG_C48502D94140-1 2](https://user-images.githubusercontent.com/16630145/137012891-0479c299-74fe-45aa-8251-e52c9bf5855b.jpeg)

adamrushy:
Thank you for reporting @whiplash00 

I am currently investigating and will come back to you :]

kishinmanglani:
This happened once for us as well recently. Someone on our team thought it was because they posted in image on web and viewed on mobile.

nuno-vieira:
@whiplash00 @kishinmanglani Is this still reproducible on the most recent versions of the SDK? 

We did try to follow all the steps and still were not able to reproduce this issue.

Thank you for your help!

tbarbugli:
@kishinmanglani good news, we finally managed to reproduce this and have a good idea on how to fix this. We will post an update soon

polqf:
@kishinmanglani this issue should be fixed now, and will be part of the upcoming release. Thank you for your report!

polqf:
We just released 4.9.0, which contains the fix for this issue: https://github.com/GetStream/stream-chat-swift/releases/tag/4.9.0

PR #1747, opened 2022-01-03
[CIS-1475] Fix wrong image resolution when images are being quoted

polqf:
### 🔗 Issue Link
https://stream-io.atlassian.net/browse/CIS-1475

Fixes: https://github.com/GetStream/stream-chat-swift/issues/1541

### 🎯 Goal
It's been reported several times that we were randomly showing low resolution images. We recently discovered it was due to a combination of an image and a message quoting an image that this was reproducible. Our image cache was not properly distinguishing between the two, and in the event of the quoted message being downloaded before the full resolution image, it would use the cached quoted image in the full size version too.

### 🛠 Implementation
This PR changes the cacheKey to use the url with the size parameters, when possible. It will also keep the original image size if `preferredSize` is not passed, to make sure we don't have conflicting caching keys

### ☑️ Checklist

### DIFF
--- a/Sources/StreamChatUI/Utils/ImageLoading/NukeImageLoader.swift
+++ b/Sources/StreamChatUI/Utils/ImageLoading/NukeImageLoader.swift
@@ -101,12 +101,14 @@ open class NukeImageLoader: ImageLoading {
 
         let urlRequest = imageCDN.urlRequest(forImage: url)
         let size = preferredSize ?? imageView.bounds.size
+
         if resize && size != .zero {
             url = imageCDN.thumbnailURL(originalURL: url, preferredSize: size)
         }
+
         let cachingKey = imageCDN.cachingKey(forImage: url)
         let processors: [ImageProcessing] = resize
-            ? [ImageProcessors.LateResize(url: url, sizeProvider: { imageView.bounds.size })]
+            ? [ImageProcessors.LateResize(id: cachingKey, sizeProvider: { imageView.bounds.size })]
             : []
 
         let request = ImageRequest(
--- a/Sources/StreamChatUI/Utils/ImageLoading/NukeImageLoader.swift
+++ b/Sources/StreamChatUI/Utils/ImageLoading/NukeImageLoader.swift
@@ -100,8 +100,7 @@ open class NukeImageLoader: ImageLoading {
         }
 
         let urlRequest = imageCDN.urlRequest(forImage: url)
-        let cachingKey = imageCDN.cachingKey(forImage: url)
-        
+
         let processors: [ImageProcessing] = resize
             ? [ImageProcessors.LateResize(sizeProvider: { imageView.bounds.size })]
             : []
@@ -110,7 +109,8 @@ open class NukeImageLoader: ImageLoading {
         if resize && size != .zero {
             url = imageCDN.thumbnailURL(originalURL: url, preferredSize: size)
         }
-        
+        let cachingKey = imageCDN.cachingKey(forImage: url)
+
         let request = ImageRequest(
             urlRequest: urlRequest,
             processors: processors,
--- a/Sources/StreamChatUI/Utils/ImageLoading/NukeImageLoader.swift
+++ b/Sources/StreamChatUI/Utils/ImageLoading/NukeImageLoader.swift
@@ -93,30 +93,33 @@ open class NukeImageLoader: ImageLoading {
         completion: ((_ result: Result<UIImage, Error>) -> Void)? = nil
     ) -> Cancellable? {
         imageView.currentImageLoadingTask?.cancel()
-        
+
         guard var url = url else {
             imageView.image = placeholder
             return nil
         }
 
         let urlRequest = imageCDN.urlRequest(forImage: url)
-        let size = preferredSize ?? imageView.bounds.size
+        let size = preferredSize ?? .zero
+        let canResize = resize && size != .zero
 
-        if resize && size != .zero {
+        // If we don't have a valid size, we will not be able to resize using our CDN.
+        if canResize {
             url = imageCDN.thumbnailURL(originalURL: url, preferredSize: size)
         }
 
+        // At this point, if the image will be resized using our CDN, the url will contain the size values as
+        // parameters, and this will create a unique caching key for that url with this size. Only in this case we can
+        // successfully cache the resized image.
         let cachingKey = imageCDN.cachingKey(forImage: url)
-        let processors: [ImageProcessing] = resize
+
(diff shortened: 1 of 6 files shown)


Sources/StreamChatUI/Utils/ImageLoading/NukeImageLoader.swift
```diff
@@ -110,7 +109,8 @@ open class NukeImageLoader: ImageLoading {
         if resize && size != .zero {
             url = imageCDN.thumbnailURL(originalURL: url, preferredSize: size)
         }
-        
+        let cachingKey = imageCDN.cachingKey(forImage: url)
```
polqf, line 112: At this point, the url contains the size parameters if size is not zero, and `resize` is true

evsaev: Can we double-check we provide non-nil `preferred size` and `resize: true` in all the places?

bielikb (dismissed):

Sources/StreamChatUI/Utils/ImageLoading/NukeImageLoader.swift
tbarbugli: I think something is weird here (not necessarily related to this change). What if I send an image with an image attachment that is not hosted on the CDN?

You can reproduce this using https://github.com/GetStream/sdk-tester#test-messages-with-attached-images if needed

polqf: We are now passing `resize: true` everywhere.

But we are not passing `preferredSize` anywhere but where we have avatars. But in the case of not passing preferred size it would use the imageView.size, or it would store it using a caching key that has no parameters, and would not resize the image. Either way, we should be safe.

Where we are not passing preferredSize, is because we don't know its size (Name an uploaded image), and it should be kept like this. In most cases image is not resized because imageView.frame is zero.

polqf: @tbarbugli in the case of not being a url that's coming from our CDN, it should be keeping the whole url as caching key. And thumbnailURL returns the same url when the host is not our CDN.

What I just spotted based on your comment is that the url will be kept the same, but the image can be resized 🤦 , which puts us in the same situation for non CDN originated urls 🦆

nuno-vieira: @tbarbugli All images are loaded through this centralized function. So all images will be passed by the `imageCDN` AFAIK. 

@polqf This `resize: true` is very confusing. On the notion doc we should make sure we change this in the future to something like `aspectRatio` or `contentMode`. Or at least `clip: true` if on the backend we only support clipping.

nuno-vieira (approved):
LGTM! ✅

Sources/StreamChatUI/Utils/NukeImageProcessor.swift
```diff
@@ -63,11 +63,13 @@ extension ImageProcessors {
...
         }
 
+        private let url: URL
         private let sizeProvider: () -> CGSize
-        
+
         /// Initializes the processor with size providing closure.
         /// - Parameter sizeProvider: Closure to obtain size after the image is loaded.
-        public init(sizeProvider: @escaping () -> CGSize) {
+        public init(url: URL, sizeProvider: @escaping () -> CGSize) {
```
nuno-vieira, line 71: Can we add documentation to the new URL parameter?

nuno-vieira: Also, we changed the init of this component, how there are no changes in the rest of the codebase? 🤔

Sources/StreamChatUI/Utils/ImageLoading/NukeImageLoader.swift
polqf: @nuno-vieira what we mean by CDN here is "the real CDN", so the CDN that's serving the images server side.

Also, `resize: true` does exactly what it says now, upon downloading the image, it will resize the image, and use that one. What it's not clear from my side is if the resized image is the cached one (which I'm guessing it's the case).

Sources/StreamChatUI/Utils/NukeImageProcessor.swift
polqf: Maybe you missed things: https://github.com/GetStream/stream-chat-swift/pull/1747/commits/16b06b0b2110afd136330aa9f6455cd1060f78ba

I will add docu, yeah

nuno-vieira: I see... I thought this was the init from the `NukeImageLoader`, sometimes the diffing on GitHub is not clear 😄 

I didn't even know this LateResize was from our SDK... It is odd that this util is separate from the `ImageLoading` folder and I think it should even be part of the `NukeImageLoader` file. Anyways this is kinda a breaking change since it is a public init, and we are adding a new required argument 🤔 I'm sure it is not a big deal and nobody is using this, but in practice this is a breaking change :/

nuno-vieira (approved):

Sources/StreamChatUI/Utils/ImageLoading/NukeImageLoader.swift
tbarbugli: one cause of confusion here is that we are mixing different responsibilities:

- uploading an image and get its URL in return
- retrieving an image via URL
- refreshing an expire URL 
- apply some processing to an image (ie. resizing)
- caching an image based on its URL
- caching the result of image processing (when its done on the device)

Sources/StreamChatUI/Utils/NukeImageProcessor.swift
polqf: Good point! Fixed here: https://github.com/GetStream/stream-chat-swift/pull/1747/commits/7fc2d9c2864b3e658bdddfdea072f522b5ea73e2

Sources/StreamChatUI/Utils/ImageLoading/NukeImageLoader.swift
polqf: Resolving this conversation for the sake of readability given that the implementation has been changed after the recent changes.

I completely agree with you @tbarbugli wrt the mixing of responsibilities. This is agreed to be taken care of in the upcoming work that is being debated in Notion (https://www.notion.so/stream-chat-fe-sdk/Image-Loading-375fcf50af6f4753964cb574e0d2bf6d). I'm adding your comment there for the sake of not losing all the findings.

adamrushy (approved):
Looks good, just a CHANGELOG entry required ✅

PR #1752, opened 2022-01-10
[CIS-1288] Fix Message List NSInternalInconsistencyException Crash

nuno-vieira:
### 🔗 Issue Link
CIS-1288

### 🎯 Goal
Fix a crash that happened when switching channels very quickly, and when the messages were not prefetched from a channel list.

### 🛠 Implementation
Call `reloadData()` when the data source of the message list is changed. When switching channels in quick succession it could happen that there were conflicts in the message list with the old data.

### 🎨 Changes
N/A

### 🧪 Testing
Steps to reproduce this:
1. Apply this patch: [dontSaveChannelListMessages.patch.zip](https://github.com/GetStream/stream-chat-swift/files/7841000/dontSaveChannelListMessages.patch.zip)
2. Open the demo app with an iPad (Easier to quickly switch channels)
3. Quickly switch channels
4. Observe there are no crashes.

In order to reproduce the crash, make sure to revert this PR and delete the `reloadData()` from here: [DemoAppCoordinatior.swift#L167](https://github.com/GetStream/stream-chat-swift/pull/1752/files#diff-55ec0ba42c0fa93f6c625aa3f8b251ebbe2d462a043f460d84ed7635df61bb46L167) (This is why it was really hard to reproduce this issue because we were fixing it on the demo app).

### ☑️ Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### 🐞 Fixed
 - Fix wrong image resolution when images are being quoted [#1747](https://github.com/GetStream/stream-chat-swift/pull/1747)
+- Fix message list NSInternalInconsistencyException crash [#1752](https://github.com/GetStream/stream-chat-swift/pull/1752)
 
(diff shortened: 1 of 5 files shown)


bielikb (approved):
LGTM 🚢

PR #1753, opened 2022-01-11
[CIS-1489] Fix Image and Video sharing behaviour

nuno-vieira:
### 🔗 Issue Link
CIS-1489

### 🎯 Goal
When sharing a video or image, it should not share an URL, but the asset itself.

### 🛠 Implementation
When it is an image, we share the actual `UIImage`. When it is a video, we share the actual video file.

### 🎨 Changes

https://user-images.githubusercontent.com/12814114/149010574-930e9d90-3eac-4235-ac04-1e257b279286.MP4

### 🧪 Testing
- Tap on a video/image
- Tap share
- Click on "mail"
- The video/image should appear on the content of the email instead of an URL

### ☑️ Checklist

### DIFF
--- a/Sources/StreamChatUI/Gallery/GalleryVC.swift
+++ b/Sources/StreamChatUI/Gallery/GalleryVC.swift
@@ -412,8 +412,8 @@ open class GalleryVC:
                 return nil
             }
 
-            let fileName = itemAttachment.id.messageId.lowercased()
-            let filePath = NSTemporaryDirectory().appending("\(fileName).mov")
+            let fileName = itemAttachment.payload.title ?? itemAttachment.id.messageId.lowercased() + ".mp4"
+            let filePath = NSTemporaryDirectory().appending("\(fileName)")
             let url = URL(fileURLWithPath: filePath)
             do {
                 try urlData.write(to: url)
--- a/Sources/StreamChatUI/Gallery/GalleryVC.swift
+++ b/Sources/StreamChatUI/Gallery/GalleryVC.swift
@@ -400,12 +400,28 @@ open class GalleryVC:
     /// - Returns: An item to share.
     open func shareItem(at indexPath: IndexPath) -> Any? {
         guard let item = getItem(at: indexPath) else { return nil }
-        
-        if let image = item.attachment(payloadType: ImageAttachmentPayload.self) {
-            return image.imageURL
-        } else if let video = item.attachment(payloadType: VideoAttachmentPayload.self) {
-            return video.videoURL
-        } else {
+
+        switch item.type {
+        case .image:
+            let cell = attachmentsCollectionView
+                .cellForItem(at: indexPath) as? ImageAttachmentGalleryCell
+            return cell?.imageView.image
+        case .video:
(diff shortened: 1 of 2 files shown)


Sources/StreamChatUI/Gallery/GalleryVC.swift
```diff
@@ -400,12 +400,28 @@ open class GalleryVC:
...
+                .cellForItem(at: indexPath) as? ImageAttachmentGalleryCell
+            return cell?.imageView.image
+        case .video:
+            guard let itemAttachment = item.attachment(payloadType: VideoAttachmentPayload.self),
+                  let urlData = try? Data(contentsOf: itemAttachment.videoURL) else {
+                return nil
+            }
+
+            let fileName = itemAttachment.id.messageId.lowercased()
+            let filePath = NSTemporaryDirectory().appending("\(fileName).mov")
```
bielikb, line 416: @nuno-vieira Why are we locking to `.mov` format only? Don't we get mime type/extension along with the remote resource?

Android leverages eg `.mp4` format as seen eg [here](https://github.com/GetStream/stream-chat-android/blob/f44d0a5e5b2c324ea7ff8868875e09dda7c3cc0e/stream-chat-android/src/main/java/com/getstream/sdk/chat/utils/CaptureController.java#L35).

Sources/StreamChatUI/Gallery/GalleryVC.swift
```diff
@@ -400,12 +400,28 @@ open class GalleryVC:
...
+            let url = URL(fileURLWithPath: filePath)
+            do {
+                try urlData.write(to: url)
```
bielikb, line 419: @nuno-vieira 
This will cause incremental increase of the app size everytime the end user tries to share new video via our chat. We should make sure, to clean up the storage once the video is shared/eg when activity controller is dismissed.

bielikb: Here are the docs for `UIActivityViewController`s [completionWithItemsHandler](https://developer.apple.com/documentation/uikit/uiactivityviewcontroller/1622022-completionwithitemshandler).

Sources/StreamChatUI/Gallery/GalleryVC.swift
nuno-vieira: Makes sense 👍 I will parse the file extension type nice catch!

Sources/StreamChatUI/Gallery/GalleryVC.swift
nuno-vieira: When the user closes the app the files will be removed, since we are using NSTemporaryDirectory. If you are talking about the memory while the app is being used, I think it does not make sense that if we keep sharing the same video multiple times that it will always take some time to process it but maybe this is a rare use case 🤔 What if we clean this only when we receive a memory warning ?

bielikb (approved):

b-onc:
Tapping "share" button and seeing the sharing sheet has 2 seconds delay. Many apps on iOS has that, but is there a way to escape that?

b-onc (approved):

PR #1768, opened 2022-01-26
[CIS-1254] Fix Composer not showing files when >3 files are selected

b-onc:
### 🔗 Issue Link
CIS-1254

### 🎯 Goal

The composer attachment view should be flexible in height, it should show up to 3 attachments and then user should scroll to see more.

#### Issue description

When the files are added as attachments, they're added to `composerVC.content` one by one, no matter how many files are selected. This causes `AttachmentsPreviewVC.setupVerticalStackView` to be called N times for N files added.
The current implementation assumes that `AttachmentsPreviewVC.scrollView` will have its correct size as soon as content is updated, but it's not true: one layout loop is required until `scrollView` has sized itself.

Here's a table showing expected/actual behavior:

Case | Expected | Actual
-- | -- | --
Add 1 file, and then 3 files | `scrollView` sized itself so 3 are visible, scrolling shows the 4th file | `scrollView` shows a buggy view, files are rendered on top of each other
Add 2 files, and then 2 files | `scrollView` sized itself so 3 are visible, scrolling shows the 4th file | `scrollView` shows 2 files, scrolling shows last 2 files
Add 3 files, and then 1 file | `scrollView` sized itself so 3 are visible, scrolling shows the 4th file | `scrollView` shows 3 files, scrolling shows the last file
Add 4 files at the same time | `scrollView` sized itself so 3 are visible, scrolling shows the 4th file | `scrollView` doesn't show at all

The reason for the behavior is, when user adds N (<3) files, `scrollView` sizes itself correctly. After this, when user adds M files (where N+M>3), `scrollView` freezes its size and doesn't allow height to grow anymore, enabling scroll behavior. This works for N,M = 2,2 and 3,1 pairs, but not for 1,3 (causing views to be rendered on top of each other) and 4,0 (not showing at all). For 4,0 `scrollView` freezes it's current size, which is 0, and doesn't show at all.

#### Tried solutions

1. Force `scrollView` to layout itself so it has the correct size
Using `scrollView.setNeedsLayout()` and `scrollView.layoutIfNeeded()`, I hoped `scrollView` would have correct height, so the constraint would have the correct value. This didn't work.
1. Add a constraint to `attachmentViews.first.heightAnchor * 3`
This didn't work, since by the time constraint are evaluated, `attachmentViews.first` had a height of 0
1. Use `attachmentViews.first.intrinsicContentSize.height`
The intrinsic height was -1 (AutoLayout decides)
1. Get a file's height by using `FileAttachmentView().height`
Not cool, better to use the last solution
1. Some breaking changes to make this work
We want to avoid breaking changes, so I haven't looked much into this. We can add `preferredHeight` to `AttachmentPreviewProvider` and use it to calculate the desired height, for example.
1. Use a magic number for the height
This was the last, and ugliest solution, but it's the cleanest and it works for all cases. Makes the code simpler too. I choose 54, since it's the same height as `FileAttachmentView.height`, and also it shows 2.5 files, prompting user to scroll for more.

### 🎨 Changes

Screen Shot 2022-01-26 at 14 50 44

### 🧪 Testing

Snapshot tests are added.

Launch a chat, add 4 files to composer, in any order you want.

### ☑️ Checklist


Sources/StreamChatUI/CommonViews/Attachments/AttachmentsPreviewVC.swift
```diff
@@ -115,25 +120,6 @@ open class AttachmentsPreviewVC: _ViewController, ComponentsProvider {
...
-        scrollView.isScrollEnabled = content.count > maxNumberOfVerticalItems
-
-        // If the content is bigger than the max vertical items and the scroll view height
-        // constraint is not yet created, append to the vertical constraint and activate it.
-        if content.count > maxNumberOfVerticalItems {
-            if scrollViewHeightConstraint == nil {
-                scrollViewHeightConstraint = scrollView.heightAnchor.pin(
-                    lessThanOrEqualToConstant: scrollView.frame.size.height
-                )
-                scrollViewHeightConstraint?.isActive = true
```
nuno-vieira, line 128: If we are not using `scrollViewHeightConstraint` anymore we should remove this property. Otherwise, this is dead code. If we are not removing it just because of breaking changes, it doesn't make sense because it is still a breaking change, but a silent one, which is worse than having a normal breaking change.

nuno-vieira, line 128: The `setupVerticialStackView()` is always called on the update content, so why don't we always update the constant value of the scrollViewHeightConstraint based on the current content? I'm really not convinced with the current solution since besides causing a breaking change, it is not very flexible. At least we need to extract the hard coded "54" to an open variable so that is easily customizable, otherwise, the customer needs to add additional constraints for this. Also, Visually we can also break the UI for some customers if they have already attachments views that have a bigger height, and I know some customers who have this, so with the current solution not only we will break the API but also the UI, so we need to come up with something better, even if that involves some workarounds.

nuno-vieira (changes requested):
Requesting changes mainly for these reasons:
- Current solution breaks the API
- Current solution will break the UI for some customers
- Looking into the snapshots it looks like the attachment preview is not working properly for the Large Text trait, we should try to make the text shrink, or the attachment preview height should grow, which with the current approach is not possible because it has a fixed height.

nuno-vieira:
@b-onc  This one is indeed tricky. But I think I have a working solution, at least from my testing it is working in all scenarios, let me know what you think: 

https://github.com/GetStream/stream-chat-swift/commit/e461865e9094c0fea05c91369826d235b3043fb5

Basically, I moved the scroll view height calculation to the `viewDidLayoutSubviews` lifecycle, and the `scrollViewHeight` is based on the number of items.

b-onc:
@nuno-vieira Thanks for the points and discussion! I went back and reviewed https://developer.apple.com/videos/play/wwdc2018/220/ and found a solution: `systemLayoutSizeFitting` to the rescue. Now it works with no breaking changes, and the resulting code is much closer to what we currently have.

I'm not handling the "Attachments are not resizing correctly" issue in this PR to keep the scope limited, I'm creating a ticket for it.

Sources/StreamChatUI/CommonViews/Attachments/AttachmentsPreviewVC.swift
```diff
@@ -117,13 +117,21 @@ open class AttachmentsPreviewVC: _ViewController, ComponentsProvider {
...
         // If the content is bigger than the max vertical items and the scroll view height
         // constraint is not yet created, append to the vertical constraint and activate it.
-        if content.count > maxNumberOfVerticalItems {
+        if content.count > maxNumberOfVerticalItems, let firstAttachmentView = attachmentViews.first {
             if scrollViewHeightConstraint == nil {
+                let attachmentHeight = firstAttachmentView
+                    .systemLayoutSizeFitting(.init(width: CGFloat.infinity, height: CGFloat.infinity))
+                    .height
+                let maxScrollViewHeight: CGFloat = CGFloat(maxNumberOfVerticalItems) * attachmentHeight +
+                    CGFloat(attachmentViews.count + 1) * verticalStackView.spacing.rawValue
```
nuno-vieira, line 131: Nit: This will make the code a little bit tidier and also clearer
```suggestion
                let attachmentHeight = firstAttachmentView
                    .systemLayoutSizeFitting(.init(width: CGFloat.infinity, height: CGFloat.infinity))
                    .height
                let spacingSize = CGFloat(attachmentViews.count + 1) * verticalStackView.spacing.rawValue
                let maxScrollViewHeight: CGFloat = CGFloat(maxNumberOfVerticalItems) * attachmentHeight + spacingSize
```

nuno-vieira (approved):
LGTM! ✅ Left just one comment that I think it will make a the code a little easier to understand 👍

PR #1770, opened 2022-01-28
[CIS-1288] Add Support for Message List Data Source Diffing (Experimental)

nuno-vieira:
### 🔗 Issue Link
CIS-1288

### 🎯 Goal
Add support for message list data source diffing to improve the stability of the message list, this replaces the usage of  `performBatchUpdates`.

### 🛠 Implementation
The proposed implementation creates a diffable data source snapshot from the list of changes reported by core data. We can't use the `NSFetchResultsController.didChangeContentWithSnapshot()` API because of our inverted table view, so we need to do some manual work for the diffing snapshots to works.

Only on iOS 13+ the diffing mechanism will be applied. And all customizations of the customer will still be applied since everything is backwards compatible.

**Usage:**
To enable this, you need to set to `true` the following flag: `Components.default._messageListDiffingEnabled = true`. 

### 🎨 Changes
N/A

### 🧪 Testing
Full regression QA Test should be done to make sure everything is working as before.

### ☑️ Checklist


nuno-vieira (commented):
~QA Failed: Currently it is crashing on iPad Split View (Looking into it)~

QA is passing ✅  Crashes have been solved!

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -533,3 +606,103 @@ open class ChatMessageListVC:
...
+            .sorted(by: { $0.row < $1.row })
+            .map(\.0)
+
+        if hasNewInsertions, let currentFirstMessage = snapshot.itemIdentifiers.first {
+            // Insert new messages at the bottom.
+            snapshot.insertItems(sortedInsertedMessages, beforeItem: currentFirstMessage)
+        } else if let currentLastMessage = snapshot.itemIdentifiers.last {
+            // Load new messages at the top.
+            snapshot.insertItems(sortedInsertedMessages, afterItem: currentLastMessage)
+        }
```
evsaev, line 671: This may be not the case but I have a suspicion that if the channel is empty, message insertions will never happen because of `snapshot.itemIdentifiers` being empty. Would it be possible to check it?

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -533,3 +606,103 @@ open class ChatMessageListVC:
...
+        // is the same as calling `reloadData()`. Info: https://developer.apple.com/videos/play/wwdc2021/10252/?time=158
+        UIView.performWithoutAnimation {
+            diffableDataSource?.apply(snapshot, animatingDifferences: true) { [weak self] in
+                let newestMessage = snapshot.itemIdentifiers.first
+                if hasNewInsertions && newestMessage?.isSentByCurrentUser == true {
+                    self?.listView.scrollToMostRecentMessage()
+                }
+
+                // When new message is inserted, update the previous message to hide the timestamp if needed.
+                if hasNewInsertions, let previousMessage = snapshot.itemIdentifiers[safe: 1] {
```
evsaev, line 692: The previous cell also needs to be reloaded when the message is deleted

Sources/StreamChatUI/ChatMessageList/ChatMessageListVCDataSource.swift
```diff
@@ -7,6 +7,9 @@ import StreamChat
 
 /// The object that acts as the data source of the message list.
 public protocol ChatMessageListVCDataSource: AnyObject {
+    /// Asks the data source to return all the available messages.
+    var messages: [ChatMessage] { get }
+
```
evsaev, line 12: I find this change useful in terms of data source protocol consistency but are we using it somewhere in diffing logic?

evsaev (commented):
That's an amazing piece of work 👏 Left some comments, mostly about the edge cases

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -153,6 +171,19 @@ open class ChatMessageListVC:
         listView.backgroundColor = appearance.colorPalette.background
     }
 
+    override open func updateContent() {
+        super.updateContent()
+
+        listView.delegate = self
```
evsaev, line 177: nit: it's fine to have it here just curious if there was a strong reason to move the delegate assignment outside list view builder closure?

tbarbugli (commented):
Why do we require `Components.default.messageListDiffingEnabled = true`? Unless we see valid use-cases to keep the previous approach, this only makes sense if:

- we consider this an experimental feature 
- we are going to have enough developers opting in for this and willing to share their experience
- we have a clear plan to measure success and decide when its the time to flip the witch to false (if we don't we are simply postponing)

Did you do some benchmarking / performance check?

Why don't we embed the diffing in the SDK? Your approach has a few problems:

- it needs to support diff and no-diff (complexity is higher, more testing is needed, ...)
- iOS 12 is recent enough and will not get this fix

nuno-vieira:
> Why do we require `Components.default.messageListDiffingEnabled = true`? Unless we see valid use-cases to keep the previous approach, this only makes sense if:
> 
> * we consider this an experimental feature
> * we are going to have enough developers opting in for this and willing to share their experience
> * we have a clear plan to measure success and decide when its the time to flip the witch to false (if we don't we are simply postponing)
> 
> Did you do some benchmarking / performance check?
> 
> Why don't we embed the diffing in the SDK? Your approach has a few problems:
> 
> * it needs to support diff and no-diff (complexity is higher, more testing is needed, ...)
> * iOS 12 is recent enough and will not get this fix

- Yes we already have a customer that will test drive this and will share the results
- And yes if results are positive we plan on moving to `true` as the default
- The benchmarking and performance as not yet been performed because it is not the goal of this PR, but one thing that I was going to still do it to run our performance load script (meanwhile I was on support, so this PR was suspended)
- What do you mean by "Why don't we embed the diffing in the SDK?" ?
- We don't have enough data currently to measure the impact on iOS 12 but I'm pretty sure iOS 12 is currently minimal.
- It is indeed a con of this approach, it makes it hard to test, and iOS 12 is kinda ignored, but it is also the fastest one and uses Apple's native solution. We wrote a document explaining the pros/cons of this approach.

We are going to perform more tests, benchmarking and soon a Team QA so that we can gather more data and make a decision if this is the right thing to do 👍

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
nuno-vieira: Fixed ✅

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
nuno-vieira: This is fixed now too 👍 ✅ Although I which this was not required by our UX. It adds unnecessary complexity if we had a different design

Sources/StreamChatUI/ChatMessageList/ChatMessageListVCDataSource.swift
nuno-vieira: This is needed when adding the initial data. Because we open a channel with the local data already, we don't get notified of all the messages available through the `updateMessages` event. So we need to get the messages from the channel controller. 

This is used here: https://github.com/GetStream/stream-chat-swift/pull/1770/files/98fc37e0e04f6954a1f7ff4c139e3ef367b41dff#diff-30faf0001fbdda5a68aa7eca1fe1d37eeef9053d44f81afc1c6365ed66f44869R630

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -533,3 +606,103 @@ open class ChatMessageListVC:
...
+
+        /// Populate the Initial messages data.
+        var snapshot = NSDiffableDataSourceSnapshot<Int, ChatMessage>()
+        snapshot.appendSections([0])
+        snapshot.appendItems(dataSource?.messages ?? [], toSection: 0)
+        diffableDataSource.apply(snapshot, animatingDifferences: false)
+    }
+
+    /// Transforms an array of changes to a diffable data source snapshot.
+    func updateMessagesSnapshot(with changes: [ListChange<ChatMessage>], completion: (() -> Void)?) {
```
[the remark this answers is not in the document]
nuno-vieira: Unfortunately, if we have inconsistent updates it will still crash. The hope here is that applying the snapshots has better robustness in terms of concurrency.

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -533,3 +606,103 @@ open class ChatMessageListVC:
...
+        movedMessages.forEach {
+            snapshot.moveItem($0.from, afterItem: $0.to)
+            snapshot.reloadItems([$0.from, $0.to])
+        }
+
+        // The reason we call `performWithoutAnimation` and `animatingDifferences: true` at the same time
+        // is because we don't want animations, but on iOS 14 calling `animatingDifferences: false`
+        // is the same as calling `reloadData()`. Info: https://developer.apple.com/videos/play/wwdc2021/10252/?time=158
+        UIView.performWithoutAnimation {
+            diffableDataSource?.apply(snapshot, animatingDifferences: true) { [weak self] in
```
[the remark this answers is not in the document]
nuno-vieira: As mentioned above, the `apply()` already handles this, all updates are done in a serial queue, to make sure there are no concurrency issues.

DemoApp/DemoAppCoordinator.swift
```diff
@@ -152,24 +153,24 @@ final class DemoAppCoordinator: NSObject, UNUserNotificationCenterDelegate {
...
                 channelListQuery: channelListVC.controller.query
             )
+            channelVC.channelController = channelController
+            return UINavigationController(rootViewController: channelVC)
         }
 
-        let channelVC = CustomChannelVC()
-        channelVC.channelController = makeChannelController("unknown")
+        let splitController = UISplitViewController()
+        splitController.viewControllers = [channelListVC, makeChannelVC("unknown")]
```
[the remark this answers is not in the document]
nuno-vieira: Fixed 👍

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
nuno-vieira: There wasn't, it was to make the code more cleaner and tidy :)

Sources/StreamChatUI/Components.swift
```diff
@@ -116,6 +116,11 @@ public struct Components {
     /// A boolean value that determines wether date separators should be shown between each message.
     public var messageListDateSeparatorEnabled = false
 
+    /// A boolean value that determines wether the message list should use a diffable data source.
+    /// Note: This is currently an experimental feature that we are actively
+    /// working on and testing to make sure it is 100% stable.
```
evsaev, line 121:
```suggestion
    /// A boolean value that determines whether the message list should use a diffable data source.
    /// Note: This is currently an experimental feature that we are actively
    /// working on and testing to make sure it is stable.
```

evsaev (approved):
LGTM 🚢

adamrushy (approved):
LGTM ✅

PR #1773, opened 2022-02-01
Updated the docs for presenting channels

martinmitrevski:
### 🔗 Issue Link
[JIRA](https://stream-io.atlassian.net/browse/SWUI-72?atlOrigin=eyJpIjoiYzQ3YzQ5MzZjMDI3NGM2NWIzZDU1MmU5ZTFlMWQxMmIiLCJwIjoiaiJ9)

### 🎯 Goal

Update the docs to reflect the latest changes in the SwiftUI SDK.

### 🛠 Implementation

Updated docusaurus docs.

### 🧪 Testing

Run docusaurus locally.

### ☑️ Checklist

### DIFF
--- a/docusaurus/docs/iOS/swiftui/components/helper-views.md
+++ b/docusaurus/docs/iOS/swiftui/components/helper-views.md
@@ -61,9 +61,9 @@ public func makeChannelListItem(
     avatar: UIImage,
     onlineIndicatorShown: Bool,
     disabled: Bool,
-    selectedChannel: Binding<ChatChannel?>,
+    selectedChannel: Binding<ChannelSelectionInfo?>,
     swipedChannelId: Binding<String?>,
-    channelDestination: @escaping (ChatChannel) -> ChannelDestination,
+    channelDestination: @escaping (ChannelSelectionInfo) -> ChannelDestination,
     onItemTap: @escaping (ChatChannel) -> Void,
     trailingSwipeRightButtonTapped: @escaping (ChatChannel) -> Void,
     trailingSwipeLeftButtonTapped: @escaping (ChatChannel) -> Void,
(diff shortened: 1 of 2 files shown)


docusaurus/docs/iOS/swiftui/components/helper-views.md
```diff
@@ -94,7 +94,7 @@ In the channel list item creation method, you are provided with several paramete
 - `avatar`: the avatar of the channel.
 - `onlineIndicatorShown`: whether the online indicator (about last active members) is shown on the avatar.
 - `disabled`: whether the user interactions with the channel are disabled. You should use this value while the view is being swiped, in order to avoid clicking the channel list item instead.
-- `selectedChannel`: binding of the currently selected channel.
+- `selectedChannel`: binding of the currently selected channel selection info (channel and optional message).
```
nuno-vieira, line 97: Maybe this is better like this? **selected channel selection info** sounds a bit weird and repetitive  🤔 
```suggestion
- `selectedChannel`: binding of the current channel selection info (channel and optional message).
```

nuno-vieira, line 97: Shouldn't the property be renamed as well?
```suggestion
- `channelSelectionInfo`: binding of the currently selected channel selection info (channel and optional message).
```

docusaurus/docs/iOS/swiftui/components/list-tap-events.md
```diff
@@ -31,9 +31,9 @@ class CustomFactory: ViewFactory {
     
     public static let shared = CustomFactory()
     
-    func makeChannelDestination() -> (ChatChannel) -> CustomChannelDestination {
-        { channel in
-            CustomChannelDestination(channel: channel)
+    func makeChannelDestination() -> (ChannelSelectionInfo) -> CustomChannelDestination {
+        { selectionInfo in
+            CustomChannelDestination(channel: selectionInfo.channel)
```
nuno-vieira, line 36: This actually surprising in my view, it feels the older a approach made more sense. You would give the Channel Info, and get back a destination, why only the selection matters now? 🤔 This is just out of curiosity

martinmitrevski: We need an optional message too. Basically, when you perform search on messages from the channel list, they are pairs of (channel, message). When you select a search result, you should scroll up to the message in the message list. It will also be needed for deeplinking.

nuno-vieira: Then, wouldn't it make more sense to have `ChannelInfo` where it contains the `channel` and `selectedMessage`? From what I understood the selection info is actually from the message and not the channel itself?

martinmitrevski: It's already like that. The selection info comes from the channel, and the message is optional parameter that, if present, makes the message list scroll to the provided message.

nuno-vieira: Ah I see... I understood it now

nuno-vieira (approved):
LGTM! ✅

PR #1776, opened 2022-02-02
Offline Support - Recovery Flow (Part 2 - Tests) 

polqf:
### 🔗 Issue Link
https://stream-io.atlassian.net/browse/CIS-1365

### 🎯 Goal

### 🛠 Implementation

_Provide a description of the implementation._

### 🎨 Changes

_Add relevant screenshots or videos showcasing the changes._

### 🧪 Testing

_Describe the steps how this change can be tested (or why it can't be tested)._

### ☑️ Checklist


martinmitrevski (commented):
Great work, a lot of tests and coverage. I also like the refactoring in the `SyncRepository` and the operations creations and the `Spy` stuff. 🥇 
Things I think we should improve:
- code repetition in the tests
- start using given/when/then
- some other smaller remarks (see comments)

Sources/StreamChat/Repositories/Operations/SyncOperations_Tests.swift
```diff
@@ -0,0 +1,251 @@
...
+        let context = SyncContext()
+        let operation = GetChannelIdsOperation(database: database, context: context)
+        operation.startAndWaitForCompletion()
+        
+        XCTAssertEqual(context.localChannelIds.count, 1)
+    }
+
+    // MARK: - GetPendingConnectionDateOperation
+
+    func test_GetPendingConnectionDateOperation_noUser() {
```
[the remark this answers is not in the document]
polqf: We have the intention to use block based syntax, so any work I do to get it closer to that will be trashed.

Sources/StreamChat/Repositories/Operations/SyncOperations_Tests.swift
```diff
@@ -0,0 +1,251 @@
...
+        context.lastPendingConnectionDate = nil
+        let operation = SyncEventsOperation(database: database, syncRepository: syncRepository, context: context)
+
+        operation.startAndWaitForCompletion()
+
+        XCTAssertEqual(context.synchedChannelIds.count, 0)
+        XCTAssertTrue("syncMissingEvents(using:channelIds:bumpLastSync:completion:)".wasNotCalled(on: syncRepository))
+    }
+
+    func test_SyncEventsOperation_pendingDate_syncFailure_shouldRetry2times() throws {
```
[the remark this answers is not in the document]
polqf: I think this could make it less readable. Skipping this for now

martinmitrevski (approved):
LGTM 👍

Sources/StreamChat/Repositories/Operations/SyncOperations.swift
```diff
@@ -0,0 +1,140 @@
...
+                return
+            }
+
+            // Reset only if it needs recovery
+            if let cid = controller.cid, context.synchedChannelIds.contains(cid) {
+                done(.continue)
+                return
+            }
+
+            let cidString = (controller.cid?.rawValue ?? "unknown")
```
adamrushy, line 93: Do we want to fall back to "unknown" here or is a missing cid worthy of an error?

adamrushy: I guess it will fall through but still make a query with unknown?

adamrushy (approved):
Looks solid overall, adding a comment but it's more a question :)

StreamChat.xcodeproj/project.pbxproj
```diff
@@ -5815,6 +5834,8 @@
 			children = (
 				C1E8AD5A278DDEBB0041B775 /* Operations */,
 				C1E8AD55278C8A440041B775 /* SyncRepository.swift */,
+				C186BFA927AA979B0099CCA6 /* SyncRepository_Tests.swift */,
+				C186BFA227A048690099CCA6 /* SyncRepository_Mock.swift */,
```
nuno-vieira, line 5838: I think that we could improve the folder structure. Right now we have something like this:

```bash
Repositories/
|——— SyncRepository.swift
|——— Operations/
|	|—— AsyncOperation.swift
|	|—— SyncOperation.swift
```

I would propose something like this:
```bash
Repositories/
 |——— SyncRepository/
 |    |——— SyncRepository.swift
 |    |——— SyncOperation.swift
```
Then the `AsyncOperation.swift` would live in the Utils or any top level Common folder (Currently the utils folder is totally crowded and there are things that shouldn't be there but anyhow..).

The reason behind this is because the Operations is an implementation detail of the `SyncRepository`. Operations can totally be used outside of the `Repositories` context, hence the `AsynOperation.swift` should live in a common place. By having an `Operations` folder in the `Repositories` folder, it gives an impression that each repository will have an operation, which is not true, since like I said before, it is an implementation detail, that is why I believe the `SyncOperatior` should live inside the `SyncRepository` context.

Sources/StreamChat/Repositories/Operations/SyncOperations_Tests.swift
nuno-vieira, line 94: nit: Using values on the names of the tests usually make them outdated pretty fast, in this case, it is already from what I can see the assertion it says "3" times instead of "2"
```suggestion
    func test_SyncEventsOperation_pendingDate_syncFailure_shouldRetry() throws {
```

Sources/StreamChatTestTools/Spy.swift
```diff
@@ -0,0 +1,37 @@
+//
+// Copyright © 2022 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+public protocol Spy: AnyObject {
+    var recordedFunctions: [String] { get set }
+}
```
nuno-vieira, line 9: My only concern with this is that we are adding yet another API to how we write tests and mocks. I usually defend using the native XCTest tools as much as we can or use a proper mocking library in this case. It doesn't feel to me this is adding much value in comparison to just adding a property `xxxxCallCount = 0` on a mock and then simply calling `xxxxCallCount += 1` on the desired function. 🤔

nuno-vieira (approved):
LGTM! ✅ The operations refactor was much needed 👍 

Added a couple of comments but they are not blockers by any means.

Sources/StreamChat/Repositories/Operations/SyncOperations.swift
polqf: A cid is not needed here (AKA it might not be there yet), this `cidString` is only for debug purposes

Sources/StreamChatTestTools/Spy.swift
polqf: It is true that it is yet another way of doing it, but it can much better scale than having one xxxxCallCount per each mock. Don't you think?

PR #1781, opened 2022-02-04
[CIS-1554] Make `SendButton` animation overridable

nuno-vieira:
### 🔗 Issue Link
CIS-1554

### 🎯 Goal
Currently, the animation of SendButton is done on the didSet observer of the `SendButton.isEnabled`, the problem with this is that the `didSet` implementation is not overridable. We need to extract the animation to a different function so that the customer can override it.

### 🛠 Implementation
Extract the animation to a different function so that the customer can override it.

### 🎨 Changes
N/A

### 🧪 Testing
N/A

### ☑️ Checklist

### DIFF
--- a/Sources/StreamChatUI/CommonViews/SendButton/SendButton.swift
+++ b/Sources/StreamChatUI/CommonViews/SendButton/SendButton.swift
@@ -10,7 +10,7 @@ open class SendButton: _Button, AppearanceProvider {
     /// Override this variable to enable custom behavior upon button enabled.
     override open var isEnabled: Bool {
         didSet {
-            isEnabledChangeAnimation()
+            isEnabledChangeAnimation(isEnabled)
         }
     }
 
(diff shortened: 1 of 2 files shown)


Sources/StreamChatUI/CommonViews/SendButton/SendButton.swift
```diff
@@ -28,4 +24,13 @@ open class SendButton: _Button, AppearanceProvider {
         let disabledStateImage = appearance.images.sendArrow.tinted(with: buttonColor)
         setImage(disabledStateImage, for: .disabled)
     }
+
+    /// The animation when the `isEnabled` state changes.
+    open func isEnabledChangeAnimation() {
```
adamrushy, line 29: Would it make more sense for this public function to pass in the `isEnabled` param? Approving anyway more of a small suggestion :)

adamrushy (approved):

Sources/StreamChatUI/CommonViews/SendButton/SendButton.swift
nuno-vieira: I see.. yes it kinda makes sense 👍

PR #1784, opened 2022-02-08
[CIS-1393] Fix crash on iOS 12 when local storage enabled

nuno-vieira:
### 🔗 Issue Link
CIS-1393

### 🎯 Goal
Fix a crash on iOS 12 that whenever the app was launched, it would instantly crash.

### 🛠 Implementation

The crash was happening because of a mismatch between `UserDTO.teams` property marked as `[String]?` in db schema but being a `Set<String>?` in managed object subclass. Exactly what the error was saying: 
```
reason: 'Object of class Swift.__EmptySetSingleton is not among allowed top level class list (
  NSArray,
  NSDictionary,
  NSString,
  NSNumber,
  NSDate,
  NSData,
  NSURL,
  NSUUID,
  NSNull
)'
```
The `teams` property is updated to be an array, made non-optional, and ensured to exist at the moment the `UserDTO` is saved to the persistent store.

### 🎨 Changes
N/A

### 🧪 Testing
- Run Slack Clone App on iOS 12

### ☑️ Checklist

### DIFF
--- a/Sources/StreamChat/Database/DTOs/UserDTO.swift
+++ b/Sources/StreamChat/Database/DTOs/UserDTO.swift
@@ -23,7 +23,7 @@ class UserDTO: NSManagedObject {
 
     @NSManaged var members: Set<MemberDTO>?
     @NSManaged var currentUser: CurrentUserDTO?
-    @NSManaged var teams: Set<TeamId>?
+    @NSManaged var teams: [TeamId]
     @NSManaged var channelMutes: Set<ChannelMuteDTO>
 
     /// Returns a fetch request for the dto with the provided `userId`.
@@ -87,6 +87,7 @@ extension UserDTO {
         
         let new = NSEntityDescription.insertNewObject(forEntityName: Self.entityName, into: context) as! UserDTO
         new.id = id
+        new.teams = []
         return new
     }
     
@@ -132,7 +133,7 @@ extension NSManagedObjectContext: UserDatabaseSession {
             dto.extraData = Data()
         }
 
-        dto.teams = Set(payload.teams)
+        dto.teams = payload.teams
 
         // payloadHash doesn't cover the query
         if let query = query, let queryDTO = try saveQuery(query: query) {
@@ -217,7 +218,7 @@ extension ChatUser {
             createdAt: dto.userCreatedAt,
             updatedAt: dto.userUpdatedAt,
             lastActiveAt: dto.lastActivityAt,
-            teams: dto.teams ?? [],
+            teams: Set(dto.teams),
             extraData: extraData
         )
     }
--- a/Sources/StreamChat/Database/DatabaseContainer_Tests.swift
+++ b/Sources/StreamChat/Database/DatabaseContainer_Tests.swift
@@ -38,6 +38,7 @@ class DatabaseContainer_Tests: XCTestCase {
                 userDTO.userCreatedAt = .init()
                 userDTO.userUpdatedAt = .init()
                 userDTO.userRoleRaw = "user"
+                userDTO.teams = []
             }, completion: { error in
                 XCTAssertNil(error)
                 goldenPathExpectation.fulfill()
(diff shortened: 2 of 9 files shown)


adamrushy (approved):
LGTM

adamrushy:
A few failing checks to look into @nuno-vieira

nuno-vieira:
> A few failing checks to look into @nuno-vieira

I'm trying to add iOS 12 running tests to the CI, but is kinda difficult because iOS 12.4 right now is only available on Xcode 10.3 in Github Actions. This make it really difficult for our code to run

nuno-vieira:
⚠️ Do not Merge. Waiting on @polqf to test this on iOS 12.

nuno-vieira:
@evsaev will continue to work on this. I can't run iOS 12 on my Xcode right now, so he will continue from here 👍

Sources/StreamChat/Database/DatabaseContainer_Tests.swift
```diff
@@ -38,6 +38,7 @@ class DatabaseContainer_Tests: XCTestCase {
                 userDTO.userCreatedAt = .init()
                 userDTO.userUpdatedAt = .init()
                 userDTO.userRoleRaw = "user"
+                userDTO.teams = []
```
evsaev, line 41: Would be nice to dedicate some time and revisit the API of creating DTOs. Ideally, there should be a single way of creating an object with all non-optional attributes as input params (`init` 😄?) leaving a developer 0 chance to instantiate the instance with an unassigned mandatory field.

Sources/StreamChat/Database/DTOs/UserDTO.swift
```diff
@@ -23,7 +23,7 @@ class UserDTO: NSManagedObject {
 
     @NSManaged var members: Set<MemberDTO>?
     @NSManaged var currentUser: CurrentUserDTO?
-    @NSManaged var teams: Set<TeamId>?
+    @NSManaged var teams: [TeamId]
```
nuno-vieira, line 26: Does this actually work? 🤔 I thought that Core Data only supported NSSet or NSOrderedSet 🤔 
Also, does this mean that we need to change all `Set<MemberDTO>` to `Set<MemberDTO>?` This might also be the reason for the crashes we are having on DTO's. CC @adamrushy

Sources/StreamChat/Database/StreamChatModel.xcdatamodeld/StreamChatModel.xcdatamodel/contents
```diff
@@ -248,7 +248,7 @@
         <attribute name="isOnline" attributeType="Boolean" usesScalarValueType="YES"/>
         <attribute name="lastActivityAt" optional="YES" attributeType="Date" usesScalarValueType="NO"/>
         <attribute name="name" optional="YES" attributeType="String"/>
-        <attribute name="teams" optional="YES" attributeType="Transformable" valueTransformerName="NSSecureUnarchiveFromData" customClassName="[String]"/>
+        <attribute name="teams" attributeType="Transformable" valueTransformerName="NSSecureUnarchiveFromData"/>
```
nuno-vieira, line 251: I think we should use this: `NSSecureUnarchiveFromDataTransformer` instead of `NSSecureUnarchiveFromData`

Sources/StreamChat/Database/DTOs/UserDTO.swift
evsaev: Using sets is fine for **relations** but not supported for `Transformable` attributes (at least on `<iOS13` which can be the reason why the crash is not happening does not pop up on later versions). Haven't yet found the docs with regard to `Transformable` on iOS13+ but keep looking

evsaev: Speaking about relations that can cause issues, I believe it's OK with those that have `to many` rule (since the empty set is a valid value) but it is worth checking `to one` for optionality mismatch between attribute and property. 

The same needs to be done for other primitive attributes. When I was debugging this crash I found some inconsistencies (e.g. `UserDTO.isOnline` is non-optional but the default value is `none`)
Screenshot 2022-02-09 at 20 00 56

Sources/StreamChat/Database/StreamChatModel.xcdatamodeld/StreamChatModel.xcdatamodel/contents
evsaev: Nice catch 👏

evsaev: Fixed ✅

Sources/StreamChat/Database/DTOs/UserDTO.swift
nuno-vieira: @evsaev Nice find! If you create a ticket please link it to this one: CIS-1558

nuno-vieira (commented):
I can't approve since I created the PR, but 
LGTM! 🎉

evsaev (approved):

PR #1785, opened 2022-02-09
[CIS-1557] Fix channel not removed from channel list when user leaves the channel

nuno-vieira:
### 🔗 Issue Link
CIS-1557

### 🎯 Goal
Change the default logic of `shouldAddNewChannelToList` and `shoudListUpdatedChannel`. Currently, whenever a user removes itself from the channel, the channel is still present in the channel list. This shouldn't be the default use case, because most customers want the opposite to happen. This will reduce the customer support tickets we are having.

### 🛠 Implementation
Move the `DemoApp` implementation to be the default one.

### 🎨 Changes
N/A

### 🧪 Testing
- Go to a channel
- Click on the Debug icon
- Remove yourself from the members
- Channel list should remove the channel

### ☑️ Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 # Upcoming
 
+### 🔄 Changed
+- `ChannelListVC` now keeps track of channels where user is a member only instead of all channels loaded in the SDK. [#1785](https://github.com/GetStream/stream-chat-swift/pull/1785)
+
 ### 🐞 Fixed
 - Make SendButton animation overridable [#1781](https://github.com/GetStream/stream-chat-swift/issues/1781)
 - Make ChannelId.rawValue public [#1780](https://github.com/GetStream/stream-chat-swift/pull/1780)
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### 🐞 Fixed
 - Make SendButton animation overridable [#1781](https://github.com/GetStream/stream-chat-swift/issues/1781)
 - Make ChannelId.rawValue public [#1780](https://github.com/GetStream/stream-chat-swift/pull/1780)
+- Fix channel not removed from channel list when user leaves the channel [#1785](https://github.com/GetStream/stream-chat-swift/pull/1785)
```
b-onc, line 9: This is not a fix, but a change in behavior: now default behavior is "ChannelListVC keeps track of channels where user is a member" instead of "all channels loaded in the SDK"

nuno-vieira: Hum 🤔 But in one way, this is going to be easier to find and understand if someone is looking for this issue. I'm going to add that to the change section, but it is worth to leave this one on the fix section as well

tbarbugli: please keep it as a fix, the previous implementation did not make any sense

nuno-vieira: @tbarbugli  Yes, we leaving it as a fix  👍

tbarbugli (approved):

evsaev (approved):
Looks good! 🚢 Would be nice to communicate to enterprise customers once this update is released with instructions on how to get the previous behavior back so it's not a surprise for those whose chat is **not** based on current user membership

nuno-vieira:
> Looks good! 🚢 Would be nice to communicate to enterprise customers once this update is released with instructions on how to get the previous behavior back so it's not a surprise for those whose chat is **not** based on current user membership

I'm pretty sure no customer wants this behaviour TBH 🤔

PR #1786, opened 2022-02-09
[CIS-1420] Fix `ChannelListController.loadNextChannels` using incorrect `limit`

b-onc:
### 🔗 Issue Link
CIS-1420

### 🎯 Goal

`ChannelListController.loadNextPages(limit:)` should use the `ChannelListController.query`s `pageSize` param when `limit` is omitted

### 🛠 Implementation

We forgot to set the `requestedChannelsLimit` value in `init`, so 25 was always used

### 🧪 Testing

Unit tests are added for the fix.

### ☑️ Checklist


Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift
```diff
@@ -42,7 +42,7 @@ public class ChatChannelListController: DataController, DelegateCallable, DataSt
             client.apiClient
         )
     
-    private let requestedChannelsLimit = 25
+    private let requestedChannelsLimit: Int
```
nuno-vieira, line 45:
```suggestion
    private let requestedChannelsLimit: Int
```
```suggestion
    private let requestedChannelsLimit: Int
```

nuno-vieira, line 45: Do we even need this?
```suggestion
```

We can then later just use:
`let limit = limit ?? query.pagination.pageSize` instead of `let limit = limit ?? requestedChannelsLimit`

nuno-vieira (approved):
LGTM! ✅ Just added on small nit comment, feel free to update it, but not a big deal 👍

Sources/StreamChat/Controllers/ChannelListController/ChannelListController_Tests.swift
```diff
@@ -676,14 +714,53 @@ class ChannelListController_Tests: XCTestCase {
...
+        XCTAssertFalse(completionCalled)
+        
+        // Assert correct `Pagination` is created
+        XCTAssertEqual(
+            env!.channelListUpdater?.update_queries.first?.pagination,
+            .init(pageSize: pageSize, offset: controller.channels.count)
+        )
+        
+        // Keep a weak ref so we can check if it's actually deallocated
+        weak var weakController = controller
```
evsaev, line 747: The same comment about 2 test vs 1 applies

evsaev (commented):
Looks good 👍 Just a minor comment with regard to multi-purpose tests.

Another thing I noticed (unrelated to the PR) is that we are currently not consistent about how we pass the page size. There're 3 options for how we do that:
1. `limit: Int? = nil` (ChannelListController.loadNextChannels)
2. `limit: Int = 25` (ChatMessageSearchController.loadNextMessages)
3. `limit: Int = .channelWatchersPageSize` (ChatChannelWatcherListController.loadNextWatchers)

Not a big deal but having a single way would make it a bit more organized 🙂 I vote for 3rd option but with a dedicated scope for constants to not pollute built-in types namespaces

nuno-vieira:
> Looks good 👍 Just a minor comment with regard to multi-purpose tests.
> 
> 
> 
> Another thing I noticed (unrelated to the PR) is that we are currently not consistent about how we pass the page size. There're 3 options for how we do that:
> 
> 1. `limit: Int? = nil` (ChannelListController.loadNextChannels)
> 
> 2. `limit: Int = 25` (ChatMessageSearchController.loadNextMessages)
> 
> 3. `limit: Int = .channelWatchersPageSize` (ChatChannelWatcherListController.loadNextWatchers)
> 
> 
> 
> Not a big deal but having a single way would make it a bit more organized 🙂 I vote for 3rd option but with a dedicated scope for constants to not pollute built-in types namespaces

Nice catch @evsaev ! We should have consistency. 1. or 2. is what makes more sense

Sources/StreamChat/Controllers/ChannelListController/ChannelListController_Tests.swift
```diff
@@ -164,6 +164,44 @@ class ChannelListController_Tests: XCTestCase {
...
+            AssertTestQueue(withId: queueId)
+            completionCalled = true
+        }
+        
+        // Keep a weak ref so we can check if it's actually deallocated
+        weak var weakController = controller
+        
+        // (Try to) deallocate the controller
+        // by not keeping any references to it
+        controller = nil
```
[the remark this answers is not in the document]
b-onc: Since nearly all `ChannelListController` tests have this logic, I think we should leave this to a bigger refactoring task for now

PR #1787, opened 2022-02-09
[CIS-1553] Fix Message Input Accessibility for Large Text

nuno-vieira:
### 🔗 Issue Link
CIS-1553

### 🎯 Goal
When the accessibility for the input text was Large, the placeholder was being wrapped.

### 🛠 Implementation
Make sure the placeholder has a limited width and that the font size is adjusted to a certain point to fit the width of the input.

### 🎨 Changes
Before:
image

After: 
image

### 🧪 Testing
- Open Accessibility Inspector  (Xcode > Open Developer Tool > Accessibility Inspector)
- Select the parameters icon in the top right corner
- Change the text size to large

### ☑️ Checklist


Sources/StreamChatUI/CommonViews/InputTextView/InputTextView.swift
```diff
@@ -98,10 +100,10 @@ open class InputTextView: UITextView, AppearanceProvider {
             )
         )
         placeholderLabel.pin(anchors: [.centerY], to: self)
```
evsaev, line 102: Hmm, why do we need this constraint? Aren't we pinning all frame anchors to the superview?

evsaev (approved):
👏

Sources/StreamChatUI/CommonViews/InputTextView/InputTextView.swift
nuno-vieira: I also though it was weird, but we need it some how, otherwise the placeholder is not centered in the y axis 🤔

PR #1788, opened 2022-02-11
Offline Support - APIClient Recovery Mode

polqf:
### 🔗 Issue Link
Par of https://stream-io.atlassian.net/browse/CIS-1484

### 🎯 Goal

The goal of this PR is to allow the APIClient to enter in a recovery mode where no other requests than the ones involved in the recovery/offline sync. During that process, requests in the APIClient should be executed in serial

### 🛠 Implementation

This PR introduces the concept of Operation Queues in the APIClient. There are two queues:
- Requests queue: Concurrent operations. Gives us the option to stop the queue and to easily queue a request.
- Recovery queue: Serial queue, where only recovery related operations are handled

The idea of having two queues is to limit the complexity added by having recovery mode. When in recovery mode, the requests queue is paused, but requests can still be queued. Once the recovery is done, we simply resume the queue so the requests can go through.

### ☑️ Checklist


Sources/StreamChat/APIClient/APIClient.swift
```diff
@@ -28,6 +28,18 @@ class APIClient {
...
+    private let recoveryQueue: OperationQueue = {
+        let operationQueue = OperationQueue()
+        operationQueue.name = "com.stream.api-client.recovery"
+        operationQueue.maxConcurrentOperationCount = 1
+        return operationQueue
+    }()
+
+    /// Determines whether the APIClient is in recovery mode. During recovery period we limit the concurrent operations to 1, and we only allow recovery related
+    /// requests to be run,
+    @Atomic private var isRecoveryModeOn: Bool = false
```
bielikb, line 41: Something like
```suggestion
    @Atomic private var isInRecoveryMode: Bool = false
```

might be more readable. Furtherly, does it makes sense to turn this property into readonly?
```
private (set) var ...
```

Sources/StreamChat/Repositories/SyncRepository_Tests.swift
```diff
@@ -190,10 +196,11 @@ final class SyncRepository_Tests: XCTestCase {
 
         if let result = requestResult {
             // Simulate API Failure
-            AssertAsync {
-                Assert.willNotBeNil(self.apiClient.request_completion)
+            AssertAsync.willBeTrue(apiClient.recoveryRequest_completion != nil)
+            guard let callback = apiClient.recoveryRequest_completion as? (Result<MissingEventsPayload, Error>) -> Void else {
+                XCTFail()
```
bielikb, line 201: Can we populate the `XCTFail` call here? If this test fails on CI it will be hard to reason about the issue.

Sources/StreamChat/APIClient/APIClient.swift
```diff
@@ -145,10 +179,9 @@ class APIClient {
                     completion(.success(decodedResponse))
                 } catch {
                     if error is ClientError.ExpiredToken {
-                        self.refreshToken {
-                            log.info("Token Refreshed", subsystems: .httpRequests)
+                        self.refreshToken { refreshError in
+                            completion(.failure(refreshError ?? error))
```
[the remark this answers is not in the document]
polqf: Added new documentation here. Let me know 😄

bielikb (approved):

evsaev (approved):

PR #1789, opened 2022-02-11
Offline Support - APIClient with OperationQueue

polqf:
### 🔗 Issue Link
Part of https://stream-io.atlassian.net/browse/CIS-1484

### 🎯 Goal

The goal of this PR is to move the APIClient to use an OperationQueue to handle incoming requests

### 🛠 Implementation

Requests queue: Concurrent operations. Gives us the option to stop the queue and to easily queue a request.

### ☑️ Checklist

### DIFF
--- a/Sources/StreamChat/APIClient/APIClient.swift
+++ b/Sources/StreamChat/APIClient/APIClient.swift
@@ -226,11 +226,19 @@ class APIClient {
         progress: ((Double) -> Void)?,
         completion: @escaping (Result<URL, Error>) -> Void
     ) {
-        cdnClient.uploadAttachment(
-            attachment,
-            progress: progress,
-            completion: completion
-        )
+        let cdnClient = self.cdnClient
+        let uploadOperation = AsyncOperation(maxRetries: maximumRetries) { [weak self] operation, done in
+            cdnClient.uploadAttachment(attachment, progress: progress) { result in
+                switch result {
+                case let .failure(error) where self?.shouldRetry(error, operation: operation) == true:
+                    done(.retry)
+                case .success, .failure:
+                    completion(result)
+                    done(.continue)
+                }
+            }
+        }
+        operationQueue.addOperation(uploadOperation)
     }
 
     func flushRequestsQueue() {
--- a/Sources/StreamChat/APIClient/APIClient.swift
+++ b/Sources/StreamChat/APIClient/APIClient.swift
@@ -43,7 +43,17 @@ class APIClient {
 
     /// Stores request failed with token expired error for retrying them later
     private var requestsQueue = [RequestsQueueItem]()
-    
+
+    /// Queue in charge of handling incoming requests
+    private let operationQueue: OperationQueue = {
+        let operationQueue = OperationQueue()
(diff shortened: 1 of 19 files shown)


Sources/StreamChat/APIClient/CDNClient_Mock.swift
```diff
@@ -5,21 +5,23 @@
...
-                progress,
-                completion
-            )
-        )
+        record()
+        uploadAttachmentProgress.map { progress?($0) }
+
+        DispatchQueue.main.asyncAfter(deadline: .now() + 0.1) {
+            self.uploadAttachmentResult.map(completion)
+        }
```
evsaev, line 25: nit: I'd prefer to save the `completion` and call it manually, tbh. It just makes debugging easier when the stack trace clearly shows what the sequence of calls is

Sources/StreamChat/Repositories/SyncOperations.swift
```diff
@@ -34,7 +34,7 @@ class GetChannelIdsOperation: AsyncOperation {
 
 class GetPendingConnectionDateOperation: AsyncOperation {
     init(database: DatabaseContainer, context: SyncContext) {
-        super.init(maxRetries: 2) { [weak database] done in
+        super.init(maxRetries: 2) { [weak database] _, done in
```
evsaev, line 37: nit: that's not directly related to the PR but would be nice to extract `maxRetries` to a constant at some point as it's reused across all sync flow operations.

Sources/StreamChat/APIClient/APIClient.swift
```diff
@@ -192,84 +157,67 @@ class APIClient {
...
+        // Increase the amount of consecutive failures
+        _tokenRefreshConsecutiveFailures.mutate { $0 += 1 }
+
+        tokenRefresher { [weak self] in
+            self?.isRefreshingToken = false
+            // We restart the queue now that token refresh is completed
+            self?.operationQueue.isSuspended = false
+            completion()
+        }
+    }
```
evsaev, line 179: The expired token error can also come via the web socket so, eventually, we'll need to encapsulate the refreshing logic with all the state info.

It's also weird we don't propagate token refreshing results so we don't know if the token is refreshed without letting the next request go, right? 🤔

evsaev (approved):
Well done 👏 Some minor comments about the new approach to mocks

Sources/StreamChat/APIClient/CDNClient_Mock.swift
```diff
@@ -5,21 +5,23 @@
 import StreamChat
 import StreamChatTestTools
 
-final class CDNClient_Mock: CDNClient {
+final class CDNClient_Mock: CDNClient, Spy {
+    var recordedFunctions: [String] = []
```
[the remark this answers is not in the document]
polqf: I do agree that we need to stick with one approach, definitely. But the truth is that MockFunc is pretty obscure and not so easy to understand, and the variables with counters/booleans do not scale.

@bielikb this is something to keep in mind for the Testing Strategy 🙏

PR #1791, opened 2022-02-14
Offline Support - Offline Actions Queuing + Running

polqf:
### 🔗 Issue Link
Part of https://stream-io.atlassian.net/browse/CIS-1484

### 🎯 Goal

This PR adds the final bits to get the offline queued requests executed when the connection is recovered

### 🛠 Implementation

`OfflineRequestsRepository` handles both the enqueuing and the execution of the requests when needed.
When running the queued requests, it basically passes the requests on to the APIClient, and waits for its result.

- If the requests succeeds -> The request is removed from the pending ones
- If the request fails with a connection error -> The request is kept to be executed once the connection is back (we are not putting it back at the queue to make sure we respect the order)
- If the request fails with any other error -> We are dismissing the request, and removing it from the queue

### ☑️ Checklist

### DIFF
--- a/Sources/StreamChat/APIClient/APIClient.swift
+++ b/Sources/StreamChat/APIClient/APIClient.swift
@@ -226,11 +226,19 @@ class APIClient {
         progress: ((Double) -> Void)?,
         completion: @escaping (Result<URL, Error>) -> Void
     ) {
-        cdnClient.uploadAttachment(
-            attachment,
-            progress: progress,
-            completion: completion
-        )
+        let cdnClient = self.cdnClient
+        let uploadOperation = AsyncOperation(maxRetries: maximumRetries) { [weak self] operation, done in
+            cdnClient.uploadAttachment(attachment, progress: progress) { result in
+                switch result {
+                case let .failure(error) where self?.shouldRetry(error, operation: operation) == true:
+                    done(.retry)
+                case .success, .failure:
+                    completion(result)
+                    done(.continue)
+                }
+            }
+        }
+        operationQueue.addOperation(uploadOperation)
     }
 
     func flushRequestsQueue() {
--- a/Sources/StreamChat/APIClient/APIClient.swift
+++ b/Sources/StreamChat/APIClient/APIClient.swift
@@ -43,7 +43,17 @@ class APIClient {
 
     /// Stores request failed with token expired error for retrying them later
     private var requestsQueue = [RequestsQueueItem]()
-    
+
+    /// Queue in charge of handling incoming requests
+    private let operationQueue: OperationQueue = {
+        let operationQueue = OperationQueue()
+        operationQueue.name = "com.stream.api-client"
+        return operationQueue
+    }()
+
+    private let maximumRetries = 3
+    private let timeout: TimeInterval = 60
+
     /// Shows whether the token is being refreshed at the moment
     @Atomic private var isRefreshingToken: Bool = false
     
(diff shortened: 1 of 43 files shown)


Sources/StreamChat/APIClient/APIClient.swift
```diff
@@ -94,9 +99,8 @@ class APIClient {
         endpoint: Endpoint<Response>,
         completion: @escaping (Result<Response, Error>) -> Void
     ) {
-        guard isRecoveryModeOn else {
+        if !isRecoveryModeOn {
             log.assertionFailure("We should not call this method if not in recovery mode")
```
adamrushy, line 103: Before I forget, we should probably assert here or at least return out?

Sources/StreamChat/APIClient/APIClient.swift
```diff
@@ -252,12 +270,14 @@ class APIClient {
...
+        log.debug("Entering recovery mode", subsystems: .offlineSupport)
         isRecoveryModeOn = true
         operationQueue.isSuspended = true
     }
 
     func exitRecoveryMode() {
         // Once recovery is done, regular requests can go through again.
+        log.debug("Leaving recovery mode", subsystems: .offlineSupport)
         isRecoveryModeOn = false
         operationQueue.isSuspended = false
```
adamrushy, line 282: Before I forget, does this automatically kick off the OQ? From memory, it doesn't but I could be wrong...

polqf: It is resuming the Operation Queue just by setting this value 😄

polqf: `You may continue to add operations to a queue that is suspended but those operations are not scheduled for execution until you change this property to false.`

adamrushy: No problem it was just one to check 💯

Sources/StreamChat/Repositories/OfflineRequestsRepository.swift
```diff
@@ -0,0 +1,134 @@
...
+
+    init(database: DatabaseContainer, apiClient: APIClient) {
+        self.database = database
+        self.apiClient = apiClient
+    }
+
+    /// - If the requests succeeds -> The request is removed from the pending ones
+    /// - If the request fails with a connection error -> The request is kept to be executed once the connection is back (we are not putting it back at the queue to make sure we respect the order)
+    /// - If the request fails with any other error -> We are dismissing the request, and removing it from the queue
+    func runQueuedRequests(completion: @escaping () -> Void) {
```
evsaev, line 40: This part is a bit confusing. If the goal is to not execute any API requests while the recovery flow, why it's not enough to just add them to `operationQueue` which gets suspended when we enter recovery flow and resumed when recovery is done?

Sources/StreamChat/Repositories/OfflineRequestsRepository.swift
```diff
@@ -0,0 +1,134 @@
...
+        var pendingActions: [(String, Data)] = []
+
+        let readContext = database.backgroundReadOnlyContext
+        readContext.performAndWait {
```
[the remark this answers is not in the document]
nuno-vieira: Yes, good point, we need to make sure we don't run this from the MainThread. But sync this is part of the Sync Operations, I don't think that will be the case?

Sources/StreamChat/Repositories/OfflineRequestsRepository.swift
```diff
@@ -0,0 +1,134 @@
...
+            pendingActions = QueuedRequestDTO.loadAllPendingRequests(context: readContext).map {
+                ($0.id, $0.endpoint)
+            }
```
nuno-vieira, line 47: Shouldn't we read these pending requests from the retry queue as well? What happens if meanwhile, we add an offline request to the database while we are reading from it? Won't there be any concurrency issues?

nuno-vieira (commented):
Looks really good overall 👍  Just added some comments and questions 🙂

Sources/StreamChat/Repositories/OfflineRequestsRepository.swift
```diff
@@ -0,0 +1,134 @@
...
+    ) {
+        // TODO: Temporary String check. We only process the response for sent messages
+        guard endpoint.path.hasSuffix("/message") else {
+            completion()
+            return
+        }
+        completion()
+    }
+
+    func queueOfflineRequest(endpoint: DataEndpoint, completion: (() -> Void)? = nil) {
```
martinmitrevski, line 111: Does it makes sense to provide more info in the completion block? At the moment the caller doesn't know which of the different cases called it.

Sources/StreamChat/APIClient/APIClient.swift
```diff
@@ -220,18 +237,14 @@ class APIClient {
...
     }
 
-    private func shouldRetry(_ error: Error, operation: AsyncOperation) -> Bool {
+    private func isConnectionError(_ error: Error) -> Bool {
         // We only retry transient errors like connectivity stuff or HTTP 5xx errors
         guard ClientError.isEphemeral(error: error) else {
             return false
         }
 
         let offlineErrorCodes: Set<Int> = [NSURLErrorDataNotAllowed, NSURLErrorNotConnectedToInternet]
```
martinmitrevski, line 246: Are these enough for offline error codes? I would suggest going through the URLError.Codes to possibly include others too (e.g. `networkConnectionLost`, `internationalRoamingOff` etc). 
Also, I think it's better to use the more Swifty syntax here (`URLError.Code.notConnectedToInternet`).

martinmitrevski (commented):
Looks great in general! Added some comments and @nuno-vieira and @evsaev also have some valid remarks worth checking.

Sources/StreamChat/Repositories/OfflineRequestsRepository.swift
polqf: Discussed offline

Sources/StreamChat/APIClient/APIClient.swift
```diff
@@ -118,8 +122,21 @@ class APIClient {
...
                     done(.retry)
                     operation.resetRetries()
-                case let .failure(error) where self?.shouldRetry(error, operation: operation) == true:
-                    done(.retry)
+                case let .failure(error) where self?.isConnectionError(error) == true:
+                    // Do not retry unless its a connection problem and we still have retries left
+                    if operation.canRetry {
+                        done(.retry)
+                        return
+                    }
```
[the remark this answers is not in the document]
polqf: Done in further PR

Sources/StreamChat/APIClient/APIClient.swift
```diff
@@ -118,8 +122,21 @@ class APIClient {
                     // Retry request. Expired token has been refreshed
                     done(.retry)
                     operation.resetRetries()
```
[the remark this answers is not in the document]
polqf: Discussed offline, and fixed in following PRs

Sources/StreamChat/APIClient/Endpoints/Endpoint.swift
```diff
@@ -27,9 +27,50 @@ struct Endpoint<ResponseType: Decodable> {
...
+    }
+
+    init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        path = try container.decode(String.self, forKey: .path)
+        method = try container.decode(EndpointMethod.self, forKey: .method)
+        queryItems = try container.decodeIfPresent(Data.self, forKey: .queryItems)
+        requiresConnectionId = try container.decode(Bool.self, forKey: .requiresConnectionId)
+        requiresToken = try container.decode(Bool.self, forKey: .requiresToken)
+        body = try container.decodeIfPresent(Data.self, forKey: .body)
```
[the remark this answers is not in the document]
polqf: This allows for better debugging

Sources/StreamChat/Repositories/SyncRepository_Tests.swift
```diff
@@ -408,4 +419,62 @@ final class SyncRepository_Tests: XCTestCase {
...
+            path: "",
+            method: .post,
+            queryItems: nil,
+            requiresConnectionId: true,
+            requiresToken: true,
+            body: nil
+        )
+        repository.queueOfflineRequest(endpoint: endpoint)
+
+        XCTAssertNotCall("queueOfflineRequest(endpoint:completion:)", on: offlineRequestsRepository)
```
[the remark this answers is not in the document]
polqf: As discussed in the tech meeting, completely agree with needed some standarization. As of now, let's keep it, and reevaluate once there's a final standard in place

Sources/StreamChat/APIClient/APIClient.swift
polqf: Actually, this is "improved" in a following PR where only `ClientError.isEphemeral` is used. ATM we are checking codes differently in 3 parts of the LLC...

Sources/StreamChat/Repositories/OfflineRequestsRepository.swift
polqf: You cannot actually do that as the contexts have their own queue. Even if I move it to the retry queue, they will be then moved to the context's queue, leaving the end result as it is right now

martinmitrevski (approved):

Sources/StreamChat/Repositories/OfflineRequestsRepository.swift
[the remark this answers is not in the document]
polqf: Yep, exactly what Nuno mentioned. But I will add a note to improve it in the following PRs

nuno-vieira (approved):
LGTM! ✅

PR #1780, opened 2022-02-04
[CIS-1549] Make `ChannelId.rawValue` public

nuno-vieira:
### 🔗 Issue Link
CIS-1549

### 🎯 Goal

Simple one-liner, just to fix the `ChannelId.rawValue` access control. 

### 🛠 Implementation

Change `ChannelId.rawValue` access control to public. 

### 🎨 Changes
N/A

### 🧪 Testing
N/A

### ☑️ Checklist
- [x] I have signed the [Stream CLA](https://docs.google.com/forms/d/e/1FAIpQLScFKsKkAJI7mhCr7K9rEIOpqIDThrWxuvxnwUq2XkHyG154vQ/viewform) (required)
- [x] This change follows zero ⚠️ policy (required)
- [x] Changelog is updated with client-facing changes

PR #1792, opened 2022-02-15
4.10.1 Release

nuno-vieira:
### 🔄 Changed
- `ChannelListVC` now keeps track of channels where user is a member only instead of all channels loaded in the SDK. [#1785](https://github.com/GetStream/stream-chat-swift/pull/1785)

### 🐞 Fixed
- Make SendButton animation overridable [#1781](https://github.com/GetStream/stream-chat-swift/issues/1781)
- Make ChannelId.rawValue public [#1780](https://github.com/GetStream/stream-chat-swift/pull/1780)
- Fix channel not removed from channel list when user leaves the channel [#1785](https://github.com/GetStream/stream-chat-swift/pull/1785)
- Fix `ChannelListController.loadNextChannels` using incorrect `limit` when argument is omitted [#1786](https://github.com/GetStream/stream-chat-swift/issues/1786)
- Fix Message Input Accessibility for Large Text [#1787](https://github.com/GetStream/stream-chat-swift/pull/1787)
- Fix crash on iOS 12 when local storage enabled [#1784](https://github.com/GetStream/stream-chat-swift/pull/1784)


Examples/SlackClone/ChatClient.swift
```diff
@@ -27,7 +27,8 @@ extension ChatClient {
         Appearance.default = appearance
         Components.default = components
         
-        let config = ChatClientConfig(apiKey: APIKey("q95x9hkbyd6p"))
+        var config = ChatClientConfig(apiKey: APIKey("q95x9hkbyd6p"))
+        config.isLocalStorageEnabled = true
```
bielikb, line 31: @nuno-vieira @evsaev is this left over from CoreData related PR?

nuno-vieira: This is fine. We need an example app with this enabled so that we can easily observe these issues 👍 Ideally in the future, we want to perform UI Tests on this example app to make sure it works on iOS 12. But Github Actions is limiting use on doing this for now.

CHANGELOG.md
```diff
@@ -5,6 +5,20 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### 🔄 Changed
 
+# [4.10.1](https://github.com/GetStream/stream-chat-swift/releases/tag/4.10.1)
+_February 15, 2022_
```
bielikb, line 9:
```suggestion
_February 16, 2022_
```

Examples/SlackClone/ChatClient.swift
bielikb: Ok. If this is intentional, this comment can be resolved.

PR #1798, opened 2022-02-16
[CIS-879] Enable Camera Functionality on Composer

adamrushy:
### 🔗 Issue Link
https://stream-io.atlassian.net/browse/CIS-879

### 🎯 Goal

To bring us in line with the other SDK's we are missing the camera functionality on the composer. The goal here is to enable the camera option so users can take a picture or video during the app.

### 🛠 Implementation

The implementation is simple, create a new UIImagePicker which has the Camera source type. Then split the logic for displaying the available actions to make it easier for consumers to override either the file attachments action / cancel action or the media actions.

### 🧪 Testing

1. Launch the Demo App
2. Enter a Channel
3. Tap on Attachments
4. Tap on Camera
5. Take Picture or Video
6. Media is attached

### ☑️ Checklist


Sources/StreamChatUI/Composer/ComposerVC.swift
```diff
@@ -431,26 +440,50 @@ open class ComposerVC: _ViewController,
         present(filePickerVC, animated: true)
     }
     
-    /// Returns actions for attachments picker.
-    open var attachmentsPickerActions: [UIAlertAction] {
```
b-onc, line 435: getting rid of this is a breaking change, IMHO we should keep it and play along (add a `showCameraAction` in this variable)

Sources/StreamChatUI/Composer/ComposerVC.swift
```diff
@@ -243,13 +243,22 @@ open class ComposerVC: _ViewController,
...
 
     /// The view controller for selecting image attachments.
-    open private(set) lazy var mediaPickerVC: UIViewController = {
+    open private(set) lazy var mediaPickerVC: UIImagePickerController = {
         let picker = UIImagePickerController()
         picker.mediaTypes = UIImagePickerController.availableMediaTypes(for: .savedPhotosAlbum) ?? ["public.image"]
         picker.sourceType = .savedPhotosAlbum
         picker.delegate = self
         return picker
     }()
```
nuno-vieira, line 252: This is also a breaking change. I guess this is a `UIViewController` so that customers can use a different picker, but then it is kinda useless to have a property with `UIViewController` anyway 🤔 But since this is a breaking change, what we can do is create a new property, and maybe call it `imagePicker` and deprecate the `mediaPickerVC`.

nuno-vieira (changes requested):

Sources/StreamChatUI/Composer/ComposerVC.swift
nuno-vieira: But it doesn't seem we are using this property any way 🤔 Why is the type changed?

Sources/StreamChatUI/Composer/ComposerVC.swift
adamrushy: Yep good points, resolved now.

Sources/StreamChatUI/Composer/ComposerVC.swift
adamrushy: Yeah, I have reverted back to keep the same API 👍🏼

Sources/StreamChatUI/Composer/ComposerVC.swift
```diff
@@ -250,6 +250,15 @@ open class ComposerVC: _ViewController,
...
+    
+    /// The View Controller for taking a picture.
+    open private(set) lazy var cameraVC: UIViewController = {
```
evsaev, line 255: nit: In this case, we can actually go without `private(set)` allowing SDK integrators to assign a VC they wanna show

evsaev: With settable `cameraVC`, the customization of a composer from channel/thread VC can look like this: 
```swift
messageComposerVC.cameraVC = MyCameraVC(didSelectPhoto: { [weak messageComposerVC] url in
  try? messageComposerVC?.addAttachmentToContent(from: url, type: .image)
})
```

With `private(set)` the only way to go is:
```swift
class CustomComposerVC: ComposerVC {
    override var cameraVC: UIViewController {
        MyCameraVC(didSelectPhoto: { [weak self] url in
          try? self?.addAttachmentToContent(from: url, type: .image)
        })
    }
}
```

PR #1802, opened 2022-02-18
[CIS-1474] Use Xcode 13.2.1 on CI

bielikb:
### 🔗 Issue Link
CIS-1474
CIS-1526

### 🎯 Goal

Update CI to Xcode 13.2.1

### 📝 Changes

* Lanes are now leveraging Xcode selected by `xcversion` (if provided via env variable `XCODE_VERSION`)
* Xcode project is updated to Xcode 13.2.1
* Unused workflows were removed
* Updates snapshots (fonts diff)

### ☑️ Contributor Checklist


.github/workflows/smoke-checks.yml
```diff
@@ -156,11 +159,11 @@ jobs:
     - uses: ./.github/actions/bootstrap
     - name: Build LLC (Xcode 12)
       env:
-        DEVELOPER_DIR: "/Applications/Xcode_12.5.1.app/Contents/Developer"
+        XCODE_VERSION: "12.5.1"
```
b-onc, line 162: Why don't we put this to env too?

.github/workflows/macos-checks.yml
```diff
@@ -1,103 +0,0 @@
-name: macOS checks
```
b-onc, line 1: IMHO we should keep this workflow, we support macos and we should fix and enable this workflow... at some point in time...

b-onc (approved):
LGTM

.github/workflows/macos-checks.yml
bielikb: We decided to remove worfklows that are not running/leveraged right now. What I can do is tag this commit so we will be able to come back to it. WDYT?

b-onc: alright

.github/workflows/smoke-checks.yml
bielikb: @b-onc Can you elaborate a bit more. Thanks!

b-onc:
```
XCODE_VERSION: ${{ env.XCODE_VERSION }}
```

bielikb: The issue with the above proposal is that anytime we try to update our CI, we will change the Xcode version globally for all CI runs.

b-onc: gotcha. in my hurry i missed that it's Xcode 12 and not 13

.github/workflows/smoke-checks.yml
```diff
@@ -113,14 +116,14 @@ jobs:
...
-    - name: Prepare iPhone 12 with iOS 15.0
+    - name: Prepare iPhone 12 with iOS 15.2
       uses: ./.github/actions/prepare-ios-simulator
       with:
-        device: "iPhone 12"
-        ios: "15.0"
-        xcode: "13.0"
+        device: ${{ env.IOS_SIMULATOR_DEVICE }}
+        ios: ${{ env.IOS_VERSION }}
+        xcode: ${{ env.XCODE_VERSION }}
```
nuno-vieira, line 124: Do we really need to update the version for the UI Tests?

This will require re-recording all screenshots. If we do this for every version update the size of the repo will increase exponentially. So I would recommend updating the iOS Version for the UI Snapshot Tests very rarely. At least until we use something like Git LFS.

I'm also not sure if it is a good idea to use these environment variables because most of the time we need different version and environments for the tests 🤔

nuno-vieira (changes requested):
LGTM Overall but I'm going to requests changes for now because I believe it is better to have the UI Snapshot Tests with different version than the default ones. We don't want to update the UI Snapshots everytime, this will increase a lot the repo size

.github/workflows/smoke-checks.yml
bielikb: Hi @nuno-vieira,

thanks for the comment.
I distilled 3 topics from your concerns:

1. Update of the Xcode version require us to everytime update the screenshots that are now part of the repo
- Yeah, this is definitely an issue, yet this issue is out of the scope of this PR. 
Altogether we updated cca 16 screenshots, where average size per screenshot is cca 6KB = +96KB in total. Is this a concern at this point in time?

2. Updating the iOS version requires us to re-record all the screenshots
- This is actually not the case. In this PR we got to update cca 16 screenshots generated by just a few snapshot tests.

3. _I'm also not sure if it is a good idea to use these environment variables because most of the time we need different version and environments for the tests_
If that was the case, the env variables can be overriden for such use cases. At this time they're shared across the actions. The added value of envs is that they're easily spotted and on top of the worfklow == easy to update across the stack if needed.

nuno-vieira: @bielikb It depends on the iOS Versions. Some iOS Versions require more tests to be re-recorded. For every iOS update, this will add-up quite significantly

nuno-vieira (approved):

.github/workflows/smoke-checks.yml
nuno-vieira: @bielikb I'll approve the PR anyways. But I can forsee this being a problem in the future. We can revisit this in the future in case we see major differences between iOS Versions. 👍

polqf (approved):

PR #1810, opened 2022-02-23
Offline Support: Move to EndpointPath

polqf:
### 🔗 Issue Link
-

### 🎯 Goal

This is a precondition for the following PR, where we need to have a way to compare different Endpoint paths.

### 🛠 Implementation

It is not scalable to compare string values, therefore this EndpointPath enum has been created.
When it comes to tests, we were lucky enough to have all the endpoints covered with tests, and this move should be secure.

### 📝 Changes

- Move to use EndpointPath when creating an Endpoint object. No more string based paths

### 🧪 Testing

- Unit tested adapting the current tests

### ☑️ Contributor Checklist


adamrushy (approved):
Looks good, just need to checking failing test on LLC

Sources/StreamChat/APIClient/Endpoints/EndpointPath.swift
```diff
@@ -0,0 +1,87 @@
...
+        case let .message(messageId): return "messages/\(messageId)"
+        case let .editMessage(messageId): return "messages/\(messageId)"
+        case let .deleteMessage(messageId): return "messages/\(messageId)"
+        case let .replies(messageId): return "messages/\(messageId)/replies"
+        case let .reactions(messageId): return "messages/\(messageId)/reactions"
+        case let .addReaction(messageId): return "messages/\(messageId)/reaction"
+        case let .deleteReaction(messageId, reaction): return "messages/\(messageId)/reaction/\(reaction.rawValue)"
+        case let .messageAction(messageId): return "messages/\(messageId)/action"
+
+        case .banMember: return "moderation/ban"
```
martinmitrevski, line 81: Don't we have unban member?

Sources/StreamChat/APIClient/Endpoints/EndpointPath.swift
```diff
@@ -0,0 +1,87 @@
...
+    case channelUpdate(String)
+    case muteChannel(Bool)
+    case showChannel(String, Bool)
+    case truncateChannel(String)
+    case markChannelRead(String)
+    case markAllChannelsRead
+    case channelEvent(String)
+    case stopWatchingChannel(String)
+    case pinnedMessages(String)
+    case uploadAttachment(String, String)
```
martinmitrevski, line 29: Might be good to document the not so obvious associated values. For example, `uploadAttachment`, what do we pass in the two `String`s.

Sources/StreamChat/APIClient/Endpoints/EndpointPath.swift
```diff
@@ -0,0 +1,87 @@
...
+    case addReaction(MessageId)
+    case deleteReaction(MessageId, MessageReactionType)
+    case messageAction(MessageId)
+
+    case banMember
+    case flagUser(Bool)
+    case flagMessage(Bool)
+    case muteUser(Bool)
+
+    var value: String {
```
martinmitrevski, line 46: Do we have existing tests for all the cases listed here?

martinmitrevski (approved):
This is a very nice improvement to the SDK ✅ . 
However, it also impacts big chunk of the codebase, so we need to make sure we didn't break anything. My assumption is that we have tests for all the URLs impacted by the change.

Sources/StreamChat/Repositories/OfflineRequestsRepository.swift
```diff
@@ -100,30 +101,22 @@ class OfflineRequestsRepository {
         data: Data,
         completion: @escaping () -> Void
     ) {
-        // TODO: Temporary String check. We only process the response for sent messages
```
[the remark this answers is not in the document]
polqf: The following PR #1811 handles this piece

Sources/StreamChat/APIClient/Endpoints/EndpointPath.swift
polqf: It uses the same path, different parameters

polqf: https://github.com/GetStream/stream-chat-swift/pull/1810/files#diff-887fd13025bf4b9e693ada9eec5f554291c79eca1b8c83a54ad8ffb717e749e2R33

Sources/StreamChat/APIClient/Endpoints/EndpointPath+OfflineRequest.swift
```diff
@@ -0,0 +1,25 @@
...
+//
+
+import Foundation
+
+extension EndpointPath {
+    var shouldBeQueuedOffline: Bool {
+        switch self {
+        case .createChannel, .sendMessage, .editMessage, .deleteMessage, .addReaction, .deleteReaction:
+            return true
+        case .connect, .sync, .users, .guest, .members, .search, .devices, .channels, .updateChannel, .deleteChannel,
```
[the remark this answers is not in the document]
polqf: I'm pretty much against default clauses. If we add any new case in the future, there is no warning to look at this pieces of code and take a decision.

Sources/StreamChat/APIClient/Endpoints/EndpointPath.swift
polqf: Yeah, fortunately enough, the current XXXEndpoint_Tests were covering those, and I updated those.

PR #1811, opened 2022-02-23
Offline Support: Recovery Post Queued Requests

polqf:
### 🔗 Issue Link
https://stream-io.atlassian.net/browse/CIS-1588

### 🎯 Goal

This PR handles the necessary actions post offline queued request.
Whenever we queue a request that happened while we are offline we lost its database context, therefore we need to perform some actions to make sure the state in the server post request matches our local database

On channel:
- Creation

On messages:
- Create
- Edit
- Delete

On reactions:
- Add
- Remove

### 🛠 Implementation

Using the enum (EndpointPath) introduced in #1810 we know which is the request that was just run, and we can execute the needed actions to leave the database in the right state.
Because this logic is already present in some workers, and using them directly would not be appropriate, this logic has been extracted in new repositories.

### 📝 Changes

- Introduction of `MessageRepository`

### 🧪 Testing

_Explain how this change can be tested (or why it can't be tested)_

### ☑️ Contributor Checklist


Sources/StreamChat/Repositories/MessageRepository_Tests.swift
```diff
@@ -0,0 +1,347 @@
...
+    // MARK: saveSuccessfullySentMessage
+
+    func test_saveSuccessfullySentMessage_noChannel() {
+        let id = MessageId.unique
+        let payload = MessagePayload.dummy(messageId: id, authorUserId: .anonymous, channel: nil)
+        let message = runSaveSuccessfullySentMessageAndWait(payload: payload)
+        XCTAssertNil(message)
+    }
+
+    func test_saveSuccessfullySentMessage_channelPayload_sending() throws {
```
martinmitrevski, line 155: The following few are very similar tests. Might be good to extract re-use the logic and only pass the local state as a parameter.

Sources/StreamChat/Repositories/OfflineRequestsRepository.swift
```diff
@@ -101,8 +107,36 @@ class OfflineRequestsRepository {
         data: Data,
         completion: @escaping () -> Void
     ) {
-        // TODO:
-        completion()
+        func decodeTo<T: Decodable>(_ type: T.Type) -> T? { try? JSONDecoder.stream.decode(T.self, from: data) }
```
martinmitrevski, line 110: IMO One-liner is not really readable here.

Sources/StreamChat/Repositories/OfflineRequestsRepository.swift
```diff
@@ -101,8 +107,36 @@ class OfflineRequestsRepository {
...
+
+        switch endpoint.path {
+        case .createChannel:
+            guard let payload = decodeTo(ChannelPayload.self) else {
```
martinmitrevski, line 114: Just a summary of this case - If the decoding fails we delete the queued request and the change is not applied. In which scenarios can this happen?

Sources/StreamChat/Repositories/MessageRepository.swift
```diff
@@ -0,0 +1,159 @@
...
+
+                let endpoint: Endpoint<MessagePayload.Boxed> = .sendMessage(cid: cid, messagePayload: requestBody)
+                self?.apiClient.request(endpoint: endpoint) {
+                    switch $0 {
+                    case let .success(payload):
+                        self?.saveSuccessfullySentMessage(cid: cid, message: payload.message) { message in
+                            if let message = message {
+                                completion(.success(message))
+                            } else {
+                                completion(.failure(.failedToSendMessage))
```
martinmitrevski, line 72: Some logging would be nice here too.

Sources/StreamChat/Repositories/OfflineRequestsRepository_Tests.swift
```diff
@@ -46,11 +48,109 @@ final class OfflineRequestsRepository_Tests: XCTestCase {
         XCTAssertEqual(apiClient.recoveryRequest_allRecordedCalls.count, 1)
         let pendingRequests = QueuedRequestDTO.loadAllPendingRequests(context: database.viewContext)
         XCTAssertEqual(pendingRequests.count, 0)
+        XCTAssertNotCall("saveSuccessfullySentMessage(cid:message:completion:)", on: messageRepository)
+    }
+
+    func test_runQueuedRequestsWithPendingRequests_createChannel() throws {
```
martinmitrevski, line 54: Few comments on the tests would be nice too.

martinmitrevski (commented):
Nice work 👍 ! I've added few comments and questions.

Sources/StreamChat/Repositories/MessageRepository.swift
```diff
@@ -0,0 +1,159 @@
...
+        cid: ChannelId,
+        message: MessagePayload,
+        completion: @escaping (ChatMessage?) -> Void
+    ) {
+        var messageModel: ChatMessage?
+        database.write({
+            guard let messageDTO = try $0.saveMessage(payload: message, for: cid, syncOwnReactions: false) else {
+                return
+            }
+            if messageDTO.localMessageState == .sending || messageDTO.localMessageState == .sendingFailed {
```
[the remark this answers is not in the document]
polqf: Yep, because you might be retrying a message that failed to be sent

Sources/StreamChat/Repositories/OfflineRequestsRepository.swift
polqf: It should actually not happen at all. Because we know the path of the request, and the return type is always the same (or at least it should), it should never happen. If the response were to come malformed, that could be this one case.

Sources/StreamChat/Repositories/MessageRepository.swift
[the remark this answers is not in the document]
martinmitrevski: Maybe the method's name is a bit misleading, since successfully sent implies that everything went fine.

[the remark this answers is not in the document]
polqf: We'll, that's actually the case. If a message was failed, but now you retry it and it gets sent, it was successfully sent 😄

Sources/StreamChat/Repositories/MessageRepository.swift
polqf: Logging is produced by `saveSuccessfullySentMessage`

Sources/StreamChat/Repositories/MessageRepository_Tests.swift
polqf: I tried, but then the whole "script" of the test is lost. No Give-When-Then kinda way of reading it and/or the function that extracts it gets way too many parameters

martinmitrevski (approved):

PR #1812, opened 2022-02-23
[CIS-1522] Fix Memory Leaks when opening and closing channels

nuno-vieira:
### 🔗 Issue Link
CIS-1522

### 🎯 Goal
The main goal was to fix the memory leaks around opening and closing channels which was mostly caused by the `GalleryVC`, but an overall cleanup of the memory leaks on the SDK was done as well.

### 🛠 Implementation
Break reference cycles in multiple places of the codebase.

### 📝 Changes
- Make sure to invalidate the `URLSession` in the `URLWebSocketEngine`. According to Apple, the `URLSession.delegate` is strongly retained and we need to make sure we call `invalidateAndCancel()` to clean up the `URLSession`.
- Fix Ref Cycle in `GalleryVC`.
- Fix multiple ref cycles in  Updaters and Middlewares.
- Correctly unsubscribe from video player observers.
- Make sure to destroy the `ChatClient` in the `LoginViewController` so that is easier to analyse memory leaks
- Improve the feedback from the connection banner of the DemoApp

### 🎨 Changes
These are the instances that are retained after playing with the application and then coming back to the login screen before the fixes and after the fixes.
| Before | After |
| --- | --- |
| image | image |

### 🧪 Testing
- Run the DemoApp
- Select a User
- Open a Channel
- Scroll and open some images and videos
- Close a Channel
- Repeat the last 3 steps multiple times
- Logout
- Open the Memory Graph and observe the live instances

### ☑️ Contributor Checklist


Sources/StreamChat/WebSocketClient/Engine/URLSessionWebSocketEngine.swift
```diff
@@ -65,46 +79,75 @@ class URLSessionWebSocketEngine: NSObject, WebSocketEngine, URLSessionDataDelega
...
+    deinit {
+        disconnect()
+    }
+}
     
+@available(iOS 13, *)
+class URLSessionDelegateHandler: NSObject, URLSessionDataDelegate, URLSessionWebSocketDelegate {
+    var didOpenWith: ((_ protocol: String?) -> Void)?
+    var didCloseWith: ((_ code: URLSessionWebSocketTask.CloseCode, _ reason: Data?) -> Void)?
+    var didCompleteWith: ((Error?) -> Void)?
```
polqf, line 131: Should we name them `onCompletion`/`onClose`/ etc? The current naming is imperative, while we are just creating blocks for later execution
We might need to standarize this naming convention, as we are using different patterns now.

Sources/StreamChat/WebSocketClient/Engine/URLSessionWebSocketEngine.swift
```diff
@@ -65,46 +79,75 @@ class URLSessionWebSocketEngine: NSObject, WebSocketEngine, URLSessionDataDelega
...
+
+            self?.callbackQueue.async { [weak self] in
+                self?.delegate?.webSocketDidDisconnect(error: WebSocketEngineError(error: error))
+            }
+        }
+
+        return urlSessionDelegateHandler
+    }
+
+    deinit {
```
polqf, line 122: Shouldn't the deinit be close (or above) the init?

Sources/StreamChat/WebSocketClient/Engine/URLSessionWebSocketEngine.swift
```diff
@@ -5,40 +5,54 @@
...
         task?.resume()
     }
     
     func disconnect() {
         task?.cancel(with: .normalClosure, reason: nil)
+        session?.invalidateAndCancel()
+
+        session = nil
+        task = nil
+        urlSessionDelegateHandler = nil
```
polqf, line 55: No need to nil these values given the current code

Sources/StreamChat/WebSocketClient/Engine/URLSessionWebSocketEngine.swift
nuno-vieira: Yes exactly, makes total sense, will change this 👍

Sources/StreamChat/WebSocketClient/Engine/URLSessionWebSocketEngine.swift
nuno-vieira: I usually use it at the end of the classes. In the current project, this is what we do more often, although in some places I've seen it closer to the `init`. For me I'm fine with both approaches if we keep it consistent across the SDK, maybe let's bring this discussion to a tech meeting to discuss this style guide? At least from what I can found this is not mentioned in any style guide.

Sources/StreamChat/WebSocketClient/Engine/URLSessionWebSocketEngine.swift
nuno-vieira: Unfortunately, it is. There are some leaks inside the Foundation library when using the WebSockets API, so this makes sure that at least our objects are freed from memory.

bielikb (approved):

bielikb:
@nuno-vieira codecov reports quite huge decrease in the test coverage.

Sources/StreamChat/WebSocketClient/Engine/URLSessionWebSocketEngine.swift
polqf: +1 on the Style Guide. We might need to start documenting some standards 👌 ! Let's ignore my comment for now, then

Sources/StreamChat/WebSocketClient/Engine/URLSessionWebSocketEngine.swift
polqf: Oh true, forgot how amazingly URLSession was built 😂

polqf (approved):

PR #1824, opened 2022-03-03
[CIS-1627] Fix MessageSearchController not respecting sorting

### LINKED ISSUE #1821: ChatMessageSearchController's sorting is not working
## What did you do?
I searched message by using ChatMessageSearchController
```Swift
let query = MessageSearchQuery(
    channelFilter: .equal(.cid, to: cid),
    messageFilter: .and([.autocomplete(.text, text: query)]),
    sort: [.init(key: .id, isAscending: false)]
  )
messageSearchController = client.messageSearchController()
messageSearchController.search(query: query)
```

I expected the results to be sorted in descending order of created time

## What happened instead?
The results were the same(in ascending order) regardless of the `sort` (`sort: [ .init(key: .id, isAscending: false)]` and `sort: [ .init(key: .id, isAscending: true)]` result was same)
Am I using the wrong query????

## GetStream Environment
**GetStream Chat version:** 4.10.1
**GetStream Chat frameworks:** StreamChat
**iOS version:** 15.x, 14.x
**Swift version:** 5.5.2
**Xcode version:** 13.2.1
**Device:** all of iPhone, iPad supporting iOS 14+

b-onc:
### 🔗 Issue Link
CIS-1627
Fixes #1821

### 🎯 Goal

`MessageSearchController` has various issues:
1. It doesn't clear old search results when a new search is made
1. It doesn't respect `sort` passed from user
1. It's not unit tested
1. It makes empty search queries which results in 400 error returned from BE

### 🛠 Implementation

Implementation is now very similar to `UserSearchController`, with `sort` obtained from `query.sort` and `UpdatePolicy` reused.

### 📝 Changes

Unit tests are added, sort param is now respected, search results are now cleared, and empty searches are avoided.

### 🧪 Testing

There's a search option in `SwiftUI` demo (in `SampleApp`) if anyone wants interactive testing.

### ☑️ Contributor Checklist


Sources/StreamChat/Controllers/SearchControllers/MessageSearchController/MessageSearchController.swift
```diff
@@ -162,19 +171,25 @@ public class ChatMessageSearchController: DataController, DelegateCallable, Data
...
         query.filterHash = explicitFilterHash
         
         lastQuery = query
         
-        messageUpdater.search(query: query) { error in
+        // To respect sorting the user passed, we must reset messagesObserver
+        state = .initialized
+        setMessagesObserver()
+        _messagesObserver.reset()
+        startObserversIfNeeded()
```
nuno-vieira, line 190: Maybe we can extract this to an helper function `resetMessageObserver()` ?

Sources/StreamChat/Controllers/SearchControllers/MessageSearchController/MessageSearchController_Tests.swift
```diff
@@ -0,0 +1,649 @@
+//
+// Copyright © 2022 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+@testable import StreamChat
+@testable import StreamChatTestTools
+import XCTest
+
+class MessageSearchController_Tests: XCTestCase {
```
nuno-vieira, line 10: Nice!

Sources/StreamChat/Controllers/SearchControllers/MessageSearchController/MessageSearchController_Tests.swift
```diff
@@ -0,0 +1,649 @@
...
+    override func tearDown() {
+        controllerCallbackQueueID = nil
+        
+        env.messageUpdater?.cleanUp()
+        
+        AssertAsync {
+            Assert.canBeReleased(&controller)
+            Assert.canBeReleased(&client)
+            Assert.canBeReleased(&env)
+        }
```
nuno-vieira, line 46: I would be careful in using this. When I fixed some memory leaks in the previous PR, this `Assert.canBeReleased()` function was not actually catching the memory leak.

Sources/StreamChat/Controllers/SearchControllers/MessageSearchController/MessageSearchController_Tests.swift
```diff
@@ -0,0 +1,649 @@
...
+        
+        let filterHash = controller.query.filterHash
+        // Deallocate controller
+        controller = nil
+        
+        // Assert query doesn't exist in DB anymore
+        AssertAsync.willBeNil(client.databaseContainer.viewContext.messageSearchQuery(filterHash: filterHash))
+        
+        // Assert the message is still here
+        AssertAsync.staysTrue(message != nil)
```
nuno-vieira, line 94: If our goal in the future is to move away from `AssertAsync` I think it would be better to stop using it for new tests. Not going to ask you to re-write all these tests ofc, but I think it is worth keeping that in mind for the future.

Sources/StreamChat/Query/MessageSearchQuery.swift
```diff
@@ -4,9 +4,52 @@
...
     case relevance
+    
+    /// Sort messages by their `id`.
     case id
+    
+    /// Sort messages by their `created_at` dates.
+    case createdAt
+    
+    /// Sort messages by their `updated_at` dates.
+    case updatedAt
```
nuno-vieira, line 20: Isn't this a breaking change? 🤔 Well, in practice, no one should be using a switch case for this... This is a hint that this should be a Static String like we use already in other places.

Sources/StreamChat/Workers/MessageUpdater.swift
```diff
@@ -563,11 +563,16 @@ class MessageUpdater: Worker {
         })
     }
     
-    func search(query: MessageSearchQuery, completion: ((Error?) -> Void)? = nil) {
+    func search(query: MessageSearchQuery, policy: UpdatePolicy = .merge, completion: ((Error?) -> Void)? = nil) {
```
nuno-vieira, line 566: Any reason for the default argument? Since this is an internal API, the default argument might hide an important decision in a Higher Level API.

nuno-vieira (commented):
Looks good to me Overall 👍 Just added some questions and small comments, after that should be ✅

Sources/StreamChat/Controllers/SearchControllers/MessageSearchController/MessageSearchController.swift
```diff
@@ -141,13 +141,22 @@ public class ChatMessageSearchController: DataController, DelegateCallable, Data
             completion?(ClientError.CurrentUserDoesNotExist("For message search with text, a current user must be logged in"))
             return
         }
+        // We don't choose `guard` here since the expression
+        // guard !text.isEmpty, lastQuery != nil else { ...clearSearch(for: lastQuery!).. }
+        // would be much more complex to parse than this `if` statement
```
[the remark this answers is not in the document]
b-onc: I don't think it's very apparent here, so I decided to comment the reasoning. There's nothing wrong with decision comments IMHO

Sources/StreamChat/Controllers/SearchControllers/MessageSearchController/MessageSearchController_Tests.swift
b-onc: Here, it works as expected

Sources/StreamChat/Controllers/SearchControllers/MessageSearchController/MessageSearchController_Tests.swift
b-onc: Tests are mostly copy/paste from `UserSearchController_Tests`

Sources/StreamChat/Query/MessageSearchQuery.swift
b-onc: ¯\_(ツ)_/¯

Sources/StreamChat/Workers/MessageUpdater.swift
b-onc: To mirror `UserListUpdater`, no strong reason here

Sources/StreamChat/Query/MessageSearchQuery.swift
nuno-vieira: No need to change this tho, don't think it is blocking 👍

Sources/StreamChat/Controllers/SearchControllers/MessageSearchController/MessageSearchController_Tests.swift
nuno-vieira: I see, makes sense then

nuno-vieira (approved):
LGTM! ✅ Would be nice to extract the `resetMessagesObserver()`. But other than that, LGTM.

PR #1829, opened 2022-03-07
Implement missing syntesized Codable methods for EndpointPath on Swift < 5.5

polqf:
### 🎯 Goal

Offline Support's Head branch was failing on Xcode 12. This was due to the fact that before Swift 5.5 we did not have syntesized Codable implementations for enumerations with associated objects.

See proposal here: https://github.com/apple/swift-evolution/blob/main/proposals/0295-codable-synthesis-for-enums-with-associated-values.md

### ☑️ Contributor Checklist

- [x] I have signed the [Stream CLA](https://docs.google.com/forms/d/e/1FAIpQLScFKsKkAJI7mhCr7K9rEIOpqIDThrWxuvxnwUq2XkHyG154vQ/viewform) (required)
- [x] Changelog is updated with client-facing changes
- [x] New code is covered by unit tests
- [x] This change follows zero ⚠️ policy (required)
- [ ] Comparison screenshots added for visual changes
- [ ] Affected documentation updated (docusaurus, tutorial, CMS)

Issue #1794, opened 2022-02-15
ThreadList tap state (row highlight state)  doesn't go away in airplane mode


linhewinner:
https://user-images.githubusercontent.com/97132583/154114497-77afc1ec-9c43-4bdf-b031-ab89a3b834e2.mp4

## What did you do?
Turn on airplane mode. Tap on a row in Thread List. Navigate back.

## What did you expect to happen?
The tap state of the row should go away.

## What happened instead?
The tap state of the row doesn't go away (the tapped row is still displayed as highlighted).
Note that this only happens in airplane mode. When network is healthy, this bug doesn't reproduce.

## GetStream Environment
**GetStream Chat version:** 4.10.0
**GetStream Chat frameworks:** StreamChat, StreamChatUI
**iOS version:** 15.3
**Swift version:** 5.5.2
**Xcode version:** 13.2.1
**Device:** iPhone 13 Pro

nuno-vieira:
Hi @linhewinner!

Thank you for the report! We will investigate this as soon as possible and will update you whenever we have a fix. 

Best,
Nuno

polqf:
This issue will be fixed once https://github.com/GetStream/stream-chat-swift/pull/1831 is merged.
You can expect this to be included in the following release, most probably.

polqf:
Hello! We just released 4.12.0, you can check the changelog [here](https://github.com/GetStream/stream-chat-swift/releases/tag/4.12.0).
This issue should be fixed now. Feel free to reopen it if needed

linhewinner:
confirmed that this is fixed. thanks!

PR #1831, opened 2022-03-08
[CIS-1275] Offline Support v1

polqf:
### 🔗 Issue Link
https://stream-io.atlassian.net/browse/CIS-1550
Bonus: https://github.com/GetStream/stream-chat-swift/issues/1794

### 🎯 Goal

Goals:
- Connection Recovery
- Events Recovery
- Queue Offline Actions

PRs contributing to it:
- #1623
- #1630
- #1653
- #1654 
- #1673 
- #1754 
- #1776 
- #1789 
- #1788 
- #1791 
- #1810 
- #1811 
- #1813 
- #1829 

### 🛠 Implementation

https://www.notion.so/stream-wiki/Connection-recovery-d3152716f9654d85b9e417e15a19c71b
https://www.notion.so/stream-wiki/Offline-Actions-Queuing-da47cdbcf3c04485b9e551dd48735dd0

### 📝 Changes

- Properly recover from downtimes (both connection and events)
- Enqueue requests that happen while there is no connection

### 🧪 Test plan:

https://www.notion.so/stream-wiki/Offline-Support-Testing-Plan-2afe1518bdf546878766568907d74619

### ☑️ Contributor Checklist


Sources/StreamChat/Controllers/SearchControllers/UserSearchController/UserSearchController.swift
```diff
@@ -195,16 +96,129 @@ public class ChatUserSearchController: DataController, DelegateCallable, DataSto
...
+        // TODO: Remove with the next major
+        //
+        // This is needed to make the delegate fire about state changes at the same time with the same
+        // values as it was when query was persisted.
+        setLocalDataFetchedStateIfNeeded()
+        
+        userQueryUpdater.fetch(userListQuery: query) { result in
+            switch result {
+            case let .success(page):
+                self.save(page: page) { loadedUsers in
```
bielikb, line 127: Possible memory leak

Sources/StreamChat/Controllers/SearchControllers/UserSearchController/UserSearchController.swift
```diff
@@ -195,16 +96,129 @@ public class ChatUserSearchController: DataController, DelegateCallable, DataSto
...
+                        loadedPage: loadedUsers,
+                        updatePolicy: query.pagination?.offset == 0 ? .replace : .merge
+                    )
+                    
+                    self.query = query
+                    self._users = self.userList(after: listChanges)
+                    self.state = .remoteDataFetched
+                    
+                    self.callback {
+                        self.multicastDelegate.invoke { $0.controller(self, didChangeUsers: listChanges) }
```
bielikb, line 138: Possible memory leak

Sources/StreamChat/Utils/EventBatcher.swift
```diff
@@ -0,0 +1,79 @@
...
+        handler: @escaping ([Item]) -> Void
+    ) {
+        self.period = max(period, 0)
+        self.timerType = timerType
+        self.handler = handler
+    }
+    
+    func append(_ item: Item) {
+        timerType.schedule(timeInterval: 0, queue: queue) {
+            self.currentBatch.append(item)
```
bielikb, line 55: Possible memory leak

Sources/StreamChat/Utils/EventBatcher.swift
```diff
@@ -0,0 +1,79 @@
...
+                timeInterval: self.period,
+                queue: self.queue,
+                onFire: self.process
+            )
+        }
+    }
+    
+    func processImmediately() {
+        timerType.schedule(timeInterval: 0, queue: queue) {
+            self.process()
```
bielikb, line 69: Possible memory leak

Sources/StreamChat/Workers/Background/ConnectionRecoveryHandler.swift
```diff
@@ -0,0 +1,257 @@
...
+        
+        switch state {
+        case .connecting:
+            cancelReconnectionTimer()
+            
+        case .connected:
+            reconnectionStrategy.resetConsecutiveFailures()
+            syncRepository.syncLocalState {
+                log.info("Local state sync completed", subsystems: .offlineSupport)
+            }
```
bielikb, line 158: This part needs to be improved and should be planned for offline support to be 100% reliable. 

The only indicator about the connectivity is the `.connected` status. At the same time, we fire another `async` operation to synchronize local state with the remote.

This creates the "buggy" window, in which the integrator might start mutating local state while it's being synced with the remote.

Since the synchronized state is not given to us at the `connected` state, I'd suggest to treat the "connecteded" state that the integrator can listen to as the "websocket.connected && local.state synchronised"

Sources/StreamChat/Workers/ChannelMemberListUpdater.swift
```diff
@@ -54,7 +54,7 @@ private extension ChannelMemberListUpdater {
     
     func fetchAndSaveChannel(with cid: ChannelId, completion: @escaping (Error?) -> Void) {
         let query = ChannelQuery(cid: cid)
-        apiClient.request(endpoint: .channel(query: query)) {
+        apiClient.request(endpoint: .updateChannel(query: query)) {
             switch $0 {
             case let .success(payload):
                 self.database.write({ session in
```
bielikb, line 60: looks like memory leak too

bielikb (commented):
Few comments from me

Sources/StreamChat/Workers/Background/ConnectionRecoveryHandler.swift
polqf: While you are right on one side, this has massive implications. Also, bear in mind that the thing that we would end up doing is we would be queuing DB actions (because API calls are already queued, and those undo DB changes if the request then fails). In the case of DB actions, it does not really matter that much when those are done, because they are all going to get through when the queue gets executed, there is no backend that can tell you what can/needs to be done when launching them.
Also, if you you were to do that, we would need to block the whole UI from any kind of actions. This is something that would need a lot of changes in the LLC as well.

All in all, I completely get your point, but TBH I don't see it a major issue (Maybe we discover it is, don't get me wrong, but it doesn't seem to be the case now)

Sources/StreamChat/APIClient/APIClient.swift
```diff
@@ -64,92 +66,113 @@ class APIClient {
...
-        if tokenRefreshConsecutiveFailures > maxTokenRefreshAttempts {
+        if tokenRefreshConsecutiveFailures > maximumTokenRefreshAttempts {
             return completion(.failure(ClientError.TooManyTokenRefreshAttempts()))
         }
 
-        if isRefreshingToken {
-            return requeueRequestOnTokenExpired(endpoint: endpoint, timeout: timeout, completion: completion)
+        guard !isRefreshingToken else {
+            completion(.failure(ClientError.RefreshingToken()))
+            return
```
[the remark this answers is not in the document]
polqf: What do you mean we lose track of those? This request will be enqueued, and when the refresh token is retrieved, it will get executed.

Sources/StreamChat/APIClient/APIClient.swift
```diff
@@ -182,94 +207,103 @@ class APIClient {
...
+                    }
+
+                    /// If the error is ExpiredToken, we need to refresh it. There are 2 possibilities here:
+                    /// 1. The token is not being refreshed, so we start the refresh, and we wait until it is completed. Then the request will be retried.
+                    /// 2. The token is already being refreshed, so we just put back the request to the queue (Cannot happen when running the queue in serial)
+                    ///
+                    /// This is done leveraging 2 error types. When ClientError.RefreshingToken is returned, we put back the request on the queue.
+                    /// But when ClientError.TokenRefreshed is returned, just retry the execution.
+                    /// This is done because we want to make sure that when the queue is running serial, there order is kept.
+                    self.refreshToken { refreshResult in
```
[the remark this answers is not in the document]
polqf: Not the case, check line 194

Sources/StreamChat/Controllers/SearchControllers/UserSearchController/UserSearchController.swift
polqf: Whilst those should not be leaks, I will address your comments for now. We might need to add some conversation wrt these. Weakifying everywhere is "fine", but also adds some overhead

DemoApp/CreateChatViewController.swift
```diff
@@ -20,8 +20,10 @@ class CreateChatViewController: UIViewController {
                   let controller = channelController else { return }
             // Create the Channel on backend
             controller.synchronize { error in
-                // TODO: handle error
-                if let error = error { print("###", error) }
+                if let error = error {
+                    self.presentAlert(title: "Error when creating the channel", message: error.localizedDescription)
```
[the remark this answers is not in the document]
polqf: This is why I did it like that:`Using 'super' in a closure where 'self' is explicitly captured is not yet supported`
But it is true that this is in our demo, and we should fix it

Sources/StreamChat/APIClient/APIClient.swift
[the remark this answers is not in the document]
bielikb: Oh I see now. thanks

Sources/StreamChat/Workers/Background/ConnectionRecoveryHandler.swift
bielikb: imo this needs follow-up. It's behaviour undefined and should be addressed to strengthen LLC Foundation.

Sources/StreamChat/Controllers/SearchControllers/UserSearchController/UserSearchController.swift
nuno-vieira: It is safer just to default weakify every closure. Especially in an SDK, we can't afford to have crashes or memory leaks, so we should try to do our best to avoid this. Also it future proofs possible refactorings, since right now it might not leak, but if some refactoring happens in the Updater or something like this, that could change.

Sources/StreamChat/Controllers/SearchControllers/UserSearchController/UserSearchController.swift
```diff
@@ -22,72 +22,29 @@ public class ChatUserSearchController: DataController, DelegateCallable, DataSto
...
+    public private(set) var query: UserListQuery?
     
-    /// The users matching the query of this controller.
-    ///
-    /// To observe changes of the users, set your class as a delegate of this controller or use the provided
-    /// `Combine` publishers.
-    ///
-    public var users: LazyCachedMapCollection<ChatUser> {
-        startUserListObserverIfNeeded()
-        return userListObserver.items
```
nuno-vieira, line 50: This is a breaking change 🔴  Why are we removing this one? We need at least to deprecate it, not remove it completely.

nuno-vieira (dismissed):
I'm still finishing the review, but I will request changes for now, since I spotted a breaking change 👍

Sources/StreamChat/Controllers/SearchControllers/UserSearchController/UserSearchController.swift
```diff
@@ -22,72 +22,29 @@ public class ChatUserSearchController: DataController, DelegateCallable, DataSto
...
-        // The initial DB fetch will return 0 users and this is expected
-        // In the future we'll implement DB search too
-        query.filter?.explicitHash = explicitFilterHash
-        
-        return query
-    }()
-    
     /// Copy of last search query made, used for getting next page.
-    var lastQuery: UserListQuery?
+    public private(set) var query: UserListQuery?
```
nuno-vieira, line 26: Any reason why is this public now? And before it was internal? 🤔

Sources/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -402,6 +402,7 @@ extension NSManagedObjectContext: MessageDatabaseSession {
         if let parentMessageId = parentMessageId,
            let parentMessageDTO = MessageDTO.load(id: parentMessageId, context: self) {
             parentMessageDTO.replies.insert(message)
+            parentMessageDTO.replyCount += 1
```
nuno-vieira, line 405: Interesting, I wonder if this fixes the following issue: https://github.com/GetStream/stream-chat-swift/issues/1806

Worth to double-check 🙂

Sources/StreamChat/WebSocketClient/RetryStrategy.swift
```diff
@@ -0,0 +1,65 @@
+//
+// Copyright © 2022 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+/// The type encapsulating the logic of computing delays for the failed actions that needs to be retried.
+protocol RetryStrategy {
+    /// Rerturns the # of consecutively failed retries.
```
nuno-vieira, line 9:
```suggestion
    /// Returns the # of consecutively failed retries.
```

Sources/StreamChat/Workers/Background/NewUserQueryUpdater.swift
```diff
@@ -14,59 +14,56 @@ import CoreData
 ///     corresponding queries to the user.
 final class NewUserQueryUpdater: Worker {
```
nuno-vieira, line 15: Any reason why we have only Updaters, but they are actually in a "Workers" folder, and each updater is a "Worker". Can't we just rename "Worker" to "Updater"? @b-onc

Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
```diff
@@ -199,6 +199,9 @@ open class ChatChannelListVC: _ViewController,
     }
         
     open func collectionView(_ collectionView: UICollectionView, didSelectItemAt indexPath: IndexPath) {
+        defer {
+            collectionView.deselectItem(at: indexPath, animated: true)
+        }
```
nuno-vieira, line 204: Does this fixes this issue: https://github.com/GetStream/stream-chat-swift/issues/1794 ?

nuno-vieira (commented):
Finished the overall Review. Looks really good overall ✅  Just left a couple of comments 👍

Sources/StreamChat/Database/DTOs/MessageDTO.swift
nuno-vieira: If yes, can we add this fix to the changelog?

Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
nuno-vieira: If yes, can we add this fix to the changelog?

Sources/StreamChat/Database/DTOs/MessageDTO.swift
polqf: While they may be connected, in #1806 I believe the issue is around Publishers itself.
It is a pretty common mistake/misconception where Publishers send the value prior to the change.

I will look for a post describing this scenario

Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
polqf: It does, yeah! Thanks for spotting it 😄

Sources/StreamChat/Database/DTOs/MessageDTO.swift
polqf: Here it is: https://stackoverflow.com/questions/58403338/is-there-an-alternative-to-combines-published-that-signals-a-value-change-afte

polqf: Added a comment there: https://github.com/GetStream/stream-chat-swift/issues/1806#issuecomment-1064969997

Sources/StreamChat/Controllers/SearchControllers/UserSearchController/UserSearchController.swift
polqf: I moved the deprecation back to this file, as agreed.

Sources/StreamChat/Controllers/SearchControllers/UserSearchController/UserSearchController.swift
polqf: Moved it back to be internal

Sources/StreamChat/Database/DTOs/ChannelListQueryDTO.swift
```diff
@@ -16,14 +16,20 @@ class ChannelListQueryDTO: NSManagedObject {
     
     @NSManaged var channels: Set<ChannelDTO>
     
-    /// Channels that are being read by the current user (the subset of `channels`)
-    @NSManaged var openChannels: Set<ChannelDTO>
-
```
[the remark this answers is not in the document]
polqf: We are tracking this using ChatClient's `activeChannelControllers`

Sources/StreamChat/Workers/Background/NewUserQueryUpdater.swift
polqf: No clue. I cannot answer that question, sorry 😐

bielikb (approved):

martinmitrevski (changes requested):
Great work! Just one minor thing: the deprecation of the `user` array should be public, otherwise it's a breaking change.

Sources/StreamChat/Controllers/SearchControllers/UserSearchController/UserSearchController.swift
```diff
@@ -22,72 +22,34 @@ public class ChatUserSearchController: DataController, DelegateCallable, DataSto
...
+    /// The users matching the last query of this controller.
+    private var _users: [ChatUser] = []
+    public var userArray: [ChatUser] {
+        setLocalDataFetchedStateIfNeeded()
+        return _users
     }
-    
+
+    @available(*, deprecated, message: "Please, switch to `userArray: [ChatUser]`")
+    var users: LazyCachedMapCollection<ChatUser> {
```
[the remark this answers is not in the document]
polqf: Good catch!

martinmitrevski (approved):
LGTM ✅ 🎉

PR #1844, opened 2022-03-11
[CIS-1635] Make it possible to customize the message view only in the popup actions

nuno-vieira:
### 🔗 Issue Link
CIS-1635

### 🎯 Goal
Right now is not possible to provide customization to the message content view only in the context of the popup actions.

### 🛠 Implementation
We extract the logic to create the message content view copy to a separate function so that it can be easily changed.

### 🧪 Testing
One way to test this, is to provide a custom actions transition controller and for example remove the timestamp of the message content view:

```swift
class CustomMessageActionsTransitionController: ChatMessageActionsTransitionController {
    override func makeMessageContentView(fromOriginalView originalView: ChatMessageContentView) -> ChatMessageContentView {
        let messageContentView = super.makeMessageContentView(fromOriginalView: originalView)
        messageContentView.layoutOptions?.remove(.timestamp)
        return messageContentView
    }
}

Components.default.messageActionsTransitionController = CustomMessageActionsTransitionController.self
```

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChatUI/MessageActionsPopup/MessageActionsTransitionController.swift
+++ b/Sources/StreamChatUI/MessageActionsPopup/MessageActionsTransitionController.swift
@@ -85,18 +85,11 @@ open class ChatMessageActionsTransitionController: NSObject, UIViewControllerTra
         else { return }
         
         selectedMessageId = originalMessageContentView.content?.id
-        
+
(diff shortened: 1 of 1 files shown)


martinmitrevski (approved):
LGTM ✅

PR #1866, opened 2022-03-23
[CIS-1523] Create mock web socket client for UI tests

bielikb:
### 🔗 Issue Link
https://stream-io.atlassian.net/browse/CIS-1523
https://stream-io.atlassian.net/browse/CIS-1654
https://stream-io.atlassian.net/browse/CIS-1656
https://stream-io.atlassian.net/browse/CIS-1657

### 🎯 Goal

Bring basic testing infra for UI tests. Please note, that these UI tests are not yet incorporated into our CI.
Testing infra is not done yet and will be gradually improved in upcoming tasks. At the same time, this gives us the possibility to start building automation on a solid foundation.

### 🛠 Implementation

In this PR, we covered several points:

**Chat UI Tests App**
A new UI testing app was added to the project. The purpose of this app is to be used for UI tests only.
UI tests

### Mock Server

**StreamMockServer**
New mocked web socket server runs & acts as actual backend once the UI tests are run. The spoofed web server is run on localhost & will respond/send predefined requests to the running app.
In order to achieve this functionality, we brought [Swifter](https://github.com/httpswift/swifter) as a new dependency to the testing target of our UI test app.

 **Mocked Responses**
To imitate a real API we captured and parametrized `mock responses` for API requests. They act as a template for generating both HTTP and WebSocket responses.

Examples:
* `http_message.json` is used to send messages
* `ws_reaction.json` is used to receive reactions, and so on.

#### 💪 Mock server current capabilities

- Read message
- Send message
- Edit message
- Delete message
- Receive message
- Add reaction
- Remove reaction
- Receive reaction
- Observe participant starts typing
- Observe participant stops typing
- Observe participant edits their message
- Observe participant deletes their message
- Observe participant removes their reaction
 
#### Launch Arguments
To run mocked web socket server, we had to modify the production code.
When `BaseURL` is initialized, we inject the web socket server configuration only if these two following conditions are met:
1. App runs in `Debug` build configuration
2. `USE_MOCK_SERVER` launch argument was passed at the launch of the app. 

Once these two conditions are met, we can start leveraging the power of our `StreamMockServer.`

#### Robot pattern
This work declares three robots (`UserRobot` - acts as end-user, `ParticipantRobot` - serves as another participant in Chat, and `DeviceRobot` - allows us to cover cases such as going to background, foreground, rotating the device).

Example of test written in Robot pattern leveraging the `GIVEN` `WHEN` `THEN` notation.
```swift
func testReceiveMessage() throws {
    let message = "test message"
    let author = "Han Solo"
    
    GIVEN("user opens the channel") {
        userRobot.login().openChannel()
    }
    WHEN("participant sends the message: '\(message)'") {
        participantRobot
            .startTyping()
            .stopTyping()
            .sendMessage(message)
    }
    THEN("the message is delivered") {
        userRobot
            .waitForParticipantsMessage()
            .assertMessageAuthor(author)
    }
}
```
 
#### Pages
To identify & select UI elements on the screen, we declared `pages.` These act as descriptions of the UI elements shown on the UI flow screen. `ChannelListPage` - Channel List, `MessageListPage` - Message List, `ThreadPage` and so on.
 
#### Accessibility Identifiers
To be able to test the `StreamChatUI` SDK, we added accessibility labels to the UI elements.
There are 2 main ways how assigning accessibility labels can happen (to cover all 3ish use cases).

1. Via protocol conformance - this is applied to all SDK's UI Views, Controls, etc. By default, the accessibility identifier matches the name of the class.
```swift
// Protocol that provides accessibility features
protocol AccessibilityView {
    // Identifier for view
    var accessibilityViewIdentifier: String { get }

    // This function is called once the view is being added to the view hierarchy
    func setAccessibilityIdentifier()
}
```

2. Via building function - this covers mostly the case when we declare `View` that doesn't inherit from our `_View`/etc subclass. This function be also leveraged in cases when we declare more than 1 appearance of certain class - eg several container stacks in 1 view hierarchy.

```swift
func withAccessibilityIdentifier(identifier: String) -> Self {
    accessibilityIdentifier = identifier
    return self
}

// Example
open var actionButton: UIButton = UIButton()
    .withAccessibilityIdentifier(identifier: "actionButton")
```

The "chain of responsibility" in this PR currently follows rule like this:
1. If the `View` class was declared in our SDK, it will share the identifier which equals to its class name, eg `ChatMessageCell` instance will get `ChatMessageCell` accessibility identifier set by default.
2. If the `View` is UIKit based and lazily instantiated -> we leverage the `withAccessibilityIdentifier(identifier:)` helper function.

#### fastlane action
`test_e2e_mock` action allows us to run UI tests from the command line.

```
bundle exec fastlane test_e2e_mock device:"iPhone 11"
```

### Showcase
gif

### ☑️ Contributor Checklist

### 🎉 GIF

UI Tests on the way


Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
```diff
@@ -30,50 +30,63 @@ open class ChatChannelListItemView: _View, ThemeProvider, SwiftUIRepresentable {
...
         .withAdjustingFontForContentSizeCategory
         .withBidirectionalLanguagesSupport
+        .withAccessibilityIdentifier(identifier: "timestampLabel")
     
     /// The view used to show channels avatar.
     open private(set) lazy var avatarView: ChatChannelAvatarView = components
         .channelAvatarView
         .init()
         .withoutAutoresizingMaskConstraints
+        .withAccessibilityIdentifier(identifier: "avatarView")
```
martinmitrevski, line 83: What would happen if the same identifier is used more than once on the same screen? For example, for the avatar, it's easy to name it "avatarView" both as part of the channel list item, as well as in the navigation bar. Maybe we should use less generic identifiers?

Sources/StreamChat/Config/BaseURL.swift
```diff
@@ -46,6 +46,23 @@ public struct BaseURL: CustomStringConvertible {
...
         
         urlString = urlString.trimmingCharacters(in: CharacterSet(charactersIn: "/"))
+
+        // Leverages mocked web socket server when running the UI tests.
+        // Two conditions need to be met in order to leverage the web socket server.
+        //   1. App runs in Debug build configuration
+        //   2. `USE_MOCK_SERVER` is set as launch argument
+        #if DEBUG
+        if ProcessInfo.processInfo.arguments.contains("USE_MOCK_SERVER") {
+            let env = ProcessInfo.processInfo.environment
```
martinmitrevski, line 56: Maybe move this code in a separate private method?

StreamChatUITestsApp/AppDelegate.swift
```diff
@@ -0,0 +1,36 @@
...
+        // Empty root view controller
+        let rootViewController = ViewController()
+
+        /// Embed in navigation controller
+        window = UIWindow()
+        window?.rootViewController = UINavigationController(rootViewController: rootViewController)
+        window?.makeKeyAndVisible()
+    }
+    
+    func disableAnimations() {
```
martinmitrevski, line 30: Out of curiosity - why do we disable animations for the ui tests? I think they are relevant for the UI tests.

StreamChatUITestsApp/StreamChat/StreamChatWrapper.swift
```diff
@@ -0,0 +1,42 @@
+//
+// Copyright © 2022 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+import StreamChat
+import StreamChatUI
+import UIKit
+
+final class StreamChatWrapper {
```
martinmitrevski, line 10: Nice wrapper 👍  Wonder if we can/should add offline mode support for the UI tests?

StreamChatUITestsAppUITests/Extensions/XCUITest.swift
```diff
@@ -0,0 +1,265 @@
...
+        activate()
+    }
+
+    func bundleId() -> String {
+        Bundle.main.bundleIdentifier ?? ""
+    }
+
+}
+
+extension XCTest {
```
martinmitrevski, line 241: Nice one!

StreamChatUITestsAppUITests/Robots/BasicAssertions.swift
```diff
@@ -0,0 +1,63 @@
+//
+// Copyright © 2022 Stream.io Inc. All rights reserved.
+//
+
+import XCTest
+
+extension Robot {
```
martinmitrevski, line 7: Would be nice to document the methods here, since I guess we will be using them a lot.

StreamChatUITestsAppUITests/Tests/MessagingTests.swift
```diff
@@ -0,0 +1,119 @@
+//
+// Copyright © 2022 Stream.io Inc. All rights reserved.
+//
+
+import XCTest
+
+final class MessagingTests: StreamTestCase {
```
martinmitrevski, line 7: Really like the structure and ease of readability in these tests 👏

StreamChatUITestsAppUITests/Utilities/LaunchArgument.swift
```diff
@@ -0,0 +1,33 @@
...
+
+struct MockServerConfiguration {
+    static var port = Int.random(in: 61000..<62000)
+    static var websocketHost = "ws://localhost"
+    static var httpHost = "http://localhost"
+}
+
+enum EnvironmentVariable: String {
+
+    /// This changes the base url to localhost & assigns port when our LLC is built in Debug build configuration && `USE_MOCK_SERVER` launch argument is set.
```
martinmitrevski, line 15: Would be nice to break down the comment in few lines, it's hardly readable like this.

martinmitrevski (commented):
Great work @bielikb and @alteral, really like how things are structured. I've added few comments. Another concern is the increased app size of almost 700kb. Any idea why is this happening?

Sources/StreamChat/Config/BaseURL.swift
bielikb: @martinmitrevski this logic happens within constructor.  Both `restAPIBaseURL` and `webSocketBaseURL` needs to be set before the `BaseURL` is fully initialized.

One way to leverage `private func` for this logic would be by turning the `restAPIBaseURL` and `webSocketBaseURL` into readonly force unwrapped vars. I would like to avoid that :) 

Let me know your thoughts ;)

Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
bielikb: Thanks for the comment @martinmitrevski .

Unless 2 views are in the very same view hierarchy, this naming logic won't harm. 
In the other cases, when several containers were used within the same view hiereachy, we applied different accessibility identifiers.

As in, it's always possible to change the identifiers to make them unique.

DemoApp/Info.plist
```diff
@@ -2,6 +2,13 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+	<key>NSAppTransportSecurity</key>
+	<dict>
+		<key>NSAllowsLocalNetworking</key>
+		<true/>
+	</dict>
+	<key>LSApplicationCategoryType</key>
+	<string></string>
```
evsaev, line 11: Why do we need this setting for `DemoApp`? Looks like smth required only for `StreamChatUITestsApp`

StreamChatUITestsApp/AppDelegate.swift
bielikb: When running UI tests you want to optimize the time it takes to run single UI tests. Disabling animations is one of the optimizations applied. 
At the moment we don't focus on the definition/performance of the animations. I believe we will have dedicated logic/test target for that particular logic.

Sources/StreamChat/APIClient/Endpoints/Payloads/MessageReactionPayload.swift
```diff
@@ -6,7 +6,7 @@ import Foundation
 
 /// The type describes the incoming message-reaction JSON.
 struct MessageReactionPayload: Decodable {
-    private enum CodingKeys: String, CodingKey, CaseIterable {
+    internal enum CodingKeys: String, CodingKey, CaseIterable {
```
evsaev, line 9: That's a good one to be discussed. I would either always explicitly specify the `internal` access level or always skip it. I'm OK with both as long as we're have consistency
```suggestion
    enum CodingKeys: String, CodingKey, CaseIterable {
```

StreamChatUITestsApp/StreamChat/StreamChatWrapper.swift
bielikb: That part is planned in our epic (CIS-1509) - see tasks.

This PR provides basic testing infra in place that already allows us (not only me & Alex) to build a robust test suite for the most common cases.

StreamChatUITestsAppUITests/Robots/BasicAssertions.swift
bielikb: @martinmitrevski Thanks for the comment. We will add more comments to this section :)

StreamChatUITestsAppUITests/Utilities/LaunchArgument.swift
bielikb: Thanks for the comment!. Will do:)

Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
martinmitrevski: thanks for the clarification @bielikb 👍

DemoApp/Info.plist
bielikb: Good catch! :)

Sources/StreamChat/Config/BaseURL.swift
martinmitrevski: also not a fan of force unwrapping them. :) Well, in that case, either make the private func static or leave everything as it is. In any case, not a major thing to spend time on.

bielikb: Thanks. Ill keep this discussion open, it's a good comment - I also spent some time playing around alternatives prior opening this PR == Im open for suggestions :)

Sources/StreamChatUI/CommonViews/BaseViews.swift
```diff
@@ -6,6 +6,46 @@ import Foundation
...
+        "\(type(of: self))"
+    }
+
+    func setAccessibilityIdentifier() {
+        accessibilityIdentifier = accessibilityViewIdentifier
+    }
+}
+
+@propertyWrapper
+public struct AccessibleView<T: UIView> {
```
evsaev, line 29: All UI components instantiate their subviews **lazily** -> a property wrapper is not applicable for UI components except message content view but increases public API surface with smth not directly related to chat. Using internal `withAccessibilityIdentifier` in message content view feels more reasonable, tbh

Sources/StreamChat/Config/BaseURL.swift
martinmitrevski: Maybe sth like this, but not sure if it brings any additional value: 
```swift
        #if DEBUG
        if ProcessInfo.processInfo.arguments.contains("USE_MOCK_SERVER") {
            let mockServerUrls = Self.mockServerUrls(with: urlString)
            restAPIBaseURL = mockServerUrls.restAPIBaseURL
            webSocketBaseURL = mockServerUrls.webSocketBaseURL
            return
        }
        #endif
```

```swift
private static func mockServerUrls(with urlString: String) -> (restAPIBaseURL: URL, webSocketBaseURL: URL) {
        let env = ProcessInfo.processInfo.environment
        let httpHost = env["MOCK_SERVER_HTTP_HOST"] ?? "https://\(urlString)/"
        let websocketHost = env["MOCK_SERVER_WEBSOCKET_HOST"] ?? "wss://\(urlString)/"
        let port = env["MOCK_SERVER_PORT"] ?? "443"
        let restAPIBaseURL = URL(string: "\(httpHost):\(port)/")!
        let webSocketBaseURL = URL(string: "\(websocketHost):\(port)/")!
        return (restAPIBaseURL: restAPIBaseURL, webSocketBaseURL: webSocketBaseURL)
    }
```

StreamChatUITestsApp/StreamChat/ChannelListVC.swift
```diff
@@ -0,0 +1,18 @@
...
+import StreamChatUI
+
+final class ChannelListVC: ChatChannelListVC {
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        // Do any additional setup after loading the view.
+        
+        let query = ChannelListQuery.init(filter: .containMembers(userIds: [ChatClient.shared.currentUserId!]))
+        self.controller = ChatClient.shared.channelListController(query: query)
```
evsaev, line 16: That seems like a miss-use 🙂

In `super.viewDidLoad` custom VC lifecycle methods are triggered (`setUp/etc.`) where `controller` property is accessed. Since we assign `controller` after `super.viewDidLoad` it would crash because in `setUp` a controller would be nil. However, it's not crashing because `controller` is already assigned in `StreamChatWrapper.makeChannelListViewController` but here we create & assign another LLC controller instance.

The right way would be to remove this custom subclass and user factory method `ChatChannelListVC.make` in `makeChannelListViewController` that ensures `controller` is assigned before the VC is shown:

```swift
func makeChannelListViewController() -> UIViewController {
        // UI
        let query = ChannelListQuery(filter: .containMembers(userIds: [userCredentials?.id ?? ""]))
        let controller = ChatClient.shared.channelListController(query: query)
        
        return ChatChannelListVC.make(with: controller)
    }
```

StreamChatUITestsAppUITests/Utilities/LaunchArgument.swift
```diff
@@ -0,0 +1,33 @@
...
+
+extension ProcessInfo {
+    static func contains(_ argument: LaunchArgument) -> Bool {
+        processInfo.arguments.contains(argument.rawValue)
+    }
+
+    static subscript(_ environmentVariable: EnvironmentVariable) -> String? {
+        processInfo.environment[environmentVariable.rawValue]
+    }
+}
```
evsaev, line 33: Are we using those helpers?

Sources/StreamChatUI/CommonViews/BaseViews.swift
bielikb: Hi @evsaev thanks for the commentt.

 All 3 cases how we're setting the accesibility identifiers are explained in the PR description. 


"All UI components instantiate their subviews lazily" - this claim is not true. There are cases when the views are not instantiated lazily, but rather kept `nil`. In those cases the property wrapper makes sense. It also allows us to leverage similar semantics in our custom apps. I see the benefit of this solution.

StreamChatUITestsAppUITests/MockServer/MockServerAttributes.swift
```diff
@@ -0,0 +1,47 @@
...
+    static var reaction = "/messages/:message_id/reaction"
+    static var reactionUpdate = "/messages/:message_id/reaction/:reaction_type"
+    static var channels = "/channels"
+    static var query = "/channels/messaging/:channel_id/query"
+}
+
+enum MessageTypes: String {
+    case regular
+    case deleted
+}
```
evsaev, line 32: nit: I would reuse domain types from LLC (e.g. `MessageType`, `MessageReactionType`) where possible so things don't go out of sync.

StreamChatUITestsAppUITests/MockServer/MockServerAttributes.swift
```diff
@@ -0,0 +1,47 @@
...
+    static var connect = "/connect"
+    static var message = "/channels/messaging/:channel_id/message"
+    static var messageUpdate = "/messages/:message_id"
+    static var event = "/channels/messaging/:channel_id/event"
+    static var messageRead = "/channels/messaging/:channel_id/read"
+    static var reaction = "/messages/:message_id/reaction"
+    static var reactionUpdate = "/messages/:message_id/reaction/:reaction_type"
+    static var channels = "/channels"
+    static var query = "/channels/messaging/:channel_id/query"
+}
```
evsaev, line 27:
```suggestion
enum MockEndpoints {
    static let connect = "/connect"
    static let message = "/channels/messaging/:channel_id/message"
    static let messageUpdate = "/messages/:message_id"
    static let event = "/channels/messaging/:channel_id/event"
    static let messageRead = "/channels/messaging/:channel_id/read"
    static let reaction = "/messages/:message_id/reaction"
    static let reactionUpdate = "/messages/:message_id/reaction/:reaction_type"
    static let channels = "/channels"
    static let query = "/channels/messaging/:channel_id/query"
}
```

evsaev: Would be nice to reuse `EndpointPath` one time

StreamChatUITestsApp/StreamChat/User.swift
```diff
@@ -0,0 +1,28 @@
...
+// Copyright © 2021 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+import StreamChat
+
+let apiKey = "8br4watad788"
+let applicationGroupIdentifier = "group.io.getstream.iOS.StreamChatUITestsApp"
+
+public struct UserCredentials {
```
evsaev, line 11:
```suggestion
struct UserCredentials {
```

evsaev (commented):
@bielikb @alteral that's an incredible chunk of work 😍 I'm very happy with how simple it is to write a UI test, well-done guys 👏 Submitting what I've spotted so far and will have 2nd review round tomorrow morning

StreamChatUITestsAppUITests/Utilities/LaunchArgument.swift
bielikb: This code will be used in following PRs. 

It was used also for dev purposes == we will keep it in place for now == the aim of this PR is to bring basic testing infra in place.

Sources/StreamChatUI/CommonViews/BaseViews.swift
evsaev: It's only `ChatMessageCell/ChatMessageContentView` which are special, that's true for the rest of the UI codebase. Having multiple ways of doing the same thing (setting accessibility identifier in our case) is what we'd like to avoid as it's not obvious which way to choose 🙂

> similar semantics in our custom apps

Which apps do you mean?

StreamChatUITestsAppUITests/Utilities/LaunchArgument.swift
bielikb: Done ;)

StreamChatUITestsApp/StreamChat/User.swift
bielikb: Done :)

StreamChatUITestsApp/StreamChat/ChannelListVC.swift
bielikb: Thanks for the insight. Done :)

StreamChatUITestsAppUITests/Pages/MessageListPage.swift
```diff
@@ -0,0 +1,147 @@
...
+        
+        private static func action(label: String) -> XCUIElement {
+            let predicate = NSPredicate(format: "label LIKE '\(label)'")
+            return app.otherElements[id].staticTexts.matching(predicate).firstMatch
+        }
+    }
+    
+    struct Attributes {
+        static func reactionButton(messageCell: XCUIElement) -> XCUIElement {
+            messageCell.buttons["ChatMessageReactionItemView"]
```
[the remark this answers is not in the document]
bielikb: Good call. Noted down. We will revisit this comment once the test suite becomes more mature.

StreamChatUITestsAppUITests/Pages/MessageListPage.swift
```diff
@@ -0,0 +1,147 @@
...
+    struct Composer {
+        static var sendButton: XCUIElement { app.buttons["SendButton"] }
+        static var confirmButton: XCUIElement { app.buttons["ConfirmButton"] }
+        static var attachmentButton: XCUIElement { app.buttons["AttachmentButton"] }
+        static var commandButton: XCUIElement { app.buttons["CommandButton"] }
+        static var inputField: XCUIElement { app.otherElements["InputChatMessageView"] }
+    }
+    
+    
+    struct Reactions {
```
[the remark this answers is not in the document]
bielikb: Yeah sure. good call. Incoming :)

Sources/StreamChatUI/CommonViews/BaseViews.swift
bielikb: @evsaev Thanks for comment. There will always be at least 2-3 ways how & where to assign the accessibility identifier irregardless what helpers we provide.

The "chain of command" comes like this.
1. If the `View` class was declared in our SDK, it will share the identifier == to its class name
2. If the `View` is UIKit based and lazily instantiated -> we leverage the `with` helper
3. If the general `View` (irregardless of where it was defined) is declared we will leverage the property wrapper.

The rule of thumb is laziness of the engineer here. The accessibility identifiers should live as close to the property declaration as possible, so at glance you'd be able to say what's the accessibility identifier of given element. This comes with the time benefit when eg writing the UI tests for your new feature.

martinmitrevski (approved):
LGTM ✅  Before merging, please make sure the emerge checks don't show big increase to the app size, since I've read somewhere you've already addressed the issue.

StreamChatUITestsAppUITests/Robots/BasicAssertions.swift
alteral: @martinmitrevski, covered the Robots with the docs, thanks! 🙂

Sources/StreamChatUI/CommonViews/BaseViews.swift
bielikb: @evsaev  I parked the `@propertyWrapper` for now.
The plan here is to leverage the testing suite also by our integrators. Giving us & them simple semantic to set accessibility identifier is valuable and help to spread the consistency and to keep the identifiers as close to the view declaration as possible.

bielikb: Meaning. We will provide that `propertyWrapper` at later stage. For now, we need to progress with this work. So guess reverting this functionality unblocks this PR.

evsaev (approved):
🚢

Issue #1756, opened 2022-01-14
Cannot perform actions on long messages.


virbedi:
## What did you do?
Sent a long message in chat.
Tried to long press on message to get action list.

## What did you expect to happen?
Action list for message is visible

## What happened instead?
Could not see the action list, long message took up entire screen.

## GetStream Environment
**GetStream Chat version:** = StreamChatUI 4.8
**GetStream Chat frameworks:** StreamChat, StreamChatUI
**iOS version:** iOS 15.2
**Swift version:** Swift 5
**Xcode version:** 13.2.1
**Device:** iPhone 7 Plus, iPhone 13 Pro Max, iPhone 13 (Simulator)

## Additional context

https://user-images.githubusercontent.com/43729852/149596872-0b01b93e-c9f2-4829-b083-d0c97e1bd08c.MP4

bielikb:
Hi @virbedi, 

I was able to reproduce this issue. Ill file a bug internally and we will cover this issue in one of the following releases.
Ill let you know once this is resolved.

Thanks for reporting!

Best,
Boris

virbedi:
Hello, is there any update on this? 
As of `4.11.0` I am still facing this issue.

nuno-vieira:
Hi @virbedi!

We are really sorry! This was a mistake on our side and probably a miss-click on the closing of the issue. This is not yet fixed you are correct 👍 Anyways, this is still being tracked internally so we would not forget about fixing it 🙂 

Best,
Nuno

virbedi:
Thanks for the quick response @nuno-vieira. I'll keep an eye out for the fix

b-onc:
Internally tracked as CIS-1219

nuno-vieira:
Hi @virbedi!

Thank you for your patience on this one! This fix will be part of the next release 👍

We will close the issue once the next release is published.

Best,
Nuno

virbedi:
@nuno-vieira Thanks for the update, looking forward to the next release 🚀

PR #1871, opened 2022-03-23
[CIS-1219] - Fixing message actions on the Chat Message

adamrushy:
### 🔗 Issue Link
CIS-1219
https://github.com/GetStream/stream-chat-swift/issues/1756

### 🎯 Goal

We're not currently utilising the height of the StackView so the MessageContentView is pushing the ActionsController further down on larger messages which creates the bug seen in the ticket.

### 🛠 Implementation

The implementation is to remove the fixed height on the StackView, and allow the message to be truncated on larger messages.

### 🎨 UI Changes

The main change in the UI is that on larger messages you now see ellipses at the tail of the message.

### 🧪 Testing

1. Launch the Demo App
2. Create a large message that covers the whole screen
3. Tap and Hold on to that message to show the actions
4. The message is truncated and you can see all the reactions & actions.

### ☑️ Contributor Checklist

### 🎉 GIF

RocketSim_Recording_iPhone_13_2022-03-23_17 47 27

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Assertions are no longer thrown by default. Check `StreamRuntimeCheck` to enable them [#1885](https://github.com/GetStream/stream-chat-swift/pull/1885)
 - Local Storage is enabled by default. You can read more [here](https://getstream.io/chat/docs/sdk/ios/guides/offline-support) [#1890](https://github.com/GetStream/stream-chat-swift/pull/1890)
 
+## StreamChatUI
+### ⚠️ Changed
+### ✅ Added
+### 🐞 Fixed
+- MessageActions shrinks on larger messages [#1871](https://github.com/GetStream/stream-chat-swift/pull/1871)
+
 
 # [4.13.0](https://github.com/GetStream/stream-chat-swift/releases/tag/4.13.0)
 _March 29, 2022_


Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
```diff
@@ -599,6 +599,7 @@ open class ChatMessageContentView: _View, ThemeProvider {
             textView?.dataDetectorTypes = .link
             textView?.isScrollEnabled = false
             textView?.backgroundColor = .clear
+            textView?.textContainer.lineBreakMode = .byTruncatingTail
```
nuno-vieira, line 602: This will affect the message list as well. If we want to add the truncating in the message list too, we need a button to expand the message in the message list. 

At least for now, we can add this only in the Popup Actions view. Recently I've added a new function `ChatMessageActionsTransitionController.makeMessageContentView()` where we can change the message content view only for the popup actions view. The problem here is the `textView` is not created yet at this point :/ It will only be created once `createTextView()` is called.

So I can see two ways we can fix this:
1. Intoduce a new property `ChatMessageContentView.isPresentedInPopupActions: Bool` and set it to true in `ChatMessageActionsTransitionController.makeMessageContentView()` and then in `createTextView()` we only set the `.byTruncatingTail` if it is presented in popup actions.
2. Introduce closures or a factory of the different components in `ChatMessageContentView` so that we can easily replace the logic of `createTextView()` and the other views.

I actually prefer option 2, since it is more scalable and flexible, where 1. feels like a quick and dirty soluton.

Sources/StreamChatUI/MessageActionsPopup/MessageActionsTransitionController.swift
```diff
@@ -148,12 +148,10 @@ open class ChatMessageActionsTransitionController: NSObject, UIViewControllerTra
             usingSpringWithDamping: 0.7,
             initialSpringVelocity: 4,
             options: [.curveEaseInOut],
-            animations: {
+            animations: { [weak self] in
```
nuno-vieira, line 151: there's no memory leaks in animation closure, it is safe to remove this 👍

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
adamrushy: @nuno-vieira I don't see this behaviour 🤔

nuno-vieira: @adamrushy Nevermind, I see, this is only truncated when the height is limited 👍

Sources/StreamChatUI/MessageActionsPopup/MessageActionsTransitionController.swift
adamrushy: Hmm, do we know this for sure? I always added this but I couldn't be 100% sure 🤔

Issue #1854, opened 2022-03-16
ChannelList Recovery doesn't work properly because It couldn't process query with extra data field in ChannelDTO for local storage(CoreData)


tokirin:
## What did you do?
We were waiting for merging CIS-1275, and as seen it merged, tested it.
and we found bug for our usecase.

```Swift
    let queryA = ChannelListQuery(
      filter: .and([.containMembers(userIds: [userId]), .notEqual(.init(rawValue: "owner_id"), to: userId)])
    )
   let controllerA = chatClient.channeListController(query: queryA)

    let queryB = ChannelListQuery(
      filter: .and([.containMembers(userIds: [userId]), .equal(.init(rawValue: "owner_id"), to: userId)])
    )
    let controllerB = chatClient.channeListController(query: queryB)
```

We use 2 `ChatChannelListController`s with different `ChannelListQuery`.
The two `ChannelListQuery`'s are mutually exclusive as you can see.
and the queries have extra data field named `owner_id` for identifying whether the channel is created by current user or not

## What did you expect to happen?
It should synchronize correctly for both of `ChatChannelListController` when connection is recovered

## What happened instead?
When background recovery begins, two of `ChatChanneListController` begin to remove each other's `ChatChannel`s and  messages

## The Problem I found in ChannelListUpdater.swift
I started to analyze problem from `SyncRepository.swift` and found something weird
```Swift
//ChannelListUpdater.swift
guard let queryDTO = session.channelListQuery(filterHash: updatedQuery.filter.filterHash) else { return }

let localQueryCIDs = Set(queryDTO.channels.compactMap { try? ChannelId(cid: $0.cid) })
let remoteQueryCIDs = Set(channelListPayload.channels.map(\.channel.cid))
let queryAlreadySynched = remoteQueryCIDs.intersection(synchedChannelIds)

// We are going to unlink & clear those channels that are not present in the remote query,
// and that have not been synched nor watched. Those are outdated.
let cidsToRemove = localQueryCIDs
    .subtracting(queryAlreadySynched)
    .subtracting(watchedChannelIds)
```
The problem is that **the `localQueryCIDs` are same for both of above two queries.**
so I checked `ChannelDTO`'s CoreData Model(in `StreamChatModel.xcdatamodel`) and confirmed the `extraData` field is just binary field.

**So my guess is:**
the sample filter's string literal (by printing `filterHash` method) are
```Swift
//Participating Chatroom List Filter Hash 
[...]

PR #1879, opened 2022-03-25
[CIS-1631] Support multiple active channel lists at the same time

polqf:
### 🔗 Issue Link
https://stream-io.atlassian.net/browse/CIS-1631
#1854

### 🎯 Goal

The objective of this PR is to allow for multiple Channel Lists to be active at the same time.
Up until now, it was pretty much broken, and channels would go to both of the lists.

When it comes to offline support, the recovery was also faulty, and has been fixed to only delete those channels that are not used by any of the active channel list controllers

### 🛠 Implementation
Context:
At the moment, we cannot use the query's filter object at runtime. Its structure is too complex to resolve, and it would lead to a lot of overcomplexity to have a generic solution for all the possible use cases.
Instead, we are adding the possibility for our clients to control which channels should make it to which lists. This is expected to be much cheaper than a "catch them all" solution. 

- ChatChannelListController now accepts a `filterBlock` in its initializer, to allow for what's exposed above.
- ChatChannelListControllerDelegate has 2 deprecated methods:
```
    @available(*, deprecated, message: "Please use `filterBlock` when initializing a `ChatChannelListController`")
    func controller(_ controller: ChatChannelListController, shouldAddNewChannelToList channel: ChatChannel) -> Bool
    @available(*, deprecated, message: "Please use `filterBlock` when initializing a `ChatChannelListController`")
    func controller(_ controller: ChatChannelListController, shouldListUpdatedChannel channel: ChatChannel) -> Bool 
```
The usage of the above mentioned was misleading to the customers, and we were delegating the responsibility outside (in our default implementation to our ViewController 🤦).

IMO it is much cleaner and scalable to do it like that, but open to thoughts

> 2. What happens to Filter+ChatChannel?

For now, we are mimicking Android's behaviour as a default implementation

> 3. Why aren't you adding deprecation annotations to ChatChannelListControllerDelegate's definition?

We are extensively using this protocol across our codebase, and it causes a lot of warnings. Fixing those warning would incur in breaking changes

> 4. I would like to know more

https://www.notion.so/stream-wiki/Chat-Channel-List-Query-69b99c1b68df4d308a9a4feada9e5d8f

### ☑️ Contributor Checklist


Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift
```diff
@@ -164,18 +185,32 @@ public class ChatChannelListController: DataController, DelegateCallable, DataSt
...
-        
+
         let channels = changes.compactMap { change -> ChatChannel? in
             switch change {
             case let .insert(channel, _):
+                guard channelBelongsToController(channel) else { return nil }
                 return (delegate?.controller(self, shouldAddNewChannelToList: channel) ?? true) ? channel : nil
             case let .update(channel, _):
+                guard channelBelongsToController(channel) else { return nil }
                 return (delegate?.controller(self, shouldListUpdatedChannel: channel) ?? true) ? channel : nil
```
nuno-vieira, line 214: Does this break the current behaviour somehow? We need to make sure the previous behaviour is still the default, otherwise, we might break some customers 🤔 

This query problem is getting a bit out of control. From what I understood, the problem is that we can't currently link the query to a controller right? 🤔 (I'm not sure even, I'm confused)

Possible solutions:
- Since we can't make the queries equatable because of their complexity, would making the queries reference types instead of value types, help here? 🤔 
- Another solution could be to have the query contain the object address of the ChannelController so that we know what channel controller does the query belong to.

evsaev (commented):
When I was doing research in Dec I, actually, came up with the same approach since `filter: (ChatChannel) -> Bool` & `compare: (ChatChannel, ChatChannel) -> Void` is all needed to maintain channel list consistency. And since we don’t know anything about the desired filter/sorting, both closures should be non-optional and do not have default values.

Such approach has the following **pros**:
1. compile-time guarantee an integrator doesn't forget to pass it
2. `filter` is encapsulated in a single closure and used for new & updated channels
3. `filter` setup is next to `query.filter` and makes it easy to see that both describe the same conditions
4. Android handles it in the same way

But since we’re avoiding breaking changes, we have to provide default value for `filter` which kinda breaks the idea. We’re bringing the 2nd way to do filtering which will be confusing for customers, which can still be misused (if filter block is not provided), and it requires some kind of `channel.meets(filter)` which is very tough to implement especially when it comes to extra data.

To me, the approach with non-optional closures seems the right way to move forward 👍 And making a channel list fully functional is a good reason to bump major (but that would also involve `compare` part).

polqf:
I completely agree with your points above @evsaev. This is not ideal the ideal move, and it would be nice to be able to request both filter and compare functions as mandatory.

On the other hand, the issue that this PR describes is only present for clients using 2 Chat Channel Lists active at the same time. For the rest, this should not be an issue. 
Also, we are mimicking the behaviour of Android with this change, which is also a "nice" in the sense of consistency.

All in all, I see this as a move towards an approach that we want to take, whilst allowing those who need it to use it in the meantime. We'll have to discuss about the next steps here, that's for sure.

Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift
polqf: It does not break the current behaviour, not. That is because the default implementation of `channelBelongsToController` just returns true, unless a filterBlock is passed.

The problem around query is not that we cannot link a query to a controller, but the fact that we are using query's linked channels as a source of data for our channel list. So the solutions posted below do not solve the issue.

I will shoot a Notion doc to make sure the issues are understood 🙏

nuno-vieira: Thanks for the explanation 🙏

nuno-vieira:
> I completely agree with your points above @evsaev. This is not ideal the ideal move, and it would be nice to be able to request both filter and compare functions as mandatory.
> 
> On the other hand, the issue that this PR describes is only present for clients using 2 Chat Channel Lists active at the same time. For the rest, this should not be an issue. Also, we are mimicking the behaviour of Android with this change, which is also a "nice" in the sense of consistency.
> 
> All in all, I see this as a move towards an approach that we want to take, whilst allowing those who need it to use it in the meantime. We'll have to discuss about the next steps here, that's for sure.

This is probably useful to have in our documentation 🤔 We should have this documented in our docs, otherwise we will always have new support tickets if customers want to have 2 active channel lists at the same time. A new section talking about 2 active channel lists could be added here: https://getstream.io/chat/docs/sdk/ios/uikit/controllers/channels/

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### ✅ Added
 - Introduce message localization. See [docs](https://getstream.io/chat/docs/ios-swift/translation/?language=swift) for more info [#1867](https://github.com/GetStream/stream-chat-swift/issues/1867)
 - Add support for multiple push providers [#1864](https://github.com/GetStream/stream-chat-swift/issues/1864)
+- Add support for  multiple active channel lists at the same time [#1879](https://github.com/GetStream/stream-chat-swift/pull/1879)
```
nuno-vieira, line 10: ⚠️  Don't forget to properly edit the changelog, after 4.13.0 release

polqf: Good one!

Sources/StreamChat/Workers/ChannelListUpdater.swift
```diff
@@ -51,23 +52,34 @@ class ChannelListUpdater: Worker {
...
-                            
-                            channelDTO.resetEphemeralValues()
-                            channelDTO.messages.removeAll()
                             queryDTO.channels.remove(channelDTO)
                         }
+
+                        // We are going to clean those channels that are present in the both the local and remote query,
+                        // and that have not been synched nor watched. Those are outdated, can contain gaps.
+                        let cidsToClean = localInRemote.subtracting(updatedChannels)
+                        session.cleanChannels(cids: cidsToClean)
```
evsaev, line 69: I would only make `RefetchChannelListQueryOperation` operation responsible for unlinking channels from a query and move channels clean-up fully to `CleanUnwantedChannelsOperation` as it makes easier to understand what's happening.

If there's a channel `C1` locally and it is linked to `Q1` and `Q2` local queries, then when the 1st page for `Q1` is fetched `C1` can be outside this page but can still be part of `Q2's` first page. In that case it's easier to make a decision which channels are/can get outdated when all queries are refetched.

evsaev, line 69: I would only make `RefetchChannelListQueryOperation` operation responsible for unlinking channels from a query and move channels clean-up fully to `CleanUnwantedChannelsOperation` as it makes easier to understand what's happening.

If there's a channel `C1` locally and it is linked to `Q1` and `Q2` local queries, then when the 1st page for `Q1` is fetched `C1` can be outside this page but can still be part of `Q2's` first page. 

I see that we're subtracting `updatedChannels` which is a union of synced & watched but it's easier to make a decision which channels are/can get outdated when all queries are refetched.

evsaev: What if we just clean all channels that are not linked to any query after connection recovery subtracting watched channels? 🤔 because the math is too complex, I'm already lost with it 😄

Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift
```diff
@@ -7,14 +7,27 @@ import Foundation
...
+    /// When passing `filterBlock`, make sure the runtime logic matches the one expected by the filter passed in the query object.
+    /// If they don't match, there can be jumps when loading the list.
+    ///
+    /// - Parameters:
+    ///   - query: The query specify the filter and sorting of the channels the controller should fetch.
+    ///   - filterBlock: A block that determines whether the channels belongs to this controller.
+    /// - Returns: A new instance of `ChatChannelListController`
+    public func channelListController(
+        query: ChannelListQuery,
+        filterBlock: ((ChatChannel) -> Bool)? = nil
```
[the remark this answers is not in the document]
polqf: Fine with that

Sources/StreamChat/Workers/ChannelListUpdater.swift
polqf: There are 2 things here:

> I would only make `RefetchChannelListQueryOperation` operation responsible for unlinking channels
It is something needed to leave that particular channel list ready for it to be used. I don't think the responsibility should be offloaded to the operation. Actually, the operation should only know about the result of it, IMO.

> What if we just clean all channels that are not linked to any query after connection recovery subtracting watched channels? 🤔 because the math is too complex, I'm already lost with it 😄
Happy to receive suggestions here. What I can tell you now is that I tried many things, and this was the least complex while still readable that I managed to do 😂

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
```diff
@@ -252,6 +252,14 @@ extension NSManagedObjectContext {
         
         delete(dto)
     }
+
+    func cleanChannels(cids: Set<ChannelId>) {
+        for cid in cids {
+            guard let channelDTO = channel(cid: cid) else { continue }
+            channelDTO.resetEphemeralValues()
+            channelDTO.messages.removeAll()
```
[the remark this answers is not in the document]
evsaev: I'd also do that for:
- pinnedMessages
- channelReads
- members

Sources/StreamChat/Workers/ChannelListUpdater.swift
evsaev: > It is something needed to leave that particular channel list ready for it to be used

Yes, but we can't make a decision to clean up a local channel matching a query when it's outside the 1st page from remote as it can be on the 1st page of another query and would get synced in the end if we didn't clean it up.

evsaev: > Happy to receive suggestions here. What I can tell you now is that I tried many things, and this was the least complex while still readable that I managed to do 😂

If we make only `CleanUnwantedChannelsOperation` responsible for cleaning channels, the operation key par could look like that:
```swift
let channelsToClean = context
  .localChannelIds
  .compactMap { session.channel(cid: $0) }
   // clean a channel that is not linked to any query excluding watched channels
  .filter { $0.queries.isEmpty && !context.watchedChannelIds.contains($0.cid) }
```swift

The code above could be done with 1 fetch request with a compound predicate to avoid multiple fetches to the persistent store for each single channel

File: Sources/StreamChat/Database/DTOs/ChannelDTO.swift (hunk 5 above)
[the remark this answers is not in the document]
Comment (polqf) [reply Sources/StreamChat/Database/DTOs/ChannelDTO.swift:260]: Would you also remove attachments?

File: Sources/StreamChat/Workers/ChannelListUpdater.swift (hunk 3 above)
Comment (polqf) [reply Sources/StreamChat/Workers/ChannelListUpdater.swift:69]: > Yes, but we can't make a decision to clean up a local channel matching a query when it's outside the 1st page from remote as it might be on the 1st page of another query and might get synced in the end if we didn't clean it up. CleanUnwantedChannelsOperation seems to be designed for that very purpose

We actually can because at that step we are only cleaning those which are present in the remote query, which means they will get populated with new data.

> If we make only CleanUnwantedChannelsOperation responsible for cleaning channels, the operation key part could look like that:

IMO it is better to have the current logic than involve fetching all the channels as you suggested above

Comment (evsaev) [reply Sources/StreamChat/Workers/ChannelListUpdater.swift:69]: Discussed on the call ✅ Pol was saying about local channels that are on remote but not synced - we can clean them as part of query re-fetch operation. I was saying about channels that are outside the 1st page that will go out of sync if not watched. Closing this as we're on the same page now 🙂

File: Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift [hunk 6]
```diff
@@ -187,7 +222,8 @@ public class ChatChannelListController: DataController, DelegateCallable, DataSt
         let channels = changes.compactMap { change -> ChatChannel? in
             switch change {
             case let .update(channel, _):
-                // We unlink the channels where `shouldListUpdatedChannel` returns `false`
+                // We unlink the channels that do not belong to this controller
+                guard channelBelongsToController(channel) else { return channel }
```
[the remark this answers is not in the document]
evsaev: Hm, my understanding is that if `filter` is present, the delegate should not be called 🤔 If I were a customer and move to the new `filter` API (to avoid deprecation warnings), I would provide the `filter` for custom query and remove overrides for `shouldAdd/shouldList`. But removing overrides from custom channel list VC would lead to default implementation in `ChatChannelVC`  being used where we check `channel.membership`. Or I'm missing smth?

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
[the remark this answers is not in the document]
evsaev: We can remove that relation, it's not used anywhere 🙂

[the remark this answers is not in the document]
polqf: It's gone 😄

Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift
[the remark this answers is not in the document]
polqf: That's actually a good point

docusaurus/docs/iOS/uikit/controllers/channels.md
```diff
@@ -17,6 +17,23 @@ func controller(
...
+    // Your filtering logic
+    return true
+})
+```
+
+It is expected that the logic on the `filter` block matches the one in `query.filter`.
+
+:::note
+Given the nature of `Filter`, having generic logic capable of using `query.filter` at runtime to determine if a ChatChannel meets the criteria is extremely expensive and complex. Instead, having a block that only checks for the particular needs of the use case is much cheaper.
+:::
```
evsaev, line 35: We don't perform any computations so far so what's written here is actually not true. I also have a strong feeling that non-optional filter would be the way to go in v5 which is on the roadmap. When the filter is required, we won't need expensive `channel.meets(filter)` block. Do we really need this note in place?

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## StreamChat
 ### ✅ Added
 - `quotesEnabled` property is added to the `ChannelConfig` [#1891](https://github.com/GetStream/stream-chat-swift/issues/1891)
+- Add support for  multiple active channel lists at the same time [#1879](https://github.com/GetStream/stream-chat-swift/pull/1879)
```
evsaev, line 8: To me, it feels more like a fix because we didn't explicitly forbid customers to have multiple channel lists. But leaving it up to you 👍

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
```diff
@@ -252,6 +251,17 @@ extension NSManagedObjectContext {
         
         delete(dto)
     }
+
+    func cleanChannels(cids: Set<ChannelId>) {
+        for cid in cids {
+            guard let channelDTO = channel(cid: cid) else { continue }
```
evsaev, line 257: Would it make sense to have a simple fetch request to load all channels in batch other than fetching each one separately?

docusaurus/docs/iOS/uikit/controllers/channels.md
polqf: > having generic logic capable of using `query.filter` at runtime 

The note states that we don't have it, and the reason we don't have it is because it would be expensive. I would leave it as an explanation

CHANGELOG.md
polqf: Well, I get your point. But effectively we never had support for it, so we are adding it 🤔

b-onc: Nope, we did. Using the delegates, you can have multiple active channel lists, it's also used in DemoApp

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
polqf: Done!

CHANGELOG.md
polqf: Understood! Changed it

Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift
```diff
@@ -164,19 +185,47 @@ public class ChatChannelListController: DataController, DelegateCallable, DataSt
...
+        let deprecatedFallback: () -> Bool? = { [unowned self] in
+            switch change {
+            case .insert:
+                return self.delegate?.controller(self, shouldAddNewChannelToList: channel)
+            case .update:
+                return self.delegate?.controller(self, shouldListUpdatedChannel: channel)
+            default:
+                return nil
+            }
+        }
```
evsaev, line 207: nit: no need to weakify, the lifetime of that closure won't exceed method's execution time
```suggestion
        let deprecatedFallback: () -> Bool? = {
            switch change {
            case .insert:
                return self.delegate?.controller(self, shouldAddNewChannelToList: channel)
            case .update:
                return self.delegate?.controller(self, shouldListUpdatedChannel: channel)
            default:
                return nil
            }
        }
```

Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift
```diff
@@ -164,19 +185,47 @@ public class ChatChannelListController: DataController, DelegateCallable, DataSt
...
+
+        do {
+            return try deprecatedFallback() ?? channel.meets(query.filter)
+        } catch {
+            log.error("Unable to resolve complex filter. Please pass a `filter` block when intializing ChatChannelListController")
+        }
+        return true
```
evsaev, line 214: nit: `meets` is empty and most likely we won't implement it. If we make it non-throwing the code would get a bit simpler:
```suggestion
        return deprecatedFallback() ?? channel.meets(query.filter)
```

evsaev (approved):
🚢

Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift
polqf: The idea is for us to provide a simple non expensive implementation for it in the future, and throw in the case of it getting to complex to resolve in a reasonable time

Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift
polqf: You're right, this is a leftover of something else I tried

PR #1885, opened 2022-03-29
[CIS-1424] Hide Assertions under StreamRuntimeCheck

polqf:
### 🔗 Issue Link
https://stream-io.atlassian.net/browse/CIS-1424

### 🎯 Goal

The purpose of this PR is to not trigger assertions when people is testing our SDK. Assertions are sometimes misunderstood as crashes, and that's not a good first impression.

### 🛠 Implementation

A new namespaced `assertionsEnabled` flag is available under a `StreamRuntimeCheck` enum. The value of this property is used to determine if the assertion should be thrown or not.

It also adds a new way of having a slightly different configuration internally than the one we want to have publicly.
Read more about this here: https://www.notion.so/stream-wiki/Dev-Setup-84b1cbc6ddad4bfaa80afe18552543a9#c5a1bafc323e4448828c5691bef162d8

### ☑️ Contributor Checklist


DemoApp/DemoAppCoordinator.swift
```diff
@@ -103,9 +107,11 @@ final class DemoAppCoordinator: NSObject, UNUserNotificationCenterDelegate {
         Components.default.messageContentView = CustomMessageContentView.self
         Components.default.messageListDateSeparatorEnabled = true
         Components.default.messageListDateOverlayEnabled = true
-        Components.default._messageListDiffingEnabled = true
+        Components.default._messageListDiffingEnabled = isStreamInternalConfiguration
```
nuno-vieira, line 110: We will soon move this to the default as well, but either way, I don't think there is any problem if customers use this in our demo app 🤔

bielikb (approved):

PR #1894, opened 2022-04-04
[Feature Branch] Enable SwiftLint

adamrushy:
### 🔗 Issue Link
https://stream-io.atlassian.net/browse/CIS-1629

### 🎯 Goal

The goal of this PR is to enable SwiftLint to our project _again_. This is WIP and is the main feature branch as we start fixing/agreeing on the rules to bring forward.

### 🛠 Implementation

The main implementation here is to add the SwiftLint script and the rules we had previously.

### 🎨 UI Changes

N/A

### 🧪 Testing

You'll need to install SwiftLint first but then CMD + B ;]

### ☑️ Contributor Checklist


StreamChat.xcodeproj/project.pbxproj
```diff
@@ -8309,6 +8310,24 @@
...
+			);
+			name = SwiftLint;
+			outputFileListPaths = (
+			);
+			outputPaths = (
+			);
+			runOnlyForDeploymentPostprocessing = 0;
+			shellPath = /bin/sh;
+			shellScript = "export PATH=\"$PATH:/opt/homebrew/bin\"\nif which swiftlint > /dev/null; then\n  swiftlint\nelse\n  echo \"warning: SwiftLint not installed, download from https://github.com/realm/SwiftLint\"\nfi\n";
+		};
```
evsaev, line 8330: Having the phase in DemoApp means that when compiling just LLC or UI rule violations won't be highlighted. It's fine for now since we don't have another place where linting for both SDKs can happen but let's keep an eye on it and maybe have a tech discussion point. Linting SDKs separately is what we might finally end up with.

evsaev (commented):
Well done @adamrushy 💪 2 things caught my eye: excluded files and enabling/disabling colon rule around type declarations. Will LGTM once discussed/resolved

StreamChat.xcodeproj/project.pbxproj
adamrushy: Yeah, I agree on this @evsaev and it's a very good point

.swiftlint.yml
```diff
@@ -0,0 +1,88 @@
...
+  - Tests
+  - Sources/StreamChatUI/Deprecations.swift
+  - Sources/StreamChatUI/**/**/AttachmentViewCatalog.swift
+  - Sources/StreamChat/Controllers/DataController.swift
+
+disabled_rules:
+  - large_tuple
+  - multiple_closures_with_trailing_closure
+  - todo
+  - function_parameter_count
```
nuno-vieira, line 21: We actually violate a lot this rule in our project, and I think we shouldn't. We should try and extract parameters to structs, to avoid this, and sometimes it is actually a code smell if you have a function with many arguments. Ideally, we would disallow this rule in the places we can't break it right now, but for future functions, we would be mindful when doing this. But maybe this is spread across the codebase right now?

.swiftlint.yml
```diff
@@ -0,0 +1,88 @@
...
+  - force_cast
+  - force_try
```
nuno-vieira, line 23: Why are we disabling these ones? 🤔

Sources/StreamChatUI/ChatChannel/ChatChannelHeaderView.swift
```diff
@@ -5,11 +5,15 @@
...
 import UIKit
 
+// swiftlint:disable colon
+
 /// The view that displays channel information on the message list header
 open class ChatChannelHeaderView:
     _View,
     ThemeProvider,
     ChatChannelControllerDelegate {
+    // swiftlint:enable colon
```
nuno-vieira, line 15: I think this rule is actually better if we fix it like this: Otherwise, we will have this disabled rule all over the place. WDYT?
```suggestion
/// The view that displays channel information on the message list header
open class ChatChannelHeaderView: _View,
    ThemeProvider,
    ChatChannelControllerDelegate {
```

.swiftlint.yml
adamrushy: Yeah, so the ones which are widespread I was thinking to ignore and create a new ticket to cover those rules specifically?

Sources/StreamChatUI/ChatChannel/ChatChannelHeaderView.swift
adamrushy: Yep, great suggestion - I have pushed this change-up :]

.swiftlint.yml
adamrushy: As mentioned above, this is also widespread - my thinking is to get this covered in a new ticket, just otherwise this PR will become huge WDYT?

StreamChatUITestsAppUITests/MockServer/MessageDetails.swift
```diff
@@ -0,0 +1,47 @@
+//
+// Copyright © 2022 Stream.io Inc. All rights reserved.
+//
+
+@testable import StreamChat
+import XCTest
+
+extension StreamMockServer {
+    
+    func saveMessageDetails(
```
alteral, line 10: Not sure what's going on here, but I think this file (`MessageDetails.swift`) is out of date. I can't even find it in the source code. `MessageList.swift` is the right one.

StreamChatUITestsAppUITests/MockServer/MessagingResponses.swift
```diff
@@ -0,0 +1,95 @@
+//
+// Copyright © 2022 Stream.io Inc. All rights reserved.
+//
+
+@testable import StreamChat
+import Swifter
+
+extension StreamMockServer {
+    
+    func configureMessagingEndpoints() {
```
alteral, line 10: Same here with `MessagingResponses.swift`, this file is out of date. No trace in the source code. `MessageResponses.swift` is the right one.

alteral (changes requested):
Nicely done, @adamrushy 🚀
I have just a couple of comments about MockServer.

.swiftlint.yml
nuno-vieira: Makes sense 🙂

.swiftlint.yml
nuno-vieira: For now, would be nice to just ignore them, and if we think it is necessary we can create tickets in the future for this case 👍

StreamChatUITestsAppUITests/MockServer/MessagingResponses.swift
adamrushy: Yeah, I can't see this either in my Xcode, it's the correct version

StreamChatUITestsAppUITests/Pages/MessageListPage.swift
```diff
@@ -5,7 +5,7 @@
 import Foundation
 import XCTest
 
-class MessageListPage {
+enum MessageListPage {
```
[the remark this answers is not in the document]
adamrushy: Okay, I have ignored this rule and pushed up :]

nuno-vieira (approved):

StreamChatUITestsAppUITests/MockServer/MessagingResponses.swift
alteral: Seems like, for some reason, your PR brings these files back, but without reference in Xcode. Could you please manually remove `StreamChatUITestsAppUITests/MockServer/MessagingResponses.swift` and `StreamChatUITestsAppUITests/MockServer/MessageDetails.swift` in Finder?

bielikb (approved):
Ship it!

Issue #1830, opened 2022-03-07
AdvancedOptionsViewController in Demo App project isn't functional


glennposadas:
The controller just doesn't work. I'm trying to learn how to use stream chat by using my own keys and tokens so I thought this advanced option is a good thing but isn't completed yet.

<img width="869" alt="Screen Shot 2022-03-08 at 1 52 33 AM" src="https://user-images.githubusercontent.com/12502679/157090001-412eaa85-3847-43da-a264-b85bce493054.png">

Also, I have a side question. The Stream can be possibly integrated into an Objective-C project, yes?

Thanks!

nuno-vieira:
Hi @glennposadas!

Thank you for opening the issue!
We will investigate this as soon as we can and will get back to you 👍

> Also, I have a side question. The Stream can be possibly integrated into an Objective-C project, yes?
Yes in theory it should. But we don't have many customers with Obj-C projects, so maybe you can find minor problems. But if you do, please open an issue here and we will get that fixed for sure 🙂 

Best,
Nuno

glennposadas:
Thanks, Nuno!

PR #1906, opened 2022-04-07
[CIS-1555] Fix crash when changing users within an active chat

b-onc:
### 🔗 Issue Link
CIS-1555 (see the zendesk ticket related to this)

### 🎯 Goal

One of the usecases we support is: app opens with guest/anonymous user, which can send/read messages. At one point, user can login and become a signed in user, without reloading the chat screen

### 🛠 Implementation

Let's understand our DB cleanup mechanism first.
When a new user logs in, we call `databaseContainer.removeAllData(force: true)` non-force clean is not implemented, and it's not clear what it'll be. Perhaps it'll only remove user-related data (such as removing only `CurrentUserDTO` but keeping channels/messages intact). Anyways, it's not the point of this PR.

`removeAllData` rips out the underlying DB file. In this case, none of the DB observers will get `object deleted` notification, because we don't play by the CoreData rules. We "workaround" this by sending our own "willDelete" and "didDelete" notifications.

`willDelete` prepares `delete` events for all the objects we're currently aware of (in the context of observer)
`didDelete` notification is sent _after_ the DB file is removed. This causes the `delete` events to be published to the observing object.

This is where it gets fun. Whenever an update is published from observers (`EntityChangeObserver` or `ListChangeObserver`), we call `asModel` on the DTO, lazily. Since the DTO is deleted, all fields are reset: strings are "", numbers are 0. `Date`s are different: if we try to get the `Date` from an erased object, we get
```swift
Date._unconditionallyBridgeFromObjectiveC(_:)
```
our beloved crash.

The solution is to publish the changes _before_ the underlying DTOs are deleted, so calling `asModel` on them succeeds. The DTOs are removed immediately after the changes are published.

### 🧪 Testing

First, checkout the first commit of this repo. It adds the possibility of logging in as `anonymous` and `guest` users.
Due to our app configuration, anonymous users don't see any channels. So:
1. Login as guest user. ID doesn't matter, leave it empty, defaults to `guest`
1. Open the first channel you see
1. Tap the debug icon in the navigation bar.
1. Tap "switch user"
1. Select any member
You'll see the `Date` crash.

Now, checkout the final commit, and follow the same steps. The chat screen will remove all messages and load them again, without any crash.

### 🎉 GIF

Simulator Screen Recording - iPhone 13 Pro - 2022-04-07 at 17 40 36

### ☑️ Contributor Checklist

### DIFF
--- a/DemoApp/LoginViewController.swift
+++ b/DemoApp/LoginViewController.swift
@@ -41,7 +41,7 @@ class LoginViewController: UIViewController {
     @IBOutlet var tableView: UITableView!
     var didRequestChatPresentation: ((DemoUserType) -> Void)!
     
-    let builtInUsers = UserCredentials.builtInUsers
+    let users: [DemoUserType] = UserCredentials.builtInUsers.map { DemoUserType.credentials($0) } + [.guest("guest"), .anonymous]
     
     override func viewDidLoad() {
         super.viewDidLoad()
@@ -75,48 +75,42 @@ class LoginViewController: UIViewController {
 
 extension LoginViewController: UITableViewDelegate, UITableViewDataSource {
     func tableView(_ tableView: UITableView, numberOfRowsInSection section: Int) -> Int {
-        builtInUsers.count + 3 // +1 for the last static cell
+        users.count
     }
     
     func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
         let cell = tableView.dequeueReusableCell(withIdentifier: "cell", for: indexPath) as! UserCredentialsCell
         
-        if indexPath.row == builtInUsers.count {
-            // Anonymous user
-            cell.nameLabel.text = "Anonymous user"
-            cell.descriptionLabel.text = ""
-            cell.avatarView.image = UIImage(systemName: "person")
+        let user = users[indexPath.row]
+        
+        switch user {
+        case let .credentials(userCredentials):
+            Nuke.loadImage(with: userCredentials.avatarURL, into: cell.avatarView)
             cell.avatarView.backgroundColor = .clear
-        } else if indexPath.row == builtInUsers.count + 1 {
-            // Guest user
+            cell.nameLabel.text = userCredentials.name
+            cell.descriptionLabel.text = "Stream test user"
+        case .guest:
             cell.nameLabel.text = "Guest user"
             cell.descriptionLabel.text = "user id: guest"
             cell.avatarView.image = UIImage(systemName: "person.fill")
             cell.avatarView.backgroundColor = .clear
-        } else if indexPath.row == builtInUsers.count + 2 {
-            // Advanced options
-            cell.nameLabel.text = "Advanced Options"
-            cell.descriptionLabel.text = "Custom settings"
-            cell.avatarView.image = UIImage(named: "advanced_settings")
+        case .anonymous:
+            cell.nameLabel.text = "Anonymous user"
+            cell.descriptionLabel.text = ""
+            cell.avatarView.image = UIImage(systemName: "person")
             cell.avatarView.backgroundColor = .clear
-        } else {
-            // Normal cell
-            let user = builtInUsers[indexPath.row]
-            Nuke.loadImage(with: user.avatarURL, into: cell.avatarView)
-            cell.avatarView.backgroundColor = .clear
-            cell.nameLabel.text = user.name
-            cell.descriptionLabel.text = "Stream test user"
         }
                 
         return cell
     }
     
     func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
-        if indexPath.row == builtInUsers.count {
-            // Anonymous user
-            didRequestChatPresentation(.anonymous)
-        } else if indexPath.row == builtInUsers.count + 1 {
-            // Guest user
+        let user = users[indexPath.row]
+        
+        switch user {
+        case .credentials, .anonymous:
+            didRequestChatPresentation(user)
+        case .guest:
             presentAlert(title: "Input a user id", message: nil, textFieldPlaceholder: "guest") { [weak self] userId in
                 if let userId = userId, !userId.isEmpty {
                     self?.didRequestChatPresentation(.guest(userId))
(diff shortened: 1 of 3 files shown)


nuno-vieira (changes requested):
Nice findings 👍  But the app is not working as expected in some scenarios:

**Scenario 1:**
1. Open app with User 1
2. Switch to User 2
3. List only has 1 item

https://user-images.githubusercontent.com/12814114/162274523-8ba1643d-fe41-4e62-bf46-0fd956865d54.mp4

 **Scenario 2:**
 1. Open app with User 1
 2. Logout
 3. Open app with User 2
 4. Channel List is empty and chat client is disconnected

This is not always reproducible, but in the video below you can see it is reproducible the second time:

https://user-images.githubusercontent.com/12814114/162274770-37588154-4b17-4680-9cad-68a5998a5f97.mp4

Although the PR looks good overall, I'm requesting changes since it is failing QA 👍

DemoApp/LoginViewController.swift
```diff
@@ -75,19 +75,30 @@ class LoginViewController: UIViewController {
...
+            cell.avatarView.image = UIImage(systemName: "person.fill")
+            cell.avatarView.backgroundColor = .clear
+        } else if indexPath.row == builtInUsers.count + 2 {
+            // Advanced options
             cell.nameLabel.text = "Advanced Options"
             cell.descriptionLabel.text = "Custom settings"
             cell.avatarView.image = UIImage(named: "advanced_settings")
-            cell.avatarView.backgroundColor = .systemGray
-            
+            cell.avatarView.backgroundColor = .clear
```
nuno-vieira, line 101: We probably can leverage `DemoUserType` and have the data of this table view be:

`var users: [DemoUserType]`

And maybe introduce a new type `DemoUserType.custom(...)`

With this, all the data of the table is derived from this property, and we don't need hard coded counts etc...

DemoApp/LoginViewController.swift
```diff
@@ -101,12 +112,24 @@ extension LoginViewController: UITableViewDelegate, UITableViewDataSource {
...
+        } else if indexPath.row == builtInUsers.count + 1 {
+            // Guest user
+            presentAlert(title: "Input a user id", message: nil, textFieldPlaceholder: "guest") { [weak self] userId in
+                if let userId = userId, !userId.isEmpty {
+                    self?.didRequestChatPresentation(.guest(userId))
+                } else {
+                    self?.didRequestChatPresentation(.guest("guest"))
+                }
+            }
+        } else if indexPath.row == builtInUsers.count + 2 {
```
nuno-vieira, line 127: As mentioned before, the enums will also make this easier. Right now it is very fragile, if add new options, it will be required to change all these hardcoded values.

nuno-vieira:
It seems that this is fixing CIS-1743 and CIS-1741 as well 👌 That's great news!

Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
```diff
@@ -315,7 +315,10 @@ open class ChatChannelListVC: _ViewController,
             completion: { _ in
                 // Move changes from NSFetchController also can mean an update of the content.
                 // Since a `moveItem` in collections do not update the content of the cell, we need to reload those cells.
-                self.collectionView.reloadItems(at: Array(indices.move.map(\.toIndex)))
+                let moveIndexes = Array(indices.move.map(\.toIndex))
+                if !moveIndexes.isEmpty {
+                    self.collectionView.reloadItems(at: Array(indices.move.map(\.toIndex)))
+                }
```
nuno-vieira, line 321:
```suggestion
                let moveIndexes = Array(indices.move.map(\.toIndex))
                if !moveIndexes.isEmpty {
                    self.collectionView.reloadItems(at: moveIndexes)
                }
```

evsaev (commented):
Great find with models evaluated after the deletion 👏 

Just wanna point out that when the channel list is shown, switching a user is not only about reconnecting `ChatClient` but also about updating the channel list query (e.g. if the filter contains a condition involving current user) of existed LLC controller OR building the new LLC controller with the new query and feeding it into `ChatChannelListVC`.

I'm happy with LLC changes but adding `switch user` option to DemoApp would involve some extra efforts before we can make it part of the pubic-facing application.

b-onc:
@evsaev Good point. I've iterated over these ideas:
1. `switch user` re-creates the view hierarchy
In this case, our filters are correct, but the bug is not reproducible.
2. `switch user` just switches user
As you point out, filters are not correct. Our VCs are not designed for this use case. For example, changing `controller` in `ChannelListVC` doesn't have any affect in its own. You have to call `synchronize` and `delegate = vc` on it separately. I didn't implement `didSet` on the `controller` in this PR to keep the scope limited. 

Logging in with anonymous or guest user creates the channel list with `type equals messaging` filter, so changing user after logging in as guest still works to some degree. The usecase this PR is showing is being used in production. As you said, logging in with a user and then switching to another user, channel list is not going to be correct. We can't avoid this without making the scope bigger.

b-onc:
@nuno-vieira Scenario 1 is the "issue" @evsaev described above. See my comment above.
Scenario 2 is not related to this PR. Chewbacca is a deleted user, you get the same screen testing in develop. We don't currently handle deleted user logins correctly.

DemoApp/LoginViewController.swift
b-onc: Did you know that "Advanced Options" didn't work at all? Like, the button is not wired 😂 I didn't know!
I'll switch to using enums and remove "Advanced Options" for now

nuno-vieira: We already have that issue reported: https://stream-io.atlassian.net/browse/CIS-1633 https://github.com/GetStream/stream-chat-swift/issues/1830

nuno-vieira: Not sure if it is a good idea to remove that. Is that much work to wire that button? Or leave it non wired then 🤔

nuno-vieira:
> @evsaev Good point. I've iterated over these ideas:
> 
> 1. `switch user` re-creates the view hierarchy
>    In this case, our filters are correct, but the bug is not reproducible.
> 2. `switch user` just switches user
>    As you point out, filters are not correct. Our VCs are not designed for this use case. For example, changing `controller` in `ChannelListVC` doesn't have any affect in its own. You have to call `synchronize` and `delegate = vc` on it separately. I didn't implement `didSet` on the `controller` in this PR to keep the scope limited.
> 
> Logging in with anonymous or guest user creates the channel list with `type equals messaging` filter, so changing user after logging in as guest still works to some degree. The usecase this PR is showing is being used in production. As you said, logging in with a user and then switching to another user, channel list is not going to be correct. We can't avoid this without making the scope bigger.

Can't we just recreate the ChatChannetListVC in this case? Merging this with the current issue is not a good idea, since it will be easily reproducible by clients.

bielikb (approved):
QA'd. This looks neat.

Thanks for adding the new capabilities to the Demo App!

b-onc:
Closing this since it's incomplete. Will open a new PR for this ticket.

PR #1911, opened 2022-04-08
[CIS-1687] Channel Header would not update its name when it is updated

hugobernalstream:
Make ChatChannelHeaderView to update always whenever the channel cont……roller is updated.

### 🔗 Issue Link
https://stream-io.atlassian.net/browse/CIS-1687

### 🎯 Goal

Fix issue that wasn't updating the channel group name whenever it was changed from another ViewController on top of that Channel.

### 🛠 Implementation

Call `updateContent()` instead of `updateContentIfNeeded()` whenever the channel controller is updated. Validation for superview if the updating VC is not in the hierarchy always fails.

### 🎨 UI Changes

Simulator Screen Recording - iPhone 13 Pro - 2022-04-08 at 13 07 08

### 🧪 Testing

Jira ticket contains a .patch file to add a VC to push over the Chat channel and make the change of the channel name there.

### ☑️ Contributor Checklist


Tests/StreamChatUITests/SnapshotTests/CommonViews/ChatChannelHeaderView/ChatChannelHeaderView_Tests.swift
```diff
@@ -0,0 +1,65 @@
+//
+// Copyright © 2022 Stream.io Inc. All rights reserved.
+//
+
+import StreamChat
+@testable import StreamChatTestTools
+@testable import StreamChatUI
+import XCTest
+
+class ChatChannelHeaderView_Tests: XCTestCase {
```
evsaev, line 10:
```suggestion
final class ChatChannelHeaderView_Tests: XCTestCase {
```

Tests/StreamChatUITests/SnapshotTests/CommonViews/ChatChannelHeaderView/ChatChannelHeaderView_Tests.swift
```diff
@@ -0,0 +1,65 @@
...
+    }
+    
+    override func tearDown() {
+        view = nil
+        mockChannelController = nil
+        
+        super.tearDown()
+    }
+    
+    func test_SetNewChannelName_IsSuccessfullyChanged() {
```
evsaev, line 31: As I remember, recently there were some changes in terms of naming convention in tests. Let's check with @bielikb on that.

evsaev (commented):
In most cases, content is assigned **before** the view is part of the view heirachy. By doing  `updateContent` in didSet, we start accessing the subviews and they get instantiated (all subviews are `lazy`) with types taken from **Components.default**. Any custom components injected into a parent component would get ignored because the responder chain is not yet established.

```swift
public extension ComponentsProvider where Self: UIResponder {
    var components: Components {
        get {
            if let components = anyComponents as? Components {
                return components
            }
            
             // This lookup does not work because there's no `next` before the view is part of view heirarchy
            var _next = next 
            while _next != nil {
                if let _next = _next as? ComponentsProvider {
                    return _next.components
                } else {
                    _next = _next?.next
                }
            }
            
            // We're always using `Components.default`
            return .default
        }
        set { register(components: newValue) }
    }
}
```
However, the outcome of the last tech discussion was to always use `Components.default` and move away from responder chain injection so I think we can make this fix even though it's not aligned with the rest of the codebase.

WDYT @GetStream/ios-developers?

Sources/StreamChatUI/CommonViews/TitleContainerView/TitleContainerView.swift
```diff
@@ -10,7 +10,7 @@ open class TitleContainerView: _View, AppearanceProvider, SwiftUIRepresentable {
     /// Content of the view that contains title (first line) and subtitle (second nil)
     open var content: (title: String?, subtitle: String?) = (nil, nil) {
         didSet {
-            updateContentIfNeeded()
+            updateContent()
```
nuno-vieira, line 13: Do we need to change this one as well? I think only the update from the delegate is what we need to change 🤔

Tests/StreamChatUITests/SnapshotTests/CommonViews/ChatChannelHeaderView/ChatChannelHeaderView_Tests.swift
```diff
@@ -0,0 +1,65 @@
...
+    func test_ChannelNameSet() {
+        let newChannelName = "New Channel Name"
+        mockChannelController.simulateInitial(
+            channel: .mockDMChannel(name: newChannelName),
+            messages: [],
+            state: .localDataFetched
+        )
+        
+        AssertSnapshot(view)
+    }
```
nuno-vieira, line 55: It is possible to add mocked messages in this case? To test the subtitle of the header view? Otherwise, this test is not required, because it is already covered by the first one 👍

Tests/StreamChatUITests/SnapshotTests/CommonViews/ChatChannelHeaderView/ChatChannelHeaderView_Tests.swift
```diff
@@ -0,0 +1,65 @@
...
+}
+
+private extension ChatChannelHeaderView {
+    func addSizeConstraints() {
+        NSLayoutConstraint.activate([
+            widthAnchor.constraint(equalToConstant: 360),
+            heightAnchor.constraint(equalToConstant: 360)
```
nuno-vieira, line 62: Maybe we can decrease the height here to mimic the navbar?

Tests/StreamChatUITests/SnapshotTests/CommonViews/ChatChannelHeaderView/ChatChannelHeaderView_Tests.swift
hugobernalstream: Done

Tests/StreamChatUITests/SnapshotTests/CommonViews/ChatChannelHeaderView/ChatChannelHeaderView_Tests.swift
hugobernalstream: Checked with new templates and code snippets and updated accordingly

Sources/StreamChatUI/CommonViews/TitleContainerView/TitleContainerView.swift
hugobernalstream: Updated so we don't have to change it here.

Tests/StreamChatUITests/SnapshotTests/CommonViews/ChatChannelHeaderView/ChatChannelHeaderView_Tests.swift
hugobernalstream: Adjusted to also show and test the subtitle. You can see it on the new snapshots added to this PR.

Tests/StreamChatUITests/SnapshotTests/CommonViews/ChatChannelHeaderView/ChatChannelHeaderView_Tests.swift
hugobernalstream: Constrants not needed now. This part was removed and now it adjusts the frame depending on the content of the title and subTitle labels.

Tests/StreamChatUITests/SnapshotTests/CommonViews/ChatChannelHeaderView/ChatChannelHeaderView_Tests.swift
```diff
@@ -0,0 +1,59 @@
+//
+// Copyright © 2022 Stream.io Inc. All rights reserved.
+//
+
+import StreamChat
+@testable import StreamChatTestTools
+@testable import StreamChatUI
+import XCTest
+
+final class ChatChannelHeaderView_Tests: XCTestCase {
```
nuno-vieira, line 10: Would be nice to add unit tests to the missing subtitleText coverage: 
- WhenIsDirectMessageChannel_AndIsMemberOnline
- WhenIsDirectMessageChannel_AndIsLastActive
- WhenIsDirectMessageChannel_AndIsOffline
Screenshot 2022-04-12 at 00 43 27

Tests/StreamChatUITests/SnapshotTests/CommonViews/ChatChannelHeaderView/ChatChannelHeaderView_Tests.swift
evsaev: Perfect! 🙌 Would you also lowercase the first letters after `_` to have it consistent with the rest of the codebase?

hugobernalstream: Done

Tests/StreamChatUITests/SnapshotTests/CommonViews/ChatChannelHeaderView/ChatChannelHeaderView_Tests.swift
hugobernalstream: Tests added

PR #1913, opened 2022-04-08
[Docs] Add missing ChatMessageAttachment typealias in custom attachments guide

nuno-vieira:
### 🔗 Issue Link
None

### 🎯 Goal
Fix a missing step from the custom attachments guide.


adamrushy (approved):
Looks good :]

PR #1915, opened 2022-04-11
[CIS-1742] Save channel config as DTO instead of binary data

nuno-vieira:
### 🔗 Issue Link
CIS-1742

### 🎯 Goal
Change how we save the channel config to avoid crashes in schema changes.

### 🛠 Implementation
Instead of storing it as binary data in `ChannelDTO`, we create a new `ChannelConfigDTO`. Storing as binary data is dangerous, because if the structure changes, CoreData doesn't know the schema has changed, so it doesn't do migration or wipe the database.

### 🎨 UI Changes
N/A

### 🧪 Testing

1. Uninstall DemoApp
2. git checkout 73d32a60992b237383cb4f83579720b292123fff
3. Open Han Solo user
4. Close the app
5. git checkout develop
6. Open Han Solo User
7. Should not crash

### ☑️ Contributor Checklist


Sources/StreamChat/Database/StreamChatModel.xcdatamodeld/StreamChatModel.xcdatamodel/contents
```diff
@@ -99,6 +117,13 @@
         <relationship name="channel" optional="YES" maxCount="1" deletionRule="Nullify" destinationEntity="ChannelDTO" inverseName="reads" inverseEntity="ChannelDTO"/>
         <relationship name="user" optional="YES" maxCount="1" deletionRule="Nullify" destinationEntity="UserDTO" inverseName="channelReads" inverseEntity="UserDTO"/>
     </entity>
+    <entity name="CommandDTO" representedClassName="CommandDTO" syncable="YES">
+        <attribute name="args" attributeType="String"/>
```
evsaev, line 121: Is there any reason why `args` is the only field that does not have default value?

Sources/StreamChat/Database/StreamChatModel.xcdatamodeld/StreamChatModel.xcdatamodel/contents
```diff
@@ -14,9 +14,26 @@
...
+        <attribute name="readEventsEnabled" attributeType="Boolean" defaultValueString="NO" usesScalarValueType="YES"/>
+        <attribute name="repliesEnabled" attributeType="Boolean" defaultValueString="NO" usesScalarValueType="YES"/>
+        <attribute name="searchEnabled" attributeType="Boolean" defaultValueString="NO" usesScalarValueType="YES"/>
+        <attribute name="typingEventsEnabled" attributeType="Boolean" defaultValueString="NO" usesScalarValueType="YES"/>
+        <attribute name="updatedAt" attributeType="Date" usesScalarValueType="NO"/>
+        <attribute name="uploadsEnabled" attributeType="Boolean" defaultValueString="0.0" usesScalarValueType="YES"/>
+        <attribute name="urlEnrichmentEnabled" attributeType="Boolean" defaultValueString="0.0" usesScalarValueType="YES"/>
+        <relationship name="channel" optional="YES" maxCount="1" deletionRule="Nullify" destinationEntity="ChannelDTO" inverseName="config" inverseEntity="ChannelDTO"/>
+        <relationship name="commands" optional="YES" toMany="YES" deletionRule="Nullify" destinationEntity="CommandDTO" inverseName="channelConfig" inverseEntity="CommandDTO"/>
```
evsaev, line 33: I think for commands we can/should go with `cascade`

Sources/StreamChat/Database/StreamChatModel.xcdatamodeld/StreamChatModel.xcdatamodel/contents
```diff
@@ -34,6 +51,7 @@
         <attribute name="typeRawValue" attributeType="String" defaultValueString=""/>
         <attribute name="updatedAt" attributeType="Date" usesScalarValueType="NO"/>
         <attribute name="watcherCount" attributeType="Integer 64" defaultValueString="0" usesScalarValueType="YES"/>
+        <relationship name="config" maxCount="1" deletionRule="Nullify" destinationEntity="ChannelConfigDTO" inverseName="channel" inverseEntity="ChannelConfigDTO"/>
```
evsaev, line 54: Here `cascade` also makes sense as there's no point in keeping the config in DB when a channel is removed. We don't reference it anywhere else

Sources/StreamChat/Database/DTOs/ChannelConfigDTO.swift
```diff
@@ -0,0 +1,76 @@
...
+            messageRetention: messageRetention,
+            maxMessageLength: Int(maxMessageLength),
+            commands: commands.map { $0.asModel() },
+            createdAt: createdAt,
+            updatedAt: updatedAt
+        )
+    }
+}
+
+extension ChannelConfig {
```
evsaev, line 47: Since the PR is focusing on `ChannelConfig`, can we clean it up a bit? Make it `Decodable` instead of `Codable` (as we no longer encode it), make it a struct to be aligned with other models/payloads, remove manually written initializer and leverage the generated one?

evsaev (changes requested):

Tests/StreamChatTestTools/TestData/DummyData/ChannelPayload.swift
```diff
@@ -50,7 +50,7 @@ extension ChannelDetailPayload {
                     )
                 ],
                 createdAt: XCTestCase.channelCreatedDate,
-                updatedAt: .unique
+                updatedAt: XCTestCase.channelUpdateDate
```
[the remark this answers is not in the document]
nuno-vieira: @evsaev This dummy is used in multiple places (30+). For now, I'll leave it like it is, to avoid a lot of changes in this PR. I just used the same pattern as before, change it in this PR is a bit out of scope

Sources/StreamChat/Database/StreamChatModel.xcdatamodeld/StreamChatModel.xcdatamodel/contents
nuno-vieira: It is a mistake, thanks! 🙏

Sources/StreamChat/Database/StreamChatModel.xcdatamodeld/StreamChatModel.xcdatamodel/contents
```diff
@@ -14,9 +14,26 @@
...
+        <attribute name="mutesEnabled" attributeType="Boolean" defaultValueString="NO" usesScalarValueType="YES"/>
+        <attribute name="quotesEnabled" attributeType="Boolean" defaultValueString="NO" usesScalarValueType="YES"/>
+        <attribute name="reactionsEnabled" attributeType="Boolean" defaultValueString="NO" usesScalarValueType="YES"/>
+        <attribute name="readEventsEnabled" attributeType="Boolean" defaultValueString="NO" usesScalarValueType="YES"/>
+        <attribute name="repliesEnabled" attributeType="Boolean" defaultValueString="NO" usesScalarValueType="YES"/>
+        <attribute name="searchEnabled" attributeType="Boolean" defaultValueString="NO" usesScalarValueType="YES"/>
+        <attribute name="typingEventsEnabled" attributeType="Boolean" defaultValueString="NO" usesScalarValueType="YES"/>
+        <attribute name="updatedAt" attributeType="Date" usesScalarValueType="NO"/>
+        <attribute name="uploadsEnabled" attributeType="Boolean" defaultValueString="0.0" usesScalarValueType="YES"/>
+        <attribute name="urlEnrichmentEnabled" attributeType="Boolean" defaultValueString="0.0" usesScalarValueType="YES"/>
```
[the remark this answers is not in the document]
nuno-vieira: This is probably a mistake on my end as well I guess 🤔 Will check

Sources/StreamChat/Database/StreamChatModel.xcdatamodeld/StreamChatModel.xcdatamodel/contents
nuno-vieira: Makes sense 👍

Sources/StreamChat/Database/DTOs/ChannelConfigDTO.swift
nuno-vieira: This was changed to a `class` to reduce the size of the SDK -> https://github.com/GetStream/stream-chat-swift/pull/1549

The Codable to Decodable, and the manual initializer makes sense 👍

Sources/StreamChat/Database/DTOs/ChannelConfigDTO.swift
```diff
@@ -0,0 +1,76 @@
...
+            connectEventsEnabled: connectEventsEnabled,
+            uploadsEnabled: uploadsEnabled,
+            repliesEnabled: repliesEnabled,
+            quotesEnabled: quotesEnabled,
+            searchEnabled: searchEnabled,
+            mutesEnabled: mutesEnabled,
+            urlEnrichmentEnabled: urlEnrichmentEnabled,
+            messageRetention: messageRetention,
+            maxMessageLength: Int(maxMessageLength),
+            commands: commands.map { $0.asModel() },
```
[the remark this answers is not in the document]
nuno-vieira: I see, then here we need to use an `NSOrderedSet` to keep the order 👍

Sources/StreamChat/Database/DTOs/ChannelConfigDTO.swift
nuno-vieira: Actually, @evsaev I'm not going to do this change. Changing from `Codable -> Decodable` is a breaking change. If a customer is using for example a JSONEncoder to mock the ChannelConfig, this will stop working. Since Codable API will be public as well in this case.

nuno-vieira: The root cause of this issue is that we are missing a data structure to define the config from the Web API. Right now our domain modal is also the web model as well

evsaev: Yeah, ideally there should always be a payload/model pair. The same for command. Can we create a chore task for that?

> Actually, @evsaev I'm not going to do this change. Changing from Codable -> Decodable is a breaking change. If a customer is using for example a JSONEncoder to mock the ChannelConfig, this will stop working. Codable API is public

Let's add this to v5 wish list so we don't forget to fix it 👍

Tests/StreamChatTests/Database/DTOs/ChannelDTO_Tests.swift
```diff
@@ -677,4 +677,37 @@ final class ChannelDTO_Tests: XCTestCase {
...
+            messageRetention: "",
+            maxMessageLength: 5000,
+            commands: [
+                .init(name: "giphy", description: "", set: "", args: ""),
+                .init(name: "workout", description: "", set: "", args: ""),
+                .init(name: "location", description: "", set: "", args: "")
+            ],
+            createdAt: .unique,
+            updatedAt: .unique
+        )
```
evsaev, line 703:
```suggestion
        let config: ChannelConfig = mock(
            commands: [
                .init(name: "giphy", description: "", set: "", args: ""),
                .init(name: "workout", description: "", set: "", args: ""),
                .init(name: "location", description: "", set: "", args: "")
            ]
        )
```

evsaev (approved):
LGTM

Tests/StreamChatTests/Database/DTOs/ChannelDTO_Tests.swift
nuno-vieira: Thanks! 👍  Done ✅

PR #1924, opened 2022-04-12
[CIS-1624] Fix not watching channel when added as a member

nuno-vieira:
### 🔗 Issue Link
CIS-1624

### 🎯 Goal
Fix a channel not receiving new messages and unread count updates, when a user is added as a member and the user is in the channel list.

### 🛠 Implementation
When a new channel is inserted into the channel list we now make sure that not only do we link the channel but also start watching it.

### 🧪 Testing
1. Log in to two different accounts (e.g.: Leia on one device and C-3PO on another one)
2. Create a new channel using one of those accounts inviting another one
3. Send a message on the new channel

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/Workers/ChannelListUpdater.swift
+++ b/Sources/StreamChat/Workers/ChannelListUpdater.swift
@@ -89,22 +89,33 @@ class ChannelListUpdater: Worker {
         }
     }
 
-    private func writeChannelListPayload(
-        payload: ChannelListPayload,
-        query: ChannelListQuery,
-        initialActions: ((DatabaseSession) -> Void)? = nil,
-        completion: ((Result<[ChatChannel], Error>) -> Void)? = nil
-    ) {
-        var channels: [ChatChannel] = []
-        database.write { session in
-            initialActions?(session)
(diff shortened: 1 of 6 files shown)


Sources/StreamChat/Workers/ChannelListUpdater.swift
```diff
@@ -89,22 +89,25 @@ class ChannelListUpdater: Worker {
...
+    ///   - completion: The callback once the request is complete.
+    func startWatchingChannels(withIds ids: [ChannelId], completion: ((Error?) -> Void)?) {
+        var query = ChannelListQuery(filter: .in(.cid, values: ids))
+        query.options = .all
+
+        fetch(channelListQuery: query) { [weak self] in
+            switch $0 {
+            case let .success(payload):
+                self?.database.write { session in
+                    try session.saveChannelList(payload: payload, query: query)
```
evsaev, line 105: @b-onc we shouldn't be calling `session.saveChannelList` because it will save query to DB, right?

evsaev: Fixed ✅

b-onc (approved):

PR #1925, opened 2022-04-12
[CIS-1764] Fix connection recovery issues

evsaev:
### 🔗 Issue Link

CIS-1764

### 🎯 Goal

Fix connection recovery issues, mainly:
- recovery flow is triggered on the first login
- recovery flow is not canceled when `ChatClient` is deallocated
- recovery flow is not canceled on disconnect 
- active components are not untracked when chat client is connected with another user
- `/sync` cooldown is applied to connection recovery flow

### 🛠 Implementation

Update `SyncRepository` to run recovery flow only for further connections. If it's the 1st connection, the `lastSyncedAt` is set to the current timestamp and will be used as a reference date for `/sync` endpoint when the web socket is reconnected.

### 📝 Changes

N/A

### 🎨 UI Changes

N/A

### 🧪 Testing

Use `LogConfig.level = .debug; LogConfig.subsystems = [.offlineSupport]` in DemoApp
Run `DemoApp` on the device

**Recovery flow is not triggered for the 1st session:**
1. Sign in as any user
1. Check the logs, make sure connection recovery flow is **not triggered**
1. Send app to background
1. Wait for web-socket to be disconnected
1. Send app to foreground
1. Check the logs, make sure connection recovery flow is **triggered**

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/ChatClient.swift
+++ b/Sources/StreamChat/ChatClient.swift
@@ -66,25 +66,29 @@ public class ChatClient {
 
     // MARK: Repositories
 
-    private(set) lazy var messageRepository = MessageRepository(database: databaseContainer, apiClient: apiClient)
-    private(set) lazy var offlineRequestsRepository = OfflineRequestsRepository(
-        messageRepository: messageRepository,
-        database: databaseContainer,
-        apiClient: apiClient
+    private(set) lazy var messageRepository = environment.messageRepositoryBuilder(
+        databaseContainer,
+        apiClient
+    )
+    
+    private(set) lazy var offlineRequestsRepository = environment.offlineRequestsRepositoryBuilder(
+        messageRepository,
+        databaseContainer,
+        apiClient
     )
 
     /// A repository that handles all the executions needed to keep the Database in sync with remote.
     private(set) lazy var syncRepository: SyncRepository = {
         let channelRepository = ChannelListUpdater(database: databaseContainer, apiClient: apiClient)
-        return SyncRepository(
-            config: config,
-            activeChannelControllers: activeChannelControllers,
-            activeChannelListControllers: activeChannelListControllers,
-            channelRepository: channelRepository,
-            offlineRequestsRepository: offlineRequestsRepository,
-            eventNotificationCenter: eventNotificationCenter,
-            database: databaseContainer,
-            apiClient: apiClient
+        return environment.syncRepositoryBuilder(
+            config,
+            activeChannelControllers,
+            activeChannelListControllers,
+            channelRepository,
+            offlineRequestsRepository,
+            eventNotificationCenter,
+            databaseContainer,
+            apiClient
         )
     }()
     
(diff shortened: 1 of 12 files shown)


evsaev:
We do not cancel connection recovery flow/don't stop tracking active components when the connection is initiated for another user. Will convert back from draft when it's fixed

Sources/StreamChat/Repositories/SyncRepository.swift
```diff
@@ -64,6 +63,32 @@ class SyncRepository {
...
+        cancelRecoveryFlow()
+    }
+    
+    func syncLocalState(completion: @escaping () -> Void) {
+        cancelRecoveryFlow()
+        
+        getUser { [weak self] in
+            guard let currentUser = $0 else {
+                log.error("Current user must exist", subsystems: .offlineSupport)
+                completion()
```
b-onc, line 77: This errors but nothing is done?

b-onc (approved):

Sources/StreamChat/Repositories/SyncRepository.swift
evsaev: Yep, it's weird but I decided not to touch the API of `SyncRepository`, just adjust the implementation. 

We can update the completion closure in `syncLocalState` to receive error but the thing is we don't interrupt the reconnection flow when some operation fails with an unrecoverable error we just move on to the next step.

Unless we propagate error from operation somehow, adding `Error` to completion won't change much because even if recovery flow has failed, completion would be called with `nil` error.

bielikb (approved):

Sources/StreamChat/Workers/ChatClientUpdater.swift
```diff
@@ -54,7 +54,11 @@ class ChatClientUpdater {
 
             // Re-create backgroundWorker's since they are related to the previous user.
             client.createBackgroundWorkers()
-
+            
+            // Stop tracking active components
+            client.activeChannelControllers.removeAllObjects()
+            client.activeChannelListControllers.removeAllObjects()
```
polqf, line 60: We should be calling a method on the client, and not know that much from here.

It could also be recreating background workers in an opaque way. This "updater" component knows too much now

Sources/StreamChat/Workers/ChatClientUpdater.swift
```diff
@@ -161,6 +165,9 @@ class ChatClientUpdater {
     /// Disconnects the chat client the controller represents from the chat servers. No further updates from the servers
     /// are received.
     func disconnect(source: WebSocketConnectionState.DisconnectionSource = .userInitiated) {
+        client.apiClient.flushRequestsQueue()
+        client.syncRepository.cancelRecoveryFlow()
```
polqf, line 169: Same as before

Tests/StreamChatTests/Repositories/SyncRepository_Tests.swift
```diff
@@ -83,10 +135,15 @@ final class SyncRepository_Tests: XCTestCase {
...
+        
+        try prepareForSyncLocalStorage(
+            createUser: true,
+            lastSynchedEventDate: .init(),
+            createChannel: false
+        )
 
         waitForSyncLocalStateRun()
 
-        XCTAssertNil(lastSyncAtValue)
```
polqf, line 89: Shouldn't this be replaced with the right assertion instead of deleting it?

Sources/StreamChat/Repositories/SyncOperations.swift
```diff
@@ -6,10 +6,15 @@ import Foundation
 
 /// A final class that holds the context for the ongoing operations during the sync process
 final class SyncContext {
+    let lastSyncAt: Date
```
polqf, line 9: I would add a more descriptive name or documentation to it. Remember how we both were understanding different things around this topic :joy:

Tests/StreamChatTests/Repositories/SyncRepository_Tests.swift
```diff
@@ -239,10 +278,53 @@ final class SyncRepository_Tests: XCTestCase {
...
+        let eventsPayload1 = messageEventPayload(with: [firstDate, secondDate])
+        waitForSyncLocalStateRun(requestResult: .success(eventsPayload1))
+        
+        XCTAssertEqual(lastSyncAtValue, secondDate)
+                
+        let thirdDate = secondDate.addingTimeInterval(1)
+        let eventsPayload2 = messageEventPayload(with: [thirdDate])
+        waitForSyncLocalStateRun(requestResult: .success(eventsPayload2))
+        
+        XCTAssertEqual(lastSyncAtValue, thirdDate)
```
polqf, line 301: 👏

Tests/StreamChatTests/Workers/ChatClientUpdater_Tests.swift
```diff
@@ -215,113 +230,181 @@ final class ChatClientUpdater_Tests: XCTestCase {
...
+        XCTAssertTrue(client.completeTokenWaiters_called)
+        // Assert `completeTokenWaiters` was called with updated token.
+        XCTAssertEqual(client.completeTokenWaiters_token, updatedToken)
+        // Assert background workers stay the same.
+        XCTAssertEqual(client.testBackgroundWorkerId, oldWorkerIDs)
+        // Assert database is not flushed.
+        XCTAssertFalse(client.mockDatabaseContainer.removeAllData_called)
+        // Assert active components are preserved
+        XCTAssertTrue(client.activeChannelControllers.contains(channelController))
+        XCTAssertTrue(client.activeChannelListControllers.contains(channelListController))
```
polqf, line 295: Both the tests and the logic can be simplified a lot if all the actions performed on properties of the ChatClient are under one method. Less knowledge, more testable. And this particular logic to be performed when this event happens should be tested from the ChatClient, then

Sources/StreamChat/ChatClient.swift
```diff
@@ -348,7 +352,6 @@ public class ChatClient {
     public func disconnect() {
         clientUpdater.disconnect()
         userConnectionProvider = nil
-        apiClient.flushRequestsQueue()
```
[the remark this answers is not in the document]
evsaev: It's moved to `clientUpdater.disconnect`

Sources/StreamChat/Workers/ChatClientUpdater.swift
evsaev: Yep, we can have a chore task to refactor this 👍

PR #1926, opened 2022-04-13
[CIS-1555] Fix crash when changing users within an active chat

b-onc:
### 🔗 Issue Link
CIS-1555 (see the zendesk ticket related to this)
CIS-1743
CIS-1741

### 🎯 Goal

One of the usecases we support is: app opens with guest/anonymous user, which can send/read messages. At one point, user can login and become a signed in user, without reloading the chat screen

Another usecase: at any point, user can logout and login with the same / another user credential.

### 🛠 Implementation

Let's understand our DB cleanup mechanism first.
When a new user logs in, we call `databaseContainer.removeAllData(force: true)` non-force clean is not implemented, and it's not clear what it'll be. Perhaps it'll only remove user-related data (such as removing only `CurrentUserDTO` but keeping channels/messages intact). Anyways, it's not the point of this PR.

`removeAllData` rips out the underlying DB file. In this case, none of the DB observers will get `object deleted` notification, because we don't play by the CoreData rules. We "workaround" this by sending our own "willDelete" and "didDelete" notifications.

`willDelete` prepares `delete` events for all the objects we're currently aware of (in the context of observer)
`didDelete` notification is sent _after_ the DB file is removed. This causes the `delete` events to be published to the observing object.

This is where it gets fun. Whenever an update is published from observers (`EntityChangeObserver` or `ListChangeObserver`), we call `asModel` on the DTO, lazily. Since the DTO is deleted, all fields are reset: strings are "", numbers are 0. `Date`s are different: if we try to get the `Date` from an erased object, we get
```swift
Date._unconditionallyBridgeFromObjectiveC(_:)
```
our beloved crash.

The solution is to publish the changes _before_ the underlying DTOs are deleted, so calling `asModel` on them succeeds. The DTOs are removed immediately after the changes are published.

Another problem was that we were sending `didDelete` notification _before_ the actual DB file was removed, so observers were re-subscribing to DB, receiving further updates.

Another problem was that removing the DB file was an async operation but we treated as sync. User connection cannot be completed before DB wipe is completed. This caused crashes, too.

### 🧪 Testing

First, checkout the first commit of this repo. It adds the possibility of logging in as `anonymous` and `guest` users.
Due to our app configuration, anonymous users don't see any channels. So:
1. Login as guest user. ID doesn't matter, leave it empty, defaults to `guest`
1. Open the first channel you see
1. Tap the debug icon in the navigation bar.
1. Tap "switch user"
1. Select any member
You'll see the `Date` crash.

Now, checkout the final commit, and follow the same steps. The chat screen will remove all messages and load them again, without any crash.

### 🥸 Known Issues 

For some reason, it takes ~10 secs for the ChannelListController to dealloc. If one swaps users too quickly (logout/login), there'll be 2 ChatClient instances alive for ~10 secs. This causes a race condition, and sometimes a crash. This shouldn't be a problem in production, since switching users within 10 seconds of logging in is too rare, and it's not caused by this PR, so the investigation will continue without blocking this PR.

The DB errors when launching are not properly handled. Again, not related to the PR. It's a bit hard to resolve, since `init` is sync but errors come async. Investigation will continue without blocking the PR.

### 🎉 GIF

Simulator Screen Recording - iPhone 13 Pro - 2022-04-07 at 17 40 36

### ☑️ Contributor Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Fix cooldown being applied to /sync endpoint in connection recovery flow [#1925](https://github.com/GetStream/stream-chat-swift/pull/1925)
 - Fix active components not being reset when another user is connected [#1925](https://github.com/GetStream/stream-chat-swift/pull/1925)
 - Fix unusable firebase push provider [#1934](https://github.com/GetStream/stream-chat-swift/pull/1934)
+- Fix DB errors happening when logging in after a logout / user switch [#1926](https://github.com/GetStream/stream-chat-swift/issues/1926)
 
 ## StreamChatUI
 ### 💥 Removed
--- a/DemoApp/ChatPresenter.swift
+++ b/DemoApp/ChatPresenter.swift
@@ -29,19 +29,25 @@ class DemoChatChannelListRouter: ChatChannelListRouter {
     
     override func showCurrentUserProfile() {
         rootViewController.presentAlert(title: nil, actions: [
+            .init(title: "Switch User", style: .default, handler: { [unowned self] _ in
+                let actions = UserCredentials.builtInUsers.map { user in
+                    UIAlertAction(title: user.name, style: .default) { _ in
+                        ChatClient.shared.connectUser(
+                            userInfo: user.userInfo,
+                            token: Token(stringLiteral: user.token)
+                        )
+                    }
+                }
+                self.rootViewController.presentAlert(
+                    title: "Select a user",
+                    message: "Channel list won't be correct until you logout and login again.",
+                    actions: actions,
+                    cancelHandler: nil
+                )
+            }),
             .init(title: "Logout", style: .destructive, handler: { _ in
                 let window = self.rootViewController.view.window!
-                guard let navigationController = UIStoryboard(name: "Main", bundle: Bundle.main)
-                    .instantiateInitialViewController() as? UINavigationController else {
-                    return
-                }
-                guard let sceneDelegate = window.windowScene?.delegate as? SceneDelegate else {
-                    return
-                }
-                sceneDelegate.coordinator = DemoAppCoordinator(navigationController: navigationController)
-                UIView.transition(with: window, duration: 0.3, options: .transitionFlipFromLeft, animations: {
-                    window.rootViewController = navigationController
-                })
+                DemoAppCoordinator.logout(window: window)
             })
         ])
     }
@@ -77,6 +83,25 @@ class DemoChatChannelListRouter: ChatChannelListRouter {
     override func didTapMoreButton(for cid: ChannelId) {
         let channelController = rootViewController.controller.client.channelController(for: cid)
         rootViewController.presentAlert(title: "Select an action", actions: [
+            .init(title: "Switch User", style: .default, handler: { [unowned self] _ in
+                let channelUsers = UserCredentials.builtInUsers.filter { user in
+                    channelController.channel?.lastActiveMembers.contains(where: { $0.id == user.id }) ?? false
+                }
+                let actions = channelUsers.map { user in
+                    UIAlertAction(title: user.name, style: .default) { _ in
+                        ChatClient.shared.connectUser(
+                            userInfo: user.userInfo,
+                            token: Token(stringLiteral: user.token)
+                        )
+                    }
+                }
+                self.rootViewController.presentAlert(
+                    title: "Select a user",
+                    message: "Channel list won't be correct until you logout and login again.",
+                    actions: actions,
+                    cancelHandler: nil
+                )
+            }),
             .init(title: "Change nav bar translucency", style: .default, handler: { [unowned self] _ in
                 self.rootViewController.presentAlert(
                     title: "Change nav bar translucency",
--- a/Sources/StreamChat/Controllers/EntityDatabaseObserver.swift
+++ b/Sources/StreamChat/Controllers/EntityDatabaseObserver.swift
@@ -137,6 +137,9 @@ class EntityDatabaseObserver<Item, DTO: NSManagedObject> {
             cacheName: nil
         )
         
+        // We want item to report nil until `startObserving` is called
+        _item.computeValue = { nil }
+        
         listenForRemoveAllDataNotifications()
     }
     
@@ -149,11 +152,8 @@ class EntityDatabaseObserver<Item, DTO: NSManagedObject> {
     ///
     /// - Throws: An error if the provided fetch request fails.
     func startObserving() throws {
-        try frc.performFetch()
-        
         _item.computeValue = { [weak self] in
-            guard let self = self else { return nil }
-            guard let fetchedObjects = self.frc.fetchedObjects else { return nil }
+            guard let self = self, let fetchedObjects = self.frc.fetchedObjects else { return nil }
             log.assert(
                 fetchedObjects.count <= 1,
                 "EntityDatabaseObserver predicate must match exactly 0 or 1 entities. Matched: \(fetchedObjects)"
@@ -169,6 +169,7 @@ class EntityDatabaseObserver<Item, DTO: NSManagedObject> {
         }
         _item.reset()
         
+        try frc.performFetch()
         frc.delegate = changeAggregator
     }
     
@@ -225,7 +226,11 @@ class EntityDatabaseObserver<Item, DTO: NSManagedObject> {
             guard let self = self else { return }
             
             // Reset FRC which causes the current `frc.fetchedObjects` to be reloaded
-            try! self.startObserving()
+            do {
+                try self.startObserving()
+            } catch {
+                log.error("Error when starting observing: \(error)")
+            }
         }
         
         releaseNotificationObservers = { [weak notificationCenter] in
--- a/Sources/StreamChat/Controllers/EntityDatabaseObserver.swift
+++ b/Sources/StreamChat/Controllers/EntityDatabaseObserver.swift
@@ -137,22 +137,6 @@ class EntityDatabaseObserver<Item, DTO: NSManagedObject> {
             cacheName: nil
         )
         
-        _item.computeValue = { [weak frc] in
-            guard let fetchedObjects = frc?.fetchedObjects else { return nil }
-            log.assert(
-                fetchedObjects.count <= 1,
-                "EntityDatabaseObserver predicate must match exactly 0 or 1 entities. Matched: \(fetchedObjects)"
-            )
-            
-            return fetchedObjects.first.flatMap { dto in
-                var result: Item?
-                context.performAndWait {
-                    result = itemCreator(dto)
-                }
-                return result
-            }
-        }
-        
         listenForRemoveAllDataNotifications()
     }
     
@@ -167,6 +151,22 @@ class EntityDatabaseObserver<Item, DTO: NSManagedObject> {
     func startObserving() throws {
         try frc.performFetch()
         
+        _item.computeValue = { [weak self] in
+            guard let self = self else { return nil }
+            guard let fetchedObjects = self.frc.fetchedObjects else { return nil }
+            log.assert(
+                fetchedObjects.count <= 1,
+                "EntityDatabaseObserver predicate must match exactly 0 or 1 entities. Matched: \(fetchedObjects)"
+            )
+            
+            return fetchedObjects.first.flatMap { dto in
+                var result: Item?
+                self.context.performAndWait {
+                    result = self.itemCreator(dto)
+                }
+                return result
+            }
+        }
         _item.reset()
         
         frc.delegate = changeAggregator
@@ -201,12 +201,20 @@ class EntityDatabaseObserver<Item, DTO: NSManagedObject> {
                     newIndexPath: nil
                 )
             }
+            
+            // Publish the changes
+            self.changeAggregator.controllerDidChangeContent(self.frc as! NSFetchedResultsController<NSFetchRequestResult>)
+            
+            // Remove delegate so it doesn't get further removal updates
+            self.frc.delegate = nil
+            
+            // Remove the cached item since it's now deleted, technically
+            self._item.computeValue = { nil }
+            self._item.reset()
         }
         
         // When `DidRemoveAllDataNotification` is received, we need to reset the FRC. At this point, the entities are removed but
-        // the FRC doesn't know about it yet. Resetting the FRC removes the content of `FRC.fetchedObjects`. We also need to
-        // call `controllerDidChangeContent` on the change aggregator to finish reporting about the removed object which started
-        // in the `WillRemoveAllDataNotification` handler above.
+        // the FRC doesn't know about it yet. Resetting the FRC removes the content of `FRC.fetchedObjects`.
         let didRemoveAllDataNotificationObserver = notificationCenter.addObserver(
             forName: DatabaseContainer.DidRemoveAllDataNotification,
             object: context,
@@ -218,9 +226,6 @@ class EntityDatabaseObserver<Item, DTO: NSManagedObject> {
             
             // Reset FRC which causes the current `frc.fetchedObjects` to be reloaded
             try! self.startObserving()
-            
-            // Publish the changes started in `WillRemoveAllDataNotification`
-            self.changeAggregator.controllerDidChangeContent(self.frc as! NSFetchedResultsController<NSFetchRequestResult>)
         }
         
         releaseNotificationObservers = { [weak notificationCenter] in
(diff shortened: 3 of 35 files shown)


Sources/StreamChat/Database/DatabaseContainer.swift
```diff
@@ -219,17 +219,26 @@ class DatabaseContainer: NSPersistentContainer {
...
+                    }
+                    
+                    // If the current persistent store is a SQLite store, this method will reset and recreate it.
+                    try self?.recreatePersistentStore(completion: completionWithNotification)
+                } catch {
+                    completion?(error)
+                }
+            }
+        } else {
+            fatalError("Non-force flush is not implemented.")
```
evsaev, line 240: nit: can we preserve the structure to reduce the # of changes?
```swift
func removeAllData(force: Bool = true, completion: ((Error?) -> Void)? = nil) {
 if !force {
  fatalError("Non-force flush is not implemented yet.")
  }
  
  ...
}
```

Sources/StreamChat/Database/DatabaseContainer.swift
```diff
@@ -219,17 +219,26 @@ class DatabaseContainer: NSPersistentContainer {
...
+                do {
+                    self?.sendNotificationForAllContexts(name: Self.WillRemoveAllDataNotification)
+                    
+                    let completionWithNotification: ((Error?) -> Void) = { error in
+                        self?.sendNotificationForAllContexts(name: Self.DidRemoveAllDataNotification)
+                        completion?(error)
+                    }
+                    
+                    // If the current persistent store is a SQLite store, this method will reset and recreate it.
+                    try self?.recreatePersistentStore(completion: completionWithNotification)
```
evsaev, line 234:
```suggestion
                    // If the current persistent store is a SQLite store, this method will reset and recreate it.
                    try self?.recreatePersistentStore { error in
                        self?.sendNotificationForAllContexts(name: Self.DidRemoveAllDataNotification)
                        completion?(error)
                    }
```

Sources/StreamChat/Database/DatabaseContainer.swift
```diff
@@ -255,53 +264,56 @@ class DatabaseContainer: NSPersistentContainer {
...
-    private func setupPersistentStore() throws {
-        var storeLoadingError: Error?
-        
-        loadPersistentStores { _, error in
-            storeLoadingError = error
-        }
-        
-        if storeLoadingError != nil {
-            try recreatePersistentStore()
+    private func setupPersistentStore() {
```
evsaev, line 268: Can we also add a completion closure here? This one is called from `DatabaseContainer.init` and followed by `resetEphemeralValues` which can also bite us back if we don't wait for the store to be loaded/recreated.

Sources/StreamChat/Workers/ChatClientUpdater.swift
```diff
@@ -93,22 +100,28 @@ class ChatClientUpdater {
...
                     }
-
-                    self.connect(
+                    
+                    try self.prepareEnvironment(
                         userInfo: userInfo,
-                        completion: completion
-                    )
+                        newToken: newToken
+                    ) { error in
```
evsaev, line 113: `self.prepareEnvironment` should run regardless of `active/passive` client's mode because we assign a token in there. If client is passive, it's still allowed to send API calls which won't pass through if there's no token

evsaev (changes requested):

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### 🐞 Fixed
 - Fixed support for multiple active channel lists at the same time [#1879](https://github.com/GetStream/stream-chat-swift/pull/1879)
+- Fix DB errors happening when logging in after a logout / user switch [#1926](https://github.com/GetStream/stream-chat-swift/issues/1926)
```
nuno-vieira, line 15: Whenever people have this crash, it is with the message `Date._unconditionallyBridgeFromObjectiveC`, so it makes sense to mention this in the changelog so people easily find this. 
```suggestion
- Fix the crash `Date._unconditionallyBridgeFromObjectiveC` when logging in after a logout / user switch [#1926](https://github.com/GetStream/stream-chat-swift/issues/1926)
```

Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
```diff
@@ -319,7 +319,10 @@ open class ChatChannelListVC: _ViewController,
             completion: { _ in
                 // Move changes from NSFetchController also can mean an update of the content.
                 // Since a `moveItem` in collections do not update the content of the cell, we need to reload those cells.
-                self.collectionView.reloadItems(at: Array(indices.move.map(\.toIndex)))
+                let moveIndexes = Array(indices.move.map(\.toIndex))
+                if !moveIndexes.isEmpty {
+                    self.collectionView.reloadItems(at: Array(indices.move.map(\.toIndex)))
+                }
```
nuno-vieira, line 325:
```suggestion
                let moveIndexes = Array(indices.move.map(\.toIndex))
                if !moveIndexes.isEmpty {
                    self.collectionView.reloadItems(at: moveIndexes))
                }
```

nuno-vieira:
@b-onc @evsaev @adamrushy 
We should create tickets for the known issues, and tackle them ASAP in the next sprint, especially if they produce crashes on our Demo App that are easily reproducible.

Sources/StreamChat/Database/DatabaseContainer.swift
```diff
@@ -219,17 +225,24 @@ class DatabaseContainer: NSPersistentContainer {
...
+                
+                // If the current persistent store is a SQLite store, this method will reset and recreate it.
+                try self?.recreatePersistentStore { error in
+                    self?.sendNotificationForAllContexts(name: Self.DidRemoveAllDataNotification)
+                    completion?(error)
+                }
+            } catch {
+                completion?(error)
+            }
+        }
```
evsaev, line 245: In this particular case, I wouldn't weakify self and let the DB flushing process complete.

Sources/StreamChat/Workers/ChatClientUpdater.swift
```diff
@@ -96,19 +103,25 @@ class ChatClientUpdater {
...
                         userInfo: userInfo,
                         newToken: newToken
-                    )
-
-                    // We manually change the `connectionStatus` for passive client
-                    // to `disconnected` when environment was prepared correctly
-                    // (e.g. current user session is successfully restored).
-                    if !self.client.config.isClientInActiveMode {
-                        self.client.connectionStatus = .disconnected(error: nil)
+                    ) { error in
```
evsaev, line 106: `[weak self]` to not connect the web socket if there's no strong refs left?

Sources/StreamChat/Database/DatabaseContainer.swift
```diff
@@ -255,53 +268,56 @@ class DatabaseContainer: NSPersistentContainer {
...
+                } catch {
+                    fatalError("Failed to initialize the storage with error: \(error). This is a non-recoverable error.")
+                }
+            }
         }
     }
     
     /// Removes the loaded persistent store and tries to recreate it.
-    func recreatePersistentStore() throws {
+    func recreatePersistentStore(completion: ((Error?) -> Void)? = nil) throws {
```
evsaev, line 285: nit: now, when completion with input `error` is in place, it's a bit weird the func throws, right? 🤔 but not sure if it's smth we wanna fix

Sources/StreamChat/Workers/ChatClientUpdater.swift
```diff
@@ -26,6 +27,7 @@ class ChatClientUpdater {
             client.createBackgroundWorkers()
             // Provide the token to pending API requests
             client.completeTokenWaiters(token: newToken)
+            completion(nil)
```
evsaev, line 30: Can we update tests to make sure we don't forget to call completion?

DemoApp/ChatPresenter.swift
```diff
@@ -29,19 +29,25 @@ class DemoChatChannelListRouter: ChatChannelListRouter {
     
     override func showCurrentUserProfile() {
         rootViewController.presentAlert(title: nil, actions: [
+            .init(title: "Switch User", style: .default, handler: { [unowned self] _ in
```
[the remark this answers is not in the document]
b-onc: What do you mean?

Sources/StreamChat/Database/DatabaseContainer.swift
b-onc: It actually makes sense: if the error is sync, we can act on it right away (switch to in-memory DB). if the error is async, we can't really do much

Sources/StreamChat/Workers/ChatClientUpdater.swift
b-onc: `prepareEnvironment` is called from `reloadUserIfNeeded` which had `completion` block already, and it's tested

DemoApp/ChatPresenter.swift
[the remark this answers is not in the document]
evsaev: I'm suggesting to fix the crash first, then add this action to the DemoApp 🙂

Sources/StreamChat/Workers/ChatClientUpdater.swift
evsaev: It's some kind of integration tests between 2 funcs. Testing `reloadUserIfNeeded` shouldn't check `prepareEnvironment` behaves as expected, just check it's invoke inside of it. 

Instead, a bunch of tests for `prepareEnvironment ` should be in place to check the behavior. 

Funny fact, it was me who wrote that 😄 Feel free to create a task in tech debt epic to fix it and assign on me 👍

Sources/StreamChat/Workers/ChatClientUpdater.swift
```diff
@@ -13,7 +13,8 @@ class ChatClientUpdater {
 
     func prepareEnvironment(
         userInfo: UserInfo?,
-        newToken: Token
+        newToken: Token,
+        completion: @escaping (Error?) -> Void
     ) throws {
```
polqf, line 18: Same as before. Either a throwing function or reporting an error via completion block

polqf (changes requested):
We might be missing tests to cover the changes

Sources/StreamChat/Controllers/EntityDatabaseObserver.swift
```diff
@@ -167,6 +151,22 @@ class EntityDatabaseObserver<Item, DTO: NSManagedObject> {
...
+                fetchedObjects.count <= 1,
+                "EntityDatabaseObserver predicate must match exactly 0 or 1 entities. Matched: \(fetchedObjects)"
+            )
+            
+            return fetchedObjects.first.flatMap { dto in
+                var result: Item?
+                self.context.performAndWait {
+                    result = self.itemCreator(dto)
+                }
+                return result
```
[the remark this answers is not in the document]
b-onc: yes, it's existing code and tested

Sources/StreamChat/Controllers/EntityDatabaseObserver.swift
```diff
@@ -201,12 +201,20 @@ class EntityDatabaseObserver<Item, DTO: NSManagedObject> {
                     newIndexPath: nil
                 )
             }
+            
+            // Publish the changes
+            self.changeAggregator.controllerDidChangeContent(self.frc as! NSFetchedResultsController<NSFetchRequestResult>)
```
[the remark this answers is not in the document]
b-onc: existing code and tested

Sources/StreamChat/Controllers/EntityDatabaseObserver.swift
```diff
@@ -201,12 +201,20 @@ class EntityDatabaseObserver<Item, DTO: NSManagedObject> {
...
+            
+            // Remove delegate so it doesn't get further removal updates
+            self.frc.delegate = nil
+            
+            // Remove the cached item since it's now deleted, technically
+            self._item.computeValue = { nil }
+            self._item.reset()
```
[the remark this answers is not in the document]
b-onc: no, `reset` is for resetting the cached value. `computeValue` is used when there's no cached value. See `@Cached` for implementation

Sources/StreamChat/Controllers/ListDatabaseObserver.swift
```diff
@@ -175,6 +162,20 @@ class ListDatabaseObserver<Item, DTO: NSManagedObject> {
     func startObserving() throws {
         try frc.performFetch()
         frc.delegate = changeAggregator
+        
+        _items.computeValue = { [weak self] in
+            guard let self = self else { return [] }
+            var result = LazyCachedMapCollection<Item>()
+            result = (self.frc.fetchedObjects ?? []).lazyCachedMap { dto in
```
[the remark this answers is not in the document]
b-onc: existing code, tested

Sources/StreamChat/Controllers/ListDatabaseObserver.swift
```diff
@@ -213,12 +214,20 @@ class ListDatabaseObserver<Item, DTO: NSManagedObject> {
                     newIndexPath: nil
                 )
             }
+            
+            // Publish the changes
+            self.changeAggregator.controllerDidChangeContent(self.frc as! NSFetchedResultsController<NSFetchRequestResult>)
```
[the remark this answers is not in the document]
b-onc: not in scope of the PR, let's move to Jira

Sources/StreamChat/Database/DatabaseContainer.swift
```diff
@@ -128,11 +128,21 @@ class DatabaseContainer: NSPersistentContainer {
...
+                self?.resetEphemeralValues()
+            }
+        }
                 
         if shouldFlushOnStart {
-            try recreatePersistentStore()
+            try recreatePersistentStore(completion: persistentStoreCreatedCompletion)
         } else {
-            try setupPersistentStore()
+            setupPersistentStore(completion: persistentStoreCreatedCompletion)
```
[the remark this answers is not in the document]
b-onc: I agree but more refactoring and the scope gets too big. we need a ticket to refactor `DatabaseContainer` and handle errors more robustly

Sources/StreamChat/Database/DatabaseContainer.swift
```diff
@@ -255,53 +268,56 @@ class DatabaseContainer: NSPersistentContainer {
...
+                } catch {
+                    fatalError("Failed to initialize the storage with error: \(error). This is a non-recoverable error.")
+                }
+            }
         }
     }
     
     /// Removes the loaded persistent store and tries to recreate it.
-    func recreatePersistentStore() throws {
+    func recreatePersistentStore(completion: ((Error?) -> Void)? = nil) throws {
```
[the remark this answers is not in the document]
b-onc: it makes sense, actually. some errors come sync, which we can act on them immediately: if we get a sync error, we can fallback to in memory storage, which we do in ChatClient. some errors come async, and we cannot actually act on them, so we just `fatalError` now. we need to refactor DatabaseContainer so these stuff are handled more robustly, but in the meantime, this approach makes it more robust.

Sources/StreamChat/Database/DatabaseContainer.swift
```diff
@@ -255,53 +268,56 @@ class DatabaseContainer: NSPersistentContainer {
...
-        var storeLoadingError: Error?
         
-        loadPersistentStores { _, error in
-            storeLoadingError = error
-        }
+        log.debug("Reloading persistent store", subsystems: .database)
         
-        if let error = storeLoadingError {
-            throw error
+        loadPersistentStores { _, error in
```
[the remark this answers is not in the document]
b-onc: can you explain?

Sources/StreamChat/Workers/ChatClientUpdater.swift
b-onc: it actually makes sense here, refactoring

b-onc (commented):
I'm updating the PR to include changes for the discussions i've 👍 'd

Sources/StreamChat/Database/DatabaseContainer.swift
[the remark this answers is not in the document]
b-onc: I actually did it..

Sources/StreamChat/Database/DatabaseContainer.swift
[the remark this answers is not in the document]
b-onc: done

polqf (approved):
LGTM as long as the comments not solved are on JIRA

Sources/StreamChat/Workers/ChatClientUpdater.swift
b-onc: CIS-1780 created for this

PR #1927, opened 2022-04-13
[CIS-1131] Support marking all channels as read at a user level

adolfogarza:
### 🔗 Issue Link
[_CIS-1131_](https://stream-io.atlassian.net/browse/CIS-1131)

### 🎯 Goal

_The goal is to relocate `markAllRead()` from the `ChannelListController` to `CurrentUserController` to have parity with other platforms._

### 🛠 Implementation

_The `markAllRead()` call has been relocated to `CurrentUserController`, the specific implementation has been moved to `CurrentUserUpdater` to be consistent with the current architecture. Tests have been included to cover typical use cases. The current `markAllRead()` has been marked as deprecated in order to not disturb current implementors of that call, but let them know that there is a better way to do that now_

### 📝 Changes

_ N/A

### 🎨 UI Changes

_N/A_

| Before | After |
| --- | --- |
| img | img |

### 🧪 Testing

_Since this method not used in the demo app or in the UI SDK by default, a very simple way to test this is to generate new messages on multiple channels that your user belongs to, then call `ChatClient.shared.currentUserController().markAllRead()` on `DemoChannelListVC` at `viewDidLoad` and you should be able to see how these are marked as read._

### ☑️ Contributor Checklist


Tests/StreamChatTests/Controllers/CurrentUserController/CurrentUserController_Tests.swift
```diff
@@ -603,6 +603,64 @@ final class CurrentUserController_Tests: XCTestCase {
...
+            env.currentUserUpdater!.markAllRead_completion = nil
+        }
+        
+        THEN("the completion is indeed called") {
+            AssertAsync.willBeTrue(completionCalled)
+        }
+        
+        AND("the reference to the controller is actually deallocated") {
+            AssertAsync.canBeReleased(&weakController)
+        }
```
evsaev, line 640: nit: would be nice to have a separate test for controller being kept alive until API response comes:
```swift
func test_loadPinnedMessages_keepsControllerAlive() {
        // Simulate `loadPinnedMessages` call
        controller.loadPinnedMessages { _ in }
        
        // Keep a weak ref so we can check if it's actually deallocated
        weak var weakController = controller
        
        // (Try to) deallocate the controller
        // by not keeping any references to it
        controller = nil
        
        // Assert controller is kept alive
        AssertAsync.staysTrue(weakController != nil)
    }
```

evsaev (commented):
Well done @adolfogarza 👍 Let's check with @bielikb @alteral that `GIVEN/WHEN/THEN` were supposed to be used not only in UI tests. Once it's resolved, I'll approve 🙂

evsaev (approved):
🚢

PR #1932, opened 2022-04-15
Fixed sending device id when Firebase if used for sending push notification

madlax19:
### 🎯 Goal
- Make functions interface more simpler
- Remove complicated types dependencies for device endpoints
- Fix firebase notification support in iOS Applications

### 🛠 Implementation
- Replace PushProvider struct with PushProvider enum with nested objects for different providers type
- Add extension for provide `push provider id` and `device id` for endpoint as 'ready to use' data
- Change functions interface for use updated PushProvider type
 
### 📝 Changes
- replaced PushProvider struct with enum with nested objects (different types for different providers)
- changed functions interfaces

### 🧪 Testing
- Enable firebase notification support for application in the `Dashboard`
- call `addDevice(pushProvider: .firebase(token: "fcm-token"), completion: { _ in }`, token sent properly, and isn't got error in completion
- Push notifications start working

### ☑️ Contributor Checklist

### 🎉 GIF

_Please provide a suitable gif that describes your work on this pull request_

### DIFF
--- a/Sources/StreamChat/APIClient/Endpoints/DeviceEndpoints.swift
+++ b/Sources/StreamChat/APIClient/Endpoints/DeviceEndpoints.swift
@@ -16,7 +16,7 @@ extension Endpoint {
     static func addDevice(
         userId: UserId,
         deviceId: DeviceId,
-        pushProvider: PushProvider
+        pushProviderId: PushProviderId
     ) -> Endpoint<EmptyResponse> {
         .init(
             path: .devices,
@@ -26,7 +26,7 @@ extension Endpoint {
             body: [
                 "user_id": userId,
                 "id": deviceId,
-                "push_provider": pushProvider.rawValue
+                "push_provider": pushProviderId
             ]
         )
     }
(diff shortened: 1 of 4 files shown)


evsaev:
Hey @madlax19, thanks for spotting the issue and raising the PR! 🙌

Yeah, I can confirm that working with `firebase` push provider is super tricky now because it's not clear how to encode the firebase token as `Data` so it's later decoded as the same string for the endpoint. So we definitely need to improve the developer's experience here.

The thing is, switching to the new enum-based API would be a breaking change and would require us to bump the major version and release `5.0.0`. We have some stuff planned for v5 which is not ready yet so we'll have to find another way how to handle the push provider in a non-breaking way.

I'll communicate that to the team and get back to you with the updates. 

Thank you!

madlax19:
@evsaev Thank you for you answer! You are right, changes in this PR don't look like a minor update.

nuno-vieira:
Hi @madlax19!

We followed a similar approach but without breaking changes in this PR: https://github.com/GetStream/stream-chat-swift/pull/1934

We are very sorry for the issue!

Thank you a lot for your time in opening this PR 🙏

PR #1933, opened 2022-04-15
[CIS-1686] Support channels configuration via mock server

alteral:
### 🔗 Issue Link
- https://stream-io.atlassian.net/browse/CIS-1686

### 🎯 Goal

We are making this change to be able to pre-generate channels prior to any UI test

### 📝 Changes

- Add an ability to generate N channels before any UI test

```swift
server.generateChannels(count: 33)
```

- Add an ability to remove all channels before any UI test

```swift
server.generateChannels(count: 0)
```

- Add an ability to specify an exact participant at any time during a UI test

```swift
participantRobot.user = UserDetails.hanSolo
```

### 🧪 Testing

- To test this PR, try to use these new features above in our existing UI tests with mock server

### ☑️ Contributor Checklist

### 🎉 GIF

han_solo


polqf (commented):
I see a lot of force unwraps. IMO those should be avoided even if we are in a test target. If any of those crash, the execution will not continue, and will not provide proper context on the issues

alteral:
> I see a lot of force unwraps. IMO those should be avoided even if we are in a test target. If any of those crash, the execution will not continue, and will not provide proper context on the issues

Hey @polqf, fair enough. The main idea was to make the mock server a little more readable because there are a lot of places where we need to unwrap some things from JSON templates. What do you think about wrapping such force unwraps with the native XCTest feature `XCTUnwrap`, e.g.:

```swift
try! XCTUnwrap(array.first, "the reason why it can fail")
```

This way we'll keep it as a one-liner. The force unwrap is still in place but the message about failure will be in the test result.

StreamChatUITestsAppUITests/MockServer/ChannelResponses.swift
```diff
@@ -4,25 +4,124 @@
...
+                newChannel[ChannelPayload.CodingKeys.channel.rawValue] = mockChannelDetails(
+                    channel: newChannel,
+                    author: author,
+                    memberCount: members.count
+                )
+                channels.append(newChannel)
+            }
+        }
+        
+        return channels
```
bielikb, line 109:
```suggestion
        guard count > 0 else { return [:]() }

       var channels: [[String: Any]] = []
       var membership = sampleChannel[ChannelPayload.CodingKeys.membership.rawValue] as! [String: Any]
       membership[TopLevelKey.user] = author
            
        for _ in 1...count {
            var newChannel = sampleChannel
            newChannel[ChannelPayload.CodingKeys.members.rawValue] = members
            newChannel[ChannelPayload.CodingKeys.membership.rawValue] = membership
            newChannel[ChannelPayload.CodingKeys.channel.rawValue] = mockChannelDetails(
                channel: newChannel,
                author: author,
                memberCount: members.count
            )
            channels.append(newChannel)
        }
    }
        
    return channels
```

StreamChatUITestsAppUITests/MockServer/MessageResponses.swift
```diff
@@ -10,103 +10,146 @@ extension StreamMockServer {
...
+        mockedMessage[MessagePayloadsCodingKeys.user.rawValue] = user
         return mockedMessage
     }
     
     func mockMessage(
         _ message: [String: Any],
+        channelId: String,
         messageId: String?,
         text: String?,
+        user: [String: Any],
```
bielikb, line 38:
```suggestion
        users: [String: Any],
```

StreamChatUITestsAppUITests/MockServer/MessageResponses.swift
```diff
@@ -10,103 +10,146 @@ extension StreamMockServer {
...
+        let channelType = ChannelType.messaging.rawValue
+        mockedMessage[MessagePayloadsCodingKeys.cid.rawValue] = "\(channelType):\(channelId)"
+        mockedMessage[EventPayload.CodingKeys.channelId.rawValue] = channelId
         mockedMessage[MessagePayloadsCodingKeys.id.rawValue] = messageId
         mockedMessage[MessagePayloadsCodingKeys.createdAt.rawValue] = createdAt
         mockedMessage[MessagePayloadsCodingKeys.updatedAt.rawValue] = updatedAt
         mockedMessage[MessagePayloadsCodingKeys.text.rawValue] = text
         mockedMessage[MessagePayloadsCodingKeys.html.rawValue] = text?.html
-        mockedMessage = intercept?(&mockedMessage) ?? mockedMessage
+        mockedMessage[MessagePayloadsCodingKeys.user.rawValue] = user
```
bielikb, line 51: Let's rename the `MessagePayloadsCodingKeys.user` to `MessagePayloadsCodingKeys.users`

StreamChatUITestsAppUITests/MockServer/ChannelResponses.swift
alteral: Updated 👍

StreamChatUITestsAppUITests/MockServer/MessageResponses.swift
alteral: This is actually about only one user, why do you think it's better to rename this arg?

StreamChatUITestsAppUITests/MockServer/MessageResponses.swift
alteral: @bielikb, same here, I think `user` is grand. Btw, `MessagePayloadsCodingKeys` is a source code. Thoughts?

bielikb: @alteral does it corresponds to exactly 1 `user` or is this an array?

alteral: This is an object in the backend response. It corresponds to exactly 1 `user`. E.g.:
```json
"user": {
      "id": "luke_skywalker",
      "role": "user",
      "created_at": "2020-12-07T11:36:47.059906Z",
      "updated_at": "2022-03-08T15:30:07.960068Z",
      "last_active": "2022-03-08T15:57:33.831192082Z",
      "banned": false,
      "online": true,
      "image": "https://blahblahblah.com/1.jpg",
      "birthland": "Tatooine",
      "name": "Luke Skywalker"
    }
```

bielikb (approved):

adolfogarza (approved):

PR #1934, opened 2022-04-18
[CIS-1776] Fix unusable firebase push provider 

nuno-vieira:
### 🔗 Issue Link
CIS-1776
https://github.com/GetStream/stream-chat-swift/pull/1932

### 🎯 Goal
We recently shipped Multiple Push Provider support, but the Firebase provider requires the token to be `String` and not `Data` which makes the previous solution unusable. Since we still need to implement Firebase Push Notifications in the DemoApp as part of the Push V2 Migration work, we didn't spot this earlier. Unfortunately, this was spotted by a customer here in this PR: https://github.com/GetStream/stream-chat-swift/pull/1932 by @madlax19 🙏

The implementation by @madlax19 included breaking changes so this PR has a similar implementation but without breaking changes. 

### 🛠 Implementation
- Introduces a new `PushDevice` data structure and a new function `CurrentUserController.addDevice(_pushDevice:)`. This new data structure supports a different type of token depending on the push provider.
- Deprecates the `CurrentUserController.addDevice(token:pushProvider:)` since it is not usable with different push providers.

#### How to use the new `PushDevice`
`currentUserController.addDevice(.apn(token: apnDeviceToken))`
`currentUserController.addDevice(.firebase(token: fcmToken))`

### 🎨 UI Changes
N/A

### 🧪 Testing
N/A

### ☑️ Contributor Checklist


Sources/StreamChat/Controllers/CurrentUserController/CurrentUserController.swift
```diff
@@ -71,7 +71,7 @@ public class CurrentChatUserController: DataController, DelegateCallable, DataSt
     private lazy var chatClientUpdater = environment.chatClientUpdaterBuilder(client)
     
     /// The worker used to update the current user.
-    private lazy var currentUserUpdater = environment.currentUserUpdaterBuilder(
+    internal lazy var currentUserUpdater = environment.currentUserUpdaterBuilder(
```
evsaev, line 74: Can we keep it private and use a new API from the deprecated method?

Sources/StreamChat/Models/Device.swift
```diff
@@ -7,30 +7,20 @@ import Foundation
...
-    /// Init a device for Push Notifications.
-    ///
-    /// - Parameters:
-    ///   - id: The push notifications device identifier.
-    ///   - created: The date when the device token was created.
-    init(_ id: DeviceId, createdAt: Date? = .init()) {
-        self.id = id
-        self.createdAt = createdAt
-    }
-}
```
evsaev, line 31: 👏

Sources/StreamChat/Deprecations.swift
```diff
@@ -213,3 +213,27 @@ public typealias NotificationInviteAccepted = NotificationInviteAcceptedEvent
...
 @available(*, deprecated, renamed: "UnknownChannelEvent")
 public typealias UnknownEvent = UnknownChannelEvent
+
+public extension CurrentChatUserController {
+    @available(
+        *,
+        deprecated,
+        message: "use addDevice(_pushDevice:) instead. This deprecated function doesn't correctly support multiple push providers."
+    )
+    func addDevice(token: Data, pushProvider: PushProvider = .apn, completion: ((Error?) -> Void)? = nil) {
```
evsaev, line 223: Shall we also deprecate `PushProvider`?

Sources/StreamChat/Controllers/CurrentUserController/CurrentUserController.swift
nuno-vieira: Yes nice idea 👍

Sources/StreamChat/Deprecations.swift
nuno-vieira: We are still using `PushProvider`, especially because `PushDevice.pushProvider` is public. I think it is still useful, even tho for now it is not exposed. But it is used internally and can be access publicly

.slather.yml
```diff
@@ -12,3 +12,4 @@ ignore:
   - "Sources/StreamChatUI/StreamSwiftyGif"
   - "StreamChatUITestsApp"
   - "StreamChatUITestsAppUITests"
+  - "**/*Deprecations.swift"
```
nuno-vieira, line 15:
```suggestion
  - "**/*/Deprecations.swift"
```

.slather.yml
```diff
@@ -12,3 +12,4 @@ ignore:
   - "Sources/StreamChatUI/StreamSwiftyGif"
   - "StreamChatUITestsApp"
   - "StreamChatUITestsAppUITests"
+  - "**/*/Deprecations.swift"
```
polqf, line 15: This should not be part of this PR, right? Adam is already fixing it in his PR

Sources/StreamChat/Deprecations.swift
```diff
@@ -213,3 +213,14 @@ public typealias NotificationInviteAccepted = NotificationInviteAcceptedEvent
...
```
polqf, line 223: We agreed to keep the deprecations in the same file as they were before

evsaev (approved):
LGTM

Sources/StreamChat/Models/PushDevice.swift
```diff
@@ -0,0 +1,30 @@
...
+    /// - Returns: The Push Device details.
+    public static func apn(token: Data) -> Self {
+        .init(deviceId: token.deviceId, provider: "apn")
+    }
+
+    /// Creates a Push Device for Firebase Push Notifications.
+    /// - Parameter token: The FCM token.
+    /// - Returns: The Push Device details.
+    public static func firebase(token: String) -> Self {
+        .init(deviceId: DeviceId(token), provider: "firebase")
```
polqf, line 28: Why aren't you using the static properties in PushProvider?

.slather.yml
nuno-vieira: This is related to Sonar, not SwiftLint. I'm not aware of Adam working on Sonar stuff 🤔

Sources/StreamChat/Deprecations.swift
nuno-vieira: I thought this was a v5+ approach agreement. If we do it now, the deprecations will be all over the place. I think we should try to make it consistent for now

Sources/StreamChat/Models/PushDevice.swift
nuno-vieira: Nice spot 🤦

nuno-vieira: Done ✅

Sources/StreamChat/Deprecations.swift
polqf: Hmmm, seems like we've been having different approaches on that then 😅 . I saw other PRs where deprecations were kept. We even created a ticket in JIRA to move the ones in this file to its original file (https://stream-io.atlassian.net/browse/CIS-1552)

.slather.yml
polqf: Yeah, discard my comment!

Sources/StreamChat/Deprecations.swift
nuno-vieira: I've moved it 👍

evsaev: Moving deprecation back to files is yak shaving, IMHO. We can start following the new approach when the v5 is released. For now, the consistency matters more

nuno-vieira: I agree with @evsaev, but now I've changed it already 😄 I think we should follow the new approach only after that ticket is complete.

Issue #1805, opened 2022-02-20
ChatChannelListVC ignores taps if you tap quickly after navigating back from ChatChannelView


linhewinner:
1. Go to ChatChannelVC.
2. Return to ChatChannelListVC.
3. Tap on a row quickly ... tap is often ignored. A second tap would work.
Note that I have localStorage enabled.

I'll explain the root cause and my work-around. I hope Stream can put this fix into ChatChannelListVC so others won't have to wrestle with this.

When you return from a ChatChannelVC, I observed that some channels often get updated (not sure why ... maybe read state or watcher count related). This causes ChatChannelListController:didChangeChannels: callback to be invoked here:
https://github.com/GetStream/stream-chat-swift/blob/7fce6491dd95839c45cf8518faff0fb7eacc0ef3/Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift#L300

The changes are usually updates only (i.e. no insert/remove/move), but the code calls performBatchUpdates which triggers a server animation. During the animation duration (250ms or so), the ChannelListView appears static but taps will be ignored due to the ongoing "animation".
From a user's perspective, it feels like the ChatChannelListView is just ignoring taps for no reason.

Here is my workaround that I hope Stream team can put into the default implementation of ChatChannelListVC

```swift
    override func controller(
        _: ChatChannelListController,
        didChangeChannels changes: [ListChange<ChatChannel>]
    ) {
        let updates = changes.filter { change in
            if case .update = change {
                return true
            } else {
                return false
            }
        }

        // Changes can be insert, remove, move or update. The default
        // implementation animates these changes in the thread list.
        // Whenever you foreground the app or navigate back from a
        // thread, there are almost always some updates: e.g. read state
        // Now those updates are "animated" for 250ms or so, during which
        // time taps are ignored. This obviously creates a bug.
        //
        // Now how about we just disable the animation altogether? That
        // is not nice if threads are added, removed or moved around.
        // We still want animations in those case
        //
        // Hence here ... we disable animation if there are ONLY updates;
        // but we continue to allow animation for inserts/removes/moves.
        if updates.count == changes.count {
            UIView.performWithoutAnimation {
                super.controller(controller, didChangeChannels: changes)
            }
        } else {
            super.controller(controller, didChangeChannels: changes)
        }
    }
}
```

## GetStream Environment
**GetStream Chat version:** 4.10.0
**GetStream Chat frameworks:** StreamChat, StreamChatUI
[...]

PR #1935, opened 2022-04-18
[CIS-1721] Delivery status on channel list

evsaev:
### 🔗 Issue Link

CIS-1721

⚠️ The PR is targeting the feature branch for delivery status ⚠️

### 🎯 Goal

Show delivery status for channel preview message if it's sent by the current user

### 🛠 Implementation

- The new `preview` relation is added to `ChannelDTO`. This has to be a relation (not a computed field), otherwise it wouldn't be possible to trigger the channel update and reload channel cell when preview message is updated (e.g. edited). 
- Preview message can't be taken from latest messages because a deleted message can't serve as a channel preview. So we could iterate latest message to find the first non-deleted, but there's no guarantee it'll be one + this would happen on every channel update on UI thread and negatively affect the performance.

### 📝 Changes

- Introduce `ChatMessageDeliveryStatusCheckmarkView` component, make use of it on channel/message list
- Update `ChatChannelListItemView` to display preview message
- Expose `previewMessage` on `ChatChannel` model

### 🎨 UI Changes

https://user-images.githubusercontent.com/12818985/164048429-30403186-bd87-481e-95fa-3f1d3c92ff10.mov

### 🧪 Testing

**1. The channel preview is updated when a new message is sent to the channel** 
**GIVEN** I’m on the channel list
**WHEN** another member sends the message to the channel
**THEN** channel preview is updated to show the new message
**AND** the timestamp label shows the timestamp of the new preview message

**2. The channel preview is updated when thread reply is also sent to main channel**
**GIVEN** I’m on the channel list
**AND** channel preview shows the last non-deleted message in the channel
**WHEN** channel member sends the new thread reply **`show_in_channel = true`**
**THEN** channel preview is updated to show the new thread reply visible in the main channel

**3. The channel preview is not updated when thread reply is sent**
**GIVEN** I’m on the channel list
**AND** channel preview shows the last non-deleted message in the channel
**WHEN** channel member sends the new thread reply **`show_in_channel = false`**
**THEN** channel preview keeps showing the last non-deleted message in the channel 

**4. The channel preview is edited when the channel preview message is edited**
**GIVEN** I’m on the channel list
**AND** channel preview shows the last non-deleted message in the channel
**WHEN** the preview message is edited
**THEN** channel preview is updated and shows the edited message

**5. The channel preview is updated when the preview message is deleted**
**GIVEN** I’m on the channel list
**AND** channel preview shows the last non-deleted message in the channel
**WHEN** the preview message is deleted
**THEN** channel preview is updated to show last non-deleted message in the channel
**AND** the timestamp label shows the timestamp of new preview message

**6. The channel preview is updated when channel is truncated w/ system message**
**GIVEN** I’m on the channel list
**AND** channel preview shows the last non-deleted message in the channel
**WHEN** the channel is truncated w/ a system message 
**THEN** channel preview is updated to show this system message
**AND** the timestamp label shows the truncation date

**7. The channel preview is updated when channel is truncated w/o system message**
**GIVEN** I’m on the channel list
**AND** channel preview shows the last non-deleted message in the channel
**WHEN** the channel is truncated w/o a system message 
**THEN** channel preview shows **No messages** text
**AND** the timestamp label is hidden

**8. The channel preview shows No messages when the channel is empty**
**GIVEN** I’m on the channel list
**WHEN** the channel has no messages
**THEN** channel preview shows **No messages** text
**AND** the timestamp label is hidden

**9. Delivery status shows clocks when the last message is from the current user and it’s in the pending send state**
**GIVEN** I open the channel (no matter `read_events` are enabled/disabled)
**AND** I send the new message
**WHEN** I return to the channel list before the message is sent
**THEN** last message delivery status in the channel preview shows clocks on the left

**10. Delivery status shows** e**rror when sending fails for the last message from the current user**
**GIVEN** I open the channel (no matter `read_events` are enabled/disabled)
**AND** I send the new message
**AND** I return to the channel list
**AND** last message delivery status in the channel preview shows clocks on the left
**WHEN** message sending fails because of any reason
**THEN** last message delivery status in the channel preview shows an error indicator on the left

**11. Delivery status shows a single checkmark when the last message is from the current user and it is sent**
**GIVEN** I open the channel w/ `read_events` enabled
**AND** I send the new message
**AND** message is successfully sent
**WHEN** I return to the channel list
**THEN** last message delivery status in the channel preview shows a single checkmark on the right

**12. Delivery status shows double checkmark when the last message is from the current user and it is seen by a channel member**
**GIVEN** I open the group/DM channel w/ `read_events` enabled
**AND** I send the new message
**AND** message is successfully sent
**AND** I return to the channel list
**WHEN** message is read by channel members
**THEN** last message delivery status changes from a single to a double checkmark
**AND** read counter is always hidden

**13. Delivery status is hidden for own channel preview messages when reads are disabled**
**GIVEN** I open the channel w/ `read_events` disabled
**AND** I send the new message
**AND** message is successfully sent (or even read)
**WHEN** I return to the channel list
**THEN** last message delivery status is hidden

**14. Delivery status is hidden if the preview message is from another user**
**GIVEN** I’m on the channel list
**WHEN** the channel preview message is from another user
**THEN** last message delivery status is hidden

### ☑️ Contributor Checklist


Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageDeliveryStatusCheckmarkView.swift
```diff
@@ -0,0 +1,78 @@
...
+            case .sent:
+                return appearance.images.messageDeliveryStatusSent
+            case .read:
+                return appearance.images.messageDeliveryStatusRead
+            case .failed:
+                return appearance.images.messageDeliveryStatusFailed
+            default:
+                return nil
+            }
+        }
```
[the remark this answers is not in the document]
evsaev: `MessageDeliveryStatus` type is declared in LLC, appearance with images is in UI SDK. 

I was thinking about private extension for `MessageDeliveryStatus` with `tintColor` but that would require a `tintColor` to be a func so that we can pass `appearance` so it turned out to be unnecessary complex thing.

Sources/StreamChatUI/Components.swift
```diff
@@ -308,10 +312,6 @@ public struct Components {
     /// The view that shows a number of unread messages in channel.
     public var channelUnreadCountView: ChatChannelUnreadCountView.Type = ChatChannelUnreadCountView.self
 
-    /// The view that shows a read/unread status of the last message in channel.
-    public var channelReadStatusView: ChatChannelReadStatusCheckmarkView.Type =
```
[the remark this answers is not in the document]
evsaev: `ChatChannelReadStatusCheckmarkView` type was just declared in UI SDK but not used anywhere. Chances there's a customer who's found that type and made use of it in their integration are really low so I thought it's fine to remove it.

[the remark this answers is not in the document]
bielikb: @evsaev let's mention that under `Removed` section in our changelog along with hint.

bielikb (approved):

Sources/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -42,7 +42,8 @@ class MessageDTO: NSManagedObject {
     @NSManaged var quotedMessage: MessageDTO?
     @NSManaged var quotedBy: Set<MessageDTO>
     @NSManaged var searches: Set<MessageSearchQueryDTO>
-    
+    @NSManaged var previewIn: ChannelDTO?
```
nuno-vieira, line 45: Maybe a little bit easier to read?
```suggestion
    @NSManaged var previewFromChannel: ChannelDTO?
```

or `previewOfChannel`

Sources/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -68,6 +69,11 @@ class MessageDTO: NSManagedObject {
     override func willSave() {
         super.willSave()
         
+        if let channel = previewIn, !channel.hasChanges, isUpdated {
+            let cid = channel.cid
+            channel.cid = cid
+        }
```
nuno-vieira, line 75: What is this actually doing? 🤔 Can we provide a comment?

Sources/StreamChat/Models/Channel.swift
```diff
@@ -144,7 +144,14 @@ public struct ChatChannel {
...
     /// Additional data associated with the channel.
     public let extraData: [String: RawJSON]
-
+    
+    /// The channel message is supposed to be shown in channel preview.
+    ///
+    /// - Important: The `previewMessage` can differ from `latestMessages.first` (or even not be included into `latestMessages`)
+    /// because the preview message is the last `non-deleted` message sent to the channel.
+    public var previewMessage: ChatMessage? { _previewMessage }
+    @CoreDataLazy private var _previewMessage: ChatMessage?
```
nuno-vieira, line 153: I'm wondering if we could avoid this 🤔 One of the main issues in the channel list right now is how we update the message in the channel list item. This sometimes blocks the main thread because of the way we fetch the `latestMessages.first` from the main thread and we use `performAndWait` in CoreData. Since we are moving from `latestMessages` to `previewMessage` it could be an opportunity to improve this 🤔 But not sure if this is possible right now without breaking changes

Sources/StreamChat/Models/ChatMessage.swift
```diff
@@ -406,3 +430,24 @@ public enum LocalReactionState: String {
...
+/// The type describing message delivery status.
+public struct MessageDeliveryStatus: RawRepresentable, Hashable {
+    public let rawValue: String
+    
+    public init(rawValue: String) {
+        self.rawValue = rawValue
+    }
+    
+    /// The message delivery state for message that is being sent/edited/deleted.
+    public static let pending = Self(rawValue: "pending")
```
nuno-vieira, line 443: We can write these a little be shorter:
```suggestion
    public static let pending: Self = "pending"
```

Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
```diff
@@ -102,30 +102,38 @@ open class ChatChannelListItemView: _View, ThemeProvider, SwiftUIRepresentable {
         guard let content = content else { return nil }
         if let typingUsersInfo = typingUserString {
             return typingUsersInfo
-        } else if let latestMessage = content.channel.latestMessages.first {
-            return "\(latestMessage.author.name ?? latestMessage.author.id): \(latestMessage.textContent ?? latestMessage.text)"
+        } else if let previewMessage = content.channel.previewMessage {
```
nuno-vieira, line 105: Probably this is already a performance improvement since instead of fetching all the latest messages, we now directly fetch the preview message. 👍 Would be nice if this issue: CIS-1677 (https://github.com/GetStream/stream-chat-swift/issues/1805) is still reproducible

Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
```diff
@@ -188,6 +198,27 @@ open class ChatChannelListItemView: _View, ThemeProvider, SwiftUIRepresentable {
...
+        let deliveryStatus: MessageDeliveryStatus? = content.flatMap {
+            guard let previewMessage = $0.channel.previewMessage else { return nil }
+            
+            let show = components.messageLayoutOptionsResolver.canShowDeliveryStatus(
+                for: previewMessage,
+                in: $0.channel,
+                allowWhenFailed: true
+            )
+            
+            return show ? previewMessage.deliveryStatus : nil
```
nuno-vieira, line 211:
```suggestion
            let canShow = components.messageLayoutOptionsResolver.canShowDeliveryStatus(
                for: previewMessage,
                in: $0.channel,
                allowWhenFailed: true
            )
            
            return canShow ? previewMessage.deliveryStatus : nil
```

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageDeliveryStatusCheckmarkView.swift
```diff
@@ -0,0 +1,78 @@
...
+
+import StreamChat
+import UIKit
+
+/// A view that displays message delivery receipts.
+open class ChatMessageDeliveryStatusCheckmarkView: _View, ThemeProvider {
+    /// The data this view component shows.
+    open var content: MessageDeliveryStatus? {
+        didSet { updateContentIfNeeded() }
+    }
```
nuno-vieira, line 13: It is better to create a `Content` struct for the views, even if it only has 1 property. This will make sure we can add new data in the future

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageLayoutOptionsResolver.swift
```diff
@@ -178,26 +178,22 @@ open class ChatMessageLayoutOptionsResolver {
...
-        
-        guard message.type != .deleted, message.type != .ephemeral else {
-            // Delivery status should not be shown for deleted & ephemeral messages.
-            return false
-        }
+    func canShowDeliveryStatus(
+        for message: ChatMessage,
+        in channel: ChatChannel,
+        allowWhenFailed: Bool
+    ) -> Bool {
```
nuno-vieira, line 185: Shouldn't this be open?
```suggestion
    open func canShowDeliveryStatus(
        for message: ChatMessage,
        in channel: ChatChannel,
        allowWhenFailed: Bool
    ) -> Bool {
```

if we have it open, we don't need the `allowWhenFailed` flag. If a customer wants a different behaviour, he can do it by overriding this function and changing the failed state behaviour

nuno-vieira (commented):
Looks really good overall 👌 Just added some questions and suggestions 🙂

Sources/StreamChat/Database/DTOs/MessageDTO.swift
evsaev: When the preview message is updated (e.g. edited) the channel cell should be reloaded to reflect the latest preview state. But the update of the source entity (message) does not trigger an update on the destination entity (channel) in 1-1 relation so we manually mark the channel as `dirty`. I'll leave a comment 👍

Sources/StreamChat/Models/ChatMessage.swift
evsaev: Yeah, I thought about this. `MessageDeliveryStatus` is not representable by string literal and implementing this protocol would require adding another `init`. I think saving a couple of chars should not be the reason for adding more complexity to sources

nuno-vieira: AH never mind, I thought this was StringRepresentable, never mind then! :)

Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
evsaev: I'll run the load test and check if I it's better now 👍

Sources/StreamChat/Models/Channel.swift
evsaev: Do you want it to be just `public var previewMessage: ChatMessage?` i.e. without core data lazy?

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageDeliveryStatusCheckmarkView.swift
evsaev: It makes sense if init is kept internal. Let me add it 👍

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageLayoutOptionsResolver.swift
evsaev: It's not for customers, it's for us, actually. On the channel list, we can show an error as a delivery status (`allowWhenFailed  == true`) but not on the message list (`allowWhenFailed  == false`). But yeah, let's open it 👍

Sources/StreamChat/Database/DTOs/MessageDTO.swift
evsaev: Renamed ✅

Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
evsaev: Done ✅

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageDeliveryStatusCheckmarkView.swift
evsaev: Done ✅

Sources/StreamChat/Models/Channel.swift
nuno-vieira: Yes, although I'm not sure about the impact of this 🤔

Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
```diff
@@ -188,6 +198,30 @@ open class ChatChannelListItemView: _View, ThemeProvider, SwiftUIRepresentable {
...
+            guard
+                let previewMessage = $0.channel.previewMessage,
+                let deliveryStatus = previewMessage.deliveryStatus
+            else { return nil }
+            
+            let canShow = components.messageLayoutOptionsResolver.canShowDeliveryStatus(
+                for: previewMessage,
+                in: $0.channel,
+                allowWhenFailed: true
+            )
```
nuno-vieira, line 212: IMO the resolver shouldn't be used in the `ChatChannelListItemView`. I would prefer to separate it into 2 functions and remove the `allowWhenFailed`. The reason is that the `allowWhenFailed` looks like a code smell. In reality, the logic for showing the delivery status is different in both places (Although they diverge in just one detail in this case). 

There are multiple ways we can fix this:
- Keep the `canShowDeliveryStatus()` in the Resolver and create a new one in `ChatChannelListItemView`.
- Create a protocol `DeliveryStatusShowing` and create a `MessagePreviewDeliveryStatus` and a `MessageDeliveryStatus`

The second approach might be an overkill for this case, but what I mean is that I think we need to distinguish 2 delivery status logics, one from a regular message, and one from a preview message. I believe that adding a flag in the argument of that function it won't scale very well 🤔

evsaev: Makes sense 👍

Sources/StreamChat/Models/Channel.swift
evsaev: There'll be no impact. In both cases, DB request happens on main thread, the only different is **when** it happens. Without core data lazy, model evaluation will happen before the change is propagated to the UI. With lazy stuff, it's gonna happen when we're calculating `subtitle` for the channel cell.

nuno-vieira: So probably is better that we keep the CoreDataLazy for now? Otherwise, the impact will move to the loading of the channels right?

Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
evsaev: Done ✅

Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
```diff
@@ -102,30 +102,57 @@ open class ChatChannelListItemView: _View, ThemeProvider, SwiftUIRepresentable {
...
-            return timestampFormatter.format(lastMessageAt)
+        if let timestamp = content?.channel.previewMessage?.createdAt {
+            return timestampFormatter.format(timestamp)
         } else {
             return nil
         }
     }
+    
+    /// The delivery status to be shown for the channel's preview message.
+    open var previewMessageDeliveryStatus: ChatMessageDeliveryStatusCheckmarkView.Content? {
```
nuno-vieira, line 132: Doesn't make sense to couple this view with `ChatMessageDeliveryStatusCheckmarkView`. Also, the name of this property doesn't reflect what actually returns. I think is better to return just the `MessageDeliveryStatus`
```suggestion
    open var previewMessageDeliveryStatus: MessageDeliveryStatus? {
```

Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
```diff
@@ -188,6 +217,17 @@ open class ChatChannelListItemView: _View, ThemeProvider, SwiftUIRepresentable {
...
+        let checkmarkContent = previewMessageDeliveryStatus
+        previewMessageDeliveryStatusView.content = checkmarkContent
+        previewMessageDeliveryStatusView.isHidden = checkmarkContent == nil
+        
+        if let status = checkmarkContent?.deliveryStatus {
+            bottomContainer.insertArrangedSubview(
+                previewMessageDeliveryStatusView,
+                at: status == .pending || status == .failed ? 0 : 1
+            )
+        }
```
nuno-vieira, line 230: Then, this makes it more readable as well:
```suggestion
        previewMessageDeliveryStatusView.content = previewMessageDeliveryStatus 
        previewMessageDeliveryStatusView.isHidden = previewMessageDeliveryStatus == nil
        
        if let status = previewMessageDeliveryStatus {
            bottomContainer.insertArrangedSubview(
                previewMessageDeliveryStatusView,
                at: status == .pending || status == .failed ? 0 : 1
            )
        }
```

evsaev: This way `previewMessageDeliveryStatus` is calculated 3 times

Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
evsaev: The issue is not fully resolved because we still do a lot on the main thread but it's much better 👍Previously, the weight of `subtitle` calculation was 3.8% from everything we do on main thread, with `previewMessage` it's `0.1` 🎉

evsaev: The weight of `updateContent` has increased because we're now showing/hiding/moving around the delivery checkmark but that's unavoidable when adding the new subview

Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
nuno-vieira: Creating `ChatMessageDeliveryStatusCheckmarkView.Content` is just a initializer. One thing is to create this data structure, another is the logic when we should create it. We shouldn't couple both. Right now `previewMessageDeliveryStatus` is responsible for 2 things, deciding whether it should display the delivery status, and creating the content of the `ChatMessageDeliveryStatusCheckmarkView`. IMO it should only decide whether it should display the delivery status.

I'm not saying to calculate the logic 3 times. This what I mean:

```swift
    /// The delivery status to be shown for the channel's preview message.
    open var previewMessageDeliveryStatus: MessageDeliveryStatus? {
        guard
            let content = content,
            let deliveryStatus = content.channel.previewMessage?.deliveryStatus
        else { return nil }
        
        switch deliveryStatus {
        case .pending, .failed:
            return deliveryStatus
        case .sent, .read:
            guard content.channel.config.readEventsEnabled else { return nil }
            
            return deliveryStatus
        default:
            return nil
        }
    }
```

Then the creation of the content is in the `updateContent()` function, there's no need to create the content in the property above. If a customer wants to reuse this logic but doesn't want to use `ChatMessageDeliveryStatusCheckmarkView.Content`, it won't be possible with the current solution.

Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
nuno-vieira: Please, read my comment above 🙂

Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
evsaev: Done

Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
```diff
@@ -188,6 +217,19 @@ open class ChatChannelListItemView: _View, ThemeProvider, SwiftUIRepresentable {
...
+        previewMessageDeliveryStatusView.content = checkmarkContent
+        previewMessageDeliveryStatusView.isHidden = checkmarkContent == nil
+        
+        if let status = checkmarkContent?.deliveryStatus {
+            bottomContainer.insertArrangedSubview(
+                previewMessageDeliveryStatusView,
+                at: status == .pending || status == .failed ? 0 : 1
+            )
+        }
```
nuno-vieira, line 232: Nit:
```suggestion
        if let status = previewMessageDeliveryStatus {
            bottomContainer.insertArrangedSubview(
                previewMessageDeliveryStatusView,
                at: status == .pending || status == .failed ? 0 : 1
            )
        }
```

nuno-vieira (approved):
LGTM! 🚀 Only added a small nit, not blocking for sure

Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
evsaev: `previewMessageDeliveryStatus` is a computed property so accessing it one more time would run the calculation again.

nuno-vieira: It's not like we are doing an intensive task here 😛 I would prefer readability over premature optimization, the calculation here is 0.0000001 sec

evsaev: Readability is fine in both cases, I think. Running the code more than it's needed is a bad practice and can bite us back if the computation becomes heavier (e.g. `readByCount` will be changed to always make a count request to DB). I'd prefer to keep the variable in place 👍

PR #1937, opened 2022-04-19
Core Data Performance take 1

tbarbugli:
Writing a list of channels is 13x faster when everything needs to be written to the db and 50x faster when the same payload is sync'd to the database. This is done using a very large payload coming from our own Watercooler app (lots of duplicated members).

Notable Changes:

- in-mem cache for users, channels and members, cache is specific to a thread and does is flushed when the writable context is closed
- batch upsert for members, messages and read DTOs
- 

Minor Changes:

- Capture `own_capabilities` and `disabled`  channel fields, this avoids useless writes to the db and avoids polluting extra data
- Avoid fetching messages when ChannelDTO is already populated with messages (fetchMessages)
- Denormalize `attachmentCount` to avoid queries to fetch attachments when there are no attachments


Sources/StreamChat/Database/DTOs/ChannelDTO.swift
```diff
@@ -314,6 +330,24 @@ extension ChannelDTO {
     }
 }
 
+extension ChannelDTO {
+    func messagesArePopulated() -> Bool {
```
polqf, line 334:
```suggestion
    fileprivate var areMessagesPopulated: Bool {
```

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
```diff
@@ -314,6 +330,24 @@ extension ChannelDTO {
...
+        if lastMessageAt == nil {
+            return true
+        }
+
+        if oldestMessageAt == nil {
+            return true
+        }
```
polqf, line 341:
```suggestion
        if lastMessageAt == nil || oldestMessageAt == nil {
            return true
        }
```

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
```diff
@@ -314,6 +330,24 @@ extension ChannelDTO {
...
+
+        let newestMessage = messages.max { a, b in
```
polqf, line 343: Can't we assume the last is the newest message?

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
```diff
@@ -371,21 +405,29 @@ extension ChatChannel {
         }
         
         let fetchMessages: () -> [ChatMessage] = {
-            MessageDTO
+            let limit = dto.managedObjectContext?.localCachingSettings?.chatChannel.latestMessagesLimit ?? 25
```
polqf, line 408: We should not be accessing this properties from here, but instead pass this as parameter

Sources/StreamChat/Database/DTOs/ChannelReadDTO.swift
```diff
@@ -63,19 +69,78 @@ class ChannelReadDTO: NSManagedObject {
 // MARK: Saving and loading the data
 
 extension NSManagedObjectContext {
+    func upsertMany(payload: [ChannelReadPayload], for cid: ChannelId) throws -> [ChannelReadDTO] {
```
polqf, line 72: would be nice to have a generic upsert function that is called from these specific implementations

Sources/StreamChat/Database/DTOs/MemberModelDTO.swift
```diff
@@ -91,17 +110,60 @@ extension MemberDTO {
...
+        payload.forEach {
+            payloadById[$0.user.id] = $0
+        }
+
+        var membersByUserID = [String: MemberDTO]()
+
+        // fetch all members based on their IDs
+        let existingMembers = try fetch(MemberDTO.members(userIDs, in: channelId))
+        existingMembers.forEach {
+            // tell core data that we do not want to merge any changes for these objects (sigh)
```
polqf, line 127: Not clear to me why this is done in every upsert. Could you expand?

Sources/StreamChat/Database/DataStore.swift
```diff
@@ -4,6 +4,89 @@
...
+    func get(memberID: String) -> MemberDTO?
+    func set(member: MemberDTO, channelId: ChannelId)
+
+    func get(messageID: String) -> MessageDTO?
+    func set(message: MessageDTO)
+
+    func flush()
+}
+
+class InMemCache: EntityCache {
```
polqf, line 23: The idea looks promising. We have to be careful about the API, doesn't seem scalable atm

polqf:
I can see crashes that seem to be happening when calling asModel() as part of a ListDatabaseObserver.|

Maybe the cache should be flushed?

Sources/StreamChat/Database/DataStore.swift
tbarbugli: cache lives for 1 db context write, we dont grow past that

tbarbugli:
> I can see crashes that seem to be happening when calling asModel() as part of a ListDatabaseObserver.|
> 
> Maybe the cache should be flushed?

it already happens, can you share more info on crashes and how to repro?

Sources/StreamChat/Database/DTOs/MemberModelDTO.swift
tbarbugli: if the object is not refreshed, core data will write it as it is after the read. the same thing happens if I read a subset of fields. Alternative approach would be to populate all objects read from core data, but that's a bit annoying and expensive

Sources/StreamChat/Database/DTOs/ChannelReadDTO.swift
tbarbugli: I agree, once we are happy with the approach we can do that. Not going to do that until we are 100% sold on this approach (making this generic is not trivial)

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
tbarbugli: agreed, this is just keeping the same behavior to avoid adding more refactoring work in scope (there's already plenty of work in this PR :))

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
tbarbugli: messages is a `Set`

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
```diff
@@ -154,16 +167,20 @@ extension NSManagedObjectContext {
...
+            encodedData = try JSONEncoder.default.encode(payload.extraData)
         } catch {
             log.error(
                 "Failed to decode extra payload for Channel with cid: <\(dto.cid)>, using default value instead. "
                     + "Error: \(error)"
             )
-            dto.extraData = Data()
+            encodedData = Data()
         }
-        dto.extraData = try JSONEncoder.default.encode(payload.extraData)
```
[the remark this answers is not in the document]
tbarbugli: yup, though this is the tip of the iceberg. we currently decode from data to payload type, then encode to core data and again decode to model

luckily custom data is tiny + RawJSON is very fast to enc/dec

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
```diff
@@ -114,6 +123,10 @@ class ChannelDTO: NSManagedObject {
         
         let new = NSEntityDescription.insertNewObject(forEntityName: Self.entityName, into: context) as! ChannelDTO
         new.cid = cid.rawValue
+        
+        defer {
```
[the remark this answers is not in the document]
tbarbugli: harder to make mistakes in the future, there is no functional advantage

Sources/StreamChat/APIClient/Endpoints/Payloads/ChannelListPayload.swift
```diff
@@ -159,6 +159,11 @@ struct ChannelDetailPayload: Decodable {
         
         if var payload = try? [String: RawJSON](from: decoder) {
             payload.removeValues(forKeys: ChannelCodingKeys.allCases.map(\.rawValue))
+            // The `own_capabilities` field is not decoded, this line is here just to  be sure that we dont wrongly add it to extra data
+            payload.removeValue(forKey: "own_capabilities")
+            
+            // The `disabled` field is not decoded, this line is here just to  be sure that we dont wrongly add it to extra data
+            payload.removeValue(forKey: "disabled")
```
b-onc, line 166: What's wrong with keeping unparsed fields in extraData? IMHO it makes the SDK more flexible (future-proof for additions from backend since they'll always end up in extraData)

polqf:
I guess we can close this now

PR #1943, opened 2022-04-20
Fix crashes when sending a message

b-onc:
As discussed in slack and around

Reproduce:
1. Open a channel
2. Send a message

Interestingly, `computeValue` closure was not called when a `self` reference was kept in it.


bielikb (approved):

evsaev (approved):

nuno-vieira (dismissed):
Shouldn't this be merged to 4.14.0?

Issue #1796, opened 2022-02-15
Custom ordering of MessageReactionType


tomSidebar:
## What are you trying to achieve?
Customizing the order message reactions.

## If possible, how can you achieve this currently?
Currently not possible, reactions get sorted and displayed alphabetically (following https://getstream.io/chat/docs/sdk/ios/uikit/views/reactions/#custom-reactions-and-images).

## What would be the better way?
Retain the order in which the custom reactions are being added OR add a comparator function for MessageReactionType.

## GetStream Environment
**GetStream Chat version: 4.10.0**
**GetStream Chat frameworks: StreamChat, StreamChatUI**
**iOS version: 14.8**
**Swift version: 5.5.2**
**Xcode version: 13.2.1**
**Device: iPhone 11**

linhewinner:
Note that it's not just the reactions attached to a message (which can be accomplished by sorting by createdAt)

We are also talking about the Reaction picker itself.
That list is also sorted alphabetically.

polqf:
Hello! We just released 4.12.0, you can check the changelog [here](https://github.com/GetStream/stream-chat-swift/releases/tag/4.12.0).
This issue should be fixed now. Feel free to reopen it if needed

nuno-vieira:
Hi @linhewinner!

This is now merged into develop. We will let you know when this issue is released 🙂 
The ETA for the next release is in 2 weeks.

Best,
Nuno

evsaev:
Hi @linhewinner,

I'm happy to say that [4.15.0](https://github.com/GetStream/stream-chat-swift/releases/tag/4.15.0) containing the requested change is released 🎉 
Here is the [guide](https://getstream.io/chat/docs/sdk/ios/uikit/views/reactions/#custom-reactions-sorting) explaining how to provide custom sorting for reactions.
Thanks and enjoy your day!

Best,
Pavel

PR #1944, opened 2022-04-20
[CIS-1589] Add support for custom reactions sorting

nuno-vieira:
### 🔗 Issue Link
CIS-1589
https://github.com/GetStream/stream-chat-swift/issues/1796

### 🎯 Goal
Provide a simple way to change the order of the reactions.

### 🛠 Implementation
- Added a new `Components.reactionsSorting` function to change the order of the reactions.
- Renamed `ChatMessageReactionsVC` -> `ChatMessageReactionsPickerVC`
- Remove dead code `ChatMessageReactionsView.reactions`

### 🎨 UI Changes
N/A

### 🧪 Testing
The DemoApp has now a custom sorting. This should be the new order of reactions:
"love", "haha", "like", "sad", "wow"

### ☑️ Contributor Checklist


DemoApp/DemoAppCoordinator.swift
```diff
@@ -371,6 +372,34 @@ class HiddenChannelListVC: ChatChannelListVC {
     }
 }
 
+class CustomMessageContentView: ChatMessageContentView {
```
evsaev, line 375:
```suggestion
final class CustomMessageContentView: ChatMessageContentView {
```

DemoApp/DemoAppCoordinator.swift
```diff
@@ -441,3 +470,16 @@ class CustomChatMessageActionsVC: ChatMessageActionsVC {
...
+        switch rawValue {
+        case "love": return 0
+        case "haha": return 1
+        case "like": return 2
+        case "sad": return 3
+        case "wow": return 4
+        default: return 5
+        }
+    }
+}
```
evsaev, line 485: Having `MessageReactionType: Comparable` can make it clear what the `position` means and would allow hiding the implementation details of sorting.

```suggestion
extension MessageReactionType: Comparable {
    public static func < (lhs: MessageReactionType, rhs: MessageReactionType) -> Bool {
        lhs.position < rhs.position
    }
    
    private var position: Int {
        switch rawValue {
        case "love": return 0
        case "haha": return 1
        case "like": return 2
        case "sad": return 3
        case "wow": return 4
        default: return 5
        }
    }
}
```

evsaev (commented):
Looks good 👍 Let's add a CHANGELOG and I'll approve

DemoApp/DemoAppCoordinator.swift
nuno-vieira: Makes sense 👍

nuno-vieira: @evsaev This is actually not possible since we need to pass the closure sorting to the `sorted()` function. This would work if we just called `sorted()` without any argument. But we can't do this. We would need to provide a default `Comparable` in our SDK, and then this wouldn't be overridable in the customer app. So the code will remain as is.

evsaev (approved):

PR #1951, opened 2022-04-25
[BW] Crash related to event batching

evsaev:


Sources/StreamChat/Workers/ChatClientUpdater.swift
```diff
@@ -179,31 +186,41 @@ class ChatClientUpdater {
...
+        completion: @escaping () -> Void
+    ) {
         client.apiClient.flushRequestsQueue()
         client.syncRepository.cancelRecoveryFlow()
         
         // Disconnecting is not possible in connectionless mode (duh)
         guard client.config.isClientInActiveMode else {
             log.error(ClientError.ClientIsNotInActiveMode().localizedDescription)
-            return
+            return completion()
```
polqf, line 199: I've always thought this is not readable. What about:
```
completion()
return
```

?

Sources/StreamChat/Utils/EventBatcher.swift
```diff
@@ -59,19 +60,19 @@ final class Batcher<Item> {
             self.batchProcessingTimer = self.timerType.schedule(
                 timeInterval: self.period,
                 queue: self.queue,
-                onFire: self.process
+                onFire: { self.process {} }
```
polqf, line 63: What about making process's block optional?

PR #1973, opened 2022-05-03
[CIS-1688] CoreData performance improvements

b-onc:
This is a WIP PR. The approach is explained in the Notion doc: 
https://www.notion.so/stream-wiki/Core-Data-Performance-03b4d8691c944a7dbc49dcb008e891c1


Sources/StreamChat/Utils/Database/NSManagedObject+Extensions.swift
```diff
@@ -10,13 +10,97 @@ extension NSManagedObject {
...
+        cache[request as! NSFetchRequest<NSFetchRequestResult>]
+    }
+    
+    func clear() {
+        cache.removeAll()
+    }
+}
+
+extension NSManagedObjectContext {
+    func fetch<T>(_ request: NSFetchRequest<T>, using cache: FetchCache) throws -> [T] where T: NSFetchRequestResult {
```
polqf, line 88: This only takes into consideration using the same request. But it can also happen that objects created with a request are fetched individually in another place.

Also, the cache needs to be properly protected against deletions (not only a complete clear), which is not at the moment. (eg. I fetch the list of channels, then an event comes saying that a model has been deleted -> from that point onwards the cache is keeping a reference to an object that is deleted)
Same goes for updates (not 100% sure about this one)

Sources/StreamChat/Utils/Database/NSManagedObject+Extensions.swift
```diff
@@ -10,13 +10,97 @@ extension NSManagedObject {
...
+protocol NSFetchRequestGettable {}
+
+extension NSFetchRequestGettable where Self: NSManagedObject {
+    static func fetchRequest(id: String) -> NSFetchRequest<Self> {
+        let request = NSFetchRequest<Self>(entityName: entityName)
+        request.predicate = NSPredicate(format: "id == %@", id)
+        return request
+    }
+    
+    static func fetchRequest(keyPath: String, equalTo key: String) -> NSFetchRequest<Self> {
```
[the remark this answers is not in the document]
b-onc: > Can't we pass a real KeyPath here? A string is not scalable

We cannot pass a KeyPath to NSFetchRequest but we can make it a bit safer by using `#keyPath`, I'll update the code

> equalTo value: String, right?

right, changing

Sources/StreamChat/Utils/Database/NSManagedObject+Extensions.swift
b-onc: >This only takes into consideration using the same request. But it can also happen that objects created with a request are fetched individually in another place.

Caching objects based on their "id"s is tricky, certainly possible but I don't think it's something we need to investigate at this point

> (eg. I fetch the list of channels, then an event comes saying that a model has been deleted -> from that point onwards the cache is keeping a reference to an object that is deleted)

since cache is cleared between DB writes, it's not possible to fetch a channel that's been deleted from the cache. we fetch a list of channels, cache is cleared, `channel.delete` events comes, channel is cached, mutated, DB flush, cache cleared. if we fetch the list again, the channel is not in cache anymore, and if it's deleted from DB it won't be fetched to the cache

Sources/StreamChat/Utils/Database/NSManagedObject+Extensions.swift
```diff
@@ -10,13 +10,97 @@ extension NSManagedObject {
...
+    }
+    
+    func get<T>(_ request: NSFetchRequest<T>) -> [NSManagedObjectID]? where T: NSFetchRequestResult {
+        cache[request as! NSFetchRequest<NSFetchRequestResult>]
+    }
+    
+    func clear() {
+        cache.removeAll()
+    }
+}
```
nuno-vieira, line 85: Currently, `FetchCache` is not thread-safe, but we can replace the `cache: Dictionary<NSFetchRequest<NSFetchRequestResult>, [NSManagedObjectID]>` with `cache: NSCache<NSFetchRequest<NSFetchRequestResult>, [NSManagedObjectID]>` which is thread-safe out of the box 👍

Sources/StreamChat/Database/DatabaseContainer.swift
```diff
@@ -201,10 +201,10 @@ class DatabaseContainer: NSPersistentContainer {
     func write(_ actions: @escaping (DatabaseSession) throws -> Void, completion: @escaping (Error?) -> Void) {
         writableContext.perform {
             log.debug("Starting a database session.", subsystems: .database)
-            // Clear the FetchCache
-            FetchCache.clear()
             do {
+                FetchCache.clear()
```
polqf, line 205: Better have one before the do, and another one after the catch ends. That way you avoid repetition

Sources/StreamChat/Utils/Database/NSManagedObject+Extensions.swift
```diff
@@ -16,6 +16,7 @@ extension NSEntityDescription {
         for request: NSFetchRequest<T>
     ) -> T {
         let entity = insertNewObject(forEntityName: T.entityName, into: context)
+        request.entity = NSEntityDescription.entity(forEntityName: T.entityName, in: context)!
```
polqf, line 19: Why is this needed? Aren't we doing that already? Also, seems unsafe

Tests/StreamChatTests/APIClient/Endpoints/Payloads/ChannelListPayload_Tests.swift
```diff
@@ -21,8 +21,8 @@ final class ChannelListPayload_Tests: XCTestCase {
     func test_hugeChannelListQuery_save() throws {
         let decodedPayload = createHugeChannelList()
         
-        let databaseContainer = DatabaseContainer_Spy()
         measure {
+            let databaseContainer = DatabaseContainer_Spy()
```
polqf, line 25: This should be outside of the measure, we don't want the initialization of it to add on the time

Sources/StreamChat/Database/DatabaseContainer.swift
b-onc: we need to clear the cache before calling `save` since `save` kicks off any observers. `didChange` callbacks sometimes use DB and this causes the cache to have old info, and it causes CoreData errors

Sources/StreamChat/Utils/Database/NSManagedObject+Extensions.swift
b-onc: Casting `request` to `NSFetchRequest<NSFetchRequestResult>` requires `request` to have an `entity` set, I've seen it crashing because of this. It's safe since `T.entityName` is guaranteed to be a registered entity name

Tests/StreamChatTests/APIClient/Endpoints/Payloads/ChannelListPayload_Tests.swift
b-onc: True, but we need the DB to be "clear" so save operation actually takes place. if this line is out of `measure` block, DB already has data after the first iteration, so other iterations only overwrite data, without actually creating DTOs. If we always test in the same manner it shouldn't matter much, since improvements will be visible

Sources/StreamChat/Database/DatabaseContainer.swift
polqf: Good point. What about a local block so that we don't need to repeat `FetchCache.clear()` so often then?

Sources/StreamChat/Utils/Database/NSManagedObject+Extensions.swift
polqf: let's do an if let there, with an assertion failure in the else. To play it safe. WDYT?

Tests/StreamChatTests/APIClient/Endpoints/Payloads/ChannelListPayload_Tests.swift
polqf: Got you!

Sources/StreamChat/Utils/Database/NSManagedObject+Extensions.swift
donnywals: You should be able to write these two lines as follows and not have to worry about the entity name:

```
let entity = T(context: context)
```

That should work as long as the entity you're initializing is correctly set up in the model editor where the model name is connected to the correct class name (which is should be because you'd see various issues otherwise).

Sources/StreamChat/Utils/Database/NSManagedObject+Extensions.swift
```diff
@@ -10,13 +10,109 @@ extension NSManagedObject {
...
+    }
+    
+    static func load<T: NSManagedObject>(keyPath: String, equalTo value: String, context: NSManagedObjectContext) -> [T] {
+        let request = NSFetchRequest<T>(entityName: entityName)
+        request.predicate = NSPredicate(format: "%K == %@", keyPath, value)
+        return load(by: request, context: context)
+    }
+    
     static func load<T>(by request: NSFetchRequest<T>, context: NSManagedObjectContext) -> [T] {
+        request.entity = NSEntityDescription.entity(forEntityName: entityName, in: context)!
```
donnywals, line 55: This looks like it could be a bug waiting to happen; there is no guarantee that `T` will be the same as `Self` but `Self.entityName` used to configure the request. For example, I think I could, in theory, do something like this which IMO is undesirable:

```
let request: NSFetchRequest<ChannelDTO> = NSFetchRequest(entityName: ChannelDTO.entityName)
MessageDTO.load(request, context: someContext)
```

It's probably better to use `Self` here instead of `T` so that the fetch request that's used must return the managed object subclass type that `load` was called on.

Sources/StreamChat/Utils/Database/NSManagedObject+Extensions.swift
donnywals: It would be nice if there was some way of caching objects by their Stream ID (even though it's tricky). If you would go down that route, you can prewarm the cache by picking apart the JSON and, for example, do one fetch request for all UserDTO objects by user ids found in the JSON. 

Let's say that you have a JSON with 50 unique users, you could save lots of queries if those users already exist in the database because you can preload them all into the cache in one go.

Sources/StreamChat/Database/DatabaseContainer.swift
b-onc: i couldn't figure it out, how would that look?

Sources/StreamChat/Utils/Database/NSManagedObject+Extensions.swift
b-onc: @donnywals that's good feedback for phase 2! Notion doc has some steps for that

Sources/StreamChat/Utils/Database/NSManagedObject+Extensions.swift
b-onc: or, as you've said above:
```
request.entity = T(context: context)
```
would work?

Sources/StreamChat/Utils/Database/NSManagedObject+Extensions.swift
```diff
@@ -16,7 +16,7 @@ extension NSEntityDescription {
         for request: NSFetchRequest<T>
     ) -> T {
         let entity = insertNewObject(forEntityName: T.entityName, into: context)
-        request.entity = NSEntityDescription.entity(forEntityName: T.entityName, in: context)!
+        request.entity = T(context: context).entity
```
donnywals, line 19: This will create a new object and insert it into the specified context so you'll want to replace `let entity = insertNewObject(forEntityName: T.entityName, into: context)` with `let entity = T(context: context)` instead. And then use `request.entity = entity.entity` (might want to rename entity)

Sources/StreamChat/Utils/Database/NSManagedObject+Extensions.swift
donnywals: Not quite, that line would insert a new object of type `T` into the context.

Sources/StreamChat/Utils/Database/NSManagedObject+Extensions.swift
b-onc: fixed!

Sources/StreamChat/Utils/Database/NSManagedObject+Extensions.swift
b-onc: fixed!

Sources/StreamChat/Utils/Database/NSManagedObject+Extensions.swift
b-onc: fixed!

Sources/StreamChat/Utils/Database/NSManagedObject+Extensions.swift
```diff
@@ -10,13 +10,109 @@ extension NSManagedObject {
...
+    private var cache = [NSFetchRequest<NSFetchRequestResult>: [NSManagedObjectID]]()
+
+    fileprivate func set<T>(_ request: NSFetchRequest<T>, objectIds: [NSManagedObjectID]) where T: NSFetchRequestResult {
+        guard let request = request as? NSFetchRequest<NSFetchRequestResult> else {
+            log.assertionFailure("Request should have a generic type conforming to NSFetchRequestResult")
+            return
+        }
+        queue.sync {
+            cache[request] = objectIds
+        }
```
nuno-vieira, line 80: No need to block the thread here
```suggestion
        queue.async {
            cache[request] = objectIds
        }
```

Sources/StreamChat/Utils/Database/NSManagedObject+Extensions.swift
```diff
@@ -10,13 +10,109 @@ extension NSManagedObject {
...
         } catch {
             log.error("Could not load \(error), \(error.localizedDescription)")
             return []
         }
     }
 }
+
+class FetchCache {
+    fileprivate static let shared = FetchCache()
+    private let queue = DispatchQueue(label: "io.stream.com.fetch-cache", qos: .userInitiated)
```
nuno-vieira, line 70: We can make this more performant if we make it concurrent:
```suggestion
    private let queue = DispatchQueue(label: "io.stream.com.fetch-cache", qos: .userInitiated, attributes: .concurrent)
```

Then, we make our write async, and protect it with a barrier flag:
```swift
       queue.async(flags: .barrier) {
            objectIDs = cache[request]
        }
```

Sources/StreamChat/Utils/Database/NSManagedObject+Extensions.swift
```diff
@@ -10,13 +10,109 @@ extension NSManagedObject {
...
+    
+    static func clear() {
+        Self.shared.cache.removeAll()
+    }
+
+    private func clear() {
+        queue.sync {
+            cache.removeAll()
+        }
+    }
```
nuno-vieira, line 103: Don't quite understand why we need both? Why isn't `clear()` public? and then we just call `FetchCache.shared.clear()`. Currently, the static method is not thread-safe because it manipulates the cache directly without the `queue`.

polqf: Oh, that's my mistake. It is supposed to be using the clear instance method! Good catch

Sources/StreamChat/Utils/Database/NSManagedObject+Extensions.swift
```diff
@@ -10,13 +10,109 @@ extension NSManagedObject {
...
             log.error("Could not load \(error), \(error.localizedDescription)")
             return []
         }
     }
 }
+
+class FetchCache {
+    fileprivate static let shared = FetchCache()
+    private let queue = DispatchQueue(label: "io.stream.com.fetch-cache", qos: .userInitiated)
+    private var cache = [NSFetchRequest<NSFetchRequestResult>: [NSManagedObjectID]]()
```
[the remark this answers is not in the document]
polqf: NSCache requires both key and value to be classes and, in this case, we don't need that.
Also, NSCache comes with other logic that is not needed for us, we are just seeking thread safety

Sources/StreamChat/Utils/Database/NSManagedObject+Extensions.swift
b-onc: I fixed the thread-safety

Sources/StreamChat/Database/DTOs/ChannelListQueryDTO.swift
```diff
@@ -38,12 +40,12 @@ extension NSManagedObjectContext {
     }
     
     func saveQuery(query: ChannelListQuery) -> ChannelListQueryDTO {
-        if let existingDTO = ChannelListQueryDTO.load(filterHash: query.filter.filterHash, context: self) {
+        let request = ChannelListQueryDTO.fetchRequest(keyPath: "filterHash", equalTo: query.filter.filterHash)
+        if let existingDTO = ChannelListQueryDTO.load(by: request, context: self).first {
```
polqf, line 44: Shouldn't it use the method changed above? `static func load(filterHash: String, context: NSManagedObjectContext) -> ChannelListQueryDTO?`

Sources/StreamChat/Database/DTOs/ChannelMemberListQueryDTO.swift
```diff
@@ -19,17 +19,20 @@ final class ChannelMemberListQueryDTO: NSManagedObject {
...
+        load(
+            keyPath: #keyPath(ChannelMemberListQueryDTO.queryHash),
+            equalTo: queryHash,
+            context: context
+        ).first
     }
     
     static func loadOrCreate(queryHash: String, context: NSManagedObjectContext) -> ChannelMemberListQueryDTO {
-        if let existing = Self.load(queryHash: queryHash, context: context) {
+        let request = fetchRequest(keyPath: "queryHash", equalTo: queryHash)
```
polqf, line 30: Same as before. Shouldn't it use the method changed above?

Sources/StreamChat/Database/DTOs/DeviceDTO.swift
```diff
@@ -34,11 +32,12 @@ extension DeviceDTO {
     ///   - context: The context used to fetch/create `UserDTO`
     ///
     static func loadOrCreate(id: String, context: NSManagedObjectContext) -> DeviceDTO {
-        if let existing = Self.load(id: id, context: context) {
+        let request = fetchRequest(id: id)
+        if let existing = load(by: request, context: context).first {
```
polqf, line 36: Same as in other cases

Sources/StreamChat/Database/DTOs/MemberModelDTO.swift
```diff
@@ -69,12 +67,14 @@ extension MemberDTO {
     ///   - context: The context used to fetch/create `UserDTO`
     ///
     static func loadOrCreate(userId: String, channelId: ChannelId, context: NSManagedObjectContext) -> MemberDTO {
-        if let existing = Self.load(userId: userId, channelId: channelId, context: context) {
+        let memberId = MemberDTO.createId(userId: userId, channeldId: channelId)
+        let request = fetchRequest(id: memberId)
+        if let existing = load(by: request, context: context).first {
```
polqf, line 72: Same

Sources/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -341,17 +341,16 @@ class MessageDTO: NSManagedObject {
...
-        let request = NSFetchRequest<MessageDTO>(entityName: entityName)
-        request.predicate = NSPredicate(format: "id == %@", id)
-        return load(by: request, context: context).first
+        load(by: id, context: context).first
     }
     
     static func loadOrCreate(id: String, context: NSManagedObjectContext) -> MessageDTO {
-        if let existing = load(id: id, context: context) {
+        let request = fetchRequest(id: id)
+        if let existing = load(by: request, context: context).first {
```
polqf, line 349: same

Sources/StreamChat/Database/DTOs/MessageReactionDTO.swift
```diff
@@ -73,14 +71,14 @@ extension MessageReactionDTO {
         user: UserDTO,
         context: NSManagedObjectContext
     ) -> MessageReactionDTO {
-        let userId = user.id
-
-        if let existing = Self.load(userId: userId, messageId: message.id, type: type, context: context) {
+        let id = createId(userId: user.id, messageId: message.id, type: type)
+        let request = fetchRequest(id: id)
+        if let existing = load(by: request, context: context).first {
```
polqf, line 76: same

Sources/StreamChat/Database/DTOs/MessageSearchQueryDTO.swift
```diff
@@ -24,12 +26,15 @@ extension NSManagedObjectContext: MessageSearchDatabaseSession {
     }
     
     func saveQuery(query: MessageSearchQuery) -> MessageSearchQueryDTO {
-        if let existingDTO = MessageSearchQueryDTO.load(filterHash: query.filterHash, context: self) {
+        let request = MessageSearchQueryDTO.fetchRequest(
+            keyPath: #keyPath(MessageSearchQueryDTO.filterHash),
+            equalTo: query.filterHash
+        )
+        if let existingDTO = MessageSearchQueryDTO.load(by: request, context: self).first {
```
polqf, line 33: same

Sources/StreamChat/Database/DTOs/UserDTO.swift
```diff
@@ -80,11 +78,12 @@ extension UserDTO {
     ///   - context: The context used to fetch/create `UserDTO`
     ///
     static func loadOrCreate(id: String, context: NSManagedObjectContext) -> UserDTO {
-        if let existing = Self.load(id: id, context: context) {
+        let request = fetchRequest(id: id)
+        if let existing = load(by: request, context: context).first {
```
polqf, line 82: same

Sources/StreamChat/Database/DTOs/UserListQueryDTO.swift
```diff
@@ -45,12 +47,15 @@ extension NSManagedObjectContext {
             return nil
         }
         
-        if let existingDTO = UserListQueryDTO.load(filterHash: filterHash, context: self) {
+        let request = UserListQueryDTO.fetchRequest(
+            keyPath: #keyPath(UserListQueryDTO.filterHash),
+            equalTo: filterHash
+        )
+        if let existingDTO = UserListQueryDTO.load(by: request, context: self).first {
```
polqf, line 54: same

Sources/StreamChat/Database/DatabaseContainer.swift
```diff
@@ -153,6 +153,8 @@ class DatabaseContainer: NSPersistentContainer {
             }
         }
         
+        FetchCache.clear()
```
polqf, line 156: Why is this needed? On init there should not be anything hanging on the cache. Is this maybe a fix for the tests?

Sources/StreamChat/Utils/Database/NSManagedObject+Extensions.swift
```diff
@@ -10,13 +10,109 @@ extension NSManagedObject {
     }
 }
 
+extension NSEntityDescription {
+    class func insertNewObject<T: NSManagedObject>(
+        into context: NSManagedObjectContext,
+        for request: NSFetchRequest<T>
+    ) -> T {
+        let obj = T(context: context)
```
polqf, line 18: So this will do the same as the old `insertNewObject` function? Good to know

Sources/StreamChat/Utils/Database/NSManagedObject+Extensions.swift
```diff
@@ -10,13 +10,109 @@ extension NSManagedObject {
...
+    }
+    
+    static func load<T: NSManagedObject>(by request: NSFetchRequest<T>, context: NSManagedObjectContext) -> [T] {
+        request.entity = NSEntityDescription.entity(forEntityName: T.entityName, in: context)!
         do {
-            return try context.fetch(request)
+            return try context.fetch(
+                request,
+                using: FetchCache.shared
+            )
```
polqf, line 60: Why is this not in the same line?

Tests/StreamChatTests/StreamChatIntegrationTests/MemberEvents_IntegrationTests.swift
```diff
@@ -57,8 +60,11 @@ final class MemberEvents_IntegrationTests: XCTestCase {
         let event = try eventDecoder.decode(from: json) as? MemberUpdatedEventDTO
 
         let unwrappedEvent = try XCTUnwrap(event)
-        client.eventNotificationCenter.process(unwrappedEvent)
+        var completionCalled = false
+        client.eventNotificationCenter.process(unwrappedEvent) { completionCalled = true }
```
polqf, line 64: As discussed, let's not add more of this

Tests/StreamChatTests/StreamChatIntegrationTests/MessageEvents_IntegrationTests.swift
```diff
@@ -41,7 +41,10 @@ final class MessageEvents_IntegrationTests: XCTestCase {
         )
 
         let unwrappedEvent = try XCTUnwrap(event)
-        client.eventNotificationCenter.process(unwrappedEvent)
+        var completionCalled = false
+        client.eventNotificationCenter.process(unwrappedEvent) { completionCalled = true }
```
polqf, line 45: same as before

Tests/StreamChatTests/StreamChatIntegrationTests/MemberEvents_IntegrationTests.swift
```diff
@@ -40,7 +40,10 @@ final class MemberEvents_IntegrationTests: XCTestCase {
             try session.saveChannel(payload: self.dummyPayload(with: unwrappedEvent.cid))
         }
 
-        client.eventNotificationCenter.process(unwrappedEvent)
+        var completionCalled = false
+        client.eventNotificationCenter.process(unwrappedEvent) { completionCalled = true }
```
polqf, line 44: same

Tests/StreamChatTests/StreamChatIntegrationTests/ChannelEvents_IntegrationTests.swift
```diff
@@ -45,7 +45,10 @@ final class ChannelEventsIntegration_Tests: XCTestCase {
         let channelId: ChannelId = ChannelId(type: .messaging, id: "new_channel_7070")
 
         let unwrappedEvent = try XCTUnwrap(event)
-        client.eventNotificationCenter.process(unwrappedEvent)
+        var completionCalled = false
+        client.eventNotificationCenter.process(unwrappedEvent) { completionCalled = true }
```
polqf, line 49: same

Tests/StreamChatTests/StreamChatIntegrationTests/MessageEvents_IntegrationTests.swift
```diff
@@ -119,7 +122,10 @@ final class MessageEvents_IntegrationTests: XCTestCase {
         XCTAssertNil(client.databaseContainer.viewContext.message(id: "042772db-4af2-460d-beaa-1e49d1b8e3b9"))
 
         let unwrappedEvent = try XCTUnwrap(event)
-        client.eventNotificationCenter.process(unwrappedEvent)
+        var completionCalled = false
+        client.eventNotificationCenter.process(unwrappedEvent) { completionCalled = true }
```
polqf, line 126: same

Tests/StreamChatTests/Workers/ChannelUpdater_Tests.swift
```diff
@@ -144,11 +146,15 @@ final class ChannelUpdater_Tests: XCTestCase {
         }
 
         // Simulate `updateChannel` call
-        channelUpdater.update(channelQuery: query, isInRecoveryMode: false, channelCreatedCallback: callback, completion: nil)
+        var completionCalled = false
+        channelUpdater
+            .update(channelQuery: query, isInRecoveryMode: false, channelCreatedCallback: callback) { _ in completionCalled = true }
```
polqf, line 151: same

Tests/StreamChatTests/Workers/ChannelUpdater_Tests.swift
```diff
@@ -172,11 +178,15 @@ final class ChannelUpdater_Tests: XCTestCase {
         }
 
         // Simulate `updateChannel` call
-        channelUpdater.update(channelQuery: query, isInRecoveryMode: true, channelCreatedCallback: callback, completion: nil)
+        var completionCalled = false
+        channelUpdater
+            .update(channelQuery: query, isInRecoveryMode: true, channelCreatedCallback: callback) { _ in completionCalled = true }
```
polqf, line 183: same

Tests/StreamChatTests/Workers/ChannelUpdater_Tests.swift
```diff
@@ -1334,13 +1344,16 @@ final class ChannelUpdater_Tests: XCTestCase {
         // Call `channelWatchers` for this channel
         // This query doesn't provide any `offset` so it's requesting the first page of watchers
         let query = ChannelWatcherListQuery(cid: cid)
-        channelUpdater.channelWatchers(query: query)
+        var completionCalled = false
+        channelUpdater.channelWatchers(query: query) { _ in completionCalled = true }
```
polqf, line 1348: same

Tests/StreamChatTests/Workers/UserListUpdater_Tests.swift
```diff
@@ -113,29 +114,35 @@ final class UserListUpdater_Tests: XCTestCase {
     func test_mergePolicy_takesAffect() throws {
         // Simulate `update` call
         let query = UserListQuery(filter: .equal(.id, to: "Luke"))
-        listUpdater.update(userListQuery: query)
+        var completionCalled = false
+        listUpdater.update(userListQuery: query) { _ in completionCalled = true }
```
polqf, line 118: same

Tests/StreamChatTests/Workers/UserListUpdater_Tests.swift
```diff
@@ -160,13 +167,16 @@ final class UserListUpdater_Tests: XCTestCase {
         // Simulate `update` call
         // This call doesn't need `policy` argument specified since
         // it's the first call for this query, hence there's no data to `replace` or `merge` to
-        listUpdater.update(userListQuery: query)
+        var completionCalled = false
+        listUpdater.update(userListQuery: query) { _ in completionCalled = true }
```
polqf, line 171: same

Tests/StreamChatTests/Workers/UserListUpdater_Tests.swift
```diff
@@ -175,13 +185,16 @@ final class UserListUpdater_Tests: XCTestCase {
         AssertAsync.willBeTrue(user != nil)
         
         // Simulate consequent `update` call with new users and `.replace` policy
-        listUpdater.update(userListQuery: query, policy: .replace)
+        completionCalled = false
+        listUpdater.update(userListQuery: query, policy: .replace) { _ in completionCalled = true }
```
polqf, line 189: same

polqf (changes requested):

Sources/StreamChat/Database/DTOs/AttachmentDTO.swift
```diff
@@ -54,17 +54,16 @@ class AttachmentDTO: NSManagedObject {
...
-        let request = NSFetchRequest<AttachmentDTO>(entityName: AttachmentDTO.entityName)
-        request.predicate = NSPredicate(format: "id == %@", id.rawValue)
-        return try? context.fetch(request).first
+        load(by: id.rawValue, context: context).first
     }
     
     static func loadOrCreate(id: AttachmentId, context: NSManagedObjectContext) -> AttachmentDTO {
-        if let existing = Self.load(id: id, context: context) {
+        let request = fetchRequest(id: id.rawValue)
+        if let existing = load(by: request, context: context).first {
```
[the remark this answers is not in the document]
b-onc: doesn't really matter, since `load(id:)` will create the `request` again it's easier to use the `load(request:)` one

Sources/StreamChat/Database/DTOs/ChannelListQueryDTO.swift
b-onc: same reason as in `AttachmentDTO`. this case, lines 43-49 act like `loadOrCreate`, so calling `load(filterHash)` would cause another `request` to be created, which is not necessary since we have the `request` already created for `insertNewEntity`

Sources/StreamChat/Database/DTOs/ChannelMemberListQueryDTO.swift
b-onc: same as `AttachmentDTO`

Sources/StreamChat/Database/DTOs/DeviceDTO.swift
b-onc: same as `AttachmentDTO`

Sources/StreamChat/Database/DTOs/MemberModelDTO.swift
b-onc: same as `AttachmentDTO`

Sources/StreamChat/Database/DTOs/MessageDTO.swift
b-onc: same as `AttachmentDTO`

Sources/StreamChat/Database/DTOs/MessageSearchQueryDTO.swift
b-onc: same as `AttachmentDTO`

Sources/StreamChat/Database/DTOs/UserDTO.swift
b-onc: same as `AttachmentDTO`

Sources/StreamChat/Database/DTOs/UserListQueryDTO.swift
b-onc: same as `AttachmentDTO`

Sources/StreamChat/Database/DatabaseContainer.swift
b-onc: Yes, tests re-create the whole stack, starting from `ChatClient` but `FetchCache` was lingering around, having dirty cache.

This can also resolve a possible edge-case where 2 `ChatClient` instances are alive

Sources/StreamChat/Utils/Database/NSManagedObject+Extensions.swift
b-onc: exactly!

Sources/StreamChat/Utils/Database/NSManagedObject+Extensions.swift
b-onc: i've separated the lines to easily comment line 59 to test non-fetch performance. can be on the same line

Sources/StreamChat/Database/DatabaseContainer.swift
polqf: Interestingly enough, this opens the door for another question:

Should the DatabaseContainer hold a reference to the cache? It would make sense conceptually, as there should not be a shared cache across clients.

polqf: That is something that might be out of the scope for this PR, as there is no easy way to do that now.
But, in any case, I would try not to have Test fixes in production code.

Sources/StreamChat/Database/DTOs/ChannelListQueryDTO.swift
```diff
@@ -40,11 +40,11 @@ extension NSManagedObjectContext {
     }
     
     func saveQuery(query: ChannelListQuery) -> ChannelListQueryDTO {
-        let request = ChannelListQueryDTO.fetchRequest(keyPath: "filterHash", equalTo: query.filter.filterHash)
-        if let existingDTO = ChannelListQueryDTO.load(by: request, context: self).first {
+        if let existingDTO = channelListQuery(filterHash: query.filter.filterHash) {
             return existingDTO
         }
         
+        let request = ChannelListQueryDTO.fetchRequest(keyPath: "filterHash", equalTo: query.filter.filterHash)
```
polqf, line 47: Can we use #keypath here?

Sources/StreamChat/Database/DTOs/ChannelMemberListQueryDTO.swift
```diff
@@ -27,11 +27,11 @@ final class ChannelMemberListQueryDTO: NSManagedObject {
     }
     
     static func loadOrCreate(queryHash: String, context: NSManagedObjectContext) -> ChannelMemberListQueryDTO {
-        let request = fetchRequest(keyPath: "queryHash", equalTo: queryHash)
-        if let existing = load(by: request, context: context).first {
+        if let existing = load(queryHash: queryHash, context: context) {
             return existing
         }
         
+        let request = fetchRequest(keyPath: "queryHash", equalTo: queryHash)
```
polqf, line 34: same

Sources/StreamChat/Database/DTOs/ChannelListQueryDTO.swift
b-onc: must've missed it, true!

polqf (approved):

PR #1978, opened 2022-05-05
[CIS-1763] Fix unread counts bumping logic 

evsaev:
### 🔗 Issue Links

CIS-1763

### 🎯 Goal

- Fix local unread counts going out of sync with the remote
- Fix channel not being marked as read when it's muted OR when all messages after last read are thread replies/silent/from muted users

### 📝 Summary

Fix unread counts being **incremented** when:
- the channel is muted
- the message author is muted

Fix unread counts not being **decremented**:
- the message is hard deleted

### 🛠 Implementation

- Add handling of missing cases to `ChannelReadUpdaterMiddleware`
- Move channel mutes from `UserDTO` to `CurrentUserDTO`
- Rename `mentionedMessages` to `mentioningMessages` in channel unread counts
- Remove unused parts of `ChannelMuteDatabaseSession`

#### New message:

Message in muted channel does not increment unread counts:
**GIVEN** I'm on the channel list
**AND** The channel is muted
**WHEN** Channel member sends a new message to the muted channel
**THEN** unread messages count is not incremented

Message in unmuted channel from muted user does not increment unread counts:
**GIVEN** I'm on the channel list
**AND** The channel is unmuted
**AND** The channel has a muted member
**WHEN** Muted member sends a new message to the channel
**THEN** unread messages count is not incremented

The silent message in unmuted channel does not increment unread counts:
**GIVEN** I'm on the channel list
**AND** The channel is unmuted
**AND** The channel has unmuted member
**WHEN** Member sends a new `silent` message to the channel
**THEN** unread messages count is not incremented

The system message in unmuted channel does not increment unread counts:
**GIVEN** I'm on the channel list
**AND** The channel is unmuted
**AND** The channel has unmuted member
**WHEN** Member performs action that results in system message posted to the channel
**THEN** unread messages count is not incremented

Thread reply in unmuted channel does not increment unread counts:
**GIVEN** I'm on the channel list
**AND** The channel is unmuted
**AND** The channel has unmuted member
**WHEN** Member leaves a new thread reply in the channel
**THEN** unread messages count is not incremented

Own message in unmuted channel does not increment unread counts:
**GIVEN** I'm logged in using 2 devices as the same user
**AND** I'm on the channel list on device 1
**AND** I'm in the channel on device 2
**AND** The channel is unmuted
**WHEN** I send the new message to the channel on device 2
**THEN** unread messages count is not incremented on device 1

Regular message in unmuted channel from unmuted member increments from unread counts:
**GIVEN** I'm on the channel list
**AND** The channel is unmuted
**AND** The channel has unmuted member
**WHEN** Member sends a regular message to the channel
**THEN** unread messages count is incremented

Thread reply sent to main channel in unmuted channel from unmuted member increments from unread counts:
**GIVEN** I'm on the channel list
**AND** The channel is unmuted
**AND** The channel has unmuted member
**WHEN** Member leaves a new thread reply in the channel
**AND** Member selects `send also to main channel` checkmark
**THEN** unread messages count is incremented

#### Deleted message:

Hard delete of unseen message decrements unread counts:
**GIVEN** I'm on the channel list
**AND** The channel is unmuted
**AND** The channel has unmuted member
**AND** Member send the new message to the channel
**AND** unread messages count is incremented
**WHEN** Member performs hard deleted on his message 
**THEN** unread messages count is decremented

Soft delete of unseen message does not decrement unread counts:
**GIVEN** I'm on the channel list
**AND** The channel is unmuted
**AND** The channel has unmuted member
**AND** Member send the new message to the channel
**AND** unread messages count is incremented
**WHEN** Member performs soft deleted on his message 
**THEN** unread messages count is not decremented

Hard delete of seen message does not decrement unread counts:
**GIVEN** I'm on the channel list
**AND** The channel is unmuted
**AND** The channel has unmuted member
**AND** Member send the new message to the channel
**AND** unread messages count is incremented
**AND** I open the channel and go back to channel list
**AND** unread messages count become 0 
**WHEN** Member performs hard deleted on his message 
**THEN** unread messages is not decremented

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/Models/Channel.swift
+++ b/Sources/StreamChat/Models/Channel.swift
@@ -251,8 +251,8 @@ public struct ChannelUnreadCount: Decodable, Equatable {
     public static let noUnread = ChannelUnreadCount(messages: 0, mentionedMessages: 0)
     
     /// The total number of unread messages in the channel.
-    public internal(set) var messages: Int
+    public let messages: Int
     
     /// The number of unread messages that mention the current user.
-    public internal(set) var mentionedMessages: Int
+    public let mentionedMessages: Int
 }
--- a/Sources/StreamChat/Models/Channel.swift
+++ b/Sources/StreamChat/Models/Channel.swift
@@ -248,11 +248,16 @@ extension ChatChannel: Hashable {
 /// A struct describing unread counts for a channel.
 public struct ChannelUnreadCount: Decodable, Equatable {
     /// The default value representing no unread messages.
-    public static let noUnread = ChannelUnreadCount(messages: 0, mentionedMessages: 0)
+    public static let noUnread = ChannelUnreadCount(messages: 0, mentions: 0)
     
     /// The total number of unread messages in the channel.
     public let messages: Int
     
     /// The number of unread messages that mention the current user.
-    public let mentionedMessages: Int
+    public let mentions: Int
+}
+
+public extension ChannelUnreadCount {
+    @available(*, deprecated, renamed: "mentions")
+    var mentionedMessages: Int { mentions }
 }
--- a/Sources/StreamChat/Models/Channel.swift
+++ b/Sources/StreamChat/Models/Channel.swift
@@ -227,8 +227,8 @@ extension ChatChannel {
     /// so backend creates a `cid` based on member's `id`s
     public var isDirectMessageChannel: Bool { cid.id.hasPrefix("!members") }
     
-    /// returns `true` if the channel has one or more unread messages for the current user.
-    public var isUnread: Bool { unreadCount.messages > 0 }
+    /// Returns `true` if the channel has one or more unread messages for the current user.
+    public var isUnread: Bool { unreadCount != .noUnread }
 }
 
 /// A type-erased version of `ChannelModel<CustomData>`. Not intended to be used directly.
--- a/Sources/StreamChat/WebSocketClient/EventMiddlewares/ChannelReadUpdaterMiddleware.swift
+++ b/Sources/StreamChat/WebSocketClient/EventMiddlewares/ChannelReadUpdaterMiddleware.swift
@@ -9,16 +9,22 @@ struct ChannelReadUpdaterMiddleware: EventMiddleware {
     func handle(event: Event, session: DatabaseSession) -> Event? {
         switch event {
         case let event as MessageNewEventDTO:
-            increaseUnreadCountIfNeeded(
+            incrementUnreadCountIfNeeded(
                 for: event.cid,
                 message: event.message,
                 session: session
             )
 
         case let event as NotificationMessageNewEventDTO:
-            increaseUnreadCountIfNeeded(
+            incrementUnreadCountIfNeeded(
                 for: event.channel.cid,
                 message: event.message,
+                session: session
+            )
+        
+        case let event as MessageDeletedEventDTO:
+            decrementUnreadCountIfNeeded(
+                event: event,
                 session: session
             )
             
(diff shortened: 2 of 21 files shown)


Sources/StreamChat/Models/Channel.swift
```diff
@@ -248,11 +248,16 @@ extension ChatChannel: Hashable {
...
+    public let messages: Int
     
     /// The number of unread messages that mention the current user.
-    public internal(set) var mentionedMessages: Int
+    public let mentioningMessages: Int
+}
+
+public extension ChannelUnreadCount {
+    @available(*, deprecated, renamed: "mentioningMessages")
+    var mentionedMessages: Int { mentioningMessages }
```
nuno-vieira, line 262: Is the renaming really necessary? The previous one is already fine IMO 🤔 Not sure if it is worth deprecating

Sources/StreamChat/WebSocketClient/EventMiddlewares/ChannelReadUpdaterMiddleware.swift
```diff
@@ -50,36 +56,104 @@ struct ChannelReadUpdaterMiddleware: EventMiddleware {
...
+        channelRead: ChannelReadDTO,
+        message: MessagePayload
+    ) -> String? {
+        guard channelRead.channel.mute == nil else {
+            return "Channel is muted"
+        }
+        
+        guard message.user.id != currentUser.user.id else {
+            return "Own messages do not affect unread counts"
+        }
```
nuno-vieira, line 135: `guards` should protect against the regular logic flow. In this case, the function name is "getSkippingReason", so we are actually interested in the skipping reason. Using regular `ifs`, in this case, would be much easier to follow the logic. Where right now we need to think backwards with the guard statements. We should try to avoid using guards in this case

Sources/StreamChat/WebSocketClient/EventMiddlewares/ChannelReadUpdaterMiddleware.swift
```diff
@@ -50,36 +56,104 @@ struct ChannelReadUpdaterMiddleware: EventMiddleware {
...
+        channelRead.unreadMessageCount = max(0, channelRead.unreadMessageCount - 1)
+    }
+        
+    private func unreadCountUpdateSkippingReason(
+        currentUser: CurrentUserDTO,
+        channelRead: ChannelReadDTO,
+        message: MessagePayload
+    ) -> String? {
+        guard channelRead.channel.mute == nil else {
+            return "Channel is muted"
```
nuno-vieira, line 130: Wouldn't it be more scalable if we introduced a UnreadSkippingReason? Also, is this logic meant to be overridable? Could the customers provide their only logic, or we don't want that?

nuno-vieira (approved):
Overall LGTM ✅ Added only some questions and nit comments 👍

Sources/StreamChat/Models/Channel.swift
evsaev: The current naming is confusing, IMO. It's not clear what are these mentioned messages, who has mentioned them and why we're interested in this number 😄 Whereas `mentioning` is more clear because it states that some of the unread messages mention the current user.

Sources/StreamChat/WebSocketClient/EventMiddlewares/ChannelReadUpdaterMiddleware.swift
evsaev: This logic just follows what's on the backend and keeps local unread counts in sync with the remote so, yeah, that's not supposed to be customizable by an integrator.

evsaev: Actually, an integrator will have control over some parts of this logic by specifying flags on app-level config (e.g. if they want to increment unread count when the channel is muted) and we will respect those flags

Sources/StreamChat/WebSocketClient/EventMiddlewares/ChannelReadUpdaterMiddleware.swift
evsaev: Done ✅ Also introduced private `UnreadSkippingReason` enum

Sources/StreamChat/Models/Channel.swift
evsaev: Done ✅

Sources/StreamChat/WebSocketClient/EventMiddlewares/ChannelReadUpdaterMiddleware.swift
```diff
@@ -50,36 +56,124 @@ struct ChannelReadUpdaterMiddleware: EventMiddleware {
...
+            return .messageIsSilent
+        } else if message.parentId != nil && !message.showReplyInChannel {
+            return .messageIsThreadReply
+        } else if message.type == .system {
+            return .messageIsSystem
+        } else if message.createdAt <= (channelRead.lastReadAt ?? Date.distantPast) {
+            return .messageIsSeen
+        } else {
+            return nil
+        }
```
nuno-vieira, line 145: nit: The `else if` are not necessary because we use returns, we could do just:
```swift
if ... {
    return 
}

if ... {
    return 
}

if ... {
    return 
}

....
```

nuno-vieira (approved):

PR #1984, opened 2022-05-06
[CIS-1828] Increase safety when mapping Core Data entities

polqf:
### 🔗 Issue Links

https://stream-io.atlassian.net/browse/CIS-1828

### 🎯 Goal

Improve safety when mapping Core Data entities into models. Big focus on avoid accesses to deleted objects.

### 🛠 Implementation

- Make conversions to model throwable
  - `asModel` and similars now throw `InvalidModel` when it is not safe to access them
- Added a flag to disable lazy mappings
  - LazyCachedMapCollection can now map all the models on initialization, instead of lazily doing it.
- Fix "Switch User" on DemoApp
- New library (conditionally enabled when using developers target) to display FPS on the status bar

### 🎨 Showcase

Screenshot 2022-05-06 at 14 20 15

### 🧪 Manual Testing Notes

1. Open App
2. Select a user
3. Paginate channels list
4. Switch user

Previous result:
Crash 💥 

Actual result:
No crash should occur

### ☑️ Contributor Checklist


Sources/StreamChat/Controllers/ListDatabaseObserver.swift
```diff
@@ -172,7 +172,11 @@ class ListDatabaseObserver<Item, DTO: NSManagedObject> {
                 // `itemCreator` returns non-optional value, so we can use implicitly unwrapped optional
                 var result: Item!
                 context.performAndWait {
-                    result = itemCreator(dto)
+                    do {
+                        result = try itemCreator(dto)
+                    } catch {
+                        log.assertionFailure("Unable to convert a DB entity to model")
+                    }
```
b-onc, line 179: Won't this cause a crash anyway? `result` is still `nil` if error is thrown

nuno-vieira (approved):
LGTM! ✅

DemoApp/ChatPresenter.swift
```diff
@@ -325,6 +319,17 @@ class DemoChatChannelListRouter: ChatChannelListRouter {
         ])
     }
 
+    /// *Only for testing purposes*. Please do not replicate this. When changing users, you should recreate the controller and the view controller
+    private func switchTo(user: UserCredentials) {
```
evsaev, line 323: I would completely remove switching users from demo app, tbh. If we have it just for crash testing purposes, it could be in a separate branch rebased on top of `develop`.

Sources/StreamChat/Utils/LazyCachedMapCollection.swift
```diff
@@ -29,7 +29,16 @@ public struct LazyCachedMapCollection<Element>: RandomAccessCollection {
...
+
+        /// This is just an internal test to see how we behave when the DB models are immediately mapped instead of being lazily mapped
+        if StreamRuntimeCheck._isLazyMappingEnabled {
+            cache = .init(capacity: source.count)
+        } else {
+            if Thread.isMainThread {
+                log.debug("This should not be happening on the Main Thread")
+            }
+            cache = .init(elements: source.map(map))
+        }
```
evsaev, line 41: If there's no drop in performance, perhaps, we can just go for it and evaluate models immediately? If that's the case, can we start exposing a plain array of channels/messages by controllers?

polqf: This is what I want to work on the following PR. For now, I want to keep it easy, and fully validate the assumptions

Sources/StreamChat/Controllers/ListDatabaseObserver.swift
polqf: Good point. Adding it

DemoApp/ChatPresenter.swift
polqf: I was inclined to do so, tbh. Let's remove it

evsaev (approved):
🚢

b-onc:
This should go into today's release too

PR #2006, opened 2022-05-16
Introduce `ListChangeUpdater` to have consistent updates in list components

nuno-vieira:
### 🔗 Issue Links
None

### 🎯 Goal
Improve the efficiency of message list updates when diffing is disabled caused by the change in this PR: https://github.com/GetStream/stream-chat-swift/pull/2001

### 📝 Summary
- Use `performBatchUpdates()` on message list when diffing is disabled
- Remove old Database List Observer hacks
- Removes overriding of `reloadRows()`. We can't assume that depending on some data changes, the UI doesn't need to reload, because we don't know how is the Customer's UI, so we don't know when the cell needs to reload or not (When the cell needs to change size etc)
- Introduce a `ListChangeUpdater` to process list changes and apply them to a collection or table views. Currently, this is an internal API.

### 🎨 Showcase
N/A

### 🧪 Manual Testing Notes
Requires Regression Testing

### ☑️ Contributor Checklist


Sources/StreamChatUI/ChatMessageList/ChatMessageListView.swift
```diff
@@ -157,101 +157,34 @@ open class ChatMessageListView: UITableView, Customizable, ComponentsProvider {
...
-        let hasConflictUpdates = collectionUpdatesMapper.mapToSetsOfIndexPaths(changes: changes) == nil
-        if hasConflictUpdates {
-            reloadData()
-            return
-        }
-
-        guard changes.count == 1, let change = changes.first else {
+        guard let indices = collectionUpdatesMapper.mapToSetsOfIndexPaths(
+            changes: changes
+        ) else {
```
evsaev, line 169: What if we move `collectionUpdatesMapper` to list database observer and in both UI frameworks just blindly apply what comes from LLC?

nuno-vieira: We can't do that, because we need to notify about `ListChange<T>`, and the `collectionUpdatesMapper` returns index paths only. But I thought about introducing a `TableViewSafeBatchUpdater` and `CollectionViewBatchUpdater` which would be useful to reuse this logic between all table views and collection views. 👍 Will notify you once I change this 👍

evsaev: What if it returns `[ListChange<T>]` but with conflicting updates removed?

nuno-vieira: It will be 2x the work. Because then we will need to stilll map them and separate Inserts, Updates, Deletes and Moves, and call them in the correct order. So it is better we do this in the same for-loop to improve performance

evsaev: We say that customers can use LLC without a UI framework but in fact, it's not possible if LLC keeps publishing crash-prone updates.

We indeed will have an increase in time complexity for additional changes processing but it'll be a constant one (`2*O(N)` instead of `O(N)`). The `N` of changes is relatively small while the benefit would be a fully functional LLC and no code duplication of patches in message/channel lists

evsaev: I'm ok to move it to tech discussion, no need to block a PR with that 👍

nuno-vieira: @evsaev The PR is not blocked, I just didn't have the time yet to pick it up again. Will do it in the next couple of days. And yes I was planning on exposing this to customers and writing documentation on how they can use this with their own UI 🙂 Let me work on a WIP implementation of this first, and then we can discuss it together and bring it to tech discussion 👍

nuno-vieira: @evsaev Added the implementation with the concept of `ListChangeUpdater`. IMO this is a pretty solid solution, but let's discuss this in the Tech Meeting 👍 The problem with your initial idea is that it doesn't solve the fact that we need to make sure `performBatchUpdates()` is called in the correct order. With this solution, the updates to collection/table views are abstracted from the customer and they can use these updaters easily with their custom views 🙂

Sources/StreamChatUI/Utils/ListChangeUpdater/ListChangeUpdater.swift
```diff
@@ -0,0 +1,26 @@
+//
+// Copyright © 2022 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+import StreamChat
+
+/// Component responsible to process an array of `[ListChange<Item>]`'s and apply those changes to a view.
+protocol ListChangeUpdater {
```
evsaev, line 9: nit:
Maybe smth like `ListReloader` or `ListUpdater` would work better?

Sources/StreamChatUI/Utils/ListChangeUpdater/TableViewListChangeUpdater.swift
```diff
@@ -0,0 +1,44 @@
+//
+// Copyright © 2022 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+import StreamChat
+import UIKit
+
+/// Component responsible to process an array of `[ListChange<Item>]`'s and apply those changes to a table view.
+class TableViewListChangeUpdater: ListChangeUpdater {
```
evsaev, line 10:
```suggestion
final class TableViewListChangeUpdater: ListChangeUpdater {
```

Tests/StreamChatUITests/Utils/ListChangeUpdater/TableViewListChangeUpdater_Tests.swift
```diff
@@ -0,0 +1,135 @@
+//
+// Copyright © 2022 Stream.io Inc. All rights reserved.
+//
+
+@testable import StreamChat
+@testable import StreamChatUI
+import XCTest
+
+class TableViewListChangeUpdater_Tests: XCTestCase {
```
evsaev, line 9:
```suggestion
final class TableViewListChangeUpdater_Tests: XCTestCase {
```

evsaev (approved):
LGTM! ✅

Sources/StreamChatUI/Utils/ListChangeUpdater/ListChangeUpdater.swift
nuno-vieira: I think that ListChangeUpdater is fine actually. `ListChange` prefix is related to `ListChange<Item>`. `Updater`, means it uses List Changes' to update the view 😄 Reloader is not really true because we can also delete items. `Reloader` feels like we just do `reloadData()`

nuno-vieira: Maybe `ListChangesUpdater` could also work. But since for now, this is a private API, we can re-name and think about this later, WDTY?

PR #2007, opened 2022-05-16
[CIS-1558] Fix crash `unconditionallyBridgeFromObjectiveC` from CoreData DTOs

nuno-vieira:
### 🔗 Issue Links
CIS-1558

### 🎯 Goal
Fix crash  `unconditionallyBridgeFromObjectiveC` that happens when `CoreData` runtime attemps converting a `NSDate` to `Date?`.

### 🧪 Manual Testing Notes
A swiftlint rule will be added to avoid regressions.

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/Database/DTOs/ChannelDTO.swift
+++ b/Sources/StreamChat/Database/DTOs/ChannelDTO.swift
@@ -79,8 +79,7 @@ class ChannelDTO: NSManagedObject {
         
         // Update the date for sorting every time new message in this channel arrive.
         // This will ensure that the channel list is updated/sorted when new message arrives.
-        let lastDate = lastMessageAt ?? createdAt
-        if lastDate != defaultSortingAt {
+        if let lastDate = lastMessageAt, lastDate != defaultSortingAt {
             defaultSortingAt = lastDate
         }
     }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
     - decrement when message is hard deleted
 - Fix paginated channels in channel list were left without messages when sync is executed [#1985](https://github.com/GetStream/stream-chat-swift/issues/1985)
 - Fix `deletedMessagesVisibility == .alwaysVisible` shows deleted ephemeral messages in message list [#1991](https://github.com/GetStream/stream-chat-swift/issues/1991)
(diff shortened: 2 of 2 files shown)


Sources/StreamChat/Database/DTOs/ChannelDTO.swift
```diff
@@ -79,8 +79,7 @@ class ChannelDTO: NSManagedObject {
         
         // Update the date for sorting every time new message in this channel arrive.
         // This will ensure that the channel list is updated/sorted when new message arrives.
-        let lastDate = lastMessageAt ?? createdAt
```
polqf, line 82: Doesn't this mean we will NOT always update the channel as it is stated in the comment? Is this to cause any issues?

evsaev: We need to check how it works with an empty channel. Previously we were making an assumption that `lastMessageAt ?? createdAt` always gives a non-optional date and `defaultSortingAt` is always has a value and can be used in sorting. Now, an empty channel can go to the bottom of the channel list when created

evsaev: Btw, if the sorting option is not provided, the backend returns channels sorted by `updatedAt` (not by `lastMessageAt`)

evsaev: Another issue is that we mark `createdAt/updatedAt` as non-optional fields but we don't give a default value. Can that be the root cause of the crash?

nuno-vieira: > Yeah, we need to check how it works with an empty channel
It works

> Now, an empty channel can go to the bottom of the channel list when created
This is not happening

> Btw, if the sorting option is not provided, the backend returns channels sorted by updatedAt (not by lastMessageAt)
This issue is already there if it is true

> Another issue is that we mark createdAt/updatedAt as non-optional fields but don't give a default value. Can that be the root cause of the crash?
We could also do this yes. Actually, I think we should do it for all dates, to avoid this annoying crash.

evsaev: Can we check that providing default values for `createdAt/updatedAt` also fixes the crash?

nuno-vieira: I will need to replicate this state again and confirm. I'll try to do this, but I'm not sure how I was able to reproduce this state. Will keep you posted 👍

nuno-vieira:
@evsaev Closing the ticket since we identified the root cause and will properly fix it in another PR 🙂

PR #2010, opened 2022-05-17
4.15.0 Release

evsaev:
### 🎯 Goal

Fix failing release workflow

### 📝 Summary

Pass Xcode version via environment variables

### ☑️ Contributor Checklist

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -115,7 +115,7 @@ lane :publish_release do |options|
   UI.user_error!("You need to pass the version of the release") unless version.length > 0
 
   # Create XCFrameworks
-  xcversion(version: "13.0")
+  xcversion(version: xcode_version)
   sh("cd .. && make frameworks")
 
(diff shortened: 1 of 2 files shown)


fastlane/Fastfile
```diff
@@ -115,7 +115,6 @@ lane :publish_release do |options|
   UI.user_error!("You need to pass the version of the release") unless version.length > 0
 
   # Create XCFrameworks
-  xcversion(version: "13.0")
```
polqf, line 118: Do not remove this, because the other step is pointing to ~> 13.0, which will mean 13.0.1 if there is any fix released.

It will also make it future proof

polqf (changes requested):

fastlane/Fastfile
polqf: Otherwise we need to change the `before_all` to point to an specific version, which could also be an option

polqf (approved):

PR #2011, opened 2022-05-17
[CIS-1869] Fix launching DemoApp in offline state

b-onc:
### 🔗 Issue Links

CIS-1869
Discussion in slack: https://getstream.slack.com/archives/CRB9P6L03/p1652798185424229

### 🎯 Goal

1. When DemoApp is launched offline, the error alert is displayed and the only action logs out the user, making DemoApp unusable offline.
1. The alerts displayed are very ugly
1. DemoApp doesn't have login caching, requires user to "Login" on each app start.

### 📝 Summary

1. We decided to skip the alert if we know we're offline
1. The alerts should display the error message (or reason), not the whole error.
1. DemoApp already has login caching, used for push notification app starts. We can use the same mechanism to login on app start.

https://user-images.githubusercontent.com/12818985/170296163-c5f1c253-0623-404d-8a9e-09b8ca804ce8.mov

### 🧪 Manual Testing Notes

**GIVEN** user logs in as a `built-in` user
**AND** user kills the app
**WHEN** user launches the app offline
**THEN** user is navigated to the channel list
**AND** user observes the banner with the connection status

**GIVEN** user launches app offline
**AND** user tries to log in
**THEN** user is navigated to the channel list
**AND** user observes the banner with the connection status

**GIVEN** user logs in as a `built-in` user
**AND** user kills the app
**WHEN** user launches the app again
**THEN** user is logged in automatically
**AND** user is navigated to the channel list

**GIVEN** user logs in as a `guest/anonymous`
**AND** user kills the app
**WHEN** user launches the app again
**THEN** user is NOT logged in automatically
**AND** user is navigated to the login screen

**GIVEN** user logs in as a `built-in` user
**AND** user kills the app/sends app to background
**WHEN** the new message is posted to the channel
**AND** user receives a push notification
**AND** user taps on push notification
**THEN** user is navigated to the channel with the new message

### ☑️ Contributor Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,16 +2,18 @@
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 # Upcoming
-## StreamChatUI
-### ✅ Added
-- Add Support for Slow Mode [#1953](https://github.com/GetStream/stream-chat-swift/pull/1953)
-### 🐞 Fixed
-- Fix DM Channel with multiple members displaying only 1 user avatar [#2019](https://github.com/GetStream/stream-chat-swift/pull/2019)
-
 ## StreamChat
 ### 🐞 Fixed
 - Saving payloads to local database is now 50% faster. Initial launch and displaying channel list should be noticeably faster [#1973](https://github.com/GetStream/stream-chat-swift/issues/1973)
 - Fix not waiting for last batch of events to be processed when connecting as another user [#2016](https://github.com/GetStream/stream-chat-swift/issues/2016)
+
+## StreamChatUI
+### ✅ Added
+- Add Support for Slow Mode [#1953](https://github.com/GetStream/stream-chat-swift/pull/1953)
+- Present channel screen modally when channel list in not embedded by navigation controller [#2011](https://github.com/GetStream/stream-chat-swift/pull/2011)
+- Show channel screen as right detail when channel list is embedded by split view controller [#2011](https://github.com/GetStream/stream-chat-swift/pull/2011)
+### 🐞 Fixed
+- Fix DM Channel with multiple members displaying only 1 user avatar [#2019](https://github.com/GetStream/stream-chat-swift/pull/2019)
 
 # [4.15.0](https://github.com/GetStream/stream-chat-swift/releases/tag/4.15.0)
 _May 11, 2022_
--- a/DemoApp/DemoAppCoordinator.swift
+++ b/DemoApp/DemoAppCoordinator.swift
@@ -16,21 +16,29 @@ private var isStreamInternalConfiguration: Bool {
     ProcessInfo.processInfo.environment["STREAM_DEV"] != nil
 }
 
-final class DemoAppCoordinator: NSObject, UNUserNotificationCenterDelegate {
-    var connectionController: ChatConnectionController?
-    let navigationController: UINavigationController
-    let connectionDelegate: BannerShowingConnectionDelegate
+final class DemoAppCoordinator: NSObject {
+    let window: UIWindow
 
-    init(navigationController: UINavigationController) {
-        self.navigationController = navigationController
-        connectionDelegate = BannerShowingConnectionDelegate(
-            showUnder: navigationController.navigationBar
-        )
+    init(window: UIWindow) {
+        self.window = window
+        
         super.init()
-
-        injectActions()
+        
+        UNUserNotificationCenter.current().delegate = self
     }
     
+    func start(cid: ChannelId? = nil) {
+        if let user = UserDefaults.shared.currentUser {
+            showChat(for: .credentials(user), cid: cid, animated: false)
+        } else {
+            showLogin(animated: false)
+        }
+    }
+}
+
+// MARK: - UNUserNotificationCenterDelegate
+
+extension DemoAppCoordinator: UNUserNotificationCenterDelegate {
     func userNotificationCenter(
         _ center: UNUserNotificationCenter,
         didReceive response: UNNotificationResponse,
@@ -52,25 +60,109 @@ final class DemoAppCoordinator: NSObject, UNUserNotificationCenterDelegate {
             return
         }
         
-        if let userId = UserDefaults(suiteName: applicationGroupIdentifier)?.string(forKey: currentUserIdRegisteredForPush),
-           let userCredentials = UserCredentials.builtInUsersByID(id: userId) {
-            presentChat(userType: .credentials(userCredentials), channelID: cid)
-        }
+        start(cid: cid)
     }
+}
 
-    func setupRemoteNotifications() {
-        UNUserNotificationCenter
-            .current()
-            .requestAuthorization(options: [.alert, .sound, .badge]) { granted, _ in
-                if granted {
-                    DispatchQueue.main.async {
-                        UIApplication.shared.registerForRemoteNotifications()
-                    }
-                }
-            }
+// MARK: - Navigation
+
+private extension DemoAppCoordinator {
+    func showChat(for user: DemoUserType, cid: ChannelId? = nil, animated: Bool) {
+        logIn(as: user)
+        
+        let chatVC = makeChatVC(for: user, startOn: cid) { [weak self] in
+            guard let self = self else { return }
+            
+            self.logOut()
+        }
+        
+        set(rootViewController: chatVC, animated: animated)
     }
     
+    func showLogin(animated: Bool) {
+        let loginVC = makeLoginVC { [weak self] user in
+            self?.showChat(for: user, animated: true)
+        }
+        
+        set(rootViewController: loginVC, animated: animated)
+    }
+    
+    func set(rootViewController: UIViewController, animated: Bool) {
+        if animated {
+            UIView.transition(with: window, duration: 0.3, options: .transitionFlipFromLeft) {
+                self.window.rootViewController = rootViewController
+            }
+        } else {
+            window.rootViewController = rootViewController
+        }
+    }
+}
+
+// MARK: - Screens factory
+
+private extension DemoAppCoordinator {
+    func makeLoginVC(onUserSelection: @escaping (DemoUserType) -> Void) -> UIViewController {
+        let storyboard = UIStoryboard(name: "Main", bundle: nil)
+        let loginNVC = storyboard.instantiateInitialViewController() as! UINavigationController
+        
+        let loginVC = loginNVC.viewControllers.first as! LoginViewController
+        loginVC.onUserSelection = onUserSelection
+        
+        return loginNVC
+    }
+    
+    func makeChatVC(for user: DemoUserType, startOn cid: ChannelId?, onLogout: @escaping () -> Void) -> UIViewController {
+        let channelListQuery: ChannelListQuery
+        switch user {
+        case let .credentials(userCredentials):
+            channelListQuery = .init(filter: .containMembers(userIds: [userCredentials.id]))
+        case .anonymous, .guest:
+            channelListQuery = .init(filter: .equal(.type, to: .messaging))
+        }
+        
+        let channelController = cid.map { ChatClient.shared.channelController(for: $0) }
+        let channelVC = channelController.map { makeChannelVC(controller: $0) }
+        let channelNVC = channelVC.map { UINavigationController(rootViewController: $0) }
+        
+        let selectedChannel = channelController?.channel
+        let channelListController = ChatClient.shared.channelListController(query: channelListQuery)
+        let channelListVC = makeChannelListVC(
+            controller: channelListController,
+            selectedChannel: selectedChannel,
+            onLogout: onLogout
+        )
+        let channelListNVC = UINavigationController(rootViewController: channelListVC)
+        
+        let splitVC = UISplitViewController()
+        splitVC.preferredDisplayMode = .oneBesideSecondary
+        splitVC.viewControllers = [channelListNVC, channelNVC].compactMap { $0 }
+        return splitVC
+    }
+    
+    func makeChannelListVC(
+        controller: ChatChannelListController,
+        selectedChannel: ChatChannel?,
+        onLogout: @escaping () -> Void
+    ) -> UIViewController {
+        let channelListVC = DemoChatChannelListVC.make(with: controller)
+        channelListVC.demoRouter.onLogout = onLogout
+        channelListVC.selectedChannel = selectedChannel
+        return channelListVC
+    }
+    
+    func makeChannelVC(controller: ChatChannelController) -> UIViewController {
+        let channelVC = DemoChatChannelVC()
+        channelVC.channelController = controller
+        return channelVC
+    }
+}
+
+// MARK: - Auth
+
+private extension DemoAppCoordinator {
     func setUpChat() {
+        guard ChatClient.shared == nil else { return }
+        
         // Set the log level
         LogConfig.level = .warning
         LogConfig.formatters = [
@@ -89,12 +181,12 @@ final class DemoAppCoordinator: NSObject, UNUserNotificationCenterDelegate {
         
         // Config
         Components.default.channelListRouter = DemoChatChannelListRouter.self
-        Components.default.channelVC = CustomChannelVC.self
-        Components.default.messageContentView = CustomMessageContentView.self
+        Components.default.channelVC = DemoChatChannelVC.self
+        Components.default.messageContentView = DemoChatMessageContentView.self
         Components.default.messageListDateSeparatorEnabled = true
         Components.default.messageListDateOverlayEnabled = true
         Components.default._messageListDiffingEnabled = isStreamInternalConfiguration
-        Components.default.messageActionsVC = CustomChatMessageActionsVC.self
+        Components.default.messageActionsVC = DemoChatMessageActionsVC.self
         Components.default.reactionsSorting = { $0.type.position < $1.type.position }
 
         StreamRuntimeCheck.assertionsEnabled = isStreamInternalConfiguration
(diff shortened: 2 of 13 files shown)


DemoApp/SceneDelegate.swift
```diff
@@ -34,6 +34,11 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
         self.window = window
         window.makeKeyAndVisible()
         scene.windows.forEach { $0.tintColor = .streamBlue }
+        
+        if let userId = UserDefaults(suiteName: applicationGroupIdentifier)?.string(forKey: currentUserIdRegisteredForPush),
+           let userCredentials = UserCredentials.builtInUsersByID(id: userId) {
+            coordinator.presentChat(userType: .credentials(userCredentials))
+        }
```
evsaev, line 41: I would add `start` method to `DemoAppCoordinator` and make this check inside

DemoApp/DemoAppCoordinator.swift
```diff
@@ -128,22 +128,44 @@ final class DemoAppCoordinator: NSObject, UNUserNotificationCenterDelegate {
...
         let connectCompletion: (Error?) -> Void = { [weak self] error in
+            guard let self = self else { return }
             if let error = error {
-                log.error("connecting the user failed \(error)")
-                if let self = self {
+                var errorTitle: String?
+                var errorMessage: String?
+                if let clientError = error as? ClientError,
+                   let underlyingError = clientError.underlyingError as? ClientError,
+                   let engineError = underlyingError.underlyingError as? WebSocketEngineError {
```
evsaev, line 137: What if we expose just `Error.isInternetOfflineError`?

DemoApp/SceneDelegate.swift
b-onc: good idea

DemoApp/DemoAppCoordinator.swift
```diff
@@ -227,7 +249,17 @@ final class DemoAppCoordinator: NSObject, UNUserNotificationCenterDelegate {
...
+        let currentUserController = ChatClient.shared.currentUserController()
+        if let deviceId = currentUserController.currentUser?.currentDevice?.id {
+            currentUserController.removeDevice(id: deviceId) { error in
+                if let error = error {
+                    log.warning("removing the device failed with an error \(error)")
+                }
+            }
+        }
         ChatClient.shared.disconnect()
+        UserDefaults(suiteName: applicationGroupIdentifier)?
```
evsaev, line 261: With this change, we start repeating ourselves. How about:
```swift
extension UserDefaults {
    static let shared = UserDefaults(suiteName: applicationGroupIdentifier)
}
```

DemoApp/SceneDelegate.swift
evsaev: That's a good time to refactor that part 🙂 If previously we didn't have any logic on startup, now we have. Launching from storyboard is no longer the way

DemoApp/DemoAppCoordinator.swift
b-onc: makes sense 👍

DemoApp/DemoAppCoordinator.swift
evsaev: Yeah, but exposing this is a bit scary:
```swift
                if let clientError = error as? ClientError,
                   let underlyingError = clientError.underlyingError as? ClientError,
                   let engineError = underlyingError.underlyingError as? WebSocketEngineError,
                   engineError.code == WebSocketEngineError.offlineCode ...  {
```
Can we maybe add `ClientError.isInternetOfflineError: Bool` for now?

DemoApp/SceneDelegate.swift
b-onc: > If previously we didn't have any logic on startup, now we have. 

It's literally 3 new lines added, we can't call this "we didn't have any logic, now we have" IMHO. `DemoAppCoordinator` needs a refactor for a proper `start` func, which would go over the scope of this PR.

DemoApp/DemoAppCoordinator.swift
nuno-vieira: I think it makes sense to provide the helper suggested by @evsaev, otherwise, it is very difficult to get to the `offlineCode` error.

DemoApp/DemoAppCoordinator.swift
```diff
@@ -128,22 +132,44 @@ final class DemoAppCoordinator: NSObject, UNUserNotificationCenterDelegate {
...
+                   let engineError = underlyingError.underlyingError as? WebSocketEngineError {
+                    if engineError.code == WebSocketEngineError.offlineErrorCode {
+                        log.info("connecting the user failed because we're offline")
+                    } else {
+                        errorTitle = "Connecting failed"
+                        errorMessage = engineError.reason
+                    }
+                } else if let clientError = error as? ClientError,
+                          let underlyingError = clientError.underlyingError as? ClientError,
+                          let errorPayload = underlyingError.underlyingError as? ErrorPayload {
```
nuno-vieira, line 150: Would be nice to make this nested underlying error easier to access. Can't we have a `ClientError.errorPayload: ErrorPayload?` as well as a `ClientError.webSocketEngineError` ?

nuno-vieira: I think at least we should create a ticket to try and improve the ClientError API 🤔

nuno-vieira:
> DemoApp already has login caching, used for push notification app starts. We can use the same mechanism to login on app start.

This is not working for me. I always have to select the user

DemoApp/SceneDelegate.swift
evsaev: I feel you but we never get back to fix tech debt in `DemoApp` 😄 it's like a snowball. 

The demo is the first thing devs check and it contributes to the first impression so spending some time making it look nice is worth it but that's just me.

evsaev: @GetStream/ios-developers WDYT?

b-onc:
@nuno-vieira are you testing on simulator or real device?

DemoApp/DemoAppCoordinator.swift
b-onc: this really needs a ticket since `ClientError.underlyingError` can be, and in this case _is_, another `ClientError`. so exposing `errorPayload` or anything would need ugly recursion

nuno-vieira:
> @nuno-vieira are you testing on simulator or real device?

I was using simulator, but why would this matter?

DemoApp/ChatPresenter.swift
```diff
@@ -14,7 +14,8 @@ class DemoChatChannelListRouter: ChatChannelListRouter {
     }
 
     var channelPresentingStyle: ChannelPresentingStyle = .push
-
+    var didLogout: (() -> Void)?
```
nuno-vieira, line 17: nit: Usually, closures have `on` prefix 👍 
```suggestion
    var onLogout: (() -> Void)?
```

Sources/StreamChat/Errors/ClientError.swift
```diff
@@ -19,10 +19,23 @@ public class ClientError: Error, CustomStringConvertible {
...
+            return message
+        }
+        
+        switch underlyingError {
+        case let clientError as ClientError:
+            return clientError.localizedDescription
+        case let errorPayload as ErrorPayload:
+            return errorPayload.message
+        case let engineError as WebSocketEngineError:
+            return engineError.reason
```
nuno-vieira, line 34: Is this actually correct? 🤔 From the previous commits, I remember that the websocket engine error was an underlying error of `errorPayload`?

Sources/StreamChat/WebSocketClient/WebSocketClient.swift
```diff
@@ -131,7 +131,7 @@ class WebSocketClient {
         completion: @escaping () -> Void
     ) {
         connectionState = .disconnecting(source: source)
-        engineQueue.async { [weak engine, eventsBatcher] in
+        engineQueue.async { [engine, eventsBatcher] in
```
nuno-vieira, line 134: Any reason why we are removing the `weak` here?

nuno-vieira (approved):
LGTM Overall ✅ Thanks Pavel for the improvements in the Demo App 👌 Awesome!

Just added some small questions and nits 🙂 But should be ready to 🚢  already 👍

Sources/StreamChat/Errors/ClientError.swift
evsaev: Yeah, here is the recursive case:
```swift
case let clientError as ClientError:
            return clientError.localizedDescription
```

When ws is trying to connect and fails with connection error, the nested structure is:
`ClientError.ConnectionNotSuccessful -> ClientError.WebSocket -> WebSocketEngineError -> Error`

So in case of the WS error, as a localized description, we'll take `WebSocketEngineError.reason`

Sources/StreamChat/Utils/InternetConnection/Error+InternetNotAvailable.swift
```diff
@@ -8,9 +8,23 @@ import Foundation
 /// to be extended as needed.
 private let offlineErrors: [(domain: String, errorCode: Int)] = [
     (NSURLErrorDomain, NSURLErrorNotConnectedToInternet),
+    (NSURLErrorDomain, NSURLErrorDataNotAllowed),
```
evsaev, line 11: The goal of this PR is to not show any alert if the user tries to connect offline. Each time I connect without the Internet, I receive this error. Strictly speaking, it's not an `offline` one but for some reason, URL session returns it instead of `NSURLErrorNotConnectedToInternet`.

Do you think we should include `NSURLErrorDataNotAllowed` to the offline list? @GetStream/ios-developers 

@GetStream/ios-developers

nuno-vieira: Discussing it on Slack 🙂

evsaev: @polqf @nuno-vieira updated the logic to show a banner. Please 👀 when you have time

Sources/StreamChatUI/Navigation/ChatChannelListRouter.swift
```diff
@@ -28,13 +28,14 @@ open class ChatChannelListRouter: NavigationRouter<ChatChannelListVC>, Component
...
-        }
         
-        navController.show(vc, sender: self)
+        if let splitVC = rootViewController.splitViewController {
+            splitVC.showDetailViewController(UINavigationController(rootViewController: vc), sender: self)
+        } else if let navigationVC = rootViewController.navigationController {
+            navigationVC.show(vc, sender: self)
+        } else {
+            rootViewController.show(UINavigationController(rootViewController: vc), sender: self)
+        }
```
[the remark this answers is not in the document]
nuno-vieira: @evsaev Why are we making this change as part of this PR? I'm afraid this might conflict with some customers' logic.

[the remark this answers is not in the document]
nuno-vieira: Another idea: If we are doing this, would it makes sense to set the `navigationController.transitionDelegate` to our own `StreamModalTransitioningDelegate`? (Check `DemoChatChannelListRouter`). We have a custom modal transition delegate to fix the Inverted message list

CHANGELOG.md
```diff
@@ -2,17 +2,19 @@
...
+## StreamChat
+### 🐞 Fixed
+- Saving payloads to local database is now 50% faster. Initial launch and displaying channel list should be noticeably faster [#1973](https://github.com/GetStream/stream-chat-swift/issues/1973)
+- Fix not waiting for last batch of events to be processed when connecting as another user [#2016](https://github.com/GetStream/stream-chat-swift/issues/2016)
+
 ## StreamChatUI
 ### ✅ Added
 - Add Support for Slow Mode [#1953](https://github.com/GetStream/stream-chat-swift/pull/1953)
+- Present channel screen modally when channel list in not embedded by navigation controller [#2011](https://github.com/GetStream/stream-chat-swift/pull/2011)
+- Show channel screen as right detail when channel list is embedded by split view controller [#2011](https://github.com/GetStream/stream-chat-swift/pull/2011)
```
nuno-vieira, line 14: I'm a bit concerned about this. `SplitViewController` has additional logic. Can we test these scenarios currently in the DemoApp? If not, how do we know they are working?

Sources/StreamChat/Utils/InternetConnection/Error+InternetNotAvailable.swift
nuno-vieira: Looks good 👌

nuno-vieira: Hum 🤔 The only problem now is "Connecting..." is displayed always, even if the connection is good. Why did we make this change? Wasn't the goal to just discard any change for now. If the connection is offline, we don't do nothing for now

DemoApp/BannerShowingConnectionDelegate.swift
```diff
@@ -34,6 +34,7 @@ extension BannerShowingConnectionDelegate: ChatConnectionControllerDelegate {
             bannerView.update(text: "Disconnecting...")
         case .connecting:
             bannerView.update(text: "Connecting...")
+            showBanner()
```
nuno-vieira, line 37: This always displays "Connecting..." when you login :/ Should we for now just discard this change and wait for channel list states implementation?

nuno-vieira (changes requested):
@evsaev These changes are breaking the split view in the Demo App. You can see here:

`This Branch`:

https://user-images.githubusercontent.com/12814114/170305004-1ccf5152-ebba-4538-9f63-1f5738052f51.mp4

`Develop`:

https://user-images.githubusercontent.com/12814114/170304696-3ea16480-51f0-4d84-b7ef-f3152ab0dcc7.mp4

You can see that the cell gets deselected + the UI feels very laggy when selecting new channels

Sources/StreamChatUI/Navigation/ChatChannelListRouter.swift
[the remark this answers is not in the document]
evsaev: > I'm not sure if we want to bring this logic to the SDK tho

With this, we cover all the scenarios of showing the channel list (in navigation/split/on its own) and execute correct navigation when the channel is opened.

Previously, if the channel list was not embedded into navigation, nothing was happening when channel is selected. We also saw how complex it was with split vc.

So for me it feels like a good change.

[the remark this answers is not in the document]
evsaev: > I'm afraid this might conflict with some customers' logic.

If customers have custom presentation logic (other then push), they currently have no other option but to override without calling `super`. So we are safe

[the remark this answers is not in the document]
evsaev: > Another idea: If we are doing this, would it makes sense to set the navigationController.transitionDelegate to our own StreamModalTransitioningDelegate? (Check DemoChatChannelListRouter). We have a custom modal transition delegate to fix the Inverted message list modal issue

Nice one, will add 👍

CHANGELOG.md
evsaev: I tested the demo on all size classes including iPhones Pro/Pro max transitioning from portrait (cr) to landscape (rc) when split appears. Also tried scenarios without split/navigation. I can add manual tests but I don't think it makes much sense for DemoApp

DemoApp/BannerShowingConnectionDelegate.swift
evsaev: > This always displays "Connecting..." when you login

Showing `Connecting...` when the user logs in online is expected. However, when launching the app offline, showing `connecting` when nothing really happens in LLC is confusing. That's probably smth we might wanna invest some time into (cc @polqf)

> Should we for now just discard this change and wait for channel list states implementation?
The user should at least be aware, that the chat is not connected. Showing `connecting` is better than nothing, IMO.  If we want better UX we could show the connection state in channel list navigation subtitle like Telegram to make the connection state look like some small bit of info but not a warning (that annoys by overlapping the channel list)

evsaev:
> @evsaev These changes are breaking the split view in the Demo App. You can see here:
> 
> `This Branch`:
> 
>  Simulator.Screen.Recording.-.iPad.Pro.9.7-inch.-.2022-05-25.at.16.48.39.mp4 
> 
> `Develop`:
> 
>  Simulator.Screen.Recording.-.iPad.Pro.9.7-inch.-.2022-05-25.at.16.50.23.mp4 
> 
> You can see that the cell gets deselected + the UI feels very laggy when selecting new channels

Good catch, I fixed the selection 👍

Sources/StreamChat/Utils/InternetConnection/Error+InternetNotAvailable.swift
evsaev: Responded [here](https://github.com/GetStream/stream-chat-swift/pull/2011#discussion_r882474905)

evsaev:
@bielikb @nuno-vieira would be really nice to find less hacky solution for how to keep the selection of the channel cell the user has tapped on.

Sources/StreamChatUI/Navigation/ChatChannelListRouter.swift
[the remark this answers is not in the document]
nuno-vieira: @evsaev The problem is that if they dig into the code and see the `split` implementation, they will think that the split is supported out of the box, and it is not **yet**. Currently, that Split implementation was added to the DemoApp to debug stuff only

nuno-vieira:
> But yeah, would be really nice to come up with a less hacky solution on how to keep the selection of the channel cell the user has tapped on and make it available out of the box
> 
> cc @bielikb

https://github.com/GetStream/stream-chat-swift/pull/2011#discussion_r882543788

The Split Implementation out of the box, should be out of the scope of this PR. We are already changing too much things in this PR already

Sources/StreamChatUI/Navigation/ChatChannelListRouter.swift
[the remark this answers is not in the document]
evsaev: Yeah, but we can say the same about the navigation controller. As of now, we require `ChatChannelListVC` to be embedded into the navigation controller **by the integrator**, otherwise, the channel is not opened. We explain that in docs and probably we'll need a section for split

DemoApp/BannerShowingConnectionDelegate.swift
nuno-vieira: Like we agreed on Slack, we would not do anything, not show any alert until we tackle the channel states. IMO I rather not have any feedback for now when your connection is Offline than have a "Conecting..." appearing every time I log in with a user.

nuno-vieira: The goal of this PR is to "When you open the channel list while offline, the data should be there, and should not present any alert". This is fine. It is just the demo app, I would not overthink it and make it perfect in this PR.

evsaev: We agreed to have a banner, didn't we? 🤔

nuno-vieira (approved):
LGTM! ✅

Sources/StreamChatUI/Navigation/ChatChannelListRouter.swift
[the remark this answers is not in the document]
evsaev: > The Split Implementation is out of the box and should be out of the scope of this PR. We are already changing too many things in this PR already. IMO we should keep the Split implementation a DemoApp only feature for now.

Demo is an integration example where we put ourselves into customers' shoes. In this case, refactoring has highlighted the complexity of working with split-view and other issues. Changes introduced in the router make it much simpler (before, it felt like fighting against the SDK).

If we refer to the current implementation, I actually think it needs an improvement 🙂 Like, why do we make an assumption that the channel list is embedded into navigation and freak out when it's not? We either leave `showChannel` empty and leave it to the customer to implement or handle all the common use-cases (split, navigation, none of it)

@GetStream/ios-developers what's your opinion?

PR #2015, opened 2022-05-18
[BW] Fix smoke-checks on push

alteral:
### 📝 Summary

Fixing workflow on `push` to `develop`. Currently, it fails because Github does not allow us to search for changed files on `push`.

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -786,7 +786,7 @@ lane :emerge_upload do
 end
 
 def testing_required?(type)
-  return true if ENV['GITHUB_EVENT_NAME'] !~ /(push|pull_request)/ || ENV['GITHUB_REF'].include?('main')
+  return true if ENV['GITHUB_EVENT_NAME'] != 'pull_request'
 
   sources = {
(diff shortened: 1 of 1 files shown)


fastlane/Fastfile
```diff
@@ -786,22 +786,17 @@ lane :emerge_upload do
 end
 
 def testing_required?(type)
```
evsaev, line 788: `type` input param is not used 🤔

fastlane/Fastfile
```diff
@@ -786,22 +786,17 @@ lane :emerge_upload do
...
-  return true if ENV['GITHUB_EVENT_NAME'] !~ /(push|pull_request)/ || ENV['GITHUB_REF'].include?('main')
+  return true if ENV['GITHUB_EVENT_NAME'] != 'pull_request'
```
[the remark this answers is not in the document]
evsaev: Never mind, it's just the early exit while I thought it was the entire func.

fastlane/Fastfile
alteral:
```ruby
sources[type.to_sym].any? { |required| path.start_with?(required) }
```

PR #2016, opened 2022-05-18
[CIS-1873] Wait for last events batch to be processed to disconnect

evsaev:
### 🔗 Issue Links

CIS-1873

### 🎯 Goal

Fix crash happening when the chat client is connected with another user. The issue was connected to the last batch of events from the previous user posted **after** the new user is connected. The crash's root cause were models held by events that no longer exist after database removal.

### 🛠 Implementation

Add `completion` to disconnection methods called after processing the last batch of events.

### ☑️ Contributor Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## StreamChat
 ### 🐞 Fixed
 - Saving payloads to local database is now 50% faster. Initial launch and displaying channel list should be noticeably faster [#1973](https://github.com/GetStream/stream-chat-swift/issues/1973)
+- Fix not waiting for last batch of events to be processed when connecting as another user [#2016](https://github.com/GetStream/stream-chat-swift/issues/2016)
 
 # [4.15.0](https://github.com/GetStream/stream-chat-swift/releases/tag/4.15.0)
 _May 11, 2022_
--- a/Sources/StreamChat/ChatClient.swift
+++ b/Sources/StreamChat/ChatClient.swift
@@ -611,6 +611,12 @@ extension ClientError {
             """
         }
     }
+    
+    public class ClientHasBeenDeallocated: ClientError {
+        override public var localizedDescription: String {
+            "ChatClient has beed deallocated, make sure to keep at least one strong reference to it."
+        }
+    }
 }
 
 /// `APIClient` listens for `WebSocketClient` connection updates so it can forward the current connection id to
--- a/Sources/StreamChat/ChatClient.swift
+++ b/Sources/StreamChat/ChatClient.swift
@@ -351,7 +351,9 @@ public class ChatClient {
     /// Disconnects the chat client from the chat servers. No further updates from the servers
     /// are received.
     public func disconnect() {
-        clientUpdater.disconnect()
+        clientUpdater.disconnect(source: .userInitiated) {
+            log.info("The `ChatClient` has been disconnected.", subsystems: .webSocket)
+        }
         userConnectionProvider = nil
     }
 
@@ -614,7 +616,7 @@ extension ClientError {
     
     public class ClientHasBeenDeallocated: ClientError {
         override public var localizedDescription: String {
-            "ChatClient has beed deallocated, make sure to keep at least one strong reference to it."
+            "ChatClient has been deallocated, make sure to keep at least one strong reference to it."
         }
     }
 }
--- a/Sources/StreamChat/Controllers/ConnectionController/ConnectionController.swift
+++ b/Sources/StreamChat/Controllers/ConnectionController/ConnectionController.swift
@@ -96,7 +96,9 @@ public extension ChatConnectionController {
     /// Disconnects the chat client the controller represents from the chat servers.
     /// No further updates from the servers are received.
     func disconnect() {
-        chatClientUpdater.disconnect()
+        chatClientUpdater.disconnect(source: .userInitiated) {
+            log.info("The `ChatClient` has been disconnected.", subsystems: .webSocket)
+        }
     }
 }
 
--- a/Sources/StreamChat/Utils/EventBatcher.swift
+++ b/Sources/StreamChat/Utils/EventBatcher.swift
@@ -7,12 +7,13 @@ import Foundation
 /// The type that does events batching.
 protocol EventBatcher {
     typealias Batch = [Event]
+    typealias BatchHandler = (_ batch: Batch, _ completion: @escaping () -> Void) -> Void
     
     /// The current batch of events.
     var currentBatch: Batch { get }
     
     /// Creates new batch processor.
-    init(period: TimeInterval, timerType: Timer.Type, handler: @escaping (Batch) -> Void)
+    init(period: TimeInterval, timerType: Timer.Type, handler: @escaping BatchHandler)
     
     /// Adds the item to the current batch of events. If it's the first event also schedules batch processing
     /// that will happen when `period` has passed.
@@ -21,7 +22,7 @@ protocol EventBatcher {
     func append(_ event: Event)
     
     /// Ignores `period` and passes the current batch of events to handler as soon as possible.
-    func processImmediately()
+    func processImmediately(completion: @escaping () -> Void)
 }
 
 extension Batcher: EventBatcher where Item == Event {}
@@ -34,7 +35,7 @@ final class Batcher<Item> {
     /// The timer that  calls `processor` when fired.
     private var batchProcessingTimer: TimerControl?
     /// The closure which processes the batch.
-    private let handler: ([Item]) -> Void
+    private let handler: (_ batch: [Item], _ completion: @escaping () -> Void) -> Void
     /// The serial queue where item appends and batch processing is happening on.
     private let queue = DispatchQueue(label: "io.getstream.Batch.\(Item.self)")
     /// The current batch of items.
@@ -43,7 +44,7 @@ final class Batcher<Item> {
     init(
         period: TimeInterval,
         timerType: Timer.Type = DefaultTimer.self,
-        handler: @escaping ([Item]) -> Void
+        handler: @escaping (_ batch: [Item], _ completion: @escaping () -> Void) -> Void
     ) {
         self.period = max(period, 0)
         self.timerType = timerType
@@ -59,19 +60,19 @@ final class Batcher<Item> {
             self.batchProcessingTimer = self.timerType.schedule(
                 timeInterval: self.period,
                 queue: self.queue,
-                onFire: self.process
+                onFire: { self.process() }
             )
         }
     }
     
-    func processImmediately() {
+    func processImmediately(completion: @escaping () -> Void) {
         timerType.schedule(timeInterval: 0, queue: queue) { [weak self] in
-            self?.process()
+            self?.process(completion: completion)
         }
     }
     
-    private func process() {
-        handler(currentBatch)
+    private func process(completion: (() -> Void)? = nil) {
+        handler(currentBatch) { completion?() }
         currentBatch.removeAll()
         batchProcessingTimer?.cancel()
         batchProcessingTimer = nil
--- a/Sources/StreamChat/Workers/ChatClientUpdater.swift
+++ b/Sources/StreamChat/Workers/ChatClientUpdater.swift
@@ -48,22 +48,29 @@ class ChatClientUpdater {
             client.currentToken = newToken
 
             // Disconnect from web-socket.
-            disconnect(source: .userInitiated)
-            
-            // Update web-socket endpoint.
-            client.webSocketClient?.connectEndpoint = .webSocketConnect(
-                userInfo: userInfo ?? .init(id: newToken.userId)
-            )
+            disconnect(source: .userInitiated) { [weak client] in
+                guard let client = client else {
+                    completion(ClientError.ClientHasBeenDeallocated())
+                    return
+                }
+                
+                // Update web-socket endpoint.
+                client.webSocketClient?.connectEndpoint = .webSocketConnect(
+                    userInfo: userInfo ?? .init(id: newToken.userId)
+                )
 
-            // Re-create backgroundWorker's since they are related to the previous user.
-            client.createBackgroundWorkers()
+                // Re-create backgroundWorker's since they are related to the previous user.
+                client.createBackgroundWorkers()
+                
+                // Stop tracking active components
+                client.activeChannelControllers.removeAllObjects()
+                client.activeChannelListControllers.removeAllObjects()
+                
+                // Reset all existing local data.
+                client.databaseContainer.removeAllData(force: true, completion: completion)
+            }
             
-            // Stop tracking active components
-            client.activeChannelControllers.removeAllObjects()
-            client.activeChannelListControllers.removeAllObjects()
-            
-            // Reset all existing local data.
-            return client.databaseContainer.removeAllData(force: true, completion: completion)
+            return
         }
 
         // Set the web-socket endpoint
@@ -179,13 +186,17 @@ class ChatClientUpdater {
 
     /// Disconnects the chat client the controller represents from the chat servers. No further updates from the servers
     /// are received.
-    func disconnect(source: WebSocketConnectionState.DisconnectionSource = .userInitiated) {
+    func disconnect(
+        source: WebSocketConnectionState.DisconnectionSource = .userInitiated,
+        completion: @escaping () -> Void
+    ) {
         client.apiClient.flushRequestsQueue()
         client.syncRepository.cancelRecoveryFlow()
         
         // Disconnecting is not possible in connectionless mode (duh)
         guard client.config.isClientInActiveMode else {
             log.error(ClientError.ClientIsNotInActiveMode().localizedDescription)
+            completion()
             return
         }
 
@@ -193,17 +204,20 @@ class ChatClientUpdater {
             if source == .userInitiated {
                 log.warning("The client is already disconnected. Skipping the `disconnect` call.")
             }
+            completion()
             return
         }
 
         // Disconnect the web socket
-        client.webSocketClient?.disconnect(source: source)
+        client.webSocketClient?.disconnect(source: source) { [weak client] in
+            // Reset `connectionId`. This would happen asynchronously by the callback from WebSocketClient anyway, but it's
+            // safer to do it here synchronously to immediately stop all API calls.
+            client?.connectionId = nil
 
-        // Reset `connectionId`. This would happen asynchronously by the callback from WebSocketClient anyway, but it's
-        // safer to do it here synchronously to immediately stop all API calls.
-        client.connectionId = nil
-
-        // Remove all waiters for connectionId
-        client.completeConnectionIdWaiters(connectionId: nil)
+            // Remove all waiters for connectionId
+            client?.completeConnectionIdWaiters(connectionId: nil)
+            
+            completion()
+        }
     }
 }
--- a/Tests/StreamChatTests/StreamChat/ChatClient_Tests.swift
+++ b/Tests/StreamChatTests/StreamChat/ChatClient_Tests.swift
@@ -568,6 +568,8 @@ final class ChatClient_Tests: XCTestCase {
         XCTAssertTrue(testEnv.apiClient! === client.apiClient)
         // Assert `disconnect` on updater is triggered
         XCTAssertTrue(testEnv.clientUpdater!.disconnect_called)
+        // Assert source is user initiated
+        XCTAssertEqual(testEnv.clientUpdater!.disconnect_source, .userInitiated)
     }
     
     // MARK: - Background workers tests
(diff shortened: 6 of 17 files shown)


Sources/StreamChat/Controllers/ConnectionController/ConnectionController.swift
```diff
@@ -92,11 +92,15 @@ public extension ChatConnectionController {
...
-    /// No further updates from the servers are received.
-    func disconnect() {
-        chatClientUpdater.disconnect()
+    
+    /// Stops active web-socket connection for the underlaying `ChatClient` and
+    /// cancels all queued API requests. If there is a connection recovery flow running, it will be interrupted.
+    ///
+    /// - Parameter completion: The completion handler invoked when the `ChatClient` has been disconnected.
+    func disconnect(completion: (() -> Void)? = nil) {
+        chatClientUpdater.disconnect {
```
evsaev, line 101: @polqf passing `completion` to `chatClientUpdater.disconnect` is not possible since the closure is optional here but updater requires a non-optional one.

What we could is:
```swift
chatClientUpdater.disconnect(completion: completion ?? {})
```

Sources/StreamChat/Utils/EventBatcher.swift
```diff
@@ -59,19 +60,19 @@ final class Batcher<Item> {
...
+    func processImmediately(completion: @escaping () -> Void) {
         timerType.schedule(timeInterval: 0, queue: queue) { [weak self] in
-            self?.process()
+            self?.process(completion: completion)
         }
     }
     
-    private func process() {
-        handler(currentBatch)
+    private func process(completion: (() -> Void)? = nil) {
```
evsaev, line 74: @polqf made `completion` optional here so we don't need to pass it when the timer fires

Sources/StreamChat/Workers/ChatClientUpdater.swift
```diff
@@ -179,31 +186,43 @@ class ChatClientUpdater {
...
+        source: WebSocketConnectionState.DisconnectionSource = .userInitiated,
+        completion: @escaping () -> Void
+    ) {
         client.apiClient.flushRequestsQueue()
         client.syncRepository.cancelRecoveryFlow()
         
         // Disconnecting is not possible in connectionless mode (duh)
         guard client.config.isClientInActiveMode else {
             log.error(ClientError.ClientIsNotInActiveMode().localizedDescription)
+            completion()
```
evsaev, line 199: @polqf addressed your comment about calling `completion` on a separate line before the return.

Sources/StreamChatUI/ChatMessageList/ChatMessageListView.swift
```diff
@@ -178,7 +178,7 @@ open class ChatMessageListView: UITableView, Customizable, ComponentsProvider {
         }
 
         switch change {
-        case let .insert(message, index: index):
+        case let .insert(_, index: index):
```
nuno-vieira, line 181: 😨 Did we actually release 4.15.0 with this warning?

Sources/StreamChat/ChatClient.swift
```diff
@@ -348,10 +348,12 @@ public class ChatClient {
...
-    /// Disconnects the chat client from the chat servers. No further updates from the servers
-    /// are received.
-    public func disconnect() {
-        clientUpdater.disconnect()
+    /// Stops active web-socket connection and cancels all queued API requests.
+    /// If there is a connection recovery flow running, it will be interrupted.
+    ///
+    /// - Parameter completion: The completion handler invoked when the `ChatClient` has been disconnected.
+    public func disconnect(completion: (() -> Void)? = nil) {
+        clientUpdater.disconnect { completion?() }
```
nuno-vieira, line 356: Do we need to expose this completion to the `ChatClient`? How would this be useful for the customers?

nuno-vieira (commented):
Overall looks good, just added a couple of questions 👍

Sources/StreamChatUI/ChatMessageList/ChatMessageListView.swift
evsaev: Yep, we did 😞 We need to add lint phase to Allure to prevent it from happening in the future

Sources/StreamChat/ChatClient.swift
evsaev: I don't have a concrete use-case in mind but someone might be interested when the disconnection process (or `logout` that we've discussed recently) is completed. Show loaded that the disconnection is happening/keeping the chat screen opened till chat-client is disconnected/sending a request to the backend/tracking an analytics event). It does not cost us anything to expose it so why not to have it?

Sources/StreamChat/Controllers/ConnectionController/ConnectionController.swift
evsaev: Sonar cloud treats `completion ?? {}` as a code smell. Let's keep the current version

Sources/StreamChat/ChatClient.swift
nuno-vieira: > Show spinner when the disconnection is happening/keeping the chat screen opened till chat-client

Why would customers do this? Usually, disconnection is a fire and forget action.

> It does not cost us anything to expose it so why not have it?

It will cause confusion to the customers. They will think "Should I actually wait for the disconnection to be completed?", "What if I don't wait, what is the problem?". 

IMO, I think, for now, we should not add this completion. It will add confusion to the customers, and in the future we might not even need this completion at all. So for this reasons, I would not introduce this. If in the future we encounter a valid use case, or a customer wants this, we can introduce it.

evsaev: > if we add it now, we can't remove it

But why would we limit an integrator to know when disconnection is completed? The param is `nil` by default so if they are not interested in completion they choose not to pass it

Screenshot 2022-05-19 at 14 51 06

evsaev: We have optional `completion` in every async method in SDK so I wouldn't make that one an exception 🙂

nuno-vieira: I think we can’t compare this one with the other methods you are describing. Because on the other ones we complete with an Error, so that is a valid use case to wait for it to complete and make an action if there is an error. In this case, disconnect calls usually are fire and forget. Adding a completion will only add confusion to the customers. There's no reason they should wait for a disconnect call to finish. Imagine closing a Chat UI, and waiting a little bit before it closes, this doesn't make sense.

We need to think if there is actually a valid case for this completion. If there isn't, we shouldn't add a new API that we can't remove it in the future. IMO we should try to minimise the risks here. We can always add a new API in the future, but we can't remove it. 

I have one question tho: What happens if you disconnect, and connect very fast. Is there any problem with this? Can a user connect before the disconnect finishes? 

CC @GetStream/ios-developers

b-onc: PR description:
>Fix crash happening when the chat client is connected with another user. The issue was connected to the last batch of events from the previous user posted after the new user is connected. The crash's root cause were models held by events that no longer exist after database removal.

so @nuno-vieira 
> I have one question tho: What happens if you disconnect, and connect very fast. Is there any problem with this? Can a user connect before the disconnect finishes?

it's a race condition, and can cause a crash

nuno-vieira: @evsaev ☝️ Is this true? If so, we are not fixing anything, we are leaving it to the customer to fix it himself and we should clarify this in the changelog

evsaev: @nuno-vieira we did fix the race condition. When the chat client is reconnected with another user, we wait for disconnection to happen before establishing the new ws connection for the new user

CHANGELOG.md
```diff
@@ -3,7 +3,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 # Upcoming
 
-### 🔄 Changed
+## StreamChat
+### ✅ Added
+- Add `completion` to disconnect APIs [#2016](https://github.com/GetStream/stream-chat-swift/issues/2016)
+### ✅ Fixed
+- Fix not waiting for last batch of events to be processed when connecting as another user [#2016](https://github.com/GetStream/stream-chat-swift/issues/2016)
```
nuno-vieira, line 10: Did we fix it out-of-the-box, or does the customer need to do anything about it? @evsaev

Sources/StreamChat/ChatClient.swift
evsaev: see [here](https://github.com/GetStream/stream-chat-swift/blob/47f6a629021eccd1c7854565342aa8187ab2671f/Sources/StreamChat/Workers/ChatClientUpdater.swift#L51)

nuno-vieira: Yes, that is what I understood from the PR, just wanted to double-check. CC @b-onc 

Then, I vote for removing the `completion` then. I don't see any valid use case for this, it will just confuse the customers. In this case I think we should follow YAGNI principle 😄

b-onc: I've meant before the PR, sorry for confusion.

I vote for adding the completion, since YAGNI doesn't apply to us - there's no single Y (you) for an SDK. customer 1 might not need it, customer 2 might.

nuno-vieira: Another valid point: No other SDK has this callback in the disconnect call. This is an implementation detail of our SDK and we need this internally to fix problems in our SDK. The disconnect call should be a fire and forget action, just like all the other platfroms.

CHANGELOG.md
```diff
@@ -3,7 +3,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 # Upcoming
 
-### 🔄 Changed
+## StreamChat
+### ✅ Added
+- Add `completion` to disconnect APIs [#2016](https://github.com/GetStream/stream-chat-swift/issues/2016)
+### ✅ Fixed
```
polqf, line 9: Seems like the icon does not match here

Sources/StreamChat/ChatClient.swift
```diff
@@ -611,6 +613,12 @@ extension ClientError {
             """
         }
     }
+    
+    public class ClientHasBeenDeallocated: ClientError {
+        override public var localizedDescription: String {
+            "ChatClient has beed deallocated, make sure to keep at least one strong reference to it."
```
polqf, line 619:
```suggestion
            "ChatClient has been deallocated, make sure to keep at least one strong reference to it."
```

Sources/StreamChat/Utils/EventBatcher.swift
```diff
@@ -34,7 +35,7 @@ final class Batcher<Item> {
     /// The timer that  calls `processor` when fired.
     private var batchProcessingTimer: TimerControl?
     /// The closure which processes the batch.
-    private let handler: ([Item]) -> Void
+    private let handler: (_ batch: [Item], _ completion: @escaping () -> Void) -> Void
```
polqf, line 38:
```suggestion
    private let handler: BatchHandler
```

Sources/StreamChat/Utils/EventBatcher.swift
```diff
@@ -43,7 +44,7 @@ final class Batcher<Item> {
     init(
         period: TimeInterval,
         timerType: Timer.Type = DefaultTimer.self,
-        handler: @escaping ([Item]) -> Void
+        handler: @escaping (_ batch: [Item], _ completion: @escaping () -> Void) -> Void
```
polqf, line 47:
```suggestion
        handler: @escaping BatchHandler
```

I'm actually not sure if this (👆) compiles 🤔

Tests/StreamChatTests/StreamChat/ChatClient_Tests.swift
```diff
@@ -570,6 +570,33 @@ final class ChatClient_Tests: XCTestCase {
...
     
+    func test_disconnect_callsCompletion() {
+        // Create a chat client
+        let client = ChatClient(
+            config: inMemoryStorageConfig,
+            environment: testEnv.environment
+        )
+        
+        // Disconnect chat client
+        var completionCalled = false
```
polqf, line 581: Better use expectations 🙏

Tests/StreamChatTests/Utils/EventBatcher_Tests.swift
```diff
@@ -83,15 +85,24 @@ final class Batch_Tests: XCTestCase {
         batcher.append(event)
         
         // Ask to process immidiately
-        batcher.processImmediately()
+        var completionCalled = false
```
polqf, line 88: same as before

Tests/StreamChatTests/WebSocketClient/WebSocketClient_Tests.swift
```diff
@@ -500,9 +501,19 @@ final class WebSocketClient_Tests: XCTestCase {
         XCTAssertFalse(eventsBatcher.mock_processImmediately.called)
         
         // Simulate disconnection
-        webSocketClient.disconnect()
+        var completionCalled = false
```
polqf, line 504: same

Tests/StreamChatTests/Workers/ChatClientUpdater_Tests.swift
```diff
@@ -17,7 +17,10 @@ final class ChatClientUpdater_Tests: XCTestCase {
         let updater = ChatClientUpdater(client: client)
 
         // Simulate `disconnect` call.
-        updater.disconnect()
+        var completionCalled = false
```
polqf, line 20: same here and the rest of the file

Sources/StreamChat/ChatClient.swift
evsaev: Changed ✅

CHANGELOG.md
evsaev: Fixed ✅

Sources/StreamChat/ChatClient.swift
evsaev: Fixed ✅

Sources/StreamChat/Utils/EventBatcher.swift
evsaev: Yeah, tried that but it does not compile because the of `Batcher` type being generic

Sources/StreamChat/Utils/EventBatcher.swift
evsaev: Yeah, it does not, see the comment above

Tests/StreamChatTests/StreamChat/ChatClient_Tests.swift
evsaev: Updated ✅

Tests/StreamChatTests/Utils/EventBatcher_Tests.swift
evsaev: Updated ✅

Tests/StreamChatTests/WebSocketClient/WebSocketClient_Tests.swift
evsaev: Updated ✅

Tests/StreamChatTests/Workers/ChatClientUpdater_Tests.swift
evsaev: Updated ✅

CHANGELOG.md
```diff
@@ -4,8 +4,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 # Upcoming
 
 ## StreamChat
+### ✅ Added
+- Add `completion` to disconnect APIs [#2016](https://github.com/GetStream/stream-chat-swift/issues/2016)
```
nuno-vieira, line 8: It looks like we actually removed this. Can we remove this changelog? 🤔

Tests/StreamChatTests/WebSocketClient/WebSocketClient_Tests.swift
```diff
@@ -500,9 +501,18 @@ final class WebSocketClient_Tests: XCTestCase {
...
+        }
         
         // Assert `processImmediately` is triggered
         AssertAsync.willBeTrue(eventsBatcher.mock_processImmediately.called)
+        
+        // Simulate batch processing completion
+        eventsBatcher.mock_processImmediately.calls.last?()
+        
+        // Assert completion called
+        wait(for: [expectation], timeout: 0)
```
nuno-vieira, line 516: We need to provide a timeout. Depending on the load in the CI, this might take more time.
```suggestion
        wait(for: [expectation], timeout: 0.5)
```

Tests/StreamChatTests/Utils/EventBatcher_Tests.swift
```diff
@@ -83,15 +85,22 @@ final class Batch_Tests: XCTestCase {
...
         
         // Assert handler is called sooner
         XCTAssertEqual(handlerCalls, [[event]])
+
+        // Complete batch processing
+        handlerCompletion?()
         
         // Assert current batch is empty
         XCTAssertEqual(batcher.currentBatch, [])
+        wait(for: [expectation], timeout: 0)
```
nuno-vieira, line 104: same here

Tests/StreamChatTests/Workers/ChatClientUpdater_Tests.swift
```diff
@@ -26,6 +29,9 @@ final class ChatClientUpdater_Tests: XCTestCase {
         XCTAssertCall("flushRequestsQueue()", on: client.mockAPIClient, times: 1)
         // Assert recovery flow is cancelled.
         XCTAssertCall("cancelRecoveryFlow()", on: client.mockSyncRepository, times: 1)
+        
+        // Assert completion called
+        wait(for: [expectation], timeout: 0)
```
nuno-vieira, line 34: same here

Tests/StreamChatTests/Workers/ChatClientUpdater_Tests.swift
```diff
@@ -36,19 +42,28 @@ final class ChatClientUpdater_Tests: XCTestCase {
...
+        // Assert `flushRequestsQueue` is called on API client.
+        XCTAssertCall("flushRequestsQueue()", on: client.mockAPIClient, times: 1)
+        // Assert recovery flow is cancelled.
+        XCTAssertCall("cancelRecoveryFlow()", on: client.mockSyncRepository, times: 1)
+        
+        // Simulate completed disconnection
+        client.mockWebSocketClient.disconnect_completion!()
+        
+        // Assert completion is called
+        wait(for: [expectation], timeout: 0)
```
nuno-vieira, line 61: same here

Tests/StreamChatTests/Workers/ChatClientUpdater_Tests.swift
```diff
@@ -59,22 +74,28 @@ final class ChatClientUpdater_Tests: XCTestCase {
...
+        }
 
         // Assert `connect` was not called on `webSocketClient`.
         XCTAssertEqual(client.mockWebSocketClient.disconnect_calledCounter, 0)
         // Assert `flushRequestsQueue` is called on API client.
         XCTAssertCall("flushRequestsQueue()", on: client.mockAPIClient, times: 1)
         // Assert recovery flow is cancelled.
         XCTAssertCall("cancelRecoveryFlow()", on: client.mockSyncRepository, times: 1)
+        // Assert completion is called
+        wait(for: [expectation], timeout: 0)
```
nuno-vieira, line 98: same here

Sources/StreamChat/ChatClient.swift
```diff
@@ -348,10 +348,12 @@ public class ChatClient {
         )
     }
     
-    /// Disconnects the chat client from the chat servers. No further updates from the servers
-    /// are received.
+    /// Stops active web-socket connection and cancels all queued API requests.
+    /// If there is a connection recovery flow running, it will be interrupted.
```
nuno-vieira, line 352: I actually think the previous documentation was better. What the customer needs to know is that by calling `disconnect()` he will stop getting updates from the server. The changed documentation is describing on what happens internally. This documentation is good for `ConnectionController` but it should be in the public API.

Tests/StreamChatTests/WebSocketClient/WebSocketClient_Tests.swift
```diff
@@ -99,7 +99,7 @@ final class WebSocketClient_Tests: XCTestCase {
         // Save currently existed engine.
         let oldEngine = webSocketClient.engine
         // Disconnect the client.
-        webSocketClient.disconnect()
+        webSocketClient.disconnect {}
```
nuno-vieira, line 102: If this is async now, don't we need to wait for an expectation here as well? 🤔

Tests/StreamChatTests/WebSocketClient/WebSocketClient_Tests.swift
evsaev: The `completion` is synchronously called on the previous line `eventsBatcher.mock_processImmediately.calls.last?()` so the expectation has been already fulfilled at this point.

The same is valued for all other places where `timeout: 0` is used

CHANGELOG.md
evsaev: Fixed

Tests/StreamChatTests/WebSocketClient/WebSocketClient_Tests.swift
evsaev: It's fine, what matters in this test is 2nd `connect` call which should not recreate the engine

bielikb (approved):

PR #2022, opened 2022-05-25
[BW] Fix flaky UI test for Slow mode.

hugobernalstream:
### 🎯 Goal

Fix flaky UI test for Slow mode.

### 🛠 Implementation

Increase duration time for cooldown so UI test can consistently validate cooldown view is shown when UI tests are running on CI.

### ☑️ Contributor Checklist

### DIFF
--- a/StreamChatUITestsAppUITests/Tests/SlowMode_Tests.swift
+++ b/StreamChatUITestsAppUITests/Tests/SlowMode_Tests.swift
@@ -20,7 +20,7 @@ final class SlowMode_Tests: StreamTestCase {
         linkToScenario(withId: 186)
 
         GIVEN("user opens a channel") {
-            backendRobot.setCooldown(enabled: true, duration: 3)
+            backendRobot.setCooldown(enabled: true, duration: 10)
             userRobot
                 .login()
                 .openChannel()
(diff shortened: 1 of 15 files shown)


StreamChatUITestsAppUITests/Tests/SlowMode_Tests.swift
```diff
@@ -20,7 +20,7 @@ final class SlowMode_Tests: StreamTestCase {
         linkToScenario(withId: 186)
 
         GIVEN("user opens a channel") {
-            backendRobot.setCooldown(enabled: true, duration: 3)
```
nuno-vieira, line 23: We probably should increate all tests which use duration "3". Since the failure scenarios might always pass but it could mean the cooldown is already finished. Even better would be to extract the duration to a constant in the test, and use it for every test to keep it consistent 👍

hugobernalstream: Done

adamrushy: The change looks good, although the value of slow tests vs having UI tests for this feature is potentially questionable... especially as our UI tests grow in time.

adamrushy (approved):

StreamChatUITestsAppUITests/Tests/SlowMode_Tests.swift
bielikb: @adamrushy are you concerned about the size of our UI test suite?
We leverage parallelisation on GH level, so that engineers are not required to question if +1 automated UI scenario might prolong the CI running times. 

The rule of thumb is to have the features covered by UI tests.

nuno-vieira (approved):
LGTM! ✅

PR #2027, opened 2022-05-27
[CIS-1901] Use NSDate in DTO to avoid crashes on deleted objects

polqf:
### 🔗 Issue Links

https://stream-io.atlassian.net/browse/CIS-1901

### 🎯 Goal

Fix `Date._unconditionallyBridgeFromObjectiveC(NSDate?)` crash

### 📝 Summary

We've been chasing this crash for a while now. A recent PR focused on increasing safety in our `NSManagedObject`s improved the situation but did not fix the crash for good.
After a nice investigation from @nuno-vieira , he discovered that the real root of this issue is that Core Data does not play nicely with models that use Date instead of NSDate. It is internally using the ObjC representation for a date, and it tries to bridge from one type to another. 
Whenever the DTO is deleted, all the properties are filled with default values. All, but `Date` values. Core Data is unable to nicely do the bridging, and this leads to a crash.  

### 🛠 Implementation

This PR introduces extensions over Date and NSDate to map from the ObjC type to the Swift one, and vice-versa.
It is **intentional** that the variables are named the same, reducing the complexity of these mappings.

It also introduces a typealias, `DBDate`, so that we don't easily fall in the trap of changing NSDate to Date as we see those on the code. In the definition of this typealias, we also add the necessary documentation so that there's a logical explanation to it.

As per performance, the mappings from Date <-> NSDate should be super lightweight, as stated here:
https://github.com/apple/swift-evolution/blob/master/proposals/0069-swift-mutability-for-foundation.md#from-objective-c-to-swift

### 🧪 Manual Testing Notes

### ☑️ Contributor Checklist

### 🎁 Meme


Tests/StreamChatTestTools/Extensions/Unique/Date+Unique.swift
```diff
@@ -18,3 +18,13 @@ extension Date {
...
+
+extension NSDate {
+    static var unique: NSDate {
+        Date.unique as NSDate
+    }
+
+    static func unique(after date: NSDate) -> NSDate {
+        Date.unique(after: date as Date) as NSDate
+    }
+}
```
evsaev, line 30: nit: I would go with typealias here so if we decide to remove `DBDate` we get a compilation error in test target and won't forget to remove this leftover
```suggestion
extension DBDate {
    static var unique: DBDate {
        Date.unique.bridgedDate
    }

    static func unique(after date: DBDate) -> DBDate {
        Date.unique(after: date.bridgedDate).bridgedDate
    }
}
```

Tests/StreamChatTests/Database/DTOs/ChannelDTO_Tests.swift
```diff
@@ -487,7 +487,7 @@ final class ChannelDTO_Tests: XCTestCase {
             // Truncate the channel to leave only 10 newest messages
             // We're dropping 9 messages to fullfil the predicate: createdAt >= channel.truncatedAt"
             let truncateDate = channelDTO.messages
-                .sorted(by: { $0.createdAt < $1.createdAt })
+                .sorted(by: { $0.createdAt.bridgeDate < $1.createdAt.bridgeDate })
```
evsaev, line 490:
```suggestion
                .sorted(by: { $0.createdAt < $1.createdAt })
```

evsaev (commented):
LGTM! I'll approve when the PR is converted from draft and missing bits like docs & lint rule are in place 👍

Sources/StreamChat/Database/DTOs/DBDate.swift
```diff
@@ -0,0 +1,14 @@
+//
+// Copyright © 2022 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+typealias DBDate = NSDate
+extension DBDate {
+    var bridgeDate: Date { self as Date }
```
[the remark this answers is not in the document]
evsaev: Ouch, that is a very good point.

Speaking about the fallback date, maybe `Date(timeIntervalSince1970: 0)/Date.distantPast` would better indicate that the date is invalid?

Tests/StreamChatTests/Database/DTOs/ChannelDTO_Tests.swift
polqf: NSDate is not comparable, so converting both to Date makes sense

Sources/StreamChat/Database/DTOs/DBDate.swift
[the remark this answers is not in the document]
polqf: Date(timeIntervalSince1970: timeIntervalSince1970) makes more sense, will be adding documentation

.swiftlint.yml
```diff
@@ -86,3 +86,7 @@ custom_rules:
     match_kinds:
       - identifier
     message: "Plain constraints methods are forbidden. Use `pin(...)` alternative"
+  coredata_date_forbidden:
+    included: "Sources/StreamChat/Database/DTOs"
+    regex: "@NSManaged(| \\S*)* var \\S*(: Date|: NSDate)"
```
polqf, line 91: Matches all these variants:
```
    @NSManaged var test1: Date
    @NSManaged private(set) var test2: NSDate
    @NSManaged public private(set) var test23: NSDate
```

.swiftlint.yml
```diff
@@ -86,3 +86,7 @@ custom_rules:
     match_kinds:
       - identifier
     message: "Plain constraints methods are forbidden. Use `pin(...)` alternative"
+  coredata_date_forbidden:
+    included: "Sources/StreamChat/Database/DTOs"
+    regex: "@NSManaged(| \\S*)* var \\S*: (NS)?Date"
```
polqf, line 91: Matches all these variants:
```
    @NSManaged var test1: Date
    @NSManaged private(set) var test2: NSDate
    @NSManaged public private(set) var test23: NSDate
```

evsaev (approved):
🚢

.swiftlint.yml
nuno-vieira: @polqf Do we need to included paths? I'm concerned that if we change the folder structure of the SDK this stops working. Can we simply discard this, or maybe limit files that end in "DTO.swift" ? 🤔

Sources/StreamChat/Database/DTOs/ChannelConfigDTO.swift
```diff
@@ -41,8 +41,8 @@ final class ChannelConfigDTO: NSManagedObject {
             commands: Array(Set(
                 commands.compactMap { try? ($0 as? CommandDTO)?.asModel() }
             )),
-            createdAt: createdAt,
-            updatedAt: updatedAt
+            createdAt: createdAt.bridgeDate,
```
nuno-vieira, line 44: I'm fine with `bridgeDate` but IMO `toDate()` or `asDate()` would be more consistent with the rest of the SDK + it would be easier to understand and use 🤔

.swiftlint.yml
polqf: I did it to avoid having an expensive Regex running through the whole code 🤔
I guess it is a trade off, not sure which one I prefer 🙃 thoughts?

nuno-vieira (approved):
LGTM! ✅ Added some small comments and suggestions, but not blocking for sure 👍

.swiftlint.yml
nuno-vieira: Let's keep the current one, for now, 👍 I think it is fine 🙂 We can change this later if we see any problems with it

PR #2028, opened 2022-05-27
[CIS-1481] - Changing the Role Decoding Key to be Channel_Role

adamrushy:
### 🔗 Issue Links

CIS-1481

### 🎯 Goal

We have implemented V2 of the permissions within the backend of Stream, which means role for a channel is now deprecated. We are receiving the same permissions as before but allowing custom permissions to be parsed.

### 📝 Summary

- I have added the decoding key to be channel_role
- Added more tests to cover custom roles.

### 🛠 Implementation

The implementation is rather simple, decode role with channel_role because we already cover the roles inside the enum but also include custom roles as well.

### 🧪 Manual Testing Notes

To test this feature you'll need to enable V2 permissions on your own Stream application. You can then update a users permissions and look for it to be parsed within the Demo app.

### ☑️ Contributor Checklist


Tests/StreamChatTestTools/Mocks/MockEndpointResponses/Member.json
```diff
@@ -5,7 +5,7 @@
   "shadow_banned" : true,
   "ban_expires" : "2021-03-08T15:42:31.355923Z",
   "user_id" : "broken-waterfall-5",
-  "role" : "owner",
+  "channel_role" : "custom_role",
```
evsaev, line 8: nit: what if we change only the JSON key here? This would keep the existed test untouched and make sure it keeps working. 

Then the new JSON (`Member+CustomRole`) can contain the custom role + the new test would verify custom roles are also successfully decoded.
```suggestion
  "channel_role" : "owner",
```

evsaev (approved):
Looks good! 🚢 Just one suggestion

Tests/StreamChatTestTools/Mocks/MockEndpointResponses/Member.json
adamrushy: Ah! I thought I changed this back, yep this makes sense - let me push the change

bielikb (approved):

polqf (approved):

davefoxy:
@adamrushy I think you're missing this change from here too? https://github.com/GetStream/stream-chat-swift/blob/develop/Sources/StreamChat/APIClient/Endpoints/Payloads/UserPayloads.swift#L11

davefoxy:
@adamrushy  Ah yes, you're right. Apologies 🙇

adamrushy:
@davefoxy no problem at all, thanks again for the initial find 👍🏼

PR #2031, opened 2022-05-27
[CIS-1185] Token provider

evsaev:
### 🔗 Issue Links

CIS-1185

### 🎯 Goal

Make it possible to pass a `tokenProvider` when connecting the chat-client.

### 📝 Summary

1. chat client has `connect` method that takes a token provider as an argument (the same as on other platforms)
1. chat client still has `connect` method that requires a token but with the note that a token must not expire
1. chat client no longer have `tokenProvider` exposed

⚠️ The **3rd** change is a **breaking** one but it might be OK to ship it: 
- customers who is using **non-expiring** tokens will not be affected by this change
- customers who is using **expiring** and a `tokenProvider` that was removed would be glad to hear that the whole flow is simplified:

**Before**:
```swift
// 1. Get a user
let user = 

// 2. Create a token provider
let tokenProvider: TokenProvider = { completion in
   authService.fetchToken(for: user, completion: completion)
}

// 3. Call the token provider to receive the 1st token
tokenProvider { tokenResult in
 guard case .success(let token) = tokenResult else { return }

 // 4. Create and connect the chat-client
 let chatClient: ChatClient = ...
 chatClient.tokenProvider = tokenProvider

 OR

 let chatClient = ChatClient(config, tokenProvider: tokenProvider)

 // 5. Connect the chat client with received token
 chatClient.connect(userInfo: user, token: token, completion: { /*handle connection error*/ })
}
```

**After**:
```swift
// 1. Get a user
let user: UserInfo = 

// 2. Create a token provider
let tokenProvider: TokenProvider = { completion in
   authService.fetchToken(for: user, completion: completion)
}

// 3. Create and connect chat-client with a token provider
let chatClient: ChatClient = ...
chatClient.connect(userInfo: user, tokenProvider: tokenProvider, completion: { /*handle connection error*/ })
```

### 🧪 Manual Testing Notes

Sign in to the `DemoApp` as a **normal/guest/anon** user and make sure the connection flow works as before.

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/ChatClient.swift
+++ b/Sources/StreamChat/ChatClient.swift
@@ -227,10 +227,7 @@ public class ChatClient {
     
     /// The token of the current user. If the current user is anonymous, the token is `nil`.
     @Atomic var currentToken: Token?
-
-    /// In case of token expiration this property is used to obtain a new token
-    public var tokenProvider: TokenProvider?
-
+    
     /// Sets the user token to the client, this method is only needed to perform API calls
     /// without connecting as a user.
     /// You should only use this in special cases like a notification service or other background process
@@ -244,8 +241,7 @@ public class ChatClient {
     ///   - config: The config object for the `Client`. See `ChatClientConfig` for all configuration options.
     ///   - tokenProvider: In case of token expiration this closure is used to obtain a new token
     public convenience init(
-        config: ChatClientConfig,
-        tokenProvider: TokenProvider? = nil
+        config: ChatClientConfig
     ) {
         var environment = Environment()
         
@@ -255,7 +251,6 @@ public class ChatClient {
         
         self.init(
             config: config,
-            tokenProvider: tokenProvider,
             environment: environment
         )
     }
@@ -268,11 +263,9 @@ public class ChatClient {
     ///
     init(
         config: ChatClientConfig,
-        tokenProvider: TokenProvider? = nil,
         environment: Environment
     ) {
         self.config = config
-        self.tokenProvider = tokenProvider
         self.environment = environment
 
         setupConnectionRecoveryHandler(with: environment)
@@ -300,19 +293,44 @@ public class ChatClient {
         )
     }
     
-    /// Connects authorized user
+    /// Connects the client with the given user.
+    ///
+    /// - Parameters:
+    ///   - userInfo: The user info passed to `connect` endpoint.
+    ///   - tokenProvider: The closure used to retreive a token. Token provider will be used to establish the initial connection and also to obtain the new token when the previous one expires.
+    ///   - completion: The completion that will be called once the **first** user session for the given token is setup.
+    ///
+    /// - Note: Connect endpoint uses an upsert mechanism. If the user does not exist, it will be created with the given `userInfo`. If user already exists, it will get updated with non-nil fields from the `userInfo`.
+    public func connectUser(
+        userInfo: UserInfo,
+        tokenProvider: @escaping TokenProvider,
+        completion: ((Error?) -> Void)? = nil
+    ) {
+        setConnectionInfoAndConnect(
+            userInfo: userInfo,
+            userConnectionProvider: .init(tokenProvider: tokenProvider),
+            completion: completion
+        )
+    }
+    
+    /// Connects the client with the given user.
+    ///
     /// - Parameters:
     ///   - userInfo: User info that is passed to the `connect` endpoint for user creation
     ///   - token: Authorization token for the user.
     ///   - completion: The completion that will be called once the **first** user session for the given token is setup.
+    ///
+    /// - Note: Connect endpoint uses an upsert mechanism. If the user does not exist, it will be created with the given `userInfo`. If user already exists, it will get updated with non-nil fields from the `userInfo`.
+    ///
+    /// - Important: This method can only be used when `token` does not expire. If the token expores, the `connect` API with token provider has to be used.
     public func connectUser(
         userInfo: UserInfo,
         token: Token,
         completion: ((Error?) -> Void)? = nil
     ) {
-        setConnectionInfoAndConnect(
+        connectUser(
             userInfo: userInfo,
-            userConnectionProvider: .static(token),
+            tokenProvider: { $0(.success(token)) },
             completion: completion
         )
     }
@@ -329,6 +347,7 @@ public class ChatClient {
         setConnectionInfoAndConnect(
             userInfo: userInfo,
             userConnectionProvider: .guest(
+                client: self,
                 userId: userInfo.id,
                 name: userInfo.name,
                 imageURL: userInfo.imageURL,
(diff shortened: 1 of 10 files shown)


bielikb:
This is much needed change! The approach is basically the same I took. Great minds think alike :)

Sources/StreamChat/ChatClient.swift
```diff
@@ -227,10 +227,7 @@ public class ChatClient {
     
     /// The token of the current user. If the current user is anonymous, the token is `nil`.
     @Atomic var currentToken: Token?
-
-    /// In case of token expiration this property is used to obtain a new token
-    public var tokenProvider: TokenProvider?
-
```
nuno-vieira, line 233: Could we just deprecate this, instead of totally removing it?

nuno-vieira (commented):
The change LGTM ✅ I would prefer to deprecate the old API instead of removing it though. Is this possible?

polqf (commented):
If I were to be the client, I would not be happy with either the previous nor the proposed approaches. Chances are that I want to configure all the things in an initial point in the execution of the app, and only call connect whenever needed, without the need of knowing more about it.

Is this something we should maybe consider? `connect` should know nothing of tokens IMO

Sources/StreamChat/ChatClient.swift
evsaev: @nuno-vieira when they are 2 different variables we have to manually keep them in sync: when any of it changes, another one needs to be changed as well. E.g. when a user connects as a `guest` but then assigns the new `tokenProvider`, what should we do with `userConnectionProvider`? ATM there's no way to understand what's the type of `userConnectionProvider` since it's just a closure.

It was a mistake to keep a `tokenProvider` separately from `userConnectionProvider` when they are essentially the same thing.

We could try to gracefully deprecate `tokenProvider` but it's gonna be a lot of boilerplate with the low-value outcome, I think. Customers leveraging `tokenProvider` will get a warning instead of a compilation error and by trying to fix it will end up with the implementation shown in PR description.

I'm against breaking changes in minor releases but this one seems like one customers will be happy about

polqf:
> @polqf yeah, was also thinking about that.
> 
> The tokenProvider could be passed once via init when ChatClient is created but in that case it needs a user to fetch a token for:
> 
> (userID: UserId, completion: @escaping (Result<Token, Error>) -> Void) -> Void
> Tho it can be done this way, I see 2 potential issues:
> 
> it is not aligned with other SDKs
> it can be misleading because for guest/anonymous users the tokenProvider won't be triggered but has to be provided
> The 2nd is more or less fine and expected but 1st contradicts with API alignment we're aiming for.

Yeah. I do agree with 1, but tbh I don't thing this should be something we base our decisions off. Otherwise there's always a blocking point towards our ideal "excellence" direction
On the init approach, I always thought it is better to have a `configure` method that does exactly that. The same way we were discussing before that connect and init could happen at different places/moments, I believe this is the case too.

Maybe I am proposing something that it is too big for this PR, but I think it worths a discussion, at least.

evsaev:
> Yeah. I do agree with 1, but tbh I don't thing this should be something we base our decisions off. Otherwise there's always a blocking point towards our ideal "excellence" direction On the init approach, I always thought it is better to have a `configure` method that does exactly that. The same way we were discussing before that connect and init could happen at different places/moments, I believe this is the case too.
> 
> Maybe I am proposing something that it is too big for this PR, but I think it worths a discussion, at least.

Ok, I see the point of initializing and configuring chat-client in different places, that makes sense 👍 But if we go with `configure`, someone can ask why we provide `ChatClientConfig` via `ChatClient.init` and not via dedicated `configure` method? And changing it gonna time because we won't longer be able to create `database/apiClient` without having a config at the client initialization point OR will need to recreate them after `configure`is called.

> I don't thing this should be something we base our decisions off. Otherwise there's always a blocking point towards our ideal "excellence" direction

As mentioned, the approach with `configure` method is also a good one but since it's connection API we can't immediately introduce it on iOS. This way, we'll never get aligned with other platforms. We can suggest the new approach to other teams and if they like it - plan migration to the new `connect` API in the next major.

@polqf WDYT?

polqf:
@evsaev yeah, as long as there is a discussion to improve this, I am happy to let this through for now 😄 Having it aligned with other SDKs is always better that having it as we have it now.

bielikb (approved):
Great stuff!

bradleyrzeller:
Hi - 

If I understand this change correctly, there is now no way for the token provider to differentiate between a token needed for the "initial" connection and one that is needed because the current one expired (a refresh). This distinction is critical for the token provider to determine whether or not a cached token can be attempted or not.

Before, you could provide the initial token in the `connect()` method and the token provider could be set afterwards. In my testing, the token provider was only ever invoked upon token invalidation / refreshes.

There are some workarounds to the new API here but I'm curious what the stream team recommends for this situation.

nuno-vieira:
Hi @bradleyrzeller!

You are correct. For now, you can do the workarounds, and maybe add a boolean flag to check if the method is the first time being called. Currently, we don't have plans to change this, especially because it is aligned with the rest of the platforms. But we will discuss this internally and will let you know if we have updates on this in the future 👍 

Best,
Nuno

bradleyrzeller:
Ok, thanks for the reply @nuno-vieira.

I've been testing my workaround here and I am actually not seeing the token provider be called automatically after the token has expired.

I am generating a token using your [token generator](https://getstream.io/chat/docs/react/token_generator/) with 1 minute expiry. The token works initially as expected. I can send messages up until it expires ... and then the client disconnects. However, as I mentioned, the client is not invoking my token provider when it clearly knows that the token has expired (api calls return 401s etc.).

Can you provide some additional insight here? Thanks!

nuno-vieira:
Hi, @bradleyrzeller can you provide a snippet of your workaround?

bradleyrzeller:
It's essentially this:

```
        client.connectUser(
            userInfo: .init(id: "foobar"),
            tokenProvider: { callback in
                if self.allowCachedCredentials {
                    self.allowCachedCredentials = false
                    callback(.success(cachedToken))
                } else {
                    fetchNewToken { tokenResult in
                        callback(tokenResult)
                    }
                }
        )
```

The token provider is only ever invoked once, upon initial connection. I cannot get it to be invoked again even after my token has expired and the client disconnects.

bradleyrzeller:
@nuno-vieira it seems to be because the `RequestDecoder` is failing to decode an error payload into a `ErrorPayload` on this [line](https://github.com/GetStream/stream-chat-swift/blob/3f748ee181c4097964858deae47c2ba8cb811dae/Sources/StreamChat/APIClient/RequestDecoder.swift#L49). The actual payload I am seeing is 
```
{
  "code" : 40,
  "message" : "SendEvent failed with error: \"JWTAuth error: token is expired (exp)\"",
  "more_info" : "https:\/\/getstream.io\/chat\/docs\/api_errors_response",
  "StatusCode" : 401,
  "duration" : ""
}
```

but the ErrorPayload expects a `details` key.

This throws off [downstream logic](https://github.com/GetStream/stream-chat-swift/blob/3f748ee181c4097964858deae47c2ba8cb811dae/Sources/StreamChat/APIClient/RequestDecoder.swift#L58) that detects token expiration and triggers a refresh...

bradleyrzeller:
Just filed an [issue](https://github.com/GetStream/stream-chat-swift/issues/2247) for visibility.

PR #2032, opened 2022-05-30
Fix crash when accessing NSHashTable from different threads

polqf:
### 🔗 Issue Links

https://stream-io.atlassian.net/browse/CIS-1803

### 🎯 Goal

Fix crash when accessing NSHashTable from multiple threads at the same time.

### 📝 Summary

In the ChatClient, we had the following:
```
    private(set) var activeChannelListControllers = NSHashTable<ChatChannelListController>.weakObjects()
    private(set) var activeChannelControllers = NSHashTable<ChatChannelController>.weakObjects()
```
These are used to keep a weak reference to all the controllers we have active.
Because of those are currently accessed from different thread, there is chance of a race condition, which in this case leads to a crash

### 🛠 Implementation

To have a thread safe weak collection, a new class (`ThreadSafeWeakCollection`) has been introduced. It is basically a thread safe wrapper around NSHashTable

### 🎨 Showcase

Screenshot 2022-05-30 at 18 12 31

### ☑️ Contributor Checklist


Sources/StreamChat/Utils/ThreadSafeWeakCollection.swift
```diff
@@ -0,0 +1,46 @@
...
+
+    var count: Int {
+        var count: Int = 0
+        queue.sync {
+            count = storage.count
+        }
+        return count
+    }
+
+    func add(_ object: T?) {
```
evsaev, line 27: It's a bit unclear what we expect adding `nil` object to the collection. Can we make the param non optional?
```suggestion
    func add(_ object: T) {
```

evsaev (approved):
That's a well-thought implementation, reads can be concurrent but all mutating operations are barriers. Amazing 👏

Sources/StreamChat/Utils/ThreadSafeWeakCollection.swift
polqf: I am matching the signature of NSHashMap methods

Sources/StreamChat/Utils/ThreadSafeWeakCollection.swift
```diff
@@ -0,0 +1,46 @@
...
+        }
+    }
+
+    func removeAllObjects() {
+        queue.async(flags: .barrier) {
+            self.storage.removeAllObjects()
+        }
+    }
+
+    func contains(_ anObject: T?) -> Bool {
```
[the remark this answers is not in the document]
polqf: I am matching NSHashMap's API

Sources/StreamChat/Utils/ThreadSafeWeakCollection.swift
```diff
@@ -0,0 +1,46 @@
+//
+// Copyright © 2022 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+final class ThreadSafeWeakCollection<T: AnyObject> {
+    private let queue = DispatchQueue(label: "io.stream.com.weak-collection")
```
nuno-vieira, line 8: This means that all `TheadSafeWeakCollection`'s will use the same queue. Shouldn't we pass the label in the initializer, so that each collection has its own queue?

polqf: Nope, not at all. The label is mainly used for debugging purposes, DispatchQueue will not create unique queues based on it.

nuno-vieira: You are right 👍 This a wrong assumption from me

nuno-vieira (approved):
LGTM! ✅

Issue #1762, opened 2022-01-21
QuotedChatMessageView causes message list flicker


vladpre92:
## What did you do?
Added a quoted message with the last DempApp

## What did you expect to happen?
Everything to be good looking

## What happened instead?
Layout vertical spacings are updating late causing a flicker.

## GetStream Environment
**GetStream Chat version:** 4.9.0
**GetStream Chat frameworks:** StreamChat, StreamChatUI
**iOS version:** 15.0
**Swift version:** 5.0
**Xcode version:** 13.1
**Device:** iPhone 7, iOS Simulator 13 Pro Max (not tested on other simulators)

## Additional context
This happens only when having quoted messages visible when message list appears.
Investigated a little subclassing QuotedChatMessageView. It seems that setting ` preservesSuperviewLayoutMargins = true` inside `setUpLayout()` method causes this. Setting `preservesSuperviewLayoutMargins = false` removes the problem but you lose the margins.

https://user-images.githubusercontent.com/3681654/150523248-98521a57-b359-47be-a186-58eb8d9bf54f.mov

Thank you!

adamrushy:
Thank you for reporting @vladpre92 I can see the issue. I am going to create a ticket internally to get this fixed :] will update you on the progress, thanks again!

adamrushy:
@vladpre92 Hey, just an update to say we have this ticket scoped out and is in the backlog to be worked on, will update when it's completed :]

linhewinner:
I just upgraded to 4.12.0 and I can still reproduce this bug. See attached screen recording (the first time there is a repro but the second time there isn't). Something I noticed:

1. On 4.10.0, I could reproduce this nearly 100% but with 4.12.0, it's much harder to repro.
2. It seems like the bug only happens when I background and foreground the app (where the WS connection would reconnect and trigger a reload)

I suspect that in order to reproduce this bug, you need some reload/syncing to be happening and causing CoreData updates of some sort.
This bug is highly visible because most commonly, users get a push and goes into the app where a reconnection/reload happens ... they go into the ChannelView and they almost always see the bug.

https://user-images.githubusercontent.com/97132583/158896301-5fbc766a-2342-4980-a6b8-8878cc9119ab.mp4

nuno-vieira:
Hi @vladpre92!

Thank you for your patience in waiting for this issue to get solved!

The fix will be finally available in the next release 🎉 

We will close it once the release is published 👍 

Best,
Nuno

[...]

PR #2050, opened 2022-06-02
[CIS-1529] Fix quoted message adding extra top spacing

nuno-vieira:
### 🔗 Issue Links
CIS-1529
https://github.com/GetStream/stream-chat-swift/issues/1762

### 🎯 Goal
Fixes quoted message list UI flicker bug. This was caused because there was an extra margin-top in the quoted message view. Sometimes the cell would reload and cause this jump, but sometimes the jump wouldn't happen but the cell with have extra padding, and only after scrolling it would go away.

### 📝 Summary
- Uses `embed()` to layout the `containerView` instead of pinning it to the layout margins. The `ContainerStackView` is not working properly when using the latter API.
- Although we could add insets to the `embed()` to replicate the previous design, it doesn't actually follow the design on Figma, here: https://www.figma.com/file/ekifwChR9tR7zRJg1QEzSM/Chat-Design-Kit---Current-version?node-id=22190%3A297522

### 🎨 Showcase

image

| Before | After |
| ------ | ----- |
| image   |  image  |

### 🧪 Manual Testing Notes
- Add a normal message
- Quote the normal message
- There shouldn't be any gap between the normal and quoted message

### ☑️ Contributor Checklist


evsaev (approved):
Nice 👏

bielikb (approved):

PR #2051, opened 2022-06-02
[CIS-1727] Expired token handling

evsaev:
### 🔗 Issue Links

CIS-1727

### 🎯 Goal

- When a user connects with the expired token, the completion needs to be called with `token expired` error 
- Improve how token refreshing is handled

### 🛠 Implementation

When WS gets disconnected with the `token expired` error, the token refresh is followed by WS reconnection.

When an API request fails with `token expired` error, the WS is manually disconnected (if it's connected), token is refreshed and WS was connected before.

`TokenHandler` encapsulating token refreshing logic is introduced. It guarantees that there's only 1 ongoing token refresh process active at a time.

### 🧪 Manual Testing Notes

Expired token error is propagated to `connectUser` completion:
**WHEN** `ChatClient` is connected with an expired token
**THEN** the expired token error is propagated to `connectUser` completion
**AND** token refresh flow is **not** triggered

Token expired error is reported by web-socket:
**GIVEN** `ChatClient` is connected with `tokenProvider` that returns a token with an expiration date
**AND** user performs no actions and just stays on the channel list
**WHEN** web-socket get's disconnected
**AND** reconnection timer is scheduled
**AND** reconnection timer fires
**AND** `connect` endpoint is called with expired token
**AND** `websocket` gets connected, reports `expired token` and disconnects
**THEN** token refresh flow is triggered and a new token if fetched
**AND** `websocket` is reconnected automatically 
**AND** recovery flow is triggered

Token expired error is reported by API request:
**GIVEN** `ChatClient` is connected with `tokenProvider` that returns a token with an expiration date
**AND** user performs actions is chat
**WHEN** one of the actions fails with an expired token error 
**THEN** web-socket is manually disconnected without reconnection scheduled
**AND** token refresh flow is triggered and the new token if fetched
**AND** web-socket is manually reconnected when the token is refreshed
**AND** failed API request is repeated

Token refresh has retry mechanism:
**GIVEN** `ChatClient` is connected with `tokenProvider` that returns a token with an expiration date
**AND** token expires
**AND** tokenProvider is asked for the new token
**WHEN** token request fails
**THEN** token request is retired with increasing delay if there're attempts left

Token expires when app is in background/offline:
**GIVEN** `ChatClient` is connected with `tokenProvider` that returns a token with an expiration date
**AND** user goes offline/sends app to background
**AND** `websocket` get's disconnected
**AND** token expires
**WHEN** user goes back online/sends the app to the foreground
**AND** recovery flow is triggered
**AND** /sync fails with expired token error
**THEN** token refresh is triggered and a new token if fetched
**AND** /sync is successfully retried
**AND** recovery flow is continued

### ☑️ Contributor Checklist


Sources/StreamChat/APIClient/APIClient.swift
```diff
@@ -167,19 +158,10 @@ class APIClient {
     /// - Parameters:
     ///   - endpoint: The `Endpoint` used to create the network request.
     ///   - completion: Called when the networking request is finished.
-    private func executeRequest<Response: Decodable>(
+    func executeRequest<Response: Decodable>(
```
evsaev, line 161: When a user is connected as a `guest` token refresh is done using stream backend. The thing is, when token expired error is reported during **connection recovery flow** when request queue in APIClient is suspended we make this request off the queue so it can pass through.

Sources/StreamChat/Workers/Background/TokenHandler.swift
```diff
@@ -0,0 +1,225 @@
...
+}
+
+final class DefaultTokenHandler: TokenHandler {
+    private let maximumTokenRefreshAttempts: Int
+    private var retryStrategy: RetryStrategy
+    private let retryTimerType: Timer.Type
+    private var retryTimer: TimerControl?
+    
+    @Atomic private var isRefreshingToken: Bool = false
+    @Atomic private var tokenWaiters: [WaiterToken: TokenWaiter] = [:]
```
evsaev, line 44: Might be better to handle it using dispatch queue, left as it was for simplicity

Sources/StreamChat/Workers/Background/TokenHandler.swift
```diff
@@ -0,0 +1,225 @@
...
+            guard let self = self else { return }
+            
+            // When `set(token:)` get's invoked while the ongoing token refresh process,
+            // we should assign the given token and cancel the refresh flow. However, cancelling refresh flow
+            // is currently not possible because `tokenProvider` is not returning the cancellable (planned for v5).
+            //
+            // To avoid overriding the token assinged by `set(token:)`, we store the token before the refresh
+            // and compare it we with the current token when refresh is completed. If the tokens does not match,
+            // it means the token was manually assigned and results from `tokenProvider` should be discarded.
+            guard self.currentToken == tokenBeforeRefresh else {
```
evsaev, line 109: To correctly handle this, the `tokenProvider` should return the cancellable request token. 

I've added the [note](https://www.notion.so/stream-wiki/StreamChat-v5-e7cd0c287fba41e8af69ae507c40d9c9#2aa41ad748874809926ce39fed3be743) to v5 document so we don't forget

Sources/StreamChat/Workers/Background/TokenHandler.swift
```diff
@@ -0,0 +1,225 @@
...
+                let tokenAssingedWhileRefresh = """
+                The `currentToken` was assinged while the refresh process.
+                The token given by `tokenProvider` will be discarded.
+                """
+                log.info(tokenAssingedWhileRefresh, subsystems: .tokenRefresh)
+                return
+            }
+            
+            switch $0 {
+            case .success(let newToken) where newToken == self.currentToken:
```
evsaev, line 119: We could also check that the new token is for the current user (there'll be one more check related to `token.userId`)

evsaev:
Handling for the following cases is currently missing:
- token provider returns an expired
- token provider returns an invalid token
- token provider does not return result in reasonable time (timeout)

PR #2053, opened 2022-06-03
Fix hidden cell content after popover transition

polqf:
### 🔗 Issue Links

Reported through Slack: https://getstream.slack.com/archives/C01FT5767AS/p1654165950668659

### 🎯 Goal

Fix hidden cell content after reacting to a message

### 📝 Summary

There was an edge case where, in ChatMessageActionsTransitionController, we would set the alpha of the cell content to 0 during an animation, and never reset it back to 1.

### 🛠 Implementation

The previous logic on the transition controller was looking for the cell both at the beginning of the animation and at the end of it. It was basically looking at the visible cells, and finding the first matching the cell content id.

In some cases, the cell content that was used at the beginning of the animation and the one that was used after were different references. That led to some cells staying with alpha 0.

This PR changes this logic by keeping a local reference to the cell content before the animation starts, to make sure the reference where we alter the alpha is always the same.

### 🎨 Showcase
Before:

https://user-images.githubusercontent.com/7887319/171851491-5e00be17-7f97-45c9-9739-607fd79d2316.mov

### ☑️ Contributor Checklist


bielikb (approved):
Looks good!

nuno-vieira (approved):
LGTM! 🙂

Issue #2039, opened 2022-06-01
Hide channel with clearHistory doesn't show when new message receive


ajay-simform:
## What did you do?
- login with users x and y on a different device
- x sends a direct message to y
- x will hide the channel with clearing history `channelController.hideChannel(clearHistory: true, completion: nil)`
- y will hide the channel with clearing history `channelController.hideChannel(clearHistory: true, completion: nil)`
- x and y close the app
- x and y re-open the app
- x send a direct message to y
- y doesn't see channel in the channelList

## What did you expect to happen?
- should the channel visible when a new message receive after hiding channel

## What happened instead?
- channel doesn't show in channel list

## GetStream Environment
**GetStream Chat version:** 4.15.0
**GetStream Chat frameworks:** StreamChat, StreamChatUI
**iOS version:** 15.2
**Swift version:** 5.3
**Xcode version:** 13.2
**Device:** iphone-XS

b-onc:
Hey @ajay-simform 

We were able to pinpoint and resolve the issue. The PR is linked to this ticket. It'd be awesome if you can test the PR & tell us if the issue is resolved!

polqf:
Hi @ajay-simform, we just released [4.17.0](https://github.com/GetStream/stream-chat-swift/releases/tag/4.17.0) which contains a fix for the issue you reported.

Let us know if we can help you with something else

PR #2056, opened 2022-06-03
Multiple fixes for channel hiding/appearing logic

b-onc:
### 🔗 Issue Links

#2039 

### 🎯 Goal

Channel hiding/appearing mechanism was broken with the introduction of offline caching

### 📝 Summary

This PR contains multiple fixes for multiple bugs.

1. DemoApp keeps a reference to last displayed ChannelVC
1. Duplicated pinned message saving in MessageDTO (performance)
1. channel.hidden event fails to decode
1. Locally saved hidden channels do not re-appear when they should
1. Hidden channel messages re-appear seemingly randomly
1. Hidden channel messages leak via channel list preview 
1. Querying a hidden channel makes the channel un-hide

### 🛠 Implementation

1. _DemoApp keeps a reference to last displayed ChannelVC_
We were always using splitVC, which keeps the ref to latest displayed VC, which keeps it alive. We should use the old logic and only use splitVC when we detect an iPad
Keeping a ref to ChannelVC causes the latest displayed channel to keep marking newly received messages as "read", which messes up the unread count

1. _Duplicated pinned message saving in MessageDTO (performance)_
This has a big performance implication since it included one `saveUser` call, effectively adding ~500 `saveUser` calls per channel list

1. _channel.hidden event fails to decode_
This only happens for channel.hidden events coming from sync endpoint. Backend explained that `clearHistory` is treated as custom data, and custom data is not sent on `/sync` call, so it's not sent. It's a bug that will be resolved by backend, but client needs a workaround meanwhile.
Also resolved that decoding errors are not printed to console.

1. _Locally saved hidden channels do not re-appear when they should_
When the client starts for the first time, the channels linked to the query will be watched. If a channel was hidden on a previous session, that channel will not be linked to the query, so will not be watched. We won't receive a `MessageNew` event for that channel (since it's not being watched) and we won't mark it as `hidden = false`, so it won't show up.
The visibility middleware should also act on `Notification` events and mark the channel accordingly. Then it can be linked to any query that's currently alive.

1. _Hidden channel messages re-appear seemingly randomly_
This happens because client uses `truncatedAt` as `hiddenAt`. Hiding channel with clearHistory flag sets truncatedAt, but backend is not aware so they send null for this value, and client uses null value to reset truncatedAt, effectively un-hiding the messages.
As per slack discussion (https://getstream.slack.com/archives/CE5N802GP/p1654094011610589?thread_ts=1654093897.531669&cid=CE5N802GP) it's not possible to un-truncate a channel by setting truncatedAt to null once it's set, so it's fine to only use non-nil truncatedAt values sent by backend

1. _Hidden channel messages leak via channel list preview_
We didn't re-set the `previewMessage` of a hidden channel so if `clearHistory == true`, last message was still set as `previewMessage` and leaked

1. _Querying a hidden channel makes the channel un-hide_
This was a backend issue, it was resolved & deployed, cannot be reproduced anymore

### 🧪 Manual Testing Notes

Reproduction videos for the bugs:

Points 5, 6, 7
https://user-images.githubusercontent.com/770464/171873446-dbbd5e9e-c6b5-42eb-9395-7f4ceb90cb92.mp4

Points 3 & 4
https://user-images.githubusercontent.com/770464/171873496-b59dd491-9a91-463b-9b13-3e519c1a2e18.mp4

### ☑️ Contributor Checklist

### DIFF
--- a/DemoApp/DemoAppCoordinator.swift
+++ b/DemoApp/DemoAppCoordinator.swift
@@ -67,7 +67,7 @@ extension DemoAppCoordinator: UNUserNotificationCenterDelegate {
 // MARK: - Navigation
 
 private extension DemoAppCoordinator {
-    func showChat(for user: DemoUserType, cid: ChannelId? = nil, animated: Bool) {
+    func showChat(for user: DemoUserType, cid: ChannelId?, animated: Bool) {
         logIn(as: user)
         
         let chatVC = makeChatVC(for: user, startOn: cid) { [weak self] in
@@ -81,7 +81,7 @@ private extension DemoAppCoordinator {
     
     func showLogin(animated: Bool) {
         let loginVC = makeLoginVC { [weak self] user in
-            self?.showChat(for: user, animated: true)
+            self?.showChat(for: user, cid: nil, animated: true)
         }
         
         set(rootViewController: loginVC, animated: animated)
@@ -131,12 +131,18 @@ private extension DemoAppCoordinator {
             selectedChannel: selectedChannel,
             onLogout: onLogout
         )
-        let channelListNVC = UINavigationController(rootViewController: channelListVC)
         
-        let splitVC = UISplitViewController()
-        splitVC.preferredDisplayMode = .oneBesideSecondary
-        splitVC.viewControllers = [channelListNVC, channelNVC].compactMap { $0 }
-        return splitVC
+        let channelListNVC = UINavigationController(rootViewController: channelListVC)
+        let isIpad = UIDevice.current.userInterfaceIdiom == .pad
+        if isIpad {
+            let splitVC = UISplitViewController()
+            splitVC.preferredDisplayMode = .oneBesideSecondary
+            splitVC.viewControllers = [channelListNVC, channelNVC].compactMap { $0 }
+            return splitVC
+        } else {
+            channelVC.map { channelListNVC.pushViewController($0, animated: false) }
+            return channelListNVC
+        }
     }
     
     func makeChannelListVC(
@@ -284,11 +290,17 @@ final class DemoChatChannelVC: ChatChannelVC {
     }
     
     @objc private func debugTap() {
-        guard
-            let cid = channelController.cid,
-            let mainVC = splitViewController?.viewControllers.first as? UINavigationController,
-            let channelListVC = mainVC.viewControllers.first as? DemoChatChannelListVC
-        else { return }
+        guard let cid = channelController.cid else { return }
+        
+        let channelListVC: DemoChatChannelListVC
+        if let mainVC = splitViewController?.viewControllers.first as? UINavigationController,
+           let _channelListVC = mainVC.viewControllers.first as? DemoChatChannelListVC {
+            channelListVC = _channelListVC
+        } else if let _channelListVC = navigationController?.viewControllers.first as? DemoChatChannelListVC {
+            channelListVC = _channelListVC
+        } else {
+            return
+        }
         
         channelListVC.demoRouter.didTapMoreButton(for: cid)
     }
--- a/Sources/StreamChat/Database/DTOs/ChannelDTO.swift
+++ b/Sources/StreamChat/Database/DTOs/ChannelDTO.swift
@@ -65,7 +65,7 @@ class ChannelDTO: NSManagedObject {
     override func willSave() {
         super.willSave()
 
-        // Change to the `trunctedAt` value have effect on messages, we need to mark them dirty manually
+        // Change to the `truncatedAt` value have effect on messages, we need to mark them dirty manually
         // to triggers related FRC updates
         if changedValues().keys.contains("truncatedAt") {
             messages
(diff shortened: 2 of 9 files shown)


Sources/StreamChat/Database/DTOs/ChannelDTO.swift
```diff
@@ -177,6 +177,14 @@ extension NSManagedObjectContext {
         dto.lastMessageAt = payload.lastMessageAt?.bridgeDate
         dto.memberCount = Int64(clamping: payload.memberCount)
         dto.truncatedAt = payload.truncatedAt?.bridgeDate
+        // We don't always set `truncatedAt` directly since we also use this property as `hiddenAt`
+        // when a channel is hidden with `clearHistory`
+        // Backend doesn't send `truncatedAt` for this case and we reset it, causing the hidden messages to re-appear
+        // It's not possible to set `truncatedAt` to `nil` once it's set on backend side
+        // so it's safe to keep client-side changes when backend sends `nil`
+        if payload.truncatedAt != nil {
+            dto.truncatedAt = payload.truncatedAt?.bridgeDate
```
polqf, line 186: This sounds dangerous, and it would be much better to clear the message history, as stated in the Slack discussion. Any reason why you went with this approach?

b-onc: Even if we cleared the message history, we still need this "workaround" as long as we use `truncatedAt` as `hiddenAt`
Alternatively,
```swift
- channelDTO.truncatedAt = event.createdAt.bridgeDate
+ channelDTO.messages = channelDTO.messages.filter({ $0.createdAt.bridgeDate > event.createdAt })
```
but i'm not sure of the performance implication of this. Doable though.

Why does it sound dangerous? There's no way to un-truncate a channel by setting `truncatedAt` to nil, so we're safe to mutate `truncatedAt` locally as long as backend keeps sending `null` for it.

adamrushy: Do we have a timescale for the backend implementation? It feels like OK to do this but we should really push for the backend implementation

Sources/StreamChat/WebSocketClient/EventMiddlewares/ChannelVisibilityEventMiddleware.swift
```diff
@@ -36,6 +37,15 @@ struct ChannelVisibilityEventMiddleware: EventMiddleware {
...
                 channelDTO.isHidden = false
                 
+            // New Message will unhide the channel
+            // but we won't get `ChannelVisibleEvent` for this case
+            case let event as NotificationMessageNewEventDTO:
+                guard let channelDTO = session.channel(cid: event.channel.cid) else {
+                    throw ClientError.ChannelDoesNotExist(cid: event.channel.cid)
+                }
+                
+                channelDTO.isHidden = false
```
adamrushy, line 47: Wondering what this is used for 🤔

DemoApp/DemoAppCoordinator.swift
```diff
@@ -131,12 +131,17 @@ private extension DemoAppCoordinator {
...
-        return splitVC
+        let channelListNVC = UINavigationController(rootViewController: channelListVC)
+        let isIpad = UIDevice.current.userInterfaceIdiom == .pad
+        if isIpad {
+            let splitVC = UISplitViewController()
+            splitVC.preferredDisplayMode = .oneBesideSecondary
+            splitVC.viewControllers = [channelListNVC, channelNVC].compactMap { $0 }
+            return splitVC
+        } else {
+            return channelListNVC
```
evsaev, line 143: If we start with `cid`, the app should start with channel screen pushed to the navigation

b-onc: That's already handled, isn't it? `makeChannelListVC` accepts `selectedChannel`

Sources/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -624,13 +624,6 @@ extension NSManagedObjectContext: MessageDatabaseSession {
            let parentMessageDTO = MessageDTO.load(id: parentMessageId, context: self) {
             parentMessageDTO.replies.insert(dto)
         }
-
-        dto.pinned = payload.pinned
-        dto.pinnedAt = payload.pinnedAt?.bridgeDate
-        dto.pinExpires = payload.pinExpires?.bridgeDate
-        if let pinnedBy = payload.pinnedBy {
-            dto.pinnedBy = try saveUser(payload: pinnedBy)
-        }
```
evsaev, line 633: Why is this gone?

Sources/StreamChat/WebSocketClient/EventMiddlewares/ChannelVisibilityEventMiddleware.swift
```diff
@@ -25,6 +25,7 @@ struct ChannelVisibilityEventMiddleware: EventMiddleware {
                 
                 if event.isHistoryCleared {
                     channelDTO.truncatedAt = event.createdAt.bridgeDate
+                    channelDTO.previewMessage = session.preview(for: event.cid)
```
evsaev, line 28: Good catch 👍

Sources/StreamChat/Database/DTOs/MessageDTO.swift
b-onc: It was duplicated, see pr description

Sources/StreamChat/WebSocketClient/EventMiddlewares/ChannelVisibilityEventMiddleware.swift
evsaev: Sorry, un-resolving because it's better to have everything related to channel preview in one place. We're doing it in `saveEvent -> updateChannelPreview`.

b-onc: done

evsaev (approved):
LGTM!

PR #2059, opened 2022-06-03
Demo App restructured

bielikb:
### 🎯 Goal

Restructure Demo App.

### 📝 Summary

* StreamChat API, components, extensions were consolidated under `StreamChat` folder in `DemoApp`
* `StreamChatWrapper` now represents the integration point for `StreamChat` SDK
* UI screens were separated into single swift files and moved under `Screens` folder
* `DemoAppConfiguration` acts as a single place to configure/initialize configuration used by the iOS dev team == `AppDelegate` became thin.
* `DemoAppCoordinator` now holds onto `StreamChatWrapper` instance and `PushNotifications` instance. Several responsibilities handled by coordinator were moved onto respective interfaces (wrapper or push notifications) == better separation of concerns.

### 🎨 Showcase

| Before | After |
| ------ | ----- |
| Screenshot 2022-06-08 at 16 48 01  |  Screenshot 2022-06-08 at 16 47 16 |

### 🧪 Manual Testing Notes

### ☑️ Contributor Checklist

### 🎁 Meme


DemoApp/DemoAppConfiguration.swift
```diff
@@ -0,0 +1,58 @@
...
+            }
+        }
+
+        // Performance tracker
+        func trackPerformanceIfNeeded() {
+            if isStreamInternalConfiguration {
+                PerformanceMonitor.shared().performanceViewConfigurator.options = [.performance]
+                PerformanceMonitor.shared().start()
+            }
+        }
```
evsaev, line 56: We declare several func level funcs but we don't call them anywhere 🤔

DemoApp/PushNotifications.swift
```diff
@@ -0,0 +1,62 @@
...
+        UNUserNotificationCenter.current()
+    }
+
+    var authorizationOptions: UNAuthorizationOptions {
+        [.alert, .sound, .badge]
+    }
+
+    var onNotificationResponse: ((UNNotificationResponse) -> Void)?
+
+    static let shared = PushNotifications()
```
evsaev, line 20: I would be avoiding singletons. Why not to have it as a property in `AppDelegate`?

DemoApp/DemoAppCoordinator.swift
```diff
@@ -34,33 +32,22 @@ final class DemoAppCoordinator: NSObject {
...
+    func handlePushNotificationResponse() {
+        pushNotifications.onNotificationResponse = { [weak self] response in
+            guard case UNNotificationDefaultActionIdentifier = response.actionIdentifier else {
+                return
+            }
 
-        guard let cid = notificationInfo.cid else {
-            return
-        }
+            guard let self = self else { return }
```
evsaev, line 42: I would do that as the first thing in closure and move checking `actionIdentifier` to the next guard OR event have a single guard for all statements

DemoApp/StreamChat/Extensions/DeletedMessageVisibility+CustomStringConvertible.swift
```diff
@@ -0,0 +1,23 @@
+//
+// Copyright © 2022 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+import StreamChat
+
+extension ChatClientConfig.DeletedMessageVisibility: CustomStringConvertible {
```
evsaev, line 8: What if we just make the change in LLC? When this setting is moved to the backend, string would be the way to encode/decode it from JSON so it's fine to make this change, I think
```swift
public enum DeletedMessageVisibility: String {
        /// All deleted messages are always hidden.
        case alwaysHidden
        /// Deleted message by current user are visible, other deleted messages are hidden.
        case visibleForCurrentUser
        /// Deleted messages are always visible.
        case alwaysVisible
    }
```

DemoApp/StreamChat/StreamChatWrapper.swift
```diff
@@ -0,0 +1,155 @@
...
+        client = nil
+    }
+}
+
+// MARK: Controllers
+
+extension StreamChatWrapper {
+    func channelController(for channelId: ChannelId?) -> ChatChannelController? {
+        channelId.map { client.channelController(for: $0) }
+    }
```
evsaev, line 122: I would make `channelId` non-optional and handle `nils` on the caller side. Asking the wrapper to give a controller for `nil` is a bit weird

DemoApp/StreamChat/StreamChatWrapper.swift
```diff
@@ -0,0 +1,155 @@
...
+    func channelListController(query: ChannelListQuery) -> ChatChannelListController {
+        client.channelListController(query: query)
+    }
+}
+
+// MARK: Push Notifications
+
+extension StreamChatWrapper {
+    func setupRemoteNotifications() {
+        PushNotifications.shared.registerForPushNotifications()
```
evsaev, line 133: what if we pass `PushNotifications` instance as a dependency via `init`?

evsaev (approved):
Looks good 💪 Left some comments, mostly minor

DemoApp/StreamChat/Extensions/DeletedMessageVisibility+CustomStringConvertible.swift
bielikb: Im happy to do so, if there are no objections from the team.

bielikb: This small bit in our demo app pinpoints another deficiency in our LLC layer. 
Our models aren't `CustomDebugStringConvertible`, thus they can't be nicely interpreted by the LLDB.

DemoApp/DemoAppConfiguration.swift
bielikb: Thanks! This must've slipped through my fingers :) Done!

DemoApp/DemoAppCoordinator.swift
bielikb: Done!

DemoApp/StreamChat/Extensions/DeletedMessageVisibility+CustomStringConvertible.swift
bielikb: I moved that conformance to LLC

DemoApp/StreamChat/StreamChatWrapper.swift
bielikb: Thanks. I provided closure to fire this logic to `StreamChatWrapper` to which the outer clients can hook into.

DemoApp/PushNotifications.swift
bielikb: Done ;)

DemoApp/StreamChat/StreamChatWrapper.swift
bielikb: Good catch! Done ;)

PR #2067, opened 2022-06-07
[CIS-42] Markdown support for messages

hugobernalstream:
### 🔗 Issue Links

https://stream-io.atlassian.net/browse/CIS-42

### 🎯 Goal

Add Markdown support for messages.

### 📝 Summary

- Add Markdown formatting for messages with proper Markdown syntax.
- Add a feature flag to enable/disable Markdown support.

### 🛠 Implementation

- Implement an Adapter over the [SwiftyMarkdown](https://github.com/SimonFairbairn/SwiftyMarkdown) library for Markdown support.
- Add a Regex validation to check for Markdown syntax before formatting the text.

### 🎨 Showcase
Screen Shot 2022-06-06 at 22 26 22
Screen Shot 2022-06-06 at 22 26 40

performance tests for `containsMarkdown` and `format` methods:

Screen Shot 2022-06-12 at 19 10 12
Screen Shot 2022-06-12 at 19 11 00

### 🧪 Manual Testing Notes

Markdown support is enabled by default.

### ☑️ Contributor Checklist

### 🎁 Meme

Bold


Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
```diff
@@ -51,6 +51,14 @@ open class ChatMessageContentView: _View, ThemeProvider {
     /// When this value is set the subviews are instantiated and laid out just once based on
     /// the received options.
     public var layoutOptions: ChatMessageLayoutOptions?
+    
+    /// The formatter used for text Mardown
```
adamrushy, line 55: nit: **Markdown

adamrushy (commented):
Love the tests 😍 just a few questions

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
```diff
@@ -51,6 +51,14 @@ open class ChatMessageContentView: _View, ThemeProvider {
...
+    public var markdownFormatter: MarkdownFormatter = DefaultMarkdownFormatter()
```
evsaev, line 56: Instead of having 2 boolean flags + `markdownFormatter` in each message content view, I would just have a single optional formatter in `Components`. If it's nil, it means the markdown is disabled. At the same time, user can inject their own formatter if they want.
```swift
public struct Components {
 public var markdownFormatter: MarkdownFormatter? = DefaultMarkdownFormatter()
}
```

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
```diff
@@ -493,6 +501,11 @@ open class ChatMessageContentView: _View, ThemeProvider {
         textView?.font = textFont
         textView?.text = content?.textContent
         
+        if markdownFormatterEnabled, markdownFormatter.containsMarkdown(text: content?.textContent ?? "") {
+            let markdownText = markdownFormatter.format(from: content?.textContent ?? "")
+            textView?.attributedText = markdownText
+        }
```
evsaev, line 507: Let's assign either `attributedText` or `text`. Currently, we always assign `text` the live above and then conditionally override it.

Sources/StreamChatUI/Utils/MarkdownFormatter.swift
```diff
@@ -0,0 +1,46 @@
...
+
+public protocol MarkdownFormatter {
+    /// Checks for Markdown patterns in the given String.
+    /// - Parameter text: The string in which Markdown patters are going to be sought.
+    /// - Returns: Returns a Boolean value that indicates whether Markdown patters where found in the given String.
+    func containsMarkdown(text: String) -> Bool
+    /// Returns an attributed string form the given Markdown-formatted string.
+    /// - Parameter string: The string to be formatted.
+    /// - Returns: An attributed string with the corresponding formatted attributes.
+    func format(from string: String) -> NSAttributedString
```
evsaev, line 17:
```suggestion
    func format(_ string: String) -> NSAttributedString
```

Sources/StreamChatUI/Utils/MarkdownFormatter.swift
```diff
@@ -0,0 +1,46 @@
...
+}
+
+/// Default implementation for the Markdown formatter
+class DefaultMarkdownFormatter: MarkdownFormatter {
```
evsaev, line 21: How about making the default implementation `open` so integrators can subclass and reuse the existed logic if they want?

Sources/StreamChatUI/Utils/MarkdownFormatter.swift
```diff
@@ -0,0 +1,46 @@
...
+
+import Foundation
+import SwiftyMarkdown
+import UIKit
+
+public protocol MarkdownFormatter {
+    /// Checks for Markdown patterns in the given String.
+    /// - Parameter text: The string in which Markdown patters are going to be sought.
+    /// - Returns: Returns a Boolean value that indicates whether Markdown patters where found in the given String.
+    func containsMarkdown(text: String) -> Bool
```
evsaev, line 13:
```suggestion
    func containsMarkdown(_ string: String) -> Bool
```

Sources/StreamChatUI/Utils/MarkdownFormatter.swift
hugobernalstream: I wanted to check about this with you guys. Inheritance makes more sense for UI components that inherit from UIView - UIViewController but this is more of a formatting class. I used the `MarkdownFormatter` protocol because  commonly for Markdown support, similar as we are doing, integrators would probably use a 3rd party library which I think is best if they create their own abstraction wrapper class around that framework, using the `MarkdownFormatter` protocol so they can integrate it in the `Components.swift` class. But let me know if you still consider the inheritance approach better for consistency; in that case I'd better remove the protocol and probably rename the class removing the `Default` word and marking it as `open` as you suggested.

evsaev: The use-case I have in mind is a customer who knows that all messages contain a markdown, let's say. In that case, they would override `conatinsMarkdown` to always return `true` (to not run the regex and save time) but keep the default formatting. That's quite an edge case but there can be other scenarios when an integrator just wants to slightly tweak the default logic without integrating another 3rd party for markdown. We had some cases when default implementation was internal but we were asked to open it (e.g. `VideoLoading`). 

But I'm not pushing hard on it 🙂 Let's also ask what other guys think @GetStream/ios-developers

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
evsaev: Or maybe always use `attributedText` but with a font if markdown is not detected?

Sources/StreamChatUI/Utils/MarkdownFormatter.swift
```diff
@@ -0,0 +1,46 @@
...
+    /// - Returns: An attributed string with the corresponding formatted attributes.
+    func format(from string: String) -> NSAttributedString
+}
+
+/// Default implementation for the Markdown formatter
+class DefaultMarkdownFormatter: MarkdownFormatter {
+    enum Attributes {
+        enum Code {
+            static let fontName: String = "CourierNewPSMT"
+            static let color: UIColor = .red
```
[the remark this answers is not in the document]
evsaev: Let's ask how other frontend SDKs draw code snippets? 🙂 Great chance for us to not spend time on alignment later

Sources/StreamChatUI/Utils/MarkdownFormatter.swift
```diff
@@ -0,0 +1,45 @@
...
+        }
+    }
+    
+    private let markdownRegex: String =
```
evsaev, line 28: is this that very that will never fail and where the force unwrap is acceptable?
```suggestion
    private let markdownRegex = NSRegularExpression(
      pattern: "((?:\\`(.*?)\\`)|(?:\\*{1,2}(.*?)\\*{1,2})|(?:\\~{2}(.*?)\\~{2})|(?:\\_{1,2}(.*?)\\_{1,2})|^(>){1}|(#){1,6}|(=){3,10}|(-){1,3}|(\\d{1,3}\\.)|(?:\\[(.*?)\\])(?:\\((.*?)\\))|(?:\\[(.*?)\\])(?:\\[(.*?)\\])|(\\]\\:))+", 
      options: .anchorsMatchLines
     )!
```

Sources/StreamChatUI/Utils/MarkdownFormatter.swift
```diff
@@ -0,0 +1,46 @@
...
+        "((?:\\`(.*?)\\`)|(?:\\*{1,2}(.*?)\\*{1,2})|(?:\\~{2}(.*?)\\~{2})|(?:\\_{1,2}(.*?)\\_{1,2})|^(>){1}|(#){1,6}|(=){3,10}|(-){1,3}|(\\d{1,3}\\.)|(?:\\[(.*?)\\])(?:\\((.*?)\\))|(?:\\[(.*?)\\])(?:\\[(.*?)\\])|(\\]\\:))+"
+    
+    public func containsMarkdown(text: String) -> Bool {
+        guard let regex = try? NSRegularExpression(pattern: markdownRegex, options: .anchorsMatchLines) else {
+            return false
+        }
+
```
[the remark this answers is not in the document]
evsaev: If we go with what was suggested above the `guard` won't be needed

Sources/StreamChatUI/Utils/MarkdownFormatter.swift
nuno-vieira: @evsaev @tbarbugli Although having the class open is more easily customizable, this is not required to reuse our logic and still provide minor customizations. You can do this with composition:

```swift
class CustomMarkdownFormatter: MarkdownFormatter {
    let streamFormatter = DefaultMarkdownFormatter()
    
    public func containsMarkdown(text: String) -> Bool {
        return true
    }
    
    public func format(from string: String) -> NSAttributedString {
        return streamFormatter.format(string)
    }
}
```

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
nuno-vieira: I get your point @evsaev, but I actually think that having the flag might be easier to understand and easier to find what you need in this case 🤔

Sources/StreamChatUI/Utils/MarkdownFormatter.swift
```diff
@@ -0,0 +1,46 @@
...
+        enum Code {
+            static let fontName: String = "CourierNewPSMT"
+            static let color: UIColor = .red
+        }
+    }
+    
+    private let markdownRegex: String =
+        "((?:\\`(.*?)\\`)|(?:\\*{1,2}(.*?)\\*{1,2})|(?:\\~{2}(.*?)\\~{2})|(?:\\_{1,2}(.*?)\\_{1,2})|^(>){1}|(#){1,6}|(=){3,10}|(-){1,3}|(\\d{1,3}\\.)|(?:\\[(.*?)\\])(?:\\((.*?)\\))|(?:\\[(.*?)\\])(?:\\[(.*?)\\])|(\\]\\:))+"
+    
+    public func containsMarkdown(text: String) -> Bool {
```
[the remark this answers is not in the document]
nuno-vieira: They are already provided in the protocol, there's no need to duplicate 👍 I believe the documentation will be picked up from the protocol 🤔

Sources/StreamChatUI/Utils/MarkdownFormatter.swift
```diff
@@ -0,0 +1,45 @@
...
+        
+        return regex.numberOfMatches(in: text, range: .init(location: 0, length: text.utf16.count)) > 0
+    }
+    
+    public func format(from string: String) -> NSAttributedString {
+        let markdownFormatter = SwiftyMarkdown(string: string)
+        markdownFormatter.code.fontName = Attributes.Code.fontName
+        markdownFormatter.code.color = Attributes.Code.color
+        return markdownFormatter.attributedString()
+    }
```
nuno-vieira, line 44: What happens if the text doesn't contain a valid markdown? Do we actually need to check if the text contains markdown? Why don't we always apply the markdown?

Sources/StreamChatUI/Utils/MarkdownFormatter.swift
```diff
@@ -0,0 +1,45 @@
...
+        enum Code {
+            static let fontName: String = "CourierNewPSMT"
+            static let color: UIColor = .red
+        }
+    }
+    
+    private let markdownRegex: String =
+        "((?:\\`(.*?)\\`)|(?:\\*{1,2}(.*?)\\*{1,2})|(?:\\~{2}(.*?)\\~{2})|(?:\\_{1,2}(.*?)\\_{1,2})|^(>){1}|(#){1,6}|(=){3,10}|(-){1,3}|(\\d{1,3}\\.)|(?:\\[(.*?)\\])(?:\\((.*?)\\))|(?:\\[(.*?)\\])(?:\\[(.*?)\\])|(\\]\\:))+"
+    
+    public func containsMarkdown(text: String) -> Bool {
```
nuno-vieira, line 31: What happens if the text doesn't contain a valid markdown? Do we actually need to check if the text contains markdown? Why don't we always apply the markdown?

Tests/StreamChatUITests/Utils/DefaultMarkdownFormatter_Tests.swift
```diff
@@ -0,0 +1,252 @@
...
+                XCTAssertEqual(expectedLinkAttributedSubstring, attributedString.attributedSubstring(from: range).string)
+                XCTAssertEqual(expectedLinkURL, url)
+            } else if let paragraphStyleAttribute = attributes[.paragraphStyle] as? NSParagraphStyle,
+                      paragraphStyleAttribute.headIndent > 0 {
+                XCTAssertEqual(
+                    true,
+                    expectedUnorderedListedSubstrings.contains(attributedString.attributedSubstring(from: range).string)
+                )
+            }
+        }
```
nuno-vieira, line 250: These tests are a bit too complex 🤔  One easy way to simplify this and to make it more robust is to actually use a snapshot test here. We can use a snapshot test in AttributedStrings as well 🙂 We just need to introduce a wrapper in AssertSnapshot.swift like this:
```swift
func AssertSnapshot(
    attributedString: NSAttributedString,
    record: Bool = false,
    line: UInt = #line,
    file: StaticString = #filePath,
    function: String = #function
) {
    assertSnapshot(
        matching: attributedString,
        as: .description,
        record: record,
        file: file,
        line: line
    )
}
```

Sources/StreamChatUI/Utils/MarkdownFormatter.swift
```diff
@@ -0,0 +1,45 @@
...
+        }
+        
+        return regex.numberOfMatches(in: text, range: .init(location: 0, length: text.utf16.count)) > 0
+    }
+    
+    public func format(from string: String) -> NSAttributedString {
+        let markdownFormatter = SwiftyMarkdown(string: string)
+        markdownFormatter.code.fontName = Attributes.Code.fontName
+        markdownFormatter.code.color = Attributes.Code.color
+        return markdownFormatter.attributedString()
```
nuno-vieira, line 43: Don't we provide a way for the customer to customize the styling? 🤔 Or is this out of the scope of this PR? 

If we add this later tho, my concern is that if we add a new method to the Formatter protocol to include a styling options struct, it will be a breaking change 🤔 This is a con of using protocols in SDKs, is that if you add a new method, it is a breaking change.

.swiftlint.yml
```diff
@@ -5,6 +5,7 @@ excluded:
   - Dependencies
   - Sources/StreamChatUI/Generated
   - Sources/StreamChatUI/StreamSwiftyGif
+  - Sources/StreamChatUI/StreamSwiftyMarkdown
```
nuno-vieira, line 8: We need. to update the documentation to add the new dependency here: https://getstream.io/chat/docs/sdk/ios/#dependencies 🙂

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
hugobernalstream: I decided to do it that way because whenever you set `attributedString`  and if you have values in `text` , `textColor`  or `textFont` , they’ll get overridden as mentioned on Apple’s docs. But I see your point and so I will wrap this with an if statement; if there is markdown, set the `attributedString`  else set text normally as we currently have.
The reason I’m not setting attributedText  on all cases even if there isn’t markdown syntax in the `String` is because I would have to wrap the String around a `NSMutableAttributedString`  first, add the existing textFont  and textColor  properties as additional attributes and then assign. Feels more verbose IMHO and there is no problem if either text  or attributedText  are set, but only one of them.

Sources/StreamChatUI/Utils/MarkdownFormatter.swift
hugobernalstream: Not the case though. As you can see in the guard statement in containsMarkdown() the NSRegularExpression  init throws because it indeed can fail so is easier to treat it as an optional with `try?` .

Sources/StreamChatUI/Utils/MarkdownFormatter.swift
```diff
@@ -0,0 +1,45 @@
...
+class DefaultMarkdownFormatter: MarkdownFormatter {
+    enum Attributes {
+        enum Code {
+            static let fontName: String = "CourierNewPSMT"
+            static let color: UIColor = .red
+        }
+    }
+    
+    private let markdownRegex: String =
+        "((?:\\`(.*?)\\`)|(?:\\*{1,2}(.*?)\\*{1,2})|(?:\\~{2}(.*?)\\~{2})|(?:\\_{1,2}(.*?)\\_{1,2})|^(>){1}|(#){1,6}|(=){3,10}|(-){1,3}|(\\d{1,3}\\.)|(?:\\[(.*?)\\])(?:\\((.*?)\\))|(?:\\[(.*?)\\])(?:\\[(.*?)\\])|(\\]\\:))+"
```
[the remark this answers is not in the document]
hugobernalstream: Yes, I tried simpler approaches with Swift String API but considered that a proper validation of the cases was a good idea and that was better to do with regex. Unfortunately none of the libraries I tested had their own way to check for Markdown syntax public and we do need a way to validate that the String contains regex syntax because we have other situations where text font and color changes depending on if is Jumbo-emoji, is a deleted message, etc. Also I did a quick performance test with the `measure {}`  test you suggested  and for ~1000 times of calling `containsMarkdown()`  (`the measure {`} test does this 10 times ) and validating with the regex, for a fairly long String with several regex elements took around ~0.1s. Let me know what you think.

Tests/StreamChatUITests/Utils/DefaultMarkdownFormatter_Tests.swift
```diff
@@ -0,0 +1,252 @@
...
+        let stringWithNoMarkdown = "Hello, This is a test String"
+
+        // WHEN
+        let containsMarkdown = sut.containsMarkdown(text: stringWithNoMarkdown)
+
+        // THEN
+        XCTAssertEqual(false, containsMarkdown)
+    }
+    
+    func test_containsMarkdown_whenCheckForItalicEmphasis_thenReturnsTrue() {
```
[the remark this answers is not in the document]
hugobernalstream: It does seem kind of verbose, but since they are unit tests, is ideal to test just one specific behavior. These are to make sure that we detect each of the markdown symbols properly. 
With the comments you refer to the // Given, When, Then? This is the approach we are following with the team for unit tests.
Since we have the validation, empty text doesn’t even get to be formatted with the MarkdownFormatter nor we have invalid markdown. Basically with the regex validation, incomplete or invalid Regex syntax is just set to the textView as it is.

Tests/StreamChatUITests/Utils/DefaultMarkdownFormatter_Tests.swift
hugobernalstream: I added the test_format_whenStringContainsManyMarkdowns_thenAttributedStringIncludesAllAttibutes() test to have some verification that the returned NSAttributedString  was returning proper formatting and unfortunately that’s Apple way to test that :sweat_smile:. Ugly but is still valid. The idea of testing with a Snapshot is really good I’m gonna try that one.

Sources/StreamChatUI/Utils/MarkdownFormatter.swift
hugobernalstream: Although the SwiftyMarkdown library does offer some level of customization, and since the base idea was to do a similar implementation as Android (feature parity), opening for customization wasn’t considered. Will check with the rest of the UIKit squad to see what they think. Since I will be going for the open/inheritance approach it won’t be a braking change anymore if we want to add a new method.

Sources/StreamChatUI/Utils/MarkdownFormatter.swift
bielikb: > Got it 👍 . Will open it then. Also will get rid of the protocol since it won’t be needed for the open class approach.

@hugobernalstream don't get rid of the `protocol` itself. That was not implied here. 
We should always prefer `composition` over `inheritance` where possible. 
Based on the comments, we're trying to go with a mixed approach, where:
1. We keep the protocol in place, so that integrator can provide his own `MarkDownFormatter` - this is great 👍
2. We turn on the inheritance on the default implementation

FYI, we took similar approach with [ChannelNameFormatter](https://github.com/GetStream/stream-chat-swift/blob/61b87c7c91fe5377c80e1a851242f1d767ea4523/Sources/StreamChatUI/Appearance%2BFormatters/ChannelNameFormatter.swift).

nuno-vieira: This also adds a new question, is this considered a formatter? If yes, we should add this to the `appearance.formatters`. Otherwise, it doesn't make sense to name MarkdownFormatter, and having it in a different place then the formatters. Since indeed this is data formatting,  this should go to `appearance.formatters`

Sources/StreamChatUI/Components.swift
```diff
@@ -123,6 +123,9 @@ public struct Components {
     /// Note: This is currently an experimental feature that we are actively
     /// working on and testing to make sure it is stable.
     public var _messageListDiffingEnabled = false
+    
+    /// A boolean value that determines whether Markdown is active for messages to be formatted.
+    public var markdownFormatterEnabled = true
```
nuno-vieira, line 128: Better to move this flag close the formatter

Sources/StreamChatUI/Components.swift
```diff
@@ -327,6 +330,9 @@ public struct Components {
     /// A view that displays the video attachment preview in composer.
     public var videoAttachmentComposerPreview: VideoAttachmentComposerPreview
         .Type = VideoAttachmentComposerPreview.self
+    
+    /// A formatter used for text Markdown
+    public var markdownFormatter: MarkdownFormatter.Type = DefaultMarkdownFormatter.self
```
nuno-vieira, line 335: In fact, the `markdownFormatter` and `markdownFormatterEnabled` should be in the `Appearance.formatters` 🙂

Tests/StreamChatUITests/Utils/DefaultMarkdownFormatter_Tests.swift
nuno-vieira: Testing with snapshot should work 👍 But we are actually testing a third-party framework, we shouldn't be testing third-party dependencies. Either way, since with snapshot testing this should be easy, we can add the test to make sure if we change the dependency, it still works 👍

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
hugobernalstream: Done

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
hugobernalstream: Keep the bool for consistency as recommended by Nuno.

Sources/StreamChatUI/Utils/MarkdownFormatter.swift
hugobernalstream: Done

Sources/StreamChatUI/Utils/MarkdownFormatter.swift
hugobernalstream: Kept both approaches and moved the formatter to Appearance+Formatters as suggested.

Sources/StreamChatUI/Utils/MarkdownFormatter.swift
hugobernalstream: Done

Sources/StreamChatUI/Utils/MarkdownFormatter.swift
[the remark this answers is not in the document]
hugobernalstream: Added screenshots with performance measures on this PR. Please take a look.

Sources/StreamChatUI/Components.swift
hugobernalstream: Done

Sources/StreamChatUI/Components.swift
hugobernalstream: Done 🙂

evsaev (approved):
LGTM!

docusaurus/docs/iOS/uikit/data-formatting.md
```diff
@@ -76,3 +76,49 @@ Appearance.default.formatters.channelName = CustomChannelNameFormatter()
 ```
 
 In the example above, if the channel has only the current user as a member, it uses the name of the current user as the channel name, otherwise, the default logic is used.
+
+## Markdown
+
+The SDK offers Markdown formatting out of the box for channel messages. It supports the most common Markdown syntax; _italic_, **bold**, ~~strikethrough~~, `code`, Headings, Links, etc. It uses [SwiftyMarkdown](https://github.com/SimonFairbairn/SwiftyMarkdown) library internally.
```
bielikb, line 82:
```suggestion
The SDK offers markdown formatting out of the box for channel messages. 
It supports the most common markdown syntax:
* _italic_
* **bold**
* ~~strikethrough~~
* `code`
* headings and subheadings
* links, etc. 

It uses [SwiftyMarkdown](https://github.com/SimonFairbairn/SwiftyMarkdown) library internally.
```

docusaurus/docs/iOS/uikit/data-formatting.md
```diff
@@ -76,3 +76,49 @@ Appearance.default.formatters.channelName = CustomChannelNameFormatter()
...
+
+<center><img src={require("../assets/markdown-formatting.png").default} width="60%" height="60%"/></center>
+
+Mardown support is enabled by default. You can disable it by setting the following flag to false in the `Appearance` config:
+
+```swift
+Appearance.default.formatters.markdownFormatterEnabled = false
```
bielikb, line 89: Looking at the API right now, by the Apple's code guidelines for `Bool` types we could rename the flag to:
`isMarkdownEnabled`

docusaurus/docs/iOS/uikit/data-formatting.md
```diff
@@ -76,3 +76,49 @@ Appearance.default.formatters.channelName = CustomChannelNameFormatter()
...
+```
+
+You can also provide you own Markdown implementation either by subclassing our `DefaultMarkdownFormatter` class or by adopting the `MarkdownFormatter` protocol:
```
bielikb, line 92: Could we add header to this section? 
Something like:
`Customizing the markdown formatter`

bielikb: `You can customize the default markdown implementation by subclassing the `DefaultMarkdownFormatter` class or by providing your own implementation of the `MarkdownFormatter` protocol:`

bielikb (commented):
Just a few comments related to docs + API

docusaurus/docs/iOS/uikit/data-formatting.md
hugobernalstream: Adjusted

docusaurus/docs/iOS/uikit/data-formatting.md
hugobernalstream: That makes more sense. Adjusted.

docusaurus/docs/iOS/uikit/data-formatting.md
hugobernalstream: Adjusted.

bielikb (approved):

PR #2070, opened 2022-06-07
CIS-1815 - Removing force_casts and force_unwraps, force_trys

adamrushy:
### 🔗 Issue Links

CIS-1815

### 🎯 Goal

We have adopted SwiftLint and we decided to separate the changes for force_cast and force_unwrap inside this ticket due to how large the changes are. This PR enables this rule and removes all usage across the SDK + Example apps.

### 📝 Summary

Removing all force unwraps and casts, the biggest changes are surrounding TableView's and CollectionView cells, and the decision is to either default to a blank cell, assert or blank cell with logging an error.

### ☑️ Contributor Checklist


Examples/SlackClone/SlackComposerVC.swift
```diff
@@ -94,10 +94,12 @@ open class ComposerEmojiSuggestionsDataSource: NSObject, UICollectionViewDataSou
...
 
     public func collectionView(_ collectionView: UICollectionView, cellForItemAt indexPath: IndexPath) -> UICollectionViewCell {
-        let cell = collectionView.dequeueReusableCell(
+        guard let cell = collectionView.dequeueReusableCell(
             withReuseIdentifier: ComposerEmojiSuggestionsCollectionViewCell.reuseId,
             for: indexPath
-        ) as! ComposerEmojiSuggestionsCollectionViewCell
+        ) as? ComposerEmojiSuggestionsCollectionViewCell else {
+            return UICollectionViewCell()
+        }
```
evsaev, line 102: How do you feel about introducing `Reusable` type for both table & collection views? 🙂 The iOS community is familiar with this approach and for us, it would allow to remove the boilerplate and make code easy to read.

```swift
protocol Reusable {
    static var reuseIdentifier: String { get }
}

extension Reusable {
    static var reuseIdentifier: String {
        String(describing: self)
    }
}

extension UITableView {
    func register<Cell: UITableViewCell>(_ type: Cell.Type) where Cell: Reusable {
        register(type, forCellReuseIdentifier: type.reuseIdentifier)
    }
    
    func dequeueReusableCell<Cell: UITableViewCell>(_ type: Cell.Type, for indexPath: IndexPath) -> Cell where Cell: Reusable {
        // here it's acceptable to disable the force-case rule, i think
        return dequeueReusableCell(withIdentifier: type.reuseIdentifier, for: indexPath) as! Cell
    }
}
```

evsaev, line 102: How do you feel about introducing `Reusable` type for both table & collection views? 🙂 The iOS community is familiar with this approach and for us, it would allow to remove the boilerplate and make code easy to read.

```swift
protocol Reusable {
    static var reuseIdentifier: String { get }
}

extension Reusable {
    static var reuseIdentifier: String {
        String(describing: self)
    }
}

extension UITableView {
    func register<Cell: UITableViewCell>(_ type: Cell.Type) where Cell: Reusable {
        register(type, forCellReuseIdentifier: type.reuseIdentifier)
    }
    
    func dequeueReusableCell<Cell: UITableViewCell>(_ type: Cell.Type, for indexPath: IndexPath) -> Cell where Cell: Reusable {
        // here it's acceptable to disable the force-case rule, i think
        return dequeueReusableCell(withIdentifier: type.reuseIdentifier, for: indexPath) as! Cell
    }
}
```

Examples/YouTubeClone/Composer/YTChatComposerViewController.swift
```diff
@@ -9,13 +9,17 @@ import UIKit
...
+    var ytMessageComposerView: YTChatMessageComposerView? {
+        composerView as? YTChatMessageComposerView
     }
     
     override func setUp() {
         super.setUp()
 
+        guard let ytMessageComposerView = ytMessageComposerView else {
+            return
+        }
```
evsaev, line 21: In that case, I would be using `if` instead of the early exist because if we put more code for other UI components but this guard fails, we will never reach the code for other components

evsaev, line 21: In that case, I would be using `if` instead of the early exist because if we put more code for other UI components but this guard fails, we will never reach the code for other components

Examples/SlackClone/SlackComposerVC.swift
adamrushy: Yeah, it's a good question - I personally would be happy with this... my biggest question really was what is the ideal approach here should we force unwrap in general on cells and if the answer is no then do we just return empty cell or something else (log, assert, etc...)

Examples/YouTubeClone/Composer/YTChatComposerViewController.swift
adamrushy: Yep good suggestion 👍🏼

Examples/SlackClone/SlackComposerVC.swift
evsaev: Yeah, that's not a conditional failure, that's a dev mistake if the force case of the cell fails so someone checking the DemoApp should never experience these error (unless they don't change sources). I honestly think `as!` was fine but if we encapsulate it in the nice new dequeue API we apply the logging for all dequeues as once.

func dequeueReusableCell<Cell: UITableViewCell>(_ type: Cell.Type, for indexPath: IndexPath) -> Cell where Cell: Reusable {
        // here it's acceptable to disable the force-case rule, i think
        guard let cell = dequeueReusableCell(withIdentifier: type.reuseIdentifier, for: indexPath) as? Cell else {
         log.error()
         return Cell.init()
        }
        return cell
    }

bielikb: I like the above suggestion from @evsaev

bielikb (approved):

Sources/StreamChatUI/Utils/Reusable+Extensions.swift
```diff
@@ -0,0 +1,58 @@
...
+    }
+    
+    /// Dequeues a `UITableView` cell with a generic type and indexPath
+    /// - Parameters:
+    ///   - type: A generic cell type
+    ///   - indexPath: The indexPath of the row in the UITableView
+    /// - Returns: A cell from the type passed through
+    public func dequeueReusableCell<Cell: UITableViewCell>(with type: Cell.Type, for indexPath: IndexPath) -> Cell {
+        dequeueReusableCell(withIdentifier: type.reuseIdentifier, for: indexPath) as! Cell
+    }
```
evsaev, line 35: I'm not sure having reusable helpers in SDK is a good idea as it can lead to collision with other 3rd party/customer's implementation. I was thinking of adding this just to the `DemoApp` (the same for storyboard helpers)

Sources/StreamChatUI/ChatMessageList/ChatMessageListView.swift
```diff
@@ -107,10 +107,12 @@ open class ChatMessageListView: UITableView, Customizable, ComponentsProvider {
...
         }
 
-        let cell = dequeueReusableCell(
+        guard let cell = dequeueReusableCell(
             withIdentifier: reuseIdentifier,
             for: indexPath
-        ) as! ChatMessageCell
+        ) as? ChatMessageCell else {
+            return ChatMessageCell()
+        }
```
evsaev, line 115: Can we leverage generic `dequeue` here?

Sources/StreamChatUI/Gallery/GalleryVC.swift
```diff
@@ -342,10 +342,12 @@ open class GalleryVC: _ViewController,
...
         }
         
-        let cell = collectionView.dequeueReusableCell(
+        guard let cell = collectionView.dequeueReusableCell(
             withReuseIdentifier: reuseIdentifier,
             for: indexPath
-        ) as! GalleryCollectionViewCell
+        ) as? GalleryCollectionViewCell else {
+            return UICollectionViewCell()
+        }
```
evsaev, line 350: The same, would generic dequeue work here as well?

Sources/StreamChatUI/Utils/UIStoryboard+Extensions.swift
```diff
@@ -0,0 +1,26 @@
+//
+// Copyright © 2022 Stream.io Inc. All rights reserved.
+//
+
+import UIKit
+
+// swiftlint:disable force_cast
+extension UIStoryboard {
```
evsaev, line 8: Can be removed since we don't leverage storyboards in the SDKs

Sources/StreamChatUI/Gallery/GalleryVC.swift
adamrushy: Hmm good spot, so this also changes things a little because we change cell type based on the index path contents, I can introduce an optional param to the reusable func that takes the reuseId in these situations, WDYT?

Sources/StreamChatUI/Utils/UIStoryboard+Extensions.swift
adamrushy: Good to keep for the future?

evsaev: it decreases the test coverage and we don't have plans to leverage storyboards internally. i'd remove but leaving it up to you

PR #2076, opened 2022-06-08
Revert ListChangeAggregator newIndexPath on Updates

nuno-vieira:
### 🔗 Issue Links
None

### 🎯 Goal
Switching the `ListChangeAggregator` to the `newIndexPath` caused multiple crashes. In Apple documentation, they use the `oldIndexPath`, so this PR reverts that change and also stops checking for conflicts between Inserts and Updates to avoid unnecessary reload data.

### 📝 Summary
- Reverts the ListChangeAggregator Update IndexPath. Switches back to the oldIndexPath.
- Don't treat Insert<->Updates has conflicts in `ListChangeIndexPathResolver`

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
+++ b/Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
@@ -245,22 +245,11 @@ open class ChatMessageListVC: _ViewController,
             // Because we use an inverted table view, we need to avoid animations since they look odd.
             UIView.performWithoutAnimation {
                 listView.updateMessages(with: changes) { [weak self] in
-                    guard let self = self else {
-                        completion?()
-                        return
-                    }
-
                     if let newMessageInserted = changes.first(where: { $0.isInsertion && $0.indexPath.row == 0 })?.item {
-                        if self.dataSource?.numberOfMessages(in: self) ?? 0 > 1 {
-                            let previousMessageIndexPath = IndexPath(row: 1, section: 0)
-                            self.listView.reloadRows(at: [previousMessageIndexPath], with: .none)
-                        }
-
                         if newMessageInserted.isSentByCurrentUser {
-                            self.listView.scrollToMostRecentMessage()
+                            self?.listView.scrollToMostRecentMessage()
                         }
                     }
-
                     completion?()
                 }
             }
--- a/Sources/StreamChatUI/Utils/ListChangeUpdater/ListChangeIndexPathResolver.swift
+++ b/Sources/StreamChatUI/Utils/ListChangeUpdater/ListChangeIndexPathResolver.swift
@@ -33,52 +33,63 @@ final class ListChangeIndexPathResolver {
     /// - Parameters:
     ///   - changes: changes
     /// - Returns: Returns the indices mapped to sets. Returns nil if there were conflicts.
-    func mapToSetsOfIndexPaths<Item>(
+    func resolve<Item>(
         changes: [ListChange<Item>]
     ) -> Indexes? {
-        var allIndexes = Set<IndexPath>()
+        var moveFromIndexes = Set<IndexPath>()
+        var moveToIndexes = Set<IndexPath>()
         var moveIndexes = Set<IndexPathMove>()
         var insertIndexes = Set<IndexPath>()
         var removeIndexes = Set<IndexPath>()
         var updateIndexes = Set<IndexPath>()
         
-        var hasConflicts = false
-        let verifyConflict = { (indexPath: IndexPath) in
-            let (inserted, _) = allIndexes.insert(indexPath)
-            hasConflicts = !inserted || hasConflicts
-        }
-        
         for change in changes {
-            if hasConflicts {
-                break
-            }
-            
             switch change {
             case let .insert(_, index):
-                verifyConflict(index)
                 insertIndexes.insert(index)
             case let .move(_, fromIndex, toIndex):
-                verifyConflict(fromIndex)
-                verifyConflict(toIndex)
                 moveIndexes.insert(IndexPathMove(fromIndex, toIndex))
+                moveFromIndexes.insert(fromIndex)
+                moveToIndexes.insert(toIndex)
             case let .remove(_, index):
-                verifyConflict(index)
                 removeIndexes.insert(index)
             case let .update(_, index):
-                verifyConflict(index)
                 updateIndexes.insert(index)
             }
         }
-        
-        if hasConflicts {
-            return nil
-        }
-        
-        return (
+
+        let indexes = Indexes(
             move: moveIndexes,
             insert: insertIndexes,
             remove: removeIndexes,
             update: updateIndexes
         )
+
+        // Check if there are conflicts between Inserts<->Moves<->Deletes or Updates<->Moves<->Deletes.
+        // We don't check for conflicts between Inserts<->Updates, since it is not a conflict.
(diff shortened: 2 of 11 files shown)


Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -245,22 +245,11 @@ open class ChatMessageListVC: _ViewController,
...
-                    guard let self = self else {
-                        completion?()
-                        return
-                    }
-
                     if let newMessageInserted = changes.first(where: { $0.isInsertion && $0.indexPath.row == 0 })?.item {
-                        if self.dataSource?.numberOfMessages(in: self) ?? 0 > 1 {
-                            let previousMessageIndexPath = IndexPath(row: 1, section: 0)
-                            self.listView.reloadRows(at: [previousMessageIndexPath], with: .none)
-                        }
```
evsaev, line 257: We should manually reload the previous cell for any insert and delete (except maybe when the new message is sent by the current user). Imagine a chat participant sending 2 messages and then hard deleting the last one. In that case, we should reload his previous message to show avatar etc.

Sources/StreamChatUI/Utils/ListChangeUpdater/TableViewListChangeUpdater.swift
```diff
@@ -22,39 +22,23 @@ final class TableViewListChangeUpdater: ListChangeUpdater {
...
+        tableView?.performBatchUpdates({
+            tableView?.deleteRows(at: Array(indices.remove), with: .none)
+            tableView?.insertRows(at: Array(indices.insert), with: .none)
+            tableView?.reloadRows(at: Array(indices.update), with: .none)
+            indices.move.forEach {
+                tableView?.moveRow(at: $0.fromIndex, to: $0.toIndex)
+            }
+        }, completion: { finished in
+            completion?(finished)
+        })
```
evsaev, line 42: happy to see it getting into the good shape 🎉

Sources/StreamChatUI/Utils/ListChangeUpdater/ListChangeIndexPathResolver.swift
```diff
@@ -33,42 +33,49 @@ final class ListChangeIndexPathResolver {
...
+        let insertsMovesRemoves = insertIndexes.union(movesAndRemoves)
+        if insertsMovesRemoves.count != insertIndexes.count + totalCountOfMovesAndDeletes {
+            hasConflicts = true
+        }
+
+        // Check if there are conflicts between Updates<->Moves<->Deletes
+        let updatesMovesRemoves = updateIndexes.union(movesAndRemoves)
+        if updatesMovesRemoves.count != updateIndexes.count + totalCountOfMovesAndDeletes {
+            hasConflicts = true
+        }
```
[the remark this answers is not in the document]
nuno-vieira: It's taken from my 🧠 😂

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
nuno-vieira: I see 👍 I'll apply that fix tomorrow and test it 🙂

Sources/StreamChatUI/Utils/ListChangeUpdater/ListChangeIndexPathResolver.swift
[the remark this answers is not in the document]
nuno-vieira: This is basically making a union of the sets, and checking if the `count` was not changed, if it was not, then it means there are no conflicts. Making an intersection of the 3 sets would also work, but I don't think that is possible in swift easily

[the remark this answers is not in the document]
evsaev: I wonder if we still need conflict resolution logic? We care a lot about moves but the only case when it happens is sending a pending giphy with a delay when other messages are inserted below it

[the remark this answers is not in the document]
nuno-vieira: @evsaev I've made this much simpler now :) Please let me know if not its easier to read!

Yeah we probably don't need this anymore, but let's keep it for extra safety. I've also added a log to check if we encounter this any more times.

Sources/StreamChatUI/Utils/ListChangeUpdater/ListChangeIndexPathResolver.swift
```diff
@@ -33,52 +33,63 @@ final class ListChangeIndexPathResolver {
...
             move: moveIndexes,
             insert: insertIndexes,
             remove: removeIndexes,
             update: updateIndexes
         )
+
+        // Check if there are conflicts between Inserts<->Moves<->Deletes or Updates<->Moves<->Deletes.
+        // We don't check for conflicts between Inserts<->Updates, since it is not a conflict.
+        let movesAndRemoves = [moveFromIndexes, moveToIndexes, removeIndexes]
+        let hasInsertConflicts = insertIndexes.containsDuplicates(between: movesAndRemoves)
```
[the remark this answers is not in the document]
nuno-vieira: That is not a valid combination according to Apple, but tomorrow we can test this 👍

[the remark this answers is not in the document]
nuno-vieira: I'm pretty sure that will be resolved as an update. Which in fact is a remove and insert

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
nuno-vieira: This is not needed, it is working fine without any change 🙂

Sources/StreamChatUI/Utils/ListChangeUpdater/ListChangeIndexPathResolver.swift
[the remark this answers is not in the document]
nuno-vieira: I was testing this, but couldn't. There's currently a backend bug that stops me from testing this so lets wait for tomorrow

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
evsaev: How so? It works fine only for the new message case because previous message stops being a channel preview and receives and update. But it removal happens in the middle of the list, there’s no other way but to manually reload

nuno-vieira: I see what you mean. But we didn't have this working previously, right? Should we create an issue for this? Just want to try and avoid so many changes. This is a small edge case anyway

evsaev: We were handling this case, not sure when we stopped doing that

evsaev (approved):

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
evsaev: That's a regression introduced several releases ago. Agreed to have a ticket and fix it later because this can potentially lead to more crashes which would be more disappointing for customers then having this edge-case uncovered for 2 more weeks.

PR #2084, opened 2022-06-13
Improve JSON parsing performance

b-onc:
### 🎯 Goal

Our JSON decoding has been slow since forever. A 3MB JSON payload from a realistic app takes around 2.3 seconds to parse. It's affecting the UX badly.

### 📝 Summary

We don't need to have custom date decoding for our JSON files, Apple already has one built-in with much better performance.

### 🛠 Implementation

I've attended a WWDC22 lab "Performance" and asked the Apple engineer why our Codable implementation would be this slow. He was puzzled to hear it took around 2 seconds to parse a 3MB JSON file, it should be around 200-300ms he said. Apple has tests for Codable and they're not aware of this performance issue.

With their help, we've debugged our decoding logic and found out that 1.8 seconds from that 2.3 seconds is actually spend on `DateFormatter.date(from: String)`. We've looked into our logic and found out that we're doing custom stuff for the standard dates.

Stream uses RFC3339 (ISO8601) dates. Codable's default `.iso8601` date decoding strategy uses `ISO8601DateFormatter.Options.withInternetDateTime`, which looks like:
> 2020-06-09T08:10:40Z

so it doesn't include miliseconds. But RFC3339 dates can also look like:

> 2020-06-09T08:10:40.800912Z

which stream uses.

Trying to parse the first date with `.iso8601` parsing fails, expectedly. A workaround is to use a custom date decoding _with_ default `ISO8601DateFormatter` with options `[,withInternetDateTime, .withFractionalSeconds]`, which parses the date our backend is sending perfectly.

As an extra safety feature, I've left the code we've been using to decode dates, in case our new decoding logic fails. This doesn't have any performance implication since it doesn't run unless it's needed.

The performance test runs for 10 iterations so it takes ~9 seconds. I've disabled it for now, potentially we can enable it for a smaller JSON, since performance will scale and we'll be able to catch issues.

### 🎨 Showcase

image
The first iteration always takes longer since `DateFormatter` is created within it.

### ☑️ Contributor Checklist


bielikb (approved):
🙌 thanks!

b-onc:
Failing test is known to be flaky so i'm merging this

Issue #2091, opened 2022-06-15
Crash on direct chat, force unwrap on channelController.cid!


michallaskowski:
## What did you do?
Tried to open a direct chat with a user.

## What did you expect to happen?
Chat to be opened.

## What happened instead?
App crashes on `messageComposerVC.channelController = client.channelController(for: channelController.cid!)`
Line 75, ChatChannelVC.swift
`channelController.cid` is nil, and force unwrap crashes the app.

## GetStream Environment
**GetStream Chat version:** 4.16.0
**GetStream Chat frameworks:** StreamChat, StreamChatUI
**iOS version:** 15.4.1
**Swift version:** Swift 5.6
**Xcode version:** 13.4.1
**Device:** iPhone 13

## Additional context
For some reason seems to be working fine on our production builds (TestFlight), fails on Staging builds (TestFlight) and local builds from Xcode using the same Production scheme, or Staging scheme.

b-onc:
Hello @michallaskowski ,
Can you share the full code?

polqf:
Hi @michallaskowski, we just released [4.17.0](https://github.com/GetStream/stream-chat-swift/releases/tag/4.17.0) which contains a fix for the crash you reported.

Let us know if we can help you with something else

PR #2097, opened 2022-06-17
Fix ChatChannelVC crash for DM channels

b-onc:
### 🔗 Issue Links

#2091 
https://getstream.slack.com/archives/C01FT5767AS/p1655306508062099

### 🎯 Goal

To prevent the crash

### 📝 Summary

`ChatChannelVC` has this line in `setUp` func:
```swift
messageComposerVC.channelController = client.channelController(for: channelController.cid!)
```
SDK assumes that at this point, `channelController` will have a valid `cid`, but there's no code path to guarantee `synchronize` to be called before this line. So, if a controller doesn't have a valid `cid`, (which is the case if it's been created with `createDirectMessageChannelWith` factory), and `synchronize` hasn't been called on it yet.

### 🛠 Implementation

`ChannelVC` already calls `synchronize` on the controller, but only after this line. So we should only assign a new controller to composer when we're sure we have the valid `cid`

### 🧪 Manual Testing Notes

Use this patch file in the DemoApp and run the app:
[ChatChannelVC_Crash_Reproduce.patch.zip](https://github.com/GetStream/stream-chat-swift/files/8926861/ChatChannelVC_Crash_Reproduce.patch.zip)

### ☑️ Contributor Checklist


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Fix last message of hidden channel with `clearHistory` visible in channel list [#2056](https://github.com/GetStream/stream-chat-swift/issues/2056)
 - Message action title now supports displaying 2 lines of text instead of 1 [#2082](https://github.com/GetStream/stream-chat-swift/pull/2082)
 - Fix Logger persisting config after usage, preventing changing parameters (such as LogLevel) [#2081](https://github.com/GetStream/stream-chat-swift/issues/2081)
+- Fix crash in `ChannelVC` when it's initialized using a `ChannelController` created with `createDirectMessageChannelWith` factory [#2095](https://github.com/GetStream/stream-chat-swift/issues/2095)
```
evsaev, line 15:
```suggestion
- Fix crash in `ChannelVC` when it's initialized using a `ChannelController` created with `createDirectMessageChannelWith` factory [#2097](https://github.com/GetStream/stream-chat-swift/issues/2097)
```

b-onc: I really need to fix our `get_next_issue_number` lane :D

Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
```diff
@@ -71,12 +71,23 @@ open class ChatChannelVC: _ViewController,
...
-
-        messageComposerVC.channelController = client.channelController(for: channelController.cid!)
+        
         messageComposerVC.userSearchController = userSuggestionSearchController
+        
+        func setChannelControllerToComposerIfNeeded(cid: ChannelId?) {
+            guard messageComposerVC.channelController == nil else { return }
+            let composerChannelController = channelController.cid.map { client.channelController(for: $0) }
+            messageComposerVC.channelController = composerChannelController
+        }
```
evsaev, line 81: I'd make it `private` in `ChatChannelVC` scope to keep the lifecycle methods short

evsaev (approved):
Looks good!

PR #2101, opened 2022-06-17
[CIS-1900] Push Notifications - multi bundle support

bielikb:
### 🔗 Issue Links

CIS-1900

### 🎯 Goal

Allow integrator to leverage multi-bundle feature via exposing new parameter to `addDevice` API called `provider name`.
Provider name represents the name of the push configuration created in Stream dashboard.
Push configurations are located under `Push Notifications` tab in the dashboard.
 
Multi-bundle support allows the integrator to list multiple client apps under one Stream app in the dashboard. (eg. when leveraging `Staging` vs. `Production` setup).
The default configuration will be used if no `provider name` is specified.

Available API:
APN:
```
client?.currentUserController().addDevice(.apn(token: deviceToken, providerName: "Name of the push configuration listed in Stream's Dashboard"))
```

Firebase:
```
client?.currentUserController().addDevice(.firebase(token: deviceToken, providerName: "Name of the push configuration listed in Stream's Dashboard"))
```

### 📝 Changes

* New optional `providerName` parameter can be specified for both `apn` and `firebase` integrations.
* Existing `provider` parameter in the `PushDevice` struct was renamed to `pushProvider` to gain API alignment with Android.
* In order to have a way to test the to multi-bundle support, new `DemoApp` target was added to the [integration apps repo](https://github.com/GetStream/stream-chat-swift-integration-apps/tree/multi-bundle) with bundle identifier `io.getstream.iOS.ChatDemoAppTwo`.
appstore

* Additionally, a new Info section was added to our `DemoApp` to clarify which push configuration is used by our default develop app.
Simulator

* `client` property inside `StreamChatWrapper` is no longer forced-cast.

# Important Note 📜
Docs will be added in separate PR, this PR contains only the  implementation part.

### 🧪 Manual Testing Notes

1. Build multi bundle app on `multi-bundle` branch in [integration apps repo](https://github.com/GetStream/stream-chat-swift-integration-apps/tree/multi-bundle)
3. Send a message to `Luke` from a different app/web
4. Verify the push notification is received.
IMG_1087

### ☑️ Contributor Checklist

### 🎁 Meme


Sources/StreamChat/Models/PushDevice.swift
```diff
@@ -6,25 +6,46 @@ import Foundation
...
+    /// Device token used to register a device for push notifications
     public let deviceId: DeviceId
-    public let provider: PushProvider
 
-    private init(deviceId: DeviceId, provider: PushProvider) {
+    /// Provider of the remote push notification service (eg. `APN`, `Firebase`)
+    public let pushProvider: PushProvider
+
+    /// Name of your push configuration in the Stream's dashboard.
+    public let providerName: String?
```
evsaev, line 16: Why do we call configuration a provider? That's a bit confusing IMO, because currently when we say `provider` we mean either APN or Firebase

bielikb: This follows the naming adopted by the backend. It's good question, I'll pipe it up to workgroup. At this moment the naming change would cause `iOS` to diverge from the rest of the SDKs.

evsaev (commented):
What if we just have another `configuration` (e.g. `alpha`) instead of 2nd target? Because that's the approach enterprise usually follows: they ship 1 app version for internal testing purposes via TestFlight and another app is a production app available in AppStore for end users

bielikb:
Usually you have separate targets per each type of app (staging, production). The separation of concerns via another app target for multi bundle app makes it easy to configure the build settings/phases the way they make sense for you. eg adding support for `Firebase` in MB target won't interfere with our main `DemoApp` target.

> What if we just have another `configuration` (e.g. `alpha`) instead of 2nd target? Because that's the approach enterprise usually follows: they ship 1 app version for internal testing purposes via TestFlight and another app is a production app available in AppStore for end users. They have single target but multiple configurations which bundle identifier depends on

evsaev (approved):
LGTM

Sources/StreamChat/Models/PushDevice.swift
ferhatelmas: provider is APN, firebase, etc. but they can be multiple whose name defined by users such as staging, prod, etc.

PR #2113, opened 2022-06-22
[CIS-1894] Pre-Warm DB objects before saving big payloads

polqf:
### 🔗 Issue Links

https://stream-io.atlassian.net/browse/CIS-1894

### 🎯 Goal

Saving big payloads with a lot of children is really expensive because we perform a lot of individual fetches per each item that we want to check whether it is already there or not. This means a lot of trips to the database.

The goal of this PR is to introduce a mechanism that will
1. Get all the ids in the payload separated by entity
2. Perform 1 fetch per entity 
3. Create a mapping between the ids on the payload and the model ids (objectID)
4. Use this with CoreData's in-memory cache to quickly get all those elements instead of consulting the database each time.

End result: we only perform one try to the database per each entity, and then we only consult the cache.

In the event that the object is not there, we just create it.

### 📝 Summary
Saving a big channel list payload went from 3.4s to 1.6s 🎉 

### 🛠 Implementation

The initial implementation was based on Reflection but ended up not being super performant.
The current approach leverages Protocols to define an API for all the Payloads/Models that are used in the process.
It is basically a depth-first search through all the elements and its children looking for the ids that are needed.

### ☑️ Contributor Checklist


Sources/StreamChat/APIClient/Endpoints/Payloads/IdentifiableModel.swift
```diff
@@ -0,0 +1,61 @@
+//
+// Copyright © 2022 Stream.io Inc. All rights reserved.
+//
+
+import CoreData
+import Foundation
+
+typealias PreWarmedCache = [String: [String: NSManagedObjectID]]
```
b-onc, line 8: This comment will apply throughout this file

Can we make these more readable by typealiasing a bunch of stuff
```swift
typealias DatabaseName = String
typealias DatabaseId = String
```
then
```
[String: Set<String>] => [DatabaseName: Set<DatabaseId>]
[String: [String: NSManagedObjectId]] => [DatabaseName: [DatabaseId: NSManagedObjectId]]
```
also
```
protocol IdentifiableModel {
     static var className: DatabaseName { get }
     static var idKeyPath: String? { get }
     static func id(for model: NSManagedObject) -> DatabaseId?
 }
```

Sources/StreamChat/APIClient/Endpoints/Payloads/IdentifiableModel.swift
```diff
@@ -0,0 +1,61 @@
...
+    static func id(for model: NSManagedObject) -> String?
+}
+
+private extension IdentifiableModel {
+    static var _className: String { String(describing: self) }
+}
+
+extension ChannelDTO: IdentifiableModel {
+    static let className: String = _className
+    static let idKeyPath: String? = "cid"
```
b-onc, line 29:
```suggestion
    static let idKeyPath: String? = #keyPath(ChannelDTO.cid)
```
applies to all others as well

Sources/StreamChat/APIClient/Endpoints/Payloads/IdentifiablePayload.swift
```diff
@@ -0,0 +1,224 @@
...
+    func recursivelyGetAllIds() -> [String: Set<String>] {
+        var cache: [String: Set<String>] = [:]
+        fillIds(cache: &cache)
+        return cache
+    }
+
+    func getPayloadToModelIdMappings(context: NSManagedObjectContext) -> PreWarmedCache {
+        let payloadIdsMappings = recursivelyGetAllIds()
+
+        var modelToIdMappings: [String: [String: NSManagedObjectID]] = [:]
```
b-onc, line 31:
```suggestion
        var preWarmedCache: [String: [String: NSManagedObjectID]] = [:]
```
I'm confused by the `modelToIdMappings`, it's more like `modelToIdToModelIdMappings` (which is just more confusing)

Sources/StreamChat/APIClient/Endpoints/Payloads/IdentifiablePayload.swift
```diff
@@ -0,0 +1,224 @@
+//
+// Copyright © 2022 Stream.io Inc. All rights reserved.
+//
+
+import CoreData
+import Foundation
+
+protocol IdentifiablePayload {
+    var databaseId: String? { get }
+    static var modelClass: (NSManagedObject & IdentifiableModel).Type? { get }
```
b-onc, line 10:
```swift
typealias IdentifiableDatabaseObject = NSManagedObject & IdentifiableModel
```
since I saw this multiple times

Sources/StreamChat/APIClient/Endpoints/Payloads/IdentifiablePayload.swift
```diff
@@ -0,0 +1,224 @@
...
+
+    func fillIds(cache: inout [String: Set<String>]) {
+        addId(cache: &cache)
+        user.fillIds(cache: &cache)
+    }
+}
+
+private extension NSManagedObject {
+    static func batchFetch(keyPath: String, equalTo values: [String], context: NSManagedObjectContext) -> [NSManagedObject] {
+        let request = NSFetchRequest<Self>(entityName: entityName)
```
b-onc, line 220: You don't need to fetch all the objects, you can set
```
request = .managedObjectIDResultType
```
and then CoreData will only return ManagedObjectIds, resulting in better performance

Sources/StreamChat/APIClient/Endpoints/Payloads/IdentifiablePayload.swift
```diff
@@ -0,0 +1,224 @@
...
+                    return nil
+                }
+            }()
+
+            guard let modelClass = modelClass, let keyPath = modelClass.idKeyPath else { continue }
+
+            let values = Array(identifiableValues)
+            var results: [NSManagedObject]?
+            context.performAndWait {
+                results = modelClass.batchFetch(keyPath: keyPath, equalTo: values, context: context)
```
b-onc, line 58: You can have a `batchFetchIds` func that'll only fetch ManagedObjectIds (_how_ is explained below), will have better performance since this part doesn't require whole objects

Sources/StreamChat/APIClient/Endpoints/Payloads/IdentifiablePayload.swift
```diff
@@ -0,0 +1,224 @@
...
+    static let modelClass: (NSManagedObject & IdentifiableModel).Type? = MemberDTO.self
+
+    func fillIds(cache: inout [String: Set<String>]) {
+        addId(cache: &cache)
+        user.fillIds(cache: &cache)
+    }
+}
+
+extension ChannelReadPayload: IdentifiablePayload {
+    var databaseId: String? { nil } // Needs a composed predicate 'channel.cid == %@ && user.id == %@'
```
b-onc, line 209: Shouldn't we add `id` to DTO then? Like `AttachmentId` or `MessageReactionId`

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
```diff
@@ -111,7 +111,11 @@ class ChannelDTO: NSManagedObject {
         return load(by: request, context: context)
     }
     
-    static func loadOrCreate(cid: ChannelId, context: NSManagedObjectContext) -> ChannelDTO {
+    static func loadOrCreate(cid: ChannelId, context: NSManagedObjectContext, cache: PreWarmedCache?) -> ChannelDTO {
+        if let cachedObject = cache?.model(for: cid.rawValue, context: context, type: ChannelDTO.self) {
```
b-onc, line 115: With a caller like this, you don't need to pass `type` param to `model(for:context:type:)` func, since T will get resolved correctly

Sources/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -647,7 +661,19 @@ extension NSManagedObjectContext: MessageDatabaseSession {
         return dto
     }
 
-    func saveMessage(payload: MessagePayload, for cid: ChannelId?, syncOwnReactions: Bool = true) throws -> MessageDTO? {
+    func saveMessages(messagesPayload: MessageListPayload, for cid: ChannelId?, syncOwnReactions: Bool = true) -> [MessageDTO] {
+        let cache = messagesPayload.getPayloadToModelIdMappings(context: self)
+        return messagesPayload.messages.compactMap {
+            try? saveMessage(payload: $0, for: cid, syncOwnReactions: syncOwnReactions, cache: cache)
```
b-onc, line 667: (Applies to multiple places) Thrown errors are just ignored now, they should at least be logged

Tests/StreamChatTests/APIClient/Endpoints/Payloads/ChannelListPayload_Tests.swift
```diff
@@ -30,22 +30,36 @@ final class ChannelListPayload_Tests: XCTestCase {
...
-                writeCompleted.fulfill()
-            })
-            wait(for: [writeCompleted], timeout: 10)
+            saveChannelListPayload(decodedPayload, database: databaseContainer)
+        }
+    }
+
+    func test_hugeChannelListQuery_save_cached() throws {
+        let decodedPayload = createHugeChannelList()
+        let databaseContainer = DatabaseContainer_Spy()
```
b-onc, line 57: `databaseContainer` creation should happen in `measure` block (as it's in the uncached test) since DB creation makes sure we're writing the payload on an empty DB

We can introduce another set of tests where DB creation happens outside `measure` block and this'll measure payload saving when DB already has the data

Tests/StreamChatTests/APIClient/Endpoints/Payloads/IdentifiableModel_Tests.swift
```diff
@@ -0,0 +1,107 @@
...
+//
+
+@testable import StreamChat
+@testable import StreamChatTestTools
+import XCTest
+
+final class IdentifiableModel_Tests: XCTestCase {
+    var database: DatabaseContainer!
+
+    private lazy var deviceDTO: DeviceDTO = {
```
b-onc, line 12: We should use the existing `TestManagedObject` as the DTO for unit tests when the test itself is irrelevant of a specific DTO

Tests/StreamChatTests/APIClient/Endpoints/Payloads/IdentifiableModel_Tests.swift
```diff
@@ -0,0 +1,107 @@
...
+        dto.id = "DeviceDTO_123"
+        return dto
+    }()
+
+    override func setUp() {
+        super.setUp()
+        database = DatabaseContainer_Spy()
+    }
+
+    func test_ChannelDTO_conformsToIdentifiableModel() {
```
b-onc, line 23: Can we generalize these tests somehow?
1. Get all Models registered in the DB scheme
2. Iterate over them and do what's done in these tests

Tests/StreamChatTests/APIClient/Endpoints/Payloads/IdentifiablePayload_Tests.swift
```diff
@@ -0,0 +1,407 @@
...
+
+        let reactionIds = try XCTUnwrap(cache[MessageReactionDTO.className])
+        let userIds = try XCTUnwrap(cache[UserDTO.className])
+
+        XCTAssertEqual(cache.keys.count, 2)
+        XCTAssertEqual(reactionIds, ["u3/m2/r1"])
+        XCTAssertEqual(userIds, ["u3"])
+    }
+
+    func test_bigChannelListPayload_recursivelyIdentifiablePayload() throws {
```
b-onc, line 218: Should we add a performance test for our `recursivelyGetAllIds` algorithm?

evsaev:
@polqf I had a suggestion but was not sure it would work so I actually tried it in a separate `feature/pre-warm-suggestion` branch and it seems to be working fine and doing the same thing in a bit another way. Would give it a quick look?

Sources/StreamChat/APIClient/Endpoints/Payloads/IdentifiablePayload.swift
polqf: We should, yeah, but outside of the scope of this PR for now

Tests/StreamChatTests/APIClient/Endpoints/Payloads/ChannelListPayload_Tests.swift
polqf: This is exactly what we don't want to do. In a pre-warm test, what you want is to have the DB filled, and measure how much it takes to save the payload when the data is already there.

Sources/StreamChat/APIClient/Endpoints/Payloads/IdentifiableModel.swift
polqf: You can do that, but then you need to get the results in the same order to be able to match id to objectID. Which doesn't seem to be the case.

Sources/StreamChat/APIClient/Endpoints/Payloads/IdentifiablePayload.swift
polqf: Let's continue the discussion in the other comment

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
polqf: Not unless you cast on the if let, which I want to avoid

Tests/StreamChatTests/APIClient/Endpoints/Payloads/IdentifiableModel_Tests.swift
polqf: It is throwing an error when using TestManagedObject because it cannot get its entity description

`An NSManagedObject of class 'TestManagedObject' must have a valid NSEntityDescription`

Sources/StreamChat/Database/DTOs/MessageDTO.swift
polqf: There is already a perf test for it. Line 47 - test_measureBigPayload_recursivelyGetAllIds

evsaev:
> @evsaev I've been checking your suggestion and seems pretty similar to my initial try. As shared in the description, Reflection was my first approach, and actually accomplished the same "result" with half of the code or less. BUT:
> 
> > The initial implementation was based on Reflection but ended up not being super performant.
> 
> Performance was degraded. 5x if I remember it correctly. So that's why it was discarded. Also, I see a lot of recursion in your approach, which might even degrade that further. On top of that, dynamic castings also add quite a toll there.

@polqf Yes, the performance degraded because for each model type we iterate all the way through the nested payload structure checking its field type. Instead of the reflection, we can take the approach with the protocol and concrete implementations for each payload type.

How about protocols with associated types?

Sources/StreamChat/APIClient/Endpoints/Payloads/IdentifiableModel.swift
b-onc: Can you clarify? Why do we need to get the results in the same order? It's just eye-sugar

Tests/StreamChatTests/APIClient/Endpoints/Payloads/ChannelListPayload_Tests.swift
b-onc: The other test is measuring how much time it takes to save the payload _when DB is empty_. To be able to compare, this test should also measure that. As I said, we can have another set of tests where we measure this on a pre-filled DB.

Pre-warm doesn't only benefit us when DB is filled, it also benefits when DB is empty since # of fetches are greatly reduced

Tests/StreamChatTests/APIClient/Endpoints/Payloads/IdentifiableModel_Tests.swift
b-onc: Alright, then we should have another DTO for this test. Basically, it shouldn't be bound to any DTO config

Tests/StreamChatTests/APIClient/Endpoints/Payloads/IdentifiableModel_Tests.swift
b-onc: I guess the answer to this was:
> I am in favor of making it a bit more readable than parametrizing tests. In the end, more human errors can happen if we do what you suggest.

Fair enough, I wanted to handle the case where we add a new DTO and forget to add a test for it in this file

Sources/StreamChat/APIClient/Endpoints/Payloads/IdentifiableModel.swift
polqf: I swear I put this comment in another thread..

Sources/StreamChat/APIClient/Endpoints/Payloads/IdentifiablePayload.swift
polqf: You can do that, but then you need to get the results in the same order to be able to match id to objectID. Which doesn't seem to be the case.

b-onc: I now realized the snippet i've pasted was missing some important characters :D 

```
request.resultType = .managedObjectIDResultType
```

this won't change the ordering or anything, it'll just make sure only objectIDs are returned

polqf:
> @polqf Yes, the performance degraded because for each model type we iterate all the way through the nested payload structure checking each field. Instead of the reflection, we can take the approach with the protocol and concrete implementations for each payload type you're using atm.
> 
> How about protocols with associated types?

I basically tested 4 different approaches. Those include Reflection, Protocol Conformances, Associated Types, and this last one which was more "manual".

Anything that includes reflection or type check-ups is going to be far slower

b-onc:
@polqf @evsaev My 2cents is, this part of the code should be focused on efficiency and not readability. Let's write it in inline assembly if it means we'll have much better performance :D

Tests/StreamChatTests/APIClient/Endpoints/Payloads/ChannelListPayload_Tests.swift
polqf: I guess you are misled by the test name. What I mean by cached is with the DB filled.

I both cases Pre-warming is enabled.

polqf:
> @polqf @evsaev My 2cents is, this part of the code should be focused on efficiency and not readability. Let's write it in inline assembly if it means we'll have much better performance :D

Agree with the above. And that's why I sacrificed scalability and "magic" for the sake of a much better performance

Sources/StreamChat/APIClient/Endpoints/Payloads/IdentifiablePayload.swift
polqf: Yeah, but that is not enough. We need to create a mapping between id and objectID. If we only get the objectID, we don't know to which id it belongs

b-onc (approved):

PR #2115, opened 2022-06-23
Add UserPayload caching during decoding

b-onc:
### 🎯 Goal

With #2084 we improved JSON decoding performance, but we can still do better

### 📝 Summary

JSON decoding bottleneck appears to be the repeated User payloads. Each UserPayload has 3 Dates to be decoded, which is the most expensive primitive to decode.

### 🛠 Implementation

We can cache Dates to improve the performance by a lot, but caching UserPayloads improves the performance further and it's more logical, since UserPayload is repeated multiple times, and Dates are repeated because of this.

UserPayload is safe to decode with any JSONDecoder since we use `userInfo` to check if we have caching enabled on that specific decoder.

### 🎨 Showcase

| Before | User caching | User + Date caching (#2128)
| ------ | ----- | ----- |
|  970ms   |  514ms  | 185ms

### 🧪 Manual Testing Notes

You can use the disabled `ChannelListPayload_Tests.test_decode_bigChannelListPayload` test to measure.

### ☑️ Contributor Checklist


Sources/StreamChat/Utils/Codable+Extensions.swift
```diff
@@ -4,12 +4,22 @@
 
 import Foundation
 
+class StreamJSONDecoder: JSONDecoder {
```
bielikb, line 7:
```suggestion
final class StreamJSONDecoder: JSONDecoder {
```

Sources/StreamChat/APIClient/Endpoints/Payloads/UserPayloads.swift
```diff
@@ -73,29 +75,50 @@ class UserPayload: Decodable {
...
+            } catch {
+                log.error(
+                    "Failed to decode extra data for User with id: <\(userId)>, using default value instead. "
+                        + "Error: \(error)"
+                )
+                extraData = [:]
+            }
+            
+            if isCachingEnabled {
+                UserPayload.userDecodingCache[userId] = self
```
bielikb, line 120: It looks like we don't invalidate the cache.

Could we invalidate the cache eg when `user disconnects` or `ChatClient` is initialized?

Sources/StreamChat/Utils/Codable+Extensions.swift
```diff
@@ -4,12 +4,26 @@
 
 import Foundation
 
+final class StreamJSONDecoder: JSONDecoder {
+    override func decode<T>(_ type: T.Type, from data: Data) throws -> T where T: Decodable {
```
tbarbugli, line 8: why did you have this here and not in the response decoder?

Sources/StreamChat/APIClient/Endpoints/Payloads/UserPayloads.swift
```diff
@@ -29,6 +29,8 @@ enum UserPayloadsCodingKeys: String, CodingKey, CaseIterable {
 
 /// An object describing the incoming user JSON payload.
 class UserPayload: Decodable {
+    static var userDecodingCache = [String: UserPayload]()
```
tbarbugli, line 32: this is not thread-safe

Sources/StreamChat/Utils/Codable+Extensions.swift
b-onc: Can you clarify?

Sources/StreamChat/APIClient/Endpoints/Payloads/UserPayloads.swift
b-onc: Alright, I've switched to NSCache which is thread-safe:
> You can add, remove, and query items in the cache from different threads without having to lock the cache yourself.

Sources/StreamChat/Utils/Codable+Extensions.swift
polqf, line 8: What about having another name for it? And then you have it in a fileprivate extension in the UserPayload?

polqf: I don't see a need for a subclass here 🤔

Sources/StreamChat/Utils/Codable+Extensions.swift
```diff
@@ -37,6 +51,9 @@ extension JSONDecoder {
             throw DecodingError.dataCorruptedError(in: container, debugDescription: "Invalid date: \(dateString)")
         }
         
+        // We enable the UserPayload caching by setting this key
+        decoder.userInfo[userPayloadCachingFlagKey] = true
```
polqf, line 55: Why is there this flag in `userInfo` while it is always true and integrators cannot change it? I would skip it and treat it as always on

polqf: Or is there any reason why we want to enable/disable it?

Sources/StreamChat/APIClient/Endpoints/Payloads/UserPayloads.swift
```diff
@@ -29,6 +29,8 @@ enum UserPayloadsCodingKeys: String, CodingKey, CaseIterable {
 
 /// An object describing the incoming user JSON payload.
 class UserPayload: Decodable {
+    static var userDecodingCache = NSCache<NSString, UserPayload>()
```
polqf, line 32: with the other suggestion, make it fileprivate

Sources/StreamChat/Utils/Codable+Extensions.swift
b-onc: I don't follow? The aim is to make sure cache is reset before any decoding session, since cache is only valid for a single decoding session

Sources/StreamChat/Utils/Codable+Extensions.swift
b-onc: We want the `UserPayload` to be decodable with any JSONDecoder, not just with ours. Without the flag/subclass, cache will always fill, in turn we'll get invalid cache across decodings. Without this flag, decoding `UserPayload` with other JSONDecoder's would fill the cache, and would result in incorrect decodings

Sources/StreamChat/Utils/Codable+Extensions.swift
polqf: If you have a fileprivate extension (in UserPayload), you avoid overriding this method, and only add this logic there.
That avoids having a subclass, too.

Sources/StreamChat/Utils/Codable+Extensions.swift
polqf: UserPayload is decodable with any JSONDecoder. And with the current logic, if the value is not in the cache, it will decode it from the json. I don't see a problem here. Am I missing anything?

Sources/StreamChat/Utils/Codable+Extensions.swift
b-onc: fileprivate extension to which class exactly? And where would we call this new func? Where would we reset the cache?

Sources/StreamChat/Utils/Codable+Extensions.swift
b-onc: Without the flag, using any other JSONDecoder:
1. You decode a ChannelListPayload. All UserPayloads are cached.
2. You decode another ChannelListPayload, with the same users, but they're updated. All UserPayloads are received from the cache (because their id is not changed), but they're outdated

Tests/StreamChatTests/Utils/JSONDecoder_Tests.swift
```diff
@@ -118,13 +106,49 @@ final class JSONDecoderTests: XCTestCase {
             second: 13
         )
     }
-}
+    
+    func test_userPayloadCaching_isThreadSafe() {
+        for offset in 0..<100 {
```
bielikb, line 111: This is nice touch!

bielikb (approved):
This solution looks good also from a reliability perspective. +1 from me.

Sources/StreamChat/Utils/Codable+Extensions.swift
polqf: Forget about it, I was misled by the same signature in JSONDecoder and its container.

Worth adding a comment there on the overriden `decode` function to call out it is only affecting the top level payload/data, as the children of the payload call `decode` on the container, and not on the decoder.

nuno-vieira:
@b-onc How is the cache invalidated? What happens if the user was updated? I'm a bit concerned this will add some data inconsistencies issues.

Sources/StreamChat/Utils/Codable+Extensions.swift
nuno-vieira: This cache invalidation doesn't seem right 🤔 Does this mean that for a session, if any user updates their data, you won't get the data correctly? 

Example:
1. You fetch the message list with user avatars
2. User updates the avatar, this updates local DB correclty
3. you fetch again the message list, since the user is in the cache, it overrides the updated user
4. You get the data incorrectly

It seems to me that instead of removing all cache here, can we remove the user payload from the cache whenever we have a user updated event?

nuno-vieira (dismissed):
I'm requesting changes since I think that the cache invalidation logic is wrong, but maybe I'm missing something 🤔

Sources/StreamChat/Utils/Codable+Extensions.swift
b-onc: @nuno-vieira Cache is invalidated before and after any decoding session, so it's not possible for it to become invalid. I've added thread safety checks for this too, to make sure cache doesn't leak inter-session. You can QA this branch to see that the scenario you're describing doesn't happen.

PR #2119, opened 2022-06-24
Add missing channel / member list filters

b-onc:
### 🎯 Goal

Add missing Channel List and Member List filter keys so they can be used in a type-safe manner (and autocompletion rocks)

### 📝 Summary

We were missing a couple of Channel List filter keys:
https://getstream.io/chat/docs/ios-swift/query_channels/?language=javascript#channel-queryable-built-in-fields
and Member list filter keys:
https://getstream.io/chat/docs/ios-swift/query_members/?language=javascript#member-queryable-built-in-fields

Those are added to the already-existing filter infra, and comments are updated to reflect which operators are possible (this info already exists in our docs)

### 🛠 Implementation

Hypothetically, we can implement `var supportedOperators: [FilterOperator]` to `FilterKey` so we can know exactly which filter is usable with which operator. This is a bit more complicated, specifically because `FilterKey` is a struct and not an enum, so I'm leaving that out for now.

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/Query/ChannelListQuery.swift
+++ b/Sources/StreamChat/Query/ChannelListQuery.swift
@@ -47,10 +47,24 @@ extension FilterKey where Scope: AnyChannelListFilterScope {
     static var members: FilterKey<Scope, UserId> { "members" }
 }
 
-/// Non extra-data-specific filer keys for channel list.
+/// Filter values to be used with `.invite` FilterKey.
+public enum InviteFilterValue: String, FilterValue {
+    case pending
+    case accepted
+    case rejected
+}
+
+/// Filter keys for channel list.
 public extension FilterKey where Scope: AnyChannelListFilterScope {
     /// A filter key for matching the `cid` value.
+    /// Supported operators: `in`, `equal`
     static var cid: FilterKey<Scope, ChannelId> { "cid" }
+    
+    /// A filter key for matching the `id` value.
+    /// Supported operators: `in`, `equal`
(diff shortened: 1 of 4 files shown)


bielikb (approved):

Sources/StreamChat/Query/ChannelListQuery.swift
```diff
@@ -47,46 +47,91 @@ extension FilterKey where Scope: AnyChannelListFilterScope {
     static var members: FilterKey<Scope, UserId> { "members" }
 }
 
-/// Non extra-data-specific filer keys for channel list.
+/// Filter values to be used with `.invite` FilterKey.
+public enum InviteFilterValue: String, FilterValue {
+    case pending
+    case accepted
+    case rejected
+}
```
nuno-vieira, line 55: Any reason we don't use a struct here? Are we sure this enum won't be changed? Would it be useful for customers to extend this?

b-onc: From backend documentation:
>invite: type string, must be one of these values: (pending, accepted, rejected)

SSaleemSSI:
Hi - How can i use this filter if i want to filter channels in which i haven't accepted the request yet and is pending.
.equal(.invite, to: true)
this lines gives error 
Member 'invite' in 'FilterKey<ChannelListFilterScope, Bool>' produces result of type 'FilterKey<ChannelListFilterScope, InviteFilterValue>', but context expects 'FilterKey<ChannelListFilterScope, Bool>'

PR #2126, opened 2022-06-28
[CIS-1407] Improve JSON decoding robustness

b-onc:
### 🔗 Issue Links

CIS-1407
https://getstream.zendesk.com/agent/tickets/24700

### 🎯 Goal

* JSON decoding should be much more robust
* JSON decoding errors must be more understandable

### 📝 Summary

This PR makes it so a single deformed entity in JSON arrays won't disable decoding for the whole entity itself, but will print a fat, very verbose error instead.

### 🛠 Implementation

We already have `decodeArrayIgnoringFailures` which decodes the array elements where decodable and prints the error where they are deformed. This was used for Event decoding. We can safely use this for any array decoding.

The error function didn't print understandable information. The errors look like:
> 2022-06-27 11:41:05.837 [ERROR] [com.apple.NSURLSession-delegate] [RequestDecoder.swift:79] [decodeRequestResponse(data:response:error:)] > valueNotFound(Swift.KeyedDecodingContainer<StreamChat.UserPayloadsCodingKeys>, Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "channels", intValue: nil), _JSONKey(stringValue: "Index 1", intValue: 1), CodingKeys(stringValue: "messages", intValue: nil), _JSONKey(stringValue: "Index 22", intValue: 22), MessagePayloadsCodingKeys(stringValue: "own_reactions", intValue: nil), _JSONKey(stringValue: "Index 0", intValue: 0), CodingKeys(stringValue: "user", intValue: nil)], debugDescription: "Cannot get keyed decoding container -- found null value instead.", underlyingError: nil))

With the PR, errors look like:
> 2022-06-27 11:41:05.837 [ERROR] [com.apple.NSURLSession-delegate] [RequestDecoder.swift:79] [decodeRequestResponse(data:response:error:)] > Failed to decode MessageReactionPayload in array: {
  "score" : 1,
  "message_id" : "8268DBDE-6B10-4326-8F1E-2564CA3F20B9",
  "created_at" : "2022-05-15T19:31:25.96832Z",
  "user_id" : "8bee5134-5d06-43a2-b2f7-76e4f12aeb8b",
  "updated_at" : "2022-05-15T19:31:25.96832Z",
  "type" : "like",
  "enforce_unique" : false,
  "user" : null
}, error: valueNotFound for value KeyedDecodingContainer<UserPayloadsCodingKeys>, path: <channels[8].messages[15].own_reactions[0].user>, debugDescription: Cannot get keyed decoding container -- found null value instead.

The error has all the information - the failed payload, legible coding path, and the description - and it's much more legible.

More importantly, 1 single deformed object won't cause whole channel list to go blank.

### 🧪 Manual Testing Notes

You can use any channel list payload, just deform an object and run this test with the payload:
```swift
    func test_decode_deformedChannelListPayload() {
        let url = XCTestCase.mockData(fromJSONFile: "<#Deformed Channel List Payload#>")

        do {
                _ = try JSONDecoder.default.decode(ChannelListPayload.self, from: url)
            } catch {
                XCTFail("Failed to parse JSON: \(error)")
            }
    }
```
This will print the `log.error` to the console.

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/APIClient/Endpoints/Payloads/ChannelListPayload.swift
+++ b/Sources/StreamChat/APIClient/Endpoints/Payloads/ChannelListPayload.swift
@@ -58,10 +58,10 @@ extension ChannelPayload: Decodable {
             channel: try container.decode(ChannelDetailPayload.self, forKey: .channel),
             watcherCount: try container.decodeIfPresent(Int.self, forKey: .watcherCount),
             watchers: try container.decodeIfPresent([UserPayload].self, forKey: .watchers),
-            members: try container.decode([MemberPayload].self, forKey: .members),
+            members: try container.decodeArrayIgnoringFailures([MemberPayload].self, forKey: .members),
             membership: try container.decodeIfPresent(MemberPayload.self, forKey: .membership),
-            messages: try container.decode([MessagePayload].self, forKey: .messages),
-            pinnedMessages: try container.decode([MessagePayload].self, forKey: .pinnedMessages),
+            messages: try container.decodeArrayIgnoringFailures([MessagePayload].self, forKey: .messages),
+            pinnedMessages: try container.decodeArrayIgnoringFailures([MessagePayload].self, forKey: .pinnedMessages),
             channelReads: try container.decodeIfPresent([ChannelReadPayload].self, forKey: .channelReads) ?? [],
             isHidden: try container.decodeIfPresent(Bool.self, forKey: .hidden)
         )
--- a/Sources/StreamChat/APIClient/Endpoints/Payloads/MessagePayloads.swift
+++ b/Sources/StreamChat/APIClient/Endpoints/Payloads/MessagePayloads.swift
@@ -110,12 +110,12 @@ class MessagePayload: Decodable {
         parentId = try container.decodeIfPresent(String.self, forKey: .parentId)
         showReplyInChannel = try container.decodeIfPresent(Bool.self, forKey: .showReplyInChannel) ?? false
         quotedMessage = try container.decodeIfPresent(MessagePayload.self, forKey: .quotedMessage)
-        mentionedUsers = try container.decode([UserPayload].self, forKey: .mentionedUsers)
+        mentionedUsers = try container.decodeArrayIgnoringFailures([UserPayload].self, forKey: .mentionedUsers)
         // backend returns `thread_participants` only if message is a thread, we are fine with to have it on all messages
         threadParticipants = try container.decodeIfPresent([UserPayload].self, forKey: .threadParticipants) ?? []
         replyCount = try container.decode(Int.self, forKey: .replyCount)
-        latestReactions = try container.decode([MessageReactionPayload].self, forKey: .latestReactions)
-        ownReactions = try container.decode([MessageReactionPayload].self, forKey: .ownReactions)
+        latestReactions = try container.decodeArrayIgnoringFailures([MessageReactionPayload].self, forKey: .latestReactions)
+        ownReactions = try container.decodeArrayIgnoringFailures([MessageReactionPayload].self, forKey: .ownReactions)
 
         reactionScores = try container
             .decodeIfPresent([String: Int].self, forKey: .reactionScores)?
--- a/Sources/StreamChat/Utils/KeyedDecodingContainer+Array.swift
+++ b/Sources/StreamChat/Utils/KeyedDecodingContainer+Array.swift
@@ -6,6 +6,7 @@ import Foundation
 
 private struct ElementWrapper<T: Decodable>: Decodable {
     let value: T?
+    var rawJSON: [String: RawJSON]?
     var error: Error?
     init(from decoder: Decoder) throws {
         do {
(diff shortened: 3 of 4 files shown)


bielikb (approved):
Excellent!

take my money

nuno-vieira (approved):
LGTM! ✅

Sources/StreamChat/APIClient/Endpoints/Payloads/ChannelListPayload.swift
```diff
@@ -58,10 +58,10 @@ extension ChannelPayload: Decodable {
             channel: try container.decode(ChannelDetailPayload.self, forKey: .channel),
             watcherCount: try container.decodeIfPresent(Int.self, forKey: .watcherCount),
             watchers: try container.decodeIfPresent([UserPayload].self, forKey: .watchers),
```
polqf, line 60: We should introduce a `decodeArrayIfPresentIgnoringFailures`
`decodeIfPresent` is not enough because it guards us against it being nil, but whenever it is not nil, it fails the same way as a regular decode.

Sources/StreamChat/APIClient/Endpoints/Payloads/ChannelListPayload.swift
```diff
@@ -58,10 +58,10 @@ extension ChannelPayload: Decodable {
...
-            members: try container.decode([MemberPayload].self, forKey: .members),
+            members: try container.decodeArrayIgnoringFailures([MemberPayload].self, forKey: .members),
             membership: try container.decodeIfPresent(MemberPayload.self, forKey: .membership),
-            messages: try container.decode([MessagePayload].self, forKey: .messages),
-            pinnedMessages: try container.decode([MessagePayload].self, forKey: .pinnedMessages),
+            messages: try container.decodeArrayIgnoringFailures([MessagePayload].self, forKey: .messages),
+            pinnedMessages: try container.decodeArrayIgnoringFailures([MessagePayload].self, forKey: .pinnedMessages),
             channelReads: try container.decodeIfPresent([ChannelReadPayload].self, forKey: .channelReads) ?? [],
```
polqf, line 65: Same

Sources/StreamChat/APIClient/Endpoints/Payloads/MessagePayloads.swift
```diff
@@ -110,12 +110,12 @@ class MessagePayload: Decodable {
         parentId = try container.decodeIfPresent(String.self, forKey: .parentId)
         showReplyInChannel = try container.decodeIfPresent(Bool.self, forKey: .showReplyInChannel) ?? false
         quotedMessage = try container.decodeIfPresent(MessagePayload.self, forKey: .quotedMessage)
-        mentionedUsers = try container.decode([UserPayload].self, forKey: .mentionedUsers)
+        mentionedUsers = try container.decodeArrayIgnoringFailures([UserPayload].self, forKey: .mentionedUsers)
         // backend returns `thread_participants` only if message is a thread, we are fine with to have it on all messages
         threadParticipants = try container.decodeIfPresent([UserPayload].self, forKey: .threadParticipants) ?? []
```
polqf, line 115: Same

Sources/StreamChat/Utils/KeyedDecodingContainer+Array.swift
```diff
@@ -6,25 +6,83 @@ import Foundation
...
+    var rawJSON: [String: RawJSON]?
     var error: Error?
     init(from decoder: Decoder) throws {
         do {
             value = try T(from: decoder)
         } catch {
             value = nil
             self.error = error
+            // We decode the payload as RawJSON to be able to display the payload in case of error
+            rawJSON = try? [String: RawJSON](from: decoder)
```
polqf, line 18: Can we measure if this adds a lof of overhead?

Sources/StreamChat/APIClient/Endpoints/Payloads/ChannelListPayload.swift
polqf: Let's add tests to cover for this before introducing it. Make it fail, and then introduce `decodeArrayIfPresentIgnoringFailures`

Sources/StreamChat/Utils/KeyedDecodingContainer+Array.swift
```diff
@@ -6,25 +6,83 @@ import Foundation
...
 extension KeyedDecodingContainer {
     func decodeArrayIgnoringFailures<T: Decodable>(_ type: [T].Type, forKey key: KeyedDecodingContainer<K>.Key) throws -> [T] {
-        let wrapper = try decode([ElementWrapper<T>].self, forKey: key)
-        let errors = wrapper.compactMap(\.error)
-        if !errors.isEmpty {
-            let errorsDescription = errors.map(String.init(describing:))
-            log.error("Failed decoding elements from array: \(errorsDescription)", subsystems: .httpRequests)
+        let wrappers = try decode([ElementWrapper<T>].self, forKey: key)
+        for wrapper in wrappers where wrapper.error != nil {
+            let rawJSONPrettyPrinted = (try? JSONEncoder.default.encode(wrapper.rawJSON))?.debugPrettyPrintedJSON
```
polqf, line 27: I am worried about the double decode/encode here. Why are you decoding `rawJSON` in the first place if you are going to encode it here?

polqf (commented):
I am a bit worried about the overhead can have by all detailed logging you want to add

Sources/StreamChat/APIClient/Endpoints/Payloads/ChannelListPayload.swift
b-onc: I'll open a new PR for this, while approved let's keep the scope small & get it merged

Sources/StreamChat/Utils/KeyedDecodingContainer+Array.swift
b-onc: RawJSON conversion is only done when a decoding fails, so no overhead for the good path. Worst case, every element in an array where we use `decodeArrayIgnoringFailures` will fail, so we'll decode them into RawJSON, which will double the decoding time for the said array.

There's no way (that I could find) to get the Raw JSON data from decoder, that's why I opted for the RawJSON decoding path, to be able to print the erroring payload

Sources/StreamChat/Utils/KeyedDecodingContainer+Array.swift
b-onc: Here's a test I've written to measure the impact of this PR, I'll put the test at the end of this comment so the measurements stay on top.

I've used a message array (since MessagePayload is quite big - impact should be big too), with 1000 messages.

(1) Using `decode` | (2) Using `decodeArrayIgnoringFailures` | (3) Using `decodeArrayIgnoringFailures` when `updated_at` is deleted from payload
---- | ----- | -----
460ms | 461ms | 533ms

Case (3) is the extreme case where every object fails to decode (hence they're decoded as RawJSON, encoded to String and printed to console) and the array is quite big, we never have this big arrays. Performance scales linearly so for an array of 100 MessagePayloads I've observed 46ms for (1). I can say that performance impact is around 10%, which is good enough for this kind of verbosity for our errors.

The test code:
```swift
func test_decode_withErroringPayloads() throws {
    struct MessageArrayContainer: Decodable {
        enum CodingKeys: String, CodingKey {
            case messages
        }
        
        let messages: [MessagePayload]
        
        init(from decoder: Decoder) throws {
            
            messages = try container.decodeArrayIgnoringFailures([MessagePayload].self, forKey: .messages)
        }
    }
    
    let messagePayload = """
        {
        "id" : "broken-waterfall-5-7aede36b-b89f-4f45-baff-c40c7c1875d9",
        "reaction_counts" : {
        
        },
        "silent" : false,
        "created_at" : "2020-06-09T08:10:40.800912Z",
        "reaction_scores" : {
        
        },
        "type" : "regular",
        "latest_reactions" : [
        
        ],
        "text" : "sadfadf",
        "attachments" : [
        
        ],
        "own_reactions" : [
        
        ],
        "updated_at" : "2020-06-09T08:10:40.800912Z",
        "reply_count" : 0,
        "user" : {
        "id" : "broken-waterfall-5",
        "banned" : false,
        "unread_channels" : 0,
        "extraData" : {
        "name" : "Tester"
        },
        "totalUnreadCount" : 0,
        "last_active" : "2020-06-10T13:24:00.501797Z",
        "created_at" : "2019-12-12T15:33:46.488935Z",
        "unreadChannels" : 0,
        "unread_count" : 0,
        "image" : "https://getstream.io/random_svg/?id=broken-waterfall-5&amp;name=Broken+waterfall",
        "updated_at" : "2020-06-10T14:11:29.946106Z",
        "role" : "user",
        "total_unread_count" : 0,
        "online" : true,
        "name" : "broken-waterfall-5"
        },
        "mentioned_users" : [
        
        ],
        "html" : "<p>sadfadf</p>",
        "pinned": false,
        "pinned_at": null,
        "pin_expires": null,
        "pinned_by": null
        }
        """
    let messageArrayPayload = "{ \"messages\": [\(Array(repeating: messagePayload, count: 1000).joined(separator: ",\n"))]}"
    let messageArrayPayloadData = messageArrayPayload.data(using: .utf8)!
    
    measure {
        do {
            _ = try JSONDecoder.default.decode(MessageArrayContainer.self, from: messageArrayPayloadData)
        } catch {
            XCTFail("Failed to parse JSON: \(error)")
        }
    }
}
```

polqf: Good to go then 🚀

polqf (approved):

PR #2143, opened 2022-07-05
Lowercase UUID for new sent messages

Aurazion:
### 🎯 Goal

Hi,
Currently, when a new message is created in the database (Core Data), a new UUID is generated with `uuidString` property before the message is sent to Stream backend. Before the new iOS SDK was released, no UUID was generated locally, and the server returned a lowercase UUID. Android Stream SDK also have a lowercased UUID.
We noticed that your API is case sensitive, which is causing us some issues when we want to make messages requests.

### 📝 Summary

- [x] Lowercase UUID in MessageDTO for new messages.

### 🛠 Implementation

Added `lowercased()` next to UUID generation of a new message in `MessageDTO.swift` to be consistent with other Stream environments.

### ☑️ Contributor Checklist


Sources/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -455,7 +455,7 @@ extension NSManagedObjectContext: MessageDatabaseSession {
             throw ClientError.ChannelDoesNotExist(cid: cid)
         }
 
-        let message = MessageDTO.loadOrCreate(id: .newUniqueId, context: self, cache: nil)
+        let message = MessageDTO.loadOrCreate(id: .newUniqueId.lowercased(), context: self, cache: nil)
```
nuno-vieira, line 458: Hi @Aurazion ! Thanks for the PR! Can you actually put the `lowercased()` inside the `newUniqueId`? So that it is consistent across all our UUID's generations? Thanks!

Also, can you specify which issues are you having?

Aurazion: Hi @nuno-vieira ! I applied your changes thanks 👍

We having issues when we tried to request some messages on your API (case sensitive visibly) because we store all UUID in lowercase in our backend for later use. And we saw that every other platforms generate it by default in lowercase. So when we try from our backend to query some messages sent from iOS, your API is unable to fetch the message because of the case sensitivity.

nuno-vieira (approved):
LGTM! ✅

PR #2156, opened 2022-07-07
[CIS-2026] Avoid triggering CoreData updates in willSave()

nuno-vieira:
### 🔗 Issue Links
CIS-2026

### 🎯 Goal
Fix CoreData crash when triggering updates on `willSave()`

### 📝 Summary
- Removes triggers which are not needed anymore
- Only perform `willSave()` if the message is not deleted
- Try to reduce the number of times the triggers are executed to not a recursive dirtying crash

For more information, read these resources:
- https://developer.apple.com/documentation/coredata/nsmanagedobject/1506209-willsave#:~:text=If%20you%20want,SaveNotification).
- https://cocoacasts.com/three-features-of-nsmanagedobject-you-may-not-know-about#:~:text=On%20Save,have%20side%20effects

### ☑️ Contributor Checklist


Sources/StreamChat/Database/DTOs/ChannelReadDTO.swift
```diff
@@ -15,16 +15,6 @@ class ChannelReadDTO: NSManagedObject {
     @NSManaged var channel: ChannelDTO
     @NSManaged var user: UserDTO
     
-    override func willSave() {
```
martinmitrevski, line 18: Does this mean that read updates won't trigger channel changes anymore?

Sources/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -68,14 +68,18 @@ class MessageDTO: NSManagedObject {
...
         super.willSave()
+
+        guard !isDeleted else {
+            return
+        }
         
         // Manually mark the channel as dirty to trigger the entity update and give the UI a chance
         // to reload the channel cell to reflect the updated preview.
-        if let channel = previewOfChannel, !channel.hasChanges, isUpdated {
+        if let channel = previewOfChannel, !channel.hasChanges, !channel.isDeleted, channel.lastMessageAt == createdAt {
```
martinmitrevski, line 78: Can you please elaborate on the last check `channel.lastMessageAt == createdAt`?
 
Updates should be triggered when there are changes to other messages too (e.g. reactions), but I might be lacking some context here.

martinmitrevski (commented):
Was there a way to reproduce this on our side?

Sources/StreamChat/Database/DTOs/MessageDTO.swift
nuno-vieira: This part is to change the preview message of the channel (In the channel list), so the preview message that we will be analysing is always the last message of the channel. To limit the number of times this part of the codebase is executed, we make sure that the channel is not deleted and that is in fact the last message.

martinmitrevski: got it, thanks for the explanation 👍

martinmitrevski (approved):
LGTM ✅

PR #2163, opened 2022-07-12
Sync active channels when no channel list

martinmitrevski:
### 🔗 Issue Links

https://github.com/GetStream/stream-chat-swiftui/issues/126

### 🎯 Goal

Fix a bug when syncing active channels w/o channel list.

### 📝 Summary

When there's a channel screen shown directly (w/o channel list), the sync endpoint was not called, which lead to losing messages sent in the meantime.

### 🛠 Implementation

If there are no local channel ids to sync, we take the ones from the active channel ids.

### 🧪 Manual Testing Notes

You need an app which directly shows channel. Put the app into background, send messages from another device, come back and the messages are loaded now (previously they weren't).

### ☑️ Contributor Checklist


polqf (commented):
Looks good! Just missing some tests here

nuno-vieira (dismissed):
LGTM! ✅

Tests/StreamChatTests/Repositories/SyncOperations_Tests.swift
```diff
@@ -138,8 +138,8 @@ final class SyncOperations_Tests: XCTestCase {
 
         operation.startAndWaitForCompletion()
 
-        XCTAssertEqual(context.watchedAndSynchedChannelIds.count, 0)
-        XCTAssertNotCall("recoverWatchedChannel(completion:)", on: controller)
+        XCTAssertEqual(context.watchedAndSynchedChannelIds.count, 1)
+        XCTAssertCall("recoverWatchedChannel(completion:)", on: controller)
```
nuno-vieira, line 142: Why are we reverting the test here? Although the test passes now, it is not what it is supposed to be testing. Right? @polqf

Sources/StreamChat/Repositories/SyncOperations.swift
```diff
@@ -72,12 +72,6 @@ final class WatchChannelOperation: AsyncOperation {
                 return
             }
 
-            // Reset only if it needs recovery
-            if let cid = controller.cid, context.synchedChannelIds.contains(cid) {
-                done(.continue)
-                return
-            }
```
nuno-vieira, line 79: Are we sure about this?

Tests/StreamChatTests/Repositories/SyncOperations_Tests.swift
polqf: This test should be split in 2.
- One that checks that given a channelId that was synched (present in `context.synchedChannelIds `), it calls `recoverWatchedChannel`, but the end result of `context.watchedAndSynchedChannelIds` is 1
- One that checks that given  a channelId that was NOT synched (NOT present in `context.synchedChannelIds `), it calls `recoverWatchedChannel`, and the result of `context.watchedAndSynchedChannelIds` is 1 and `context.synchedChannelIds` keeps being 0

Sources/StreamChat/Repositories/SyncOperations.swift
polqf: @nuno-vieira this is a fix that's meant to be safe given the error we are fixing, but maybe not the "most performant" as it can lead to unnecessary calls. (1 more call at most)

There is a task in JIRA to review this in depth and prevent that unnecessary call if possible

Tests/StreamChatTests/Repositories/SyncOperations_Tests.swift
martinmitrevski: @polqf, the current test was already doing the first check. I've added another one that tests the second point.

Tests/StreamChatTests/Repositories/SyncOperations_Tests.swift
```diff
@@ -141,6 +141,21 @@ final class SyncOperations_Tests: XCTestCase {
         XCTAssertEqual(context.watchedAndSynchedChannelIds.count, 1)
         XCTAssertCall("recoverWatchedChannel(completion:)", on: controller)
     }
+    
+    func test_WatchChannelOperation_availableOnRemote_notSynched() {
+        let context = SyncContext(lastSyncAt: .init())
+        let controller = ChatChannelController_Spy(client: client)
+        controller.state = .remoteDataFetched
+        context.localChannelIds.insert(controller.cid!, at: 0)
```
polqf, line 149: Why is this needed here?

polqf (commented):
Just one comment. Rest looks good

Tests/StreamChatTests/Repositories/SyncOperations_Tests.swift
martinmitrevski: it's not needed, fixed

polqf (approved):

PR #2165, opened 2022-07-12
Revert willSave() removal changes

nuno-vieira:
### 🔗 Issue Links
None

### 🎯 Goal

Reverted some of the changes introduced here: https://github.com/GetStream/stream-chat-swift/pull/2156 to reduce the risk. Since I observed some data update problems after these changes.


martinmitrevski (approved):
LGTM ✅

PR #2200, opened 2022-07-27
[CIS-2087] Improve consistency when retrieving Message after Push Notification

polqf:
### 🔗 Issue Links

https://stream-io.atlassian.net/browse/CIS-2087

### 🎯 Goal

It's been a while that we randomly get reports of inconsistencies when retrieving a message from a Push Notification

### 📝 Summary

ChatRemoteNotificationHandler was using MessageController to retrieve the message that was coming from a Push Notification. This was both unreliable and overcomplex. This PR move the responsibility of doing so to the MessageRepository, to ensure we don't rely on the DB to return it.

### 🛠 Implementation

MessageRepository now has a new method, getMessage, that returns a ChatMessage object. ChatRemoteNotificationHandler directly requests the message to MessageRepository, reducing inconsistencies

### ☑️ Contributor Checklist


martinmitrevski (approved):
nice improvements @polqf 👏 Two things:
- any specific case to test or just receiving pushes?
- we should update the changelog

polqf:
@martinmitrevski this change only affects Push Notifications. it is still to be reviewed if it completely fixes the fact that channel is sometimes not fully updated when opening a push, but this will be tackled in another PR

PR #2204, opened 2022-08-02
[CIS-2048] Make sure ChannelDTO is still valid when accessing Lazy blocks

polqf:
### 🔗 Issue Links

Related to https://stream-io.atlassian.net/browse/CIS-2048

### 🎯 Goal

Mitigate crash when accessing unread count for a channel

### 📝 Summary

The crash suggests that there are values that are null when trying to evaluate the predicate. This smells like a predicate being evaluated on a deleted object. Because the unread count is accessed on demand, and not when creating ChatChannel, it can be that the DTO is deleted whenever the values is accessed.

Whenever a DTO is deleted, its values are reset to defaults. BUT, CoreData is not able to represent a `Date?` whenever this happens, leading to a crash.

All the above is just a theory, and there was no way to consistently reproduce the scenario where it crashed. In any case, it adds a safety net.

### 🛠 Implementation

We now check if the ChannelDTO is valid before performing any lazy fetch request using it.

### 🧪 Manual Testing Notes

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/Database/DTOs/ChannelDTO.swift
+++ b/Sources/StreamChat/Database/DTOs/ChannelDTO.swift
@@ -381,7 +381,7 @@ extension ChatChannel {
         let reads: [ChatChannelRead] = try dto.reads.map { try $0.asModel() }
         
         let unreadCount: () -> ChannelUnreadCount = {
-            guard let currentUser = context.currentUser else { return .noUnread }
+            guard dto.isValid, let currentUser = context.currentUser else { return .noUnread }
             
             let currentUserRead = reads.first(where: { $0.user.id == currentUser.user.id })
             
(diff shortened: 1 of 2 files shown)


martinmitrevski (approved):
LGTM ✅

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
```diff
@@ -381,7 +381,7 @@ extension ChatChannel {
         let reads: [ChatChannelRead] = try dto.reads.map { try $0.asModel() }
         
         let unreadCount: () -> ChannelUnreadCount = {
-            guard let currentUser = context.currentUser else { return .noUnread }
+            guard dto.isValid, let currentUser = context.currentUser else { return .noUnread }
```
nuno-vieira, line 384: And what about checking if it is deleted as well? If the Channel is deleted, we are not interested in fetching its data right?

polqf: `isValid` will return false if the dto is deleted.

nuno-vieira (approved):
LGTM! ✅

Issue #2094, opened 2022-06-16
Message List Scroll Issue


jitendra-simform:
## What did you do?
Send message to other user

## What did you expect to happen?
Receiver user message list screen should not jump or auto scroll 

## What happened instead?
Receiver user message list screen jump and auto scroll

## GetStream Environment
GetStream Chat version: 4.15
GetStream
iOS version: 15.2
Swift version: 5
Xcode version: 13.2
Device: Simulator

https://user-images.githubusercontent.com/97155807/174076637-6a6c4164-94b3-45af-816d-694d49e02c01.mov

b-onc:
Hello @jitendra-simform ,

This is known and internally tracked as CIS-1347.

nuno-vieira:
Hi @jitendra-simform!

This issue will be released in the 4.21.0 version.

We'll let you know, once we publish the release. The ETA is next week. 

Best,
Nuno

jitendra-simform:
> 

Thanks for update @nuno-vieira

nuno-vieira:
Hi!

This issue is now solved in the **4.21.1** release.

Thank you for your patience! 

Best,
Nuno

PR #2208, opened 2022-08-02
[CIS-2084] Fix message list jumps by removing the inverted table view

nuno-vieira:
### 🔗 Issue Links
CIS-2084
https://github.com/GetStream/stream-chat-swift/issues/2094

### 🎯 Goal
Fx the jumping issues we currently have with the inverted table view solution.

### 📝 Summary
The inverted table view is the root cause of the jumps. Since our message list cells have unpredictable heights, especially because they can be totally customizable, the table view can't properly calculate the new content size of the table when new messages are being inserted. The jumps are basically caused by this, and when the message is not inverted, this is not a problem, because the table view grows to the bottom of the screen.

**This solution tries to remove the need for inverting the table view by:**
- If we were previously at the bottom of the message list, and a new message is inserted, keep the message list at the bottom by scrolling to the newest message always
- When loading the initial messages, we also make sure it starts at the bottom, by scrolling to the bottom.
- When loading previous messages, prepend the messages at the top and adjust the scroll content offset.

**This implementation requires some UX changes:**
- The loading of the previous messages is not prefetched anymore, and we show a loading spinner at the top. This is how iMessage, Instagram and FB Messenger are doing it as well. The reason is that without this, the scrolling would stop weirdly because we need to adjust the content offset of the table view when prepending new items.
- When opening a message list with few messages, they stick to the top instead of the bottom like it was before. This is also how the other major apps are doing it as well, so it doesn't seem like a problem.

**Known Issues:**
- When opening the message list, the start at the bottom is not yet very stable.
- Keyboard is not adjusting properly when it is open (Working on it at the moment)

**Why it is draft:**
- We are currently exploring another solution which is actually the one used by the SwiftUI SDK, and that is to keep everything like it is right now, and skip insertions at the bottom when the user is scrolled to the top. But this requires using a proper diffing algorithm, so until we check which is the best solution, we will keep this one in Draft.

### 🎨 Showcase

| Loading Previous Messages | Few messages UX | Stick to the bottom |
| ------ | ----- | ----- |
|  <video src="https://user-images.githubusercontent.com/12814114/182442291-6d7473c8-6519-4d4f-87d1-5a726181d421.mp4"/>   |  <video src="https://user-images.githubusercontent.com/12814114/182442421-76b080a3-6c15-49ff-ae02-570d02041309.mp4"/>  |  <video src="https://user-images.githubusercontent.com/12814114/182442467-6eaffcc0-4f1a-4cb3-a51a-a88bb9fcd639.mp4"/>  |

### 🧪 Manual Testing Notes

Full regression is required

### ☑️ Contributor Checklist

### DIFF
--- a/DemoApp/DemoAppConfiguration.swift
+++ b/DemoApp/DemoAppConfiguration.swift
@@ -32,7 +32,6 @@ enum DemoAppConfiguration {
 
         configureAtlantisIfNeeded()
         trackPerformanceIfNeeded()
-        enableMessageDiffingIfNeeded()
     }
 
     // HTTP and WebSocket Proxy with Proxyman.app
@@ -50,10 +49,5 @@ enum DemoAppConfiguration {
             PerformanceMonitor.shared().performanceViewConfigurator.options = [.performance]
             PerformanceMonitor.shared().start()
         }
-    }
-
-    // Enable message diffing in Message List
-    private static func enableMessageDiffingIfNeeded() {
-        StreamChatWrapper.shared.setMessageDiffingEnabled(isStreamInternalConfiguration)
     }
 }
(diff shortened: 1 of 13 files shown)


nuno-vieira:
Closing the PR for now because the DifferenceKit solution is looking much better.

PR #2216, opened 2022-08-04
[CIS-2083] Fix hidden channels showing past history

polqf:
### 🔗 Issue Links

https://stream-io.atlassian.net/browse/CIS-2083

### 🎯 Goal

Fix an issue where hidden history would be visible for some time after hiding a channel

### 📝 Summary

There was an error in the logic to handle incoming truncatedAt values from the Backend, setting an old value for it. This is because we currently use `truncatedAt` for both hiding and truncating channels.

### 🛠 Implementation

The value for `truncatedAt` is now only updated if the backend returns a date for `truncatedAt` AND if that date is older than the currently set in the DTO

### 🧪 Manual Testing Notes

1. Create a 1-1 channel
2. Send few messages
3. User A hides the channel, clearing history
4. User B sends more messages

Expected result:
- Only the new messages are shown

Previous result:
- Some old messages were shown

### ☑️ Contributor Checklist


nuno-vieira (approved):
LGTM! ✅

Issue #2184, opened 2022-07-22
Share picture blinks


jrBordet:
## What did you do?

- Share picture or take photo
- Send it on chat
- The images sent <b>blinks</b> one time on the thread

## What did you expect to happen?

The image should not <b>blink</b>

## GetStream Environment
**GetStream Chat version: 4.18.0**
**GetStream Chat frameworks:** StreamChat, StreamChatUI
**iOS version:**
**Swift version:**
**Xcode version:**
**Device:**

polqf:
Hi @jrBordet,

Thank you for your report. We've internally logged it and we will work on it!

jrBordet:
Hi guys, is there some news about this issue? Can I do something?

nuno-vieira:
Hi @jrBordet,

This issue will be fixed by https://github.com/GetStream/stream-chat-swift/pull/2226. It will come on the next 4.21.0 release.

We will let you know when we officially release 4.21.0. The ETA is next week.

Best,
Nuno

nuno-vieira:
Hi!

This issue is now solved in the **4.21.1** release.

Thank you for your patience! 

Best,
Nuno

PR #2226, opened 2022-08-09
[CIS-2084, CIS-2085] Refactor Message List with DifferenceKit

nuno-vieira:
### 🔗 Issue Links
CIS-2084
CIS-2085
https://github.com/GetStream/stream-chat-swift/issues/2094

Bonus:
https://github.com/GetStream/stream-chat-swift/issues/2184

### 🎯 Goal
The goal of this PR is to fix the crashes in the message list related to data inconsistencies in `performBatchUpdates` and also to fix the message list jumps when receiving multiple messages and the user is scrolled up.

### 📝 Summary
- Adds [DifferenceKit](https://github.com/ra1028/DifferenceKit) dependency to `StreamChatUI`.
- Removes the`_messageListDiffingEnabled` flag, since now we only have 1 implementation of the message list and it is not experimental anymore
- Instead of relying on `ListChange` updates from `StreamChat`, now the `StreamChatUI` applies changes from the diffing calculation of the old messages snapshot, and the newest message snapshot
- Skips bottom insertions at the bottom when the user is scrolled up.  With the diffing logic in place, now it is easier to skip bottom insertion updates when the user is scrolled up. This avoids the message list jumps because the insertions at the bottom don't actually happen until the user scrolls to the bottom again. This not only avoids the message list jumps but also makes the message list more performant and efficient.
- Improves the performance of the message list when multiple messages are being inserted at the same time and there are already quite some messages loaded. (Most of this improvement was by skipping channel list updates while the channel list is not visible. After profiling the application, the main thread was mostly blocked by the channel list updates, and the root cause is being fixed here: https://github.com/GetStream/stream-chat-swift/pull/2168. Either way, it doesn't make sense for the channel list to impact the message list while the user is on the message list. The https://github.com/GetStream/stream-chat-swift/pull/2168 PR will also improve even further the message list performance especially when it is being overloaded. Right now the DTO mappings on the main thread are the only optimisation left in the Message List. Even though we have right now a good 56/60FPS average when the message list is being overloaded.

Missing Tasks:
- [ ] Skipping logic test coverage
- [ ] Properly add the DifferenceKit dependency using the bash script

### 🎨 Showcase
TODO

### 🧪 Manual Testing Notes
- [ ] Mob Regression Testing

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
+++ b/Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
@@ -195,7 +195,7 @@ open class ChatChannelVC: _ViewController,
             return
         }
 
-        if indexPath.row < channelController.messages.count - 10 {
+        if indexPath.row < messages.count - 10 {
             return
         }
 
--- a/Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
+++ b/Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
@@ -237,6 +237,11 @@ open class ChatChannelVC: _ViewController,
             channelController.markRead()
 
             messageListVC.scrollToLatestMessageButton.content = .noUnread
+
+            skippedMessages = []
+            if messages.count != channelController.messages.count {
+                channelController(channelController, didUpdateMessages: [])
+            }
         }
     }
 
@@ -250,8 +255,10 @@ open class ChatChannelVC: _ViewController,
 
     // MARK: - ChatChannelControllerDelegate
 
+    var skippedMessages: Set<MessageId> = []
+
     open func channelControllerWillUpdateMessages(_ channelController: ChatChannelController) {
-        messageListVC.listView.previousMessagesSnapshot = Array(channelController.messages)
+        messageListVC.listView.previousMessagesSnapshot = channelController.messages.filter { !skippedMessages.contains($0.id) }
     }
 
     open func channelController(
@@ -262,7 +269,17 @@ open class ChatChannelVC: _ViewController,
             channelController.markRead()
         }
 
-        messageListVC.listView.newMessagesSnapshot = Array(channelController.messages)
+        debugPrint("Insertions:", changes.filter(\.isInsertion).map(\.indexPath))
+        if !isLastMessageFullyVisible && changes.first(where: { $0.indexPath.item == 0 }) != nil {
+            changes.filter(\.isInsertion).forEach {
+                skippedMessages.insert($0.item.id)
+            }
+            // Don't update messages if we are skipping
+            return
+        }
+
+        let newMessages = channelController.messages.filter { !skippedMessages.contains($0.id) }
+        messageListVC.listView.newMessagesSnapshot = newMessages
         messageListVC.updateMessages(with: changes)
     }
 
--- a/Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
+++ b/Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
@@ -2,7 +2,6 @@
 // Copyright © 2022 Stream.io Inc. All rights reserved.
 //
 
-import DifferenceKit
 import StreamChat
 import UIKit
 
@@ -99,6 +98,9 @@ open class ChatChannelVC: _ViewController,
             setChannelControllerToComposerIfNeeded(cid: self?.channelController.cid)
             self?.messageComposerVC.updateContent()
         }
+
+        // Initial messages data
+        messages = Array(channelController.messages)
     }
 
     override open func setUpLayout() {
@@ -151,15 +153,7 @@ open class ChatChannelVC: _ViewController,
 
     // MARK: - ChatMessageListVCDataSource
 
-    private var _messages: [DiffChatMessage] = []
-
-    public var messages: [ChatMessage] {
-        if _messages.isEmpty {
-            return Array(channelController.messages)
-        }
-
-        return _messages.map(\.message)
-    }
+    public var messages: [ChatMessage] = []
     
     open func channel(for vc: ChatMessageListVC) -> ChatChannel? {
         channelController.channel
@@ -256,75 +250,20 @@ open class ChatChannelVC: _ViewController,
 
     // MARK: - ChatChannelControllerDelegate
 
-    var previousMessagesSnapshot: [DiffChatMessage] = []
-
     open func channelControllerWillUpdateMessages(_ channelController: ChatChannelController) {
-        previousMessagesSnapshot = channelController.messages.map(DiffChatMessage.init)
+        messageListVC.listView.previousMessagesSnapshot = Array(channelController.messages)
     }
 
     open func channelController(
         _ channelController: ChatChannelController,
         didUpdateMessages changes: [ListChange<ChatMessage>]
     ) {
-        let target = Array(channelController.messages).map(DiffChatMessage.init)
-        let changeset = StagedChangeset(
-            source: previousMessagesSnapshot,
-            target: target
-        )
-
         if isLastMessageFullyVisible {
             channelController.markRead()
         }
-//        messageListVC.updateMessages(with: changes)
 
-//        messageListVC.listView.reload(
-//            using: changeset,
-//            with: .fade,
-//            setData: { [weak self] newData in
-//                self?._messages = newData
-//            },
-//            completion: {
-//                if changeset.first?.elementInserted.contains(.init(element: 0, section: 0)) == true {
-//                    self.messageListVC.listView.scrollToRow(at: IndexPath(item: 0, section: 0), at: .bottom, animated: true)
-//                }
-//            }
-//        )
-
-        let newMessageInserted = changeset.first?.elementInserted.contains(.init(element: 0, section: 0)) == true
-
-        let reload = {
-            self.messageListVC.listView.reload(
-                using: changeset,
-                with: .fade,
-                setData: { [weak self] newData in
-                    self?._messages = newData
-                }
-            )
-        }
-
-        if messageListVC.listView.isLastCellFullyVisible {
-            UIView.performWithoutAnimation {
-                reload()
-            }
-        } else {
-            reload()
-        }
-
-        if newMessageInserted && messageListVC.listView.isLastCellFullyVisible {
-            UIView.performWithoutAnimation {
-                self.messageListVC.listView.reloadRows(at: [IndexPath(item: 1, section: 0)], with: .none)
-            }
-        }
-
-        if newMessageInserted && target.first?.message.isSentByCurrentUser == true {
-            DispatchQueue.main.asyncAfter(deadline: .now() + 0.3) {
-                self.messageListVC.listView.scrollToRow(at: IndexPath(item: 0, section: 0), at: .bottom, animated: true)
-            }
-        }
-
-//        if changeset.first?.elementInserted.contains(.init(element: 0, section: 0)) == true {
-//            messageListVC.listView.scrollToRow(at: IndexPath(item: 0, section: 0), at: .bottom, animated: true)
-//        }
+        messageListVC.listView.newMessagesSnapshot = Array(channelController.messages)
+        messageListVC.updateMessages(with: changes)
     }
 
     open func channelController(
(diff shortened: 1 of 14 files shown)


Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -117,6 +97,11 @@ open class ChatMessageListVC: _ViewController,
     open var isDateSeparatorEnabled: Bool {
         components.messageListDateSeparatorEnabled
     }
+
+    /// The message cell height caches. This makes sure that the message list doesn't
+    /// need to recalculate the cell height every time. This improve the scrolling
+    /// experience since the content size calculation is more precise.
+    private var cellHeightsCache: [MessageId: CGFloat] = [:]
```
martinmitrevski, line 104: I'm not sure if the `MessageId` is the right key here. Imagine a case where we edit a message and add larger / smaller text. The id of the message would be the same, but the cell size not. 
Another thing to consider is if adding reactions / thread replies affects this size. 
To sum up, I would go with a computed `Differentiable` ID as a key instead.

Sources/StreamChatUI/ChatMessageList/ChatMessageListView+DiffKit.swift
```diff
@@ -0,0 +1,59 @@
...
+            && extraData == source.extraData
+            && quotedMessage == source.quotedMessage
+            && isShadowed == source.isShadowed
+            && reactionCounts.count == source.reactionCounts.count
+            && reactionScores.count == source.reactionScores.count
+            && threadParticipants.count == source.threadParticipants.count
+            && localState == source.localState
+            && isFlaggedByCurrentUser == source.isFlaggedByCurrentUser
+            && readBy.count == source.readBy.count
+            && giphyAttachments.map(\.previewURL) == source.giphyAttachments.map(\.previewURL)
```
martinmitrevski, line 52: Might be good to extract these in computed variables, e.g. uploadingStatesId, reactionScoresId, etc. If we group them it would be easier to reason about them, and potentially write tests.

Sources/StreamChatUI/ChatMessageList/ChatMessageListView.swift
```diff
@@ -9,10 +9,52 @@ import UIKit
...
+            let isNotSkippedMessage: (ChatMessage) -> Bool = {
+                !self.skippedMessages.contains($0.id)
+            }
+            previousMessagesSnapshot = previousMessagesSnapshot.filter(isNotSkippedMessage)
+        }
+    }
+
+    /// The new messages snapshot reported by the channel or message controller.
+    private var _newMessagesSnapshotWithSkipped: [ChatMessage] = []
+    internal var newMessagesSnapshot: [ChatMessage] = [] {
```
martinmitrevski, line 43: While testing, sth weird happened when I scrolled to the top and scrolled down fast. Only 5 messages were shown all the time (in the channel `hello everyone` that has much more) - it seems we might have an edge case in this logic. My assumption is that the skipped messages were not cleared?

Sources/StreamChatUI/ChatMessageList/ChatMessageListView.swift
```diff
@@ -156,12 +195,52 @@ open class ChatMessageListView: UITableView, Customizable, ComponentsProvider {
...
         completion: (() -> Void)? = nil
     ) {
-        listChangeUpdater.performUpdate(with: changes) { _ in
-            completion?()
+        let bottomChange = changes.first(where: { $0.indexPath.item == 0 })
+        if !isLastCellFullyVisible && bottomChange?.isInsertion == true && bottomChange?.item.isSentByCurrentUser == false {
+            changes.filter(\.isInsertion).forEach {
+                skippedMessages.insert($0.item.id)
+            }
+            // Don't update UI if we are skipping the inserted messages
```
martinmitrevski, line 203: If we handle it like this, then when a new message arrives, the reactions of the visible messages are not reflected in the UI. We need to apply the updates in cases like this, just ignore the inserts.

martinmitrevski (commented):
Great work @nuno-vieira 👏 . It seems much more stable now, and the performance is better.
I've noticed few issues that we need to fix before merging:
- bug with pagination broken (see comments)
- reactions don't work when in "skip inserts" mode (also in comments)
Additionally, we would need some tests before we can merge it.

Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
```diff
@@ -98,6 +98,9 @@ open class ChatChannelVC: _ViewController,
             setChannelControllerToComposerIfNeeded(cid: self?.channelController.cid)
             self?.messageComposerVC.updateContent()
         }
+
+        // Initial messages data
+        messages = Array(channelController.messages)
```
polqf, line 103: Just a note: We should not expect messages to be available in setUp, specially with the coming up changes on Background Mappings.

Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
```diff
@@ -81,6 +81,11 @@ open class ChatChannelListVC: _ViewController,
     private lazy var listChangeUpdater: ListChangeUpdater = CollectionViewListChangeUpdater(
         collectionView: collectionView
     )
+
+    /// Currently there are some performance problems in the Channel List which
+    /// is impacting the message list performance as well, so we skip channel list
+    /// updates when the channel list is not visible in the window.
+    private var skippedRendering = false
```
polqf, line 88: That's a bit sketchy... Is this meant to stay forever or just temporarily?

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
polqf: I think computing ids might be a bit more costly. Is there any way for us to clear the cache whenever the message is updated, somehow?

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -449,13 +419,31 @@ open class ChatMessageListVC: _ViewController,
...
+        return UITableView.automaticDimension
+    }
+
     open func scrollViewDidScroll(_ scrollView: UIScrollView) {
         delegate?.chatMessageListVC(self, scrollViewDidScroll: scrollView)
 
         setScrollToLatestMessageButton(visible: isScrollToBottomButtonVisible)
+
+        // If the user scrolled to the bottom, update the UI for the skipped messages
+        if listView.isLastCellFullyVisible && !listView.skippedMessages.isEmpty {
```
polqf, line 443: I am a bit concerned about the pieces of custom logic. Is this something we can solve any other way? This can be one of those things that we end up maintaining without knowing how to remove

Sources/StreamChatUI/ChatMessageList/ChatMessageListView+DiffKit.swift
```diff
@@ -0,0 +1,59 @@
...
+            with: animation()
+        ) { [weak self] newMessages in
+            self?.onNewDataSource?(newMessages)
+        }
+        CATransaction.commit()
+    }
+}
+
+extension ChatMessage: Differentiable {
+    public func isContentEqual(to source: ChatMessage) -> Bool {
```
polqf, line 35: Just a note: This can actually be quite expensive right now

Sources/StreamChatUI/ChatMessageList/ChatMessageListView.swift
```diff
@@ -9,10 +9,52 @@ import UIKit
...
+        didSet {
+            let isNotSkippedMessage: (ChatMessage) -> Bool = {
+                !self.skippedMessages.contains($0.id)
+            }
+            previousMessagesSnapshot = previousMessagesSnapshot.filter(isNotSkippedMessage)
+        }
+    }
+
+    /// The new messages snapshot reported by the channel or message controller.
+    private var _newMessagesSnapshotWithSkipped: [ChatMessage] = []
```
polqf, line 42: Yeah, the more I see skipped messages related logic, the more concerned I get 🙏

polqf (commented):
Overall looks really good, I am just concerned with certain pieces of custom logic. We don't want this evolution to end up being like the previous one where it was full of patches and custom logic. 

Also, we should be embedding DifferenceKit as part of StreamChatUI

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
nuno-vieira: It doesn't really matter when you edit the message or update the message. Keep in mind this is just the cell height estimation. It doesn't need to be exact, and not even close to the real height. This is only important when loading the previous messages. We can indeed check if a message was updated, and remove it from the cache, I actually did this already, but from my experience, this doesn't have any real improvement. So, for me, it does not make sense to make this more complex if we don't need to. For now, everything is working fine :)

Sources/StreamChatUI/ChatMessageList/ChatMessageListView.swift
nuno-vieira: Do you mean, you couldn't load more messages? Or for a split second, you only saw 5 messages? Because the CATransition hack, sometimes doesn't work 100% and there could be animation glitches

Sources/StreamChatUI/ChatMessageList/ChatMessageListView.swift
nuno-vieira: Initially, I had it like you are saying, but it was causing some problems from what I can remember. So I decided to make it more simple. The reality is that usually, this is not a visible problem, because the messages being added and updated are 90% of the time out of the screen when the index == 0. But I'll double-check this one and see what I can do 👍

Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
nuno-vieira: For now, these are used to render the initial offline messages only. Without this, it won't work with the current setup. If we change to background mappings , then we can think about how to improve this 👍

Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
nuno-vieira: For now, it is temporary, yes, this was making the message list totally unusable with 100+ messages, especially if the stress test script was running at the same time. If we fix the channel list later, we probably can remove this all together.

Sources/StreamChatUI/ChatMessageList/ChatMessageListView+DiffKit.swift
martinmitrevski: This is more subjective, like better readability if they are split up. However, totally up to you to decide if it makes sense.

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
nuno-vieira: Unfortunately, we don't have any other option. Of all the solutions I tried, none fixes the jumps in the message list. We have been dealing with jumps in the message list since I joined Stream 😅 When I thought about this solution I was thinking it could be problematic and error-prone, but actually after talking with @martinmitrevski, this is what they are actually doing on SwiftUI. So if it is working for them, and it is stable, it means we can do it as well.

Sources/StreamChatUI/ChatMessageList/ChatMessageListView.swift
martinmitrevski: No, I had only 5 messages the whole time. I had to leave the channel and come back to get it back to working. Will try to have proper steps on the mob testing session.

Sources/StreamChatUI/ChatMessageList/ChatMessageListView+DiffKit.swift
nuno-vieira: Yes, this is not perfect right now, because it runs some lazy properties which hit the DB. Still, for most use cases we have really nice performance. But when the message list is being completely stressed (1000+ messages and the stress test running) it drops a little bit the FPS, but currently, we don't have another option to improve this.

Sources/StreamChatUI/ChatMessageList/ChatMessageListView.swift
nuno-vieira: This is more complex than it actually should be because we are trying to avoid breaking changes. Otherwise, this would look very simple. This [commit](https://github.com/GetStream/stream-chat-swift/pull/2226/commits/955802ea856f903089547ec88674c71d7b2ad87e) shows all the skipped message logic before moving the code to the `ChatMessageListView`. Moving the logic to the message list view, in order to make it work for both the Channel List and Thread, makes things a little bit more tricky.

Sources/StreamChatUI/ChatMessageList/ChatMessageListView.swift
martinmitrevski: The main issue here is that when I've scrolled up and read older messages, if I react to one of them, there's no visual feedback. And the use-case is pretty common, so I think it's worth revisiting this.

Sources/StreamChatUI/ChatMessageList/ChatMessageListView+DiffKit.swift
nuno-vieira: IMO I think having the equality in this function, and clear see what is important and not, make it easy to and understand in comparison of hiding it behind a computed property, but maybe I'm not understanding the full picture 🤔

Sources/StreamChatUI/ChatMessageList/ChatMessageListView.swift
nuno-vieira: Ah I see, this is worrying then, definitely should not happen.  I'll investigate this one 👍

Sources/StreamChatUI/ChatMessageList/ChatMessageListView.swift
nuno-vieira: When you scroll down to read older messages, the cell will be reused and the data will be updated by that time 🤔 But I'll investigate this one a little bit more too and will comeback with feedback soon 👍

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
martinmitrevski: yes, this works well in SwiftUI, so far we haven't received any complaints. However, it's important that we find a way to apply any other updates (e.g. reactions, deletes), since this is the part where it differs from SwiftUI atm.

Sources/StreamChatUI/ChatMessageList/ChatMessageListView.swift
nuno-vieira: I was able to understand the problem and now it is fixed, all updates are still executed, but the insertions :)

Sources/StreamChatUI/ChatMessageList/ChatMessageListView.swift
nuno-vieira: While fixing the other problem, I noticed I was not using the previous snapshot correctly! So I think the fix I did for the missing updates, also most likely fixed this one :) But on the mob testing we should be able to confirm it 👍

Sources/StreamChatUI/ChatMessageList/ChatMessageListView.swift
nuno-vieira: I've tried to improve a little bit the readability as much as I could with the limited constraints we have, but let me know what you think !

polqf (approved):

PR #2237, opened 2022-08-16
[CIS-1854] [Internal] Move Model mappings to Background

polqf:
### 🔗 Issue Links

https://stream-io.atlassian.net/browse/CIS-1854
https://stream-io.atlassian.net/browse/CIS-2059

### 🎯 Goal

The goal of this PR is to investigate the viability of moving the mapping of DTOs to Model to a background thread, reducing the pressure we put in the view/main thread.

### 📝 Summary

This PR introduces a new ListDatabaseObserver, matching the same API for simplicity purposes. This new object, named BackgroundListDatabaseObserver does exactly the same as its predecessor, but in a background and asynchronous way, and only notifies about results when those are already mapped.

### 🛠 Implementation

As of now:

- The observer listens to a particular fetch request and processes the results of any update that might be coming from the Database
- It converts the latest snapshot of DTOs into Models in a background thread, thus not leading to pressure on the main thread.
- It calls its delegate (in this case a block) whenever all the objects have been parsed
- Because all the mapped objects are assigned to a property on the same (internal) thread, that should result in no data races
- CoreDataLazy is effectively disabled under the runtime flag `_isBackgroundMappingEnabled `, unless `forceLazy` is passed

All this implementation is guarded now behind `_isBackgroundMappingEnabled`. When the flag is false, the same behaviour as before will be present.

When using DemoApp-StreamDevelopers, it uses Background Mapping by default

Pitfalls:

Because now background mapping tries to map every single updated dto as new updates come in, we end up resolving the whole tree for each object, which results in way too many fetches. This has no direct impact on the main thread per se, but the overhead we add to Core Data can be significant in some cases. 

Also, in some cases, there are cyclic relationships, leading to Stack Overflow. This is now prevented from passing `forceLazy: true` to some models. This preserves the laziness of those properties. This is a temporary solution to those problems, but it is not meant to be the final one

### ☑️ Contributor Checklist


martinmitrevski (commented):
Code-wise looks really good. I did some testing and there are solid performance improvements in the message list in UIKit. In SwiftUI, there's still a noticeable glitch when loading new messages, but we can iterate on this in the future.
To merge this, we would need some tests, and let's see how we communicate this change. Will there be doc update, where customers can opt in?

polqf:
> Will there be doc update, where customers can opt in?
Not as of now, the idea is to use it internally, make sure things get to a better stage, and then allow customers to use it.
At some point down the line, when things are stable enough, enable it for everyone

Sources/StreamChat/Controllers/BackgroundListDatabaseObserver.swift
```diff
@@ -0,0 +1,297 @@
...
+    /// Called with the aggregated changes after the internal `NSFetchResultsController` calls `controllerWillChangeContent`
+    /// on its delegate.
+    var onWillChange: (() -> Void)? {
+        didSet {
+            if isBackground {
+                background?.onWillChange = onWillChange
+            } else {
+                foreground?.onWillChange = onWillChange
+            }
+        }
```
nuno-vieira, line 33: All these ifs can be avoided if we create a `ListDatabaseObserving` protocol since the background and the regular observer have the same interface. Then, we would only need to do the if once, when creating the observer.

nuno-vieira (commented):
Overall LGTM, the only thing I think we could improve is the Wrapper, and use a protocol instead. Or still, use the Wrapper, but inside, we could just move the ifs statements, to only one if statement, when creating the observer.

Sources/StreamChat/Controllers/BackgroundListDatabaseObserver.swift
```diff
@@ -0,0 +1,297 @@
...
+                fetchedResultsControllerType: fetchedResultsControllerType
+            )
+        } else {
+            foreground = ListDatabaseObserver(
+                context: database.viewContext,
+                fetchRequest: fetchRequest,
+                itemCreator: itemCreator,
+                fetchedResultsControllerType: fetchedResultsControllerType
+            )
+        }
```
nuno-vieira, line 70: Another option, is to maintain the Wrapper, but here, we only have one observer, that is initialised depending on isBackground. In that case, the if, is only here, and not spread across the wrapper

nuno-vieira (approved):
LGTM! ✅

martinmitrevski (approved):

Issue #1793, opened 2022-02-15
In ChannelView, Swipe-to-navigate-back causes the scrollView to shift offset if keyboard is up.


linhewinner:
https://user-images.githubusercontent.com/97132583/154113306-521e72f9-77c7-4f86-875f-9b6be433391d.mp4

## What did you do?
In ChatView, bring up the keyboard. Swipe to navigate back to thread list. Notice that the message list scrollView is scrolling down as you swipe.

## What did you expect to happen?
The scrollView offset should be stable during the swipe

## What happened instead?
The scrollView offset steadily decreases as you swipe.
Try Signal, iMessages, WhatsApp, Messenger. None of them have this behavior.

## GetStream Environment
**GetStream Chat version:** 4.10.0
**GetStream Chat frameworks:** StreamChat, StreamChatUI
**iOS version:** iOS 15.3
**Swift version:** 5.5.2
**Xcode version:** 13.2.1
**Device:** iPhone 13 Pro

nuno-vieira:
Hi @linhewinner !

Thank you for the report! We will investigate this as soon as possible and will update you whenever we have a fix. 

Best,
Nuno

nuno-vieira:
Hi @linhewinner!

Sorry for a long time without updates. I've already identified the root cause of the issue and will provide a fix very soon. It should be available on the next release 4.21.0. 

We will let you know once we officially release it. 

Thank you for your patience.

Best,
Nuno

nuno-vieira:
Hi!

This issue is now solved in the **4.21.1** release.

Thank you for your patience! 

Best,
Nuno

PR #2239, opened 2022-08-17
[CIS-1591] Fix message list scrolling when popping from navigation stack

nuno-vieira:
### 🔗 Issue Links
CIS-1591
https://github.com/GetStream/stream-chat-swift/issues/1793

### 🎯 Goal
Fix the message list scrolling when popping the message list from the navigation stack.

### 🛠 Implementation
The reason this happened was that the Keyboard hide notification is still triggered when popping from the navigation stack, and the keyboard is dismissed but there's no animation, only the message list animates. To fix this, on the keyboard handler we check if the view controller is still present in the navigation stack. If not, we don't hide the keyboard.

### 🎨 Showcase

| Before | After |
| ------ | ----- |
|  <video src="https://user-images.githubusercontent.com/12814114/185121501-01f8bb92-d1a7-4474-947f-f5e445635c52.mov"/>  |  <video src="https://user-images.githubusercontent.com/12814114/185127580-486a01ac-bf06-4499-801a-aef747f63f31.mp4"/>  |

### 🧪 Manual Testing Notes
- Open a Channel
- Dismiss a Channel
- The message list should not move

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChatUI/Composer/ComposerKeyboardHandler.swift
+++ b/Sources/StreamChatUI/Composer/ComposerKeyboardHandler.swift
@@ -63,8 +63,18 @@ open class ComposerKeyboardHandler: KeyboardHandler {
               let frame = (userInfo[UIResponder.keyboardFrameEndUserInfoKey] as? NSValue)?.cgRectValue,
               let duration = userInfo[UIResponder.keyboardAnimationDurationUserInfoKey] as? Double,
               let curve = userInfo[UIResponder.keyboardAnimationCurveUserInfoKey] as? UInt,
-              let composerParentView = composerParentVC?.view else {
+              let composerParentView = composerParentVC?.view,
+              let composerParentVC = composerParentVC else {
             return
+        }
+
+        // This fixes a UI Glitch when popping the channel from the navigation stack,
+        // which makes the table view size to change because the hide notification is triggered.
+        if let viewControllersInNavigationStack = composerParentVC.navigationController?.viewControllers {
+            let isBeingPopped = !viewControllersInNavigationStack.contains(composerParentVC)
+            if isBeingPopped {
+                return
+            }
         }
 
         // When hiding, we reset the bottom constraint to the original value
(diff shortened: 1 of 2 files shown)


Sources/StreamChatUI/Composer/ComposerKeyboardHandler.swift
```diff
@@ -63,10 +63,20 @@ open class ComposerKeyboardHandler: KeyboardHandler {
               let frame = (userInfo[UIResponder.keyboardFrameEndUserInfoKey] as? NSValue)?.cgRectValue,
               let duration = userInfo[UIResponder.keyboardAnimationDurationUserInfoKey] as? Double,
               let curve = userInfo[UIResponder.keyboardAnimationCurveUserInfoKey] as? UInt,
-              let composerParentView = composerParentVC?.view else {
+              let composerParentView = composerParentVC?.view,
```
martinmitrevski, line 66: Small detail that doesn't really change anything - only one unwrapping is needed here now, for the composerParentVC.
```
guard composerParentVC?.presentedViewController == nil,
              let userInfo = notification.userInfo,
              let frame = (userInfo[UIResponder.keyboardFrameEndUserInfoKey] as? NSValue)?.cgRectValue,
              let duration = userInfo[UIResponder.keyboardAnimationDurationUserInfoKey] as? Double,
              let curve = userInfo[UIResponder.keyboardAnimationCurveUserInfoKey] as? UInt,
              let composerParentVC = composerParentVC else {
            return
        }
        
        let composerParentView = composerParentVC.view
```

martinmitrevski (approved):
✅

Sources/StreamChatUI/Composer/ComposerKeyboardHandler.swift
nuno-vieira: It's actually true because the view from the view controller is also optional, so we need both 👍

martinmitrevski: You are right, disregard my comment 😁

PR #2241, opened 2022-08-17
[CIS-1771] Expose method for clearing local DB data

hugobernalstream:
Add logout method for properly disconnecting users and clearing local…… data.

### 🔗 Issue Links

[CIS-1771](https://stream-io.atlassian.net/browse/CIS-1771)

### 🎯 Goal

Provide a way to properly logout the users by disconnecting them form the service (as currently done) and clearing the related stored local data.

### 🛠 Implementation

- Disconnect and remove all data from the databaseContainer when calling logout.

### 🎨 Showcase

logout - data cleared and reloaded

Files being updated after removing stores and being refreshed.

### 🧪 Manual Testing Notes

Implemented on the DemoApp so can be tested when logging out the user.

### ☑️ Contributor Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 # Upcoming
 
 ## StreamChat
+### ✅ Added
+- Logout from `ChatClient` [#2241](https://github.com/GetStream/stream-chat-swift/pull/2241)
 ### 🐞 Fixed
 - Fix hidden channels showing past history [#2216](https://github.com/GetStream/stream-chat-swift/pull/2216)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## StreamChat
 ### ✅ Added
-- Logout from `ChatClient` [#2241](https://github.com/GetStream/stream-chat-swift/pull/2241)
+- From now on, if you want to logout the user from the app, especially when switching users, you should call the `client.logout()` method instead of `client.disconnect()`. Read more [here](https://getstream.io/chat/docs/sdk/ios/uikit/getting-started/#disconnect--logout) [#2241](https://github.com/GetStream/stream-chat-swift/pull/2241)
 ### 🐞 Fixed
 - Fix hidden channels showing past history [#2216](https://github.com/GetStream/stream-chat-swift/pull/2216)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### 🐞 Fixed
 - Fix hidden channels showing past history [#2216](https://github.com/GetStream/stream-chat-swift/pull/2216)
 
+## StreamChatUI
+### 🐞 Fixed
+- Fix message list scrolling when popping from navigation stack [#2239](https://github.com/GetStream/stream-chat-swift/pull/2239)
+
 # [4.20.0](https://github.com/GetStream/stream-chat-swift/releases/tag/4.20.0)
 _August 02, 2022_
 
--- a/Sources/StreamChat/ChatClient.swift
+++ b/Sources/StreamChat/ChatClient.swift
@@ -373,6 +373,16 @@ public class ChatClient {
         }
         userConnectionProvider = nil
     }
+    
+    /// Disconnects the chat client form the chat servers and removes all the local data related.
+    ///
+    /// - Parameters:
+    ///   - completion: The completion. Will be called when the logout is completed successfully.
+    ///                 If request fails, the completion will be called with an error.
+    public func logout(completion: @escaping ((Error?) -> Void)) {
+        disconnect()
+        databaseContainer.removeAllData(force: true, completion: completion)
+    }
 
     func fetchCurrentUserIdFromDatabase() -> UserId? {
         var currentUserId: UserId?
--- a/Sources/StreamChat/ChatClient.swift
+++ b/Sources/StreamChat/ChatClient.swift
@@ -375,13 +375,16 @@ public class ChatClient {
     }
     
     /// Disconnects the chat client form the chat servers and removes all the local data related.
-    ///
-    /// - Parameters:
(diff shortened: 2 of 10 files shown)


Sources/StreamChat/ChatClient.swift
```diff
@@ -373,6 +373,16 @@ public class ChatClient {
...
+    
+    /// Disconnects the chat client form the chat servers and removes all the local data related.
+    ///
+    /// - Parameters:
+    ///   - completion: The completion. Will be called when the logout is completed successfully.
+    ///                 If request fails, the completion will be called with an error.
+    public func logout(completion: @escaping ((Error?) -> Void)) {
+        disconnect()
+        databaseContainer.removeAllData(force: true, completion: completion)
+    }
```
nuno-vieira, line 385: Do we really need a completion block here? 🤔 If we do, better to provide a default to `nil` as most likely, customers, won't need to wait for this. And also, how could the logout error? 🤔 

CC @polqf

docusaurus/docs/iOS/guides/working-with-user.md
```diff
@@ -118,11 +118,20 @@ chatClient.currentUserController().reloadUserIfNeeded()
...
+    if let error = error {
+        // Handle error
+    }
+
+    // Execute additional actions after logout succeeded such as going back to login view, etc.
+}
 ```
 
+This way the logged-in user will disconnect from the servers and any local stored data related to the user will be deleted.
+
```
nuno-vieira, line 134: We need to add documentation about when to use logout vs disconnect only

docusaurus/sidebars-ios.json
```diff
@@ -113,6 +113,7 @@
         "guides/customize-message-actions",
         "guides/push-notifications",
         "guides/offline-support",
+        "guides/working-with-user",
```
nuno-vieira, line 116: Was this not on the section? 🤔 The rest of the documentation might be outdated then 🤔

Sources/StreamChat/ChatClient.swift
hugobernalstream: I added it mainly because in this case the `removeAllData` could fail and probably is a good idea that the implementers have feedback on that if they want to act depending on the error, but I also agree that a logout might be better without parameters and we address the completion internally by logging for instance similar to what I added on the `StreamChatWrapper` when using the `logout()` method. The default to `nil` approach might be a good idea too, but probably is better to remove the parameters altogether.

docusaurus/sidebars-ios.json
hugobernalstream: I found this because I was able to got the page on the current implementation when searching if there was something related to log out in the docs and to see where to add this docs properly, and was able to open it from the search, but the left side bar wasn't showing so I noticed this wasn't added and it appeared again when added to the side-bars JSON. I remember you had a similar issue with the Extra Data docs if I can remember well?

nuno-vieira: In my case the guide was new and created by me, and it was accidentally deleted by a SwiftUI Docs PR. In this case, every document that is not referenced in the sections means it is an incomplete document and should not be used. 

We should move this documentation to here: https://getstream.io/chat/docs/sdk/ios/uikit/getting-started after the "connect user" section, we should create a "Disconnect vs logout" section. This should be the best place for this 👍

docusaurus/docs/iOS/guides/working-with-user.md
```diff
@@ -118,21 +118,29 @@ chatClient.currentUserController().reloadUserIfNeeded()
...
 to make sure `ChatClient` calls your `tokenProvider` closure and acquires a new token for the newly logged-in user.
 
-You don't need to "logout" in the common sense. Deallocating `ChatClient` instances for the currently logged-in user means the user will disconnect. If you want to keep your "ChatClient" instance (eg you're using a singleton for it) you can call:
+If you're using static tokens for developing quick prototypes and want to test login/logout, you can assign a new `tokenProvider` to the `ChatClient` and then call `reloadUserIfNeeded` to "login" using the new token:
+```swift
+chatClient.tokenProvider = .static(newToken)
+chatClient.currentUserController().reloadUserIfNeeded()
+```
+
+## Disconnect & Logout
```
nuno-vieira, line 127: Move this section to the Getting Started guide, after the "Connect User" section 👍

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 # Upcoming
 
 ## StreamChat
+### ✅ Added
+- Logout from `ChatClient` [#2241](https://github.com/GetStream/stream-chat-swift/pull/2241)
```
nuno-vieira, line 8: It's better to provide more information here, something like:
"From now on, if you want to log out the user from the app, especially when switching users, you should call the client.logout() instead of client.disconnect(). Read more [here](link)"

Sources/StreamChat/ChatClient.swift
polqf: I think it is essential for us to guide our customers to use our APIs in the correct way. The message we send if we don't add a completion block is wrong. We might find customers calling logout and just after trying to log in. This will end up in a race condition where the DB is still being cleared while we are trying to perform additions to the DB

nuno-vieira: I find that a bit difficult, especially because the user will then need to log in, which takes quite some time. I'm fine with adding a completion block, but from a developer perspective is weird. Especially the error. How do you react to a failed logout?

docusaurus/docs/iOS/uikit/getting-started.md
```diff
@@ -72,6 +72,19 @@ ChatClient.shared.connectUser(userInfo: userInfo, tokenProvider: tokenProvider)
 }
 ```
 
+### Disconnect & Logout
+
+Whenever your users leave the chat component, you should use disconnect to stop receiving chat updates and events while using other features of your app. You disconnect by calling:
+```swift
+chatClient.connectionController().disconnect()
```
nuno-vieira, line 79:
```suggestion
chatClient.disconnect()
```

Sources/StreamChat/ChatClient.swift
nuno-vieira: IMO the completion block is not necessary. In real use cases, there’s no case a user could log in as fast as the completion block. If we think it is necessary and some customers still have issues, we can add the completion block later

nuno-vieira (approved):
LGTM! ✅

PR #2253, opened 2022-08-26
[CIS-2045] Add highlighted mentioned users support

hugobernalstream:
### 🔗 Issue Links

CIS-2045

### 🎯 Goal

Highlight user mentions in chat message and provide a delegate for when they are tapped.

### 🛠 Implementation

Highlight mentioned users as if they were links and provide a delegate to act whenever they are tapped for instance to show the user profile or go to a direct message with user, etc.

### 🎨 Showcase
Simulator Screen Recording - iPhone 12 Pro - 2022-08-26 at 13 48 31

### 🧪 Manual Testing Notes

Tap on mentioned users either on message list or in composer.

### ☑️ Contributor Checklist


Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
```diff
@@ -69,6 +73,9 @@ open class ChatMessageContentView: _View, ThemeProvider {
 
     /// The delegate responsible for action handling.
     public weak var delegate: ChatMessageContentViewDelegate?
+    
+    /// The delegate responsible for interactions with mentions.
+    public weak var mentionDelegate: ChatMessageMentionedUserDelegate?
```
nuno-vieira, line 78: Is there any reason why we didn't use the `ChatMessageContentViewDelegate`? IMO it makes sense to keep things consistent for now and move this to the `ChatMessageContentViewDelegate`

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
```diff
@@ -645,6 +662,19 @@ open class ChatMessageContentView: _View, ThemeProvider {
     @objc open func handleTapOnDeliveryStatusView() {
         delegate?.messageContentViewDidTapOnDeliveryStatusIndicator(indexPath?())
     }
+    
+    // MARK: - UITextViewDelegate
+    
+    public func textView(_ textView: UITextView, shouldInteractWith URL: URL, in characterRange: NSRange, interaction: UITextItemInteraction) -> Bool {
```
nuno-vieira, line 668:
```suggestion
   open func textView(_ textView: UITextView, shouldInteractWith URL: URL, in characterRange: NSRange, interaction: UITextItemInteraction) -> Bool {
```

Sources/StreamChatUI/Composer/ComposerVC.swift
```diff
@@ -909,6 +922,17 @@ open class ComposerVC: _ViewController,
...
+    public func textView(_ textView: UITextView, shouldInteractWith URL: URL, in characterRange: NSRange, interaction: UITextItemInteraction) -> Bool {
+        guard let text = textView.text,
+              let range = Range(characterRange, in: text)
+        else { return true }
+        
+        let string = String(text[range].replacingOccurrences(of: "@", with: ""))
+        let user = content.mentionedUsers.first(where: { $0.name == string })
+        didTapOnMentionedUser(user)
+        return true
+    }
```
nuno-vieira, line 935: This is duplicated logic in the Message Content View. We need to reuse this logic somehow. Maybe you can extract this to a component like `TextViewMentionedUserDetector` 🤔

nuno-vieira (commented):
Overall looks good 👍 We just need to improve the code reusability and be more consistent with the current codebase.

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
hugobernalstream: I used a separated delegate because since we also have to provide a similar implementation in the ComposerVC, the ComposerVC would have to conform to the `ChatMessageContentViewDelegate` too, which is unrelated to the ComposerVC, and would have to implement all of the current methods on that protocol which we would have to leave empty.

nuno-vieira: I get the idea, but we are not using the `ChatMessageMentionedUserDelegate` in the `ComposerVC`. We conform to it, but the delegate is never called, we just call directly a function which is part of the `ComposerVC`, that is not how the delegate pattern should be used, the delegate pattern is about delegating responsibility to another component. Here, we are delegating the responsibility to the same component, which does not make sense. Besides that, it is very unlikely that we need to handle clicking on the composer. IMO there's no need to reuse the same delegate in both places (we are not doing it right now either way).

Sources/StreamChatUI/Utils/TextViewMentionedUserDetector.swift
```diff
@@ -0,0 +1,18 @@
...
+    open func handleMentionedUserInteraction(on textView: UITextView, in characterRange: NSRange, _ mentionedUsers: Set<ChatUser>?, onTap: ((ChatUser?) -> Void)?) {
+        guard let text = textView.text,
+              let range = Range(characterRange, in: text)
+        else { return }
+        
+        let string = String(text[range].replacingOccurrences(of: "@", with: ""))
+        let user = mentionedUsers?.first(where: { $0.name == string })
+        onTap?(user)
+    }
+}
```
nuno-vieira, line 18: -  `onTap: ((ChatUser?) -> Void)?` should be a reference property, or it could also be a delegate. But closure are more friendly on SDKs so that we don't break the API easily.
- `mentionedUsers: Set<ChatUser>?` should not be nullable. If there are no mentionedUsers, then don't call this method. 

Here is a suggestion of improvement:
```suggestion
open class TextViewMentionedUsersHandler {
    var onMentionedUserTap: ((ChatUser?) -> Void)? // This can also be a delegate

    open func handleInteraction(
        on textView: UITextView,
        in characterRange: NSRange,
        withMentionedUsers mentionedUsers: Set<ChatUser> // This should not be nil, if there are no mentioned users, don't call this method
    ) {
        guard let text = textView.text,
              let range = Range(characterRange, in: text)
        else { return }

        let string = String(text[range].replacingOccurrences(of: "@", with: ""))
        let user = mentionedUsers?.first(where: { $0.name == string })
        onMentionedUserTap?(user)
    }
}
```

Then, in the `ComposerVC` we simply handle the closure of `TextViewMentionedUsersHandler`, and nothing else. In the `ChatMessageListVC` we handle the closure and forward it to the `ChatMessageContentViewDelegate` just to make it easier on the message list. With this, we can handle interaction in both components, and on the message list, we follow the current pattern to handle message content view events.

Let me know if you have any questions.

Sources/StreamChatUI/Composer/ComposerVC.swift
```diff
@@ -909,6 +922,12 @@ open class ComposerVC: _ViewController,
         guard let maxMessageLength = channelConfig?.maxMessageLength else { return true }
         return textView.text.count + (text.count - range.length) <= maxMessageLength
     }
+    
+    open func textView(_ textView: UITextView, shouldInteractWith URL: URL, in characterRange: NSRange, interaction: UITextItemInteraction) -> Bool {
+        let detector = TextViewMentionedUserDetector()
```
nuno-vieira, line 927: This should be a reference property so that it can be easy replaced by the customer and in tests.
```suggestion

```

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
```diff
@@ -645,6 +669,29 @@ open class ChatMessageContentView: _View, ThemeProvider {
...
+    // MARK: - UITextViewDelegate
+    
+    open func textView(_ textView: UITextView, shouldInteractWith URL: URL, in characterRange: NSRange, interaction: UITextItemInteraction) -> Bool {
+        if let mentionedUsers = content?.mentionedUsers, !mentionedUsers.isEmpty {
+            let tappedMentionedUser = textViewUserMentionsHandler.mentionedUserTapped(
+                on: textView,
+                in: characterRange,
+                withMentionedUsers: mentionedUsers
+            )
+            tappedMentionedUser.map {
```
martinmitrevski, line 682: Not sure what exactly happens here. Map is used for transforming arrays of elements, and here we're using it to call a delegate method. Could you provide more details why is this the case?

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -547,6 +547,12 @@ open class ChatMessageListVC: _ViewController,
         )
     }
 
+    /// Gets called when mentioned user is tapped.
+    /// - Parameter mentionedUser: The mentioned user that was tapped on.
+    open func messageContentViewDidTapOnMentionedUser(_ mentionedUser: ChatUser?) {
```
martinmitrevski, line 552: Do we need an optional `ChatUser` here? We shouldn't call the method if there's no user available.

Sources/StreamChatUI/Utils/TextViewMentionedUsersHandler.swift
```diff
@@ -0,0 +1,23 @@
+//
+// Copyright © 2022 Stream.io Inc. All rights reserved.
+//
+
+import StreamChat
+import UIKit
+
+open class TextViewMentionedUsersHandler {
```
martinmitrevski, line 8: The class is open, we need docs for it and the method.

martinmitrevski (commented):
Looks good in general. What's missing before merging this:
- feature flag: off by default, on in the demo app (as already discussed)
- docusaurus docs
- few PR comments

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
nuno-vieira: Map here is being to unwrap the optional. But I agree this can be confusing to read, so I will change it 👍 The advantage is that it makes the code a bit more concise, but indeed not very readable

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
nuno-vieira: Yes this is wrong, updating 👍 Missed this from the previous implementation

Sources/StreamChatUI/Utils/TextViewMentionedUsersHandler.swift
nuno-vieira: Added docs already ✅

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
nuno-vieira: Fixed it ✅

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
nuno-vieira: Done ✅

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
```diff
@@ -645,6 +669,24 @@ open class ChatMessageContentView: _View, ThemeProvider {
     @objc open func handleTapOnDeliveryStatusView() {
         delegate?.messageContentViewDidTapOnDeliveryStatusIndicator(indexPath?())
     }
+    
+    // MARK: - UITextViewDelegate
+    
+    open func textView(_ textView: UITextView, shouldInteractWith URL: URL, in characterRange: NSRange, interaction: UITextItemInteraction) -> Bool {
+        if let mentionedUsers = content?.mentionedUsers, !mentionedUsers.isEmpty {
+            let tappedMentionedUser = textViewUserMentionsHandler.mentionedUserTapped(
```
polqf, line 677:
```suggestion
            let mentionedUser = textViewUserMentionsHandler.mentionedUserTapped(
```

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
```diff
@@ -645,6 +669,24 @@ open class ChatMessageContentView: _View, ThemeProvider {
     @objc open func handleTapOnDeliveryStatusView() {
         delegate?.messageContentViewDidTapOnDeliveryStatusIndicator(indexPath?())
     }
+    
+    // MARK: - UITextViewDelegate
+    
+    open func textView(_ textView: UITextView, shouldInteractWith URL: URL, in characterRange: NSRange, interaction: UITextItemInteraction) -> Bool {
+        if let mentionedUsers = content?.mentionedUsers, !mentionedUsers.isEmpty {
```
polqf, line 676: guard?

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
```diff
@@ -645,6 +669,24 @@ open class ChatMessageContentView: _View, ThemeProvider {
...
+            let tappedMentionedUser = textViewUserMentionsHandler.mentionedUserTapped(
+                on: textView,
+                in: characterRange,
+                withMentionedUsers: mentionedUsers
```
polqf, line 680: Doesn't look to swifty

Sources/StreamChatUI/Composer/TypingSuggester.swift
```diff
@@ -100,17 +100,10 @@ public struct TypingSuggester {
             return nil
         }
 
-        // Fetch the suggestion text. The suggestions can't have spaces.
```
polqf, line 103: Why do we allow spaces now?

Sources/StreamChatUI/Utils/TextViewMentionedUsersHandler.swift
```diff
@@ -0,0 +1,30 @@
...
+    ///   - mentionedUsers: The current mentioned users in the message.
+    /// - Returns: The `ChatUser` in case it tapped a mentioned user.
+    open func mentionedUserTapped(
+        on textView: UITextView,
+        in characterRange: NSRange,
+        withMentionedUsers mentionedUsers: Set<ChatUser>
+    ) -> ChatUser? {
+        guard let text = textView.text,
+              let range = Range(characterRange, in: text)
+        else {
```
polqf, line 23: Weird formatting

Sources/StreamChatUI/Utils/TextViewMentionedUsersHandler.swift
```diff
@@ -0,0 +1,30 @@
...
+            return nil
+        }
+        let name = String(text[range].replacingOccurrences(of: "@", with: ""))
+        let user = mentionedUsers.first(where: { $0.name == name })
+        return user
```
polqf, line 28:
```suggestion
        return mentionedUsers.first(where: { $0.name == name })
```

Sources/StreamChatUI/Utils/UITextView+Extensions.swift
```diff
@@ -5,31 +5,17 @@
...
-        textStorage.addLayoutManager(layoutManager)
-        if let font = font {
-            textStorage.addAttribute(.font, value: font, range: .init(0..<textStorage.length))
+    func highlightMention(mention: String) {
+        let attributeText = NSMutableAttributedString(attributedString: attributedText)
+        let string = attributeText.string
+        guard let range = string
+            .range(of: mention, options: .caseInsensitive)
+            .map({ NSRange($0, in: string) })
+        else {
```
polqf, line 14: weird formatting again

polqf (commented):
Some of the processes performed to highlight can be really costly. Would be nice to disable all of it if the client is not using mentions. Do we have any channel config?

martinmitrevski (approved):

Sources/StreamChatUI/Composer/TypingSuggester.swift
nuno-vieira: It is a bug. If you typed "@" in a new line, it wouldn't show the suggestions

nuno-vieira: It looks this was wrong before, user names can have spaces. The other platforms also allow spaces in user suggestions

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
nuno-vieira: Feel free to suggest better naming

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
nuno-vieira: The normal branch of this logic is to actually open a URL, if we in the future introduce more logic here, it will be easy to introduce at the bottom. The early return pattern here doesn't make much sense to me IMO

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
nuno-vieira: This will clash with the property below. The reason we don't use the if let here, is to make the code easier to read

Sources/StreamChatUI/Utils/TextViewMentionedUsersHandler.swift
nuno-vieira: It looks fine to me, if we want to be more strict here, we should introduce rules in SwiftFormat, to avoid wasting time with formatting comments on PR Reviews

Sources/StreamChatUI/Utils/UITextView+Extensions.swift
nuno-vieira: It is normal to pipe functions at the bottom in a new line in Swift. Just like we do in the E2E Tests. 

But same as above, if we want to be more strict here, we should introduce rules in SwiftFormat

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
polqf: `with mentionedUsers: [ChatUser]` would be enough

nuno-vieira: ✅

docusaurus/docs/iOS/uikit/navigation.md
```diff
@@ -0,0 +1,43 @@
...
+
+The StreamChatUI SDK not only provides reusable UI Components, but also it comes with navigation flows out of the box, that you can customize.
+
+## Routers
+
+The navigation of StreamChatUI SDK is handled by `NavigationRouter`'s. Currently, the available routers in the SDK are the following:
+
+- `ChatChannelListRouter`: The Channel List Router is responsible to handle the navigation of the Channel List, like showing the Channel View.
+- `ChatMessageListRouter`: The Message List Router is responsible to handle the navigation of the Message List, like showing the User Profile View, the Gallery View, the Message Actions Popup, etc.
+- `AlertsRouter`: The Alerts Router is responsible to show alerts in the Chat SDK, like confirmation dialogs.
```
nuno-vieira, line 13:
```suggestion
- `ChatChannelListRouter`: The Channel List Router is responsible to handle the navigation of the Channel List, like showing the Channel View.
- `ChatMessageListRouter`: The Message List Router is responsible to handle the navigation of the Message List, like showing the User Profile View, the Gallery View, the Message Actions Popup, etc.
- `AlertsRouter`: The Alerts Router is responsible to show alerts in the Chat SDK, like confirmation dialogues.
```

docusaurus/docs/iOS/uikit/navigation.md
```diff
@@ -0,0 +1,43 @@
...
+
+### Presenting a custom user profile view
+
+Currently, the StreamChatUI SDK does not come with a user profile view out of the box, since most apps already have their own user profile view. If you want to show your user profile view whenever the the user clicks on an avatar or clicks on a user mention, you can do so my customizing the `ChatMessageListRouter`.
```
nuno-vieira, line 17:
```suggestion
Currently, the StreamChatUI SDK does not come with a user profile view out of the box, since most apps already have their own user profile view. If you want to show your user profile view whenever the user clicks on an avatar or clicks on a user mention, you can do so my customizing the `ChatMessageListRouter`.
```

nuno-vieira (commented):
Fix doc typos

nuno-vieira (approved):

nuno-vieira:
> Some of the processes performed to highlight can be really costly. Would be nice to disable all of it if the client is not using mentions. Do we have any channel config?

Right now, there is no such flag. I believe this the case because it is like a core feature of the SDK 🤔

Sources/StreamChatUI/Navigation/ChatMessageListRouter.swift
```diff
@@ -191,6 +191,17 @@ open class ChatMessageListRouter:
         rootViewController.present(galleryVC, animated: true)
     }
 
+    /// Shows the user profile view.
+    /// - Parameter user: The user profile to show.
+    open func showUser(_ user: ChatUser) {
+        log.info(
+            """
+            Performed action to show user profile view of \(user.name ?? user.id).
+            To provide your own profile view, please override ChatMessageRouter.showUser()"
```
nuno-vieira, line 200:
```suggestion
            To provide your own profile view, please override ChatMessageListRouter.showUser()"
```

martinmitrevski (approved):

PR #2258, opened 2022-08-30
[CIS-2150, CIS-2115 ] Fix broken pagination when quoting or pinning old messages

polqf:
### 🔗 Issue Links

https://stream-io.atlassian.net/browse/CIS-2150
https://stream-io.atlassian.net/browse/CIS-2115

### 🎯 Goal

This PR fixes a case where quoting or pinning old messages can lead to broken pagination

### 📝 Summary

When we fetch a page, it can contain pinned and/or quoted messages. Those are also saved in the database.
Because those messages meet the current criteria of our DB listener, they appear in the channel.

If there is a case where the quoted/pinned message is old, we were using it as the last id to fetch the following page.
That meant that we were leaving a gap in some cases.

### 🛠 Implementation

This PR addresses the above-mentioned issue by keeping track of the last *fetched* message id. This id will be the one used, instead of the last visible one.

### ⚠️ Caveat:
When scrolling really fast, you might see the old messages at the top while waiting for the page to be loaded

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
+++ b/Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
@@ -232,6 +232,7 @@ public class ChatChannelController: DataController, DelegateCallable, DataStoreP
     )
     
     private var markingRead: Bool = false
+    private var lastFetchedMessageId: MessageId?
 
     /// A type-erased delegate.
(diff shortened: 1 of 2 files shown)


martinmitrevski (approved):
LGTM ✅

nuno-vieira (approved):
LGTM! ✅

Issue #2251, opened 2022-08-25
notification.removed_from_channel fails to decode event from event payload


AdamAexol:
## What did you do?
I removed user from channel when that user was disconnected and logged out of the app.

## What did you expect to happen?
After connecting that user again stream-chat should get event about removal from that channel and that channel should also  disappear from that user's channel list.

## What happened instead?
Stream-chat receives event about removal from the channel, but cannot handle that event. It fails on decoding event object
```
2022-08-23 15:39:47.646 [INFO] [com.apple.NSURLSession-delegate] [SyncRepository.swift:243] [syncMissingEvents(using:channelIds:isRecoveryRequest:completion:)] > Processing pending events. Count 1
2022-08-23 15:39:47.654 [ERROR] [com.apple.NSURLSession-delegate] [EventPayload.swift:176] [asEvents()] > Failed to decode event from event payload: 
-----EventPayload-----
     eventType: EventType(rawValue: "notification.removed_from_channel")
     connectionId: nil
     cid: Optional(group_chat:fZiaIB4w_group)
     currentUser: nil
     user: Optional(StreamChat.UserPayload)
     createdBy: nil
     memberContainer: Optional(StreamChat.MemberContainerPayload(member: nil, invite: nil, memberRole: nil))
     channel: nil
     message: nil
     reaction: nil
     watcherCount: nil
     unreadCount: nil
     createdAt: Optional(2022-08-23 13:30:32 +0000)
     isChannelHistoryCleared: nil
     banReason: nil
     banExpiredAt: nil
     parentId: nil
     hardDelete: false
, error: Error EventDecoding in /Users/adamgrabowski/Projects/ios-parlorsocial/ParlorSocial/Pods/StreamChat/Sources/StreamChat/WebSocketClient/Events/EventPayload.swift:160 -> `Swift.KeyPath<StreamChat.EventPayload, Swift.Optional<StreamChat.MemberPayload>>` field can't be `nil` for the `notification.removed_from_channel` event.
2022-08-23 15:39:47.668 [INFO] [NSManagedObjectContext 0x2827852b0] [SyncRepository.swift:277] [processMissingEventsPayload(_:completion:)] > Successfully processed pending events. Count 1
```
Probably member field in memberContainer is required and in this event it's nil.
Also after launching app again, above message is still in logs, as if this event was hanging on the pending events list.
[...]

PR #2259, opened 2022-08-30
Fix parsing `member` field in `notification.removed_from_channel` event

nuno-vieira:
### 🔗 Issue Links
https://github.com/GetStream/stream-chat-swift/issues/2251

### 🎯 Goal
In the `notification.removed_from_channel` event, the `user` field is not present inside the `member` field, although the `user_id` is. This PR adds a fallback so that if the `user` is not present, we use the id from the `user_id` property. 

The backend will change this meanwhile, and make sure the `user` field is always present. But since we don't have an ETA for this, it makes sense to add this fallback.

### ☑️ Contributor Checklist


martinmitrevski (approved):
LGTM ✅

polqf (commented):
The only comment I would add is that we should be using optionals instead of unwrapped values in tests

nuno-vieira:
@polqf That will take some time to change. Besides that, most of the tests are performing against optional values, so if we just use the optional value on the actuals, there's a chance of having false positives, if both are nil

Issue #2284, opened 2022-09-10
Unread count gets reset after sending a message even though the message list screen was dismissed


matthewjsloane:
## What did you do?
I opened the message list for a channel and then dismissed it. From another user on a different device, I sent a message to that channel.

## What did you expect to happen?
I expected the unread count to go from 0 to 1.

## What happened instead?
The unread count went from 0 to 1, but then the channel was marked as read afterwards, causing it to reset to 0. Subsequent messages started incrementing the unread count again as expected.

## GetStream Environment
**GetStream Chat version:** 4.21.1
**GetStream Chat frameworks:** StreamChat, StreamChatUI
**iOS version:** 15
**Swift version:** 5
**Xcode version:** 13.4.1
**Device:** iPhone 13 Pro Max

## Additional context
It's almost as if the message channel is "staying open" longer than it should, and causing messages to be marked as read even when the view is no longer on the screen. Possibly some sort of race condition? Let me know if you need any additional details.

nuno-vieira:
Hi @matthewjsloane,

We were able to reproduce this in our demo app. We will investigate it and will let you know once we have more details.

Thank you for the report!

Best,
Nuno

nuno-vieira:
Hi!

The issue has been fixed in the 4.22.0 release! 🚀 

Thank you for your patient!

Best,
Nuno

matthewjsloane:
Thanks for the update and your quick support!

PR #2288, opened 2022-09-12
[CIS-2168] Do not mark channels as read when the controller is not on screen

polqf:
### 🔗 Issue Links

https://stream-io.atlassian.net/browse/CIS-2168
https://github.com/GetStream/stream-chat-swift/issues/2284

### 🎯 Goal

Do not mark channels as read when the controller is not in the stack

### 🛠 Implementation

There was a temporary leak during the execution of ChatChannelVC.synchronize that was keeping the ChatController alive. That was translated into the controller marking the channel as read even if the chat was not visible on screen.

### 🧪 Manual Testing Notes

1. Open a channel
2. Go back to the channel list
3. From another device, send a message
4. Wait until the unread count show as 1 in the list

Previous result:
The unread count was showing 1 for a second, and then going back to 0

Expected result:
The unread count should stay as 1

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## StreamChat
 ### 🐞 Fixed
 - Fixed pagination in message list not working when synchronize does not succeed [#2241](https://github.com/GetStream/stream-chat-swift/pull/2241)
+- Do not mark channels as read when the controller is not on screen [#2288](https://github.com/GetStream/stream-chat-swift/pull/2288)
 
 ## StreamChatUI
 ### ✅ Added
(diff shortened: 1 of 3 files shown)


martinmitrevski (approved):

Sources/StreamChat/Database/DTOs/ChannelReadDTO.swift
```diff
@@ -85,6 +85,8 @@ extension NSManagedObjectContext {
     }
     
     func markChannelAsRead(cid: ChannelId, userId: UserId, at: Date) {
+        guard currentUser?.user.id == userId else { return }
```
nuno-vieira, line 88: Worth adding a test to avoid regressions 👍

polqf: Yep, still in draft state because many things are missing 😄

nuno-vieira (approved):
Don't forget the Changelog 👍 Other than that, LGTM! ✅

PR #2292, opened 2022-09-13
[BW] Turn off LLC tests on Xcode 12.5.1 due to a syntax error

alteral:


polqf:
Maybe it is time to start moving things to Xcode 13, and move the main flows to Xcode 14

polqf (approved):

PR #2294, opened 2022-09-14
[BW] Don't register for push notifications in Test App on iOS <14

alteral:
### 🔗 Issue Links

- N/A

### 🎯 Goal

- To fix a crash in UI tests on iOS 12 that occurs due to a push notification alert


nuno-vieira (changes requested):
LGTM! ✅

nuno-vieira (approved):

PR #2307, opened 2022-09-20
[CIS-2177] Add support for Thread List pagination to Mock Server

alteral:
### 🔗 Issue Links

- https://stream-io.atlassian.net/browse/CIS-2177
- https://stream-io.atlassian.net/browse/CIS-1583

### 🎯 Goal

- Add support for pagination in Thread List
- Automate related test-cases

### 📝 Summary

- From now on the mock server supports pagination in Thread List

### 🧪 Manual Testing Notes

1. Generate replies using the mock server, e.g.:
```swift
backendRobot.generateChannels(count: 1, messagesCount: 1, replyCount: 60)
```
2. Freeze the test app:
```swift
sleep(10000)
```
3. Manually Open a Thread List and load the history
4. Observe pagination requests

### ☑️ Contributor Checklist

### 🎁 Meme

gif


TestTools/StreamChatTestMockServer/MockServer/MessageList.swift
```diff
@@ -40,7 +51,7 @@ public extension StreamMockServer {
     
     func findMessagesByParrentId(_ parentId: String) -> [[String: Any]] {
```
nuno-vieira, line 52:
```suggestion
    func findMessagesByParentId(_ parentId: String) -> [[String: Any]] {
```

TestTools/StreamChatTestMockServer/MockServer/ChannelResponses.swift
```diff
@@ -338,19 +339,29 @@ public extension StreamMockServer {
...
-                    messages.append(message)
-                    saveMessage(message)
+                    messages.append(newMessage)
+                    
+                    if replyCount > 0 {
+                        for replyIndex in 1...replyCount {
+                            generateMessage(
+                                withIndex: replyIndex,
+                                withId: TestData.uniqueId,
+                                inThread: true,
```
nuno-vieira, line 358: Instead of inThread: true, why don't we use the parentId, wouldn't that simplify the `saveReply` function? 
```suggestion
                                parentId: newMessage.id,
```

StreamChatUITestsAppUITests/Tests/MessageList_Tests.swift
```diff
@@ -501,6 +501,26 @@ extension MessageList_Tests {
...
+    func test_paginationOnThread() throws {
+        linkToScenario(withId: 55)
+        
+        try XCTSkipIf(ProcessInfo().operatingSystemVersion.majorVersion == 12,
+                      "[CIS-2020] Scroll on message list does not work well enough")
+        
+        let replyCount = 60
+        
+        GIVEN("user opens the channel") {
+            backendRobot.generateChannels(count: 1, messagesCount: 1, replyCount: replyCount)
```
nuno-vieira, line 513: My only concern with this function is that it is not very flexible, you can't decide which messages have replies, and which don't. For these use case, it is fine now, but if we want to add more complex scenarios it will start to be difficult to use this function 🤔 Maybe creating a tech debt ticket would be nice for this

TestTools/StreamChatTestMockServer/MockServer/MessageList.swift
```diff
@@ -18,6 +18,17 @@ public extension StreamMockServer {
...
+    func saveReply(_ message: [String: Any]?) {
+        guard let newMessage = message else { return }
+        if let index = threadList.firstIndex(where: { (message) -> Bool in
+            (newMessage[messageKey.id.rawValue] as? String) == (message[messageKey.id.rawValue] as? String)
+        }) {
+            threadList[index] = newMessage
+        } else {
+            threadList.append(newMessage)
+        }
+    }
```
nuno-vieira, line 30: This function looks a bit confusing and hard to read, would my suggestion of putting the parent Id of the message, make it easier?

TestTools/StreamChatTestMockServer/MockServer/StreamMockServer.swift
```diff
@@ -16,6 +16,7 @@ public final class StreamMockServer {
     public private(set) var server: HttpServer = HttpServer()
     private weak var globalSession: WebSocketSession?
     private var channelConfigs = ChannelConfigs()
+    public var threadList: [[String: Any]] = []
```
nuno-vieira, line 19: Does this mean we can only have one thread list? 🤔 What is the thread list dictionary used for? Couldn't each message in the `messageList` have a key-value of `replies: [[String: Any]]` ?

TestTools/StreamChatTestMockServer/MockServer/MessageResponses.swift
```diff
@@ -534,9 +537,61 @@ public extension StreamMockServer {
         return .ok(.json(json))
     }
     
-    private func mockMessageReplies(_ messageId: String) -> HttpResponse {
+    private func limitReplies(_ request: HttpRequest) throws -> HttpResponse {
```
nuno-vieira, line 540: I still think it makes sense to name this mock replies, instead of limitReplies, the limit pagination is an implementation detail in this case 
```suggestion
    private func mockMessageReplies(_ request: HttpRequest) throws -> HttpResponse {
```

TestTools/StreamChatTestMockServer/MockServer/MessageResponses.swift
```diff
@@ -534,9 +537,61 @@ public extension StreamMockServer {
...
+            }
+            if let messageIndex = messageIndex {
+                let messageCount = messages.count - 1
+                let plusLimit = messageIndex + limit
+                let endWith = plusLimit < messageCount ? plusLimit - 1 : messageCount
+                messages = Array(messages[messageIndex...endWith])
+            }
+        } else {
+            messages = Array(messages.suffix(limit))
+        }
```
nuno-vieira, line 593: I can see we also have a `limitQuery` function that does exactly this implementation 🤔 

Ideally, we should also change the `limitQuery` to `mockMessages()` which is more intuitive. And then, we should try to extract this logic to a function and reuse it in both places. Something like `fakeMessagePagination()`

WDYT?

nuno-vieira (commented):
Overall looks good, but I think we can improve it to make it more maintainable in the future 🙂

TestTools/StreamChatTestMockServer/MockServer/MessageResponses.swift
alteral: Done 👍

TestTools/StreamChatTestMockServer/MockServer/MessageList.swift
alteral: Hmm, not sure I got you. I did update it a little bit (as well as `saveMessage()` as they are similar). Could you please check it out?

StreamChatUITestsAppUITests/Tests/MessageList_Tests.swift
alteral: Yeah, it is a bit tricky func, makes sense to create a separate ticket for improving it.

TestTools/StreamChatTestMockServer/MockServer/ChannelResponses.swift
alteral: Nice catch, done 👍

TestTools/StreamChatTestMockServer/MockServer/MessageList.swift
alteral: Done 👍

TestTools/StreamChatTestMockServer/MockServer/StreamMockServer.swift
alteral: Tbh, this `threadList` var was introduced only for the pagination scenario and it's used only when you'd like to generate loads of thread replies prior to the test. By default all replies live inside `messageList`, this just does not work with pagination, because it loads them straight away.

I could add a comment to this var, but not sure it makes sense to make `messageList` more complex just because of one scenario with pagination in the thread. WDYT?

TestTools/StreamChatTestMockServer/MockServer/MessageResponses.swift
alteral: Done, check it out 🤞

TestTools/StreamChatTestMockServer/MockServer/StreamMockServer.swift
nuno-vieira: Makes sense then 👍

nuno-vieira (approved):
LGTM! 🚀

PR #2314, opened 2022-09-22
[CIS-2184] Ignore channel.created events

polqf:
### 🔗 Issue Links

https://stream-io.atlassian.net/browse/CIS-2184

### 🎯 Goal

Stop logging error when 'channel.created' events are received

### 📝 Summary

We are already handling new channels when ‘notification.added_to_channel' are received.
’channel.created' events are only received through the sync endpoint, but we were logging an error as it was unhandled, but it seemed like there was a bug.

### 🛠 Implementation

When receiving 'channel.created', we are throwing `ClientError.UnsupportedEventType`, which is then only logged as info

### 🧪 Manual Testing Notes

1. Logged with user A, and the app closed
2. With user B, create a group including user A
3. Open the app with user A

Expected result:
No error logs regarding 'channel.created' should appear

Previous result:
An error log would be logged for 'channel.created'

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/WebSocketClient/Events/EventType.swift
+++ b/Sources/StreamChat/WebSocketClient/Events/EventType.swift
@@ -42,6 +42,8 @@ public extension EventType {
     // MARK: Channel Events
     
     /// When a channel was updated.
+    static let channelCreated: Self = "channel.created"
+    /// When a channel was updated.
     static let channelUpdated: Self = "channel.updated"
     /// When a channel was deleted.
     static let channelDeleted: Self = "channel.deleted"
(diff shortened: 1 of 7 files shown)


nuno-vieira (approved):
LGTM! ✅

Sources/StreamChat/WebSocketClient/Events/EventType.swift
```diff
@@ -120,6 +122,7 @@ extension EventType {
         case .userBanned: return try UserBannedEventDTO(from: response)
         case .userUnbanned: return try UserUnbannedEventDTO(from: response)
         
+        case .channelCreated: throw ClientError.UnsupportedEventType()
```
tbarbugli, line 125: @polqf bit of weird/confusing approach, isn't better to parse the event correctly and simply do nothing about it when it's received?

tbarbugli: marking "channel.created" as an unsupported event is the strange bit, totally fine to do nothing when such event is received though

polqf: The idea is not to have to maintain any kind of code for an event we don't even need. Otherwise both an Event and an EventDTO is required as per our contract.

Maybe the naming is what can be changed here. Might rename it to IgnoredEventType

PR #2322, opened 2022-10-04
[BW] Fix flakiness in the message delivery status tests

alteral:
### 🔗 Issue Links

- N/A

### ☑️ Contributor Checklist

### DIFF
--- a/StreamChatUITestsAppUITests/Tests/Message Delivery Status/MessageDeliveryStatus+ChannelList_Tests.swift
+++ b/StreamChatUITestsAppUITests/Tests/Message Delivery Status/MessageDeliveryStatus+ChannelList_Tests.swift
@@ -102,7 +102,7 @@ final class MessageDeliveryStatus_ChannelList_Tests: StreamTestCase {
             userRobot.tapOnBackButton()
         }
         WHEN("participant reads the user's message") {
-            participantRobot.wait(0.3).readMessage()
+            participantRobot.readMessage(after: 0.3)
         }
         THEN("user spots double checkmark next to the message") {
             userRobot.assertMessageDeliveryStatusInChannelPreview(.read)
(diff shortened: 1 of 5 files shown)


StreamChatUITestsAppUITests/Tests/Message Delivery Status/MessageDeliveryStatus+ChannelList_Tests.swift
```diff
@@ -102,7 +102,7 @@ final class MessageDeliveryStatus_ChannelList_Tests: StreamTestCase {
             userRobot.tapOnBackButton()
         }
         WHEN("participant reads the user's message") {
-            participantRobot.readMessage()
+            participantRobot.wait(0.3).readMessage()
```
martinmitrevski, line 105: Wouldn't it be better if we make a new `readMessage(after seconds: Double = 0.3)` to avoid duplicating the same logic?

martinmitrevski (commented):
Looks good, I would just create a new method for reading message with delay (see comment).

StreamChatUITestsAppUITests/Tests/Message Delivery Status/MessageDeliveryStatus+ChannelList_Tests.swift
polqf: +1

alteral: Sure thing, done 👍

martinmitrevski: since it already has a default value, might be good to skip passing the 0.3 everywhere. If it collides with the other `readMessage` method, might be good to rename this one.

alteral: If we skip passing the 0.3 everywhere then it will use the function without `wait`. I removed the default value, does it look better?

martinmitrevski (approved):
✅

PR #2347, opened 2022-10-24
[CIS-2113] Allow Message Search pagination with sorting option

polqf:
### 🔗 Issue Links

https://stream-io.atlassian.net/browse/CIS-2113

### 🎯 Goal

Allow pagination when using search in combination with sort parameters

### 🛠 Implementation

When using sort parameters, we use the `next` value that backend returns us to perform the subsequent API call to load more messages

### ☑️ Contributor Checklist


martinmitrevski (approved):
LGTM ✅ 
We just need to see why the checks are failing + increase the coverage.

Sources/StreamChat/Query/MessageSearchQuery.swift
```diff
@@ -86,7 +109,7 @@ public struct MessageSearchQuery: Encodable {
     
     public let sort: [Sorting<MessageSearchSortingKey>]
     
-    public var pagination: Pagination?
+    public var pagination: SearchPagination?
```
nuno-vieira, line 112: Isn't this a breaking change? 🤔 Can't we add a new property `cursor` to `Pagination` to avoid the breaking change? And then `Pagination` would only have 2 constructors, one where we use the cursor, and one where use Offset and limit.

nuno-vieira (commented):
Overall LGTM 👍 Just worried about the breaking change + would be nice to have a way to manually QA this. Maybe changing the ChatSampleApp? I think we are using the message search there 🤔 If we add a sort parameter, then we can manually test this

Sources/StreamChat/Query/MessageSearchQuery.swift
polqf: Good catch!

martinmitrevski (approved):

Tests/StreamChatTests/Controllers/SearchControllers/MessageSearchController/MessageSearchController_Tests.swift
```diff
@@ -605,14 +630,36 @@ final class MessageSearchController_Tests: XCTestCase {
...
+        try client.databaseContainer.createMessage(id: messageId, searchQuery: controller.query)
+
+        // Simulate update call response
+        env.messageUpdater?.search_completion?(.success(MessageSearchResultsPayload(results: [], next: responseNextCursor)))
+
+        let message = try XCTUnwrap(client.databaseContainer.viewContext.message(id: messageId)?.asModel())
+
+        AssertAsync.willBeEqual(controller.messages.count, 1)
+        return message
+    }
```
nuno-vieira, line 662: Can we move this to the bottom of the file?

nuno-vieira (approved):
LGTM! ✅ Just added a small nit. And let's make sure CI is 🟢

Issue #2305, opened 2022-09-19
Allow developers to filter message list


virbedi:
## What are you trying to achieve?
We are trying to hide/show threaded replies sent to the main channel based on a boolean flag. 

## If possible, how can you achieve this currently?
We are checking the content of each message and passing empty layoutOptions for ones we should hide. 
This affects scroll performance and does not always provide clean hide/show views of the channel.

## What would be the better way?
Making `ChatMessageListVC` > `dataSource` open so that we can override it or allow the ability for users to perform a query on the message list in order to decide which messages are rendered.

## GetStream Environment
**GetStream Chat version:**
4.15
**GetStream Chat frameworks:** StreamChat, StreamChatUI
**iOS version:**
16
**Swift version:**
5
**Xcode version:**
13.3
**Device:**
iPhone 13 Pro Max

## Additional context
This request was initially opened as a support ticket - 27570

polqf:
Hi @virbedi ,

As we discussed in a support ticket, this is not a functionality we currently have in the SDK.
We will open a feature request and will prioritize it accordingly.

Will keep you updated

nuno-vieira:
Hi @virbedi,

You can try overriding the following methods from `ChatChannelVC`: 

<img width="1004" alt="image" src="https://user-images.githubusercontent.com/12814114/191239240-e6bba4ed-eb8f-4cdf-91d0-5cf95405be34.png">

Your suggestion is already possible. The `ChatMessageListVCDataSource` is already open, but you need to override it in the `ChatChannelVC`, not `ChatMessageListVC`. 

For your use case, we are not sure what could be the consequences of overriding this, so keep that in mind. 

Let us know if this worked for you.

Best,
Nuno

virbedi:
Hello @nuno-vieira,
Overriding these functions seem to lead to frequent crashes related to mismatched indexes. 

```
*** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'attempt to delete row 124 from section 0 which only contains 48 rows before the update'
terminating with uncaught exception of type NSException
CoreSimulator 802.6 - Device: iPhone 13 (E33F5BC3-4E30-40F1-B369-6FDF712889AC) - Runtime: iOS 15.4 (19E240) - DeviceType: iPhone 13
```

nuno-vieira:
Hi @virbedi!

Yes, that could be happening. Can you share how are you overriding all the methods above?

Best,
Nuno

[...]

PR #2348, opened 2022-10-24
[CIS-2179] Add support for bypassing the messages and replies data source

nuno-vieira:
### 🔗 Issue Links
CIS-2179
https://github.com/GetStream/stream-chat-swift/issues/2305

### 🎯 Goal
Add support for bypassing the messages and replies data source.

### 🛠 Implementation
Provide a `Components.messagesBypass` and `Components.repliesBypass` to be able to override the message list data source. It is also possible to change directly inside the `ChatThreadVC` and `ChatChannelVC`.

TODO:
- [ ] Add DemoApp example to test this
- [ ] Finish test coverage
- [ ] Refactor and reuse some logic between Thread and Channel

Limitations:
- The pagination number won't be respected. If the pagination is 25 by 25, for example, since the filter happens after fetching from the DB. If the bypass removes 2 messages, it will be loading 23, and not 25. Either way, the same thing happens already when hiding deleted messages on the client side.

### 🧪 Manual Testing Notes
TODO: Add an example in the DemoApp

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChatUI/ChatThread/ChatThreadVC.swift
+++ b/Sources/StreamChatUI/ChatThread/ChatThreadVC.swift
@@ -53,12 +53,17 @@ open class ChatThreadVC: _ViewController,
 
     public var messageComposerBottomConstraint: NSLayoutConstraint?
 
+    /// A closure to filter replies and override the message list data source.
+    public var repliesFilter: ((ChatMessage) -> Bool)?
+
     private var isLoadingPreviousMessages: Bool = false
 
     private var currentlyTypingUsers: Set<ChatUser> = []
 
     override open func setUp() {
         super.setUp()
+
+        repliesFilter = components.repliesFilter
 
         let notificationCenter = NotificationCenter.default
         notificationCenter.addObserver(
@@ -150,12 +155,17 @@ open class ChatThreadVC: _ViewController,
 
     // MARK: - ChatMessageListVCDataSource
 
+    // The replies data source.
     public var messages: [ChatMessage] {
+        set {
+            if let repliesFilter = repliesFilter {
+                replies = newValue.filter(repliesFilter)
+                return
+            }
+            replies = newValue
+        }
         get {
             replies
-        }
-        set {
-            replies = newValue
         }
     }
 
@@ -321,8 +331,16 @@ open class ChatThreadVC: _ViewController,
 
     private func updateMessages(with changes: [ListChange<ChatMessage>]) {
         messageListVC.setPreviousMessagesSnapshot(self.messages)
+
         let messages = getRepliesWithThreadRootMessage(from: messageController)
-        messageListVC.setNewMessagesSnapshot(messages)
+        let newMessages: [ChatMessage]
+        if let repliesFilter = repliesFilter {
+            newMessages = messages.filter(repliesFilter)
+        } else {
+            newMessages = messages
+        }
+
+        messageListVC.setNewMessagesSnapshot(newMessages)
         messageListVC.updateMessages(with: changes)
     }
 
--- a/Sources/StreamChatUI/ChatThread/ChatThreadVC.swift
+++ b/Sources/StreamChatUI/ChatThread/ChatThreadVC.swift
@@ -53,8 +53,8 @@ open class ChatThreadVC: _ViewController,
 
     public var messageComposerBottomConstraint: NSLayoutConstraint?
 
-    /// A closure to filter replies and override the message list data source.
-    public var repliesFilter: ((ChatMessage) -> Bool)?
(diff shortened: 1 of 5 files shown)


martinmitrevski (commented):
Looks good 👍 
My two cents on: "Another option would be to provide a ([ChatMessage] -> [ChatMessages]) closure instead of ([ChatMessage] -> Bool) to allow for even more flexibility" - I would go with this option, since it allows more flexibility. Apart from adding more messages, you could also check dependencies between messages.
Additionally, we would need changelog update and docusaurus docs.

polqf:
Not really a fan of having filtering logic duplicated in each file. Is there any way to abstract it?

nuno-vieira:
> Not really a fan of having filtering logic duplicated in each file. Is there any way to abstract it?

What do you mean by each file? 🤔

Sources/StreamChatUI/ChatThread/ChatThreadVC.swift
```diff
@@ -321,8 +331,16 @@ open class ChatThreadVC: _ViewController,
...
-        messageListVC.setNewMessagesSnapshot(messages)
+        let newMessages: [ChatMessage]
+        if let repliesBypass = repliesBypass {
+            newMessages = repliesBypass(messages)
+        } else {
+            newMessages = messages
+        }
+
+        messageListVC.setNewMessagesSnapshot(newMessages)
         messageListVC.updateMessages(with: changes)
```
nuno-vieira, line 344: This is a problem. We are not filtering the `changes` and use it for some logic inside the `messageListVC` 🤔

nuno-vieira: This is now fixed 🙂

Sources/StreamChatUI/ChatThread/ChatThreadVC.swift
```diff
@@ -320,10 +330,22 @@ open class ChatThreadVC: _ViewController,
...
+        let newMessages: [ChatMessage]
+        let newChanges: [ListChange<ChatMessage>]
+        if let repliesBypass = repliesBypass {
+            newMessages = repliesBypass(messages)
+            let messageIds = Set(newMessages.map(\.id))
+            newChanges = changes.filter { messageIds.contains($0.item.id) }
+        } else {
+            newMessages = messages
+            newChanges = changes
+        }
```
nuno-vieira, line 344: This can be easily extracted and reused between Channel and Thread

Sources/StreamChatUI/ChatThread/ChatThreadVC.swift
```diff
@@ -150,11 +155,16 @@ open class ChatThreadVC: _ViewController,
 
     // MARK: - ChatMessageListVCDataSource
 
+    // The replies data source.
     public var messages: [ChatMessage] {
         get {
             replies
```
ipavlidakis, line 161: I have to admit that seeing returning replies for messages is a little bit confusing (i know this is not part of your changes just mentioning it)

Tests/StreamChatUITests/SnapshotTests/ChatChannel/ChatChannelVC_Tests.swift
```diff
@@ -549,6 +549,24 @@ final class ChatChannelVC_Tests: XCTestCase {
...
+
+        /// Regular case
+        vc.messagesBypass = { messages in messages.filter(\.isBounced) }
+        vc.messages = [.mock(isBounced: false), .mock(isBounced: true), .mock(isBounced: true)]
+        XCTAssertEqual(vc.messages.count, 2)
+    }
+
+    func test_messagesBypass_whenDidChangeMessages_shouldOverrideNewSnapshot_shouldOverrideChanges() {
+        XCTFail()
+    }
```
ipavlidakis, line 569: Do you mind adding a test with the messageBypass being nil just to ensure that we are not affecting the default behaviour?

Sources/StreamChatUI/Components.swift
```diff
@@ -100,6 +100,12 @@ public struct Components {
     /// Used in both the Channel and Thread view controllers.
     @available(iOSApplicationExtension, unavailable)
     public var messageListVC: ChatMessageListVC.Type = ChatMessageListVC.self
+
+    /// A closure to bypass the message list channel data source.
+    public var messagesBypass: (([ChatMessage]) -> [ChatMessage])?
```
ipavlidakis, line 105: One thing that I would like to point out is that blocks are more difficult to debug and provide useful information in stacktraces (after a crash). I would be more confident with a new protocol(maybe MessageFiltering) & object(provided from clients) pair.

PR #2361, opened 2022-11-02
[CIS-2234, CIS-2239, CIS-1727, CIS-2172] Improve Authentication Logic

polqf:
### 🔗 Issue Links

https://stream-io.atlassian.net/browse/CIS-2234
https://stream-io.atlassian.net/browse/CIS-2239
https://stream-io.atlassian.net/browse/CIS-1727
https://stream-io.atlassian.net/browse/CIS-2172

### 🎯 Goal

Improve overall logic around Authentication

### 📝 Summary

- Make sure waiters for token and connectionId timeout after certain time to avoid never completing calls to `connect()`
- Stop spamming errors to console while we refresh tokens upon expiration
- Update documentation around `currentUser` stating that a non-nil value does not mean that it is authenticated and the token is valid
- Add a development only way to test refresh token mechanisms
- Allow calling `connect()` after calling `setToken()`. This was an issue before as it was not creating background workers.

### 🛠 Implementation

- Token/ConnectionId requests are by default agnostic of whether they need a timeout or not. This is applied on the called function itself as a mechanism to streamline this logic.
- We are now not logging errors when "Socket is not connected" error is received. This is misleading, and is returned when the token expires.

### ⏭ Next steps

- Create internal Keychain to store token so that we
    - Add public APIs to know if token is expired
- Move token/connectionId waiters logic + protocols to a place where it makes sense
- Move token attempts logic away from the APIClient and into the Authentication Repository
- Properly handle expired tokens on open (either cached logic or initial token received being expired)

### ☑️ Contributor Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,12 @@ _October 27, 2022_
 - Fix CurrentChatUserController+Combine initialValue hard coded to `.noUnread` instead of using the initial value from the current user data model [#2334](https://github.com/GetStream/stream-chat-swift/pull/2334)
 - Allow Message Search pagination when using sort parameters [#2347](https://github.com/GetStream/stream-chat-swift/pull/2347)
 - Fix TokenProvider sometimes being invoked two times when token is expired [#2337](https://github.com/GetStream/stream-chat-swift/pull/2347)
+- Add timeout for token/connectionId providers so that `ChatClient.connect()` completes even in edge cases where we cannot get the needed data [#2361](https://github.com/GetStream/stream-chat-swift/pull/2361)
+- Stop spamming the console with "Socket is not connected" error when token is being refreshed [#2361](https://github.com/GetStream/stream-chat-swift/pull/2361)
+- Update documentation around `CurrentUserController.currentUser` to state that a non-nil value does not mean there is a valid authentication [#2361](https://github.com/GetStream/stream-chat-swift/pull/2361)
+- Allow flow where `ChatClient.setToken()` is called before `ChatClient.connect()` [#2361](https://github.com/GetStream/stream-chat-swift/pull/2361)
+- Properly recover from a missing/expired token on the first execution of `TokenProvider` [#2361](https://github.com/GetStream/stream-chat-swift/pull/2361)
+- Fix data races created by `AsyncOperation` looped execution when refreshing tokens [#2361](https://github.com/GetStream/stream-chat-swift/pull/2361)
 
 ## StreamChatUI
 ### ✅ Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 # Upcoming
 
-### 🔄 Changed
+## StreamChat
+### 🐞 Fixed
+- Add timeout for token/connectionId providers so that `ChatClient.connect()` completes even in edge cases where we cannot get the needed data [#2361](https://github.com/GetStream/stream-chat-swift/pull/2361)
+- Stop spamming the console with "Socket is not connected" error when token is being refreshed [#2361](https://github.com/GetStream/stream-chat-swift/pull/2361)
+- Update documentation around `CurrentUserController.currentUser` to state that a non-nil value does not mean there is a valid authentication [#2361](https://github.com/GetStream/stream-chat-swift/pull/2361)
+- Allow flow where `ChatClient.setToken()` is called before `ChatClient.connect()` [#2361](https://github.com/GetStream/stream-chat-swift/pull/2361)
+- Properly recover from a missing/expired token on the first execution of `TokenProvider` [#2361](https://github.com/GetStream/stream-chat-swift/pull/2361)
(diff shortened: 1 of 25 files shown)


DemoApp/Shared/Token+Development.swift
```diff
@@ -0,0 +1,79 @@
...
+    }
+}
+
+#if GENERATE_JWT
+
+import CryptoKit
+import Foundation
+import StreamChat
+
+let appSecret = ""
```
martinmitrevski, line 37: How's the secret going to be set? Manually or would there be another way?

DemoApp/Shared/Token+Development.swift
```diff
@@ -0,0 +1,79 @@
...
+                let token: Token
+                #if GENERATE_JWT
+                let timeInterval = TimeInterval(tokenDurationInMinutes * 60)
+                let generatedToken = _generateUserToken(
+                    secret: appSecret,
+                    userID: initialToken.userId,
+                    expirationDate: Date().addingTimeInterval(timeInterval)
+                )
+                if generatedToken == nil {
+                    print("Unable to generate token. Falling back to initialToken")
```
martinmitrevski, line 19: Although it's only the demo app, still better to use log instead of print.

Sources/StreamChat/Errors/ErrorPayload.swift
```diff
@@ -28,11 +28,21 @@ public struct ErrorPayload: LocalizedError, Codable, CustomDebugStringConvertibl
     }
 }
 
+/// https://getstream.io/chat/docs/ios-swift/api_errors_response/
+private enum StreamCode {
```
martinmitrevski, line 32: Isn't it better if we call it `StreamErrorCode`? Only `StreamCode` doesn't imply that we're talking about error codes.

Sources/StreamChat/Utils/InternetConnection/Error+InternetNotAvailable.swift
```diff
@@ -40,4 +40,8 @@ extension Error {
         }
         return false
     }
+
+    var isSocketNotConnectedError: Bool {
+        has(parameters: (NSPOSIXErrorDomain, 57))
```
martinmitrevski, line 45: Is this enough to check for socket errors? I also see https://developer.apple.com/documentation/foundation/nsstreamsockserrordomain.

martinmitrevski: I guess it's fine, since it's only used for logging.

martinmitrevski (approved):
Looks great 👏  I've added some minor comments.
Additionally, we should update our CHANGELOG and see why our checks fail. Also, would be nice to sync with @alteral about adding this to the e2e test suite.

Sources/StreamChat/Utils/Operations/AsyncOperation.swift
```diff
@@ -15,7 +15,7 @@ class AsyncOperation: BaseOperation {
     private var executedRetries = 0
 
     var canRetry: Bool {
-        executedRetries < maxRetries && !isCancelled
+        executedRetries < maxRetries && !isCancelled && !isFinished
```
polqf, line 18: Looks inoffensive but this actually created an edge case where we would be calling the completion of an API request up to 10 times at the SAME exact moment, leading to really bad data races.

DemoApp/Shared/Token+Development.swift
polqf: As of now, manually. We don't want anyone to be using this

CHANGELOG.md
```diff
@@ -17,6 +17,12 @@ _October 27, 2022_
 - Fix CurrentChatUserController+Combine initialValue hard coded to `.noUnread` instead of using the initial value from the current user data model [#2334](https://github.com/GetStream/stream-chat-swift/pull/2334)
 - Allow Message Search pagination when using sort parameters [#2347](https://github.com/GetStream/stream-chat-swift/pull/2347)
 - Fix TokenProvider sometimes being invoked two times when token is expired [#2337](https://github.com/GetStream/stream-chat-swift/pull/2347)
+- Add timeout for token/connectionId providers so that `ChatClient.connect()` completes even in edge cases where we cannot get the needed data [#2361](https://github.com/GetStream/stream-chat-swift/pull/2361)
```
martinmitrevski, line 20: This should go in the upcoming version, not in 4.23.0.

martinmitrevski (approved):
✅ Only the CHANGELOG needs to be updated, it's added to the previous version.

CHANGELOG.md
polqf: 🤦 😂

Sources/StreamChat/Utils/Operations/AsyncOperation.swift
nuno-vieira: Can we add a comment explaining that then? Someone might remove this since it looks inoffensive 🤔

nuno-vieira (approved):
LGTM! ✅

Although wasn't part of the https://stream-io.atlassian.net/browse/CIS-2172 to add the refresh token to the Test App as well?

PR #2366, opened 2022-11-08
Package: update macOS to "11.0"

AndrewSB:
this appears to still not be working: i initially offered over at https://github.com/GetStream/stream-chat-swift/pull/2357. i'm still getting this error on `develop`:

```
error: the library 'StreamChatTestMockServer' requires macos 10.15, but depends on the product 'StreamChatTestHelpers' which requires macos 11.0; consider changing the library 'StreamChatTestMockServer' to require macos 11.0 or later, or the product 'StreamChatTestHelpers' to require macos 10.15 or earlier.
```

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -8,7 +8,7 @@ let package = Package(
     name: "StreamChat",
     defaultLocalization: "en",
     platforms: [
-        .iOS(.v11), .macOS(.v10_15)
+        .iOS(.v11), .macOS("11.0")
     ],
     products: [
         .library(


alteral:
Hi @AndrewSB,

Thanks for the PR. I suppose you are using `StreamChatSwiftUI`. If so, unfortunately, you need to wait for a new release because the fix was provided on `develop` but `StreamChatSwiftUI` points to the latest version of `StreamChat`.

Best wishes,
Alexey

AndrewSB:
`develop` on this repository still is set to 10.15, not 11. am i misunderstanding the branching strategy you use for this project?

https://github.com/GetStream/stream-chat-swift/blob/develop/Package.swift

nuno-vieira:
> `develop` on this repository still is set to 10.15, not 11. am I misunderstanding the branching strategy you use for this project?
> 
> https://github.com/GetStream/stream-chat-swift/blob/develop/Package.swift

Hi @AndrewSB,

We downgraded the other packages to 10_15. But if you are using SwiftUI, you need to wait for a new StreamChat release , so that SwiftUI pickups up this change. So, in the next release, this will be fixed. We could not bump to v11, since that could break some customers.

AndrewSB:
ahh i understand, you lowered the test helper version in https://github.com/GetStream/stream-chat-swift-test-helpers/commit/237da98e4946ed8e2dac51d9e82e713b0d47a298

got it, thanks!

PR #2369, opened 2022-11-09
[CIS-2262] Allow changing attachment info with custom CDN

nuno-vieira:
### 🔗 Issue Links
CIS-2262

### 🎯 Goal
Currently, our `CDNClient` protocol expects always to return a single `URL`. This is very limited since a customer might want to update the attachment info using their file uploader API, like changing the preview URL or adding extra data.

The goal of this PR is to improve this and allow the customer to change the final attachment when uploading to their CDN. 

### 🛠 Implementation
- Introduces a new `AttachmentUploader` and renames the old one to `AttachmentQueueUploader`.
- `channelController.uploadFile()` and `channelController.uploadImage()` are now deprecated in favour of `channelController.uploadAttachment()`

The `CDNClient` can be customized if you are only interested in changing the CDN Url and do not want to update the attachment payload.

The `AtttachmentUploader` can be customized for more fine-grain control, in case you want to update the final attachment with additional information.

### 🧪 Manual Testing Notes
Probably worth doing smoke testing related to loading images.

### ☑️ Contributor Checklist


Sources/StreamChat/APIClient/CDNClient/CDNClient.swift
```diff
@@ -81,20 +76,28 @@ class StreamCDNClient: CDNClient {
...
                     )
-                    completion(.success(decodedResponse.file))
+
+                    let uploadedFile = UploadedFile(
+                        remoteURL: decodedResponse.fileURL,
+                        remotePreviewURL: decodedResponse.previewURL
+                    )
+
+                    // TODO: For v5, CDNClient should return the `UploadedFile` directly.
+                    // Especially if we want to support video thumbnails.
```
martinmitrevski, line 94: Wasn't the video thumbnail support part of the CDN implementation?

Sources/StreamChat/APIClient/CDNClient/CDNClient.swift
```diff
@@ -4,15 +4,15 @@
...
     static var maxAttachmentSize: Int64 { get }
 
+    /// Uploads attachment as a multipart/form-data and returns only the uploaded remote file.
     /// - Parameters:
-    ///   - attachment: An attachment to upload
-    ///   - progress: A closure that broadcasts upload progress
-    ///   - completion: Returns uploading result on upload completion or failure
+    ///   - attachment: An attachment to upload.
+    ///   - progress: A closure that broadcasts upload progress.
+    ///   - completion: Returns the file uploaded information.
```
martinmitrevski, line 15: maybe "the uploaded file's information."?

Sources/StreamChat/Config/ChatClientConfig.swift
```diff
@@ -112,18 +112,35 @@ public struct ChatClientConfig {
     /// - Parameter apiKey: The API key of the chat app the `ChatClient` connects to.
     ///
     
-    /// Allows to inject a custom API client for uploading attachments, if not specified `StreamCDNClient` is used
+    /// Allows to inject a custom API client for uploading attachments, if not specified, `StreamCDNClient` is used.
     public var customCDNClient: CDNClient?
+
+    /// Allows to inject a custom attachment uploader. It can be used to have more
+    /// control than `CDNClient` to allow changing the attachment payload.
+    public var customAttachmentUploader: AttachmentUploader?
```
martinmitrevski, line 120: Maybe we should specify that if this one is used, the `customCDNClient` above doesn't have any effect?

Sources/StreamChat/Models/Attachments/AnyAttachmentUpdater.swift
```diff
@@ -0,0 +1,32 @@
...
+        _ updates: ((inout Payload) -> Void)
+    ) {
+        do {
+            if let attachmentPayload = attachment.attachment(payloadType: payloadType)?.payload {
+                var payload = attachmentPayload
+                updates(&payload)
+                attachment.payload = try JSONEncoder.stream.encode(payload.asAnyEncodable)
+            }
+        } catch {
+            log.debug(error.localizedDescription)
```
martinmitrevski, line 29: `log.error` would be more appropriate here.

Sources/StreamChat/Models/Attachments/ChatMessageImageAttachment.swift
```diff
@@ -94,15 +118,12 @@ extension ImageAttachmentPayload: Decodable {
                 container.decodeIfPresent(String.self, forKey: .name)
         )?.trimmingCharacters(in: .whitespacesAndNewlines)
 
-        let previewUrl = try container.decodeIfPresent(URL.self, forKey: .thumbURL) ?? imageURL
```
martinmitrevski, line 97: Did we decode this value from the payload? If yes, then this would be a breaking change.

docusaurus/docs/iOS/client/custom-cdn.md
```diff
@@ -0,0 +1,141 @@
...
+                ]
+            }
+
+            completion(.success(uploadedAttachment))
+        }
+    }
+}
+```
+
+The `AnyAttachmentUpdater` is an helper component provided by Stream, to make it easier to update the underlying payload of a type-erased attachment. You should pass a reference of the attachment with `&` and say which payload to update depending on what type is the attachment.
```
martinmitrevski, line 135: small typo: "is a helper component"

martinmitrevski (commented):
Looks good in general. I'm only worried about this line:
`let previewUrl = try container.decodeIfPresent(URL.self, forKey: .thumbURL) ?? imageURL`.
It seems we were previously decoding the thumbURL and setting the previewURL after all. Now this logic is gone, which might be a breaking change.

nuno-vieira:
> Looks good in general. I'm only worried about this line:
> 
> `let previewUrl = try container.decodeIfPresent(URL.self, forKey: .thumbURL) ?? imageURL`.
> 
> It seems we were previously decoding the thumbURL and setting the previewURL after all. Now this logic is gone, which might be a breaking change. 

This does not exist, our API does not return any thumb_url for Images. Also, our ImageAttachmentPayload does not encode that value, so it would never be sent to the server even if has a customer you updated the previewUrl. So this is fine to remove

Sources/StreamChat/APIClient/CDNClient/CDNClient.swift
nuno-vieira: The only thing missing is the thumbnail for videos, which at the moment on iOS it is not a priority since we already have it locally. So a separate PR should handle that in the future

Sources/StreamChat/APIClient/CDNClient/CDNClient.swift
nuno-vieira: The only thing missing is the thumbnail for videos, which at the moment on iOS it is not a priority since we already have it locally. So a separate PR should handle that in the future

Sources/StreamChat/Config/ChatClientConfig.swift
nuno-vieira: Yes make sense. Also good to say this in docusaurud

Sources/StreamChat/Models/Attachments/ChatMessageImageAttachment.swift
nuno-vieira: This does not exist, our API does not return any thumb_url for Images. Also, our ImageAttachmentPayload does not encode that value, so it would never be sent to the server even if has a customer you updated the previewUrl. So this is fine to remove

martinmitrevski (approved):
great work @nuno-vieira 👏

PR #2370, opened 2022-11-10
[CIS-2235] Properly arrange Authentication and Connection logic

polqf:
### 🔗 Issue Links

https://stream-io.atlassian.net/browse/CIS-2235

### 🎯 Goal

### 📝 Summary

### 🛠 Implementation

### 🎨 Showcase

| Before | After |
| ------ | ----- |
|  img   |  img  |

### 🧪 Manual Testing Notes

### ☑️ Contributor Checklist

### 🎁 Meme


Sources/StreamChat/Repositories/ConnectionRepository.swift
```diff
@@ -0,0 +1,225 @@
...
+        self.webSocketClient = webSocketClient
+        self.apiClient = apiClient
+        self.timerType = timerType
+    }
+
+    /// Connects the chat client the controller represents to the chat servers.
+    ///
+    /// When the connection is established, `ChatClient` starts receiving chat updates, and `currentUser` variable is available.
+    ///
+    /// - Parameter completion: Called when the connection is established. If the connection fails, the completion is
```
martinmitrevski, line 54: We should also document the `userInfo` here.

Sources/StreamChat/ChatClient.swift
```diff
@@ -479,15 +446,40 @@ public class ChatClient {
...
+            createBackgroundWorkers()
+        case .newToken:
+            if backgroundWorkers.isEmpty {
+                createBackgroundWorkers()
+            }
         }
     }
 }
 
+/// `APIClient` listens for `WebSocketClient` connection updates so it can forward the current connection id to
```
martinmitrevski, line 461: Is this comment still relevant?

Sources/StreamChat/ChatClient.swift
```diff
@@ -229,15 +235,6 @@ public class ChatClient {
     private let sessionHeaders: [String: String] = [
         "X-Stream-Client": SystemEnvironment.xStreamClientHeader
     ]
-    
-    /// The current connection id
-    @Atomic var connectionId: String?
-    
-    /// An array of requests waiting for the connection id
-    @Atomic private(set) var connectionIdWaiters: [String: (String?) -> Void] = [:]
```
martinmitrevski, line 237: Great that we're getting rid of these - now they're in the correct place and no `@Atomic` is needed. 👍

Sources/StreamChat/Repositories/ConnectionRepository.swift
```diff
@@ -0,0 +1,225 @@
...
+import Foundation
+
+class ConnectionRepository {
+    private let connectionQueue: DispatchQueue = DispatchQueue(label: "io.getstream.connection-repository", attributes: .concurrent)
+    private var _connectionIdWaiters: [String: (String?) -> Void] = [:]
+    private var _connectionId: ConnectionId?
+    private var _connectionStatus: ConnectionStatus = .initialized
+
+    private var connectionIdWaiters: [String: (String?) -> Void] {
+        get { connectionQueue.sync { _connectionIdWaiters } }
```
martinmitrevski, line 14: Not in the scope of this PR, but food for thought: since we will be adding more thread-safe repos in the future, maybe we should think about abstracting this logic for getters and setters, since it will repeat itself?

martinmitrevski (commented):
Looks really nice @polqf, much cleaner and in line with the direction we want to go in 👏 

As already discussed, we should merge this after our next release. Also, @alteral's token-related e2e tests would need to be run before merging. We might even do a common session to do some QA, if it makes sense.

Additionally, we need tests and see why the checks are failing.

Sources/StreamChat/Repositories/ConnectionRepository.swift
polqf: I did not do that because it requires us to make that wrapping class as public in some cases...
But might be trying to find a workaround on that soon

Sources/StreamChat/Repositories/AuthenticationRepository.swift
```diff
@@ -140,6 +153,98 @@ class AuthenticationRepository {
...
+        } else if newToken.userId == currentUserId {
+            state = .newToken
+        } else {
+            state = .newUser
+        }
+
+        log.assert(delegate != nil, "Delegate should not be nil at this point")
+
+        switch state {
+        case .firstConnection:
```
martinmitrevski, line 173: The handlings for both `firstConnection` and `newToken` are the same. Maybe we can put them in the same case (or extract these lines in a method)?

Sources/StreamChat/Repositories/AuthenticationRepository.swift
```diff
@@ -140,6 +153,98 @@ class AuthenticationRepository {
...
+            delegate?.didFinishSettingUpAuthenticationEnvironment(for: state)
+
+            // Setting a new connection is not possible in connectionless mode.
+            guard connectionRepository.isClientInActiveMode else {
+                completion(ClientError.ClientIsNotInActiveMode())
+                return
+            }
+
+            connectionRepository.disconnect(source: .userInitiated) { [weak delegate, weak connectionRepository] in
+                connectionRepository?.updateWebSocketEndpoint(with: newToken, userInfo: userInfo)
```
martinmitrevski, line 197: I think I've commented somewhere about the order of these two - it's clearer if we first clear current user data and then update the connection repository with a new token.

Sources/StreamChat/Repositories/AuthenticationRepository.swift
```diff
@@ -163,16 +268,48 @@ class AuthenticationRepository {
...
+        let onCompletion: (Error?) -> Void = { [weak self] error in
+            guard let self = self else { return }
+
             if let error = error {
                 log.error("Error when getting token: \(error)", subsystems: .authentication)
             } else {
                 log.debug("Successfully retrieved token", subsystems: .authentication)
             }
-            let completionBlocks: [(Error?) -> Void]? = self?.tokenRequestCompletions
+            let completionBlocks: [(Error?) -> Void]? = self.tokenQueue.sync {
```
martinmitrevski, line 280: This is already handled in the other PR.

martinmitrevski (approved):
looks really good @polqf 👍 Added few comments, but nothing major. 
Should we have a CHANGELOG entry for this? Also, let's merge it after the patch release.

Sources/StreamChat/Repositories/AuthenticationRepository.swift
polqf: Oh wow, did not realize 😂 After abstracting the logic, there are no more differences

Sources/StreamChat/Repositories/AuthenticationRepository.swift
polqf: It looks semantically incorrect. The second method has a completion block, and we should not make assumptions on whether the completion block will be run sync or async. That's why I kept that order.

Another option would be to add this line to the completion of `clearCurrentUserData`, but seems too much for this.

martinmitrevski: ok, let's keep it in this order then.

Sources/StreamChat/Repositories/ConnectionRepository.swift
nuno-vieira: I think writing Repositories as concrete classes usually are not the best practice. Repositories by definition are interface adapters, they abstract how we fetch data. By creating concrete classes instead it does not give us the flexibility that repositories should give. With interfaces, we can easily create Decorators, which abstract the Threading code. This will also improve a lot how we test this. 

As abstracting all together thread-safe code, I don't think it will be possible, unless we go back to @Atomic 🤔 Which we want to move away from.

Sources/StreamChat/Controllers/ConnectionController/ConnectionController.swift
```diff
@@ -93,8 +93,8 @@ public extension ChatConnectionController {
     /// called with an error.
     ///
     func connect(completion: ((Error?) -> Void)? = nil) {
-        chatClientUpdater.connect { error in
-            self.callback {
+        connectionRepository.connect { [weak self] error in
+            self?.callback {
```
nuno-vieira, line 97: I remember before, we could not have weak-self, otherwise the Updater would not work. This is not the case anymore?

nuno-vieira, line 97: I remember before, we could not have weak-self, otherwise, the Updater would not work. This is not the case anymore?

Sources/StreamChat/Repositories/ConnectionRepository.swift
```diff
@@ -0,0 +1,225 @@
...
+                }
+            }
+        }
+        webSocketClient?.connect()
+    }
+
+    /// Disconnects the chat client the controller represents from the chat servers. No further updates from the servers
+    /// are received.
+    func disconnect(
+        source: WebSocketConnectionState.DisconnectionSource = .userInitiated,
```
nuno-vieira, line 94: Can we actually make this argument without a default value? On the client side, maybe it makes sense, but here in the repository, it makes sense to not omit this decision 🤔

nuno-vieira (approved):
LGTM! ✅ Just added a couple of NITs, not a blocker at all

Sources/StreamChat/Controllers/ConnectionController/ConnectionController.swift
polqf: Tests were failing because we test that the controller is successfully deallocated, so had to change that.
Still, rest of the tests are passing fine, so no issue there.

Sources/StreamChat/Repositories/ConnectionRepository.swift
polqf: Good point

PR #2371, opened 2022-11-11
[CIS-2264] Make sure to reload rows after moving them

polqf:
### 🔗 Issue Links

https://stream-io.atlassian.net/browse/CIS-2264

### 📝 Summary

UIKit does not reload rows when they are moved. This is really bad for us because when we receive a .move update, it comes with changes to the cell itself. In most of the cases, because of other events, this is not an issue. But in edge cases, the cell stays with the old state until a new event comes in.

This PR ensures that whenever those .move events come, we always update the cell just in case.

### 🛠 Implementation

- Unified logic for CollectionViewListChangeUpdater and TableViewListChangeUpdater so that we don't need to repeat the code around
- Ensured cells are reloaded after a move

### 🧪 Manual Testing Notes

1. Open the channel list
2. Send A SINGLE message to a channel that is on screen, but it is not in the first position.

Expected result:
- The channel should move to the first position
- The cell should display an updated read indicator, and the typing indicator should be gone

Previous result:
- The channel would move to the first position
- BUT the cell was displaying old information, and the typing indicator was still there

### ☑️ Contributor Checklist


nuno-vieira (approved):
LGTM! ✅

PR #2372, opened 2022-11-11
[CIS-2172] Add UI tests for JWT authentication

alteral:
### 🔗 Issue Links

- https://stream-io.atlassian.net/browse/CIS-2172

### 🎯 Goal

We need to test JWT

### 🛠 Implementation

- JWT generation is implemented through [stream-chat-ruby](https://github.com/GetStream/stream-chat-ruby) (as a real backend to generate valid tokens) and [sinatra](https://github.com/sinatra/sinatra) (as a mock server to return expired and valid tokens on the fly)
    - When the test needs a valid token, `sinatra` goes to the backend and returns a brand new token that is valid for 3 seconds
    - When the test needs an expired token, `sinatra` returns the hard-coded expired token
    - When the test needs an error from the backend, `sinatra` returns status `500`
- On the client side, we have two new server methods:
    - `revokeJwt()` - to revoke the next token that will be requested by the client
    - `breakJwt()` - to break the token generation and return an error when the next token will be requested by the client
- Covered scenarios:
    - `test_tokenExpiriesBeforeUserLogsIn`
    - `test_tokenExpiriesAfterUserLoggedIn`
    - `test_tokenExpiriesWhenUserIsInBackground`
    - `test_tokenExpiriesWhileUserIsOffline`
    - `test_tokenGenerationFails`

### ☑️ Contributor Checklist

### 🎁 Meme

5834555ede7a912e00000942


nuno-vieira (approved):
LGTM! ✅

Issue #2368, opened 2022-11-08
How are we meant to determine user's visibility preference?


sgoodwin:
## What are you trying to achieve?

I see in the docs where you can set a user to "invisible" mode so they don't show as online, but I want to make that a control the user can change. How are we meant to determine that the user has been set to invisible?

## If possible, how can you achieve this currently?

I thought to check the `lastActiveAt` value, but that remains populated wether the user is invisible or not. I also looked in the `extraData` field, but there's no hint there either.

## What would be the better way?

Ideally next to `isOnline` there's also be `isVisible`.

## GetStream Environment
**GetStream Chat version:** main branch
**GetStream Chat frameworks:** StreamChat
**iOS version:** 16
**Swift version:** 5
**Xcode version:** 14.1
**Device:** iPhone 13 Pro

polqf:
Hi @sgoodwin ,

By invisible you mean a banned user? I'm sorry I am not fully understanding what you are doing and what is your expected outcome.

sgoodwin:
@polqf I mean to hide a user's online status. Is that not a thing we can do?

polqf:
This is a feature that is currently not available in iOS. We will be prioritizing it in our backlog! 🙏

sgoodwin:
From poking around, it looks like `invisible` is a property present in the `CurrentUserPayload`, it just isn't persisted into the `CurrentUserDTO` so and subsequently pulled out into `CurrentUser`.

polqf:
Correct @sgoodwin. This is something the API already returns, but we haven't implemented on the SDK side.

polqf:
Hi! This has been fixed in our latest release [4.25.0](https://github.com/GetStream/stream-chat-swift/releases/tag/4.25.0)
Closing this issue for now, feel free to reopen it if you encounter any issues.

PR #2373, opened 2022-11-11
[CIS-2295] Add support for hiding connection status with `isInvisible`

sgoodwin:
### 🔗 Issue Links

https://github.com/GetStream/stream-chat-swift/issues/2368
CIS-2295

### 🎯 Goal

The goal of this task is to be able to hide the connection status when connecting a user and also expose this property to the current user.

### 🛠 Implementation
- Adds `UserInfo.isInvisible` to be able to set the value when connecting the user
- Adds `CurrentUserDTO.isInvisible` to be able to get the value when the user is already connected

### 🧪 Manual Testing Notes
- Open a channel with User A
- With another device, before connecting User B, tap the Configuration Button in Demo App
- Enable `isInvisible` in the user info
- Open a channel with the User B
- User A should no see User B has online in the Channel Header View

### ☑️ Contributor Checklist


nuno-vieira:
Hi @sgoodwin!

Thank you a lot for your contribution 🙂 We will handle it from here 👍 I'm going to add a couple more tests and add the option to make yourself invisible when connecting.

Best,
Nuno

nuno-vieira (approved):
LGTM! ✅

Sources/StreamChat/Models/CurrentUser.swift
```diff
@@ -52,12 +52,16 @@ public class CurrentChatUser: ChatUser {
     
     /// The unread counts for the current user.
     public let unreadCount: UnreadCount
+
+    /// A Boolean value indicating if the user has opted to hide their online status.
+    public let isInvisible: Bool
```
polqf, line 57: Are we sure this value will always exist? Would be nice to check with BE

nuno-vieira: Yes, either way it has a default value of false

sgoodwin:
Thank you all for this :)

Issue #2381, opened 2022-11-20
Crash after calling logout


moubry:
## What did you do?

I implemented user sign out in my app. As part of that process, I'm calling `ChatClient.shared.logout()` function, so that the user session with Stream is destroyed, so if they sign in again with a different user, there’s nothing leftover from the previous Stream user session.

## What did you expect to happen?

I expected it to not crash.

## What happened instead?

Stream crashes intermittently.

## GetStream Environment
**GetStream Chat version:** 4.24.0
**GetStream Chat frameworks:** StreamChat, StreamChatUI
**iOS version:** latest
**Swift version:** latest
**Xcode version:** latest
**Device:** latest

## Additional context

I realize an intermittent crash during logout is likely not a very high priority issue. I’m only logging it on the off-chance that this is helpful information.

I cannot reliably reproduce it. However, this specific crash occurred while the StreamChatSwiftUI channel was visible, and we detected an expired user session in our app, so we destroyed the view, and called logout on the Stream client at the exact same time. I’m guessing it’s some kind of race condition.

<img width="1694" alt="Screen Shot 2022-11-19 at 9 33 10 PM" src="https://user-images.githubusercontent.com/144283/202882604-1b0a8cb5-26d8-4471-a4c6-11475b1cd756.png">

polqf:
Hi @moubry ,

Thank you for your report. This is something that we are actually investigating and we expect to have a fix for it soon.

Long story short, log out does not finish all the "clean up" synchronously, and we discovered a potential chance of race conditions if you log in just after calling it.

vjard33:
Same for me... logout() doesn't clean properly on 4.22.0 (same for 4.24.0)
TokenProvider seems to be broken on 4.24.0, sometimes I got some double completionHandler call from one call, really weird... That's why I'm stuck on 4.22.0 (and for these crashes...)

vjard33:
> Hi @moubry ,
> 
> Thank you for your report. This is something that we are actually investigating and we expect to have a fix for it soon.
> 
> Long story short, log out does not finish all the "clean up" synchronously, and we discovered a potential chance of race conditions if you log in just after calling it.

@polqf @nuno-vieira 
Can you tell me if this logout() bug exists on 4.22.0? I want to upgrade to 4.24.1, but don't know if this is a good idea. I didn't see any crash on logout() on 4.22.0.

polqf:
[...]

PR #2386, opened 2022-11-23
[CIS-2243, CIS-2244] Increase Safety by making Log Out and Disconnect async operations

polqf:
### 🔗 Issue Links

https://stream-io.atlassian.net/browse/CIS-2243
https://stream-io.atlassian.net/browse/CIS-2244
https://github.com/GetStream/stream-chat-swift/issues/2381

### 🎯 Goal

Avoid race conditions or crashes originated after calling log out

### 📝 Summary

`logOut` does not finish all the "clean up" synchronously, which can potentially create issues when interacting with the SDK just after calling it.

### 🛠 Implementation

Synchronous `logOut` and `disconnect` calls are deprecated in favour of its asynchronous version

### 🧪 Manual Testing Notes

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/ChatClient.swift
+++ b/Sources/StreamChat/ChatClient.swift
@@ -364,27 +364,52 @@ public class ChatClient {
             completion: { completion?($0) }
         )
     }
-    
+
     /// Disconnects the chat client from the chat servers. No further updates from the servers
     /// are received.
+    @available(*, deprecated, message: "Use the asynchronous version of `disconnect` for increased safety")
     public func disconnect() {
+        disconnect {}
+    }
+
+    /// Disconnects the chat client from the chat servers. No further updates from the servers
+    /// are received.
+    public func disconnect(completion: @escaping () -> Void) {
         connectionRepository.disconnect(source: .userInitiated) {
             log.info("The `ChatClient` has been disconnected.", subsystems: .webSocket)
+            completion()
         }
         authenticationRepository.clearTokenProvider()
     }
-    
+
     /// Disconnects the chat client form the chat servers and removes all the local data related.
+    @available(*, deprecated, message: "Use the asynchronous version of `logout` for increased safety")
     public func logout() {
-        disconnect()
+        logout {}
+    }
+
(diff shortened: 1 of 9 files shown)


martinmitrevski (commented):
LGTM ✅ We need to update the changelog, check if docs need to be updated and write some tests for the async logout (especially the scenario that previously crashed).

Sources/StreamChat/Controllers/ConnectionController/ConnectionController.swift
```diff
@@ -103,20 +103,12 @@ public extension ChatConnectionController {
     /// Disconnects the chat client the controller represents from the chat servers.
     /// No further updates from the servers are received.
     func disconnect() {
-        chatClientUpdater.disconnect(source: .userInitiated) {
+        connectionRepository.disconnect(source: .userInitiated) {
```
nuno-vieira, line 106: Shouldn't this disconnect have a completion block as well?

Sources/StreamChat/ChatClient.swift
```diff
@@ -367,27 +364,52 @@ public class ChatClient {
...
     /// are received.
+    @available(*, deprecated, message: "Use the asynchronous version of `disconnect` for increased safety")
     public func disconnect() {
-        clientUpdater.disconnect(source: .userInitiated) {
+        disconnect {}
+    }
+
+    /// Disconnects the chat client from the chat servers. No further updates from the servers
+    /// are received.
+    public func disconnect(completion: @escaping () -> Void) {
```
nuno-vieira, line 377: I'm still concerned about this API change. Would be nice to have some customers try this new API and see the potential impact on their apps. If this is problematic we could opt in the approach of moving this to the login side.

polqf: The previous API will still be available, under a deprecation note.

martinmitrevski (approved):
LGTM ✅

nuno-vieira:
@polqf The Demo App looks that is not building

nuno-vieira (approved):

PR #2394, opened 2022-11-30
[CIS-2009] Sync mock server with the backend

alteral:
### 🔗 Issue Links

- https://stream-io.atlassian.net/browse/CIS-2009

### 🎯 Goal

- To keep back-end HTTP and Websocket responses up-to-date in the Mock Server

### 📝 Summary

- Sync will be performed every 1st and 15th of the month at 0:00
    - Json templates will be collected via [AIOHTTP](https://github.com/aio-libs/aiohttp) and [Stream REST API](https://getstream.io/chat/docs/rest)
- PR with updated json templates will be opened from `ci/sync-mock-server-{timestamp}` to `develop`
    - E2E tests will be executed based on new templates

### 🎨 Showcase

- [Sync example](https://github.com/GetStream/stream-chat-swift/actions/runs/3594999477):

Screen Shot 2022-12-01 at 4 06 54 PM

### 🧪 Manual Testing Notes

1. Install dependencies: `pip3 install -r requirements.txt`
2. Run `bundle exec fastlane sync_mock_server`
3. Check out local git changes

**PS**: tested both UIKit (on CI) and SwiftUI (locally) with auto-generated json templates

### ☑️ Contributor Checklist

### 🎁 Meme

gif


TestTools/StreamChatTestMockServer/Fixtures/JSONs/ws_health_check.json
```diff
@@ -1,44 +1,2953 @@
...
-
+    "cid": "*",
+    "connection_id": "6315fc7c-0a05-4049-0000-00001647dac7",
+    "created_at": "2022-12-01T12:56:41.765102666Z",
+    "me": {
+        "banned": false,
+        "birthland": "Tatooine",
+        "channel_mutes": [
+            {
+                "channel": {
```
nuno-vieira, line 10: Why the ws_health_check is so big? Shouldn't this be split up between multiple events?

fastlane/sync_mock_server.py
```diff
@@ -0,0 +1,207 @@
+import aiohttp
+import asyncio
+import json
+import uuid
+import os
```
nuno-vieira, line 5: Any reason why we are doing this with AIOHTTP and not, for example, Sinatra, since we are using Sinatra for other things? My concern is that we might be adding too much tech here, were we could just stick to ruby 🤔

fastlane/sync_mock_server.py
```diff
@@ -0,0 +1,207 @@
...
+      await send_typing_event(session, ws, channel_id)
+      message_id = await send_regular_message(session, ws, channel_id)
+      await add_reaction(session, ws, message_id)
+      await add_member_to_channel(session, ws, channel_id)
+      await send_attachment(session, channel_id)
+      await send_ephemeral_message(session, channel_id)
+      await send_youtube_link(session, channel_id)
+      await send_unsplash_link(session, channel_id)
+      await truncate_channel_with_messsage(session, channel_id)
+      await remove_channel(session, channel_id)
```
nuno-vieira, line 205: Looks really nice 👌

nuno-vieira (approved):
Looks really good 👌 This is a really nice thing to have. Just added a couple of questions, but overall already LGTM 🙂

fastlane/sync_mock_server.py
alteral: Eh, to be honest, I'd love to, we use Ruby for almost everything around building and scripting (fastlane + sinatra), but for this particular task I just could not come up with something on Ruby - I've searched for any possible solution on Ruby to work with WebSockets, but it did not work out with any of them. Unfortunately, most of them are quite old and not supported. With Python, it just worked like a charm from the first try.

TestTools/StreamChatTestMockServer/Fixtures/JSONs/ws_health_check.json
alteral: Yeah, it is huge. This is basically what you get when connecting to a web socket in our DemoApp by luke_skywalker. I could parse its content on the fly and remove everything but the first channel, but honestly do not know if it's worth it.

alteral: Hey @nuno-vieira, check out my latest commit ([9b6187](https://github.com/GetStream/stream-chat-swift/pull/2394/commits/9b6187f869063256971f0e42ce932272d975d52d)), I've reduced the size of both `ws_health_check` and `http_channels`.

Hope the tests will pass 😄

nuno-vieira: @alteral If the response actually comes exactly like this, we can leave it. It's probably not worth adding more complexity. So I guess we can revert that change, if you want

alteral: No, tbh, I like the idea of reducing those JSON templates. We are overwriting these objects in the tests anyway. Now it will be easier to review automated PRs.

Issue #2384, opened 2022-11-22
Crash when calling `addDevice()` from background thread 


vjard33:
## What did you do?
Try to add a device after connectUser, sometimes it crashes (this is the biggest crash we have on production on our app: 256 crashes).
ChatClient.shared.currentUserController().addDevice(.firebase(token: token, providerName: providerName))

## What did you expect to happen?
Not crash. ChatClient.shared.currentUserController().addDevice should addDevice if the token doesn't exist, if it exists: nothing.
```
        // Connect the client with the token provider.
        ChatClient.shared.connectUser(userInfo: userInfo, tokenProvider: tokenProvider) { error in
            if error == nil {
                // Stream Chat push notifications.
                if ChatClient.shared.currentUserId != nil {
                    if let fcmToken = Messaging.messaging().fcmToken {
                        ChatClient.shared.currentUserController().addDevice(.firebase(token: fcmToken, providerName: ChatClient.providerName))
                    }
                }

                successCompletion?()
            } else {
                failCompletion?(error)
            }
        }
```

## What happened instead?
Crash 
<img width="1424" alt="Capture d’écran 2022-11-22 à 21 21 50" src="https://user-images.githubusercontent.com/99725306/203414485-6c4a0812-953c-40f0-8b5e-787432b52f6d.png">

## GetStream Environment
**GetStream Chat version:** 4.22.0
**GetStream Chat frameworks:** StreamChat, StreamChatUI
**iOS version:** iOS 16.0.0, iOS 16.1.0, iOS 16.1.1
**Swift version:**5.7
**Xcode version:**14.1
**Device:**iPhone XS to iPhone 14 (seems to be more reproducible on iPhone 11)

## Additional context
I can't upgrade to 4.24.0 because I've weird behavior with TokenProvider after some logout/connectUser (completion called twice or more..)

nuno-vieira:
Hi @vjard33,

When you are logging out a user, do you call `logout()`? Can you see in the analytics of your app the user journey and check if he previously logged out and then logged in? 

More details here: https://getstream.io/chat/docs/sdk/ios/uikit/getting-started/#disconnect--logout

Best,
Nuno

nuno-vieira:
@vjard33 The docs also mention to add the device on the register for push notifications call, here: https://getstream.io/chat/docs/sdk/ios/client/push-notifications/

Any reason why you are doing it after the connectUser call?

polqf:
[...]

PR #2398, opened 2022-11-30
Fix crash when calling `addDevice()` from background thread

nuno-vieira:
### 🔗 Issue Links
https://github.com/GetStream/stream-chat-swift/issues/2384

### 🎯 Goal
Fixes crash when calling `currentUserController.addDevice()` from a background thread.

### 🧪 Manual Testing Notes
N/A

### ☑️ Contributor Checklist


Sources/StreamChat/Workers/CurrentUserUpdater.swift
```diff
@@ -71,31 +71,34 @@ class CurrentUserUpdater: Worker {
...
+
+        let backgroundContext = database.backgroundReadOnlyContext
+        backgroundContext.perform { [weak self] in
+            if let currentUserDTO = backgroundContext.currentUser,
+               currentUserDTO.devices.first(where: { $0.id == deviceId }) != nil {
+                saveCurrentDevice(deviceId: deviceId, completion: completion)
+                return
+            }
+
+            self?.apiClient
```
polqf, line 83: We should not be starting an API request from a background thread 👀 . It would be better to get the model, and then switch back to the (presumably) main thread

nuno-vieira: Why not? 🤔

nuno-vieira: The request call with be added to the operation queue, so there's no point here to change it to another thread since it will be instantly changed again to the operation queue in the chat client.

martinmitrevski (approved):
LGTM ✅  Please update the branch and check the failing checks before merging.

PR #2414, opened 2022-12-12
[Internal] Add retries to Authentication Repository

alteral:
### 🎯 Goal

Fix E2E tests which were failing due to missing retries after a token fetch failure

### 🛠 Implementation

This PR adds retries when a token fetch process fails. It does it under the following circumstances:
- If the fetched token is expired
- If there is an error returned
- All the above is done a up to a maximum of 10 times

In order to do that, it also removes state tracking of token attempts refresh from the APIClient.
Now, whenever the Authentication Repository is fetching a token, it notifies the APIClient so it can temporarily suspend the queue. In the case where requests are already in the queue and are being executed, those will be requeued as soon as an expired token is received, to then be relaunched when the queue is resumed.

To be quicker detecting expired tokens, we also try to refresh tokens when the connection changes to `.disconnecting`, as well as when it does to `.disconnected`, and the server error reflects an expired token. Before we were only doing so when the state was `.disconnected`.

### 🎨 Showcase
LLC excalidraw copy

### ☑️ Contributor Checklist

### DIFF
--- a/Tests/StreamChatTests/Repositories/AuthenticationRepository_Tests.swift
+++ b/Tests/StreamChatTests/Repositories/AuthenticationRepository_Tests.swift
@@ -27,6 +27,7 @@ final class AuthenticationRepository_Tests: XCTestCase {
             tokenExpirationRetryStrategy: retryStrategy,
             timerType: DefaultTimer.self
         )
+        retryStrategy.mock_nextRetryDelay.returns(0.01)
     }
 
     func test_concurrentAccess() {
@@ -167,7 +168,11 @@ final class AuthenticationRepository_Tests: XCTestCase {
 
         // Token Provider Failure
         let testError = TestError()
-        let provider: TokenProvider = { $0(.failure(testError)) }
+        var tokenCalls = 0
+        let provider: TokenProvider = {
+            tokenCalls += 1
+            $0(.failure(testError))
+        }
 
         let completionExpectation = expectation(description: "Connect completion")
         var receivedError: Error?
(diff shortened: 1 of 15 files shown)


alteral:
JWT tests just fly! @polqf, you rock!
Also quickly tested the authentication in the DemoApp, LGTM ✅

Tests/StreamChatTests/Repositories/AuthenticationRepository_Tests.swift
```diff
@@ -526,11 +570,8 @@ final class AuthenticationRepository_Tests: XCTestCase {
             completionExpectation.fulfill()
         })
 
-        let requestCompletion = try XCTUnwrap(apiClient.request_completion as? ((Result<GuestUserTokenPayload, Error>) -> Void))
-        requestCompletion(.failure(apiError))
-
         XCTAssertNotNil(repository.tokenProvider)
-        waitForExpectations(timeout: 0.1)
+        waitForExpectations(timeout: 10000)
```
martinmitrevski, line 574: why 10000? 😁

Tests/StreamChatTests/Repositories/AuthenticationRepository_Tests.swift
```diff
@@ -549,6 +590,9 @@ final class AuthenticationRepository_Tests: XCTestCase {
         let testError = TestError()
         connectionRepository.connectResult = .failure(testError)
 
+        // API Result
+        apiClient.test_mockResponseResult(Result<GuestUserTokenPayload, Error>.success(GuestUserTokenPayload(user: CurrentUserPayload.dummy(userId: "", role: .user), token: apiToken)))
```
martinmitrevski, line 594: We should format this one for better readability. Wonder why this was not done on the pre-commit hook?

Tests/StreamChatTests/Repositories/AuthenticationRepository_Tests.swift
```diff
@@ -580,6 +621,9 @@ final class AuthenticationRepository_Tests: XCTestCase {
         // Simulate Success on Connection Repository
         connectionRepository.connectResult = .success(())
 
+        // API Result
+        apiClient.test_mockResponseResult(Result<GuestUserTokenPayload, Error>.success(GuestUserTokenPayload(user: CurrentUserPayload.dummy(userId: "", role: .user), token: apiToken)))
```
martinmitrevski, line 625: Same here (formatting)

martinmitrevski (approved):
LGTM ✅ just few short things.

Tests/StreamChatTests/Repositories/AuthenticationRepository_Tests.swift
polqf: 😂 leftover

Issue #2343, opened 2022-10-23
Message Actions Popup doesn't work in thread reply after restarting app


VictorSemenchuk:
## What did you do?
Precondition: I'm using the default UI from StreamChatUI.
1. Open any channel and add reply to any message. Long press on the left reply - context menu displays as expected.
2. Restart the app.
3. Go to the channel which was used at the 1st step and find your reply. Long press on the reply - nothing happens.

## What did you expect to happen?
Context menu displays each type I make long press on a thread reply.

## What happened instead?
Context menu is not displayed if the reply is taken from cache if I make long press on a thread reply. 
Looks like `channel` is `nil` in `MessageDTO` took via `FRC`.

## GetStream Environment
**GetStream Chat version:** 4.22.0
**GetStream Chat frameworks:** StreamChat, StreamChatUI
**iOS version:** 16.0
**Swift version:** 5
**Xcode version:** 14.0.1
**Device:** iPhone 14 Pro

nuno-vieira:
Hi @VictorSemenchuk!

I was able to reproduce the issue. Thank you for the report! We will let know once we have a fix for this one!

Best,
Nuno

AlexanderFulcrum:
Stumbled on the same issue, can be pretty pesky for the user. Is there any ETA on the fix?

nuno-vieira:
> Stumbled on the same issue, can be pretty pesky for the user. Is there any ETA on the fix? 

Hi @AlexanderFulcrum , we don't have a ETA yet, we will let you know once we pick this one up.

nuno-vieira:
Hi @AlexanderFulcrum, @VictorSemenchuk!

Good news, this fix will be shipped in the next release :) 

We let you know once that happens, the ETA is this week, worst case scenario, next week. 

Best,
Nuno

polqf:
Hi! This has been fixed in our latest release [4.25.0](https://github.com/GetStream/stream-chat-swift/releases/tag/4.25.0)
Closing this issue for now, feel free to reopen it if you encounter any issues.

PR #2415, opened 2022-12-14
[CIS-2232] Fix message actions popup in cached thread replies

nuno-vieira:
### 🔗 Issue Links
CIS-2232
https://github.com/GetStream/stream-chat-swift/issues/2343

### 🎯 Goal
Fixes message actions popup not opening in thread replies, after restarting the app (loading replies from the local cache).

### 📝 Summary
When calling `didSelectMessageCell()` the router would not be called because `message.cid` would be `nil`. This should not happen, and replies from the local cache should have the `cid` property. For now, as a workaround, we will use the cid from the data source, so that it is more reliable. I spent quite a lot of time trying to understand why the channel is not attached to the thread reply in the local database, so for now, we will go with the workaround approach.

### 🧪 Manual Testing Notes
- Open a Channel
- Send a message
- Send a thread reply
- Open the thread
- Restart the app (To make sure the threads are loaded from local db)
- Go to the thread above
- Long press a thread reply
- It should open the message actions

### ☑️ Contributor Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Fix connecting user with non-expiring tokens (ex: development token) [#2393](https://github.com/GetStream/stream-chat-swift/pull/2393)
 - Fix crash when calling `addDevice()` from background thread [#2398](https://github.com/GetStream/stream-chat-swift/pull/2398)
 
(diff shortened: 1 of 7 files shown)


alteral:
For some reason, Emerge build on the previous build was auto-cancelled by Github. I've restarted [it](https://github.com/GetStream/stream-chat-swift/actions/runs/3641352041), so Emerge build here should pass on the next run.

PR #2433, opened 2023-01-04
[CIS-2241] Do not increment channel unread count for already processed messages

polqf:
### 🔗 Issue Links

https://stream-io.atlassian.net/browse/CIS-2241

### 🎯 Goal

Fix inconsistencies for channel unread counts

### 📝 Summary

There are edge cases in which we get channel updates through HTTP requests, but we then get events for new messages that we have already processed. In this cases, we might end up double counting those messages, and showing an incorrect unread count for those

### 🛠 Implementation

With this changes, for each WebSocket event that notifies of new messages, we are going to be checking if the message is already in database. If the message is already in our database, it means it was already processed, and therefore we don't need to update the unread count based on the event.
For the messages that we receive through WebSockets that are not yet in the database, the behaviour is kept the same. 

### 🧪 Manual Testing Notes

The best way to reproduce this in real world is:
1. Open the app in device A, wait until the sync is complete
2. Stay in the channel list
3. From another device (B), send messages to A, and observe how the unread counts work properly
4. Send A to background (make sure `staysConnectedInBackground` is set to `false`)
5. Open app in device A

Expected behaviour:
6. We fetch all the content, and the unread counts are kept the same

Old behaviour:
7. The messages that were sent during the previous session are DOUBLE COUNTED.
8. In most of the cases the unread count goes back to normal, but in some exceptions the wrong count was kept.

### ☑️ Contributor Checklist

### 🎁 Meme

[CIS-2241]: https://stream-io.atlassian.net/browse/CIS-2241?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### Notes

Ended up going for the count (vs fetch) approach.
// 200 objects ----
// Count:   0,00167 0,00199 0,0006
// Fetch:   0,00129 0,00189 0,0017

// 1000 objects ---
// Count:   0,0006 0,0007
// Fetch:   0,0007 0,0006
 Faster when having ~200, and same when as high as 1000.
In any case, in most of the cases when using the /sync endpoint we would not even reach 200


Sources/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -744,6 +744,17 @@ extension NSManagedObjectContext: MessageDatabaseSession {
 
     func message(id: MessageId) -> MessageDTO? { .load(id: id, context: self) }
 
+    func messageExists(id: MessageId) -> Bool {
+        let request = NSFetchRequest<MessageDTO>(entityName: MessageDTO.entityName)
+        request.predicate = NSPredicate(format: "id == %@", id)
+        do {
+            let count = try count(for: request)
+            return count != 0
```
polqf, line 752: Still to be tested whether this approach using `count` is better than a batch `fetch` using all the ids.
Theory says that individual `count` should be better 🤷

Sources/StreamChat/WebSocketClient/EventMiddlewares/ChannelReadUpdaterMiddleware.swift
```diff
@@ -59,12 +61,19 @@ struct ChannelReadUpdaterMiddleware: EventMiddleware {
...
+        session: DatabaseSession,
+        notificationCenter: EventNotificationCenter
     ) {
         guard let currentUser = session.currentUser else {
             return log.error("Current user is missing", subsystems: .webSocket)
         }
 
+        // If the message exists in the database before processing the current batch of events, it means it was
+        // already processed and we don't have to increase the unread count
+        guard notificationCenter.newMessageIds.contains(message.id) else {
```
polqf, line 73: In line 97, we increment the unread count for this particular channel.
What this line wants to avoid is for us to increment based on a message that was already in the database before processing this event. If it was already in the database, it means it was already processed, therefore we should not be altering the unread count

Sources/StreamChat/Database/DTOs/MessageDTO.swift
martinmitrevski: do you have some resource to share about the performance comparison between `count` and `fetch`?

Sources/StreamChat/WebSocketClient/EventMiddlewares/ChannelReadUpdaterMiddleware.swift
martinmitrevski: This should be the only place where we can show wrong information based on the double handling, right? All the other event middlewares process things that can't add up (e.g. typing, watching, etc).

martinmitrevski (commented):
Looks good ✅ One thing that we should verify is the performance of `count` when there are many messages in the events.
Also, we would need some tests and update the changelog.

Sources/StreamChat/Database/DTOs/MessageDTO.swift
polqf: Added to the description

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Fix truncated channels being moved to the bottom of the channel list [#2420](https://github.com/GetStream/stream-chat-swift/pull/2420)
 - Fix reactions not insantly updating when enforce unique is true [#2421](https://github.com/GetStream/stream-chat-swift/pull/2421)
 - Fix not being able to delete messages in `pendingSend` state [#2432](https://github.com/GetStream/stream-chat-swift/pull/2432)
+- Fix innacuracies in for channel unread count [#2433](https://github.com/GetStream/stream-chat-swift/pull/2433)
```
martinmitrevski, line 12: typo, should be "inaccuracies"

martinmitrevski (approved):
LGTM! ✅ Just a small typo in the changelog.

Sources/StreamChat/WebSocketClient/EventMiddlewares/ChannelReadUpdaterMiddleware.swift
nuno-vieira, line 73: @polqf Why don't we simply fetch the message from the `DatabaseSession` to check if it already exists? Why do we need the notification centre? I'm 100% following it 🤔

polqf: It does not work as there is another middleware saving the message to the db just before the unread count is processed... And I don't want to rely on the order of the middlewares to do this

Sources/StreamChat/WebSocketClient/EventMiddlewares/ChannelReadUpdaterMiddleware.swift
```diff
@@ -6,20 +6,22 @@ import Foundation
 
 /// A middleware which updates a channel's read events as websocket events arrive.
 struct ChannelReadUpdaterMiddleware: EventMiddleware {
-    func handle(event: Event, session: DatabaseSession) -> Event? {
+    func handle(event: Event, session: DatabaseSession, notificationCenter: EventNotificationCenter) -> Event? {
```
nuno-vieira, line 9: Ideally, we should not break the `EventMiddleware` interface and only add the notification center to this middleware, which is the only one who needs it. This, or provide a protocol or closure to fetch the `newMessageIds`.

This way we only need to change this middleware, and not all of them. This PR will be much smaller as a result of this. 

```suggestion
    weak var notificationCenter: EventNotificationCenter?
    init(notificationCenter: EventNotificationCenter) {
         self.notificationCenter = notificationCenter
    }
    
    func handle(event: Event, session: DatabaseSession) -> Event? {
```

CHANGELOG.md
polqf:
```suggestion
- Fix inaccuracies in for channel unread count [#2433](https://github.com/GetStream/stream-chat-swift/pull/2433)
```

PR #2446, opened 2023-01-11
Fix missing messages in thread pagination

nuno-vieira:
### 🔗 Issue Links
N/A

### 🎯 Goal
Fix the issue spotted in 4.26.0 QA, where some paginations would be missing from Threads.

### 🛠 Implementation
For now, we are disabling the batchSize implementation until we update how we refresh the objects which don't contain changes to minimize the risk of the release

### 🧪 Manual Testing Notes
1. Open Release 4.26.0 Channel
2. Open the first thread
3. Leave the thread
4. Comeback to the thread
5. All pages should be loaded

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/Controllers/MessageController/MessageController.swift
+++ b/Sources/StreamChat/Controllers/MessageController/MessageController.swift
@@ -300,18 +300,22 @@ public extension ChatMessageController {
             return
         }
 
-        let lastLocalMessageId: () -> MessageId? = {
-            self.replies.last { !$0.isLocalOnly }?.id
-        }
-
-        let lastMessageId = messageId ?? lastFetchedMessageId ?? lastLocalMessageId()
-
         let pageSize = limit ?? repliesPageSize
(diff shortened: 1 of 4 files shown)


Sources/StreamChat/Controllers/MessageController/MessageController.swift
```diff
@@ -300,18 +300,22 @@ public extension ChatMessageController {
             return
         }
 
-        let lastLocalMessageId: () -> MessageId? = {
```
polqf, line 303: Any reason why we got rid of this? I guess `lastFetchedMessageId` should be the right source of truth, right?

polqf (approved):

Sources/StreamChat/Controllers/MessageController/MessageController.swift
nuno-vieira: Yea exactly, if we don't have it, it means we are fetching the first page

PR #2459, opened 2023-01-20
WIP Custom attachments v2

nuno-vieira:
This is a proof of concept to make it easier to handle custom attachments on the customer side. 

#### How to use
This is what a customer would need to do to add support for async attachment uploads:
1. Create the custom attachment payload and make sure to implement `func updateRemoteUrl(url:)`
```swift
extension AttachmentType {
    static var workout: Self = "workout"
}

struct WorkoutAttachmentPayload: AttachmentPayload {
    static var type: AttachmentType = .workout

    var imageURL: URL?
    var caloriesBurned: Int

    mutating func updateRemoteUrl(_ url: URL) {
        imageURL = url
    }
}
```
2. Register the custom attachment payload in `ChatClient`
```swift
ChatClient.shared.registerCustomAttachment(with: WorkoutAttachmentPayload.self)
```

#### Implementation

The basic idea is to change the `typealias AnyChatMessageAttachment` from `ChatMessageAttachment<Data>` to `ChatMessageAttachment<Any>`. Where instead of storing the Payload as `Data`, the idea is to store it as a type of `Payload: AttachmentPayload`. With this change, we can add common properties/functions to the `AttachmentPayload` protocol and update the payload data as needed.

This requires quite a few changes internally, and it adds a little bit of complexity, but with the advantage of making it really easy to add support for new custom attachments for the customer, but also for our own new attachments in the future.

**The new flow of `ChatMessageAttachment.payload` -> `AttachmentDTO.payload`:**
```mermaid
flowchart LR
    Any:AttachmentPayload-- asDTO -->Data
```
**The new flow of  `AttachmentDTO.payload` -> `ChatMessageAttachment.payload`:**
```mermaid
flowchart LR
    Data --> |asModel| Resgistered{Payload Registered?}
    Resgistered -->|Yes| A[Decoded to AttachmentPayload]
    Resgistered -->|Fallback| D[Remains Data]
    A -->|whenFetched| B[Casted to AttachmentPayload]
    D -->|whenFetched| C[Decoded to AttachmentPayload]
```

The 2nd flow diagram is important since it describes the fallback in case the customer does not register their custom attachment payload. If this is the case, the Attachment Model continues having the payload as `Data` and whenever it is accessed it is decoded to the custom payload type.

This new approach even has performance improvements, at least when fetching the attachments from the models, since at this stage, we only do a casting, instead of decoding it every time.

#### Caveats
- In theory, it is a breaking change since we are updating the typealias `AnyChatMessageAttachment` from `ChatMessageAttachment<Data>` to  `ChatMessageAttachment<Any>`, but in practice, this should not have any impact on the customer code base, so for that reason, I think this minor "breaking change" is worth it.
- The implementation relies a lot on existential types, so it means we would bump the min swift version to 5.6, which we are already only supporting officially since we only compile the project with Xcode 13.4+. We could also hide this feature behind a swift 5.6 compiler flag, but I think it would add a lot of complexity.
- Using existential types, and relying on typecasting, indeed adds a bit of complexity and unsafeness, but as long as we make sure we only have 1 place for each transformation (Model <-> DTO), it should be fine.

### DIFF
--- a/Sources/StreamChat/Database/DTOs/AttachmentDTO.swift
+++ b/Sources/StreamChat/Database/DTOs/AttachmentDTO.swift
@@ -151,12 +151,24 @@ private extension AttachmentDTO {
 }
 
 extension AttachmentDTO {
+    static var decoders: [AttachmentType: (Data) throws -> Any] = [:]
+
+    static func register<Payload: AttachmentPayload>(_ payloadType: Payload.Type) {
+        decoders[Payload.type] = { data in
+            try JSONDecoder.stream.decode(Payload.self, from: data)
+        }
+    }
+
     /// Snapshots the current state of `AttachmentDTO` and returns an immutable model object from it.
     func asAnyModel() -> AnyChatMessageAttachment {
-        .init(
+        // If a decoder is registered, we decode here. This is mandatory for attachments which are updated automatically
+        // when the remote url is available. As a fallback, it will use Data, and will only be parsed in
+        // the attachment model using `attachment(payloadType:)` function.
+        let payload = (try? Self.decoders[attachmentType]?(data)) ?? data
+        return .init(
             id: attachmentID,
             type: attachmentType,
-            payload: data,
+            payload: payload,
             uploadingState: uploadingState
         )
     }
--- a/Sources/StreamChat/Models/Attachments/AnyAttachmentPayload.swift
+++ b/Sources/StreamChat/Models/Attachments/AnyAttachmentPayload.swift
@@ -6,9 +6,22 @@ import Foundation
 
 /// A protocol an attachment payload type has to conform in order it can be
 /// attached to/exposed on the message.
-public protocol AttachmentPayload: Codable {
+public protocol AttachmentPayload: Codable, Equatable {
     /// A type of resulting attachment.
     static var type: AttachmentType { get }
+
+    /// Update the url of the attachment with the remote one.
+    /// This function will be called once an attachment has successfully uploaded.
+    /// - Parameter url: The remote url of the attachment.
+    mutating func updateRemoteUrl(_ url: URL)
+}
+
+public extension AttachmentPayload {
+    mutating func updateRemoteUrl(_ url: URL) {
+        // Implement this function to update the remote URL of the attachment.
+        // This function will be called when the attachment was successfully uploaded,
+        // in case the attachment was added to a message lazily.
+    }
 }
 
 /// A type-erased type that wraps either a local file URL that has to be uploaded
@@ -105,7 +118,7 @@ public extension AnyAttachmentPayload {
             .flatMap { try JSONEncoder.stream.encode($0.asAnyEncodable) }
             .flatMap { try JSONDecoder.stream.decode([String: RawJSON].self, from: $0) }
 
-        let payload: AttachmentPayload
+        let payload: any AttachmentPayload
         switch attachmentType {
         case .image:
             payload = ImageAttachmentPayload(
--- a/Sources/StreamChatUI/ChatMessageList/ChatMessageListView+DiffKit.swift
+++ b/Sources/StreamChatUI/ChatMessageList/ChatMessageListView+DiffKit.swift
@@ -50,6 +50,35 @@ extension ChatMessage: Differentiable {
             && localState == source.localState
             && isFlaggedByCurrentUser == source.isFlaggedByCurrentUser
             && readBy.count == source.readBy.count
-            && allAttachments == source.allAttachments
+            && allAttachments.map(\.uploadingState) == source.allAttachments.map(\.uploadingState)
+            && areEqual(allAttachments.map(\.payload), source.allAttachments.map(\.payload))
+    }
+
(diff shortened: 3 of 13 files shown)


Sources/StreamChat/Database/DTOs/AttachmentDTO.swift
```diff
@@ -151,12 +151,24 @@ private extension AttachmentDTO {
...
+        }
+    }
+
     /// Snapshots the current state of `AttachmentDTO` and returns an immutable model object from it.
     func asAnyModel() -> AnyChatMessageAttachment {
-        .init(
+        // If a decoder is registered, we decode here. This is mandatory for attachments which are updated automatically
+        // when the remote url is available. As a fallback, it will use Data, and will only be parsed in
+        // the attachment model using `attachment(payloadType:)` function.
+        let payload = (try? Self.decoders[attachmentType]?(data)) ?? data
```
nuno-vieira, line 167: This is basically the flow of the 2nd diagram. The code is a bit hard to read as it stands, so I'll make it easier to read when cleaning it up.

Sources/StreamChat/Models/Attachments/AnyAttachmentPayload.swift
```diff
@@ -6,9 +6,22 @@ import Foundation
 
 /// A protocol an attachment payload type has to conform in order it can be
 /// attached to/exposed on the message.
-public protocol AttachmentPayload: Codable {
+public protocol AttachmentPayload: Codable, Equatable {
```
nuno-vieira, line 9: Maybe this could be a potential breaking change as well, if for some reason the custom attachment payload of the customer is not automatically equitable (very unlikely as well)

Sources/StreamChatUI/ChatMessageList/ChatMessageListView+DiffKit.swift
```diff
@@ -50,6 +50,35 @@ extension ChatMessage: Differentiable {
...
+            return other.isExactlyEqual(self)
+        }
+        return self == other
+    }
+
+    private func isExactlyEqual(_ other: any Equatable) -> Bool {
+        guard let other = other as? Self else {
+            return false
+        }
+        return self == other
```
nuno-vieira, line 82: After some research, this was the only was to compare to `any Equatable` types. Kinda odd that `any Equatable` are actually not comparable 😅

Sources/StreamChat/Models/Attachments/AnyAttachmentPayload.swift
```diff
@@ -6,9 +6,22 @@ import Foundation
...
     /// A type of resulting attachment.
     static var type: AttachmentType { get }
+
+    /// Update the url of the attachment with the remote one.
+    /// This function will be called once an attachment has successfully uploaded.
+    /// - Parameter url: The remote url of the attachment.
+    mutating func updateRemoteUrl(_ url: URL)
```
martinmitrevski, line 16: The method name's a bit confusing - it doesn't update the remote url, but the attachment url with the remote one.
Maybe sth like this would be better: `updateAttachmentUrl(_ remoteUrl: URL)`?

Sources/StreamChat/Models/Attachments/AnyAttachmentPayload.swift
```diff
@@ -6,9 +6,22 @@ import Foundation
...
+}
+
+public extension AttachmentPayload {
+    mutating func updateRemoteUrl(_ url: URL) {
```
martinmitrevski, line 20: What happens if a customer doesn't implement this method? How does this compare with the current behaviour?

Sources/StreamChat/Models/Attachments/AnyAttachmentPayload.swift
```diff
@@ -105,7 +118,7 @@ public extension AnyAttachmentPayload {
             .flatMap { try JSONEncoder.stream.encode($0.asAnyEncodable) }
             .flatMap { try JSONDecoder.stream.decode([String: RawJSON].self, from: $0) }
 
-        let payload: AttachmentPayload
+        let payload: any AttachmentPayload
```
martinmitrevski, line 121: I have some doubts about using existential types - might be too soon. Swift 5.6 is still new-ish, so not sure if we should start introducing them in the codebase.

Sources/StreamChatUI/ChatMessageList/ChatMessageListView+DiffKit.swift
martinmitrevski: That actually makes sense tbh, you need to make sure that the type is the same first.

martinmitrevski (commented):
Will provide more comments internally.

Sources/StreamChat/Models/Attachments/AnyAttachmentPayload.swift
nuno-vieira: Nothing happens

nuno-vieira:
Closing this one for now. We will re-evaluate later or when v5 comes.

PR #2462, opened 2023-01-25
Filter auto-resolving for Channel List Controller

polqf:
### 🔗 Issue Links

Resolves https://github.com/GetStream/ios-issues-tracking/issues/261

### 🎯 Goal

### 📝 Summary

### 🛠 Implementation

### 🎨 Showcase

| Before | After |
| ------ | ----- |
|  img   |  img  |

### 🧪 Manual Testing Notes

### ☑️ Contributor Checklist


Sources/StreamChat/Query/Filter.swift
```diff
@@ -148,22 +185,45 @@ public extension Filter {
     }
 }
 
+struct FilterAutoKey {
```
nuno-vieira, line 188: Nice to add documentation here too

Sources/StreamChat/Query/Filter.swift
```diff
@@ -148,22 +185,45 @@ public extension Filter {
...
+    let payloadKey: CodingKey
+    let dtoKey: String
+    let value: CVarArg? // isModerator is actually role == admin
```
nuno-vieira, line 191: Better to delete this comment and provide better documentation on why we need this 👍

Sources/StreamChat/Query/UserListQuery.swift
```diff
@@ -14,46 +14,112 @@ public class UserListFilterScope: FilterScope, AnyUserListFilterScope {}
 /// Non extra-data-specific filer keys for channel list.
 public extension FilterKey where Scope: AnyUserListFilterScope {
     /// A filter key for matching the `id` value.
-    static var id: FilterKey<Scope, UserId> { "id" }
+    static var id: FilterKey<Scope, UserId> {
+        .init(
+            payloadKey: UserPayloadsCodingKeys.id,
+            dtoKey: #keyPath(UserDTO.id)
+        )
+    }
```
nuno-vieira, line 22: Does it make sense in this PR to create the auto keys for all Controllers? Maybe it is a good idea to limit the scope for now to the Channel List, which is the most problematic one. Then we can create tickets to handle the other controllers (AFAIK we already have this for some, ex: members)

Sources/StreamChat/Query/Filter+Predicate.swift
```diff
@@ -0,0 +1,88 @@
...
+                subpredicates: valueArray.map {
+                    NSPredicate(format: "ANY \(key) == %@", $0.description)
+                }
+            )
+        case .notIn:
+            guard let key = dtoKey else { return unableToResolve() }
+            guard let valueArray = value as? NSArray else { return unableToResolve() }
+            return NSCompoundPredicate(format: "NONE \(key) IN %@", valueArray)
+        case .contains:
+            return makeComparablePredicate("IN")
```
nuno-vieira, line 63: I'm not sure this is quite true. The original implementation was this:
```swift
guard let key = dtoKey else { return nil }
guard let valueArray = value as? [CVarArg] else { return nil }
return NSPredicate(format: "\(key) IN %@", valueArray)
```

So the value needs to be `[CVarArg]`. Did you test if this produced the same result?

Sources/StreamChat/Query/Filter+Predicate.swift
```diff
@@ -0,0 +1,88 @@
...
+            return makeCompoundPredicate(.and)
+        case .or:
+            return makeCompoundPredicate(.or)
+        case .nor:
+            guard let value = self.value as? [Filter] else { return unableToResolve() }
+            let predicates = value.compactMap {
+                $0.predicate.map(NSCompoundPredicate.init(notPredicateWithSubpredicate:))
+            }
+            guard !predicates.isEmpty else { return unableToResolve() }
+            return NSCompoundPredicate(type: .and, subpredicates: predicates)
```
nuno-vieira, line 79: Isn't this the same thing? (I actually don't know, didn't test this myself)
```suggestion
           return makeCompoundPredicate(.not)
```

Sources/StreamChat/Query/Filter+Predicate.swift
```diff
@@ -0,0 +1,88 @@
...
+        case .query:
+            guard let key = dtoKey else { return unableToResolve() }
+            let valueToUse: CVarArg = autoKey?.value ?? value.description
+            return NSPredicate(format: "\(key) CONTAINS %@", valueToUse)
+        case .autocomplete:
+            return unableToResolve()
+        }
```
nuno-vieira, line 86: After talking with the backend team, the `.query` I don't think we should support it, we should just return `nil` in this case. This is basically used for search, and in this case, the automatic filter has no use. Besides that, the `contains` is a pretty naive implementation which does not replicate what the server is doing. 

On the other hand, `autocomplete` we can actually support it, and this one is useful. What we should use is the `PREFIX` statement in the NSPredicate. 👍

docusaurus/docs/iOS/client/controllers/channels.md
```diff
@@ -54,28 +54,43 @@ let compoundFilter: Filter<ChannelListFilterScope> = .and([
...
+#### :warning: When the filter cannot be auto-resolved:
+
+In certain edge cases, it can be that we cannot resolve at runtime the filter you passed in the query. In that case, you will either see an assertion or a log in the console pointing you to this documentation.
+
+We have a fallback that can be used when this happens. The method that is used to create a `ChatChannelListController` also accepts a `filter` block
+```swift
+let client = ChatClient.shared
+let controller = client.channelListController(query: query, filter: { channel in
+    // The filtering logic goes here.
+    return true/false
```
nuno-vieira, line 80: We should provide a concrete example in this case, otherwise, customers are really not sure what to expect and what to write. We need to provide a real example, with a real query, and a real closure.

docusaurus/docs/iOS/client/controllers/channels.md
```diff
@@ -54,28 +54,43 @@ let compoundFilter: Filter<ChannelListFilterScope> = .and([
...
 
 When `filter` returns `false`, the new/updated channel is not linked/is unlinked from the channel list.
 
 :::note
-It is expected that the logic on the `filter` closure matches the one in `query.filter`.
-:::
-
-:::note
-Given the complex nature of `Filter`, having generic logic capable of using `query.filter` at runtime to determine if a channel meets the criteria is extremely expensive and complex. Instead, having a block that checks for the particular needs of the use case is much cheaper.
+It is expected that the logic on the `filter` closure matches exactly the one in `query.filter`.
```
nuno-vieira, line 93: This is more or less true. If, for example, a customer uses filters that we can automatically resolve, but it also has extra data fields, this means the customer can only write the block for the extra data logic, and not for everything. Should we clarify this and provide 2 examples I guess? In fact, the customer only needs to apply the filter block for the properties which are printed out in the log 🤔 But maybe this is too confusing? And maybe it is just easier to try and replicate the whole query if it doesn't work I guess?

Sources/StreamChat/Query/ChannelMemberListQuery.swift
```diff
@@ -13,38 +13,69 @@ public class MemberListFilterScope: UserListFilterScope, AnyMemberListFilterScop
...
 public extension FilterKey where Scope: AnyMemberListFilterScope {
+    #warning("Tricky! Test filter generation for BE as well as predicate generation")
     /// A filter key for matching moderators of a channel.
     /// Supported operators: `equal`, `notEqual`
-    static var isModerator: FilterKey<Scope, Bool> { "is_moderator" }
+    static var isModerator: FilterKey<Scope, Bool> {
+        .init(
+            payloadKey: MemberPayload.CodingKeys.role,
+            dtoKey: #keyPath(MemberDTO.channelRoleRaw)
+//            value: MemberRole.moderator.rawValue
```
ipavlidakis, line 23: nit: Can we remove this if it's not needed?

Sources/StreamChat/Query/Filter+Predicate.swift
```diff
@@ -0,0 +1,92 @@
...
+
+extension Filter {
+    var predicate: NSPredicate? {
+        guard let op = FilterOperator(rawValue: self.operator) else { return nil }
+
+        let unableToResolve: () -> NSPredicate? = {
+            let message = """
+            The filter passed to the query could not be resolved at runtime.
+            Please check the documentation at: https://getstream.io/chat/docs/sdk/ios/client/controllers/channels/#%EF%B8%8F-when-the-filter-cannot-be-auto-resolved
+            Otherwise you can have issues when multiples channel list controllers are alive at the same time
```
ipavlidakis, line 15: What do you think about rephrasing it to:

> ... Otherwise you may face issues when multiple channel list controllers are alive at the same time ...

Sources/StreamChat/Query/Filter+Predicate.swift
```diff
@@ -0,0 +1,92 @@
...
+            """
+            log.assertionFailure(message)
+            return nil
+        }
+
+        let dtoKey = autoKey?.dtoKey ?? key
+
+        let makeComparablePredicate: (_ compareToken: String) -> NSPredicate? = { compareToken in
+            guard let key = dtoKey else { return unableToResolve() }
```
ipavlidakis, line 24: nit: I think that keeping the name as `dtoKey` will avoid confusion with the key outside the context

Sources/StreamChat/Query/Filter+Predicate.swift
```diff
@@ -0,0 +1,92 @@
...
+            return NSPredicate(format: "\(key) \(compareToken) %@", argumentArray: [value])
+        }
+
+        let makeCompoundPredicate: (_ logicalType: NSCompoundPredicate.LogicalType) -> NSPredicate? = { logicalType in
+            guard let value = self.value as? [Filter] else { return unableToResolve() }
```
ipavlidakis, line 29: `self` is strongly captured here, which become dangerous. Maybe consider capturing self weakly. That should change much for the code as it already guards against nil or for an invalid type

Sources/StreamChat/Query/Filter+Predicate.swift
ipavlidakis: Another reason why to choose `IN` is a potential performance benefit([reference](https://stackoverflow.com/a/3074731)). SQL seems that uses binary search for IN while for AND/OR statements will evaluate them one by one.

docusaurus/docs/iOS/client/controllers/channels.md
```diff
@@ -54,28 +54,43 @@ let compoundFilter: Filter<ChannelListFilterScope> = .and([
...
+We have a fallback that can be used when this happens. The method that is used to create a `ChatChannelListController` also accepts a `filter` block
+```swift
+let client = ChatClient.shared
+let controller = client.channelListController(query: query, filter: { channel in
+    // The filtering logic goes here.
+    return true/false
 })
 ```
 
+This logic on the block should match the one in  `query.filter`
```
ipavlidakis, line 84: typo: `The logic in the block should ...`

Sources/StreamChat/Query/ChannelMemberListQuery.swift
nuno-vieira: This should not be commented out, actually 🤔 Any reason why it is now? @polqf

PR #2474, opened 2023-02-02
Replace ChannelListVC with SplashVC on startup

ipavlidakis:
### 🔗 Issue Links

[Github Ticket](https://github.com/GetStream/ios-issues-tracking/issues/307)

### 🎯 Goal

Fix crash on Example Apps.

### 📝 Summary

The SplashViewController acts as a mediator, where a simple loading indicator is presented to the user, while we are connecting the demo user with the ChatClient.

### 🛠 Implementation

On successful completion of user's connection, the window is updated with the channelList. 
On failure an error message will be presented with the options either to Cancel or Retry. On Retry the connectUser flow will be triggered again, while on Cancel the screen will be updated with some information about the not connected user and action to trigger the connectUser flow.

### 🧪 Manual Testing Notes

- Run any of the Example Apps
- The app should run successfully with no crash on startup

### ☑️ Contributor Checklist

### 🎁 Meme
meme


Examples/MessengerClone/SceneDelegate.swift
```diff
@@ -11,14 +11,57 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
...
+        activityIndicatorView.startAnimating()
+        activityIndicatorView.center = view.center
+        view.addSubview(activityIndicatorView)
+        
+        ChatClient.shared.connectUser(
+            userInfo: userInfo,
+            token: token
+        ) { [weak self] error in
+            if let error {
+                fatalError("Failed to connect user.(\(error.localizedDescription)")
```
martinmitrevski, line 59: We shouldn't crash if there's an error. Would be nicer to show an alert.

Examples/MessengerClone/SceneDelegate.swift
```diff
@@ -11,14 +11,57 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
...
+            window.rootViewController = SplashViewController { [unowned window] in
+                window.rootViewController = UIHostingController(rootView: MessengerChatChannelList())
+            }
             self.window = window
             window.makeKeyAndVisible()
         }
     }
 }
+
+private final class SplashViewController: UIViewController {
```
martinmitrevski, line 26: We could add the same `SplashViewController` to multiple targets, instead of copying the code.

martinmitrevski (commented):
Nice start 👍 We just need to make sure that we're not crashing when connect fails and also try to avoid code duplication for the splash VC.

Examples/MessengerClone/SceneDelegate.swift
ipavlidakis: What is the best place to have shared components like this one?

martinmitrevski: We can add a `Shared` folder under `Examples` and put the common code there.

Examples/Shared/SplashViewController.swift
```diff
@@ -0,0 +1,79 @@
...
+    private func presentAlert(
+        with error: Error
+    ) {
+        let alertController = UIAlertController(
+            title: "Error",
+            message: error.localizedDescription,
+            preferredStyle: .alert
+        )
+        
+        alertController.addAction(.init(
```
martinmitrevski, line 64: Just one more thing here: If the connection fails, and the user cancels, we would end up with endless spinner. Maybe just switch the spinner to a label with an error message if that happens?

ipavlidakis: Updated with one step further. Added a message and a button to restart the connectUser flow. Let me know your thoughts

Examples/Shared/SplashViewController.swift
```diff
@@ -0,0 +1,125 @@
...
+
+final class SplashViewController: UIViewController {
+    private let userInfo: UserInfo
+    private let token: Token
+    private let completionHandler: () -> Void
+    
+    private lazy var activityIndicatorView: UIActivityIndicatorView = .init(style: .large)
+    private lazy var cancellationLabel: UILabel = {
+        let label = UILabel()
+        label.text = "No connected use found.\nTap the button below to try connect the user with id \(userInfo.id)."
```
martinmitrevski, line 17: "use" -> "user"

Examples/Shared/SplashViewController.swift
martinmitrevski: great, even better 👍

martinmitrevski (approved):
Nice work 👍  Just one typo in the label, no need for re-review.

Examples/Shared/SplashViewController.swift
ipavlidakis: Good catch. Updated and i'll open it for review once the checks are finished. Thanks :)

nuno-vieira (approved):
LGTM! 🎉

PR #2476, opened 2023-02-06
[Fix]Crash on Demo when user creates channel

ipavlidakis:
### 🎯 Goal

The app is crashing due to a race condition that is happening while the tableView updates its contents and the SearchController updates its users state.

We want to fix the race condition and effectively update the UI without crashing.

### 📝 Summary

When the user is trying to create a new channel and type back and forth queries on the search box the Demo app is crashing.

### 🛠 Implementation

The issue is being caused because the SearchController state (that asynchronously fetches users) is the same as the UI state in the CreateChatViewController. The solution to the problem is to decouple the 2 states and only rely on the delegate calls from the SearchController to update the users state in the CreateChatViewController and thus update the UI.

### 🧪 Manual Testing Notes

1. Open the Create Channel
2. Wait until the user list is populated
3. Type something in the search box
4. While the activityIndicator is visible, delete the text in the search box
5. Repeat

### ☑️ Contributor Checklist

### 🎁 Meme

meme

### DIFF
--- a/DemoApp/Screens/Create Chat/CreateChatViewController.swift
+++ b/DemoApp/Screens/Create Chat/CreateChatViewController.swift
@@ -72,9 +72,7 @@ class CreateChatViewController: UIViewController {
 
     var searchController: ChatUserSearchController!
 
-    var users: [ChatUser] {
-        searchController.userArray
-    }
+    var users: [ChatUser] = []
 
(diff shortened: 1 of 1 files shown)


martinmitrevski (approved):
LGTM ✅

PR #2488, opened 2023-02-10
Implement Local Filter processing for ChannelListController

ipavlidakis:
### 🔗 Issue Links

Resolves https://github.com/GetStream/ios-issues-tracking/issues/261

### 🎯 Goal

Allow local filter processing in order to reduce the need for a local filter block in ChannelListController.

### 🛠 Implementation

This PR implements the following:
- Implement `keyPathValueProvider` for FilterKeys
- Update the ChannelList related filterKeys, that can be filtered locally, to provide a valid `keyPathValueProvider`
- Translation of API filters to DB filters (NSPredicates)
- ChannelDTO fetchRequest creation was updated to include the filter's predicate - if not nil
- Update tests

The logic is being described below:
FilterKeys that are compatible with local filter should provide a `keyPathValueProvider`. `KeyPathValueProvider` are closures that provide the keypath for the requested value or return nil if the value cannot be mapped.

The fetchRequest that the ChannelDTO, will include the filter's predicate if it's not nil. The fetchRequest can then be used to query the DB/

### 🧪 Manual Testing Notes
**- Create a new Channel**
1. Open 2 Simulators and login with different users
2. Create a new group with the 2 users + any other user you may need to add in order to ensure that a new group will be created
3. Send a message
4. The new channel should appear at the top of the list in both simulators

**- Update an existing Channel**
1. Open 2 Simulators and login with different users
2. Open an old existing channel with both our users
3. Send a message
5. The new channel should appear at the top of the list in both simulators

**- Public API Changes**
- `StreamRuntimeCheck.isChannelLocalFilteringEnabled: Bool` (default: `true`): When the flag is `true` the ChannelDTO will use the filter's predicate to create the fetchRequest.

**- Docusaurus Update Preview**

Screenshot 2023-02-16 at 4 45 58 PM

### ☑️ Contributor Checklist

### 🎁 Meme

meme


Sources/StreamChat/Query/ChannelListQuery.swift
```diff
@@ -42,9 +42,12 @@ extension Filter where Scope: AnyChannelListFilterScope {
 }
 
 // We don't want to expose `members` publicly because it can't be used with any other operator
-// then `$in`. We expose it publicly via the `containMembers` filter helper.
+// than `$in`. We expose it publicly via the `containMembers` filter helper.
+// TODO: should it be using the ChannelPayload.members or ChannelPayload.channel.members
 extension FilterKey where Scope: AnyChannelListFilterScope {
-    static var members: FilterKey<Scope, UserId> { "members" }
+    static var members: FilterKey<Scope, UserId> { .init(rawValue: "members") {
+        ($0 as? ChannelPayload)?.members.map(\.userId)
```
ipavlidakis, line 49: Is this the right keypath for the members or shall we instead get from the `($0 as? ChannelPayload)?.channel.members.map(\.userId)`?

Sources/StreamChat/Query/ChannelListQuery.swift
```diff
@@ -58,80 +61,127 @@ public enum InviteFilterValue: String, FilterValue {
...
+    static var deletedAt: FilterKey<Scope, Date> { .init(rawValue: "deleted_at") {
+        ($0 as? ChannelPayload)?.channel.deletedAt
+    }}
 
     /// A filter key for querying hidden channels.
     /// Supported operators: `equal`
-    static var hidden: FilterKey<Scope, Bool> { "hidden" }
+    // TODO: should it be using the ChannelPayload.isHidden or ChannelPayload.channel.isHidden
+    static var hidden: FilterKey<Scope, Bool> { .init(rawValue: "hidden") {
+        ($0 as? ChannelPayload)?.isHidden
```
ipavlidakis, line 130: Is this the right keypath for the members or shall we instead get from the `($0 as? ChannelPayload)?.channel.isHidden`?

Sources/StreamChat/Query/ChannelListQuery.swift
```diff
@@ -58,80 +61,127 @@ public enum InviteFilterValue: String, FilterValue {
...
-    static var team: FilterKey<Scope, TeamId?> { "team" }
+    static var team: FilterKey<Scope, TeamId?> { .init(rawValue: "team") {
+        ($0 as? ChannelPayload)?.channel.team
+    }}
 
     /// Filter for checking whether current user is joined the channel or not (through invite or directly)
     /// Supported operators: `equal`
-    static var joined: FilterKey<Scope, Bool> { "joined" }
+    static var joined: FilterKey<Scope, Bool> { .init(rawValue: "joined") {
+        ($0 as? ChannelPayload)?.membership?.role == .owner
```
ipavlidakis, line 155: Is this the right way to check if the current user has joined the channel

Sources/StreamChat/Query/ChannelListQuery.swift
```diff
@@ -58,80 +61,127 @@ public enum InviteFilterValue: String, FilterValue {
...
+            || ($0 as? ChannelPayload)?.membership?.inviteAcceptedAt != nil
+    }}
 
     /// Filter for checking whether current user has muted the channel
     /// Supported operators: `equal`
+    ///
+    // TODO: How can we filter that locally?
     static var muted: FilterKey<Scope, Bool> { "muted" }
```
ipavlidakis, line 163: What is the best way to get this one?

Sources/StreamChat/Query/ChannelListQuery.swift
```diff
@@ -58,80 +61,127 @@ public enum InviteFilterValue: String, FilterValue {
...
 
     /// Filter for checking the status of the invite
     /// Supported operators: `equal`
+    ///
+    // TODO: How can we filter that locally?
     static var invite: FilterKey<Scope, InviteFilterValue> { "invite" }
```
ipavlidakis, line 169: What is the best way to get this one?

Sources/StreamChat/Query/ChannelListQuery.swift
```diff
@@ -58,80 +61,127 @@ public enum InviteFilterValue: String, FilterValue {
...
+    // TODO: should it be using the ChannelPayload.members or ChannelPayload.channel.members
+    static var memberName: FilterKey<Scope, String> { .init(rawValue: "member.user.name") {
+        ($0 as? ChannelPayload).map { $0.members.compactMap { $0.user?.name } }
+    }}
 
     /// Filter for the time of the last message in the channel. If the channel has no messages, then the time the channel was created.
     /// Supported operators: `equal`, `greaterThan`, `lessThan`, `greaterOrEqual`, `lessOrEqual`, `notEqual`
-    static var lastUpdatedAt: FilterKey<Scope, Date> { "last_updated" }
+    static var lastUpdatedAt: FilterKey<Scope, Date> { .init(rawValue: "last_updated") {
+        ($0 as? ChannelPayload)?.channel.lastMessageAt
```
ipavlidakis, line 183: Is this the right way to get lastUpdated as it's not available?

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
```diff
@@ -351,10 +351,14 @@ extension ChannelDTO {
         // We can't pass bools directly to NSPredicate so we have to use integers
         let isHidden = NSPredicate(format: "isHidden == %i", query.filter.hiddenFilterValue == true ? 1 : 0)
 
-        let subpredicates = [
+        var subpredicates: [NSPredicate] = [
             matchingQuery, notDeleted, isHidden
         ]
 
+        if let filterPredicate = query.filter.predicate {
```
ipavlidakis, line 358: What's the best way to allow that be configured and in case it's causing issues, customers can disable it?

Tests/StreamChatTests/Controllers/ChannelListController/ChannelListController_Tests.swift
```diff
@@ -1095,6 +1092,268 @@ final class ChannelListController_Tests: XCTestCase {
...
+
+        try assertFilterPredicate(
+            .notIn(.members, values: [memberId1, memberId2]),
+            channelsInDB: [
+                .dummy(channel: .dummy(cid: cid1, members: [.dummy(), .dummy()])),
+                .dummy(channel: .dummy(cid: cid2, members: [.dummy()])),
+                .dummy(channel: .dummy(members: [.dummy(user: .dummy(userId: memberId1)), .dummy(user: .dummy(userId: memberId2))]))
+            ],
+            expectedResult: [cid1, cid2]
+        )
```
polqf, line 1263: is this working? Would be nice to member1 to cid1 and member2 to cid2

Sources/StreamChat/Query/Filter+ChatChannel.swift
```diff
@@ -13,3 +13,194 @@ extension ChatChannel {
...
+        case .in where value is [FilterValue]:
+            guard
+                let filterArray = (value as? [FilterValue])
+            else {
+                return nil
+            }
+            return NSCompoundPredicate(
+                andPredicateWithSubpredicates: filterArray.map { subValue in
+                    NSPredicate(
+                        format: "%@ IN %K",
```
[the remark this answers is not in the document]
polqf: Oh wait, was the issue in my implementation that we were using %@ instead of %K? 🤦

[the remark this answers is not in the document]
ipavlidakis: Exactly ;) %K does the trick

Sources/StreamChat/Query/Filter+ChatChannel.swift
```diff
@@ -13,3 +13,194 @@ extension ChatChannel {
...
+                return nil
+            }
+            return NSCompoundPredicate(
+                andPredicateWithSubpredicates: filterArray.map { subValue in
+                    NSPredicate(
+                        format: "NOT(%@ IN %K)",
+                        argumentArray: [subValue, keyPathValueProvider()]
+                    )
+                }
+            )
```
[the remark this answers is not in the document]
ipavlidakis: Yeap :)

Tests/StreamChatTests/Controllers/ChannelListController/ChannelListController_Tests.swift
```diff
@@ -1095,6 +1092,268 @@ final class ChannelListController_Tests: XCTestCase {
         XCTAssert(client.activeChannelListControllers.count == 1)
         XCTAssert(client.activeChannelListControllers.allObjects.first === controller)
     }
+
+    // MARK: Predicates
+
+    private func assertFilterPredicate(
```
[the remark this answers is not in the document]
ipavlidakis: Because before, we didn't have any local filtering, so we were accepting whatever the server was sending us (in this case the mocked data). 
Now, that the local filtering is in place for the data to actually be retrieved they need to match the filter.

In those cases here, the used filter is checking that the user is a member of those channels. So, my changes are making sure that every channel that is being tested, has the required user as a member

Tests/StreamChatTests/Controllers/ChannelListController/ChannelListController_Tests.swift
```diff
@@ -1095,6 +1092,268 @@ final class ChannelListController_Tests: XCTestCase {
...
+        _ = try waitFor { [unowned client] in
+            client?.databaseContainer.write({ [query] session in
+                try channelsInDB().forEach { payload in
+                    try session.saveChannel(payload: payload, query: query, cache: nil)
+                }
+            }, completion: $0)
+        }
+
+        // Assert the resulting value is updated
+        AssertAsync.willBeEqual(
```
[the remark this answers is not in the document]
ipavlidakis: I just used it because this is what the rest of the tests are using. What would be the recommended way?

Tests/StreamChatTests/Controllers/ChannelListController/ChannelListController_Tests.swift
ipavlidakis: Yeap this is working as expected

Tests/StreamChatTests/Controllers/ChannelListController/ChannelListController_Tests.swift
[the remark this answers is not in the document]
ipavlidakis: This is usually my approach, but on this specific file I noticed that `private` functions have been recently added close to their test cases and that's why i have this one here

Sources/StreamChat/Query/Filter+ChatChannel.swift
```diff
@@ -13,3 +13,194 @@ extension ChatChannel {
...
+    }
+
+    private func logicalPredicate(
+        _ operator: FilterOperator
+    ) -> NSPredicate? {
+        switch `operator` {
+        case .and where value is [Filter<Scope>]:
+            guard
+                let filters = value as? [Filter<Scope>]
+            else {
```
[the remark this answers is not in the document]
ipavlidakis: Good to know. Do we have a link where i can take a look on what the code-style is?

Tests/StreamChatTests/Controllers/ChannelListController/ChannelListController_Tests.swift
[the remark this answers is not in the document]
polqf: Yeah, it is still pretty spread out around the codebase. We usually go for expectations for new code

Sources/StreamChat/Query/Filter+ChatChannel.swift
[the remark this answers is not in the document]
ipavlidakis: > This might not work as %K is usually not a top-level property

I'm not sure what you mean by this. `%K` represents keypaths and doesn't seem to have any limitations on what level this keypath is

Tests/StreamChatTests/Controllers/ChannelListController/ChannelListController_Tests.swift
polqf: I think having member1 in cid1 and member2 in cid2 can be a good test so we make sure it not treated as an "OR", but nothing else aside from that

Tests/StreamChatTests/Controllers/ChannelListController/ChannelListController_Tests.swift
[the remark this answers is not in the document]
ipavlidakis: Aaah great. Let me try updating them

Sources/StreamChat/Query/Filter+ChatChannel.swift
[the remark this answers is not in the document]
polqf: I might have spent hours trying other things, and that was it. Great 😂

Sources/StreamChat/Query/Filter+ChatChannel.swift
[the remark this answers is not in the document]
polqf: We don't, at the moment, but it is something we've discussed we should do.

Sources/StreamChat/Query/Filter+ChatChannel.swift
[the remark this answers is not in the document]
ipavlidakis: It was literally after had spent hours on documentation and trials and couldn't figure out why it doesn't work, that i read this one 
> If you want to specify a dynamic property name, use %K in the format string, as shown in the following example.
[Apple Docs](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/Predicates/Articles/pSyntax.html)

Sources/StreamChat/Config/StreamRuntimeCheck.swift
```diff
@@ -14,4 +14,9 @@ public enum StreamRuntimeCheck {
     ///
     ///  Enables background mapping of DB models
     public static var _isBackgroundMappingEnabled = false
+
+    /// Enable/Disable local filtering for Channel lists. When enabled,
+    /// the `ChannelDTO` will include the filter's predicate (if available)
+    /// in the fetchRequest.
+    public static var isChannelLocalFilteringEnabled = true
```
nuno-vieira, line 21: I'm not sure this one is a candidate for `StreamRuntimeCheck` 🤔 We already have a `ChatClientConfig` which IMO it is a better candidate for this. CC @polqf @martinmitrevski

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
```diff
@@ -351,10 +351,14 @@ extension ChannelDTO {
         // We can't pass bools directly to NSPredicate so we have to use integers
         let isHidden = NSPredicate(format: "isHidden == %i", query.filter.hiddenFilterValue == true ? 1 : 0)
 
-        let subpredicates = [
+        var subpredicates: [NSPredicate] = [
             matchingQuery, notDeleted, isHidden
         ]
 
+        if StreamRuntimeCheck.isChannelLocalFilteringEnabled, let filterPredicate = query.filter.predicate {
```
nuno-vieira, line 358: If we use the `ChatClientConfig`, we can pass this in the `channelListFetchRequest()` method. There's no reason for this to be static. CC @polqf @martinmitrevski 

We currently have config flags spread in multiple places: `Components`, `ChatClientConfig` and `StreamRuntimeCheck`. The `Components` should be used for UI (Although it deserves a different struct as well, right now it handles too much), and the `ChatClientConfig` should be used for LLC configs. The `StreamRuntimeCheck` should be unnecessary, I actually thought this was a private API 🤔

Sources/StreamChat/Query/ChannelListQuery.swift
```diff
@@ -58,67 +58,67 @@ public enum InviteFilterValue: String, FilterValue {
...
-    static var id: FilterKey<Scope, String> { "id" }
+    static var id: FilterKey<Scope, String> { .init(rawValue: "id", keyPathValueProvider: { "cid" }) }
 
     /// A filter key for matching the `name` value.
-    static var name: FilterKey<Scope, String> { "name" }
+    static var name: FilterKey<Scope, String> { .init(rawValue: "name", keyPathValueProvider: { "name" }) }
 
     /// A filter key for matching the `image` value.
-    static var imageURL: FilterKey<Scope, URL> { "image" }
+    static var imageURL: FilterKey<Scope, URL> { .init(rawValue: "image", keyPathValueProvider: { "imageURL" }) }
```
nuno-vieira, line 73: In the initial implementation, we used something like this: `#keyPath(ChannelDTO.name)` to avoid committing mistakes, is there any reason we are not following this approach?

Sources/StreamChat/Query/Filter+ChatChannel.swift
```diff
@@ -4,12 +4,190 @@
...
+    /// The predicate will be automatically be used
+    /// by the ChannelDTO to create the
+    /// fetchRequest.
+    ///
+    /// - Important:
+    /// The behaviour of the ChannelDTO, to include or not
+    /// the predicate in the fetchRequest, it's controlled by
+    /// `StreamRuntimeCheck.isChannelLocalFilteringEnabled`
+    var predicate: NSPredicate? {
+        guard let `operator` = FilterOperator(rawValue: `operator`) else {
```
nuno-vieira, line 24: Nit: In this case we could have used just "op" to avoid always typing `` whenever we used this variable

Sources/StreamChat/Query/Filter.swift
```diff
@@ -111,23 +113,47 @@ public struct Filter<Scope: FilterScope> {
     ///   - key: The "left-hand" side of the filter. Specifies the name of the field the filter should match.
     ///   - value: The "right-hand" side of the filter. Specifies the value the filter should match.
     ///
-    public init(operator: String, key: String?, value: FilterValue) {
+    public init(
+        operator: String,
+        key: String?,
+        value: FilterValue,
+        keyPathValueProvider: (() -> String)? = nil
```
nuno-vieira, line 120: We should add init docs for the the new argument `keyPathValueProvider` 👍

Sources/StreamChat/Query/Filter.swift
```diff
@@ -99,6 +99,8 @@ public struct Filter<Scope: FilterScope> {
     /// The "right-hand" side of the filter. Specifies the value the filter should match.
     public let value: FilterValue
 
+    public let keyPathValueProvider: (() -> String)?
```
nuno-vieira, line 102: What is the reason for this to be a closure? Why isn't it just a `String`? 

Also, I think we should make this internal for now and not public since the goal is only for us to provide this automatic behaviour. It will be easy for customers to commit mistakes, especially because they don't have access to the DTOs. 

Another thing that seems to be missing, is the option to override a value. In some cases, in server, we have booleans like "isModerator", but in our data, it translates to an enum. 

This is why the original implementation was like this:
https://github.com/GetStream/stream-chat-swift/blob/36e4e8e5679832238bba0f271185125c8179606f/Sources/StreamChat/Query/Filter.swift#L187-L190

We had a payloadKey, a dtoKey, and a possibility to override the value. Since a payloadKey is a CodingKey it also provided some safety, and the dtoKey would always use the #keyPath to avoid mistakes, especially if we change the DTO models in the future.

Sources/StreamChat/Config/StreamRuntimeCheck.swift
martinmitrevski: I agree with Nuno here

Sources/StreamChat/Query/Filter.swift
ipavlidakis: > Another thing that seems to be missing, is the option to override a value. In some cases, in server, we have booleans like "isModerator", but in our data, it translates to an enum
I'm not really sure what you mean with this one. Let's catch up tomorrow about it

Sources/StreamChat/Query/Filter.swift
ipavlidakis: I moved a little bit the parts here in order to hide the keypath and not allow the customers to pass in a value (for now at least). Let me know your thoughts

Sources/StreamChat/Config/StreamRuntimeCheck.swift
ipavlidakis: Great, I'll move it to the `ChatClientConfig` then

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Add `UploadedAttachmentPostProcessor` in `ChatClientConfig` to allow changing custom attachment payloads after an attachment has been uploaded [#2457](https://github.com/GetStream/stream-chat-swift/pull/2457)
 - Add `AnyAttachmentPayload(localFileURL:customPayload:)` initializer to allow creating custom attachments without a remote URL [#2457](https://github.com/GetStream/stream-chat-swift/pull/2457)
 - Add skip push support when sending a message [#2486](https://github.com/GetStream/stream-chat-swift/pull/2486)
+- Add `isChannelLocalFilteringEnabled` in `ChatClientConfig` to allow changing whether the Channels in ChannelList will be locally filtered [#2488](https://github.com/GetStream/stream-chat-swift/pull/2488)
```
nuno-vieira, line 11: I think that `isChannelAutomaticFilteringEnabled` is a bit more suggestive for customers 🤔 Since "Local" is kinda an implementation detail and doesn't actually express what the goal of it is

nuno-vieira (commented):
Looking pretty good 👌 

The only thing missing is that we need docs to explain this flag. The original issue was actually about just improving the docs around this area. And the previous PR already had some changes: https://github.com/GetStream/stream-chat-swift/pull/2462/files#diff-21367a01e6d6533c0dff9e65bef56dbe0eefd81e2c4740290047cdd0cdf15e7aR57

Ideally, we want to touch on these points:
- By default, the SDK will try to automatically filter new channels created
- If extra data properties are used in query filters, they need to be manually filtered in the filter closure
- If the customer creates custom filters, they also need to be filtered by the closure

CHANGELOG.md
martinmitrevski: @ipavlidakis, also don't forget these docs: https://getstream.io/chat/docs/sdk/ios/client/controllers/channels/#2-create-a-controller, they might require an update.

nuno-vieira: Yes, they should touch on the points described here: https://github.com/GetStream/stream-chat-swift/pull/2488#pullrequestreview-1299334895

docusaurus/docs/iOS/client/controllers/channels.md
```diff
@@ -55,9 +55,26 @@ let compoundFilter: Filter<ChannelListFilterScope> = .and([
 The `query.sort` is an array of sorting options. Sorting options are applied based on their order in the array so the first option has the highest impact while the others are used mainly as a tiebreakers. By default, the channel list is sorted by `updated_at`.
 
 ### 2. Create a controller
-
+#### With auto Channel Filtering
```
nuno-vieira, line 58:
```suggestion
#### With Automatic Filtering
```

docusaurus/docs/iOS/client/controllers/channels.md
```diff
@@ -55,9 +55,26 @@ let compoundFilter: Filter<ChannelListFilterScope> = .and([
...
 The controller can be created using `ChatClient`:
 ```swift
+let controller = ChatClient.shared.channelListController(query: query)
+```
+The behaviour is controlled from the `ChatClientConfig` and more specifically the `isChannelAutomaticFilteringEnabled` flag.
```
nuno-vieira, line 63: I think it is better we provide a little introduction and incrementally increase the complexity, something like this:

```suggestion
### 2. Create a controller
The controller can be created using `ChatClient`:
`let controller = ChatClient.shared.channelListController(query: query)`

This is all you need to create a channel list controller, but depending on the complexity of your query, you might need to provide a custom filtering closure.

By default, the SDK will automatically handle filtering the channels as they get created. Whenever there is a web socket event that a channel has been created, the SDK will only insert it in the channel list if it matches the query. But, if the query provided contains extra data or custom filters, the SDK won't be able to automatically match the filter query, so you will need to provide a filtering closure.

#### Filtering with extra data
Here we should provide a real example of using a query + filtering closure with extra data.

#### Filtering with custom filters
Here we should provide a real example of using a query + filtering closure with a custom filter key provided.

#### Manual Filtering
Finally, here we should mention that you can disable the automatic filtering and provide a filter closure to replicate the filter logic. (We should also show a real example) 
```

docusaurus/docs/iOS/client/controllers/channels.md
```diff
@@ -55,9 +55,26 @@ let compoundFilter: Filter<ChannelListFilterScope> = .and([
...
+
+:::warning
+Custom filters are not currently supported. If you are already defining and using your own filters, then the initialization option `With Channel filtering closure` will be a better fit more your needs.
+:::
+
+#### With Channel filtering closure
+```swift
 let controller = ChatClient.shared.channelListController(query: query, filter: { channel in
     // the filtering logic goes here.
     return true
```
nuno-vieira, line 80: We definitely don't want this anymore. Customers have a really hard time understanding what is a filter closure since we don't provide a real example.

nuno-vieira (commented):
Almost there! 💪 There are some typos and I think the documentation can get some structural improvements 👍

CHANGELOG.md
nuno-vieira, line 11: I think that it also makes sense to add a point, to just mention that now, channels will be automatically filtered:
```suggestion
- Add support for automatically filtering channels in the Channel List [#2488](https://github.com/GetStream/stream-chat-swift/pull/2488)
- Add `isChannelLocalFilteringEnabled` in `ChatClientConfig` to allow changing whether the Channels in ChannelList will be locally filtered [#2488](https://github.com/GetStream/stream-chat-swift/pull/2488)
```

Sources/StreamChat/Config/ChatClientConfig.swift
```diff
@@ -182,6 +182,11 @@ public struct ChatClientConfig {
     /// It controls how long (in seconds) a network task should wait for additional data to arrive before giving up
     public var timeoutIntervalForRequest: TimeInterval = 30
 
+    /// Enable/Disable local filtering for Channel lists. When enabled,
+    /// the `ChannelDTO` will include the filter's predicate (if available)
+    /// in the fetchRequest.
```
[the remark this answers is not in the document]
ipavlidakis: Should they know about the `fetchRequest`?

docusaurus/docs/iOS/client/controllers/channels.md
ipavlidakis: You have a very good point and I like your suggestion. Before moving forward, I need to mention that as the autofiltering is happening on the query, the data that may `arrive` in the filter closure may already not containing what the customer needs. Do you think it makes more sense to disable the auto-filtering if a filter closure has been provided?

Sources/StreamChat/Config/ChatClientConfig.swift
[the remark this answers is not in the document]
nuno-vieira: They shouldn't as well, since it is an implementation detail

docusaurus/docs/iOS/client/controllers/channels.md
```diff
@@ -55,28 +55,73 @@ let compoundFilter: Filter<ChannelListFilterScope> = .and([
...
+        var config = ChatClientConfig(apiKeyString: "<# Your API Key Here #>")
+        config.isChannelAutomaticFilteringEnabled = false
+
+        // Create an instance of the `ChatClient` with the given config
+        let client = ChatClient(config: config)
 
-The controller can be created using `ChatClient`:
+        return client
+    }()
+}
```
nuno-vieira, line 82: There's no need to mention this since I believe the customer by this point should already have this in the getting started guide 🤔

Sources/StreamChat/Query/ChannelListQuery.swift
```diff
@@ -58,67 +58,67 @@ public enum InviteFilterValue: String, FilterValue {
...
 
     /// Filter for checking whether current user is joined the channel or not (through invite or directly)
     /// Supported operators: `equal`
-    static var joined: FilterKey<Scope, Bool> { .init(rawValue: "joined", keyPathValueProvider: { "membership" }) }
+    static var joined: FilterKey<Scope, Bool> { .init(rawValue: "joined", keyPathString: #keyPath(ChannelDTO.membership)) }
 
     /// Filter for checking whether current user has muted the channel
     /// Supported operators: `equal`
-    static var muted: FilterKey<Scope, Bool> { .init(rawValue: "muted", keyPathValueProvider: { "mute" }) }
+    static var muted: FilterKey<Scope, Bool> { .init(rawValue: "muted", keyPathString: #keyPath(ChannelDTO.mute)) }
```
polqf, line 121: 👏

Sources/StreamChat/Config/ChatClientConfig.swift
```diff
@@ -182,6 +182,11 @@ public struct ChatClientConfig {
...
+    public var isChannelLocalFilteringEnabled: Bool = true
```
polqf: Makes more sense here, yes!

docusaurus/docs/iOS/client/controllers/channels.md
```diff
@@ -55,49 +55,73 @@ let compoundFilter: Filter<ChannelListFilterScope> = .and([
 The `query.sort` is an array of sorting options. Sorting options are applied based on their order in the array so the first option has the highest impact while the others are used mainly as a tiebreakers. By default, the channel list is sorted by `updated_at`.
 
 ### 2. Create a controller
-#### With auto Channel Filtering
-The controller can be created using `ChatClient`:
+The simplest way to create a controller is by using the static method `channelListController(query:)` on your `ChatClient`.
```
polqf, line 58: Not an `static` method 👀

docusaurus/docs/iOS/client/controllers/channels.md
```diff
@@ -55,49 +55,73 @@ let compoundFilter: Filter<ChannelListFilterScope> = .and([
...
-By default, the SDK will handle filtering the channels as they get updated and will only return to the controller **only** the channels that match the provider `query.filter`. 
-This means, that the controller always receives channels that match its `query.filter` and then the controller is responsible for deciding if the channels need to be linked/unlinked.
+:::note
+In cases though, where the query provided contains, extra data or custom filters, **the SDK may not be able to automatically match the filter query**. In this case, you will need to provide a filtering closure.
+:::
 
-When the flag is disabled (`false`), then the new/updated channels won't be filtered as the arrive, and the controller will be responsible for filtering them first and confirm that they
-match its `query.filter` and then decide if the filtered channels need to be linked/unlinked.
+#### Filtering with extra data
+First we need to disable the Channel auto-filtering. We can do that by turning the `isChannelAutomaticFilteringEnabled` in your `ChatClient` configuration, to `false`.
```
polqf, line 69: Not sure if this should be the suggested approach. We can also suggest an "additive" block, right?

polqf (approved):
All good, just couple of comments 👏 Looking forward to getting this out!!

docusaurus/docs/iOS/client/controllers/channels.md
```diff
@@ -55,28 +55,73 @@ let compoundFilter: Filter<ChannelListFilterScope> = .and([
...
+By default, the SDK will automatically handle filtering the channels as they get created. Whenever there is a web socket event that a channel has been created, the SDK will only insert it in the channel list if it matches the query.
+
+:::note
+In cases though, where the query provided contains, extra data or custom filters, **the SDK may not be able to automatically match the filter query**. In this case, you will need to provide a filtering closure.
+:::
+
+#### Filtering with extra data
+First we need to disable the Channel auto-filtering. We can do that by turning the `isChannelAutomaticFilteringEnabled` in your `ChatClient` configuration, to `false`.
+```swift
+extension ChatClient {
```
nuno-vieira, line 71: The idea in this section is to not mention yet the `isChannelAutomaticFilteringEnabled`, only in the end, in a `#### Manual Filtering` section. The idea of this section is to show an example where automatic filtering is enabled, and only the extra data property is used in the closure. Example:

Here is an example to filter
```swift
let query = ChannelListQuery(filter: .and([
    .containMembers(userIds: [userCredentials.id])),
    .equal(.type, to: .messaging),
    .equal("myCustomBooleanKey", to: true),
])

let controller = ChatClient.shared.channelListController(query: query, filter: { channel in
      channel.extraData["myCustomBooleanKey"]?.boolValue == true
}
```

👆 Here, we should explain that only the extra data needs to be in the closure

docusaurus/docs/iOS/client/controllers/channels.md
```diff
@@ -55,28 +55,73 @@ let compoundFilter: Filter<ChannelListFilterScope> = .and([
...
-The `filter` closure is needed to dynamically link/unlink channels to the list during the controller's lifecycle to keep the list in sync with the remote:
-
-When `filter` returns `true`, the new/updated channel is linked/stays linked to the channel list
+        // Create an instance of the `ChatClient` with the given config
+        let client = ChatClient(config: config)
 
-When `filter` returns `false`, the new/updated channel is not linked/is unlinked from the channel list.
+        return client
+    }()
+}
```
nuno-vieira, line 108: We should remove this as well

docusaurus/docs/iOS/client/controllers/channels.md
```diff
@@ -55,28 +55,73 @@ let compoundFilter: Filter<ChannelListFilterScope> = .and([
...
-    // the filtering logic goes here.
-    return true
+    // We are filtering for channels that:
+    // 1. The current is user is a member
+    // 2. A value exists for the extraData key `myCustomBooleanKey` and this value is `true`
+    return channel.membership != nil 
+        && channel.extraData["myCustomBooleanKey"] as? Bool == true
 })
 ```
+#### Filtering with custom filters
```
nuno-vieira, line 94: I actually don't think we have any section in the docs on how to create custom filters 🤔 So maybe we should discard this one for now. Otherwise, we need to make an introduction here on how to create one.

docusaurus/docs/iOS/client/controllers/channels.md
ipavlidakis: You are right, Nuno had a similar idea but I have to admit that i was a little bit confused. I have now updated the docs with the `filter closure` being described as additive. Additionally I added a test that demonstrates the scenario where a custom filter is applied and the auto-filtering is ignoring it and evaluates partially the filter.

docusaurus/docs/iOS/client/controllers/channels.md
ipavlidakis: I just added it here to make it complete. How would you write it?

nuno-vieira (approved):
LGTM! ✅

CHANGELOG.md
```diff
@@ -8,6 +8,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Add `UploadedAttachmentPostProcessor` in `ChatClientConfig` to allow changing custom attachment payloads after an attachment has been uploaded [#2457](https://github.com/GetStream/stream-chat-swift/pull/2457)
 - Add `AnyAttachmentPayload(localFileURL:customPayload:)` initializer to allow creating custom attachments without a remote URL [#2457](https://github.com/GetStream/stream-chat-swift/pull/2457)
 - Add skip push support when sending a message [#2486](https://github.com/GetStream/stream-chat-swift/pull/2486)
+- Add support for automatically filtering channels in the Channel List [#2488](https://github.com/GetStream/stream-chat-swift/pull/2488)
+- Add `isChannelAutomaticFilteringEnabled` in `ChatClientConfig` to allow changing whether the Channels in ChannelList will be automatically filtered [#2488](https://github.com/GetStream/stream-chat-swift/pull/2488)
+
+### 🔄 Changed
+- Remove unused ReactionNotificationContent [#2485](https://github.com/GetStream/stream-chat-swift/pull/2485)
```
nuno-vieira, line 15: Probably a mistake in resolving the conflicts, this is duplicated:
```suggestion
```

docusaurus/docs/iOS/client/controllers/channels.md
```diff
@@ -55,28 +55,67 @@ let compoundFilter: Filter<ChannelListFilterScope> = .and([
...
+    .containMembers(userIds: [currentUserId]),
+    .equal(.type, to: .messaging),
+    .equals("myCustomBooleanKey", value: true)
+]), filter: { channel in
+    // The channel is guaranteed to:
+    // 1. contain a member with id the currentUserId
+    // 2. have type == `.messaging` 
+    // We are filtering for channels that a value exists for the extraData 
+    // key `myCustomBooleanKey` and this value is `true`
+    return channel.extraData["myCustomBooleanKey"] as? Bool == true
```
nuno-vieira, line 84:
```suggestion
    return channel.extraData["myCustomBooleanKey"]?.boolValue == true
```

docusaurus/docs/iOS/client/controllers/channels.md
```diff
@@ -55,28 +55,67 @@ let compoundFilter: Filter<ChannelListFilterScope> = .and([
...
+]), filter: { channel in
+    // As we have disabled the auto-filtering, the SDK will not try to match
+    // the channels in the filter and instead will forward them to the 
+    // filter closure where we are expected to apply our custom 
+    // filtering logic.
+    // 
+    // In this case, we need to evaluate manually all parts of the filter.
+    return channel.members.map(\user.id).contains(currentUserId) 
+        && channel.type == .messaging,
+        && channel.extraData["myCustomBooleanKey"] as? Bool == true
```
nuno-vieira, line 116:
```suggestion
        && channel.extraData["myCustomBooleanKey"]?.boolValue == true
```

PR #2489, opened 2023-02-13
[BW] Prepare simulator before first scan

alteral:
## Description

`Then` prepare_simulator lane was not triggered on cron checks
`Now` it will


nuno-vieira (approved):
LGTM! ✅

PR #2497, opened 2023-02-20
Fix auto-filtering when filter and DB value types don't match

ipavlidakis:
### 🔗 Issue Links

Resolves [Link](https://github.com/GetStream/ios-issues-tracking/issues/326)

### 📝 Summary

The following filter doesn't work correctly with auto filtering.

```swift
let controller = streamChatClient.channelListController(
      query: ChannelListQuery(filter: .and([
            .equal(.type, to: .custom("test")),
            .containMembers(userIds: ["12632287"]),
      ]))
)
```

### 🛠 Implementation

The issue appears to be that the filter's and DB's value type don't match and when they are being compared it fails.

To maintain our type-safe API while supporting auto-filtering, we will provide a valueMapper on the filter that will allow in cases where the filter's and DB's value type don't match, to disambiguate the problem by mapping the filter's value to the a value that matches the DB's value type.

### ☑️ Contributor Checklist

### 🎁 Meme

meme

### DIFF
--- a/Sources/StreamChat/Query/Filter.swift
+++ b/Sources/StreamChat/Query/Filter.swift
@@ -215,34 +215,34 @@ public struct FilterKey<Scope: FilterScope, Value: FilterValue>: ExpressibleBySt
 
     /// The mapper that will transform the input value to a value that
     /// can be compared with the DB value
-    typealias InputValueMapper = (Any) -> FilterValue?
-    typealias TypeInputValueMapper = (Value) -> FilterValue?
-    let inputValueMapper: InputValueMapper?
+    typealias ValueMapper = (Any) -> FilterValue?
+    typealias TypedValueMapper = (Value) -> FilterValue?
+    let valueMapper: ValueMapper?
 
     public init(stringLiteral value: String) {
         rawValue = value
-        inputValueMapper = nil
+        valueMapper = nil
         keyPathString = nil
     }
 
     public init(rawValue value: String) {
         rawValue = value
         keyPathString = nil
-        inputValueMapper = nil
+        valueMapper = nil
     }
 
     init(
         rawValue value: String,
         keyPathString: String,
-        inputValueMapper: TypeInputValueMapper? = nil
+        valueMapper: TypedValueMapper? = nil
     ) {
         rawValue = value
         self.keyPathString = keyPathString
-        self.inputValueMapper = {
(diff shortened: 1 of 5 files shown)


martinmitrevski (approved):
LGTM ✅

polqf (approved):
Just wondering if there is any case in which we would like to allow integrators to create a filter  key with a mapper

Sources/StreamChat/Query/Filter.swift
```diff
@@ -99,18 +99,27 @@ public struct Filter<Scope: FilterScope> {
...
+
+    /// The keypath of the DB object that will be compared with the input value during
+    /// a local filtering.
     let keyPathString: String?
 
     init(
         operator: String,
         key: String?,
         value: FilterValue,
+        valueMapper: ValueMapper?,
```
nuno-vieira, line 115: The problem I see with this approach is that it might be error prone, we need to make sure we don't forget to provide a value mapper for all enums or for every FilterValue that requires it. IMO we should make sure all FilterValues are convirtible to String raw value

ipavlidakis: We had a discussion about this. The problem with having a way to convert everything to `String` it may require us to expose functionality to customers that it's not needed. 

In this way we are very explicit about when we need to include the valueMapper. 

As this approach is meant for internal usage only and helps us quickly resolve the issue we have, I consider it safe enough. Once we start exposing more entities into auto filtering and especially when we need to expose the functionaliy publically we can reevaluate it

nuno-vieira: Exposing a filterRawValue should not be a big deal, but fine, if the team is okay with this approach. I just think we could introduce similar issues in the future. But we can improve it later if we verify it is error prone 👍

nuno-vieira: Either way, we should test all filters to make sure they are converted correctly. At least only the equal operator. For example, how is URL converted?

ipavlidakis: Exposing a `filterRawValue` means that we are moving the responsibility for converting the value to the filter when the definition of it is happening on the FilterKey. For simplicity, we should try to keep the filter as clean and unaware of the key and it's operations as possible. In that way we can have every related to the key and how it's being interpreted in one place.

PR #2507, opened 2023-02-22
Feature/restructure headers footers logic in message cell

ipavlidakis:
### 🔗 Issue Links

### 🎯 Goal

We would like to make cell decorations (headers and footers), easier to implement and customise.

### 📝 Summary

* The SDK matched and identifies the header and footer of a cell through the `ChatMessageDecorationView` type.
* The API has been added on the MessageList's delegate (`ChatMessageListVCDelegate`) following the same path that Apple's `UITableViewDelegate` is using to provide decorations for a section.
* In comparison to the `UITableView` Header and Footer, our Decoration views are not **yet** cachable. This is due to a technical issue when trying to cache any view, *that doesn't conform on `UITableViewHeaderFooterView`*, on a `UITableView`.
* Both the `ChatChannelVC` and `ChatThreadVC` have been updated to use the new API to present the DateSeparator views.
* Any time the delegate provides `nil` as a result, the `ChatMessageListVC` will ensure that the space for this decorationType has been released.

### 🛠 Implementation

The `ChatMessageListVC` will ask the `ChatMessageListVCDelegate` to provide decorationViews for header and footer everytime it goes through the `func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell`. The `ChatMessageListVC` will provide to the delegate the: *message*, *decorationType* and *indexPath* that describe the cell and the decorationView that is asking for.

- If the `ChatMessageListVCDelegate` returns a **non-nil** result, then the provided view will be added in the cell's UI in the position that is described by its decorationType.
- If the `ChatMessageListVCDelegate` return a **nil** result, the cell's UI will be updated to release the space used by the decorationType.

image

### Public Interface Changes

- Added `func chatMessageListVC(_ vc: ChatMessageListVC, headerViewForMessage message: ChatMessage, at indexPath: IndexPath)` on `ChatMessageListVCDelegate` that enables users to provide their own decorations for headers.
- Added `func chatMessageListVC(_ vc: ChatMessageListVC, footerViewForMessage message: ChatMessage, at indexPath: IndexPath)` on `ChatMessageListVCDelegate` that enables users to provide their own decorations for headers.
- Updated `ChatMessageListDateSeparatorView` to occupy the whole provided width and adjust its content's width accordingly (it will remain centred while fits its content). That enable further customisation.

### Object Responsibility Changes

- `ChatMessageListVC` is not responsible anymore about configuring the message's DateSeparator. The responsibility has now being moved to its delegate (e.g. `ChatChannelVC`, `ChatThreadVC`). In that way users have more flexibility providing their own decorationViews while `ChatMessageListVC` remains simple and agnostic about these parts of the cell.

### Alternatives Considered
**Caching**
You are already familiar with `UITableViewHeaderFooterView` from `UITableView`. Those are decoration views that can be used above or below a section in the `UITableView`. What is important here is that **the `UITableView` will cache the `UITableViewHeaderFooterView` instances and it will try to reuse them** (similarly to what is happening with `UITableViewCell`). 
Initially, our cell decorationViews were subclassing `UITableViewHeaderFooterView` and they were also going through the same caching/reusing flow as the section decorationViews. That was causing issues because, by default `UITableViewHeaderFooterView` are being managed by the `UITableView` and their layout is designed to be managed by anyone else. Due to the structure of the `ChatMessageCell` are being positioned using Autolayout, inside a `UIStackView`. That was causing the SDK to change the `translatesAutoresizingMaskIntoConstraints` to `true` which was causing UIKit to log a warning for unexpected behaviour. 
For the time being, we removed the subclassing of `UITableViewHeaderFooterView` in our decorationViews, which results in them not getting cached and reused.

**Provide decorations as LayoutOptions**
This way made sense as all the elements on the message's contentView are provided in that way. The main issue with this approach is that `LayoutOptions` are commonly customised which means that if we made the change there, that would cause a potential breaking change (where date separators are not visible) for users that override and provide their own `LayoutOptions`. 

### 🎨 Showcase

| Before | After |
| ------ | ----- |
|||

### 🧪 Manual Testing Notes

Run through channels and pay attention to the Date separator that appears above some messages. You shouldn't notice any difference compared to the version before.

### ☑️ Contributor Checklist

### 🎁 Meme


Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageCell.swift
```diff
@@ -11,22 +11,26 @@ import UIKit
...
+    /// about the message above the message's content.
+    internal lazy var headerContainerView: UIView = .init()
         .withoutAutoresizingMaskConstraints
-        .withAccessibilityIdentifier(identifier: "dateSeparatorView")
+        .withAccessibilityIdentifier(identifier: "headerContainerView")
 
     /// The message content view the cell is showing.
     public private(set) var messageContentView: ChatMessageContentView?
 
+    /// The header View of the cell. It can be used to display additional information
```
nuno-vieira, line 28:
```suggestion
    /// The footer View of the cell. It can be used to display additional information
```

Sources/StreamChatUI/CommonViews/BaseViews.swift
```diff
@@ -466,3 +466,7 @@ open class _TableViewCell: UITableViewCell, Customizable, AccessibilityView {
         super.layoutSubviews()
     }
 }
+
+/// Base class for overridable views StreamChatUI provides.
+/// All conformers will have StreamChatUI appearance settings by default.
+open class _TableViewHeaderFooterReusableView: _View {}
```
nuno-vieira, line 472: Do we still need this? Since the view is actually not Reusable with the current implementation

Sources/StreamChatUI/ChatMessageList/ChatMessageListVCDelegate.swift
```diff
@@ -46,4 +46,17 @@ public protocol ChatMessageListVCDelegate: AnyObject {
         didTapOnMessageListView messageListView: ChatMessageListView,
         with gestureRecognizer: UITapGestureRecognizer
     )
+
+    /// Ask the delegate to provide a decoration view for the specified decoration type.
+    /// - Parameters:
+    ///   - vc: The message list  informing the delegate of this event.
```
nuno-vieira, line 52: Additional spacing:
```suggestion
    ///   - vc: The message list informing the delegate of this event.
```

docusaurus/docs/iOS/uikit/components/message.md
```diff
@@ -108,7 +112,29 @@ Components.default.messageLayoutOptionsResolver = CustomMessageLayoutOptionsReso
 | ------------- | ------------- |
 | <img src={require("../../assets/message-basic-resolver-before.png").default} /> | <img src={require("../../assets/message-basic-resolver-after.png").default} /> |
 
-### Date Separators
+#### Decoration Views
+
+The SDK allows you to configure what will be presented above and below your message. Those views we are calling them `ChatMessageDecorationView`. `ChatMessageDecorationView` are fully customizable and they are even used for standard SDK components like the `Date Separators` that we will see below.
```
nuno-vieira, line 117: WDYT of this shorter version?
```suggestion
The SDK allows you to configure what will be presented above and below your message with decoration views (`ChatMessageDecorationView`). They are fully customizable, and we also use them for standard SDK components like the `Date Separators` as seen below.
```

docusaurus/docs/iOS/uikit/components/message.md
```diff
@@ -108,7 +112,29 @@ Components.default.messageLayoutOptionsResolver = CustomMessageLayoutOptionsReso
...
+) -> ChatMessageDecorationView?
+```
+`ChatMessageDecorationView` are following a similar flow as the `UITableViewHeaderFooterView` decorations for sections that `UITableView` is managing. During the preparation of a message cell, the `ChatMessageListVC` will ask the delegate to provide `ChatMessageDecorationView` for header and footer. If the delegate returns a non-nil value the 
+provided `ChatMessageDecorationView` will be placed in the cell's UI according to it's decoration type. In the case where the delegate will return a nil value then the cell will be updated to release the space for this specific decoration type, if it was previously used.
+
+:::note
+Unlike the `UITableViewHeaderFooterView`, the cell `ChatMessageDecorationView` that the SDK manages, is not cached and reused. This was decided due to some limitations of `UITableViewHeaderFooterView` that were preventing us from making `ChatMessageDecorationView` inherit from.
+
+This is a known limitation and we are working on providing a solution in a following update.
+:::
```
nuno-vieira, line 135: We should not provide implementation details in the documentation, especially this one might warn the customer for no reason. 
```suggestion
```

Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
```diff
@@ -243,6 +243,22 @@ open class ChatChannelVC: _ViewController,
...
+        decorationType: ChatMessageDecorationType,
+        at indexPath: IndexPath
+    ) -> ChatMessageDecorationView? {
+        switch decorationType {
+        case .header where vc.shouldShowDateSeparator(forMessage: message, at: indexPath):
+            let dateSeparatorView = components.messageListDateSeparatorView.init()
+            dateSeparatorView.content = vc.dateSeparatorFormatter.format(message.createdAt)
+            return dateSeparatorView
+        default:
+            return nil
```
polqf, line 258: Looks clean! Have you considered having separate delegate methods for the header and footer? It might make it a bit more readable as more customizations are added

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageCell.swift
```diff
@@ -65,12 +71,45 @@ public class ChatMessageCell: _TableViewCell, ComponentsProvider {
...
+        decorationView.translatesAutoresizingMaskIntoConstraints = false
+        container.embed(decorationView)
+        switch decorationType {
+        case .header:
+            containerStackView.insertArrangedSubview(container, at: 0)
+        case .footer:
+            containerStackView.addArrangedSubview(container)
+        }
+
+        container.pin(
```
polqf, line 109: You might end up pinning multiple times if you call updateDecoration on the same cell, right?

Sources/StreamChatUI/CommonViews/BaseViews.swift
polqf, line 472: Seems like it is a leftover?

nuno-vieira (commented):
Overall the PR looks great 👌 However, there is currently a breaking change, so we need to be careful with this. ⚠️

Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
nuno-vieira: Yes this is a good point 👍 It also seems more flexible. If for some reason we need more decoration types, we will need to add new options to the enum type, which is a breaking change

Sources/StreamChatUI/ChatThread/ChatThreadVC.swift
```diff
@@ -258,6 +258,22 @@ open class ChatThreadVC: _ViewController,
...
+    ) -> ChatMessageDecorationView? {
+        switch decorationType {
+        case .header where vc.shouldShowDateSeparator(forMessage: message, at: indexPath):
+            let dateSeparatorView = components.messageListDateSeparatorView.init()
+            dateSeparatorView.content = vc.dateSeparatorFormatter.format(message.createdAt)
+            return dateSeparatorView
+        default:
+            return nil
+        }
+    }
```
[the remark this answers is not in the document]
nuno-vieira: Another problem with this, is that it means we need to duplicate 3 times the default implementation right now 🤔 It might be worth it to export this default implementation into a component, where we just call it in the 3 different implementations. Then for specific things related to the Channel or Thread, it will be handled only there 🤔 Just thinking out loud here. But ideally, for every non-channel/thread related separators, the Channel, Thread or a Custom implementation of the Message List, would get them for fre

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageCell.swift
nuno-vieira: Yes this is also a good point. @ipavlidakis Any reason why we don't simply hide/show like the previous implementation. Keep in mind that Apple actually recommends this approach for performance reasons, it is better to hide/show views in table view cells, than removing/adding views or changing constraints ad-hoc

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageCell.swift
```diff
@@ -65,12 +71,45 @@ public class ChatMessageCell: _TableViewCell, ComponentsProvider {
...
+        headerContainerView.backgroundColor = nil
+        footerContainerView.backgroundColor = nil
     }
 
     override public func prepareForReuse() {
         super.prepareForReuse()
 
         messageContentView?.prepareForReuse()
+        headerContainerView.removeFromSuperview()
+        footerContainerView.removeFromSuperview()
```
nuno-vieira, line 84: Do we still want this based on this comment: https://github.com/GetStream/stream-chat-swift/pull/2507#discussion_r1116030956

Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
ipavlidakis: I have tried the footer/header split out in different methods a couple of times but i was reverting it back. Just couldn't make up my mind.

The only benefit that this single approach has is that for users they only have to implement/override one method instead of 2.

That being said I'm really happy that you both agree on the approach and you help choose which one to use to move forward.

@nuno-vieira more decoration types may not require addition to the enum type. Even if they do the enum is not extensible. Finally, even in the approach with the 2 methods a new decoration type will require a new method

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageCell.swift
ipavlidakis: The issue I had with hide/show is that the spacing between the views is still there, which is causing UI changes. 
What I can do is to remove the spacing from the stackview and move it in the containers instead. WDYT?

Sources/StreamChatUI/ChatThread/ChatThreadVC.swift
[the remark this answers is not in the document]
ipavlidakis: Any implementation of `ChatMessageListVCDelegate` the users have they will need to implement the 2 new methods. So, what we can do is to provide a method on an extension of `ChatMessageListVCDelegate` that can be used to create the DateSeparator. 
I don't see a need to default implement the 2 new methods. WDYT?

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageCell.swift
nuno-vieira: Yes, that makes sense since if a customer wants to customize the separator, including the spacing, it will only do it in 1 view, it does need to change the spacing in the stack view. So this is more like padding, then spacing actually 👍

Sources/StreamChatUI/ChatThread/ChatThreadVC.swift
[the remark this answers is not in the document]
nuno-vieira: We need to provide a default implementation, unfortunately. Otherwise this won't compile for customers, which is a breaking change, and would require a major bump. Protocols in SDKs can be problematic, especially if the API is not final. 

Another problem here, is that customers already had the date separators out of the box with the `ChatMessageListVC`, and with this change, they don't. So we really need to provide this implementation out of the box.

docusaurus/docs/iOS/uikit/components/message.md
```diff
@@ -108,7 +112,25 @@ Components.default.messageLayoutOptionsResolver = CustomMessageLayoutOptionsReso
 | ------------- | ------------- |
 | <img src={require("../../assets/message-basic-resolver-before.png").default} /> | <img src={require("../../assets/message-basic-resolver-after.png").default} /> |
 
-### Date Separators
+#### Decoration Views
+:::note
+Decoration Views are available on SDK version 4.28.0 and above.
```
nuno-vieira, line 117: It will be 4.29.0 actually since we just released 4.28 🙂 
```suggestion
Decoration Views are available on SDK version 4.29.0 and above.
```

docusaurus/docs/iOS/uikit/components/message.md
```diff
@@ -108,7 +112,25 @@ Components.default.messageLayoutOptionsResolver = CustomMessageLayoutOptionsReso
...
+
+In order to provide a `ChatMessageDecorationView` (either a header or a footer) for a message, you will need to implement the following method from the `ChatMessageListVCDelegate`
+```swift
+func chatMessageListVC(
+    _ vc: ChatMessageListVC,
+    decorationViewForMessage message: ChatMessage,
+    decorationType: ChatMessageDecorationType,
+    at indexPath: IndexPath
+) -> ChatMessageDecorationView?
+```
```
nuno-vieira, line 130: This is outdated right?

docusaurus/docs/iOS/uikit/components/message.md
```diff
@@ -163,11 +185,20 @@ class CustomChatMessageListDateSeparatorView: ChatMessageListDateSeparatorView {
...
 
+:::important
+`Date Separator` is a type of `ChatMessageDecorationView` and its creation flow goes through the `ChatMessageListVCDelegate`. Both `ChatChannelVC` and `ChatThreadVC` configure `Date Separator` based on the `Components.default.messageListDateSeparatorEnabled` configuration flag.
+
+In the case where you are overriding the `func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell` method from `ChatMessageListVC` you may want to ensure that:
+1. You either call `super.tableView(tableView, cellForAt: indexPath)` at some point in your code 
+or
+2. You implement and call the `ChatMessageListVCDelegate` method that manages `ChatMessageDecorationView`
+:::
+
```
nuno-vieira, line 201: For me this seems a bit unnecessary 🤔 WDYT? 
```suggestion
```

nuno-vieira (commented):
LGTM! ✅ We just need to update the docs and we are ready 🚀

ipavlidakis:
@nuno-vieira updated

nuno-vieira (approved):
LGTM! 🚀

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageCell.swift
```diff
@@ -11,27 +11,36 @@ import UIKit
...
+        .withoutAutoresizingMaskConstraints
+        .withAccessibilityIdentifier(identifier: "footerContainerView")
+
     /// The minimum spacing below the cell.
     public var minimumSpacingBelow: CGFloat = 2 {
         didSet { updateBottomSpacing() }
     }
 
+    /// The spacing between the header, content and footer
+    public var spacingBetween: CGFloat = 8 {
```
polqf, line 40: headerFooterSpacing?

polqf (approved):

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageCell.swift
ipavlidakis: If you don't mind i'll update it on the following patch that implements the thread replies counter decoration view

PR #2514, opened 2023-02-27
Fix "In" Filter only returning results when all values match

nuno-vieira:
### 🔗 Issue Links
Resolves https://github.com/GetStream/ios-issues-tracking/issues/333

### 🎯 Goal
Fix "In" Filter only returning results when all values match.

### 📝 Summary
The Mongoose style "in" operator, used by our backend API is actually an "or" operation, meaning that only one of the given values need to match, and not all of them. So this was a simple change in our side.

### 🧪 Manual Testing Notes
Unit tests were provided to cover regression.

### ☑️ Contributor Checklist


polqf (approved):
Oh nice! 👏

PR #2519, opened 2023-03-02
Mark as unread

polqf:
### 🔗 Issue Links

https://github.com/GetStream/ios-issues-tracking/issues/285

### 🎯 Goal

This PR adds an option to mark as unread a chat starting from a particular message

### 📝 Summary

Now it is possible to select a particular message, and mark the channel as unread starting from that message.

### 🛠 Implementation

ChannelController exposes the first unread message id, and this one is used in the UI to show a banner (not part of this PR) to visually distinguish new messages

### 🧪 Manual Testing Notes

1. Open a channel
2. Long press on a message
3. Select "Mark As Unread"

Expected result:
- The channel should be marked as unread starting from that message

### ☑️ Contributor Checklist

- [x] I have signed the [Stream CLA](https://docs.google.com/forms/d/e/1FAIpQLScFKsKkAJI7mhCr7K9rEIOpqIDThrWxuvxnwUq2XkHyG154vQ/viewform) (required)
- [x] This change follows zero ⚠️ policy (required)
- [x] This change should be manually QAed
- [ ] Changelog is updated with client-facing changes
- [ ] New code is covered by unit tests
- [ ] Comparison screenshots added for visual changes
- [ ] Affected documentation updated (docusaurus, tutorial, CMS)

PR #2526, opened 2023-03-09
Mark As Unread: Add UI and WS Events Handling

polqf:
### 🔗 Issue Links

#2519 

### 🎯 Goal

Show unread messages separator when marking a message as unread

### 📝 Summary

This PR adds the UI related part for Mark as unread. It also adds handling for the WS Events that are sent when this happens

### 🛠 Implementation

- Reuses `ChatThreadRepliesCountDecorationView` to show the UI separator.
    - Class has been renamed to `ChatUnreadMessagesCountDecorationView` so it is generic for Threads and Unread messages
- Mark as unread WS Events are now handled through `ChannelReadUpdaterMiddleware`
- Disable Mark as unread from Threads

### 🧪 Manual Testing Notes

1. Go to a channel
2. Long press on a message
3. Choose "Mark as unread"

Expected result:

- Should add a UI indicator on top of the message
- The message should be marked as unread

### ☑️ Contributor Checklist


Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatUnreadMessagesCountDecorationView.swift
```diff
@@ -5,8 +5,8 @@
 import StreamChat
 import UIKit
 
-/// The date separator view that groups messages from the same day.
-open class ChatThreadRepliesCountDecorationView: ChatMessageDecorationView, AppearanceProvider {
+/// The view used to show a separator when there are unread messages.
+open class ChatUnreadMessagesCountDecorationView: ChatMessageDecorationView, AppearanceProvider {
```
ipavlidakis, line 9: The 2 views are showing semantically different content. The new one shows unread count while the existing one shows number of replies. 

If we are to reuse this view for both cases, the name should something more generic to fit both cases like `ChatMessagesCountDecorationView`

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatUnreadMessagesCountDecorationView.swift
```diff
@@ -15,7 +15,7 @@ open class ChatThreadRepliesCountDecorationView: ChatMessageDecorationView, Appe
     /// The container that the contentTextLabel will be placed aligned to its centre.
     open private(set) lazy var container: UIView = UIView()
         .withoutAutoresizingMaskConstraints
-        .withAccessibilityIdentifier(identifier: "repliesCountDecorationContainer")
+        .withAccessibilityIdentifier(identifier: "unreadMessagesDecorationContainer")
```
ipavlidakis, line 18: This one will also need to be updated, as inside thread it won't represent unread messages but instead number of replies

Sources/StreamChatUI/ChatMessageList/ChatMessage/StackViewDecoratorView.swift
```diff
@@ -0,0 +1,36 @@
+//
+// Copyright © 2023 Stream.io Inc. All rights reserved.
+//
+
+import UIKit
+
+/// The decorator view that is used to stack multiple decorators.
+class StackViewDecoratorView: ChatMessageDecorationView {
```
ipavlidakis, line 8: nit: Can we make it `final`?

Sources/StreamChat/WebSocketClient/Events/NotificationEvents.swift
```diff
@@ -104,18 +104,24 @@ public struct NotificationMarkReadEvent: ChannelSpecificEvent, HasUnreadCount {
...
     public let createdAt: Date
+
+    /// The id of the first unread message id
+    public let firstUnreadMessageId: MessageId
+
+    /// The timestamp of the last read message
+    public let lastReadAt: Date
+
+    /// The number of unread messages for the channel
+    public let unreadMessages: Int
```
nuno-vieira, line 124: What about making it more clear that it is a number?
```suggestion
    public let unreadMessagesCount: Int
```

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessagesCountDecorationView.swift
```diff
@@ -28,7 +28,7 @@ open class ChatThreadRepliesCountDecorationView: ChatMessageDecorationView, Appe
         super.setUpLayout()
 
         embed(container, insets: .init(top: 0, leading: 0, bottom: 8, trailing: 0))
-        container.embed(textLabel, insets: .init(top: 3, leading: 9, bottom: 3, trailing: 9))
+        container.giembed(textLabel, insets: .init(top: 3, leading: 9, bottom: 3, trailing: 9))
```
nuno-vieira, line 31: What is giembed? 🤔

Sources/StreamChatUI/MessageActionsPopup/ChatMessageActionsVC.swift
```diff
@@ -105,7 +105,7 @@ open class ChatMessageActionsVC: _ViewController, ThemeProvider {
                 actions.append(threadReplyActionItem())
             }
 
-            if channelConfig.readEventsEnabled {
+            if channelConfig.readEventsEnabled && !message.isPartOfThread {
```
nuno-vieira, line 108: We need to keep in mind replies which are displayed in the channel
```suggestion
            if channelConfig.readEventsEnabled && !message.isPartOfThread || message.isShowReplyInChannel {
```

Tests/StreamChatUITests/SnapshotTests/ChatChannel/ChatChannelVC_Tests.swift
```diff
@@ -573,6 +573,45 @@ final class ChatChannelVC_Tests: XCTestCase {
...
+            at: .init(row: 0, section: 0)
+        )
+        let separatorView = try XCTUnwrap(headerDecorationView as? StackViewDecoratorView)
+        XCTAssertEqual(separatorView.content.count, 2)
+        let dateSeparatorView = try XCTUnwrap(separatorView.content.first as? ChatMessageListDateSeparatorView)
+        let unreadMessagesSeparator = try XCTUnwrap(separatorView.content.last as? ChatMessagesCountDecorationView)
+        XCTAssertEqual(dateSeparatorView.content, "Jan 01")
+        XCTAssertEqual(unreadMessagesSeparator.content, "1 UNREAD MESSAGE")
+
+        AssertSnapshot(vc, variants: [.defaultLight])
```
nuno-vieira, line 613: It would be nice to include more variants. How does it look in Dark theme + bigger font sizes?

Sources/StreamChatUI/ChatMessageList/ChatMessage/StackViewDecoratorView.swift
```diff
@@ -0,0 +1,36 @@
+//
+// Copyright © 2023 Stream.io Inc. All rights reserved.
+//
+
+import UIKit
+
+/// The decorator view that is used to stack multiple decorators.
+final class StackViewDecoratorView: ChatMessageDecorationView {
```
nuno-vieira, line 8: Any reason why this one is not public?

Sources/StreamChatUI/ChatMessageList/ChatMessage/StackViewDecoratorView.swift
```diff
@@ -0,0 +1,36 @@
+//
+// Copyright © 2023 Stream.io Inc. All rights reserved.
+//
+
+import UIKit
+
+/// The decorator view that is used to stack multiple decorators.
```
nuno-vieira, line 7: Can we add more context to this comment? It is not 100% clear why when need this stack of multiple decorators 🤔

nuno-vieira (commented):
Overall looks good 👌 Added a couple of questions + one missing logic required I believe

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessagesCountDecorationView.swift
polqf: 🙃

Sources/StreamChatUI/MessageActionsPopup/ChatMessageActionsVC.swift
polqf: Why? Those can actually be marked as unread

Tests/StreamChatUITests/SnapshotTests/ChatChannel/ChatChannelVC_Tests.swift
polqf: We haven't been doing so for this whole class

Sources/StreamChatUI/MessageActionsPopup/ChatMessageActionsVC.swift
nuno-vieira: Yes, right now with your current logic, you won't be able to mark them unread (since isPartOfThread is true)

nuno-vieira: I've updated the logic in the snippet, btw, it was incorrect

Tests/StreamChatUITests/SnapshotTests/ChatChannel/ChatChannelVC_Tests.swift
nuno-vieira: We should have done it 🤔  It is probably better to add it now, otherwise we will never do it. It is just a question of removing the `variants` argument, and it will add them automatically by default.

polqf: Recording in M1 has its issues, let's move this to another PR to speed it up

ipavlidakis (approved):
🚀

Sources/StreamChatUI/MessageActionsPopup/ChatMessageActionsVC.swift
```diff
@@ -105,7 +105,7 @@ open class ChatMessageActionsVC: _ViewController, ThemeProvider {
                 actions.append(threadReplyActionItem())
             }
 
-            if channelConfig.readEventsEnabled {
+            if channelConfig.readEventsEnabled && (!message.isPartOfThread || message.showReplyInChannel) {
```
nuno-vieira, line 108: There are missing tests here 🤔 We already have tests for `test_messageActions_`, we should add these ones as well ideally

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessagesCountDecorationView.swift
```diff
@@ -5,8 +5,8 @@
 import StreamChat
 import UIKit
 
-/// The date separator view that groups messages from the same day.
-open class ChatThreadRepliesCountDecorationView: ChatMessageDecorationView, AppearanceProvider {
+/// The view used to show a separator when there are unread messages.
+open class ChatMessagesCountDecorationView: ChatMessageDecorationView, AppearanceProvider {
     /// The date in string format.
     open var content: String? {
```
nuno-vieira, line 11: I know this comes from a different PR, but it would be nice if this was actually a `ChatMessage` instead of a String. It will make it hard for customers to customise this

Sources/StreamChatUI/MessageActionsPopup/ChatMessageActionsVC.swift
polqf: Added

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessagesCountDecorationView.swift
polqf: A ChatMessage? It should actually display the number of unread messages, not the message

polqf: I do agree with the above, though

ipavlidakis: I'm not sure I agree with the content being `ChatMessage`. This is a view that is being reused between Thread Replies Counter and Unread Messages. Currently the logic of what message to show is handled higher in the hierarchy from the responsible VC. Making the content `ChatMessage` it will require this view to take more responsibility and make it more complext (potentially it will need to be split into separate classes).

Regarding customers being able to customise it, we offer full control from the delegate (where this view is coming from)

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessagesCountDecorationView.swift
```diff
@@ -5,8 +5,8 @@
 import StreamChat
 import UIKit
 
-/// The date separator view that groups messages from the same day.
-open class ChatThreadRepliesCountDecorationView: ChatMessageDecorationView, AppearanceProvider {
+/// The view used to show a separator when there are unread messages.
+open class ChatMessagesCountDecorationView: ChatMessageDecorationView, AppearanceProvider {
```
nuno-vieira, line 9: I can see some issues with this view:
1. The doc comment is incorrect since it is also used for Thread replies.
2. We should not have logic in the names of the views. What I mean is that a customer might customize this and not even use the "count" of the messages. I've seen customers who just want to have a "New Messages" banner instead of displaying the messages count.
3. Having the content as a String does not provide context on what the content actually is. It also means that if you want to change the content, the customer will need to customize more than 1 view

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessagesCountDecorationView.swift
ipavlidakis: That being said, we should probably change the comment above as it's not a date anymore :P

nuno-vieira (changes requested):
Requesting changes for now only because QA is failing (The comments I provided previously are nits and not blocking for now)

When marking unread, the separator does not show:

https://user-images.githubusercontent.com/12814114/225342418-a887442b-6f55-4b2c-88f7-871b9d006680.mp4

When going to the unread message, the unread counter does not get updated:

https://user-images.githubusercontent.com/12814114/225342643-3724af29-5b69-46fd-b516-f3a086b32d6b.mp4

PR #2528, opened 2023-03-14
Feature/1 audio supporting components

ipavlidakis:
### 🔗 Issue Links

### 🎯 Goal

### 📝 Summary

### 🛠 Implementation

### 🎨 Showcase

| Before | After |
| ------ | ----- |
|  img   |  img  |

### 🧪 Manual Testing Notes

### ☑️ Contributor Checklist

### 🎁 Meme

### DIFF
--- a/Sources/StreamChatUI/Utils/AudioPlayer/AssetPropertyLoading.swift
+++ b/Sources/StreamChatUI/Utils/AudioPlayer/AssetPropertyLoading.swift
@@ -1,0 +1,106 @@
+//
+// Copyright © 2023 Stream.io Inc. All rights reserved.
+//
+
+import AVFoundation
+import Foundation
+
+/// Defines a type that represents the properties of an asset that can be request for loading
+public enum AssetProperty: String {
+    case duration
+}
+
+/// A simple type that describes the domain specific errors an instance of AssetPropertyLoading can return
+public enum AssetPropertyLoadingError: LocalizedError, Equatable {
+    case unknown
+    case cancelled
+
+    public var errorDescription: String? {
+        switch self {
+        case .unknown:
+            return "[\(type(of: self))] Asset's property loading failed with due to an unknown error"
+        case .cancelled:
+            return "[\(type(of: self))] Asset's property loading was cancelled"
+        }
+    }
+}
+
+/// A protocol that describes an object that can be used to load properties from an AVURLAsset
+public protocol AssetPropertyLoading {
+    /// A method that loads the property of an AVURLAsset asynchronously and
+    /// returns a result through a completion handler
+    ///
+    /// - Parameters:
+    /// - property: The property to load
+    /// - asset: The asset on which we will try to load the provided property
+    /// - onSuccessTransformer: If the load succeeds then the onSuccessTransformer
+    /// will be called, to provide the value to pass on the completion closure
+    /// - completion: The closure to call when we have final result to
+    /// report (success or failure)
+    func loadProperty<Value>(
+        _ property: AssetProperty,
+        of asset: AVURLAsset,
+        onSuccessTransformer: @escaping (AVURLAsset) -> Value,
+        completion: @escaping (Result<Value, Error>) -> Void
+    )
+}
+
+public struct StreamAssetPropertyLoader: AssetPropertyLoading {
+    public func loadProperty<Value>(
+        _ property: AssetProperty,
+        of asset: AVURLAsset,
+        onSuccessTransformer: @escaping (AVURLAsset) -> Value,
+        completion: @escaping (Result<Value, Error>) -> Void
+    ) {
+        asset.loadValuesAsynchronously(forKeys: [property.rawValue]) {
+            handlePropertyLoadingResult(
+                property,
+                of: asset,
+                onSuccessTransformer: onSuccessTransformer,
+                completion: completion
+            )
+        }
+    }
+
+    /// A private method that handles the result of loading the property of an AVURLAsset and returns a
+    /// result through a completion handler
+    private func handlePropertyLoadingResult<Value>(
+        _ property: AssetProperty,
+        of asset: AVURLAsset,
+        onSuccessTransformer: @escaping (AVURLAsset) -> Value,
+        completion: @escaping (Result<Value, Error>) -> Void
+    ) {
+        var error: NSError?
+        // Get the status of the loaded property and any associated error
+        let statusOfValue = asset.statusOfValue(
+            forKey: property.rawValue,
+            error: &error
+        )
+
+        /// Handle the status of the loaded property and call the appropriate completion handler
+        switch statusOfValue {
+        case .loading:
+            /// Do nothing if the property is still loading
+            break
+        case .loaded:
+            /// If the property has been loaded, apply the onSuccessTransformer to the asset and call
+            /// the completion handler with a success result
+            completion(.success(onSuccessTransformer(asset)))
+        case .cancelled:
+            /// If loading the property was cancelled, call the completion handler with a failure result
+            /// and the appropriate cancelled error
+            completion(.failure(AssetPropertyLoadingError.cancelled))
+        case .failed:
+            /// If loading the property failed, call the completion handler with a failure result and
+            /// the associated error (or an unknown error if there is none)
+            completion(.failure(error ?? AssetPropertyLoadingError.unknown))
+        case .unknown:
+            /// If the status of the loaded property is unknown, call the completion handler with a
+            /// failure result and an unknown error
+            completion(.failure(AssetPropertyLoadingError.unknown))
+        @unknown default:
+            /// Handle any future cases of the status of the loaded property
+            completion(.failure(AssetPropertyLoadingError.unknown))
+        }
+    }
+}
--- a/Sources/StreamChatUI/Utils/AudioPlayer/AudioPlaybackRate.swift
+++ b/Sources/StreamChatUI/Utils/AudioPlayer/AudioPlaybackRate.swift
@@ -1,0 +1,57 @@
+//
+// Copyright © 2023 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+// Defines an enumeration for audio playback rate with a raw value of Float
+public enum AudioPlaybackRate: Float, CustomStringConvertible {
+    case zero = 0
+    case half = 0.5
+    case normal = 1
+    case double = 2
+
+    public init?(rawValue: Float) {
+        // Switch statement that maps the rawValue to a case of AudioPlaybackRate
+        switch rawValue {
+        case 0:
+            self = .zero
+        case 0.5:
+            self = .half
+        case 1:
+            self = .normal
+        case 2:
+            self = .double
+        default:
+            self = .normal
+        }
+    }
+
+    // Computed property that returns a string representation of the playback rate
+    public var description: String {
+        switch self {
+        case .zero:
+            return "x0"
+        case .half:
+            return "x0.5"
+        case .normal:
+            return "x1"
+        case .double:
+            return "x2"
+        }
+    }
+
+    // Computed property that returns the next playback rate
+    public var next: AudioPlaybackRate {
+        switch self {
+        case .zero:
+            return .half
+        case .half:
+            return .normal
+        case .normal:
+            return .double
+        case .double:
+            return .half
+        }
+    }
+}
--- a/Sources/StreamChatUI/Utils/AudioPlayer/AudioPlaybackState.swift
+++ b/Sources/StreamChatUI/Utils/AudioPlayer/AudioPlaybackState.swift
@@ -1,0 +1,37 @@
+//
+// Copyright © 2023 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+// Defines an enumeration for audio player playback state
+public enum AudioPlaybackState: Equatable {
+    // Cases representing different playback states
+    case notLoaded
+    case loading
+    case paused
+    case playing
+    case stopped
+
+    // Equatable implementation for comparing two instances of
+    // AudioPlayerPlaybackState
+    public static func == (
+        lhs: Self,
+        rhs: Self
+    ) -> Bool {
+        switch (lhs, rhs) {
+        // Cases where both instances have the same case, return true
+        case
+            (.notLoaded, .notLoaded),
+            (.loading, .loading),
+            (.paused, .paused),
+            (.playing, .playing),
+            (.stopped, .stopped):
+            return true
+
+        // All other cases, return false
+        default:
+            return false
+        }
+    }
+}
--- a/Sources/StreamChatUI/Utils/AudioPlayer/AudioPlayerObserving.swift
+++ b/Sources/StreamChatUI/Utils/AudioPlayer/AudioPlayerObserving.swift
@@ -1,0 +1,132 @@
+//
+// Copyright © 2023 Stream.io Inc. All rights reserved.
+//
+
+import AVFoundation
+import Foundation
+
+/// Describes and object that can provide AVPlayer related updates.
+public protocol AudioPlayerObserving {
+    /// Registers an observer with the player that will receive updates when the player's
+    /// ``timeControlStatus`` gets updated.
+    /// - Parameters:
+    /// player: The player from which we would like to received updates
+    /// block: The block to call once a ``timeControlStatus`` update occurs
+    func addTimeControlStatusObserver(
+        _ player: AVPlayer,
+        using block: @escaping (AVPlayer.TimeControlStatus?) -> Void
+    )
+
+    /// Registers an observer that will periodically invoke the given block during playback to
+    /// report changing time.
+    /// - Parameters:
+    /// player: The player from which we would like to received updates
+    /// interval: The interval at which we would like to receive updates
+    /// queue: The queue on which the call to block will occur every time there is an update
+    /// block: The block to call every time there is an update
+    func addPeriodicTimeObserver(
+        _ player: AVPlayer,
+        forInterval interval: CMTime,
+        queue: DispatchQueue?,
+        using block: @escaping () -> Void
+    )
+
+    /// Registers and observer that will be called once the playback of an item stops
+    /// - Parameters:
+    /// queue: The queue on which the ``stoppedPlaybackHandler`` will be dispatched
+    /// block: The block to call once a player's item has stopped
+    func addStoppedPlaybackObserver(
+        queue: OperationQueue?,
+        using block: @escaping (AVPlayerItem) -> Void
+    )
+}
+
+open class StreamPlayerObserver: AudioPlayerObserving {
+    /// An observer token that is used to observe the ``timeControlStatus`` of the player to receive
+    /// updates when the state changes between playing and paused.
+    private var timeControlStatusObserver: NSKeyValueObservation?
+
+    /// A token referencing the periodicTimer that is registered on the player and is used to provide
+    /// time-related metadata updates.
+    private var periodicTimeObservationToken: Any?
+
+    /// This block is called on deinit to remove any registered periodicTimeObserver
+    ///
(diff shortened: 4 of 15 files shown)


Sources/StreamChatUI/Utils/AudioPlayer/AudioPlaybackRate.swift
```diff
@@ -0,0 +1,57 @@
...
+        default:
+            self = .normal
+        }
+    }
+
+    // Computed property that returns a string representation of the playback rate
+    public var description: String {
+        switch self {
+        case .zero:
+            return "x0"
```
martinmitrevski, line 34: These would be presented to the users right? Not sure if it makes sense to be localizable?

Sources/StreamChatUI/Utils/AudioPlayer/AudioPlaybackRate.swift
```diff
@@ -0,0 +1,57 @@
+//
+// Copyright © 2023 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+// Defines an enumeration for audio playback rate with a raw value of Float
+public enum AudioPlaybackRate: Float, CustomStringConvertible {
```
martinmitrevski, line 8: Not sure if an enum is the best representation here - we would need to always add additional cases, for example if customers ask for 1.25 or 1.5 (the more common ones).

Sources/StreamChatUI/Utils/AudioPlayer/AudioPlayerObserving.swift
```diff
@@ -0,0 +1,132 @@
...
+    /// block: The block to call once a ``timeControlStatus`` update occurs
+    func addTimeControlStatusObserver(
+        _ player: AVPlayer,
+        using block: @escaping (AVPlayer.TimeControlStatus?) -> Void
+    )
+
+    /// Registers an observer that will periodically invoke the given block during playback to
+    /// report changing time.
+    /// - Parameters:
+    /// player: The player from which we would like to receive updates
```
martinmitrevski, line 23: Doc formatting is not good at several places - we might have trouble generating the DocC docs.

Sources/StreamChatUI/Utils/AudioPlayer/AudioPlayerObserving.swift
```diff
@@ -0,0 +1,132 @@
+//
+// Copyright © 2023 Stream.io Inc. All rights reserved.
+//
+
+import AVFoundation
+import Foundation
+
+/// Describes an object that can provide AVPlayer related updates.
+public protocol AudioPlayerObserving {
```
martinmitrevski, line 9: No action for now, just food for thought: If we decide reusing this component in SwiftUI, we should probably also expose publishers.

Sources/StreamChatUI/Utils/AudioPlayer/AudioPlayerObserving.swift
```diff
@@ -0,0 +1,132 @@
...
+}
+
+open class StreamPlayerObserver: AudioPlayerObserving {
+    /// An observer token that is used to observe the ``timeControlStatus`` of the player to receive
+    /// updates when the state changes between playing and paused.
+    private var timeControlStatusObserver: NSKeyValueObservation?
+
+    /// A token referencing the periodicTimer that is registered on the player and is used to provide
+    /// time-related metadata updates.
+    private var periodicTimeObservationToken: Any?
```
martinmitrevski, line 51: Curious why are these `Any`?

Sources/StreamChatUI/Utils/AudioPlayer/AudioPlayerObserving.swift
```diff
@@ -0,0 +1,132 @@
...
+    ) {
+        stoppedPlaybackObservationToken = notificationCenter.addObserver(
+            forName: NSNotification.Name.AVPlayerItemDidPlayToEndTime,
+            object: nil,
+            queue: queue
+        ) { notification in
+            guard
+                let playerItem = notification.object as? AVPlayerItem
+            else {
+                return
```
martinmitrevski, line 127: Is this case possible? Would there be any implications to the UI layer?

Sources/StreamChatUI/Utils/Debouncer.swift
```diff
@@ -0,0 +1,54 @@
+//
+// Copyright © 2023 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+public protocol Debouncing {
```
martinmitrevski, line 7: I think we already had some logic with debouncing, @nuno-vieira?

Sources/StreamChatUI/Utils/AudioPlayer/AudioPlaybackState.swift
```diff
@@ -0,0 +1,37 @@
...
+    // Cases representing different playback states
+    case notLoaded
+    case loading
+    case paused
+    case playing
+    case stopped
+
+    // Equatable implementation for comparing two instances of
+    // AudioPlayerPlaybackState
+    public static func == (
```
martinmitrevski, line 18: is this really needed? Also, the doc says `AudioPlayerPlaybackState`, shouldn't it be `AudioPlaybackState`?

martinmitrevski (commented):
great stuff @ipavlidakis 👏  Added few comments, let me know what you think.

Sources/StreamChatUI/Utils/Debouncer.swift
nuno-vieira: We don't have it implemented yet, it was proposed here: https://github.com/GetStream/ios-issues-tracking/issues/217. Debouncing has some gotchas as far as I remember, worth comparing the implementations 👍

nuno-vieira, line 7: I guess this one can actually be internal, WDYT? 🤔

Sources/StreamChatUI/Utils/Debouncer.swift
```diff
@@ -0,0 +1,54 @@
...
+    // Method to debounce the handler closure.
+    func debounce(
+        _ handler: @escaping Handler
+    )
+
+    // Method to cancel any pending debounce calls.
+    func cancel()
+}
+
+open class Debouncer: Debouncing {
```
nuno-vieira, line 20: This one should not be `open` for sure, a debouncer is a debouncer and should not have any customizability 👍

nuno-vieira (commented):
Looks good overall 👌 Added some feedback, but it is in the right direction 🙂

Sources/StreamChatUI/Utils/AudioPlayer/AudioPlayerObserving.swift
ipavlidakis: Those are KVO/AVFoundation observation tokens (like subscription receipts on Combine) and the frameworks are actually returning `Any`. As their aim is to provide a strong reference to the observer so it won't be deallocated, we don't need to interact with them, just hold them. Once the token is deallocated, the observer will also be deallocated, invalidating the observation flow.

Sources/StreamChatUI/Utils/AudioPlayer/AudioPlayerObserving.swift
ipavlidakis: As this notification can be fired from anyone, including AVPlayer, I would say it's possible. I don't foresee any UI implications as if it's not our player firing it, we will ignore it.

Sources/StreamChatUI/Utils/Debouncer.swift
ipavlidakis: Comparing the implementation with what?

Sources/StreamChatUI/Utils/AudioPlayer/AudioPlaybackState.swift
ipavlidakis: just a remnant that needed clear up

Sources/StreamChatUI/Utils/AudioPlayer/AssetPropertyLoading.swift
```diff
@@ -0,0 +1,106 @@
...
+// Copyright © 2023 Stream.io Inc. All rights reserved.
+//
+
+import AVFoundation
+import Foundation
+
+/// Defines a type that represents the properties of an asset that can be loaded
+public enum AssetProperty: String {
+    case duration
+}
```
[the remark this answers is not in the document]
ipavlidakis: No worries, I figured out a type-safe way that doesn't require the introduction of a new type

Sources/StreamChatUI/Utils/AudioPlayer/AudioPlayerObserving.swift
ipavlidakis: The shortcut doesn't work for me. What exactly is it generating?

Sources/StreamChatUI/Utils/AudioPlayer/AssetPropertyLoading.swift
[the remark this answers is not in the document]
ipavlidakis: And i had to remove it :P to implement the multiple properties loading

Sources/StreamChatUI/Utils/AudioPlayer/AssetPropertyLoading.swift
```diff
@@ -0,0 +1,106 @@
...
+    ) {
+        asset.loadValuesAsynchronously(forKeys: [property.rawValue]) {
+            handlePropertyLoadingResult(
+                property,
+                of: asset,
+                onSuccessTransformer: onSuccessTransformer,
+                completion: completion
+            )
+        }
+    }
```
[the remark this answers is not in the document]
ipavlidakis: That was a tricky one. So for audio the duration is basically what we need. My idea was to implement just the bare minimum but after your feedback i tried to implement it and i realised that this will actually require work and potentially a change on the API. So, even if now we don't need it (and we may never really use it) i updated the code to load an array of properties.

Sources/StreamChatUI/Utils/AudioPlayer/AudioPlaybackRate.swift
```diff
@@ -0,0 +1,57 @@
...
+        case .zero:
+            return .half
+        case .half:
+            return .normal
+        case .normal:
+            return .double
+        case .double:
+            return .half
+        }
+    }
```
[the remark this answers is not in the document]
ipavlidakis: This logic will also be moved closer to the UI either on the audiView itself or the delegate

Sources/StreamChat/AudioPlayer/AssetPropertyLoading.swift
```diff
@@ -0,0 +1,176 @@
...
+                error: &error
+            )
+
+            /// Handle the status of the loaded property and call the appropriate completion handler
+            switch statusOfValue {
+            case .loading:
+                /// Do nothing if the property is still loading
+                break
+            case .loaded:
+                /// If the property has been loaded, apply the onSuccessTransformer to the asset and call
```
martinmitrevski, line 144: The comment here doesn't really match the code. Is it only "do nothing" in this case or sth is forgotten? If the first, we could also put it together with `loading` or maybe explain that an action would be executed after the switch.

Sources/StreamChat/AudioPlayer/AudioPlaybackRate.swift
```diff
@@ -0,0 +1,27 @@
...
+
+    public static func < (
+        lhs: AudioPlaybackRate,
+        rhs: AudioPlaybackRate
+    ) -> Bool {
+        lhs.rawValue < rhs.rawValue
+    }
+
+    /// SDK available playback rates
+    public static let zero = AudioPlaybackRate(rawValue: 0)
```
martinmitrevski, line 23: 👍 this is more flexible now

Sources/StreamChat/Utils/Debouncer.swift
```diff
@@ -0,0 +1,51 @@
+//
+// Copyright © 2023 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+protocol Debouncing {
```
martinmitrevski, line 7: Where is the debouncer going to be used? If it's not in the LLC, then we should probably move it to the UI layer, since it's not public (which is correct, we shouldn't expose this).

martinmitrevski (commented):
Almost there 👍 Just some small comments / questions. Also, don't forget to update the CHANGELOG.

ipavlidakis:
> Almost there 👍 Just some small comments / questions. Also, don't forget to update the CHANGELOG.

The PR doesn't really make any visible to customers changes as the components are not in use. My idea was to only update the CHANGELOG at the last PR. What would you add for this one?

Sources/StreamChat/AudioPlayer/AssetPropertyLoading.swift
ipavlidakis: Good catch. This is now updated

martinmitrevski:
> > Almost there 👍 Just some small comments / questions. Also, don't forget to update the CHANGELOG.
> 
> The PR doesn't really make any visible to customers changes as the components are not in use. My idea was to only update the CHANGELOG at the last PR. What would you add for this one?

There are new components that customers who don't use the UI layer can use. Question is if they can do something with it already. If yes, we can add something.

Sources/StreamChat/Utils/Debouncer.swift
ipavlidakis: My understanding is that the AudioPlayer will also be moved to the LLC. The PR is just not updated it yet

ipavlidakis:
> > > Almost there 👍 Just some small comments / questions. Also, don't forget to update the CHANGELOG.
> > 
> > 
> > The PR doesn't really make any visible to customers changes as the components are not in use. My idea was to only update the CHANGELOG at the last PR. What would you add for this one?
> 
> There are new components that customers who don't use the UI layer can use. Question is if they can do something with it already. If yes, we can add something.

As our SDK doesn't provide code/features that are using those components I'm not sure we should add anything in the CHANGELOG. If we do, we should also be able to provide examples of usage for those components, which we will be able to do with the next 2 PRs

Sources/StreamChatUI/Utils/Debouncer.swift
nuno-vieira: Nevermind 👍

Sources/StreamChat/Utils/Debouncer.swift
nuno-vieira: And do we actually need a debouncer for this? https://github.com/GetStream/stream-chat-swift/pull/2529#discussion_r1139203252

AFAIK AVFoundation is already quite forgiving if we do multiple seek requests, it should handle it automatically

Tests/StreamChatTests/Models/Attachments/AttachmentFileType_Tests.swift
```diff
@@ -0,0 +1,25 @@
...
+            .mp4,
+            .wav,
+            .ogg,
+            .m4a
+        ]
+
+        AttachmentFileType.allCases.forEach { subject in
+            XCTAssertEqual(subject.isAudio, expectedValues.contains(subject))
+        }
+    }
```
nuno-vieira, line 24: Worth testing the `returnsFalse` case as well. We should put an array of some types which are not Audio, and it should returns false

ipavlidakis: But we have expected values. It feels unnecessary and it will require to update the test even when isAudio hasn't changed

nuno-vieira: The thing here is that we are not testing all the branches. 
```swift
public var isAudio: Bool {
        switch self {
        case .mp3, .mp4, .wav, .ogg, .m4a:
            return true
        default:
            return false <--- Not tested
        }
    }
```

If in the future someone changes the implementation to this:
```swift
public var isAudio: Bool {
        return true
    }
```

The test would still pass

**Review by nuno-vieira — APPROVED** [review]
LGTM! ✅ Nice changes 👌

File: Tests/StreamChatTests/Models/Attachments/AttachmentFileType_Tests.swift (hunk 16 above)
Comment (ipavlidakis) [reply Tests/StreamChatTests/Models/Attachments/AttachmentFileType_Tests.swift:24]: Valid point. If you don't mind i'm going to update this test on the next PR

File: Sources/StreamChat/AudioPlayer/AudioPlaybackContext.swift [hunk 17]
```diff
@@ -0,0 +1,34 @@
+//
+// Copyright © 2023 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+/// A struct that represents the current state of an audio player
+public struct AudioPlaybackContext: Equatable {
+    /// The duration of the audio track in seconds
+    public var duration: TimeInterval
```
polqf, line 10: Why are these public vars?

martinmitrevski (approved):

Sources/StreamChat/AudioPlayer/AudioPlaybackRate.swift
```diff
@@ -0,0 +1,27 @@
...
+    public static let half = AudioPlaybackRate(rawValue: 0.5)
+    public static let normal = AudioPlaybackRate(rawValue: 1)
+    public static let double = AudioPlaybackRate(rawValue: 2)
```
polqf, line 26: WhatsApp has 1, 1.5 and 2, maybe it is best to copy them?

Sources/StreamChat/AudioPlayer/AudioPlayerObserving.swift
```diff
@@ -0,0 +1,136 @@
...
+    public init(
+        notificationCenter: NotificationCenter = .default
+    ) {
+        self.notificationCenter = notificationCenter
+    }
+
+    deinit {
+        timeControlStatusObserver?.invalidate()
+        periodicTimeObservationCancellationBlock?()
+        stoppedPlaybackObservationToken.map { notificationCenter.removeObserver($0) }
```
polqf, line 82: No need to remove `periodicTimeObservationToken`?

ipavlidakis: The `periodicTimeObservationToken` removal requires the player. In order to not hold a reference of it in the observer we are using the cancellationBlock which it does the following 
```
periodicTimeObservationCancellationBlock = { [weak player, periodicTimeObservationToken] in
            guard
                let player = player,
                let periodicTimeObservationToken = periodicTimeObservationToken
            else {
                return
            }
            player.removeTimeObserver(periodicTimeObservationToken)
        }
```

polqf: Oh, missed this part! Thanks!

Sources/StreamChat/AudioPlayer/AudioPlaybackContext.swift
ipavlidakis: `let` properties on classes make sense to allow for instance immutability. On structs, where immutability can be controlled at the caller's site, IMHO `let` properties are just limiting for the API and that's why i'm always starting with `var` and only changing to `let` when it's necessary.
That being said, i'm not strongly opinionated and if you think there are cases where allowing this struct to be mutable that can hurt our flows i'm happy to change the properties to `let` in a subsequent PR.

Tests/StreamChatTests/Models/Attachments/AttachmentFileType_Tests.swift
ipavlidakis: Actually, just realised it. Due to the assert used here
```
XCTAssertEqual(subject.isAudio, expectedValues.contains(subject))
```
any value that is not in the expectedValues should return false. This should be sufficient to cover all cases (including the one you mentioned above). I'll update the name on the next PR to be clearer

PR #2529, opened 2023-03-14
[Async Messages]Implementation of AudioPlayer

ipavlidakis:
### 🎯 Goal

This revision implements the AudioPlayer that will be used for the playback of async messages.

### 📝 Summary

In order to play async messages, we are using AVPlayer, which allows to stream audio from URLs. This PR is trying to hide as much as possible the complexity of dealing with media playback using AVPlayer, while providing a unified interface to communicate updates.

### 🛠 Implementation

We are introducing a protocol that describes an AudioPlayer and we provide concrete implementation (StreamAudioPlayer) that conforms to the new protocol and uses the components we introduced in [2528](https://github.com/GetStream/stream-chat-swift/pull/2528) . Finally this PR includes and extensive testing suite on the StreamAudioPlayer.

### ☑️ Contributor Checklist

### 🎁 Meme

meme


Sources/StreamChatUI/Utils/AudioPlayer/AudioPlayingDelegate.swift
```diff
@@ -0,0 +1,15 @@
+//
+// Copyright © 2023 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+/// An object that the audioPlayer can communicate updates regarding its context
+public protocol AudioPlayingDelegate: AnyObject {
+    /// The audioPlayer will calls this one to inform its delegate that the the playback's context has
+    /// been updated.
```
nuno-vieira, line 10: Some typos:
```suggestion
    /// The audioPlayer will call this one to inform its delegate that the playback's context has
    /// been updated.
```

Tip: You can turn on the spell checker in Xcode. It will help a lot 👌

nuno-vieira (commented):
Looking great 👌 Added a couple of questions and food for thought 🙂

Sources/StreamChatUI/Utils/AudioPlayer/AudioPlaying.swift
```diff
@@ -0,0 +1,383 @@
...
+    func playbackContext(for url: URL) -> AudioPlaybackContext
+
+    /// Instructs the player to load the asset from the provided URL and prepare it for streaming. If the
+    /// player's current item has as URL that matches the provided one, then we will try to play or restart
+    /// the playback while updating the new delegate.
+    /// - Parameters:
+    /// - url: The URL where the asset will be streamed from. If nil then the player will simply clear
+    /// up the current playback queue.
+    /// - delegate: The delegate that will be informed for changes on the asset's playback.
+    func loadAsset(from url: URL?, andConnectDelegate delegate: AudioPlayingDelegate)
```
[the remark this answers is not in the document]
ipavlidakis: As we are using a single player for everything, if we were to introduce multicast delegates we will need to add more logic to the view itself. For now (with the requirements we have) this doesn't seem to be necessary. Additionally, this player, is only going to be used for audio messages playback which don't really have many requirements (e.g. fast forward, rewind or even previews). For this reason I'm not sure there are going to be additional events added here. If there are, we can handle those in future itereations.
(Please note that even though we are using an AVPlayer here (the same that is being used for Video playback) our use cases are much simpler than video playback).

Sources/StreamChat/AudioPlayer/AudioPlaying.swift
```diff
@@ -0,0 +1,375 @@
...
+    /// The queue used to ensure thread-safe access to the context property
+    private lazy var contextAccessQueue: DispatchQueue = .init(
+        label: "com.getstream.audio.player.context", qos: .userInteractive
+    )
+    /// Describes the player's current playback state. The access to this property is **not** thread-safe
+    private var _context: AudioPlaybackContext = .notLoaded
+    /// Describes the player's current playback state. The access to this property is thread-safe
+    private(set) var context: AudioPlaybackContext {
+        get { contextAccessQueue.sync { _context } }
+        set { contextAccessQueue.sync { _context = newValue } }
```
polqf, line 62: You can use `async` with a barrier flag here

Sources/StreamChat/AudioPlayer/AudioPlaying.swift
```diff
@@ -0,0 +1,375 @@
...
+
+        playerObserver.addPeriodicTimeObserver(
+            player,
+            forInterval: interval,
+            queue: nil
+        ) { [weak self] in
+            guard
+                let self = self,
+                self.context.isSeeking == false
+            else {
```
polqf, line 121: Following the current codebase, should be something like
```suggestion
            guard let self = self, self.context.isSeeking == false else {
```

Sources/StreamChat/AudioPlayer/AudioPlaying.swift
```diff
@@ -0,0 +1,375 @@
...
+    ) {
+        var newContext = context
+        newContextProvider(&newContext)
+        context = newContext
+        delegate?.audioPlayer(self, didUpdateContext: context)
+    }
+
+    /// It's used by the assetPropertyLoader to handle the completion (successful or failed) of duration's
+    /// asynchronous loading.
+    private func handleDurationLoading(
```
polqf, line 201: Feels a bit unnatural that this method is defined before its usage (file placement wise)

Sources/StreamChat/AudioPlayer/AudioPlaying.swift
```diff
@@ -0,0 +1,375 @@
...
+
+        switch result {
+        /// If the assetPropertyLoaded managed to successfully load the asset's duration information
+        /// we update the context with the new information.
+        case let .success(asset):
+            player.replaceCurrentItem(with: .init(asset: asset))
+            updateContext { value in
+                value.duration = asset.duration.seconds
+                value.currentTime = 0
+                value.rate = .zero
```
polqf, line 219: Shouldn't you update `value.state`? Or you leave it up to `play` to do so?

Sources/StreamChat/AudioPlayer/AudioPlaying.swift
```diff
@@ -0,0 +1,375 @@
...
+        /// context with the notLoaded state in order to inform the delegate and we log a debug error message
+        case let .failure(error):
+            updateContext { value in
+                value.duration = 0
+                value.currentTime = 0
+                value.rate = .zero
+                value.state = .notLoaded
+                value.isSeeking = false
+            }
+            log.debug(error.localizedDescription)
```
polqf, line 235: Would be nice that you add a subsystem to the log (this function accepts `subsystems: LogSubsystem`)

Sources/StreamChat/AudioPlayer/AudioPlaying.swift
```diff
@@ -0,0 +1,375 @@
...
+        }
+    }
+
+    /// It executes a seek request at the specified time on the player in order to progress the playback.
+    private func executeSeek(
+        to time: TimeInterval
+    ) {
```
polqf, line 242: One line?

Sources/StreamChat/AudioPlayer/AudioPlaying.swift
```diff
@@ -0,0 +1,375 @@
...
+
+            self.delegate?.audioPlayer(self, didUpdateContext: self.context)
+        }
+    }
+
+    /// Provides thread-safe updates for the player's context and makes sure to forward any updates
+    /// to the the delegate
+    private func updateContext(
+        _ newContextProvider: (inout AudioPlaybackContext) -> Void
+    ) {
```
polqf, line 192: One line?

Sources/StreamChat/AudioPlayer/AudioPlaying.swift
```diff
@@ -0,0 +1,375 @@
...
+            let currentItemURL = (player.currentItem?.asset as? AVURLAsset)?.url,
+            currentItemURL == url
+        else {
+            return .notLoaded
+        }
+        return context
+    }
+
+    open func loadAsset(
+        from url: URL?,
```
polqf, line 289: Why is `url` optional here? Does not make much sense to pass a nil url

Sources/StreamChat/AudioPlayer/AudioPlaying.swift
ipavlidakis: In general I tried to keep update in their related context. For this one where we have basically loaded the asset's properties, we don't have enough context to assume confidently the state of the player. This responsibility is for the timeControlStatusObserver on the player.

Sources/StreamChat/AudioPlayer/AudioPlaying.swift
ipavlidakis: I had it initially but wasn't sure what's our "policy" for the subsystems. What do you thing about the name `audioPlayback`?

TestTools/StreamChatTestTools/Mocks/StreamChat/AudioPlayer/MockAudioPlayerDelegate.swift
```diff
@@ -0,0 +1,19 @@
+//
+// Copyright © 2023 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+import StreamChat
+
+public final class MockAudioPlayerDelegate: AudioPlayingDelegate {
+    public private(set) var didUpdateContextWasCalled: (player: AudioPlaying, context: AudioPlaybackContext)?
```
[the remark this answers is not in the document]
ipavlidakis: My idea here is to avoid having multiple variables. I'm strongly opinionated for this one and I'll change it to be consistent.

Sources/StreamChatUI/Utils/AudioPlayer/AudioPlaying.swift
```diff
@@ -0,0 +1,383 @@
...
+public protocol AudioPlaying: AnyObject {
+    /// Provides  way to get an instance of a player
+    static func build() -> AudioPlaying
+
+    /// Requests the playbackContext for the given URL. If the player's current item has as URL that
+    /// matches the provided one, it should return a context, otherwise it will return
+    /// ``AudioPlaybackContext.notLoaded``
+    /// - Parameters:
+    /// - url: The URL (provided by the asset) that is used to stream/download the content to play
+    func playbackContext(for url: URL) -> AudioPlaybackContext
```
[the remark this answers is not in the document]
nuno-vieira: But I guess it is fine for now 🤔

Sources/StreamChatUI/Utils/AudioPlayer/AudioPlaying.swift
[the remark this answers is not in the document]
nuno-vieira: I think it will be pretty common for a customer to want to observe the audio player events in a different UI component or any other component. If we add the multicastDelegate later, then we could potentially deprecate the API quickly since the delegate won't make sense. Either way, I'll leave it up to you 👍 If we can change this later whenever a customer needs it is also fine 👍

Sources/StreamChat/AudioPlayer/AudioPlaying.swift
```diff
@@ -0,0 +1,358 @@
...
+                playerItemURL == currentItemURL
+            else {
+                return
+            }
+            var newContext = self.context
+            newContext.state = .stopped
+            newContext.currentTime = 0
+            newContext.rate = .zero
+            newContext.isSeeking = false
+            self.context = newContext
```
nuno-vieira, line 272: Shouldn't this use the `updatedContext` function?

Sources/StreamChat/AudioPlayer/AudioPlaying.swift
```diff
@@ -0,0 +1,358 @@
...
+
+    /// Provides thread-safe updates for the player's context and makes sure to forward any updates
+    /// to the the delegate
+    private func updateContext(
+        _ newContextProvider: (inout AudioPlaybackContext) -> Void
+    ) {
+        var newContext = context
+        newContextProvider(&newContext)
+        context = newContext
+        delegate?.audioPlayer(self, didUpdateContext: context)
```
nuno-vieira, line 286: Can we move this delegate to the didSet of the `AudioPlaybackContext`?

Sources/StreamChat/AudioPlayer/AudioPlaying.swift
```diff
@@ -0,0 +1,358 @@
...
+
+            self.delegate?.audioPlayer(self, didUpdateContext: self.context)
+        }
+    }
+
+    /// Provides thread-safe updates for the player's context and makes sure to forward any updates
+    /// to the the delegate
+    private func updateContext(
+        _ newContextProvider: (inout AudioPlaybackContext) -> Void
+    ) {
```
nuno-vieira, line 282: I have one idea for this. It seems that it is quite important that we update the `context` through this function. But right now, nothing protects against someone manipulating the `context` property directly, which could cause issues in the future. Right now this is a bit error-prone. 

**Proposal:**
We create an `AudioPlaybackContextUpdater` component, which will have the `context` property private, and you can only change it through the `updateContext` function. WDYT?

Sources/StreamChatUI/Utils/AudioPlayer/AudioPlaying.swift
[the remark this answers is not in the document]
ipavlidakis: >  My concern is that if customers use this incorrectly and call this assuming this is the most up-to-date data, which might not be the case

The returned context will be the most up-to-date data, except if you are seeing a case that I don't see.

Sources/StreamChatUI/Utils/AudioPlayer/AudioPlaying.swift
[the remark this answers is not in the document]
ipavlidakis: I can't really see now (as there are not use cases) on what kind of other events customers may want to subscribe to. As this is currently only implemented for the playback of audiomessages and not audio, I would prefer to keep small and simple and closer to the defined needs rather than trying to guess potential usecases.

Sources/StreamChat/AudioPlayer/AudioPlaying.swift
ipavlidakis: Yes it should!

Sources/StreamChat/AudioPlayer/AudioPlaying.swift
ipavlidakis: I like the suggestion but i'm not sure about the responsibilities. Do you mind taking a look in the `ThreadSafeValueAccessor` and let me know your thoughts?

Sources/StreamChat/AudioPlayer/AudioPlaying.swift
nuno-vieira: > The blocking of unwanted threads is a potential issue in the concurrent queue with the barrier task, right?
I mean that using `sync` for writing it is. Writing usually is a fire-and-forget operation, there's no need to wait for this to be completed. 

> The audio player is one and tied to the MessageListVC and is being reused for every message and the events are only coming from the player and not outside. I don't see the need for a concurrent queue here.
Using a concurrent queue could be a premature optimization indeed, but if we stick to the Serial Queue, I would still use `async` when changing the context 🤔

Sources/StreamChat/AudioPlayer/AudioPlaying.swift
nuno-vieira: I would not introduce another abstraction for this now; we already have many of them in the SDK. 

But even with this new `ThreadSafeValueAccessor`, we guarantee thread-safeness. Still, we don't ensure that we use `updateContext()` every time so that you don't directly change the context to avoid unnecessary didUpdateContext events. 

As far as I understood, we use `updateContext()` here to only trigger the context's `didSet` after we change all the values we wanted, right? 

Ideally, we want to have a component that makes sure we don't forget any of these:

1. Update the context in a thread-safe manner
2. Don't forget to call the delegate when the context changes
3. Don't call the delegate more than necessary
    - Example, this should only trigger the delegate once instead of 3 times:
       ```swift
       context.state = .play
       context.rate = .half
       context.isSeeking = false
       ```

nuno-vieira: We could still have `ThreadSafeValueAccessor` and introduce this pattern in the SDK. But it looks to me that it still makes sense to have a `AudioPlaybackContextUpdater` to make sure those 3 requirements are met. And this component can use the  `ThreadSafeValueAccessor` inside it.

ipavlidakis: I also updated the `updateContext` and now the update to the delegate happens on didSet. Which means that there is no path you can take to update the context that is not thread safe or not updating the delegate. The `updateContext` remains as it provides an easy way to mutate (rather than having to do it inline whenever we need it to). I'll change the `ThreadSafeValueAccessor` and make it specific for the work we need it :)

nuno-vieira: Yes, but I can still change the `context` directly without using the `updateContext` right? That is why that having only one 1 to change the context would make sense https://github.com/GetStream/stream-chat-swift/pull/2529#discussion_r1146241196

Sources/StreamChat/AudioPlayer/AudioPlaying.swift
ipavlidakis: I see your point, and I can change the writing to use `async` but still on the Serial Queue.

Sources/StreamChat/AudioPlayer/AudioPlaying.swift
ipavlidakis: I don't see a need for both the `AudioPlaybackContextUpdater` and `ThreadSafeValueAccessor` TBH as in the way it is now you can't make mistakes in the future. I'm trying to not introduce a "bouncing" abstraction where the player passes itself and the delegate to the update which will then receive updates and communicate them as it seems overly complicated for the size of the object.

Isn't it something that we can do in the future if needed?

nuno-vieira: Okay makes sense, if we see issues in the future we can change it 👍

Sources/StreamChat/Utils/ThreadSafeValueAccessor.swift
```diff
@@ -0,0 +1,51 @@
...
+    private var _value: Value
+    var value: Value {
+        get { readValue() }
+        set { writeValue(newValue) }
+    }
+
+    init(
+        _ initialValue: Value,
+        with accessLevels: [AccessLevel] = [.read, .write],
+        queueLabel: String = "com.getstream.thread.safe.value.accessor.\(type(of: Value.self))",
```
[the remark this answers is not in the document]
nuno-vieira: Okay 👍 Maybe in the future, we can extract this pattern and spread it across the SDK if we think it makes sense 🙂 It could be a good replace of @Atomic 👍

PR #2537, opened 2023-03-21
Add support for the message list to start at the top

nuno-vieira:
### 🔗 Issue Links
Resolves https://github.com/GetStream/ios-issues-tracking/issues/352

### 🎯 Goal
Allow customers to configure the SDK, so the message list starts at the top.

### 📝 Summary
- Adds `Components.default.shouldMessagesStartAtTheTop`(Public)
- Adds `ChatMessageListView.adjustContentInsetToPositionMessagesAtTheTop()` (Internal)
- Removes forgotten `ChatThreadVC.appMovedToForeground()` (Internal)

### 🛠 Implementation
Because we use an inverted table view, we need to be a bit creative to support this 😄 The solution is based on changing the `tableView.contentInset` with the difference of `tableViewHeight - tableViewContentHeight`. For this we need to make sure we adjust the content inset for these 3 conditions:
- When the content is updated
- When the keyboard is shown/hidden
- When the screen orientation changes

### 🎨 Showcase

https://user-images.githubusercontent.com/12814114/226673329-f1618d8e-22a9-419d-806d-9acc669f368d.mp4

### 🧪 Manual Testing Notes
**Precondition:**
Open the configuration of the demo app and enable "shouldMessagesStartAtTheTop"

**Steps:**
- Open a Channel with a few messages
- The messages should be at the top
- Insert new messages
- The newly inserted message should be at the top
- Open and close the keyboard
- It should not have any UI glitch
- Change the orientation to landscape
- It should not have any UI glitch

### ☑️ Contributor Checklist


Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -184,12 +184,16 @@ open class ChatMessageListVC: _ViewController,
         listView.delegate = self
         listView.dataSource = self
         listView.reloadData()
+        DispatchQueue.main.async {
```
ipavlidakis, line 187:
```suggestion
        DispatchQueue.main.async { [weak self] in
```
nit: No reason to hold a strong reference here

Sources/StreamChatUI/ChatMessageList/ChatMessageListView.swift
```diff
@@ -185,13 +195,15 @@ open class ChatMessageListView: UITableView, Customizable, ComponentsProvider {
...
     ) {
+        adjustContentInsetToPositionMessagesAtTheTop()
         UIView.performWithoutAnimation {
             reloadMessages(
                 previousSnapshot: previousMessagesSnapshot,
                 newSnapshot: newMessagesSnapshot,
                 with: .fade,
                 completion: {
                     completion?()
+                    self.adjustContentInsetToPositionMessagesAtTheTop()
```
ipavlidakis, line 206: nit: same here, not really a reason to hold a strong reference here.

Sources/StreamChatUI/ChatMessageList/ChatMessageListView.swift
```diff
@@ -206,6 +218,32 @@ open class ChatMessageListView: UITableView, Customizable, ComponentsProvider {
...
     }
+
+    /// Adjusts the content inset so that messages are inserted at the top when there are few messages.
+    /// This is will be executed if the `Components.shouldMessagesStartAtTheTop` is enabled.
+    internal func adjustContentInsetToPositionMessagesAtTheTop() {
+        guard components.shouldMessagesStartAtTheTop else {
+            return
+        }
+
+        // If the height of message list is more than the content height
```
ipavlidakis, line 229:
```suggestion
        // If the height of message list is less than the content height
```

Sources/StreamChatUI/ChatMessageList/ChatMessageListView.swift
```diff
@@ -206,6 +218,32 @@ open class ChatMessageListView: UITableView, Customizable, ComponentsProvider {
...
+            contentInset.top = newContentInset
+            showsVerticalScrollIndicator = false
+            requiresContentInsetReset = true
+            // In case  we already removed the content inset, there's
+            // no need to do it every time.
+        } else if requiresContentInsetReset {
+            requiresContentInsetReset = false
+            contentInset.top = 0
+            showsVerticalScrollIndicator = true
+        }
```
ipavlidakis, line 245: nit: Just for readability purposes, can we add the `else` with `no-op` and probably `showsVerticalScrollIndicator = false`?

Sources/StreamChatUI/ChatThread/ChatThreadVC.swift
```diff
@@ -58,14 +59,6 @@ open class ChatThreadVC: _ViewController,
     override open func setUp() {
         super.setUp()
 
-        let notificationCenter = NotificationCenter.default
```
ipavlidakis, line 61: Why are we removing this?

Sources/StreamChatUI/Composer/ComposerKeyboardHandler.swift
```diff
@@ -93,6 +97,7 @@ open class ComposerKeyboardHandler: KeyboardHandler {
             options: UIView.AnimationOptions(rawValue: curve << 16)
         ) {
             composerParentView.layoutIfNeeded()
+            self.messageListVC?.listView.adjustContentInsetToPositionMessagesAtTheTop()
```
ipavlidakis, line 100: nit: Same here, strong self isn't necessary.

Tests/StreamChatUITests/SnapshotTests/ChatChannel/ChatChannelVC_Tests.swift
```diff
@@ -186,6 +186,26 @@ final class ChatChannelVC_Tests: XCTestCase {
...
+            messages: [
+                .mock(id: .unique, cid: .unique, text: "Hello", author: .mock(id: .unique)),
+                .mock(id: .unique, cid: .unique, text: "Cool", author: .mock(id: .unique))
+            ],
+            state: .localDataFetched
+        )
+
+        AssertSnapshot(
+            vc,
+            variants: [.smallDark] // This variant is the only one where the contentInset is stable between tests
```
ipavlidakis, line 205: What about `smallLight`? The contentInset shouldn't be different depending on darkMode, right?
Additionally, do you mind elaborating what's the issue with the other variants and we can't use them?

Tests/StreamChatUITests/SnapshotTests/ChatMessageList/ChatMessageListVC_Tests.swift
```diff
@@ -143,9 +144,9 @@ final class ChatMessageListVC_Tests: XCTestCase {
                 cache: nil
             )
             messageDTOWithoutCid.channel = nil
+            mockedMessageWithoutCid = try messageDTOWithoutCid.asModel()
```
ipavlidakis, line 147: What is this change affecting?

Tests/StreamChatUITests/SnapshotTests/ChatThread/ChatThreadVC_Tests.swift
```diff
@@ -78,6 +78,34 @@ final class ChatThreadVC_Tests: XCTestCase {
         )
     }
 
+    func test_whenShouldMessagesStartAtTheTopEnabled() {
```
ipavlidakis, line 81:
```suggestion
    func test_whenShouldMessagesStartAtTheTopIsTrue() {
```
As this is a boolean value, true/false is a more accurate match.

Tests/StreamChatUITests/SnapshotTests/ChatThread/ChatThreadVC_Tests.swift
```diff
@@ -78,6 +78,34 @@ final class ChatThreadVC_Tests: XCTestCase {
...
+            ],
+            state: .localDataFetched
+        )
+        messageControllerMock.simulate(state: .remoteDataFetched)
+
+        vc.view.layoutIfNeeded()
+
+        AssertSnapshot(
+            vc,
+            variants: [.smallDark]
```
ipavlidakis, line 105: Same as before, probably we can use at least `.smallLight` too

Sources/StreamChatUI/ChatMessageList/ChatMessageListView.swift
nuno-vieira: Not sure what you mean. Can you clarify?

Tests/StreamChatUITests/SnapshotTests/ChatMessageList/ChatMessageListVC_Tests.swift
nuno-vieira: This test was not stable, at least on my local machine; it was causing a crash. So now it should be good 👍

Tests/StreamChatUITests/SnapshotTests/ChatChannel/ChatChannelVC_Tests.swift
nuno-vieira: Sometimes, the contentInset does not remain 100% in the same position after the adjustment, and even in the demo app, we can see some minor little jumps when this happens. It looks like for the snapshot tests, this also happens, the adjustment is not 100% the same, which causes the snapshot test to fail between tests. I'm not sure why with small font, this does not happen, but it is what I found while trying it out.

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
nuno-vieira: Clearly there as no retain cycles here, but yes if self is already deallocated, it does not make sense for the closure to still run, so I will add the weak self 👍

Sources/StreamChatUI/ChatThread/ChatThreadVC.swift
nuno-vieira: This was forgotten from a previous PR as it says in the commit message, I just noticed now only

Tests/StreamChatUITests/SnapshotTests/ChatThread/ChatThreadVC_Tests.swift
nuno-vieira: We don't have a smallLight for now

Sources/StreamChatUI/ChatMessageList/ChatMessageListView.swift
ipavlidakis: I misread it. All good :)

Sources/StreamChatUI/ChatMessageList/ChatMessageListView.swift
ipavlidakis:
```suggestion
} else {
 // No-op
}
```

Tests/StreamChatUITests/SnapshotTests/ChatThread/ChatThreadVC_Tests.swift
ipavlidakis: Can you elaborate why we can't use the rest of the variants?

Tests/StreamChatUITests/SnapshotTests/ChatChannel/ChatChannelVC_Tests.swift
ipavlidakis: I see. Not sure if that requires further investigation to understand what's going on there. Maybe also an increased tolerance for the image comparison can help mitigate the issue and allow us to test on all variants.

Sources/StreamChatUI/ChatMessageList/ChatMessageListView.swift
ipavlidakis: It improves readability and explicitly defines all possible paths of the flow. It's a nice to have if we want to ensure that readability is maintained to high standards but I'm happy to leave it for another PR.

Tests/StreamChatUITests/SnapshotTests/ChatChannel/ChatChannelVC_Tests.swift
nuno-vieira: Unfortunately that does not work very well, and it slows down the speed of the snapshot tests if we decrease the precision :/ If a customer complains about these minor jitters, we can further investigate 👍 Right now they are barely noticeable 👍

ipavlidakis (approved):
🚢

PR #2543, opened 2023-03-23
Jump to unread messages - UI Component

polqf:
### 🔗 Issue Links

https://github.com/GetStream/ios-issues-tracking/issues/345

### 📝 Summary

Adds a component that will be used to display a banner when there are unread messages that are outside of the bounds of the screen

### 🎨 Showcase
Simulator Screen Shot - iPhone 14 Pro - 2023-03-23 at 09 31 12

### ☑️ Contributor Checklist


Sources/StreamChatUI/Appearance+ColorPalette.swift
```diff
@@ -45,6 +45,7 @@ public extension Appearance {
 
         public var messageCellHighlightBackground: UIColor = .streamYellowBackground
         public var pinnedMessageBackground: UIColor = .streamYellowBackground
+        public var jumpToUnreadMessagesBackground: UIColor = .streamGrayDisabledText
```
nuno-vieira, line 48: Maybe this is a bit clear? 🤔 WDYT?
```suggestion
        public var jumpToUnreadButtonBackground: UIColor = .streamGrayDisabledText
```

Sources/StreamChatUI/ChatMessageList/JumpToUnreadMessagesButton.swift
```diff
@@ -0,0 +1,73 @@
...
+        }
+    }
+
+    /// The view showing number of unread messages in channel if any.
+    open private(set) lazy var textLabel = UILabel()
+        .withoutAutoresizingMaskConstraints
+        .withBidirectionalLanguagesSupport
+        .withAdjustingFontForContentSizeCategory
+
+    open private(set) lazy var closeButton = UIButton().withoutAutoresizingMaskConstraints
```
nuno-vieira, line 23: I wonder if we could use a stack view to layout these 🤔 It would simplify the `setUpLayout()` code and it would also make it easier for customers to change the layout without overriding the `setUpLayout()` completely.

nuno-vieira:
Btw, right now, the button doesn't do anything, right? 🤔 Does it make sense to merge this as it is right now? This will mislead people, since they will click on it, and nothing will happen

Sources/StreamChatUI/ChatMessageList/JumpToUnreadMessagesButton.swift
polqf: I am pretty against having an Stack View that customers manually mutate, as it is something that depends on the internal layout, and does not scale well.

On the other hand, it is such a simple view, that maybe it is not worth it to allow customizability within it.

martinmitrevski: I'm with Pol on this one, there's no need to introduce a stack view here.

martinmitrevski (approved):
LGTM ✅ Will you merge it like this, or will you also add the logic for jumping here?

PR #2532, opened 2023-03-15
Don't mark as read if not onscreen

gsbernstein:
### 🔗 Issue Links

_Provide all Jira tickets and/or Github issues related to this PR, if applicable._

### 🎯 Goal

Bug:
1. put chat thread in a UITabBarController
2. switch tabs away from thread
3. receive a message on that thread

Expected: new messages not marked read until you switch back to the chat tab
Actual: new messages marked read without being seen

### 📝 Summary

Check if thread is onscreen before marking as read

### 🛠 Implementation

Window is nil when view is not part of the view hierarchy, so we can check this.

### 🎨 Showcase

_Add relevant screenshots and/or videos/gifs to easily see what this PR changes, if applicable._

| Before | After |
| ------ | ----- |
|  img   |  img  |

### 🧪 Manual Testing Notes

Performed in my app.

### ☑️ Contributor Checklist

- [x] I have signed the [Stream CLA](https://docs.google.com/forms/d/e/1FAIpQLScFKsKkAJI7mhCr7K9rEIOpqIDThrWxuvxnwUq2XkHyG154vQ/viewform) (required)
- [x] This change follows zero ⚠️ policy (required)
- [ ] This change should be manually QAed
- [ ] Changelog is updated with client-facing changes
- [ ] New code is covered by unit tests
- [ ] Comparison screenshots added for visual changes
- [ ] Affected documentation updated (docusaurus, tutorial, CMS)

PR #2544, opened 2023-03-23
Fix channel marking read when view is not visible

nuno-vieira:
### 🔗 Issue Links
https://github.com/GetStream/stream-chat-swift/pull/2532
https://github.com/GetStream/ios-issues-tracking/issues/357

### 🎯 Goal
Fixes channel being marked read when the view is not visible

### 🧪 Manual Testing Notes
- Login as User A
- In the channel list, slide the channel row and change the presentation style to TabBar
- Make sure to remain at the bottom of the channel
- Then select the dummy view (right view)
- With User B, send messages to the channel
- With User A, tap on the top left arrow to return to the Channel List
- User A should see the unread counter in the channel row

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
+++ b/Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
@@ -67,9 +67,16 @@ open class ChatChannelVC: _ViewController,
         messageListVC.listView.isLastCellFullyVisible
     }
 
+    internal var isViewVisible: ((ChatChannelVC) -> Bool) = { channelVC in
+        channelVC.viewIfLoaded?.window != nil
+    }
+
     /// A boolean value indicating whether it should mark the channel read.
     public var shouldMarkChannelRead: Bool {
-        isLastMessageFullyVisible && channelController.hasLoadedAllNextMessages && !hasMarkedMessageAsUnread
+        guard isViewVisible(self) else {
+            return false
+        }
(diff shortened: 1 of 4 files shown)


Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
```diff
@@ -67,9 +67,16 @@ open class ChatChannelVC: _ViewController,
         messageListVC.listView.isLastCellFullyVisible
     }
 
+    internal var isViewVisible: ((ChatChannelVC) -> Bool) = { channelVC in
```
ipavlidakis, line 70: nit: Why not a function? If it is only for tests, you can make it a function and add the vc.view on a dummy window in your tests

nuno-vieira: It makes the tests more simple and also a tiny bit faster most likely. IMO it is not worth to change it now.

Also, it can't be nil, otherwise we can't change it in the test 👍

gsbernstein:
I found a sort of corollary of this: the channel is still getting marked as read when the app is in the background. I added a check `guard UIApplication.shared.applicationState == .active else { return false }` to our override. Would you like me to make a ticket?

nuno-vieira:
> 

Hi @gsbernstein!

Yes please open a issue on our repo and provide reproduction steps 👍 

Best,
Nuno

PR #2551, opened 2023-03-28
Add E2E tests for bi-directional-scrolling

alteral:
### 🔗 Issue Links

Resolves https://github.com/GetStream/ios-issues-tracking/issues/361

### 🎯 Goal

- Add E2E tests for bi-directional-scrolling

### ☑️ Contributor Checklist

### 🎁 Meme

gif

### DIFF
--- a/StreamChatUITestsAppUITests/Robots/UserRobot+Asserts.swift
+++ b/StreamChatUITestsAppUITests/Robots/UserRobot+Asserts.swift
@@ -662,7 +662,7 @@ extension UserRobot {
 
     @discardableResult
     func assertQuotedMessage(
-        replyText: String = "",
+        replyText: String = "", // empty text by default for attachment messages
         quotedText: String,
         at messageCellIndex: Int? = nil,
         file: StaticString = #filePath,
--- a/StreamChatUITestsAppUITests/Robots/UserRobot+Asserts.swift
+++ b/StreamChatUITestsAppUITests/Robots/UserRobot+Asserts.swift
@@ -367,6 +367,40 @@ extension UserRobot {
         XCTAssertEqual(author, actualAuthor, file: file, line: line)
         return self
     }
+    
+    @discardableResult
+    func assertScrollToBottomButton(
+        isVisible: Bool,
+        file: StaticString = #filePath,
+        line: UInt = #line
+    ) -> Self {
+        var btn = MessageListPage.scrollToBottomButton
+        btn = isVisible ? btn.wait() : btn.waitForDisappearance()
+        XCTAssertEqual(isVisible,
+                       btn.exists,
+                       "Scroll to bottom button should be \(isVisible ? "visible" : "hidden")",
+                       file: file,
+                       line: line)
+        return self
+    }
+    
+    @discardableResult
+    func assertScrollToBottomButtonUnreadCount(
+        _ expectedCount: Int,
+        file: StaticString = #filePath,
+        line: UInt = #line
+    ) -> Self {
+        let unreadCount = MessageListPage.scrollToBottomButtonUnreadCount
+        let unreadCountShouldBeVisible = expectedCount > 0
+        if unreadCountShouldBeVisible {
+            XCTAssertEqual("\(expectedCount)", unreadCount.wait().label)
+        } else {
+            if unreadCount.exists {
+                XCTAssertEqual("\(expectedCount)", unreadCount.label)
+            }
+        }
+        return self
+    }
 
     @discardableResult
     func assertTypingIndicatorShown(
@@ -615,6 +649,12 @@ extension UserRobot {
         XCTAssertTrue(link.isHittable, "Link itself is not clickable")
         return self
     }
+    
+    @discardableResult
+    func waitForMessageVisibility(at messageCellIndex: Int) -> Self {
+        _ = messageCell(withIndex: messageCellIndex).waitForHitPoint()
+        return self
+    }
 }
 
 // MARK: Quoted Messages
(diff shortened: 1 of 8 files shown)


StreamChatUITestsAppUITests/Robots/UserRobot+Asserts.swift
```diff
@@ -367,6 +367,40 @@ extension UserRobot {
...
+        line: UInt = #line
+    ) -> Self {
+        let unreadCount = MessageListPage.scrollToBottomButtonUnreadCount
+        let unreadCountShouldBeVisible = expectedCount > 0
+        if unreadCountShouldBeVisible {
+            XCTAssertEqual("\(expectedCount)", unreadCount.wait().label)
+        } else {
+            if unreadCount.exists {
+                XCTAssertEqual("\(expectedCount)", unreadCount.label)
+            }
```
polqf, line 400: Shouldn't you XCFail if unreadCount does not exist?

alteral:
```swift
if unreadCountShouldBeVisible {
    XCTAssertEqual("\(expectedCount)", unreadCount.wait().label)
} else {
    if unreadCount.exists {
        XCTAssertEqual("\(expectedCount)", unreadCount.label)
    }
}
```

Maybe I did not get your point here, but it actually will fail if there is no `unreadCount` when it should be there. 
`Else` statement is a bit hackish here because in some scenarios, we still keep `unreadCount` in the hierarchy even though it equals to `0`. I discussed it with @nuno-vieira, it's not a big deal, but we will probably fix this in the future.

StreamChatUITestsAppUITests/Robots/UserRobot+Asserts.swift
```diff
@@ -633,6 +673,22 @@ extension UserRobot {
...
+        at messageCellIndex: Int? = nil,
+        file: StaticString = #filePath,
+        line: UInt = #line
+    ) -> Self {
+        let messageCell = messageCell(withIndex: messageCellIndex, file: file, line: line)
+        let quotedMessage = attributes.quotedText(quotedText, in: messageCell).wait()
+        XCTAssertTrue(quotedMessage.exists, "Quoted message was not showed", file: file, line: line)
+        XCTAssertFalse(quotedMessage.isEnabled, "Quoted message should be disabled", file: file, line: line)
+        XCTAssertTrue(quotedMessage.isHittable, "Quoted message is not visible", file: file, line: line)
         return self
```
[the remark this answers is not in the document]
alteral: Oops, fair point! This is something I was playing with and, eventually, did not go that road.

Fixed now. So `assertQuotedMessage()` will be for quoted messages in general, and attachments can be tested separately as usual, e.g.: 

```swift
userRobot
        .assertQuotedMessage(quotedText: "some text")
        .assertFile(isPresent: true)
```

StreamChatUITestsAppUITests/Robots/UserRobot+Asserts.swift
```diff
@@ -615,24 +649,53 @@ extension UserRobot {
...
+    }
 }
 
 // MARK: Quoted Messages
 extension UserRobot {
 
     @discardableResult
     func assertQuotedMessage(
-        replyText: String,
+        replyText: String = "",
```
nuno-vieira, line 665: Why is this empty string by default ? 🤔

StreamChatUITestsAppUITests/Robots/UserRobot.swift
```diff
@@ -232,6 +231,18 @@ extension UserRobot {
         let messageCell = messageCell(withIndex: messageCellIndex)
         return tapOnMessage(messageCell)
     }
+    
+    @discardableResult
+    func tapOnRepliedMessage(_ text: String, at messageCellIndex: Int? = 0) -> Self {
```
nuno-vieira, line 236: Isn't this the correct naming? 🤔 Otherwise, it looks we are taping on the reply of the quoted message 
```suggestion
    func tapOnQuotedMessage(_ text: String, at messageCellIndex: Int? = 0) -> Self {
```

StreamChatUITestsAppUITests/Tests/QuotedReply_Tests.swift
```diff
@@ -0,0 +1,254 @@
...
+        }
+        THEN("user observes the reply in message list") {
+            userRobot
+                .assertQuotedMessage(replyText: quotedMessage, quotedText: parentMessage)
+                .assertScrollToBottomButton(isVisible: false)
+        }
+        WHEN("user taps on a replied message") {
+            userRobot.tapOnRepliedMessage(parentMessage, at: 0)
+        }
+        THEN("user is scrolled up to the parent message") {
```
nuno-vieira, line 49:
```suggestion
        GIVEN("user opens the channel") {
            backendRobot.generateChannels(count: 1, messagesCount: 20)
            userRobot.login().openChannel()
        }
        WHEN("user adds a quoted reply to participant message") {
            userRobot
                .scrollMessageListUp(times: 4)
                .replyToMessage(quotedMessage, messageCellIndex: messageCount - 1)
                .waitForMessageVisibility(at: 0)
        }
        THEN("user observes the quote reply in message list") {
            userRobot
                .assertQuotedMessage(replyText: quotedMessage, quotedText: parentMessage)
                .assertScrollToBottomButton(isVisible: false)
        }
        WHEN("user taps on a quoted message") {
            userRobot.tapOnRepliedMessage(parentMessage, at: 0)
        }
        THEN("user is scrolled up to the quoted message") {
```

The parent message is actually for thread replies. It seems we are confusing the two concepts here 🤔 

Thread reply != Quote reply

A thread reply has a parent message. A quote reply has a quoted message id, it does not have a parent message.

StreamChatUITestsAppUITests/Tests/QuotedReply_Tests.swift
```diff
@@ -0,0 +1,254 @@
...
```
nuno-vieira, line 77: Same here 👍

StreamChatUITestsAppUITests/Tests/QuotedReply_Tests.swift
```diff
@@ -0,0 +1,254 @@
...
```
nuno-vieira, line 105: Same here

StreamChatUITestsAppUITests/Tests/QuotedReply_Tests.swift
```diff
@@ -0,0 +1,254 @@
...
```
nuno-vieira, line 131: Same here and a couple more places

TestTools/StreamChatTestMockServer/Robots/ParticipantRobot.swift
```diff
@@ -201,13 +201,34 @@ public class ParticipantRobot {
         )
         return self
     }
+    
+    @discardableResult
+    public func replyToMessage(_ text: String, withAttachment: Bool, toLastMessage: Bool = true) -> Self {
```
nuno-vieira, line 206: quoteMessage or quoteReplyMessage. We can't name `reply` only because, we can confuse it with `threadReply` 
```suggestion
    public func quoteMessage(_ text: String, withAttachment: Bool, toLastMessage: Bool = true) -> Self {
```

nuno-vieira (commented):
The implementation looks good 👍 I just think we need to improve the naming. Right now it is easy to confuse **Thread Reply vs Quote Reply**.

StreamChatUITestsAppUITests/Robots/UserRobot+Asserts.swift
alteral: To use the same assert method when the message has no text (e.g. when you send an attachment without text)

alteral:
Hey @nuno-vieira, absolutely agree. It makes more sense with the naming you proposed. Please, check out the latest commit, I think I've fixed it everywhere 👍

StreamChatUITestsAppUITests/Robots/UserRobot+Asserts.swift
nuno-vieira: Maybe better to add a comment? 🤔

PR #2557, opened 2023-04-03
Fix deprecatedFallback getting called when auto-filtering is enabled

ipavlidakis:
### 🔗 Issue Links

Resolves https://github.com/GetStream/ios-issues-tracking/issues/369

### 📝 Summary

With the auto-filtering implementation the filtering priorities were set to (higher to lower):
- `filter` closure
- auto-filtering
- `deprecatedFallback`, which is using the controller's delegate to determine if channel should be linked or not with a controller.
 
When auto-linking is enabled and no filter closure is used, we should return immediately and skip calling the controller's delegate.

### 🛠 Implementation

Implemented the filtering hierarchy that was described in Summary.

### 🧪 Manual Testing Notes

- (with auto-filtering on) Test with a user that is not part of a public livestream that they can still access the channel.
- - (with auto-filtering off) Test with a user that is not part of a public livestream that they cannot access the channel.

### ☑️ Contributor Checklist

### 🎁 Meme

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## StreamChat
 
 ### 🐞 Fixed
-- Fix channel list auto-filtering issue in cases where we are querying for channels that the user may not be a member [#2557](https://github.com/GetStream/stream-chat-swift/pull/2557)
+- Fix channel disappearing when channel list auto-filtering is enabled and the user is not a member of the channel [#2557](https://github.com/GetStream/stream-chat-swift/pull/2557)
 
 ### 🔄 Changed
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 # Upcoming
+## StreamChat
+
+### 🐞 Fixed
+- Fix channel list auto-filtering issue in cases where we are querying for channels that the user may not be a member [#2557](https://github.com/GetStream/stream-chat-swift/pull/2557)
```
nuno-vieira, line 8: Maybe we should say what is the cause of the issue (otherwise, "issue" is a bit too generic)
```suggestion
- Fix channel disappearing when channel list auto-filtering is enabled and the user is not a member of the channel [#2557](https://github.com/GetStream/stream-chat-swift/pull/2557)
```

nuno-vieira (approved):
I just added a comment to improve the changelog, otherwise, LGTM! ✅

Issue #2431, opened 2023-01-04
Intermittent 401s on launch


AndrewSB:
sometimes on launch, i'm seeing this 401 error

![Screenshot 2023-01-03 at 7 20 02 PM](https://user-images.githubusercontent.com/3814772/210479434-6ecf96fe-fd98-4a3e-b492-fbe19b24b039.png)

i'm still able to view the thread list, and send messages, etc.
not sure if this is expected, but i'm surprised to see these 401s

this is on 4.25.0

polqf:
Hi @AndrewSB,

Are you waiting until a successful connection completion before fetching the channels?
By the error, it seems like you are trying to fetch the channels before connecting and having a valid token

AndrewSB:
i double checked our sequencing, we are waiting for the connection to open before fetching controllers. by debugging the network traffic, i've noticed something i didn't expect

![Screenshot 2023-01-05 at 1 18 07 PM](https://user-images.githubusercontent.com/3814772/211401571-3daebb84-a5a8-4ffe-90d0-2a6fe3024e44.png)

these are the stream requests made in first 10 seconds after the app launch. you can see that the first 4 succeed (they have the Authorization header in the request), then for some reason the 5th doesn't include that header

any ideas on why the SDK would be doing that?

nuno-vieira:
CC @polqf

polqf:
@AndrewSB , could you please share code snippets to see your flow from opening the app until it is presenting the channel list? I was testing it in our Demo App, and I can only see 1 call to `/channels`

AndrewSB:
hmm yeah that might be challenging since we use it at a bunch of places now. i can try, but easier yet - if i sent you the stream debug logs, could you tell from that?

if not, i can try to grok our integration to see where and when we're making our stream calls

polqf:
The problem with logs with an issue like this is that they will not be able to determine what triggers those calls. We can try with those, but it would be better if you can provide the actual code.

alteral:
Hi @AndrewSB,

As mentioned above, could you please share the actual code so we could proceed investigating the issue?

Best,
Alex

AndrewSB:
getting back to this now, i'm not exactly sure what code's making this happen, we have stream code in a couple places

i've put a breakpoint on APIClient.swift:178, to here's the trace before the SDK tries to make a request with empty headers (so no authorization header)

[...]

PR #2559, opened 2023-04-05
Prevent unauthorized requests from reaching the backend

polqf:
### 🔗 Issue Links

Relates to #2431 

### 🎯 Goal

Fix unexpected 401s produced at launch, while the chat is not yet fully connected.

### 📝 Summary

The issue was produce when trying to load resources, while  connecting to the chat. 
When that happened, and if the internet connection is not ideal, it might happen that while waiting for the token to be retrieved, we end up timeout'ing. 
In that case, we were mistakenly sending the request without authentication headers and that's returning an expected 401. 

### 🛠 Implementation

This PR forwards timeouts to the APIClient for it to handle it, instead of just returning a request without authorization that gets sent to the backend.
The APIClient will retry the request (up to a max retries limit), giving time for the connection to be fully stablished

### 🧪 Manual Testing Notes

TBD

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/APIClient/RequestEncoder.swift
+++ b/Sources/StreamChat/APIClient/RequestEncoder.swift
@@ -152,7 +152,8 @@ struct DefaultRequestEncoder: RequestEncoder {
                 }
                 completion(.success(updatedRequest))
             case let .failure(error as ClientError.WaiterTimeout):
-                // The APIClient will treat a waiter timeout differently than the other ones
+                // The receiver will treat a waiter timeout differently than the other ones, and that's why we are not
+                // masking it under missing token. The receiver should retry based on their own logic
                 completion(.failure(error))
             case .failure:
                 completion(.failure(missingTokenError))
@@ -188,7 +189,8 @@ struct DefaultRequestEncoder: RequestEncoder {
                     updatedRequest.url = try updatedRequest.url?.appendingQueryItems(["connection_id": connectionId])
                     completion(.success(updatedRequest))
                 case let .failure(error as ClientError.WaiterTimeout):
-                    // The APIClient will treat a waiter timeout differently than the other ones
+                    // The receiver will treat a waiter timeout differently than the other ones, and that's why we are not
+                    // masking it under missing token. The receiver should retry based on their own logic
                     throw error
                 case .failure:
                     throw missingConnectionIdError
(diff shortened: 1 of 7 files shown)


Sources/StreamChat/APIClient/RequestEncoder.swift
```diff
@@ -151,10 +151,9 @@ struct DefaultRequestEncoder: RequestEncoder {
                     updatedRequest.setHTTPHeaders(.jwtStreamAuth, .authorization(token.rawValue))
                 }
                 completion(.success(updatedRequest))
-            case .failure(_ as ClientError.WaiterTimeout):
-                // We complete with a success to account for the most probable case for the timeout: No connection.
-                // That way, when reaching the APIClient, we would properly report a connection error.
-                completion(.success(request))
+            case let .failure(error as ClientError.WaiterTimeout):
+                // The APIClient will treat a waiter timeout differently than the other ones
+                completion(.failure(error))
```
ipavlidakis, line 156: Why are we treating it differently?

Tests/StreamChatTests/APIClient/APIClient_Tests.swift
```diff
@@ -114,6 +114,39 @@ final class APIClient_Tests: XCTestCase {
...
+            encoder?.onEncodeRequestCall = nil // So that this block does not get called again
+            encoder?.encodeRequest = .failure(testError) // This is for the following execution
+            encoder?.encodeRequest_completion?(.failure(ClientError.WaiterTimeout())) // This is for the initial execution, which we want to complete now.
+        }
+
+        waitForExpectations(timeout: defaultTimeout)
+
+        XCTAssertEqual(result?.error, testError)
+        // Should retry once, as the waiter timeout is only returned in the first try
+        XCTAssertCall("encodeRequest(for:completion:)", on: encoder, times: 2)
```
ipavlidakis, line 146: I haven't seen this before. Nice!

ipavlidakis (approved):
A comment just for clarification. Other than that it looks great!

nuno-vieira (approved):
LGTM! ✅ (Just make sure tests are passing)

Sources/StreamChat/APIClient/RequestEncoder.swift
polqf: We are treating them differently in the sense that we are no masking the error, but returning the original one so that it can be properly handled by the receiver

polqf:
> @nuno-vieira (Just make sure tests are passing)

For some reason those are only failing in the CI. Just modified them a bit to see if this fixes the problem

PR #2560, opened 2023-04-05
Edge cases app

polqf:
### 📝 Summary

This app reproduces certain edge cases that are commonly happening to our integrators. Those include:

- A connection delay bigger than the timeout in our waiters (slow connection simulation)
- Starting the chat without a prior connect call
- Showing the channel list before connecting beforehand
- Logging in without a proper logout **reusing the same channel list**
   - With the same user
   - With a different user (:exclamation:️)
- Log out / log in dance **reusing the same channel list**
   - With the same user
   - With a different user (:exclamation:️)

(Bullets marked with:exclamation:are currently producing issues)

### 🎨 Showcase

Simulator Screen Shot - iPhone 14 Pro - 2023-04-05 at 12 10 25
Simulator Screen Shot - iPhone 14 Pro - 2023-04-05 at 12 10 28

### DIFF
--- a/Examples/EdgeCases/EdgeCasesChannelList.swift
+++ b/Examples/EdgeCases/EdgeCasesChannelList.swift
@@ -1,0 +1,50 @@
+//
+// Copyright © 2023 Stream.io Inc. All rights reserved.
+//
+
+import StreamChatUI
+import UIKit
+
+class EdgeCasesChannelList: ChatChannelListVC {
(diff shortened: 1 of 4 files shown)


nuno-vieira (approved):
LGTM! ✅ This should be useful 👌 It would be even nicer if this was part of the DemoApp, but I can see it would be a bit difficult :/

PR #2566, opened 2023-04-11
Add support for bidirectional scrolling inside threads

nuno-vieira:
### 🔗 Issue Links
Resolves https://github.com/GetStream/ios-issues-tracking/issues/343

### 🎯 Goal
Adding support for bidirectional scrolling inside threads to allow jumping to a mid-page in threads.

### 📝 Summary
- Add support for bidirectional scrolling inside Threads
- Add support for jumping to a message which is not in the message list inside Threads

### 🛠 Implementation
The implementation is very close to the one used in the Channel, but with the difference that instead of using pagination cursors, we use a boolean flag to control the pagination (`showInsideThreads`). (In the future, we could even migrate the Channel with this solution since it is much simpler to manage). The reason for this implementation is that the other approach would impact the main channel, and we could have messages disappearing from the main channel, especially the ones which are `showReplyInChannel = true`. 

### 🎨 Showcase

https://user-images.githubusercontent.com/12814114/231893223-6076466c-5e67-423e-9478-0fc370ac37ca.mp4

### 🧪 Manual Testing Notes
- [x]  When I click on a quoted message that is already on the list, it should just jump to it.
- [x]  When I jump to a message a yellow background should softly appear for a couple of milliseconds
- [x]  When I click on a quoted message that is not in the list, it should fetch that message and the page/fragment it belongs to and then jump to it.
- [x]  When I slowly scroll to the bottom after jumping to a message it should load the next messages.
- [x]  When I scroll up, it should load the oldest messages, just like the already implemented loading of previous messages.
- [x]  When I’m in the middle of a message list page, and I click on the “scroll-to-bottom” button, it should scroll to the most recent message, it should hide scroll to bottom button, but it should not load all the messages in between.
- [x]  When I send a message, and I’m in the middle of a message list page, should go to the bottom, and hide the scroll-to-bottom button.
- [x]  When I’m in the middle of the page, and I receive a message, it should not appear on UI
- [x]  When I’m in the middle of a message list page, I have new messages from other users, when I go to the first page, it should load the first page
- [x]  When I jump to a message, the API call fails to load the messages around it, it should not do anything, and it should not delete the current page.

### ☑️ Contributor Checklist


Sources/StreamChat/Workers/MessageUpdater.swift
```diff
@@ -192,6 +192,7 @@ class MessageUpdater: Worker {
                 extraData: extraData
             )
 
+            newMessageDTO.showInsideThread = true
```
martinmitrevski, line 195: Does this always have to be true? Does it impact the case when "send in channel" is true?

Sources/StreamChatUI/ChatThread/ChatThreadVC.swift
```diff
@@ -202,20 +197,14 @@ open class ChatThreadVC: _ViewController,
...
     open func chatMessageListVC(
         _ vc: ChatMessageListVC,
         willDisplayMessageAt indexPath: IndexPath
     ) {
-        if indexPath.row < messages.count - 10 {
-            return
-        }
-
-        loadPreviousMessages()
+        // No-op. By default this component is not interest this event,
```
martinmitrevski, line 206: The sentence here is not clear enough.

martinmitrevski (approved):
Looks great ✅

Sources/StreamChat/Workers/MessageUpdater.swift
nuno-vieira: No this only impacts when showing replies inside the thread 👍

testableapple (approved):
LGTM 🙌

PR #2570, opened 2023-04-17
Fix crash when getting unread count in invalid state

nuno-vieira:
### 🔗 Issue Links
Resolves https://github.com/GetStream/ios-issues-tracking/issues/271

### 🎯 Goal
It fixes a rare crash when getting an unread count from a logged-out/invalid user.

### 📝 Summary
- Fixes a crash when getting an unread count from an invalid user

### 🧪 Manual Testing Notes
Unit Tests have been added to cover this scenario. 

### ☑️ Contributor Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### 🐞 Fixed
 - Fix channel disappearing when channel list auto-filtering is enabled and the user is not a member of the channel [#2557](https://github.com/GetStream/stream-chat-swift/pull/2557)
 - Fix an issues which was causing the app to terminate when using a filter with the `in` operator and `cid` values [#2561](https://github.com/GetStream/stream-chat-swift/pull/2561)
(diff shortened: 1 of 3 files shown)


Sources/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -602,7 +602,9 @@ extension NSManagedObjectContext: MessageDatabaseSession {
         message.showReplyInChannel = showReplyInChannel
         message.quotedMessage = quotedMessageId.flatMap { MessageDTO.load(id: $0, context: self) }
 
-        message.user = currentUserDTO.user
+        if let user = currentUserDTO.user {
+            message.user = user
```
polqf, line 606: Here we have the same problem as `user` is not optional for MessageDTO

polqf (commented):
I would not go ahead and make `user` optional anywhere. I suspect the problem relies with a user being deleted, and `unreadCount` being an async/lazy getter.

Just adding the defensive code checking for `isValid` in `unreadCount` block should be enough. This has been the only place where we've been having the issue as far as I can remember

polqf (approved):

PR #2571, opened 2023-04-17
Make connect/disconnect safer when network is offline

nuno-vieira:
### 🔗 Issue Links

Resolves https://github.com/GetStream/ios-issues-tracking/issues/386

### 🎯 Goal
Fixes a crash only in the develop branch, which happens when logging out and in with an unstable network connection or the network is offline.

### 🧪 Manual Testing Notes
1. Open channel list
2. Turn off Wi-fi
3. Log out
4. Log in
5. Channel List will display an error to re-fetch the channel list
6. Turn on Wi-fi and at the same time, tap to re-fetch the channel list

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/APIClient/RequestEncoder.swift
+++ b/Sources/StreamChat/APIClient/RequestEncoder.swift
@@ -56,12 +56,16 @@ extension RequestEncoder {
             dispatchGroup.leave()
         }
 
-        let waitResult = dispatchGroup.wait(timeout: .now() + 5)
+        let waitResult = dispatchGroup.wait(timeout: .now() + returningResultTimeout)
         if waitResult == .timedOut {
             result = .failure(ClientError("Encoding request timed out. Endpoint: \(endpoint)"))
         }
 
         return try result.get()
+    }
+
+    private var returningResultTimeout: TimeInterval {
+        5
     }
 }
 
--- a/Sources/StreamChat/APIClient/RequestEncoder.swift
+++ b/Sources/StreamChat/APIClient/RequestEncoder.swift
@@ -45,16 +45,23 @@ extension RequestEncoder {
             subsystems: .httpRequests
         )
 
-        var result: Result<URLRequest, Error>?
-        encodeRequest(for: endpoint) { result = $0 }
-
-        log.assert(
-            result != nil,
-            "`encodeRequest` with `requiresConnectionId == false` should return immediately.",
-            subsystems: .httpRequests
+        var result: Result<URLRequest, Error> = .failure(
+            ClientError("Unexpected error. The result was not changed after encoding the request.")
         )
 
-        return try result!.get()
+        let dispatchGroup = DispatchGroup()
+        dispatchGroup.enter()
+        encodeRequest(for: endpoint) {
+            result = $0
+            dispatchGroup.leave()
+        }
+
+        let waitResult = dispatchGroup.wait(timeout: .now() + 5)
+        if waitResult == .timedOut {
+            result = .failure(ClientError("Encoding request timed out. Endpoint: \(endpoint)"))
+        }
+
+        return try result.get()
     }
 }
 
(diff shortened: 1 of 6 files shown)


Sources/StreamChat/APIClient/RequestEncoder.swift
```diff
@@ -45,16 +45,23 @@ extension RequestEncoder {
...
 
-        return try result!.get()
+        let dispatchGroup = DispatchGroup()
+        dispatchGroup.enter()
+        encodeRequest(for: endpoint) {
+            result = $0
+            dispatchGroup.leave()
+        }
+
+        let waitResult = dispatchGroup.wait(timeout: .now() + 5)
```
ipavlidakis, line 59: Why 5 seconds? Isn't it too much?

Sources/StreamChat/WebSocketClient/WebSocketClient.swift
```diff
@@ -65,12 +65,13 @@ class WebSocketClient {
         return pingController
     }()
 
-    private func createEngineIfNeeded(for connectEndpoint: Endpoint<EmptyResponse>) -> WebSocketEngine {
+    private func createEngineIfNeeded(for connectEndpoint: Endpoint<EmptyResponse>) throws -> WebSocketEngine {
```
ipavlidakis, line 68: `throws` probably isn't needed as you are handling the creation of the engine.

It may be better to actually keep the `throws` but remove all the internal `do...catch` and just allow the caller to handle it

Sources/StreamChat/WebSocketClient/WebSocketClient.swift
```diff
@@ -65,12 +65,13 @@ class WebSocketClient {
...
         let request: URLRequest
         do {
             request = try requestEncoder.encodeRequest(for: connectEndpoint)
         } catch {
-            fatalError("Failed to create WebSocketEngine with error: \(error)")
+            log.log(.debug, message: error.localizedDescription)
```
ipavlidakis, line 73: Why not `log.error(error.localizedDescription, subsystems: .websocket)`?

Sources/StreamChat/APIClient/RequestEncoder.swift
nuno-vieira: This is not going to happen ever. That is why I put a random value. But just as safe percussion, I put a timeout. But makes sense to extract this to a variable or just use the timeout from the ClientConfig. I'll improve this 👍

Sources/StreamChat/WebSocketClient/WebSocketClient.swift
nuno-vieira: The reason this is throwing is that I want to make it clear, when using this function, that we need to handle it if it errors. Otherwise, it will be easy just to do `engine = createEngineIfNeeded()`. This will be problematic because when we use this, we want to make sure we don't proceed with the logic afterwards.

The reason I'm catching the error is that I want to log it if this fails. But indeed, I can just rethrow the error from the catch statement, instead of wrapping it in `ClientError`

Sources/StreamChat/WebSocketClient/WebSocketClient.swift
nuno-vieira: I didn't want to pollute the console with an error that is recoverable. In rare cases, this will happen, but it won't have any impact. So it should not warn the customer that something is wrong.

nuno-vieira: But I'm also not sure on this one 🤔

Sources/StreamChat/WebSocketClient/WebSocketClient.swift
ipavlidakis: > I can just rethrow the error from the catch statement, instead of wrapping it in ClientError

That sounds good :)

Sources/StreamChat/WebSocketClient/WebSocketClient.swift
nuno-vieira: Just to be safe, and since this will only happen 99% of times, I will change to `.error` 👍

ipavlidakis: If we are not handling the error higher, we should at least log something so customers have something to pinpoint to a potential issue

polqf (approved):

PR #2572, opened 2023-04-18
Avoid Crash when accessing FetchCache with an unexecuted NSFetchRequest

polqf:
### 🔗 Issue Links

https://github.com/GetStream/ios-issues-tracking/issues/385

### 🎯 Goal

Fix a crash where accessing FetchCache internal storage would crash

### 📝 Summary

```
This fetch request (0x60000263cb60) was created with a string name (QueuedRequestDTO), and cannot respond to -entity until used by an NSManagedObjectContext
```
You cannot access NSFetchRequest’s `entity` property unless the request has been executed.
Because setting it as a key for a dictionary will access its hash value, and it internally uses `entity` to compute it, it crashes.
Same goes for the other crash we've seen, where it is calling isEqual and its internal implementation makes it crash.

### 🛠 Implementation

NSFetchRequest is wrapped before adding it to the internal storage of FetchCache. This wrapper has its own implementation of Hashable and Equatable, which makes it safer.

### 🧪 Manual Testing Notes

We could not find a way to reproduce it aside from Unit tests.

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/Repositories/OfflineRequestsRepository.swift
+++ b/Sources/StreamChat/Repositories/OfflineRequestsRepository.swift
@@ -40,20 +40,21 @@ class OfflineRequestsRepository {
     /// - If the request fails with a connection error -> The request is kept to be executed once the connection is back (we are not putting it back at the queue to make sure we respect the order)
     /// - If the request fails with any other error -> We are dismissing the request, and removing it from the queue
     func runQueuedRequests(completion: @escaping () -> Void) {
-        var pendingActions: [(String, Data)] = []
-
         let readContext = database.backgroundReadOnlyContext
-        readContext.performAndWait {
-            pendingActions = QueuedRequestDTO.loadAllPendingRequests(context: readContext).map {
-                ($0.id, $0.endpoint)
+        readContext.perform { [weak self] in
+            let requests = QueuedRequestDTO.loadAllPendingRequests(context: readContext).map { ($0.id, $0.endpoint) }
+            DispatchQueue.main.async {
+                self?.executeRequests(requests, completion: completion)
             }
         }
+    }
 
-        log.info("\(pendingActions.count) pending offline requests", subsystems: .offlineSupport)
+    private func executeRequests(_ requests: [(String, Data)], completion: @escaping () -> Void) {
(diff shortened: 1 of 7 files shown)


Sources/StreamChat/Repositories/OfflineRequestsRepository.swift
```diff
@@ -40,20 +40,21 @@ class OfflineRequestsRepository {
...
     func runQueuedRequests(completion: @escaping () -> Void) {
-        var pendingActions: [(String, Data)] = []
-
         let readContext = database.backgroundReadOnlyContext
-        readContext.performAndWait {
-            pendingActions = QueuedRequestDTO.loadAllPendingRequests(context: readContext).map {
-                ($0.id, $0.endpoint)
+        readContext.perform { [weak self] in
+            let requests = QueuedRequestDTO.loadAllPendingRequests(context: readContext).map { ($0.id, $0.endpoint) }
+            DispatchQueue.main.async {
```
nuno-vieira, line 46: Why do we dispatch in the main queue here?

polqf: Because we don't want to continue the execution in the NSManagedObjectContext's thread

nuno-vieira: But the `executeRequests` has a DispatchGroup which waits for all the requests. Which means we will block the main thread until all requests are made. Or Am I missing something?

nuno-vieira: I see... the requests are then executed in the different operation queue, but then dispatching it to the main queue also does not seem to do anything. Was there a problem before? I was this done?

polqf: The change here was only to avoid the performAndWait, which could be error prone, and I did not want to risk it as I saw a crash that could look like a deadlock. Because of its async nature, now I cannot wait on the method's callee thread, so to be safe I just resume on the main thread

nuno-vieira (approved):
LGTM! ✅

PR #2575, opened 2023-04-19
Fix wrong message text font when text accessibility is larger + Snapshot Testing Framework Update

nuno-vieira:
### 🔗 Issue Links
Resolves https://github.com/GetStream/ios-issues-tracking/issues/371

### 🎯 Goal
Fixes the message text font when the text accessibility setting is set to larger. 

This issue was introduced by this [PR](https://github.com/GetStream/stream-chat-swift/pull/2328), which we tried to fix the dynamic font size for the message text, but this actually causes the text font to be with an incorrect size after closing/opening the app. So we need to revert that change. 

For now, we don't know the root cause of why the font is not changed dynamically, and it only changes after closing the app. But there is a workaround to overcome this, and it is to override the `MessageContentView.defaultMessageFont` to always return a new UIFont instance (Ex: `UIFont.preferredFont(forTextStyle: .body)`).  More context [here](https://twitter.com/nunovieira_dev/status/1578440760361644032?s=20).

**Bonus:** Fixes the message timestamp label and checkmark delivery status issues when the font size is larger. 

### 📝 Summary
- Fixes the message text font when the text accessibility setting is set to larger.
- Fixes the timestamp label being clipped when the text font size is larger.
- Fixes the checkmark in the delivery status view being too small for larger text font sizes.
- Updates the Snapshot Testing framework to improve M1 and Intel SSs

### 🎨 Showcase

| Before | After |
| ------ | ----- |
|     |    |

### 🧪 Manual Testing Notes

1. Xcode > Open Developer Tool > Accessibility Inspector
2. Processes > Simulator
3. Tap on Settings button (Right top corner)
4. Set the Dynamic font to a really big font
5. Close and open the app
6. Open a channel
7. The font of the messages should have the exact size of the Input Placeholder

### ☑️ Contributor Checklist


CHANGELOG.md
```diff
@@ -19,6 +19,19 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
 - Add support for jumping to a message which is not in the message list inside Threads [#2566](https://github.com/GetStream/stream-chat-swift/pull/2566)
 
+### 🐞 Fixed
+- Fix the message text font when the text accessibility setting is set to large after closing the app [#2575](https://github.com/GetStream/stream-chat-swift/pull/2575)
+- Fix the timestamp label being clipped when the text font size is larger [#2575](https://github.com/GetStream/stream-chat-swift/pull/2575)
+
+### ⚠️ Known Issue
+When changing the text accessibility while the app is running, it won't affect the message cell text, only after closing the app. So this is a regression of [#2325](https://github.com/GetStream/stream-chat-swift/issues/2325). This regression is necessary, since the original fix did not work 100% as it changed the text font size after closing the app, causing the font to be too big when the text accessibility is set to large. If you still want to have the dynamic font size to be working while the app is running, there is a workaround. You need to return a new `UIFont` instance everytime in `ChatMessageContentView.defaultMessageFont`. Example:
```
polqf, line 27: 👏

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
```diff
@@ -99,6 +99,54 @@ open class ChatMessageContentView: _View, ThemeProvider, UITextViewDelegate {
...
+    }
+
+    /// The font used when rendering system messages.
+    open var systemMessageFont: UIFont {
+        appearance.fonts.caption1.bold
+    }
+
+    /// The default font when rendering the message text.
+    ///
+    /// **Note:** Because of an issue which we don't know yet the root cause, if you want
```
polqf, line 114: Don't we know that it is because of UIKit's internal caching system?

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
```diff
@@ -99,6 +99,54 @@ open class ChatMessageContentView: _View, ThemeProvider, UITextViewDelegate {
...
+    /// the message font to change dynamically (live) when changing the font in accessibility
+    /// settings, you need to override this property and return a new instance of `UIFont`. Do
+    /// not use the `appearance` config, a new instance is `UIFont` is required.
```
polqf, line 117:
```suggestion
    /// not use the `appearance` config, a new instance of `UIFont` is required.
```

polqf (commented):
Did we end up changing that justifies the change on the snapshots?

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
nuno-vieira: We don't 100% know yet. Because for the TimestampLabel this works fine :/

nuno-vieira:
> Did we end up changing that justifies the change on the snapshots?

Yes, the CheckmarkView now has a different size. It is a little a bit bigger now to improve the UI when text accessibility is bigger

martinmitrevski (approved):

CHANGELOG.md
```diff
@@ -21,6 +21,19 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
```
ipavlidakis, line 29:
```suggestion
When changing the text accessibility while the app is running, it will affect the message cell text, only after closing the app. This is a regression of [#2325](https://github.com/GetStream/stream-chat-swift/issues/2325). This regression is necessary, since the original fix did not work 100% as it changed the text font size after closing the app, causing the font to be too big when the text accessibility is set to large. If you still want to have the dynamic font size to be working while the app is running, there is a workaround. You need to return a new `UIFont` instance everytime in `ChatMessageContentView.defaultMessageFont`. Example:
```

ipavlidakis (approved):

PR #2577, opened 2023-04-20
Safer `connectUser()` by logging out the user first if needed

nuno-vieira:
### 🔗 Issue Links
Resolves https://github.com/GetStream/ios-issues-tracking/issues/282

### 🎯 Goal
Make the `connectUser` call safer by logging out the user in case the user was not logged out. (If the user is a new user only)

### 🛠 Implementation
The implementation is done in the `AuthenticationRepository`, which delegates the logging out functionality to the `ChatClient`. This is useful because only in the `AuthenticationRepository` we have a centralized function to connect the user, whereas if we did it directly in the `ChatClient`, we would need to implement this in multiple places. Besides that, it would be harder to test. With the current approach, we just need to mock the delegate to check if we logged out. 

fetch2

### 🎨 Showcase

https://user-images.githubusercontent.com/12814114/233457445-29aa52a8-c685-45e4-8fb0-fc224dab4fcb.mp4

### 🧪 Manual Testing Notes
Follow the steps in the showcase. The guest user should log in successfully, even after closing the app while logged in as a guest user. (This will log out the user without calling `chatClient.logout()`)

### ☑️ Contributor Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Fix crash when accessing FetchCache with an unexecuted NSFetchRequest [#2572](https://github.com/GetStream/stream-chat-swift/pull/2572)
 - Fix an issue which was blocking a Guest Authentication operation to retrieve a connection token [#2574](https://github.com/GetStream/stream-chat-swift/pull/2574)
 - Make connect/disconnect safer when network is offline [#2571](https://github.com/GetStream/stream-chat-swift/pull/2571)
+- Make connect by logging out the user first if needed [#2577](https://github.com/GetStream/stream-chat-swift/pull/2577)
 
 ## StreamChatUI
 ### ✅ Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Fix crash when accessing FetchCache with an unexecuted NSFetchRequest [#2572](https://github.com/GetStream/stream-chat-swift/pull/2572)
 - Fix an issue which was blocking a Guest Authentication operation to retrieve a connection token [#2574](https://github.com/GetStream/stream-chat-swift/pull/2574)
 - Make connect/disconnect safer when network is offline [#2571](https://github.com/GetStream/stream-chat-swift/pull/2571)
-- Make connect by logging out the user first if needed [#2577](https://github.com/GetStream/stream-chat-swift/pull/2577)
+- Make connect safer by logging out the user first if needed [#2577](https://github.com/GetStream/stream-chat-swift/pull/2577)
 
 ## StreamChatUI
 ### ✅ Added
--- a/Sources/StreamChat/ChatClient.swift
+++ b/Sources/StreamChat/ChatClient.swift
@@ -473,21 +473,8 @@ public class ChatClient {
 }
 
 extension ChatClient: AuthenticationRepositoryDelegate {
-    /// Clears state related to the current user to leave the client ready for another user
-    /// Will clear:
-    ///     - Background workers
-    ///     - References to active controllers
-    ///     - Database
-    /// - Parameter completion: A block to be executed when the process is completed. Contains an error if something went wrong
-    func clearCurrentUserData(completion: @escaping (Error?) -> Void) {
-        createBackgroundWorkers()
-
-        // Stop tracking active components
-        activeChannelControllers.removeAllObjects()
-        activeChannelListControllers.removeAllObjects()
-
-        // Reset all existing local data.
-        databaseContainer.removeAllData(force: true, completion: completion)
+    func logoutUser(completion: @escaping () -> Void) {
+        logout(completion: completion)
     }
 
     func didFinishSettingUpAuthenticationEnvironment(for state: EnvironmentState) {
--- a/Sources/StreamChat/ChatClient.swift
+++ b/Sources/StreamChat/ChatClient.swift
@@ -403,6 +403,10 @@ public class ChatClient {
     public func logout(completion: @escaping () -> Void) {
         authenticationRepository.logOutUser()
 
+        // Stop tracking active components
+        activeChannelControllers.removeAllObjects()
+        activeChannelListControllers.removeAllObjects()
+
         let group = DispatchGroup()
         group.enter()
         disconnect {
--- a/Sources/StreamChat/ChatClient.swift
+++ b/Sources/StreamChat/ChatClient.swift
@@ -369,9 +369,7 @@ public class ChatClient {
     /// Connects anonymous user
     /// - Parameter completion: The completion that will be called once the **first** user session for the given token is setup.
     public func connectAnonymousUser(completion: ((Error?) -> Void)? = nil) {
-        authenticationRepository.connectUser(
-            userInfo: nil,
-            tokenProvider: { $0(.success(.anonymous)) },
+        authenticationRepository.connectAnonymousUser(
             completion: { completion?($0) }
         )
     }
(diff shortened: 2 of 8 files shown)


Sources/StreamChat/Repositories/AuthenticationRepository.swift
```diff
@@ -135,8 +149,23 @@ class AuthenticationRepository {
...
+            self?.scheduleTokenFetch(isRetry: false, userInfo: userInfo, tokenProvider: tokenProvider, completion: completion)
+        }
+
+        let state = EnvironmentState(currentUserId: currentUserId, newUserId: userInfo?.id)
+
+        switch state {
+        case .firstConnection, .newToken:
+            handleTokenFetch()
+        case .newUser:
+            delegate?.logoutUser {
```
ipavlidakis, line 165: At this point when we execute `logoutUser` the tokenProvider will be set to the previous value and not the newly provided one. Can that cause an issue?

Sources/StreamChat/Repositories/AuthenticationRepository.swift
```diff
@@ -294,19 +306,9 @@ class AuthenticationRepository {
...
 
         let onTokenReceived: (Token) -> Void = { [weak self, weak connectionRepository] token in
-            self?.prepareEnvironment(userInfo: userInfo, newToken: token) { error in
-                // Errors thrown during `prepareEnvironment` cannot be recovered
-                if let error = error {
-                    onCompletion(error)
-                    return
-                }
-
-                // We manually change the `connectionStatus` for passive client
```
ipavlidakis, line 304: Shall we keep the comment?

ipavlidakis (commented):
Can we add some tests, please?

Sources/StreamChat/Repositories/AuthenticationRepository.swift
```diff
@@ -195,23 +219,11 @@ class AuthenticationRepository {
...
             setToken(token: newToken, completeTokenWaiters: false)
             delegate?.didFinishSettingUpAuthenticationEnvironment(for: state)
-
-            // Setting a new connection is not possible in connectionless mode.
-            guard connectionRepository.isClientInActiveMode else {
-                completion(ClientError.ClientIsNotInActiveMode())
-                return
-            }
-
-            connectionRepository.disconnect(source: .userInitiated) { [weak delegate, weak connectionRepository] in
```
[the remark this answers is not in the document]
nuno-vieira: We don't need this anymore because we log out first, which will call disconnect. So now this is redundant 👍

nuno-vieira:
> Can we add some tests, please?

Yes I will add them, this is still in Draft phase 👍

Sources/StreamChat/Repositories/AuthenticationRepository.swift
nuno-vieira: Which previous value? I'm not sure I got it 🤔 The tokenProvider will be the one from the argument, which is the most updated one

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Fix crash when accessing FetchCache with an unexecuted NSFetchRequest [#2572](https://github.com/GetStream/stream-chat-swift/pull/2572)
 - Fix an issue which was blocking a Guest Authentication operation to retrieve a connection token [#2574](https://github.com/GetStream/stream-chat-swift/pull/2574)
 - Make connect/disconnect safer when network is offline [#2571](https://github.com/GetStream/stream-chat-swift/pull/2571)
+- Make connect by logging out the user first if needed [#2577](https://github.com/GetStream/stream-chat-swift/pull/2577)
```
nuno-vieira, line 18:
```suggestion
- Make connect safer by logging out the user first if needed [#2577](https://github.com/GetStream/stream-chat-swift/pull/2577)
```

Sources/StreamChat/Repositories/AuthenticationRepository.swift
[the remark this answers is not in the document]
polqf: `connectionRepository?.updateWebSocketEndpoint(with: newToken, userInfo: userInfo)` is missing here

Sources/StreamChat/Repositories/AuthenticationRepository.swift
```diff
@@ -10,11 +10,25 @@ enum EnvironmentState {
...
+            self = .firstConnection
+            return
+        }
+        
+        if currentUserId == newUserId {
+            self = .newToken
+            return
+        }
+
+        self = .newUser
```
polqf, line 25: Wouldn't if/else be less error-prone? Probably easier to read too

Sources/StreamChat/ChatClient.swift
```diff
@@ -473,21 +477,8 @@ public class ChatClient {
...
 
 extension ChatClient: AuthenticationRepositoryDelegate {
-    /// Clears state related to the current user to leave the client ready for another user
-    /// Will clear:
-    ///     - Background workers
-    ///     - References to active controllers
-    ///     - Database
-    /// - Parameter completion: A block to be executed when the process is completed. Contains an error if something went wrong
-    func clearCurrentUserData(completion: @escaping (Error?) -> Void) {
-        createBackgroundWorkers()
```
polqf, line 483: We are missing this, right?

Sources/StreamChat/Repositories/AuthenticationRepository.swift
```diff
@@ -135,8 +149,23 @@ class AuthenticationRepository {
     ///   - userInfo:       The user information that will be created OR updated if it exists.
     ///   - tokenProvider:  The block to be used to get a token.
     func connectUser(userInfo: UserInfo?, tokenProvider: @escaping TokenProvider, completion: @escaping (Error?) -> Void) {
```
polqf, line 151: The biggest problem I see now is that we heavily depend on UserInfo to be filled. Otherwise, we will be constantly logging out on every session.

That's why the previous approach was using the newly fetched token to determine if things were the same or not 🙃

Sources/StreamChat/Repositories/AuthenticationRepository.swift
polqf: @ipavlidakis since we will be logging out, during that process we still want to keep the old one. Even if it should not be used, tbh 🙃

Sources/StreamChat/Repositories/AuthenticationRepository.swift
polqf: UserInfo is always passed when logging a regular user. New logic made it evident

Tests/StreamChatTests/Repositories/AuthenticationRepository_Tests.swift
```diff
@@ -407,7 +410,7 @@ final class AuthenticationRepository_Tests: XCTestCase {
         XCTAssertEqual(refreshTokenCompletionCalls, 1)
         XCTAssertEqual(delegate.newStateCalls, 3)
         XCTAssertEqual(delegate.newState, .newToken)
-        XCTAssertEqual(delegate.clearDataCalls, 1)
+        XCTAssertEqual(delegate.logoutCallCount, 2)
```
ipavlidakis, line 413: Is it expected to call logout twice?

Tests/StreamChatTests/Repositories/AuthenticationRepository_Tests.swift
```diff
@@ -946,17 +1018,17 @@ final class AuthenticationRepository_Tests: XCTestCase {
 
 private class AuthenticationRepositoryDelegateMock: AuthenticationRepositoryDelegate {
```
ipavlidakis, line 1019: nit: Conforming this one to Spy will make the count for calls more consistent

Tests/StreamChatTests/Repositories/AuthenticationRepository_Tests.swift
polqf: Good catch. We should only be logging out once, when the user changes. It is a test error, though

ipavlidakis (approved):

PR #2587, opened 2023-04-28
Add support for jumping to message inside thread from channel view

nuno-vieira:
### 🔗 Issue Links

Resolves https://github.com/GetStream/ios-issues-tracking/issues/384

### 🎯 Goal
Add support for jumping to messages inside the thread from the channel view.

**Bonus:** It refactors the messages' pagination state handling so that it is centralized in one component called `MessagesPaginationStateHandler`. It was already getting hard to manage the state in two separate places. Besides that, by moving this to the `Updaters`, the logic is centralized in only one place. For the Replies, this is the `MessageUpdater.loadReplies`, and for the channel, it is the `ChannelUpdater.update()`. With the new `MessagesPaginationStateHandling`, we also guarantee thread-safeness. 

**Note:** A separate PR for the refactor is a bit hard to manage since the refactor was important and used for some of the fixes encountered, but I'm still studying the possibility of separating the refactor in a different PR.

### 📝 Summary

TODO

### 🛠 Implementation

TODO

**Note:** This is still WIP. I will provide a deeper explanation soon, as well as writing proper tests.

### 🎨 Showcase

TODO

### 🧪 Manual Testing Notes
**ACs:**
- [x]  When I tap on quoted message, which is inside a thread reply, it should open the thread and jump to the thread reply
- [x]  When opening a thread from a thread reply which was also sent to the channel, it should open the thread and jump to the reply
- [x]  When I open a channel where the initial pagination is .around(replyId), it should open the channel in the parent message location, and open the thread and jump to the reply id

### ☑️ Contributor Checklist


martinmitrevski (commented):
had a quick look at the PR. It introduces several changes in the core parts, therefore we should:
- require at least 2 approvals before merging
- manual testing and approval from @testableapple 
- potentially also a common testing session
- also would be nice to provide more implementation details in the PR description

PR #2592, opened 2023-05-02
Jump to unread messages - No interaction

polqf:
### 🔗 Issue Links
https://github.com/GetStream/ios-issues-tracking/issues/345

### 🎯 Goal

Allow jumping to unread messages when the first unread message is not on screen.

### 🛠 Implementation

The whole logic around jump to unread and mark as read relies on `lastReadMessageId`. This is a value that is both sent and retrieved from backend. It determines the last message that the user has seen.

In order to adapt this to the necessities of the SDK, there is a bidirectional transformation from last read <-> first unread. Being the **first unread** the value that matters to the SDK. That way we can either scroll to the unread messages, or show a banner on top of that one to mark that the unread history starts from there.

Code context:
[getLastReadMessageId(firstUnreadMessageId: MessageId)](https://github.com/GetStream/stream-chat-swift/pull/2592/files/e6f6e181b9f7a55c131bbba29ca0f9072be4f15a#diff-6c459ba92353f66fbf686037e6921c360306e7238b4e83de2ac852c84b0c1c1aR1445)
[getFirstUnreadMessageId()](https://github.com/GetStream/stream-chat-swift/pull/2592/files/e6f6e181b9f7a55c131bbba29ca0f9072be4f15a#diff-6c459ba92353f66fbf686037e6921c360306e7238b4e83de2ac852c84b0c1c1aR1454)

### 🎨 Showcase

https://github.com/GetStream/stream-chat-swift/assets/7887319/4fa5eb49-a28b-4cce-8d46-478c43440ad0

### 🧪 Manual Testing Notes

1. Open a channel
2. Scroll to a message that is not on screen
3. Mark it as unread
4. Scroll to bottom

Expected result:
- A pill should be displayed on top showing the unread messages count.
- ~Tapping on the pill should jump to the first unread message~
- Tapping the X button should dismiss the pill

### ☑️ Contributor Checklist

### 🧱 Pending
- [ ] Update StreamChatUI Tests


Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
```diff
@@ -84,12 +85,19 @@ open class ChatChannelVC: _ViewController,
...
 
+    /// Determines if a messaged had been marked as unread in the current session
     private var hasMarkedMessageAsUnread: Bool {
-        channelController.firstUnreadMessageId != nil
+        channelController.markedAsUnread
     }
 
+    /// Unread messages count for the session.
+    /// It can either contain the initial unread count, or the one after marking the channel as read/unread
+    public var sessionUnreadCount: ChannelUnreadCount = .noUnread
```
nuno-vieira, line 95: Why do we need to copy the unread count now? Why don't we use the `channelController.channel?.unreadCount` directly? Can't this be a bit error-prone?

Also, if the `sessionUnreadCount` lives in the UI Component, does it mean the Unread Feature is a UI feature? Meaing, that the SwiftUI SDK won't get this out of the box?

Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
```diff
@@ -351,7 +363,10 @@ open class ChatChannelVC: _ViewController,
         footerViewForMessage message: ChatMessage,
         at indexPath: IndexPath
     ) -> ChatMessageDecorationView? {
-        nil
+        guard message.id == lastReadMessageId, messages.first?.id != message.id, let channel = channelController.channel else { return nil }
+        let footer = components.unreadMessagesCounterDecorationView.init()
+        footer.content = channel
+        return footer
```
nuno-vieira, line 369: I wonder if this is a breaking change 🤔 Since we are moving the view from a header to a footer 🤔 We already have some customers who are customizing the `unreadMessagesCounterDecorationView`, and are expecting to be rendered as a header, won't this be a problem?

Sources/StreamChatUI/ChatMessageList/ChatMessage/Decorators/ChatUnreadMessagesCountDecorationView.swift
```diff
@@ -19,7 +19,7 @@ open class ChatUnreadMessagesCountDecorationView: ChatMessageDecorationView, The
     override open func setUpLayout() {
         super.setUpLayout()
 
-        embed(messagesCountDecorationView, insets: .init(top: 8, leading: 0, bottom: 0, trailing: 0))
+        embed(messagesCountDecorationView, insets: .init(top: 0, leading: 0, bottom: 8, trailing: 0))
```
nuno-vieira, line 22: Especially this part, if a customer is also adding a padding, it could break the UI of the customer 🤔

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -84,6 +84,38 @@ open class ChatMessageListVC: _ViewController,
...
 
+    /// A button to scroll the collection view to the first unread message.
+    /// Visible when there are unread messages outside of the bounds of the screen.
+    open private(set) lazy var jumpToUnreadMessagesButton: JumpToUnreadMessagesButton = components
+        .jumpToUnreadMessagesButton
+        .init()
+        .withoutAutoresizingMaskConstraints
+
+    /// A Boolean value indicating whether jump to unread messages button is visible.
+    open var showJumpToUnreadMessages: Bool {
```
nuno-vieira, line 95: We usually follow this pattern for these kinds of things, good to have consistency here:
```suggestion
    open var isJumpToUnreadMessagesButtonVisible: Bool {
```

`show` as a prefix usually feels like an action and not a question/boolean

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -84,6 +84,38 @@ open class ChatMessageListVC: _ViewController,
...
+        guard let dataSource = dataSource,
+              dataSource.sessionUnreadCount.messages > 0 else {
```
nuno-vieira, line 97: I wonder if we can move this `sessionUnreadCount` to the LLC, or if we can't somehow reuse the `channel.unreadCount`?

Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
polqf: This is in the ChatChannelVC because it is UI logic. We want to get the value for unread messages as we enter the chat and, independently of whether we mark the channel as read, keep showing the "unread messages" banner.

`channelController.channel?.unreadCount` will be updated to 0 unread as the channel is marked as read upon entering it

Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
```diff
@@ -299,13 +313,22 @@ open class ChatChannelVC: _ViewController,
             }
         case is MarkUnreadActionItem:
             dismiss(animated: true) { [weak self] in
-                self?.channelController.markUnread(from: message.id)
+                self?.channelController.markUnread(from: message.id) { [weak self] error in
+                    guard error == nil else { return }
+                    self?.updateSessionUnreadCount()
+                }
```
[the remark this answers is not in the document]
polqf: As shared [here](https://github.com/GetStream/stream-chat-swift/pull/2592#discussion_r1185807412), getting an updated value for the channel's unread count is not what we want

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
polqf: Same as before, the logic is UI only, and it does not make sense to be part of the LLC

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -228,15 +267,43 @@ open class ChatMessageListVC: _ViewController,
...
+            for: jumpToUnreadMessagesButton,
+            isVisible: showJumpToUnreadMessages,
+            animated: animated
+        )
+    }
+
+    private func updateVisibility(for view: UIView, isVisible: Bool, animated: Bool) {
+        if isVisible { view.isVisible = true }
+        Animate(
+            duration: 2,
```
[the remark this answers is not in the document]
polqf: Might be a leftover from a test, good catch

Sources/StreamChatUI/ChatMessageList/ChatMessageListVCDataSource.swift
```diff
@@ -13,6 +13,10 @@ public protocol ChatMessageListVCDataSource: AnyObject {
     /// Asks the data source to return all the available messages.
     var messages: [ChatMessage] { get set }
 
+    /// Asks the data source to return the unread messages count for the session.
+    /// It can either contain the initial unread count, or the one after marking the channel as read/unread
+    var sessionUnreadCount: ChannelUnreadCount { get }
```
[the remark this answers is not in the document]
polqf: https://github.com/GetStream/stream-chat-swift/pull/2592#discussion_r1185807412

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -84,6 +84,38 @@ open class ChatMessageListVC: _ViewController,
...
+
+    /// A Boolean value indicating whether jump to unread messages button is visible.
+    open var showJumpToUnreadMessages: Bool {
+        guard let dataSource = dataSource,
+              dataSource.sessionUnreadCount.messages > 0 else {
+            return false
+        }
+
+        guard let lastReadIndexPath = lastReadMessageIndexPath else {
+            // If there are messages, but none is read, we show the button if the first message is not on screen.
```
ipavlidakis, line 102: Wouldn't the messages that are onscreen be marked as read?

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -254,16 +321,40 @@ open class ChatMessageListVC: _ViewController,
...
+        let previousLastReadId = lastReadMessageId
+        guard previousLastReadId != lastReadId else { return }
+
         func indexPath(for id: MessageId?) -> IndexPath? {
             id.flatMap(getIndexPath)
         }
 
-        let indexPathsToReload = [indexPath(for: previousId), indexPath(for: id)].compactMap { $0 }
+        lastReadMessageId = lastReadId
+        let indexPathsToReload = [indexPath(for: previousLastReadId), indexPath(for: lastReadId)].compactMap { $0 }
```
ipavlidakis, line 333: Can we  reuse the `getIndexPath(forMessageId:)` method rather than having a new one?

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -254,16 +321,40 @@ open class ChatMessageListVC: _ViewController,
...
+
         guard !indexPathsToReload.isEmpty else { return }
         listView.reloadRows(at: indexPathsToReload, with: .automatic)
     }
 
+    private func isMessageVisible(for messageId: MessageId) -> Bool {
```
ipavlidakis, line 339:
```suggestion
    private func isMessageVisible(_ messageId: MessageId) -> Bool {
```
I believe that `isMessageVisible for id` isn't very readable here. How do you feel about using `_` instead?

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -254,16 +321,40 @@ open class ChatMessageListVC: _ViewController,
...
+        guard let indexPath = getIndexPath(forMessageId: messageId) else { return false }
+        return isMessageVisible(for: indexPath)
+    }
+
+    private func isMessageVisible(for indexPath: IndexPath) -> Bool {
```
ipavlidakis, line 344:
```suggestion
    private func isMessageVisible(at indexPath: IndexPath) -> Bool {
```
Similar as above, just a suggestion to improve the signature readability.

Tests/StreamChatTests/Database/DTOs/ChannelReadDTO_Tests.swift
```diff
@@ -22,13 +22,41 @@ final class ChannelReadDTO_Tests: XCTestCase {
         super.tearDown()
     }
 
+    // MARK: - saveChannelRead
+
+    func test_saveChannelRead() throws {
```
ipavlidakis, line 27: Is there any chance we can include a few words about what the expected outcome is?

ipavlidakis (commented):
It looks great! Just added a few comments

ipavlidakis (commented):
It looks great! Just added a few comments

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
```diff
@@ -1434,6 +1441,53 @@ private extension ChatChannelController {
             self?.state = error == nil ? .localDataFetched : .localDataFetchFailed(ClientError(with: error))
         }
     }
+
+    private func getLastReadMessageId(firstUnreadMessageId: MessageId) -> MessageId? {
```
polqf, line 1445: Because the backend only provides and requires lastReadMessageId, the controller is the one in charge of making the conversion both ways.
From the SDK point of view, getting the first unread message is the most relevant piece of information

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
```diff
@@ -1434,6 +1441,53 @@ private extension ChatChannelController {
...
+
+        let lookUpStartIndex = messages.index(before: lastReadIndex)
+
+        var id: MessageId?
+        for index in (0...lookUpStartIndex).reversed() {
+            let message = message(at: index)
+            guard message?.author.id != client.currentUserId, message?.deletedAt == nil else { continue }
+            id = message?.id
+            break
+        }
```
polqf, line 1476: We will be selecting the oldest message starting from the last read that is not a message that is deleted or that is authored by the current user

Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
```diff
@@ -397,10 +421,15 @@ open class ChatChannelVC: _ViewController,
...
     // MARK: - ChatChannelControllerDelegate
 
+    private var firstMessagesUpdate: Bool = true
     open func channelController(
         _ channelController: ChatChannelController,
         didUpdateMessages changes: [ListChange<ChatMessage>]
     ) {
+        if firstMessagesUpdate {
+            updateUnreadMessagesRelatedComponents()
+            firstMessagesUpdate = false
```
polqf, line 431: Ideally this should not be needed, and we should be using the completion of the synchronize to set the state values for the unread messages related UI, but because the data is not there, we have to do this 🙃

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
polqf: We mark the channel as read only when all the unread messages are seen

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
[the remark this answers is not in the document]
polqf: Not to self: To be reviewed

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
polqf: Not to self: To be reviewed

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
polqf: Not to self: To be reviewed

Sources/StreamChatUI/ChatMessageList/ChatMessage/Decorators/ChatUnreadMessagesCountDecorationView.swift
```diff
@@ -25,7 +25,10 @@ open class ChatUnreadMessagesCountDecorationView: ChatMessageDecorationView, The
     override open func updateContent() {
         super.updateContent()
 
-        let unreadCount = content?.unreadCount.messages ?? 0
+        // Temporarily disabling unread counts as they are not 100% accurate all the time.
+        // Passing 0 will show "Unread messages" without a number
+        // let unreadCount = content?.unreadCount.messages ?? 0
+        let unreadCount = 0
```
ipavlidakis, line 31: Is there a plan to make this accurate? If yes, shall we have a ticket somewhere about it (and maybe a TODO)?

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -86,6 +86,39 @@ open class ChatMessageListVC: _ViewController,
         return (!listView.isLastCellFullyVisible && isMoreContentThanOnePage) || dataSource?.isFirstPageLoaded == false
     }
 
+    /// A button to scroll the collection view to the first unread message.
```
ipavlidakis, line 89: Are we not using a tableview? 🤔

polqf: I think I copy pasted it from some other value 🤔 Corrected it now

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -240,14 +280,44 @@ open class ChatMessageListVC: _ViewController,
...
-        let isVisible = isScrollToBottomButtonVisible
-        if isVisible { scrollToBottomButton.isVisible = true }
+        updateVisibility(
+            for: scrollToBottomButton,
+            isVisible: isScrollToBottomButtonVisible,
+            animated: animated
+        )
+    }
+
+    open func updateJumpToUnreadMessagesVisibility(animated: Bool = true) {
```
nuno-vieira, line 303: I think it important that we add the "Button" naming, so that is clear we are talking about the button. We can also add a document comment like the one in `updateScrollToBottomButtonVisibility`
```suggestion
    open func updateJumpToUnreadButtonVisibility(animated: Bool = true) {
```

Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
```diff
@@ -340,13 +358,19 @@ open class ChatChannelVC: _ViewController,
             }
         case is MarkUnreadActionItem:
             dismiss(animated: true) { [weak self] in
-                self?.channelController.markUnread(from: message.id)
+                self?.channelController.markUnread(from: message.id) { _ in
+                    self?.updateUnreadMessagesRelatedComponents()
```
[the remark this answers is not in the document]
polqf: Unread state is local only. Relying on WS events makes it super unreliable due to delays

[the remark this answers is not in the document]
nuno-vieira: Yes but DB changes, should trigger delegate events as well 🤔

Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
```diff
@@ -317,7 +329,13 @@ open class ChatChannelVC: _ViewController,
...
         willDisplayMessageAt indexPath: IndexPath
     ) {
-        // no-op
+        let message = chatMessageListVC(vc, messageAt: indexPath)
+        if message?.id == firstUnreadMessageId {
+            hasSeenAllUnreadMessages = true
+        }
+        if isLastMessageFullyVisible {
+            hasSeenLastMessage = true
+        }
```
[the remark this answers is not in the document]
polqf: > In order for this to be correct, we would actually need to verify if hasLoadedAllNextMessages is also true.

This is taken into consideration in `shouldMarkChannelRead`

Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
polqf: ChannelUpdater does not return data, and the observer has not yet received the updates

testableapple:
Nicely done 👍

The only concern I have is that, at the moment, I can't mark as unread old thread replies sent also in channel. Other than that LGTM ✅

PR #2595, opened 2023-05-05
Reuse messages pagination logic across Channel and Thread

nuno-vieira:
### 🔗 Issue Links
Resolves https://github.com/GetStream/ios-issues-tracking/issues/410

### 🎯 Goal
It refactors the logic for handling the messages' pagination so that it is centralized in one place and not spread across the Channel and the Thread separately. This will reduce the maintainability of this part of the codebase since it will add consistency between the Channel and the Thread, and also, it will be easier to read, test and modify.

#### MessagesPaginationStateHandler
In order to reuse the logic of the messages pagination handling, a new component was created, the `MessagesPaginationStateHandler`, which conforms to the `MessagesPaginationStateHandling` protocol. This component is responsible for all the messages' pagination logic. With this, we guarantee that the pagination will work the same for the Thread and the Channel. As a bonus, it is now easier to add thread safety to the messages' pagination state.

#### Moving the logic to the Updaters
Besides introducing this new component, we also moved the pagination handling to the `MessageUpdater` and the `ChannelUpdater` instead of the `MessageController` and the `ChannelController`, respectively. The reason for this is that the updaters have only one function to load the messages/replies, whereas the controllers have multiple functions (`loadPreviousMessages()`, `loadNextMessages()`, `loadPageAroundId()`), so it would be more error-prone to call the pagination handler from multiple places.

### 🧪 Manual Testing Notes
Requires regression testing on the Bidirectional scrolling.

### ☑️ Contributor Checklist

- [ ] Affected documentation updated (docusaurus, tutorial, CMS)_


Sources/StreamChat/ChatClient.swift
```diff
@@ -132,6 +132,12 @@ public class ChatClient {
         environment.callRepositoryBuilder(apiClient)
     }()
 
+    func makeMessagesPaginationStateHandler() -> MessagesPaginationStateHandling {
+        MessagesPaginationThreadSafeDecorator(
+            decoratee: MessagesPaginationStateHandler()
```
ipavlidakis, line 137: The parameter name and the object name don't really hint any relation. Can we make it more clear that the object should be passed as this parameter?

Sources/StreamChat/Utils/MessagesPaginationStateHandling/MessagesPaginationState.swift
```diff
@@ -0,0 +1,63 @@
...
+    /// A Boolean value that returns whether the channel is currently loading next (new) messages.
+    var isLoadingNextMessages: Bool
+    /// A Boolean value that returns whether the channel is currently loading previous (old) messages.
+    var isLoadingPreviousMessages: Bool
+    /// A Boolean value that returns whether the channel is currently loading a page around a message.
+    var isLoadingMiddleMessages: Bool
+
+    // MARK: Computed Properties
+
+    /// A Boolean value that returns whether the channel is currently loading messages, either old/new or mid pages.
```
ipavlidakis, line 31:
```suggestion
    /// A Boolean value that returns whether the channel is currently loading messages on either previous, mid or next pages.
```
Better to remain consistent to the property naming and prefer previous/next over old/new.

Sources/StreamChat/Utils/MessagesPaginationStateHandling/MessagesPaginationState.swift
```diff
@@ -0,0 +1,63 @@
...
+        newestFetchedMessage?.createdAt
+    }
+
+    // MARK: Initial State
+
+    /// The initial state.
+    static var initial: Self = .init(
+        oldestFetchedMessage: nil,
+        newestFetchedMessage: nil,
+        hasLoadedAllNextMessages: true,
```
ipavlidakis, line 57: Why is this one defaulting to true?

Sources/StreamChat/Utils/MessagesPaginationStateHandling/MessagesPaginationStateHandling.swift
```diff
@@ -0,0 +1,93 @@
...
+//
+
+import Foundation
+
+/// A component responsible for handling the messages pagination state.
+protocol MessagesPaginationStateHandling {
+    /// The current state of the messages pagination.
+    var state: MessagesPaginationState { get }
+
+    /// This function should be called before performing the pagination call.
```
ipavlidakis, line 12:
```suggestion
    /// A method that will be called to inform the object that a pagination call is about to begin.
```

Sources/StreamChat/Utils/MessagesPaginationStateHandling/MessagesPaginationStateHandling.swift
```diff
@@ -0,0 +1,93 @@
...
+    func start(pagination: MessagesPagination)
+    /// This function should be called with the result of the pagination call.
```
ipavlidakis, line 14:
```suggestion
    /// A method that will be called to inform the object that a pagination call has finished with the provided result.
```

Sources/StreamChat/Utils/MessagesPaginationStateHandling/MessagesPaginationStateHandling.swift
```diff
@@ -0,0 +1,93 @@
...
+
+import Foundation
+
+/// A component responsible for handling the messages pagination state.
+protocol MessagesPaginationStateHandling {
+    /// The current state of the messages pagination.
+    var state: MessagesPaginationState { get }
+
+    /// This function should be called before performing the pagination call.
+    func start(pagination: MessagesPagination)
```
ipavlidakis, line 13: I believe `begin` is a better match than `start` as it actually means `to initiate an action`.

Sources/StreamChat/Utils/MessagesPaginationStateHandling/MessagesPaginationStateHandling.swift
```diff
@@ -0,0 +1,93 @@
...
+        }
+    }
+
+    func end(pagination: MessagesPagination, with result: Result<[MessagePayload], Error>) {
+        state.isLoadingNextMessages = false
+        state.isLoadingMiddleMessages = false
+        state.isLoadingPreviousMessages = false
+
+        guard let messages = result.value else {
+            return
```
ipavlidakis, line 56:
```suggestion
            state.oldestFetchedMessage = nil
            state.newestFetchedMessage = nil
            return
```

Sources/StreamChat/Utils/MessagesPaginationStateHandling/MessagesPaginationThreadSafeDecorator.swift
```diff
@@ -0,0 +1,33 @@
+//
+// Copyright © 2023 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+/// A decorator of the `MessagesPaginationStateHandling` to make sure it is thread safe.
+class MessagesPaginationThreadSafeDecorator: MessagesPaginationStateHandling {
```
ipavlidakis, line 8: Why is this one needed rather than making `MessagesPaginationStateHandler` thread-safe?

ipavlidakis (commented):
It's looking great. Just a few comments

Sources/StreamChat/ChatClient.swift
nuno-vieira: @ipavlidakis I'm not sure I'm following. Can you clarify?

Sources/StreamChat/Utils/MessagesPaginationStateHandling/MessagesPaginationState.swift
nuno-vieira: Because when opening a channel, you always open it on the newest page. So "allNextMessages" are loaded. This is especially useful to make sure that the "scrollToBottomButton" is hidden when opening a channel

Sources/StreamChat/Utils/MessagesPaginationStateHandling/MessagesPaginationThreadSafeDecorator.swift
nuno-vieira: This approach makes sure that the Threading logic is separate from the actual business logic. It makes the core component easier to read and also easier to test. In our codebase, we usually put this responsibility inside the actual implementation, but this time, because this component has a lot of logic, I used a decorator to make the logic easy to read. As the component grows, the threading logic will also grow, and probably will also add logging as well, so the Decorator pattern is really useful to decouple Threading, Logging, etc... from the main business logic.

martinmitrevski: I kind of agree with Ilias here, I think it's better that the component is thread-safe by default, instead of decorating it. Few arguments for that:
- someone can make the mistake of using the thread unsafe state directly.
- if the thread safe implementation is an internal detail, we can easily change it with e.g. actors in the future.
- we would actually find it harder to maintain it, since for every new addition in the state, we would need to expose sth in the decorator.
- consistency with the rest of the codebase.
- also, not sure how would this help in testing, can you clarify?

Sources/StreamChat/ChatClient.swift
martinmitrevski: I think he means the name of the parameter: the decorator seems specific to messages pagination, but the parameter name is generic and doesn't provide a clear indication what should be provided.

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
```diff
@@ -414,19 +429,12 @@ public class ChatChannelController: DataController, DelegateCallable, DataStoreP
         let limit = limit ?? channelQuery.pagination?.pageSize ?? .messagesPageSize
         channelQuery.pagination = MessagesPagination(pageSize: limit, parameter: .lessThan(messageId))
 
-        isLoadingPreviousMessages = true
```
martinmitrevski, line 417: Removing this handling would simplify things a lot 👍 We just need to make sure nothing is broken.

martinmitrevski (commented):
Looking great 👌 I'm only not a fan of the decoration of the message pagination state (see my comment).
One question: how well are we covered with existing tests here? Should we add some more?
Also, since it's a tricky change, we should do extensive manual testing (first Alexey, then us when we re-review the PR).

Sources/StreamChat/ChatClient.swift
nuno-vieira: This part of the Decorator pattern, it is meant to be generic, because it can be combined with other decorators. Example:

```swift
MessagesPaginationThreadSafeDecorator(
    decoratee: MessagesPaginationLoggingDecorator(
          decoratee: MessagesPaginationStateHandler() 
    )
}
```

The reason this name is used is for 2 reasons, to easily identify the pattern and to not limit on what the Decorator can decorate, since it can be a concrete implementation, or one more Decorator.

Sources/StreamChat/Utils/MessagesPaginationStateHandling/MessagesPaginationThreadSafeDecorator.swift
nuno-vieira: > someone can make the mistake of using the thread unsafe state directly.

I think that this is very unlikely, since the developer usually we check how it is used in other places

> if the thread-safe implementation is an internal detail, we can easily change it with e.g. actors in the future.

The goal of this decorator pattern is to actually have an "actor" feel in terms of usability.

> we would actually find it harder to maintain it, since, for every new addition in the state, we would need to expose sth in the decorator.

This is actually a good thing. It makes sure you don't forget to handle new functions that you add.

> consistency with the rest of the codebase

This one is valid, but there are multiple places in our codebase where we don't use this pattern, and 100% should. The most obvious one is the Updaters, where we want to make sure all the completions are executed in the `callbackQueue`. With this pattern it would be much cleaner, and also less error-prone.

> also, not sure how would this help in testing, can you clarify

In this particular case, it won't help because we always use `sync` instead of `async` but when using `async` it makes the test harder because you need to mock the queue.

Maybe for this scenario, it is not a clear win using this, not as much as the Updaters, and since you guys don't agree, I'll make the component thread-safe directly.

nuno-vieira: @ipavlidakis @martinmitrevski I've removed it here: https://github.com/GetStream/stream-chat-swift/pull/2595/commits/f254307e3b8daa1f7ce5cd04033aac07db0e9286, let me know what you guys think.

Keep in mind it is not a good idea to do what we usually do, something like this:
```swift
var state: MessagesPaginationState {
        get {
            queue.sync {
                _state
            }
        }
        set {
            queue.sync {
                _state = newValue
            }
        }
    }
```
Because we change **multiple** properties of the `state` at the same time, which would trigger a `queue.sync` multiple times.

Sources/StreamChat/ChatClient.swift
```diff
@@ -132,6 +132,10 @@ public class ChatClient {
         environment.callRepositoryBuilder(apiClient)
     }()
 
+    func makeMessagesPaginationStateHandler(parentMessageId: MessageId?) -> MessagesPaginationStateHandling {
```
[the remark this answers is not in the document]
nuno-vieira: You mean the `parentMessageId`? I would, but I'm actually going to remove this in a follow-up PR. After I fix a bug, we don't actually need this parentMessageId any more, so that is why I didn't explain it here. 👍

martinmitrevski (approved):
LGTM ✅

Sources/StreamChat/Utils/MessagesPaginationStateHandling/MessagesPaginationStateHandling.swift
```diff
@@ -0,0 +1,111 @@
+//
+// Copyright © 2023 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+/// A component responsible for handling the messages pagination state.
+protocol MessagesPaginationStateHandling {
```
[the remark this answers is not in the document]
ipavlidakis: You are subclassing it in `MessagesPaginationStateHandler_Mock`. The protocol so far has no benefit. It may have in the future but given that (at least I don't have enough data) it's not really obvious how this is helping here.

Sources/StreamChat/ChatClient.swift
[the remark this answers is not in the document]
ipavlidakis: I meant in general to add a comment for the method but it's internal so I'm alright not having one.

Sources/StreamChat/ChatClient.swift
ipavlidakis: I understand your intent, but I find the Decorator Pattern naming in a UI framework, to be confusing. Anyway, if you are going to introduce the Decorator Pattern in the framework, then the `MessagesPaginationThreadSafeDecorator` can be more generic and actually be a reusable component.

Sources/StreamChat/Utils/MessagesPaginationStateHandling/MessagesPaginationStateHandling.swift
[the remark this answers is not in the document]
nuno-vieira: `MessagesPaginationStateHandler_Mock` is not a subclass, it is an implementation of the protocol `MessagesPaginationStateHandling`. 

As I said before, it benefits the unit tests. I know people normally say "Do not write code for the sake of tests" (I've been there as well), but the reality is that it contributes to the developer experience. With the current approach, I'm able to write the following unit tests:
```swift
func test_begin_whenLoadingFirstPage_thenSetsStateToInitial() {
    // Given
    sut.state.isLoadingNextMessages = true
    sut.state.hasLoadedAllNextMessages = false
    sut.state.oldestFetchedMessage = .dummy()
    let pagination = MessagesPagination(pageSize: 10, parameter: nil)

    // When
    sut.begin(pagination: pagination)

    // Then
    XCTAssertEqual(sut.state, .initial)
}
```

Without the protocol, and because the `state` property would have to be `private(set)`, I would have to write them like this:
```swift
func test_begin_whenLoadingFirstPage_thenSetsStateToInitial() {
    // Given
    // simulate I already did load another page before
    let pagination = MessagesPagination(pageSize: 5, parameter: .around("122"))
    let messages: [MessagePayload] = [
        .dummy(messageId: "121"),
        .dummy(messageId: "122"),
        .dummy(messageId: "123")
    ]
    sut.end(pagination: pagination, with: .success(messages))

    // When
    let pagination = MessagesPagination(pageSize: 10, parameter: nil)
    sut.begin(pagination: pagination)

    // Then
    XCTAssertEqual(sut.state, .initial)
}
```
As you can see, it is much harder to read and understand what is happening, besides that is not really a "unit" test since it calls 2 public functions.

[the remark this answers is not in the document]
ipavlidakis: I totally agree, I misread and thought that that was a subclass.

PR #2608, opened 2023-05-11
Fix scrollToBottomButton visible when jumping to message on the first page

nuno-vieira:
### 🔗 Issue Links
Resolves https://github.com/GetStream/ios-issues-tracking/issues/411

### 🎯 Goal
Fix scrollToButtomButton appearing when jumping to a message in the first page (Specifically, opening a channel in a message that is in the first page)

### Problem
When jumping to a message on the newest page, the scrollToBottom button would appear since the `isJumpingToMessage` would be true, although we jumped to the first page. The button would only be hidden again after we tried to load the next page, which would return 0 messages.

### Solution
The solution consists of, when jumping to a message, we check where that message is located in the response. If it is in the midpoint and the response is equal to the page size, it means we have more messages to load. If the message is located after the mid point, it means we are already in the newest. If the message is before the midpoint, it means we are already on the oldest page (and so we don't need the parentMessageId anymore in the `MessagesPaginationStateHandler`.

Example jump to message response:
```
pageSize: 5
{
  messageId: 1,
  messageId: 2, <---Before midpoint we are in the oldest page
  messageId: 3, <-- Midpoint
  messageId: 4, <-- After midpoint we are in the newest page
  messageId: 5 
}
```

### 🎨 Showcase

You can see in the videos below that before when opening a channel in a message which was on the first page, the scrollToBottomButton would still appear. After the fix, that does not happen any more.

| Before | After |
| ------ | ----- |
|  <video src="https://github.com/GetStream/stream-chat-swift/assets/12814114/42852690-ea21-44f0-bb42-c5177a49bfa4"/>   |  <video src="https://github.com/GetStream/stream-chat-swift/assets/12814114/d776b83d-0255-4f1c-b3f8-97324d392026"/>  |

### 🧪 Manual Testing Notes
- Configure the DemoApp to enable the Message Debugger
- Open a channel
- Long press on one of the messages in the first page
- Tap on Message Debugger
- Copy the message Id 
- Go back to the Channel List
- Swipe left the same channel
- Tap on the 3 dots
- Tap on "Show channel with message Id""
- Past the message Id and open the channel
- The scrollToBottomButton should be hidden whether the message is from the first page, or last page

### ☑️ Contributor Checklist


ipavlidakis (approved):
LGTM! Great job :fire:

PR #2617, opened 2023-05-15
[Fix]Localization Test affecting all testCases

ipavlidakis:
commit-id:79bcdc89

---

**Stack**:
- #2623
- #2622
- #2621
- #2620
- #2619
- #2618
- #2617 ⬅


Sources/StreamChatUI/CommonViews/PillButton.swift
```diff
@@ -0,0 +1,63 @@
...
+    // MARK: - Lifecycle
+
+    override open func setUpAppearance() {
+        super.setUpAppearance()
+
+        tintColor = appearance.colorPalette.staticBlackColorText
+        setTitleColor(appearance.colorPalette.staticBlackColorText, for: .normal)
+        backgroundColor = isHighlighted
+            ? appearance.colorPalette.highlightedBackground
+            : appearance.colorPalette.staticColorText
```
polqf, line 26: Shouldn't we create a new entry for this? We are using a property that is supposed to be related to text

Sources/StreamChatUI/CommonViews/PlayPauseButton/PlayPauseButton.swift
```diff
@@ -0,0 +1,79 @@
...
+    override open func setUpAppearance() {
+        super.setUpAppearance()
+
+        setImage(appearance.images.playFill, for: .normal)
+        setImage(appearance.images.pauseFill, for: .selected)
+
+        tintColor = appearance.colorPalette.staticBlackColorText
+        backgroundColor = isHighlighted
+            ? appearance.colorPalette.highlightedBackground
+            : appearance.colorPalette.staticColorText
```
polqf, line 53: same as before

Sources/StreamChatUI/CommonViews/PlayPauseButton/PlayPauseButton.swift
```diff
@@ -0,0 +1,79 @@
+//
+// Copyright © 2023 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+import UIKit
+
+open class PlayPauseButton: _Button, AppearanceProvider {
```
polqf, line 8: Most of the logic here is the same as the one in PillButton. Would it make sense to have a common subclass that holds that logic?

Sources/StreamChatUI/VoiceRecording/Views/AudioVisualizationView.swift
```diff
@@ -0,0 +1,260 @@
...
+    public var content: [Float]? {
+        didSet {
+            updateContentIfNeeded()
+        }
+    }
+
+    // MARK: - Configuration Properties
+
+    /// The colour of the waveform bar that isn't part of the "played" duration.
+    open var barColor: UIColor { appearance.colorPalette.textLowEmphasis }
```
polqf, line 25: same as before

Tests/StreamChatUITests/SnapshotTests/CommonViews/PillButton_Tests.swift
```diff
@@ -0,0 +1,112 @@
...
+final class PillButton_Tests: XCTestCase {
+    private var subject: PillButton! = .init().withoutAutoresizingMaskConstraints
+
+    // MARK: - Lifecycle
+
+    override func setUp() {
+        super.setUp()
+        subject.setUp()
+        subject.setUpAppearance()
+        subject.setTitle("This text should appear in a pill.", for: .normal)
```
polqf, line 19: The text looks a bit too close to the edges on the screenshots, doesn't it?

Sources/StreamChatUI/CommonViews/PillButton.swift
ipavlidakis: I believe that the mistake is in the name of variable and not the usage. ColorPalettes shouldn't be semantic (Themes can be semantic) to allow flexibility. That being said I'm seeing that we are using `text` colours as non-text colours in various cases (e.g. ChatChannelListErrorView.backgroundColor, ChatMessageDeliveryStatusCheckmarkView.tintColor etc), so we are not really off here. Let's see what @nuno-vieira will also say and we can decide how to move forward. WDYT?

Tests/StreamChatUITests/SnapshotTests/CommonViews/PillButton_Tests.swift
ipavlidakis: Good catch, just added a little bit of padding.

Sources/StreamChatUI/CommonViews/PillButton.swift
```diff
@@ -0,0 +1,23 @@
+//
+// Copyright © 2023 Stream.io Inc. All rights reserved.
+//
+
+import StreamChat
+import UIKit
+
+open class PillButton: MediaButton {
```
nuno-vieira, line 8: Especially this one should be documented since it is not clear at first sight what this button is.

Sources/StreamChatUI/CommonViews/PlayPauseButton/PlayPauseButton.swift
```diff
@@ -0,0 +1,50 @@
+//
+// Copyright © 2023 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+import UIKit
+
+open class PlayPauseButton: MediaButton {
```
nuno-vieira, line 8: This one also needs documentation, we try to add documentation to all new public views (Even if the name of the view is self explanatory)

nuno-vieira, line 8: I can see that both `PillButton` and `PlayPauseButton` are subclasses of `MediaButton`. This goes against the considered best practices for the UI SDK, as mentioned [here](https://www.notion.so/stream-wiki/Writing-a-new-UI-Component-UIKit-5aeb6a53d7404079b5051f3ee7f17e41?pvs=4#32e4b15769ef40b3b36bb2a4d23c32ff). 

The main issue with this approach is that if a customer wants to customize the `MediaButton` only and for example change the highlighted logic, this change won't be reflected in all the buttons that are subclassing the `MediaButton`. So if a customer does `Components.default.mediaButton = CustomMediaButton.self` this won't change all buttons who subclass it.

I can see that for this particular scenario, it is tricky since we are handling a Button 🤔 For composition, a Button is tricky because we can't have a Button inside another Button 🤔 Or can we? Maybe we can forward the events of the Parent button to the underlying button?

Sources/StreamChatUI/CommonViews/RecordButton/RecordButton.swift
```diff
@@ -0,0 +1,86 @@
...
+
+        addTarget(self, action: #selector(didTouchDown), for: .touchDown)
+        addTarget(self, action: #selector(didTouchUp), for: .touchUpInside)
+    }
+
+    override open func setUpAppearance() {
+        super.setUpAppearance()
+
+        setImage(
+            appearance.images.mic.tinted(with: appearance.colorPalette.textLowEmphasis),
```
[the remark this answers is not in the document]
nuno-vieira: It kinda depends on this one. If this is like the "tintColor" of the button, then it is probably the same colour as all the text in the Theme 🤔

Sources/StreamChatUI/Components.swift
```diff
@@ -43,6 +43,9 @@ public struct Components {
     /// A button used for sending a message, or any type of content.
     public var sendButton: UIButton.Type = SendButton.self
 
+    /// A button used for recording a voice message,
+    public var recordButton: RecordButton.Type = RecordButton.self
```
nuno-vieira, line 47: Where are the other common views? I guess they are in another PR? Would be nice to see the structure here, I'm especially interested in the MediaButton subclasses 👍

Sources/StreamChatUI/VoiceRecording/Views/WaveformView.swift
```diff
@@ -0,0 +1,76 @@
...
+open class WaveformView: _View, ThemeProvider {
+    public struct Content: Equatable {
+        var isRecording: Bool
+        var isPlaying: Bool
+        var duration: TimeInterval
+        var currentTime: TimeInterval
+        var location: URL?
+        var waveform: [Float]
+
+        static let initial = Content(
```
nuno-vieira, line 18: I think you forgot to make the properties public

```suggestion
    public struct Content: Equatable {
        public var isRecording: Bool
        public var isPlaying: Bool
        public var duration: TimeInterval
        public var currentTime: TimeInterval
        public var location: URL?
        public var waveform: [Float]

        public static let initial = Content(
```

Another detail worth mentioning is that usually, we also manually create the `init` for these structures since they won't be public by default. It is not a big deal for now, but if a customer needs to create the this content structure, it won't be possible

nuno-vieira (commented):
Looks great 👌 Added some feedback and some minor comments 👍

nuno-vieira (commented):
Looks great 👌 Added some feedback and some minor comments 👍

Sources/StreamChatUI/CommonViews/PillButton.swift
nuno-vieira: The `ColorPallete` is a bit all over the place right now. And the name is probably misleading, as it is probably more a "Theme" than a "ColorPallete". The problem with the current structure is that customers find it hard to customize because they don't have enough context on the names of the colors. So we did discuss recently that we want to provide more context on the names of the colours. [It is also something we want to add to change on v5](https://www.notion.so/stream-wiki/StreamChat-v5-e7cd0c287fba41e8af69ae507c40d9c9?pvs=4#c70a12dfe5484a359b7aa4895a4a8183). But @ipavlidakis is correct, that for a ColorPallete, the name of this color is incorrect. I don't mind leaving it like this for now, since other views are doing the same, and the customer can still customize this in the view. But indeed what we want in the future is to start adding more context to the colors and move away from the "ColorPallete" approach.

Sources/StreamChatUI/Components.swift
ipavlidakis: The rest of the commonView are being introduced as part of this [PR](https://github.com/GetStream/stream-chat-swift/pull/2619). You can either look there for more context or in the `complete` PR [here](https://github.com/GetStream/stream-chat-swift/pull/2602).

nuno-vieira (approved):
LGTM! ✅

PR #2618, opened 2023-05-15
[Async Messages]VoiceRecordingAttachmentInjector + ItemView

ipavlidakis:
**Stack**:
- #2623
- #2622
- #2621
- #2620
- #2619
- #2618 ⬅


Sources/StreamChatUI/ChatMessageList/Attachments/ChatMessageVoiceRecordingAttachmentListView+ItemView.swift
```diff
@@ -0,0 +1,342 @@
...
+            duration: TimeInterval,
+            currentTime: TimeInterval
+        ) {
+            switch state {
+            case .notLoaded, .loading:
+                durationLabel.isHidden = false
+                durationLabel.text = appearance.formatters.videoDuration.format(duration)
+            case .stopped:
+                durationLabel.isHidden = false
+                durationLabel.text = appearance.formatters.videoDuration.format(duration)
```
polqf, line 284: These 2 cases seem to be doing exactly the same

Sources/StreamChatUI/ChatMessageList/Attachments/ChatMessageVoiceRecordingAttachmentListView+ItemView.swift
```diff
@@ -0,0 +1,342 @@
...
+            case .paused, .playing, .stopped:
+                durationLabel.isHidden = false
```
polqf, line 286: Isn't this the same for all cases?

Sources/StreamChatUI/ChatMessageList/Attachments/ChatMessageVoiceRecordingAttachmentListView+ItemView.swift
```diff
@@ -0,0 +1,342 @@
...
+
+        open func updateFileIconImageView(
+            for state: AudioPlaybackState
+        ) {
+            switch state {
+            case .notLoaded, .loading, .stopped:
+                fileIconImageView.isHidden = false
+            case .paused, .playing:
+                fileIconImageView.isHidden = true
+            default:
```
polqf, line 315: In general, I see a huge usage of `default`, which I consider quite "dangerous" or that can produce some issues when adding new cases. Is this done for any particular reason?

Sources/StreamChatUI/ChatMessageList/Attachments/ChatMessageVoiceRecordingAttachmentListView+ItemViewModel.swift
```diff
@@ -0,0 +1,187 @@
...
+// Copyright © 2023 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+import StreamChat
+import UIKit
+
+extension ChatMessageVoiceRecordingAttachmentListView {
+    /// The ViewModel that drives interactions and events for
+    internal class ItemViewPresenter: AudioPlayingDelegate {
```
polqf, line 11: Why is it called presenter if the description mentions view model?
Also, we haven't been using presenters (or at least not that I am aware of) in the SDK.

Sources/StreamChatUI/VoiceRecording/Components/Dictionary+RawJSONVoiceRecording.swift
```diff
@@ -0,0 +1,36 @@
+//
+// Copyright © 2023 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+import StreamChat
+
+/// Provides easy setters/getters for the `VoiceRecording` feature required `extraData` fields.
+extension Dictionary where Key == String, Value == RawJSON {
```
polqf, line 9: Shouldn't this be a private extension where it is used? IMO this should not be internally exposed to the SDK

Tests/StreamChatUITests/SnapshotTests/ChatMessageList/Attachments/ChatMessageVoiceRecordingAttachmentListView_Tests.swift
```diff
@@ -0,0 +1,142 @@
...
+// Copyright © 2023 Stream.io Inc. All rights reserved.
+//
+
+import StreamChat
+@testable import StreamChatTestTools
+@testable import StreamChatUI
+import UIKit
+import XCTest
+
+final class ChatMessageVoiceRecordingAttachmentListView_Tests: XCTestCase {
```
polqf, line 11: Any reason why the waveform is not shown in the snapshot specs?

Sources/StreamChatUI/ChatMessageList/Attachments/ChatMessageVoiceRecordingAttachmentListView+ItemView.swift
ipavlidakis: This is a consequence of using structs instead of enums for this kind of objects. Without the `default` there will be a compiler error.

Sources/StreamChatUI/VoiceRecording/Components/Dictionary+RawJSONVoiceRecording.swift
ipavlidakis: If this is being made `private` we can't access and we are using it from a few places (so we can't really move it in the file that is being used from). Are you too concerned about it? It's not really affecting the behaviour of the dictionary just allowing for easier access.

Tests/StreamChatUITests/SnapshotTests/ChatMessageList/Attachments/ChatMessageVoiceRecordingAttachmentListView_Tests.swift
ipavlidakis: This is because we haven't added any `waveformData`. No reason why not to have something there. Let me add some data.

Sources/StreamChatUI/ChatMessageList/Attachments/ChatMessageVoiceRecordingAttachmentListView+ItemViewModel.swift
ipavlidakis: The name of the file should become `ItemViewPresenter` and I'm going to change it. Why we are using a Presenter(or in general an interaction object here) is because the VoiceRecordingView in contrast with most of our views needs to manage events coming from the AudioPlayer (in addition to the usual user interaction events). During the implementation of the view, I realised that the file grew too big to manage easily and I decided to split it in a ViewModel and the View. After a discussion with Nuno we decided to change the ViewModel into a Presenter.

Sources/StreamChatUI/ChatMessageList/Attachments/ChatMessageVoiceRecordingAttachmentListView+ItemView.swift
nuno-vieira: Maybe we can add an assertion for the cases where adding a new enum could be problematic 🤔 WDYT?

Sources/StreamChatUI/ChatMessageList/Attachments/ChatMessageVoiceRecordingAttachmentListView+ItemViewModel.swift
nuno-vieira: Some additional context @polqf. This was a ViewModel initially, but since it holds a reference to a view, I suggested that we changed the name to a Presenter since it resembles more a Presenter than a ViewModel. Either way, I also pointed out that concern that we don't follow this pattern right now elsewhere in the SDK, although I wish we did. 
Anyways, handling video player events directly in the view is a pain in the ass, so I can understand that for this case, we can make an expectation and already introduce a UI Logic layer. But, it makes sense to make it internal for now since we are not sure yet how customers would customize this.

Sources/StreamChatUI/VoiceRecording/Components/Dictionary+RawJSONVoiceRecording.swift
nuno-vieira: Because this is our "data" this shouldn't actually be parsed as extra data 🤔 Since we have a `VoiceRecordingAttachmentPayload`, the `duration` and `waveform` properties should be part of the payload itself. Any reason it isn't this way?

nuno-vieira (commented):
Looks really good 👌 Just added some minor comments

Sources/StreamChatUI/ChatMessageList/Attachments/ChatMessageVoiceRecordingAttachmentListView+ItemViewPresenter.swift
```diff
@@ -0,0 +1,187 @@
...
+        internal weak var view: ChatMessageVoiceRecordingAttachmentListView.ItemView?
+
+        /// The rate of the currentPlayback. It's used to provide an override when the playback has been
+        /// paused, but we want to maintain showing the last non-zero rate on UI.
+        internal private(set) var currentPlaybackRate: AudioPlaybackRate = .zero
+
+        /// The debouncer that will be used to add a grace period on the presentation of the playbackLoadingIndicator.
+        private lazy var debouncer: Debouncer = .init(0.5, queue: .main)
+
+        private var isFirstSetup: Bool = true
```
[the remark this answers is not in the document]
ipavlidakis: I used this flag to ensure that the Presenter won't subscribe multiple times on the AudioPlayer. Due to the `MulticastDelegate` implementation (and the usage of `NSHashTable`) this ensured already, so the flag is not necessary any more.

Sources/StreamChatUI/VoiceRecording/Components/Dictionary+RawJSONVoiceRecording.swift
ipavlidakis: That is a very good suggestion. I (mistakenly obviously) thought that in order to change the Payload we require BE work. I have now updated this to used the payload data and not extraData. Thanks :)

Sources/StreamChatUI/ChatMessageList/Attachments/ChatMessageVoiceRecordingAttachmentListView+ItemView.swift
polqf: Missed the point on them being structs, yeah. A bit of a pain, but understandable

Sources/StreamChatUI/ChatMessageList/Attachments/ChatMessageVoiceRecordingAttachmentListView+ItemViewModel.swift
polqf: Ok, got your point. It makes sense to me for it to stay internal, though.

ipavlidakis: Yeap the plan is to keep it internal until it's (if ever) needed to be exposed.

nuno-vieira (approved):
LGTM! 🚀

polqf (approved):

ipavlidakis:

PR #2619, opened 2023-05-15
[Async Messages]VoiceRecordingVC

ipavlidakis:
**Stack**:
- #2623
- #2622
- #2621
- #2620
- #2619 ⬅
- #2618


Sources/StreamChatUI/VoiceRecording/Views/LockIndicatorView.swift
```diff
@@ -0,0 +1,107 @@
...
+    open lazy var stackView: UIStackView = .init()
+        .withoutAutoresizingMaskConstraints
+
+    open lazy var lockImageView: UIImageView = .init()
+        .withoutAutoresizingMaskConstraints
+
+    open lazy var chevronImageView: UIImageView = .init()
+        .withoutAutoresizingMaskConstraints
+
+    open lazy var bottomPaddingSpacer: UIStackView = .init()
```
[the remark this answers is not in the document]
ipavlidakis: Given that UIStackViews [are non-drawing UIViews](https://www.hackingwithswift.com/example-code/uikit/how-to-give-a-uistackview-a-background-color), I prefer to use them for things like spacers that shouldn't appear on screen.

Sources/StreamChatUI/Components.swift
```diff
@@ -428,9 +428,25 @@ public struct Components {
...
     public var voiceRecordingEnabled = false
 
+    /// When set to `true` recorded messages can be grouped together and send as part of one message.
+    public var voiceRecordingConfirmInsteadOfSendEnabled = true
+
+    /// The ViewController that handles the recording flow.
+    public var voiceRecordingViewController: VoiceRecordingVC.Type = VoiceRecordingVC.self
+
+    /// The AudioPlayer that will be used for the voiceRecording playback.
+    public var audioPlayer: AudioPlaying.Type = StreamRemoteAudioQueuePlayer.self
```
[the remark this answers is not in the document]
ipavlidakis: For now, do you have any ideas on where this kind of properties can live?

Sources/StreamChatUI/VoiceRecording/Views/LiveRecordingView.swift
```diff
@@ -0,0 +1,99 @@
+//
+// Copyright © 2023 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+import UIKit
+
+open class LiveRecordingView: _View, ThemeProvider {
```
nuno-vieira, line 8: Missing documentation on this one 👍

Sources/StreamChatUI/VoiceRecording/Views/LiveRecordingView.swift
```diff
@@ -0,0 +1,99 @@
...
+    public struct Content: Equatable {
+        var isRecording: Bool
+        var isPlaying: Bool
+        var duration: TimeInterval
+        var currentTime: TimeInterval
+        var waveform: [Float]
+
+        static var initial = Content(
```
nuno-vieira, line 16: We are missing public properties here, right? And public initialiser?

Sources/StreamChatUI/VoiceRecording/Views/LiveRecordingView.swift
```diff
@@ -0,0 +1,99 @@
...
+
+    open private(set) lazy var recordingIndicator: UIImageView = .init()
+        .withoutAutoresizingMaskConstraints
+
+    open private(set) lazy var durationLabel: UILabel = UILabel()
+        .withBidirectionalLanguagesSupport
+        .withoutAutoresizingMaskConstraints
+
+    open lazy var waveformView: WaveformView = .init()
+        .withoutAutoresizingMaskConstraints
```
nuno-vieira, line 45: Missing docs 🙏

Sources/StreamChatUI/VoiceRecording/Views/LockIndicatorView.swift
nuno-vieira, line 34: More missing docs 🙏

Sources/StreamChatUI/VoiceRecording/Views/LockIndicatorView.swift
```diff
@@ -0,0 +1,107 @@
...
+        .withoutAutoresizingMaskConstraints
+
+    open lazy var lockViewContainer: UIStackView = .init()
+        .withoutAutoresizingMaskConstraints
+
+    open lazy var container: UIView = .init()
+        .withoutAutoresizingMaskConstraints
+
+    open lazy var stackView: UIStackView = .init()
+        .withoutAutoresizingMaskConstraints
```
nuno-vieira, line 26: Especially these 3 require docs because it is difficult to distinguish them since they are all containers

Sources/StreamChatUI/VoiceRecording/Views/RecordingIndicatorView.swift
```diff
@@ -0,0 +1,52 @@
...
+
+    open private(set) lazy var container: UIStackView = .init()
+        .withoutAutoresizingMaskConstraints
+
+    open private(set) lazy var recordingIndicator: UIImageView = .init()
+        .withoutAutoresizingMaskConstraints
+
+    open private(set) lazy var durationLabel: UILabel = UILabel()
+        .withBidirectionalLanguagesSupport
+        .withoutAutoresizingMaskConstraints
```
nuno-vieira, line 24: Nit: More missing docs

Sources/StreamChatUI/VoiceRecording/VoiceRecordingVC.swift
```diff
@@ -0,0 +1,684 @@
...
+            self.duration = duration
+        }
+
+        mutating func updatedWithLocation(_ location: URL) {
+            self.location = location
+        }
+
+        mutating func updatedWithWaveform(_ dataPoint: Float) {
+            waveform.append(dataPoint)
+        }
```
nuno-vieira, line 86: All these ones seem to be missing public access control?

Sources/StreamChatUI/VoiceRecording/VoiceRecordingVC.swift
```diff
@@ -0,0 +1,684 @@
...
+
+    // MARK: - Configuration Properties
+
+    /// The distance (in pixels) the user needs to slide to the leading side of the view to cancel a recording.
+    open var slideToCancelDistance: CGFloat = 75
+
+    /// The distance (in pixels) the user needs to slide to the top side of the view to lock the recording view.
+    open var lockDistance: CGFloat = 50
+
+    open var waveformTargetSamples: Int = 100
```
nuno-vieira, line 107: Can we also add a doc here? 🙏

Sources/StreamChatUI/VoiceRecording/VoiceRecordingVC.swift
```diff
@@ -0,0 +1,684 @@
...
+    open lazy var sendButton: UIButton = components
+        .sendButton
+        .init()
+        .withoutAutoresizingMaskConstraints
+        .withAccessibilityIdentifier(identifier: "sendButton")
+
+    open lazy var confirmButton: UIButton = components
+        .confirmButton.init()
+        .withoutAutoresizingMaskConstraints
+        .withAccessibilityIdentifier(identifier: "confirmButton")
```
nuno-vieira, line 159: Maybe we can add docs for these ones as well? 🙏

Sources/StreamChatUI/VoiceRecording/VoiceRecordingVC.swift
```diff
@@ -0,0 +1,684 @@
...
+            lockIndicatorView.content = false
+            lockIndicatorView.setNeedsLayout()
+
+            Animate { [weak self] in
+                guard let self = self else { return }
+                self.lockIndicatorView.isHidden = false
+                self.lockIndicatorView.layoutIfNeeded()
+                self.centerContainer.insertArrangedSubview(self.recordingIndicatorView, at: 0)
+                self.centerContainer.insertArrangedSubview(self.slideToCancelView, at: 1)
+            }
```
nuno-vieira, line 345: Nit: Since for UIView.animate there are no risks of memory leaks, we could avoid using `weak self` to make the code a bit cleaner and shorter, but no big deal at all 👍

Tests/StreamChatUITests/SnapshotTests/VoiceRecording/Views/LiveRecordingView_Tests.swift
```diff
@@ -0,0 +1,55 @@
...
+@testable import StreamChatUI
+import XCTest
+
+final class LiveRecordingView_Tests: XCTestCase {
+    private var subject: LiveRecordingView! = .init().withoutAutoresizingMaskConstraints
+
+    // MARK: - appearance
+
+    // MARK: - appearance
+
```
nuno-vieira, line 16: Nit: Double marks

nuno-vieira (commented):
Looking super clean 🔥 Added some comments, most of them are just minor improvements or minor details missed 👌

nuno-vieira (commented):
Looking super clean 🔥 Added some comments, most of them are just minor improvements or minor details missed 👌

Sources/StreamChatUI/Components.swift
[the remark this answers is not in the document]
polqf: I think we can leave it here for now, and start debating about it when we target v5

Sources/StreamChatUI/VoiceRecording/VoiceRecordingVC.swift
ipavlidakis: Given the duration of the animation there are always risks(the animation may be on flight but the user will swipe back). Additionally I believe it's a good practice to have the weak self in closures. The difference in code will be one line (as `self.` will still be required). If you don't have strong preferences/arguments I would like to keep it otherwise we can make the change in another PR as it will require changes in a few places.

nuno-vieira: Not a big deal at all, but the behaviour you are describing I believe it should not have any significant impact 🤔  In fact, by using `guard let self = self`, it will have the same behaviour you are trying to avoid (AFAIK). If `VoiceRecordingVC` is deallocated after the `guard let self = self`, only after the closure is finished executing is when `VoiceRecordingVC` is completely deallocated. I've actually had issues with this in the past, and it is quite confusing, but `guard let self = self` is creating a strong reference inside that closure execution, so only when the closure finishes executing is when `VoiceRecordingVC` 100% deallocated 

I think it is nice to keep this in mind for the future, but for now, no need to change this since it is not that important and it won't change anything 👍

Sources/StreamChatUI/Components.swift
```diff
@@ -426,11 +426,27 @@ public struct Components {
     // MARK: - Composer VoiceRecording components
 
     /// A flag which determines if `VoiceRecording` feature will be enabled.
-    public var voiceRecordingEnabled = false
+    public var isVoiceRecordingEnabled = false
+
+    /// When set to `true` recorded messages can be grouped together and send as part of one message.
+    public var isVoiceRecordingConfirmationRequiredEnabled = true
```
nuno-vieira, line 432: Nit: I think it is beneficial if we explain what will happen if we set this to false since it could not be 100% clear. Since a customer's first idea could be, "How do I send the audio messages instantly?" and then they don't find any variable with this name, but they will find a comment. WDYT? 
```suggestion
    /// When set to `true`, recorded messages can be grouped together and sent as part of one message.
    /// When set to `false`, recorded messages will be sent instantly.
    public var isVoiceRecordingConfirmationRequiredEnabled = true
```

ipavlidakis: Shall we do it in a follow up PR please?

Sources/StreamChatUI/VoiceRecording/Components/AudioQueuePlayerNextItemProvider.swift
```diff
@@ -0,0 +1,154 @@
...
+        guard
+            let attachment = voiceRecordingAttachments.first(where: { $0.voiceRecordingURL == url }),
+            let attachmentIndex = voiceRecordingAttachments.firstIndex(of: attachment)
+        else {
+            return nil
+        }
+
+        let indexAfter = voiceRecordingAttachments.index(after: attachmentIndex)
+
+        guard voiceRecordingAttachments.indices.contains(indexAfter) else {
```
polqf, line 148: In StreamChatUI you have an extension over Collection that allows you to do `voiceRecordingAttachments[safe: indexAfter]` and will return an optional value that you can guard let against

polqf: We should probably move that to StreamChat, though

nuno-vieira (approved):
LGTM! ✅ Just added 1 nit comment, can be done in a follow up PR 🙂

Sources/StreamChatUI/Components.swift
nuno-vieira: Yes 100% ✅

ipavlidakis:

Sources/StreamChatUI/VoiceRecording/VoiceRecordingVC.swift
```diff
@@ -0,0 +1,739 @@
...
+        }
+    }
+
+    @objc private func didTapSend(_ sender: UIButton) {
+        updateContentByApplyingAction(.send)
+    }
+
+    @objc private func didTapConfirm(_ sender: UIButton) {
+        updateContentByApplyingAction(.confirm)
+    }
```
polqf, line 692: Any reason why you are passing the sender as a parameter here?

Tests/StreamChatUITests/SnapshotTests/VoiceRecording/Views/LiveRecordingView_Tests.swift
```diff
@@ -0,0 +1,53 @@
+//
+// Copyright © 2023 Stream.io Inc. All rights reserved.
+//
+
+import StreamChat
```
polqf, line 5: It looks like the current time indicator can be shorter than the slider height. Is that expected?

Sources/StreamChatUI/VoiceRecording/Components/AudioQueuePlayerNextItemProvider.swift
ipavlidakis: For now i'm going to change to use the extension. Thank you :)

Sources/StreamChatUI/VoiceRecording/VoiceRecordingVC.swift
ipavlidakis: The `sender` provides context about the caller. It's required for the `didTapPlayPause` to know which action to make. As UIKit is going to pass it for free, when we addTarget and to be consistent across the methods, I included it to all of them. I'm happy to remove them if you think that they can cause issues.

Tests/StreamChatUITests/SnapshotTests/VoiceRecording/Views/LiveRecordingView_Tests.swift
ipavlidakis: This is expected as you can also see on the snapshot from Design's Figma
image

PR #2620, opened 2023-05-15
[Async Message]Preview views

ipavlidakis:
**Stack**:
- #2623
- #2622
- #2621
- #2620 ⬅
- #2619
- #2618

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #2621, opened 2023-05-15
[Async Message]ChatMessageList & Composer

ipavlidakis:
**Stack**:
- #2623
- #2622
- #2621 ⬅
- #2620
- #2619
- #2618


Sources/StreamChatUI/Composer/ComposerVC.swift
```diff
@@ -1049,6 +1141,79 @@ open class ComposerVC: _ViewController,
...
+
+    public func voiceRecording(
+        _ voiceRecordingVC: VoiceRecordingVC,
+        addAttachmentFromLocation location: URL,
+        duration: TimeInterval,
+        waveformData: [Float]
+    ) {
+        var extraData: [String: RawJSON] = [:]
+        extraData.duration = duration
+        extraData.waveform = waveformData
```
nuno-vieira, line 1154: I see, it is because the duration and waveform come from the app that we are using extra data for it? I'm not sure if there is any limitation here, but ideally, both duration and waveform should be part of the Payload since extra data is meant for customers to use it 🤔 

Keep in mind, that extra data is just kown/unknown properties. There is not a specific field for extra data, basically whatever data we don't know from a payload, it gets parsed to extra data.

nuno-vieira (approved):
LGTM! ✅ Just added one small comment, but it is related to a different PR 👍

Sources/StreamChatUI/Composer/ComposerVC.swift
ipavlidakis: This is now being handled in the other PR :)

polqf (approved):

ipavlidakis:

PR #2626, opened 2023-05-17
Add replace query function in the Channel List + Fix muted channels query displaying empty data

nuno-vieira:
### 🔗 Issue Links
Resolves https://github.com/GetStream/ios-issues-tracking/issues/421

### 🎯 Goal
A lot of customers ask how to change the channel list query at runtime to support channel list filters or even channel searching.

The goal of this task is to provide a function to do this easily. 

### 📝 Summary

While doing this I found an issue with our Automatic Filtering. The muted channels query wouldn't show any data, because in our DB we store mutes as an actual type, instead of a Boolean. This required a "predicateMapper" to solve this, but feel free to suggest other ideas on how to fix this.

### 🎨 Showcase
TODO

### 🧪 Manual Testing Notes
TODO

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/Query/ChannelListQuery.swift
+++ b/Sources/StreamChat/Query/ChannelListQuery.swift
@@ -118,7 +118,16 @@ public extension FilterKey where Scope: AnyChannelListFilterScope {
 
     /// Filter for checking whether current user has muted the channel
     /// Supported operators: `equal`
-    static var muted: FilterKey<Scope, Bool> { .init(rawValue: "muted", keyPathString: #keyPath(ChannelDTO.mute)) }
+    static var muted: FilterKey<Scope, Bool> { .init(
+        rawValue: "muted",
+        keyPathString: #keyPath(ChannelDTO.mute),
+        predicateMapper: { muted in
+            if muted {
+                return NSPredicate(format: "mute != nil")
+            }
+            return NSPredicate(format: "mute == nil")
+        }
+    ) }
 
     /// Filter for checking the status of the invite
     /// Supported operators: `equal`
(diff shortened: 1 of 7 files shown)


Sources/StreamChat/Query/ChannelListQuery.swift
```diff
@@ -118,7 +118,16 @@ public extension FilterKey where Scope: AnyChannelListFilterScope {
...
+    static var muted: FilterKey<Scope, Bool> { .init(
+        rawValue: "muted",
+        keyPathString: #keyPath(ChannelDTO.mute),
+        predicateMapper: { muted in
+            if muted {
+                return NSPredicate(format: "mute != nil")
+            }
+            return NSPredicate(format: "mute == nil")
+        }
+    ) }
```
nuno-vieira, line 130: Hum 🤔 I think that the `PredicateMapper` might need to have as an input of the closure the operation type as well 🤔  Because right now, this one works fine, because the "muted" only works for equality, but we might have other cases where we need to change the predicate based on if the operation is `equal` , `greater` etc...

Sources/StreamChat/Query/Filter+ChatChannel.swift
```diff
@@ -46,6 +46,10 @@ extension Filter where Scope == ChannelListFilterScope {
             return nil
         }
 
+        if let overridePredicate = predicateMapper?(mappedValue) {
```
ipavlidakis, line 49: I think that ignoring/overriding the operator here is dangerous and kinda defeats the purpose of the Filter's existence.

WDYT about passing in the operator and allow the PredicateMapper to decide if there is a valid predicate to be provided. I'm talking about something like this

```
predicateMapper: { operator, value in
    switch operator {
       case .equals:
           return NSPredicate(format: value ? "mute != nil" : "mute == nil")
       default:
           return nil
    }
}
```

And then this part will become 
```
if let overridePredicate = predicateMapper?(operator, mappedValue) {
   return overridePredicate
}
```

In this way i believe that we are hooking the predicateMapper to the existing flow without bypassing it.

PR #2630, opened 2023-05-19
Add replace query function to the `ChatChannelListVC`

nuno-vieira:
### 🔗 Issue Links
Resolves https://github.com/GetStream/ios-issues-tracking/issues/421

### 🎯 Goal
A lot of customers ask how to change the channel list query at runtime to support channel list filters or even channel searching. The goal of this task is to provide a function to do this easily.

### 📝 Summary
**New public APIs:**
- `ChatChannelListVC.replaceQuery(_ query: ChannelListQuery)`
- `ChatChannelListVC.replaceChannelListController(_ controller: ChatChannelListController)`

### 🎨 Showcase
https://github.com/GetStream/stream-chat-swift/assets/12814114/d0e3f4fb-1cd3-43d6-84a0-d026e7fb10ee

### 🧪 Manual Testing Notes
Check the video above on how to test manually 

### ☑️ Contributor Checklist


Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
```diff
@@ -214,6 +214,25 @@ open class ChatChannelListVC: _ViewController,
...
+    public func replaceQuery(_ query: ChannelListQuery) {
+        let newController = controller.client.channelListController(
+            query: query
+        )
+        replaceChannelListController(newController)
+    }
+
+    /// Replaces the channel list controller and loads the new data.
+    /// - Parameter controller: The new channel list controller.
+    public func replaceChannelListController(_ controller: ChatChannelListController) {
```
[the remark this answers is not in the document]
nuno-vieira: The filter block is part of the `ChatChannelListController` that is why it makes sense to support both replacing a query and the whole controller 👍

Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
```diff
@@ -214,6 +214,25 @@ open class ChatChannelListVC: _ViewController,
...
     }
 
+    /// Replaces the channel list query and loads the new data.
+    ///
+    /// - Parameter query: The new channel list query.
+    public func replaceQuery(_ query: ChannelListQuery) {
+        let newController = controller.client.channelListController(
+            query: query
+        )
+        replaceChannelListController(newController)
```
ipavlidakis, line 224: Shouldn't we maintain the filter closure of the existing controller, here?

Otherwise we may want to mention that this method will result in the creation of a new controller with a nil filter closure.

ipavlidakis (commented):
LGTM just a comment to ensure that we are not going to lose the filter closure unknowingly.

Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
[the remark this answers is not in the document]
polqf: You're right! All good!

Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
nuno-vieira: This is why the `replaceChannelListController` exists 👍 We can't reuse the filter block for now because it is a private property. 

The `replaceQuery` will be sufficient for most cases, but if a customer wants more control, they can use `replaceChannelListController`.

There a 2 options here I guess:
1. We make the `controller.filter` public
2. We make a comment that if you are using a filter block, you need to use `replaceChannelListController()`

What do you think we should do? @ipavlidakis @polqf

ipavlidakis: I think adding a comment will be ok for now.

PR #2631, opened 2023-05-19
Update release pipeline

testableapple:
### 🎯 Goal

Share [release action](https://github.com/GetStream/fastlane-plugin-stream_actions/blob/main/lib/fastlane/plugin/stream_actions/actions/release_ios_sdk.rb) across all iOS SDKs
- [StreamChatSwiftUI](https://github.com/GetStream/stream-chat-swiftui/pull/316)
- [StreamVideo](https://github.com/GetStream/stream-video-swift/pull/36)

### ☑️ Checklist


fastlane/Fastfile
```diff
@@ -42,72 +43,51 @@ end
...
-    head: "release/#{version_number}",
-    base: 'main',
-    body: changes.to_s
+  release_ios_sdk(
+    version: options[:version],
+    sdk_names: sdk_names,
+    podspec_names: ['StreamChat', 'StreamChat-XCFramework', 'StreamChatUI', 'StreamChatUI-XCFramework'],
+    github_repo: github_repo,
+    publish_release: true,
+    upload_assets: ['Products/StreamChat.zip', 'Products/StreamChatUI.zip', 'Products/StreamChat-All.zip']
```
nuno-vieira, line 87: I wonder if should actually have 2 functions for this. One `release_ios_sdk` and another `publish_ios_sdk` 🤔 My only concern is that this could be error-prone since the way we distinguish release vs publish is a combination of arguments 🤔

nuno-vieira (commented):
This strategy looks like a good approach indeed 🤔 I just have the feeling that we should have 2 different actions for this instead of 1 with different arguments 🤔 But let me know your thoughts!

nuno-vieira (approved):
LGTM! ✅

fastlane/Fastfile
testableapple: Let me separate it into two actions, it should not take long. For SwiftUI we will just have the same `release` lane that executes two commands (`release` and `publish`) 🙂

testableapple: @nuno-vieira, done 🎉
Check out the latest commit (and [SwiftUI PR](https://github.com/GetStream/stream-chat-swiftui/pull/316/commits/a1c29f0e9adc8bae2a40c1f47b8a0a9470380421) if you wish).

nuno-vieira: Looks great 👌

PR #2634, opened 2023-05-19
Fix muted and joined channel list queries with empty data (Auto Filtering Enabled)

nuno-vieira:
### 🔗 Issue Links
Resolves https://github.com/GetStream/ios-issues-tracking/issues/423

### 🎯 Goal
Fix the channel list query being empty when the `.muted` and `.joined` filters were used with auto filtering enabled.

### 🛠 Implementation
We currently have an issue with our Auto Filtering where the `.muted` and `.joined` filters do not properly map to our DB predicates.

After some analysis and discussion with @ipavlidakis, we opted for introducing a `PredicateMapper` so that a filter can override the default Predicate logic.

### 🧪 Manual Testing Notes
Unit tests were added to prove it works correctly.

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/Query/ChannelListQuery.swift
+++ b/Sources/StreamChat/Query/ChannelListQuery.swift
@@ -113,8 +113,20 @@ public extension FilterKey where Scope: AnyChannelListFilterScope {
     static var team: FilterKey<Scope, TeamId?> { .init(rawValue: "team", keyPathString: #keyPath(ChannelDTO.team)) }
 
     /// Filter for checking whether current user is joined the channel or not (through invite or directly)
-    /// Supported operators: `equal`
-    static var joined: FilterKey<Scope, Bool> { .init(rawValue: "joined", keyPathString: #keyPath(ChannelDTO.membership)) }
+    /// Supported operators: `equal`.
+    static var joined: FilterKey<Scope, Bool> { .init(
+        rawValue: "joined",
+        keyPathString: #keyPath(ChannelDTO.membership),
+        predicateMapper: { op, joined in
+            let key = #keyPath(ChannelDTO.membership)
+            switch op {
+            case .equal:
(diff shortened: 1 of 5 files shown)


Sources/StreamChat/Query/Filter.swift
```diff
@@ -111,13 +111,19 @@ public struct Filter<Scope: FilterScope> {
     /// Whether the `keyPathString` represents an array in the DTO entities.
     let isCollectionFilter: Bool
 
+    /// The mapper that will override the DB Predicate. This might be needed
```
ipavlidakis, line 114: The comment is very helpful. Thanks for adding it!

ipavlidakis (approved):
LGTM 🚀

Sources/StreamChat/Query/ChannelListQuery.swift
```diff
@@ -113,12 +113,36 @@ public extension FilterKey where Scope: AnyChannelListFilterScope {
...
+    /// Supported operators: `equal`.
+    static var muted: FilterKey<Scope, Bool> { .init(
+        rawValue: "muted",
+        keyPathString: #keyPath(ChannelDTO.mute),
+        predicateMapper: { op, muted in
+            let key = #keyPath(ChannelDTO.mute)
+            switch op {
+            case .equal:
+                return NSPredicate(format: muted ? "\(key) != nil" : "\(key) == nil")
+            default:
```
polqf, line 141: The code looks the same in both implementations (but the keyPath). Should we be reusing it by placing it in a single place? We are probably using the same logic in other places?

polqf (approved):

Sources/StreamChat/Query/ChannelListQuery.swift
nuno-vieira: For now, I think it is fine to have a little duplication. We can introduce a helper to reuse this logic if we see more use cases like this. 👍

PR #2657, opened 2023-05-30
[BW] Use Xcode 14.2 and macOS 12 for Snapshot tests

testableapple:
Xcode 13 and macOS 13 raise some weird issues on CI for E2E and Snapshot tests. For now, we decided to stick with Xcode 14.2 for these tests and with macOS 12 for Snapshot tests only.

Test:
- https://github.com/GetStream/stream-chat-swift/actions/runs/5122487122/jobs/9211654099?pr=2657

### DIFF
--- a/.github/workflows/smoke-checks.yml
+++ b/.github/workflows/smoke-checks.yml
@@ -89,7 +89,7 @@ jobs:
 
   build-and-test-ui-debug:
     name: Test UI (Debug)
-    runs-on: macos-13
+    runs-on: macos-12
     if: ${{ github.event_name != 'push' }}
     steps:
     - uses: actions/checkout@v3.1.0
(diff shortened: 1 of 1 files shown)


polqf (approved):

nuno-vieira (approved):

fastlane/Fastfile
```diff
@@ -360,7 +360,7 @@ end
 
 desc 'Runs ui tests in Debug config'
 lane :test_ui do |options|
-  next unless is_check_required(sources: sources_matrix[:ui])
+  # next unless is_check_required(sources: sources_matrix[:ui])
```
nuno-vieira, line 363: Why is this commented out?

testableapple: To test Snapshot tests on CI (that's why I missed snapshots in the first PR). I will revert it after the test.

PR #2659, opened 2023-05-31
Allow editing messages with attachments

polqf:
### 🔗 Issue Links

https://github.com/GetStream/ios-issues-tracking/issues/14

### 🎯 Goal

Allow editing messages that contain any type of attachments

### 📝 Summary

As of now, it is not possible to edit the attachments of an already sent message.

### 🛠 Implementation

Dislpays currently attached messages in ComposerVC when editing a message, and updates the model accordingly.
Removed attachments are deleted, and new ones are uploaded and added to the message

### 🧪 Manual Testing Notes

1. Send a message with attachment(s)
2. Edit the message altering the attachments

Expected results:
- The message should only contain the attachments that were sent in the message edition

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChatUI/Composer/ComposerVC.swift
+++ b/Sources/StreamChatUI/Composer/ComposerVC.swift
@@ -156,43 +156,13 @@ open class ComposerVC: _ViewController,
         ///
         /// - Parameter message: The message that the composer will edit.
         public mutating func editMessage(_ message: ChatMessage) {
-            #warning("This process does not belong here probably!")
-            let attachments: [AnyAttachmentPayload] = message.allAttachments.compactMap { attachment in
-                let payload: AttachmentPayload?
-
-                let decoder = JSONDecoder()
-
-                let type: AttachmentPayload.Type
-
-                switch attachment.type {
-                case .image:
-                    type = ImageAttachmentPayload.self
-                case .video:
-                    type = VideoAttachmentPayload.self
-                case .audio:
-                    type = AudioAttachmentPayload.self
-                case .file:
-                    type = FileAttachmentPayload.self
-                default:
-                    return nil
-                }
-
-                payload = try? decoder.decode(type, from: attachment.payload)
-
-                guard let payload = payload else {
-                    return nil
-                }
-
-                return AnyAttachmentPayload(payload: payload)
-            }
-
             self = .init(
                 text: message.text,
                 state: .edit,
                 editingMessage: message,
                 quotingMessage: nil,
                 threadMessage: threadMessage,
-                attachments: attachments,
+                attachments: message.allAttachments.toAnyAttachmentPayload,
                 mentionedUsers: message.mentionedUsers,
                 command: command,
                 extraData: message.extraData,
@@ -767,9 +737,12 @@ open class ComposerVC: _ViewController,
             cid: cid,
             messageId: id
         )
-        // TODO: Adjust LLC to edit attachments also
         // TODO: Adjust LLC to edit mentions
-        messageController?.editMessage(text: newText, extraData: content.extraData)
+        messageController?.editMessage(
+            text: newText,
+            attachments: content.attachments,
+            extraData: content.extraData
+        )
     }
 
     /// Returns a potential user mention in case the user is currently typing a username.
@@ -1306,5 +1279,24 @@ extension ComposerVC: ChatChannelControllerDelegate {
         didUpdateMessages changes: [ListChange<ChatMessage>]
     ) {
         cooldownTracker.start(with: channelController.currentCooldownTime())
+    }
+}
+
+private extension Array where Element == ChatMessageAttachment<Data> {
+    var toAnyAttachmentPayload: [AnyAttachmentPayload] {
+        let decoder = JSONDecoder()
+        return compactMap { attachment in
+            let type: AttachmentPayload.Type
+            switch attachment.type {
+            case .image: type = ImageAttachmentPayload.self
+            case .video: type = VideoAttachmentPayload.self
+            case .audio: type = AudioAttachmentPayload.self
+            case .file: type = FileAttachmentPayload.self
+            default: return nil
+            }
+
+            guard let payload = try? decoder.decode(type, from: attachment.payload) else { return nil }
+            return AnyAttachmentPayload(payload: payload)
+        }
     }
 }
--- a/Sources/StreamChatUI/Composer/ComposerVC.swift
+++ b/Sources/StreamChatUI/Composer/ComposerVC.swift
@@ -487,7 +487,7 @@ open class ComposerVC: _ViewController,
             Animate {
                 self.composerView.confirmButton.isHidden = false
                 self.composerView.sendButton.isHidden = true
-                self.composerView.recordButton.isHidden = self.composerView.sendButton.isHidden
+                self.composerView.recordButton.isHidden = self.composerView.confirmButton.isHidden
                 self.composerView.headerView.isHidden = false
                 self.composerView.cooldownView.isHidden = true
                 self.composerView.leadingContainer.isHidden = false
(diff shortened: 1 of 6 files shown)


Sources/StreamChatUI/Composer/ComposerVC.swift
```diff
@@ -1278,3 +1281,25 @@ extension ComposerVC: ChatChannelControllerDelegate {
         cooldownTracker.start(with: channelController.currentCooldownTime())
     }
 }
+
+private extension Array where Element == ChatMessageAttachment<Data> {
+    var toAnyAttachmentPayload: [AnyAttachmentPayload] {
+        let decoder = JSONDecoder()
```
nuno-vieira, line 1287: Should we use our Stream's JSON decoder here? It is probably more lightweight since it uses the same Decoder instance instead of creating a new instance every time. 
```suggestion
        let decoder = JSONDecoder.stream
```

Sources/StreamChatUI/Composer/ComposerVC.swift
```diff
@@ -487,7 +487,7 @@ open class ComposerVC: _ViewController,
             Animate {
                 self.composerView.confirmButton.isHidden = false
                 self.composerView.sendButton.isHidden = true
-                self.composerView.recordButton.isHidden = self.composerView.sendButton.isHidden
+                self.composerView.recordButton.isHidden = self.composerView.confirmButton.isHidden
```
nuno-vieira, line 490: Was this a bug? Why is this change needed?

Sources/StreamChatUI/Composer/ComposerVC.swift
```diff
@@ -1278,3 +1281,25 @@ extension ComposerVC: ChatChannelControllerDelegate {
...
+            switch attachment.type {
+            case .image: type = ImageAttachmentPayload.self
+            case .video: type = VideoAttachmentPayload.self
+            case .audio: type = AudioAttachmentPayload.self
+            case .file: type = FileAttachmentPayload.self
+            case .voiceRecording: type = VoiceRecordingAttachmentPayload.self
+            default:
+                log.assertionFailure("Unsupported attachment")
+                return nil
+            }
```
nuno-vieira, line 1299: Does this mean that customers can only edit attachments which are our own? Not custom attachments of the customers?

Sources/StreamChatUI/Composer/ComposerVC.swift
```diff
@@ -162,7 +162,7 @@ open class ComposerVC: _ViewController,
                 editingMessage: message,
                 quotingMessage: nil,
                 threadMessage: threadMessage,
-                attachments: attachments,
+                attachments: message.allAttachments.toAnyAttachmentPayload,
```
nuno-vieira, line 165: This is the only concerning part since it looks like if the customer has custom attachments, it won't be able to edit them, right? 🤔 Can we provide a hook somewhere where customers can provide their own attachment payloads decoding logic?

nuno-vieira (commented):
It looks good overall 👍 Just added one concern about this solution not support custom attachments 🤔

Sources/StreamChatUI/Composer/ComposerVC.swift
nuno-vieira: I can see that `JSONDecoder.stream` is not accessible though it being on the LLC. Should this extension live in the LLC SDK and actually be a public API? It seems this function is important for customers wanting to support editing attachments on their custom UIs (Without using our UI SDK)

Sources/StreamChatUI/Composer/ComposerVC.swift
nuno-vieira: An idea to overcome this is to having something like this:

```swift
           let types: [AttachmentType: AttachmentPayload.Type] = [
                .image: ImageAttachmentPayload.self,
                .video: VideoAttachmentPayload.self,
                .audio: AudioAttachmentPayload.self,
                .file: FileAttachmentPayload.self,
                .voiceRecording: VideoAttachmentPayload.self
            ]

            guard let payloadType = types[attachment.type] else { return nil }
            guard let payload = try? decoder.decode(payloadType, from: attachment.payload) else { return nil }
            return AnyAttachmentPayload(payload: payload)
```

The types would be a static property or something living in the `ChatClient` where customers would be able to register their own types. Something like this:
```ChatClient.shared.registerCustomAttachment(WorkoutAttachmentPayload.self)```

This is actually similar to what I was working a while ago when I was refactoring the Attachments API to a v2 because of some limitations when customers wanted to user their custom attachment and having integrated with our own download progress view here: https://github.com/GetStream/stream-chat-swift/pull/2459/files#diff-e77d975b26106706d818e66c3c924b6021b8f5e140ad008c99b394235791417eR40

File: Sources/StreamChatUI/Composer/ComposerVC.swift (hunk 1 above)
Comment (polqf) [reply Sources/StreamChatUI/Composer/ComposerVC.swift:1287]: I am not sure about it being publicly exposed 🤔 The payloads for attachments should not depend on it, do they?

File: Sources/StreamChatUI/Composer/ComposerVC.swift (hunk 3 above)
Comment (polqf) [reply Sources/StreamChatUI/Composer/ComposerVC.swift:1299]: Moved to Slack: https://getstream.slack.com/archives/C03N8GWH8TZ/p1685975622681659

Comment (polqf) [reply Sources/StreamChatUI/Composer/ComposerVC.swift:1299]: Converted this to a follow-up task: https://github.com/GetStream/ios-issues-tracking/issues/463

**Review by ipavlidakis — COMMENTED** [review]
LGTM! One question, do you know if old attachments will be deleted on the backend? Is there something that we need to call to ensure deletion (in case its aligned with product requirements)?

File: Sources/StreamChatUI/Composer/ComposerVC.swift (hunk 1 above)
Comment (nuno-vieira) [reply Sources/StreamChatUI/Composer/ComposerVC.swift:1287]: The main problem I see here is that if a customer is not using the UIKit SDK (And even SwiftUI, for this matter), it won't be able to leverage this property to support editing attachments 🤔 But let's see what @martinmitrevski thinks as well 👍

File: Sources/StreamChatUI/Composer/ComposerVC.swift (hunk 2 above)
Comment (nuno-vieira) [reply Sources/StreamChatUI/Composer/ComposerVC.swift:490]: Was this intended? @ipavlidakis

@polqf  Should we add an entry to the changelog since it is a bug fix?

Comment (polqf) [reply Sources/StreamChatUI/Composer/ComposerVC.swift:490]: We did not allow editing attachments (and voice messages are one of those) before, so this is part of this task

**Comment by polqf** [comment]
>. @ipavlidakis  LGTM! One question, do you know if old attachments will be deleted on the backend? Is there something that we need to call to ensure deletion (in case its aligned with product requirements)?

We don't have an endpoint to delete an attachment. We should probably request one in case product want this

File: CHANGELOG.md [hunk 5]
```diff
@@ -17,6 +18,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Add jumping to a reply inside a thread when opening a thread from a thread reply in the channel view [#2647](https://github.com/GetStream/stream-chat-swift/pull/2647)
 - Add jumping to a reply inside a thread when opening a channel from a reply as the given message around id [#2647](https://github.com/GetStream/stream-chat-swift/pull/2647)
 
+### 🐞 Fixed
+- Allow editing messages with attachments [#2659](https://github.com/GetStream/stream-chat-swift/pull/2659)
```
nuno-vieira, line 22: This is not a fix, but an addition, can we move to the ✅  Added section?

nuno-vieira: I guess it could be subjective here, but we shipped the SDK without this capability on purpose (It was not an accident), so that is why I consider this a "Missing Feature" and not a "Bug". Hence the idea of adding it to the "Added" section. If we add it to the "Fixed" section, it could give a bad impression that we "forgot" about this or that we were not aware of this. 

Either way, if we keep it in the "Fixed" section, at least it sounds reasonable to re-word it to "Fix not being able to edit message attachments", since the "Allow" prefix looks like an addition, and not a fix 🤔

nuno-vieira (approved):
LGTM! ✅

ipavlidakis (approved):
🚢

PR #2667, opened 2023-06-07
Allow passing custom id when sending a message

polqf:
### 🔗 Issue Links

https://github.com/GetStream/ios-issues-tracking/issues/461

### 🎯 Goal

Allow passing a custom id when creating a new message or a new thread reply

### ☑️ Contributor Checklist


Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
```diff
@@ -629,6 +629,7 @@ public class ChatChannelController: DataController, DelegateCallable, DataStoreP
     /// Creates a new message locally and schedules it for send.
     ///
     /// - Parameters:
+    ///   - messageId: The id for the sent message.
```
nuno-vieira, line 632: WDYT?
```suggestion
    ///   - messageId: The id for the sent message. By default, it is automatically generated by Stream.
```

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
```diff
@@ -642,6 +643,7 @@ public class ChatChannelController: DataController, DelegateCallable, DataStoreP
     ///   - completion: Called when saving the message to the local DB finishes.
     ///
     public func createNewMessage(
+        messageId: MessageId? = nil,
```
nuno-vieira, line 646: Nit: Maybe it makes more sense to put messageId after the text? 🤔 Since most of the time, nobody will be interested populating the messageId 🤔

martinmitrevski (approved):

Sources/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -555,7 +556,7 @@ extension NSManagedObjectContext: MessageDatabaseSession {
             throw ClientError.ChannelDoesNotExist(cid: cid)
         }
 
-        let message = MessageDTO.loadOrCreate(id: .newUniqueId, context: self, cache: nil)
+        let message = MessageDTO.loadOrCreate(id: messageId, context: self, cache: nil)
```
nuno-vieira, line 559: Nit: if we make `messageId: MessageId?` optional here, the `messageId ?? .newUniqueId` logic is only applied here instead of duplicating it whenever we use the Updaters 👍

nuno-vieira (approved):
LGTM ✅ Just added some small suggestions 👍

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
polqf: Id is always the first thing when creating a model, I think it makes sense

PR #2677, opened 2023-06-20
[BW] Update quoted reply tests

testableapple:
Resolve https://github.com/GetStream/ios-issues-tracking/issues/469


nuno-vieira (approved):
LGTM! ✅

PR #2681, opened 2023-06-23
Channel Partial Update

polqf:
### 🔗 Issue Links

https://github.com/GetStream/ios-issues-tracking/issues/68

### 🎯 Goal

Allow channel partial updates

### 📝 Summary

We were not allowing channel partial updates until now

### 🛠 Implementation

Adds a new public method to ChannelController that allows clients to set/unset properties in order to update a channel partially, if needed.

### ☑️ Contributor Checklist


polqf:
Note to self: 
Update nessy documentation once reviewed

```swift
        let channelId = ChannelId(type: .messaging, id: "1234-abcd")
        let channelController = client.channelController(for: channelId)
        channelController.partialChannelUpdate(
            members: ["claire", "jane"],
            extraData: [
            "source": "user",
            "source_detail": ["user_id": 123],
            "channel_detail": [
                "topic": "Plants and Animals",
                "rating": "pg"
            ]
        ], completion: { error in /* Handle error if present */ })

        // let's change the source of this channel
        channelController.partialChannelUpdate(extraData: ["source": "system"], completion: { error in /* Handle response accordingly */ })

        // since it's system generated we no longer need source_detail
        channelController.partialChannelUpdate(unsetProperties: ["source_detail"], completion: { error in /* Handle response accordingly */ })

        // and finally update one of the nested fields in the channel_detail
        channelController.partialChannelUpdate(extraData: ["channel_detail.topic": "Nature"], completion: { error in /* Handle response accordingly */ })

        // and maybe we decide we no longer need a rating
        channelController.partialChannelUpdate(unsetProperties: ["channel_detail.rating"], completion: { error in /* Handle response accordingly */ })
```

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
```diff
@@ -262,6 +262,50 @@ public class ChatChannelController: DataController, DelegateCallable, DataStoreP
...
+    ///                 If request fails, the completion will be called with an error.
+    ///
+    public func partialChannelUpdate(
+        name: String? = nil,
+        imageURL: URL? = nil,
+        team: String? = nil,
+        members: Set<UserId> = [],
+        invites: Set<UserId> = [],
+        extraData: [String: RawJSON] = [:],
+        unsetProperties: [String] = [],
```
nuno-vieira, line 283: Which properties are valid to be unset? Can we improve the interface and expose an enum or a struct with static properties to make the API easier to use? Otherwise, customers must dig into the codebase or check the documentation to see which properties can be unset and how they are written.

Something as we have in `ChannelCapability` would be ideal since it is also `RawRepresentable`, so customers can also use a regular String.

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
```diff
@@ -262,6 +262,50 @@ public class ChatChannelController: DataController, DelegateCallable, DataStoreP
...
+    ///   - completion: The completion. Will be called on a **callbackQueue** when the network request is finished.
+    ///                 If request fails, the completion will be called with an error.
+    ///
+    public func partialChannelUpdate(
+        name: String? = nil,
+        imageURL: URL? = nil,
+        team: String? = nil,
+        members: Set<UserId> = [],
+        invites: Set<UserId> = [],
+        extraData: [String: RawJSON] = [:],
```
nuno-vieira, line 282: Nit: This is probably to keep it consistent with the regular `updateChannel()`, but ideally, all these properties should be a struct. Something like `struct ChannelUpdateOptions { var name; var iamgeURL, etc.. }`, otherwise, the function arguments will keep increasing in the future. This struct could then be reused for the `channelUpdate()` as well.

nuno-vieira (commented):
Overall looks good 👍 Just added some suggestions to improve the API in some areas.

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
polqf: The API for other SDKs is String-based, and there is no documentation as to which properties are possible to be unset. We could be adding those but would make it pretty maintenance heavy. Not sure if it's worth it. `FilterKey` is a good example where, IMO, it makes more sense to have such approach.

nuno-vieira (approved):

Issue #2522, opened 2023-02-15
ChannelList change index after pagination


ajay-simform:
- Let's say, I initially loaded 5 channels.
- For the next channels, I called `checkForChannels` from `ChatChannelListViewModel`
- It will fetch the next N channels. But, instead of appending the next channels after already fetched channels, it appends between them.
- e.g. I have the first 5 channels. Initial channels will be [0,1,2,3,4,5]
- After pagination, the channels look like this [0,1,2,3,6,7,8,9,10,4,5]
- The channels index should consistence even after pagination.

ChannelList query that I am using:

```
chatClient.channelListController(
            query: .init(
                filter: .or([
                    .and([
                        .equal(.type, to: "X"),
                        .nonEmpty,
                    ]),
                    .and([
                        .equal(.type, to: "Y"),
                        .nonEmpty,
                    ]),
                    .and([
                        .containMembers(userIds: [currentUserID]),
                        .nonEmpty,
                    ]),
                ]),
                sort: [.init(key: .lastMessageAt, isAscending: false)],
                pageSize: 10,
                messagesLimit: 1
            ), filter: { channel in
                if channel.lastMessageAt == nil {
                     return false
                }
                switch channel.type {
                case ChannelType.custom("X"):
                    return true
                case ChannelType.custom("Y"):
                    if ChatClient.shared.currentUserController().currentUser?.userRole == UserRoleCustomType.A.value {
                        return true
                    } else {
                        return channel.membership != nil
                    }
                default:
                    return channel.membership != nil
                }
            }
        )
```

## GetStream Environment
**GetStream Chat version:** 4.26.0
**GetStream Chat frameworks:** StreamChatSwiftUI
**iOS version:** 16.0
**Swift version:** 5
**Xcode version:** Xcode 14.2
**Device:** Iphone X

martinmitrevski:
Moving this to our low-level client repo - it seems it's not connected to the SwiftUI SDK.

ipavlidakis:
Hi @ajay-simform 

Another thing that looks suspicious is the filter closure you are providing. The filter close should be matching fully or partially the filter. In your case, I can see that you are filtering based on the user's role (our [docs](https://getstream.io/chat/docs/ios-swift/channel_capabilities/?language=swift) mentions that filtering based on the user's role is not recommended) and the membership. Both cases may cause issues.

I'm wondering if something like the following will help resolve the issue that you have
```swift
query: .init(
    filter: .and([
        .or([
[...]

PR #2684, opened 2023-06-27
Fix channel list having data from a different channel list query

nuno-vieira:
### 🔗 Issue Links
Resolves https://github.com/GetStream/ios-issues-tracking/issues/486

### 🎯 Goal
Fix a channel list being populated with data from a different channel list query.

#### Problem
Previously we were observing whenever a new channel was added to the local DB, and if that channel matched the current channel list query, we would link it to the query and show it in the channel list. This is problematic because a channel can be added to the local DB for multiple reasons, and only in some specific cases are we interested in checking if a channel should be added to a channel list query, not all of them. For example, a channel belonging to a specific channel list query can also belong to another channel list query but only later on another page. With the previous solution, this channel would appear instantly on the other channel list query, potentially breaking or creating gaps in the pagination of the channel list. 

#### Solution
The new approach, which is also aligned more with the other SDKs, is to explicitly observe the WebSocket events that are important to check if a channel should be added/removed from the channel list. Because of `NSFetchedResultsController` we now also reset the channels linked to a query on the first-page fetch since if a channel already belonged to 2 queries, whenever the channel appeared in one of the queries, it would automatically appear on the other one because of how `NSFetchedResultsController` works. As a side effect, this also fixes this issue: https://github.com/GetStream/stream-chat-swift/issues/2522, now, there are no more pagination gaps or reorderings when loading a channel list. 

### 🎨 Showcase
N/A

### 🧪 Manual Testing Notes
Smoke Testing is Required.

Important Scenarios to test:
- Sending a new message, should re-order/insert channel list
- Updating a channel which does not match the query should remove it (ex: hiding a channel)
- Creating a channel, should appear in the channel list

### ☑️ Contributor Checklist


Sources/StreamChat/Workers/ChannelListUpdater.swift
```diff
@@ -48,32 +59,33 @@ class ChannelListUpdater: Worker {
...
+//                        // We are going to clean those channels that are present in the both the local and remote query,
+//                        // and that have not been synched nor watched. Those are outdated, can contain gaps.
+//                        let cidsToClean = localInRemote.subtracting(updatedChannels)
+//                        session.cleanChannels(cids: cidsToClean)
+//
+//                        // We are also going to keep track of the unwanted channels
+//                        // Those are the ones that exist locally but we are not interested in anymore in this context.
+//                        // In this case, it is going to query local ones not appearing in remote, subtracting the ones
+//                        // that are already being watched.
+//                        unwantedCids = localNotInRemote.subtracting(watchedAndSynchedChannelIds)
```
nuno-vieira, line 88: From my understanding, since we now remove channels from the query on first fetch, we probably don't need this logic? 🤔

Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift
```diff
@@ -190,178 +244,81 @@ public class ChatChannelListController: DataController, DelegateCallable, DataSt
...
-    public func loadNextChannels(
-        limit: Int? = nil,
-        completion: ((Error?) -> Void)? = nil
-    ) {
-        if hasLoadedAllPreviousChannels {
-            completion?(nil)
-            return
+        // Fallback to legacy of checking if a channel should be linked to the current channel list query.
+        let deprecatedFallback: () -> Bool? = {
+            self.delegate?.controller(self, shouldAddNewChannelToList: channel)
```
polqf, line 303: Why did you move from the switch checking the type of change, to this that is not using `shouldListUpdatedChannel`

Sources/StreamChat/Workers/ChannelListUpdater.swift
```diff
@@ -135,6 +146,43 @@ class ChannelListUpdater: Worker {
...
+            }
+
+            queryDTO.channels.insert(channelDTO)
+        } completion: { error in
+            completion?(error)
+        }
+    }
+
+    /// Unlinks a channel to the given query.
+    func unlink(channel: ChatChannel, with query: ChannelListQuery, completion: ((Error?) -> Void)? = nil) {
```
polqf, line 170: We could probably have a private method with the shared logic of this 2 methods

polqf (commented):
I am missing some integration tests on this one, to make sure things continue to work as expected from a higher level.

Ideally, a test that receives events on a list, and then we check both the list items and the query on the DB.

nuno-vieira:
> I am missing some integration tests on this one, to make sure things continue to work as expected from a higher level.
> 
> Ideally, a test that receives events on a list, and then we check both the list items and the query on the DB.

The current tests should be enough to cover these scenarios. The ChannelListUpdater covers the DB part, and the ChannelListController tests makes sure the link/unlinking are done in the correct context. 

But I'll check if I can add some integration test as well 👍

Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift
nuno-vieira: The reason is that it is a bit redundant. The logic of the `shouldAddNewChannelToList` and the `shouldListUpdatedChannel` is the same. We just want to verify if a channel should belong to a query, and so these 2 methods are not needed, only 1. Either way, these methods are deprecated for quite a while, so customers should not be using these already. (Hopefully)

Sources/StreamChat/Workers/ChannelListUpdater.swift
nuno-vieira: The duplication is minimal, but yes I think we can probably reuse most of the logic into a private function, will do it then 👍

nuno-vieira: Done

nuno-vieira:
> I am missing some integration tests on this one, to make sure things continue to work as expected from a higher level.
> Ideally, a test that receives events on a list, and then we check both the list items and the query on the DB.

I was writing an integration test, but it is a bit redundant, the difference between the integration test, from the unit test is that we do not mock the ChannelListUpdater, but this is very minimal, and the integration test won't add any value here, IMO, since we already test that linking/unlink works in the ChannelListUpdater unit test. 

Let me know what you think. @polqf

polqf (approved):

PR #2701, opened 2023-07-12
Channel List Custom Sorting

polqf:
### 🔗 Issue Links

Resolves https://github.com/GetStream/ios-issues-tracking/issues/15

### 🎯 Goal

Allow sorting Channel List using client-defined parameters

### 📝 Summary

This PR allows clients to sort the channel list based on their need. This also includes sorting based on extra data (as shown in the documentation)

### 🛠 Implementation

`ChannelListSortingKey` has been moved from an enum to a struct, and it allows passing a KeyPath.
This KeyPath is then used to sort when necessary.
The integration point is in the ListDatabaseObserver, where we receive an array of `Sorting`. This is then mapped to `SortValue`s if possible, and this is used in the observer to sort the elements when receiving updates.

Tests are showing that there is negligible degradation with the change.

It also fixes some issues on BackgroundListDatabaseObserver

### 🧪 Manual Testing Notes

**Prerequisites:** Enable Channel Pinning on the Demo App Configuration

**Some scenarios we want to test:**
- When we fetch the channels, pinned channels should be at the top
- When I pin a channel, that channel should go to the top
- When I unpin a channel, that channel should go back to its original position.

### ☑️ Contributor Checklist


Sources/StreamChat/Models/Channel+Sorting.swift
```diff
@@ -0,0 +1,166 @@
...
+//
+
+import Foundation
+
+// last_updated, last_message_at, updated_at, created_at, member_count, unread_count or has_unread
+
+class Example {
+    let channels: [ChatChannel] = []
+
+    func sortWithDynamicMember(with sorting: [Sorting<ChannelListSortingKey>]) -> [ChatChannel] {
```
polqf, line 12: With dynamicMemberLookup, we should be able to use the existing combination of `Sorting` plus `ChannelListSortingKey `. The only problem is that ChannelListSortingKey is an enum, and does not allow custom fields. We could probably just create a new `CustomSortingKey` struct, that can be used for it 🤔 .

polqf: This solution comes at the expense of the implementation of `subscript[dynamicMember: ]`, though

Sources/StreamChat/Models/Channel+Sorting.swift
```diff
@@ -0,0 +1,166 @@
...
+            return false
+        }
+    }
+
+    struct SortValue<T> {
+        let keyPath: PartialKeyPath<T>
+        let isAscending: Bool
+    }
+
+    func sortWithKeyPath(with sorting: [SortValue<ChatChannel>]) -> [ChatChannel] {
```
polqf, line 37: This solution has the safety of using PartialKeyPath<T> to both list the possible sorting keys, and to get those values from the object in a nice and safe way. But:
- It returns values as `Any`, and then we need to do downcasting
- It does not allow extra data -> I could not think of a way to overcome this, unfortunately. Which kills this approach 🙃

Sources/StreamChat/Models/Channel+Sorting.swift
```diff
@@ -0,0 +1,166 @@
...
+            }
+
+            return false
+        }
+    }
+}
+
+extension ChatChannel {
+    // This method is meant to only be used for sorting functionality.
+    subscript(dynamicMember member: String) -> String? {
```
polqf, line 81: This is maintenance heavy, and needs a single return type. String in this case.

polqf: But seems to be the only approach that allows us to seamlessly access extra data

Sources/StreamChat/Models/Channel+Sorting.swift
```diff
@@ -0,0 +1,166 @@
...
+        case "cid":
+            return cid.rawValue
+        case "name":
+            return name
+        case "type":
+            return type.rawValue
+        case "lastMessageAt":
+            return lastMessageAt?.sortValue
+        case "createdBy":
+            return createdBy?.id
```
nuno-vieira, line 92: The keys should a actually be the ones from the backend 
```suggestion
        case "created_at":
            return createdBy?.id
```

Sources/StreamChat/Models/Channel+Sorting.swift
```diff
@@ -0,0 +1,166 @@
...
+private extension Bool {
+    var sortValue: String { self ? "0" : "1" }
+}
+
+private extension Int {
+    var sortValue: String { "\(self)" }
+}
+
+private extension Double {
+    var sortValue: String { "\(self)" }
```
nuno-vieira, line 161: Is this okay? Comparing doubles as string? For equality it be okay, but > or < also works?

Sources/StreamChat/Models/Channel+Sorting.swift
```diff
@@ -0,0 +1,166 @@
...
+        case "team":
+            return team
+        default:
+            break
+        }
+
+        // Does not work with nested values (eg. "scope.score" -> `extraData: { scope : { score : 1 }}`)
+        if let value = extraData[member] {
+            return value.sortValue
+        }
```
nuno-vieira, line 132: We could support nested values actually 🤔 On the extraData, we can check if the key contains any ".", if yes we can nest it. Example: `user.name` we would split by `.` and convert it to `extraData["user"]?extraData["name"]`

Sources/StreamChat/Models/Channel+Sorting.swift
nuno-vieira: I don't think that it is a big deal if we break a little bit the API here but make it "compatible". What I mean is that we can convert the enum to a struct with static properties like we do in other places.

This would allow for customers to do something like this:
`.custom("test", \ChatChannel.test)`

This will also work for extra data since customers can extend `ChatChannel` like so:
```swift
extension ChatChannel {
    var test: String {
        extraData["test"]?.stringValue ?? ""
    }
}
```

Sources/StreamChat/Models/Channel+Sorting.swift
nuno-vieira: Yeah, the only way this could work is to only allow Primitive types as a KeyPath, example:

`custom(key: String, keyPath: KeyPath<ChatChannel, String>)`
`custom(key: String, keyPath: KeyPath<ChatChannel, Int>)`
`custom(key: String, keyPath: KeyPath<ChatChannel, Double>)`
`custom(key: String, keyPath: KeyPath<ChatChannel, Bool>)`

If we convert the `ChannelListSortingKey` to struct, or just add this additional cases, it could potentially work

Tests/StreamChatTests/Query/Sorting/ListDatabaseObserver+Sorting.swift
```diff
@@ -0,0 +1,301 @@
...
+
+    lazy var bulkChannels: [(name: String, createdAt: Date)] = (0...50).map { _ in
+        let randomIndex = Int.random(in: 0..<defaultChannels.count)
+        return defaultChannels[randomIndex]
+    }
+
+    var database: DatabaseContainer_Spy!
+    var query: ChannelListQuery!
+    var observer: ListDatabaseObserverWrapper<ChatChannel, ChannelDTO>!
+
```
ipavlidakis, line 31: nit: Can we nullify those in `tearDown`?

Tests/StreamChatTests/Query/Sorting/ListDatabaseObserver+Sorting.swift
```diff
@@ -0,0 +1,301 @@
...
+        measureMetrics([.wallClockTime], automaticallyStartMeasuring: false) {
+            do {
+                try test_timeToProcessMultipleChatChannels_defaultSorting(isBackground: true)
+            } catch {
+                XCTFail(error.localizedDescription)
+            }
+        }
+    }
+
+    func test_timeToProcessMultipleChatChannels_defaultSorting(isBackground: Bool) throws {
```
ipavlidakis, line 228: nit: I think having those private and instead of the `test_` prefix use `assert` makes more sense as they are being reused. Passing `#file` and `#line` in their asserts will help more identifying which test case fails (when and if it does)

Tests/StreamChatTests/Query/Sorting/ListDatabaseObserver+Sorting.swift
```diff
@@ -0,0 +1,301 @@
...
+
+    lazy var defaultChannels = [
+        (name: "A", createdAt: datePast),
+        (name: "B", createdAt: dateNow),
+        (name: "C", createdAt: dateWayPast),
+        (name: "D", createdAt: dateNow),
+        (name: "E", createdAt: datePast)
+    ]
+
+    lazy var bulkChannels: [(name: String, createdAt: Date)] = (0...50).map { _ in
```
ipavlidakis, line 23: nit: Can we increases this one to 1000? Having a sample as small as 50 it will only highlight a really bad issue (e.g cubic or quadratic complexity). With bigger sample we should be able to identify even smaller issues.

polqf: You are right, but only increasing from 50 to 100 adds 15 seconds, with a total of 25s. We cannot afford to spend that much time on PR checks.
In this particular case, given the logic, I would say it is safe to leave it as is. Thoughts?

StreamChat.xcodeproj/xcshareddata/xcbaselines/799C9450247D59B1001F1104.xcbaseline/28BBA491-FD00-4587-BD5A-F64363390F25.plist
```diff
@@ -0,0 +1,52 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
polqf: This is where the baseline for the performance tests are stored

Sources/StreamChat/Query/Sorting/ChannelListSortingKey.swift
```diff
@@ -5,55 +5,53 @@
...
+    public static let createdAt = Self(keyPath: \.createdAt, remoteKey: "created_at", canUseAsDBSortDescriptor: true)
     /// Sort channels by date they were updated.
-    case updatedAt
+    public static let updatedAt = Self(keyPath: \.updatedAt, remoteKey: "updated_at", canUseAsDBSortDescriptor: true)
     /// Sort channels by the last message date..
-    case lastMessageAt
+    public static let lastMessageAt = Self(keyPath: \.lastMessageAt, remoteKey: "last_message_at", canUseAsDBSortDescriptor: true)
     /// Sort channels by number of members.
-    case memberCount
+    public static let memberCount = Self(keyPath: \.memberCount, remoteKey: "member_count", canUseAsDBSortDescriptor: true)
```
nuno-vieira, line 18: Can the Remove Key actually be the ChannelPayload Coding Keys? For better compile safe keys?

nuno-vieira (commented):
I think we are in the right direction, but it seems to me that we missing some scenarios, so I asked some questions for clarification 👍

Sources/StreamChat/Query/Sorting/ChannelListSortingKey.swift
```diff
@@ -78,4 +76,36 @@ extension Array where Element == Sorting<ChannelListSortingKey> {
...
+        SortValue(keyPath: key.keyPath, isAscending: isAscending)
+    }
+}
+
+private extension KeyPath where Root == ChatChannel {
+    var stringValue: String {
+        let value = String(describing: self)
+        let root = String(describing: Self.rootType)
+        return value.replacingOccurrences(of: "\\\(root).", with: "")
+    }
```
nuno-vieira, line 104: Can we add a comment explaining what is this doing?

nuno-vieira:
I think it is important that we add a way to test this manually. Ideally, we should try to replicate a "pinned" channel behaviour. This will also verify if everything works correctly, especially for the customers who are asking for this behaviour. 

Some scenarios we want to test:
1. When we fetch the channels, pinned channels should be at the top
2. When I pin a channel, that channel should go to the top
3. When I unpin a channel, that channel should go have the default sorting or other sorting that is part of the query.

Sources/StreamChat/Query/Sorting/ChannelListSortingKey.swift
```diff
@@ -5,55 +5,53 @@
...
-    case memberCount
+    public static let memberCount = Self(keyPath: \.memberCount, remoteKey: "member_count", canUseAsDBSortDescriptor: true)
     /// Sort channels by `cid`.
     /// **Note**: This sorting option can extend your response waiting time if used as primary one.
-    case cid
+    public static let cid = Self(keyPath: \.cid, remoteKey: "cid", canUseAsDBSortDescriptor: true)
     /// Sort channels by unread state. When using this sorting key, every unread channel weighs the same,
     /// so they're sorted by `updatedAt`
-    case hasUnread
+    public static let hasUnread = Self(keyPath: \.hasUnread, remoteKey: "has_unread", canUseAsDBSortDescriptor: false)
```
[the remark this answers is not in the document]
polqf: The idea is that if there are no custom sortings, we just use the NSFetchResultsController for that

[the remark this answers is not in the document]
polqf: I maintained the same logic that was before to avoid breaking existing customers. Runtime will only be applied if there is at least one custom sorting key.

Sources/StreamChat/Query/Sorting/SortValue.swift
```diff
@@ -0,0 +1,53 @@
...
+
+        if let lString = lhs as? String, let rString = rhs as? String {
+            return isAscending ? lString < rString : lString > rString
+        } else if let lInt = lhs as? Int, let rInt = rhs as? Int {
+            return isAscending ? lInt < rInt : lInt > rInt
+        } else if let lDouble = lhs as? Double, let rDouble = rhs as? Double {
+            return isAscending ? lDouble < rDouble : lDouble > rDouble
+        } else if let lDate = lhs as? Date, let rDate = rhs as? Date {
+            return isAscending ? lDate < rDate : lDate > rDate
+        }
```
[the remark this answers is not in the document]
polqf: Added it

Sources/StreamChat/Controllers/BackgroundListDatabaseObserver.swift
```diff
@@ -212,14 +221,24 @@ class BackgroundListDatabaseObserver<Item, DTO: NSManagedObject> {
...
 
-    private func processItems(_ changes: [ListChange<Item>] = []) {
+    private func processInitialItems() {
+        processItems(nil)
+    }
+
+    private func processItems(_ changes: [ListChange<Item>]?) {
         mapItems { [weak self] items in
             self?._items = items
+            let changes = changes ?? items.enumerated().map { .insert($1, index: IndexPath(item: $0, section: 0)) }
```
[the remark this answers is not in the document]
polqf: That is actually a good point. I missed this part in both BG and Foreground Database Observers 😄

Sources/StreamChat/Query/Sorting/SortValue.swift
```diff
@@ -47,7 +47,7 @@ extension Array {
         } else if let lDate = lhs as? Date, let rDate = rhs as? Date {
             return isAscending ? lDate < rDate : lDate > rDate
         } else if let lBool = lhs as? Bool, let rBool = rhs as? Bool {
-            return isAscending ? !lBool && rBool : lBool && !rBool
+            return isAscending ? lBool && !rBool : !lBool && rBool
```
polqf, line 50: Ascending should be from false to true 🤔

nuno-vieira: Resolving, since it was discussed on planning 👍

nuno-vieira (approved):

CHANGELOG.md
```diff
@@ -18,6 +21,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### 🔄 Changed
 - Renames `Components.mentionAvatarView` -> `Components.userAvatarView` [#2703](https://github.com/GetStream/stream-chat-swift/pull/2703)
+- Uses Diffing to update Channel List data for stability improvements [#2701](https://github.com/GetStream/stream-chat-swift/pull/2701)
```
polqf, line 24:
```suggestion
- Use Diffing to update Channel List data to improve stability  [#2701](https://github.com/GetStream/stream-chat-swift/pull/2701)
```

Sources/StreamChat/Query/Sorting/SortValue.swift
polqf: Can we add a comment here?

Tests/StreamChatUITests/SnapshotTests/ChatChannelList/Search/ChatChannelSearchVC_Tests.swift
```diff
@@ -74,16 +74,17 @@ final class ChatChannelSearchVC_Tests: XCTestCase {
 
         vc.currentSearchText = "Some message"
         vc.executeLifecycleMethods()
+        vc.channels = mockedChannelListController.channels_mock ?? []
```
polqf, line 77: I know it is nice to ease tests, but channels should be private(set) at least

nuno-vieira: This was actually not for tests, since now the data source of the channel list is the `channels` property, I don't think that `private(set)` is a good option here, just like it is not the case for the `channelListController`. Customers might want to update the data from another view as well, we don't know 🤔 I don't there is any benefit of making it private(set) but let me know if you have any other scenario in mind 👍

polqf: I don't think it is a good approach to allow others to mutate the source of truth of our views.
At most, I would have a method that explicitly mentions what it does (and internally calls reloadData or whatever is needed)

nuno-vieira: A good thing about making it public is that it could allow customers to override the DataSource and inject additional channels. We already had a couple of requests like this before in the message list for example this one: https://github.com/GetStream/stream-chat-swift/issues/2305

Either way, I'll make it private(set) for now, and if we customers asking for this, then we can open it up

nuno-vieira: @polqf Done ✅

Sources/StreamChat/Query/Sorting/SortValue.swift
nuno-vieira: @polqf ✅

testableapple (approved):
LGTM 🚢

Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
```diff
@@ -14,7 +14,7 @@ open class ChatChannelListVC: _ViewController,
     ThemeProvider,
     SwipeableViewDelegate {
     /// The data of the channel list.
-    public var channels: [ChatChannel] = []
+    public private(set) var channels: [ChatChannel] = []
```
polqf, line 17: 👏

PR #2703, opened 2023-07-14
Add Message Search and Channel Search UI Components

nuno-vieira:
### 🔗 Issue Links
Resolves https://github.com/GetStream/ios-issues-tracking/issues/217

### 🎯 Goal
Add message search and channel search UI Components. 

### 📝 Summary
- Adds Message Search UI Component
- Adds Channel Search UI Component
- Renames `Components.mentionAvatarView` -> `Components.userAvatarView`

### 🛠 Implementation
Two new UI Components were added, `ChatMessageSearchVC` and `ChatChannelSearchVC`. These two new view controllers are then used as a `resultsController` of the `UISearchController` in the Channel List.

Both `ChatMessageSearchVC` and `ChatChannelSearchVC` are subclasses of `ChatChannelListSearchVC`. The latter is an abstract class in which the responsibility is to reuse the common search logic between the Message Search and the Channel Search. In turn, the `ChatChannelListSearchVC` is a subclass of `ChatChannelListVC` since the UI is pretty much the same as the one from the Channel List.

**Class Diagram**
```mermaid
classDiagram
    ChatChannelListSearchVC <|-- ChatChannelSearchVC
    ChatChannelListVC <|-- ChatChannelListSearchVC
    ChatChannelListSearchVC <|-- ChatMessageSearchVC
    class ChatChannelSearchVC{
        loadSearchResults(text: String)*
        loadMoreSearchResults()*
        bool hasEmptyResults*
    }
    class ChatChannelListSearchVC{
        loadSearchResults(text: String)*
        loadMoreSearchResults()*
        bool hasEmptyResults*
    }
    class ChatMessageSearchVC{
        loadSearchResults(text: String)
        loadMoreSearchResults()
        bool hasEmptyResults
    }
```

**Usage:**
By default, the search is disabled. To enable it, a `ChannelListSearchStrategy` should be configured in the `Components` config. The SDK provides two search strategies out-of-the-box, searching by channels or messages. Here is how you can enable the different strategies.
```swift
// If you want to enable the search by channels
Components.default.channelListSearchStrategy = .channels
// If you want to enable the search by messages
Components.default.channelListSearchStrategy = .messages
// If you want to enable the search by channels but with a custom subclass
Components.default.channelListSearchStrategy = .channels(CustomChannelSearchVC.self)
// If you want to enable the search by messages but with a custom subclass
Components.default.channelListSearchStrategy = .messages(CustomMessageSearchVC.self)
```

### 🎨 Showcase
| Channels Search | Messages Search |
|--------|--------|
| <video src="https://github.com/GetStream/stream-chat-swift/assets/12814114/79540469-8d2d-4537-ac32-8966f735e868"/> | <video src="https://github.com/GetStream/stream-chat-swift/assets/12814114/899fa777-28f5-42e0-953e-38f80368dc53"/> | 

### 🧪 Manual Testing Notes
- Open DemoApp
- Tap on Configuration
- Make sure the channel list search strategy is set to messages or channels
- Open the Channel List
- It should have a top search bar
- Search messages or channels
- When tapping a channel, it should open that channel
- When tapping a message, it should open the channel in the message location

**Known backend issue:** If a channel is truncated, the message will still appear on the search results, so when we open the channel on this message, it won't jump to this message. The backend will fix this soon. The expected behaviour is that messages which belong to a truncated channel should not appear in the search result.

### ☑️ Contributor Checklist


.swiftlint.yml
```diff
@@ -34,12 +34,10 @@ opt_in_rules:
   - empty_string
   - fatal_error_message
   - first_where
-  - modifier_order
   - toggle_bool
   - identical_operands
   - joined_default_parameter
   - unavailable_function
-  - overridden_super_call
```
polqf, line 42: Why are those options removed?

Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
```diff
@@ -233,8 +234,38 @@ open class ChatChannelListVC: _ViewController,
...
+            channelSearchVC.didSelectChannel = { [weak self] channel in
+                self?.router.showChannel(for: channel.cid)
+            }
+            resultsController = channelSearchVC
+        }
+
+        let searchController = UISearchController(searchResultsController: resultsController)
+        searchController.searchResultsUpdater = resultsController
+        searchController.searchBar.placeholder = L10n.ChannelList.search
+        return searchController
```
polqf, line 268: Wouldn't it make more sense to have this in the subclass?

Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
```diff
@@ -13,10 +13,17 @@ open class ChatChannelListItemView: _View, ThemeProvider, SwiftUIRepresentable {
         public let channel: ChatChannel
         /// Current user ID needed to filter out when showing typing indicator.
         public let currentUserId: UserId?
-
-        public init(channel: ChatChannel, currentUserId: UserId?) {
+        /// The message part of a search result.
+        public var searchedMessage: ChatMessage?
```
polqf, line 17: What about a subclass of this Item view?

Sources/StreamChatUI/ChatChannelList/Search/ChatMessageSearchVC.swift
```diff
@@ -0,0 +1,161 @@
+//
+// Copyright © 2023 Stream.io Inc. All rights reserved.
+//
+
+import StreamChat
+import UIKit
+
+/// The view controller responsible to search messages.
+open class ChatMessageSearchVC: ChatChannelListVC, UISearchResultsUpdating, ChatMessageSearchControllerDelegate {
```
polqf, line 9: This looks quite similar to the other subclass. What about a common class that hold most of the logic, contains also some logic that should not be in ChatChannelListVC, and the final concrete implementations subclass from this intermediate class?

polqf (commented):
I think we should try to have channel list a bit more separated from search so that we can properly scale them in separate directions as needed.

.swiftlint.yml
nuno-vieira: They were annoying and incorrect. 

**Order of the override:** The order of the override is not that important, and it was the opposite of Xcode's autocomplete. 

**Overriden super call:** This was incorrect. This meant that we had to call `super` every time. In some cases we might to want to completely override the super function, just like we did here with the Search view.

Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
nuno-vieira: Which subclass? The search controller is part of the Channel List, and not other entity 🤔

Sources/StreamChatUI/ChatChannelList/Search/ChatMessageSearchVC.swift
nuno-vieira: The reason it is two separate classes for search is that otherwise, it will require a lot of if-statements which IMO will make it harder to maintain but also harder to customise for the customers. It seems to me that separating the logic into two classes can scale better, and it is also more flexible. 

I played with multiple approaches, and this one was the approach that worked best for us and for the customers. We need to keep in mind always in terms of customizability and reusability.

Because this approach is flexible enough, it means if we want to reuse some of the logic, we can in the future. Right now, it could be a premature decision to try to remove duplication, but let me know if you have a different opinion 👍

nuno-vieira:
> I think we should try to have channel list a bit more separated from search so that we can properly scale them in separate directions as needed.

I also thought about this initially, but this will complicate the customization, requiring duplication on the customer's side. Not only on the customer's side but on our side as well.

Sources/StreamChatUI/ChatChannelList/Search/ChatMessageSearchVC.swift
polqf: What I meant by my comment is not to have a common class for both use cases, but instead having an intermediate one that holds the common logic

ChatChannelListVC -> ChatChannelSearchVC (internal) -> ChatMessageSearchVC

That way you have all the common logic in ChatChannelSearchVC. If there is any issue in search, we solve it once and for all its subclasses, which makes it easier to maintain. ChatMessageSearchVC can then have only customization options that apply to the particular use case.

nuno-vieira: #### So you mean something like the following:

```mermaid
classDiagram
    ChatChannelListSearchVC <|-- ChatChannelSearchVC
    ChatChannelListVC <|-- ChatChannelListSearchVC
    ChatChannelListSearchVC <|-- ChatMessageSearchVC
    class ChatChannelSearchVC{
    }
    class ChatChannelListSearchVC{
    }
    class ChatMessageSearchVC{
    }
```

Code here:

```swift
open class ChatChannelListSearchVC: ChatChannelListVC, UISearchResultsUpdating {
    /// The component responsible to debounce search requests.
    public var debouncer = Debouncer(0.3, queue: .main)

    /// The current active search text.
    public var currentSearchText: String = ""

    /// A component responsible to handle when to load new channels.
    private lazy var viewPaginationHandler: ViewPaginationHandling = {
        ScrollViewPaginationHandler(scrollView: collectionView)
    }()

        super.setUpLayout()

        navigationItem.searchController?.searchBar.placeholder = "Why"

        view.embed(emptyView)
        emptyView.isHidden = true
        emptyView.actionButton.removeFromSuperview()
        emptyView.titleLabel.isHidden = true
    }

    override open func setUp() {
        collectionView.register(
            components.channelCell.self,
            forCellWithReuseIdentifier: collectionViewCellReuseIdentifier
        )

        collectionView.register(
            components.channelCellSeparator,
            forSupplementaryViewOfKind: ListCollectionViewLayout.separatorKind,
            withReuseIdentifier: separatorReuseIdentifier
        )

        collectionView.dataSource = self
        collectionView.delegate = self

        viewPaginationHandler.bottomThreshold = 800
        viewPaginationHandler.onNewBottomPage = { [weak self] in
            self?.loadMoreChannels()
        }
    }

    override open func setUpAppearance() {
        super.setUpAppearance()

        emptyView.iconView.image = appearance.images.emptySearch
    }

    open func updateSearchResults(for searchController: UISearchController) {
        guard let text = searchController.searchBar.text, !text.isEmpty, text != currentSearchText else {
            return
        }

        currentSearchText = text

        debouncer.execute { [weak self] in
            self?.updateSearchResults(with: text)
        }
    }

    // swiftlint:disable unavailable_function
    open func updateSearchResults(with text: String) {
        fatalError("This function should be implemented by a subclass.")
    }

    // swiftlint:disable unavailable_function
    open func loadMoreSearchResults() {
        fatalError("This function should be implemented by a subclass.")
    }

    open var hasEmptyResults: Bool {
        fatalError("This function should be implemented by a subclass.")
    }

    override open func controller(_ controller: DataController, didChangeState state: DataController.State) {
        switch state {
        case .initialized, .localDataFetched:
            if hasEmptyResults {
                loadingIndicator.startAnimating()
            } else {
                loadingIndicator.stopAnimating()
            }
        case .remoteDataFetched:
            loadingIndicator.stopAnimating()
            emptyView.subtitleLabel.text = L10n.ChannelList.Search.Empty.subtitle("\"\(currentSearchText)\"")
            emptyView.isHidden = !hasEmptyResults
        default:
            loadingIndicator.stopAnimating()
        }
    }

    deinit {
        debouncer.invalidate()
    }
}

/// The view controller responsible to search channels.
open class ChatChannelSearchVC: ChatChannelListSearchVC {
    /// The closure that is triggered whenever a channel is selected from the search result.
    public var didSelectChannel: ((ChatChannel) -> Void)?

    open override var hasEmptyResults: Bool {
        controller.channels.isEmpty
    }

    open override func updateSearchResults(with text: String) {
        guard let currentUserId = controller.client.currentUserId else { return }

        replaceQuery(.init(
            filter: .and([
                .autocomplete(.name, text: text),
                .containMembers(userIds: [currentUserId])
            ])
        ))
    }

    open override func loadMoreSearchResults() {
        loadMoreChannels()
    }

    override open func collectionView(_ collectionView: UICollectionView, didSelectItemAt indexPath: IndexPath) {
        defer {
            collectionView.deselectItem(at: indexPath, animated: true)
        }
        guard let channel = controller.channels[safe: indexPath.row] else { return }
        didSelectChannel?(channel)
    }
}

/// The view controller responsible to search messages.
open class ChatMessageSearchVC: ChatChannelListSearchVC, ChatMessageSearchControllerDelegate {
    /// The `ChatMessageSearchController` instance to perform the messages search.
    public var messageSearchController: ChatMessageSearchController!

    /// The closure that is triggered whenever a message is selected from the search result.
    public var didSelectMessage: ((ChatChannel, ChatMessage) -> Void)?

    /// Component responsible to process an array of `[ListChange<Item>]`'s and apply those changes to a view.
    private lazy var listChangeUpdater: ListChangeUpdater = CollectionViewListChangeUpdater(
        collectionView: collectionView
    )

    private var isPaginatingMessages: Bool = false

    open override var hasEmptyResults: Bool {
        messageSearchController.messages.isEmpty
    }

    override open func setUp() {
        super.setUp()

        messageSearchController.delegate = self
    }

    open override func updateSearchResults(with text: String) {
        messageSearchController.search(text: text)
    }

    open override func loadMoreSearchResults() {
        loadMoreMessages()
    }

    override open func collectionView(_ collectionView: UICollectionView, numberOfItemsInSection section: Int) -> Int {
        messageSearchController.messages.count
    }

    override open func collectionView(_ collectionView: UICollectionView, cellForItemAt indexPath: IndexPath) -> UICollectionViewCell {
        let cell = collectionView.dequeueReusableCell(with: ChatChannelListCollectionViewCell.self, for: indexPath)
        guard let message = messageSearchController.messages[safe: indexPath.item],
              let cid = message.cid,
              let channel = messageSearchController.dataStore.channel(cid: cid) else {
            return cell
        }

        cell.components = components
        cell.itemView.content = .init(
            channel: channel,
            currentUserId: messageSearchController.client.currentUserId,
            searchedMessage: message
        )

        cell.swipeableView.delegate = self
        cell.swipeableView.indexPath = { [weak cell, weak self] in
            guard let cell = cell else { return nil }
            return self?.collectionView.indexPath(for: cell)
        }

        return cell
    }

    override open func collectionView(_ collectionView: UICollectionView, didSelectItemAt indexPath: IndexPath) {
        defer {
            collectionView.deselectItem(at: indexPath, animated: true)
        }

        guard let message = messageSearchController.messages[safe: indexPath.item],
              let cid = message.cid,
              let channel = messageSearchController.dataStore.channel(cid: cid) else {
            return
        }

        didSelectMessage?(channel, message)
    }

    open func loadMoreMessages() {
        guard !isPaginatingMessages else {
            return
        }
        isPaginatingMessages = true

        messageSearchController.loadNextMessages { [weak self] _ in
            self?.isPaginatingMessages = false
        }
    }

    open func controller(_ controller: ChatMessageSearchController, didChangeMessages changes: [ListChange<ChatMessage>]) {
        listChangeUpdater.performUpdate(with: changes)
    }
}
```

My main issue here is that, as I said before, it could be a bit of premature abstraction, especially because it adds a bit of complexity. Instead of having just 2 classes only, we now have 3. Since the duplication is minimal, it could not be worth the additional code and complexity. But I'm fine with both options. Like I said before as well, if we decide not to go with this additional abstraction, we can always introduce `ChatChannelListSearchVC` in the future. But if we do introduce it now, we can never remove it.

DemoApp/StreamChat/StreamChatWrapper+DemoApp.swift
```diff
@@ -40,6 +40,7 @@ extension StreamChatWrapper {
         Components.default.messageListDateOverlayEnabled = true
         Components.default.isVoiceRecordingEnabled = true
         Components.default.messageSwipeToReplyEnabled = true
+        Components.default.channelListSearchFactory = ChannelListSearchStrategy.messages
```
polqf, line 43: Should we try to have consistent naming here? We have Factory vs Strategy. (I like Strategy better, but your call)

nuno-vieira: Yeah, I'm also not 100% ok with this. The thing is that this is actually a `Factory` since it is just a protocol that has 1 implementation to create a searchController. Our Search Strategy happens to conform to that protocol, and so it is also a factory 🤔 

I have 2 ideas:
1- Rename ChannelListSearchStrategy to `ChannelListSearchStrategyFactory` (Kinda weird)
2- Rename `channelListSearchFactory` to `channelListSearchStrategy` and remove the `ChannelListSearchFactory` protocol. Because, if we think about it, if you want to create the search controller from scratch, you don't need this protocol. You can just create it directly in the Channel List. So this factory protocol is a bit unnecessary now that I think about it 🤔

Sources/StreamChatUI/ChatChannelList/Search/ChatChannelListSearchVC.swift
```diff
@@ -0,0 +1,119 @@
...
+
+    // swiftlint:disable unavailable_function
+    /// Performs a request to fetch search results with the given text.
+    ///
+    /// - Parameter text: The text query inputted by the user.
+    open func loadSearchResults(with text: String) {
+        fatalError("This function should be implemented by a subclass.")
+    }
+
+    // swiftlint:disable unavailable_function
```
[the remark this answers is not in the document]
nuno-vieira: I'm not aware of that one, why is that one better? The warning I got was `unavailable_function`, not `next`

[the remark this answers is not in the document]
polqf: `// swiftlint:disable` will disable the rule until you enable it again (`// swiftlint:enable`).
`// swiftlint:disable:next` instead, it only disables it once

At least that's how I've been using it so far 🙃

DemoApp/StreamChat/StreamChatWrapper+DemoApp.swift
nuno-vieira: Will go with option 2 then 👍

Sources/StreamChatUI/ChatChannelList/Search/ChatChannelListSearchVC.swift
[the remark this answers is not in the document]
nuno-vieira: Ah I see, I'll change it then 👍

testableapple (approved):
LGTM 🚢

PR #2718, opened 2023-08-09
[BW] Run cron checks on release

testableapple:


polqf (approved):

.github/workflows/smoke-checks.yml
```diff
@@ -1,19 +1,10 @@
 name: Smoke Checks
 
 on:
-  push:
-    branches:
-      - main
-      - develop
-
```
nuno-vieira, line 8: Why are we removing this?

.github/workflows/smoke-checks.yml
```diff
@@ -1,19 +1,10 @@
...
   pull_request:
     branches:
       - '**'
 
-  release:
-    types:
-      - created
-
```
nuno-vieira, line 16: Here as well?

polqf (approved):

.github/workflows/smoke-checks.yml
testableapple: Why do we need to run some checks again after we executed them on release PR, then on push to `main`, and now on "release created" 🤔

nuno-vieira: What do you mean by "after we executed them on release PR"?

nuno-vieira: These checks should be run on the Release PR, that is my only concern that we are sure they are still run on Release PR

nuno-vieira: All clarified now 👍

nuno-vieira (approved):
LGTM! ✅

.github/workflows/smoke-checks.yml
testableapple: Discussed this with @nuno-vieira, and we decided to remove it eventually.

PR #2720, opened 2023-08-09
Use server generated video thumbnails

polqf:
### 🔗 Issue Links

https://github.com/GetStream/ios-issues-tracking/issues/429

### 🎯 Goal

Use server generated thumbnails when sending a video attachment so other platforms can benefit from it.

### 📝 Summary

Up until now, iOS was sending video attachments without a `thumb_url`. This was leading to issues in other platforms where this parameter is expected. This was not an issue on iOS because we were generating them client side.

### 🛠 Implementation

We are now parsing `thumb_url` as part of the response when uploading a video, and sending it as part of the attachment of the message that it belongs to.

### 🧪 Manual Testing Notes

- Send a video attachment from iOS

Expected:
- The attachment payload contains `thumb_url`
- The attachment thumbnail is properly usable from other platforms

### ☑️ Contributor Checklist


Sources/StreamChat/APIClient/CDNClient/CDNClient.swift
```diff
@@ -18,6 +24,34 @@ public protocol CDNClient {
...
+    /// Uploads attachment as a multipart/form-data and returns the uploaded remote file and its thumbnail.
+    /// - Parameters:
+    ///   - attachment: An attachment to upload.
+    ///   - progress: A closure that broadcasts upload progress.
+    ///   - completion: Returns the uploaded file's information.
+    func uploadAttachment(
+        _ attachment: AnyChatMessageAttachment,
+        progress: ((Double) -> Void)?,
+        completion: @escaping (Result<UploadedFile, Error>) -> Void
+    )
```
nuno-vieira, line 37: This one is similar to the one in the `AttachmentUploader` protocol 🤔 It feels a bit odd, but not sure if we have any other alternative.

Sources/StreamChat/Models/Attachments/ChatMessageVideoAttachment.swift
```diff
@@ -37,9 +39,10 @@ public struct VideoAttachmentPayload: AttachmentPayload {
     /// Creates `VideoAttachmentPayload` instance.
     ///
     /// Use this initializer if the attachment is already uploaded and you have the remote URLs.
-    public init(title: String?, videoRemoteURL: URL, file: AttachmentFile, extraData: [String: RawJSON]?) {
+    public init(title: String?, videoRemoteURL: URL, thumbnailURL: URL?, file: AttachmentFile, extraData: [String: RawJSON]?) {
```
nuno-vieira, line 42: Potential breaking change as well? 🤔

Sources/StreamChat/APIClient/CDNClient/CDNClient.swift
polqf: Yeah, could not find a better way... 🙃 The part that worries me the most is that it can give "ambiguous usage" errors, as the only difference is the Result.Success type.

Maybe the method signature should be a bit different?

nuno-vieira: Hum, I think it is fine to have the same name for now 🤔

nuno-vieira: What about deprecating the other one? @polqf

nuno-vieira (approved):
LGTM

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ _August 09, 2023_
 ## StreamChat
 ### 🐞 Fixed
 - Fix channel list sorting for iOS 14 and below [#2719](https://github.com/GetStream/stream-chat-swift/pull/2719)
+- Fix video attachments not being sent with `thumb_url`, which caused issues in other platforms [#2720](https://github.com/GetStream/stream-chat-swift/pull/2720)
```
nuno-vieira, line 14: This is not in the upcoming, but on 4.35.1

PR #2729, opened 2023-08-14
Update cron checks

testableapple:
### 🎯 Goal

- Previously, we executed LLC tests by cron on different versions of Xcode. Now we want to run them on different versions of iOS.

### 📝 Summary

- Updated devices and iOS versions for corn checks 🌽
- Now Sonar check won't be executed if there are no code changes in LLC
- `Xcodes` is replaced back with `xcversion` (it is faster and does not bring that amount of problems as `xcodes` does). We can replace it back when required
- `prepare_simulator` action now works more stable on older iOS versions
- [There are some flaky tests](https://github.com/GetStream/ios-issues-tracking/issues/515) on iOS 13 and 14 that will need to deal with eventually. For now, we skip them
- Tests on iOS 12 are turned off due to high % of flakiness

### ☑️ Contributor Checklist

### 🎁 Meme

gif


polqf:
A lot of corn here

Issue #2702, opened 2023-07-12
Crash in ConnectionRepository.updateWebSocketEndpoint


AndrewSB:
## What did you do?

initialized the SDK

## What did you expect to happen?

no crash

## What happened instead?

crash

<img width="993" alt="Screenshot 2023-07-12 at 10 27 17 AM" src="https://github.com/GetStream/stream-chat-swift/assets/3814772/d1e1f179-d43c-4138-ad22-839abffcb884">

i've seen this in 4.33.0 as well, i was hoping upgrading to 4.34.0 would solve

## GetStream Environment
**GetStream Chat version:** 4.34.0

nuno-vieira:
Hi @AndrewSB!

Can you export the full crashlog? Can you also share where do you call `connectUser()` ?

Best,
Nuno

AndrewSB:
[com.reclip_issue_ccdf123aa0863e5bad419491cbd5a321_crash_session_b62bdfcccf6e4476b87c5c5fd34fd2ce_DNE_0_v2_stacktrace.txt](https://github.com/GetStream/stream-chat-swift/files/12052296/com.reclip_issue_ccdf123aa0863e5bad419491cbd5a321_crash_session_b62bdfcccf6e4476b87c5c5fd34fd2ce_DNE_0_v2_stacktrace.txt)

[com.reclip_issue_baade0de3bddc89af739ca8476a79fb5_crash_session_4b9d6f81b7db4c408ac5794c4ccd8402_DNE_0_v2_stacktrace.txt](https://github.com/GetStream/stream-chat-swift/files/12052297/com.reclip_issue_baade0de3bddc89af739ca8476a79fb5_crash_session_4b9d6f81b7db4c408ac5794c4ccd8402_DNE_0_v2_stacktrace.txt)

AndrewSB:
we call connectUser soon after init, from inside a `Task {}`

nuno-vieira:
After Init? Which init? We will keep investigating this, but is there a chance you can call connectUser from the Main Thread to see if the crash goes away?

AndrewSB:
`init` of the app

AndrewSB:
yeah, we can try running the connectUser from inside a `MainActor.run {}`

nuno-vieira:
Hi @AndrewSB!

We were able to reproduce this crash and we are already working on fix 👍 

Best,
Nuno

nuno-vieira:
Hi!

[...]

PR #2762, opened 2023-09-01
Fix connecting user from background thread

nuno-vieira:
### 🔗 Issue Links
https://github.com/GetStream/stream-chat-swift/issues/2702

### 🎯 Goal
Make sure it is possible to connect a user from a background thread. Previously, this would cause a crash, or nothing would happen, and the connection would never finish.

### 🛠 Implementation
The root cause of the issue is that every dependency of `ChatClient` was lazily initialised. Since lazy properties are not thread-safe, this would cause unpredictable behaviour. Besides that, there were circular dependencies too.

### 🎨 Showcase

| Before | After |
| ------ | ----- |
| <video src="https://github.com/GetStream/stream-chat-swift/assets/12814114/4970989b-0ec4-464d-80ae-467c24e93942"/> |  <video src="https://github.com/GetStream/stream-chat-swift/assets/12814114/843c772a-47c3-426d-acbf-208afe00ac33"/> |

### 🧪 Manual Testing Notes
1. Connect the User with a Demo App
2. Should connect successfully

We now always connect from a background thread in the DemoApp to make sure we cover this scenario.

### ☑️ Contributor Checklist


Sources/StreamChat/APIClient/APIClient.swift
```diff
@@ -16,10 +16,10 @@ class APIClient {
     let decoder: RequestDecoder
 
     /// Used for reobtaining tokens when they expire and API client receives token expiration error
-    let tokenRefresher: (@escaping () -> Void) -> Void
+    var tokenRefresher: ((@escaping () -> Void) -> Void)?
```
martinmitrevski, line 19: why do we make these optional?

Sources/StreamChat/APIClient/RequestEncoder.swift
```diff
@@ -70,7 +70,7 @@ extension RequestEncoder {
 }
 
 /// The default implementation of `RequestEncoder`.
-struct DefaultRequestEncoder: RequestEncoder {
+class DefaultRequestEncoder: RequestEncoder {
```
martinmitrevski, line 73: Why do we change it to a class?

Sources/StreamChat/ChatClient.swift
```diff
@@ -49,206 +49,48 @@ public class ChatClient {
...
+    // MARK: Repositories
 
-        center.add(middlewares: middlewares)
+    private(set) var connectionRepository: ConnectionRepository
 
-        return center
-    }()
-
-    // MARK: Repositories
+    private(set) var authenticationRepository: AuthenticationRepository
```
martinmitrevski, line 58: Will these change at some point? Maybe `let` would be better?

Sources/StreamChat/ChatClient.swift
```diff
@@ -49,206 +49,48 @@ public class ChatClient {
...
-                    completion: { _ in completion() }
-                )
-            },
-            { [weak self] endpoint in
-                self?.syncRepository.queueOfflineRequest(endpoint: endpoint)
-            }
-        )
-        return apiClient
-    }()
+    var apiClient: APIClient
```
martinmitrevski, line 77: this one is both `var` and the setter is not private.

Sources/StreamChat/ChatClient.swift
```diff
@@ -285,6 +127,94 @@ public class ChatClient {
         self.config = config
         self.environment = environment
 
+        let configuration = config.urlSessionConfiguration
```
martinmitrevski, line 130: maybe a separate setup method for readability?

Sources/StreamChat/ChatClient.swift
```diff
@@ -285,6 +127,94 @@ public class ChatClient {
...
+        self.syncRepository = syncRepository
+        authenticationRepository = authRepository
+        extensionLifecycle = environment.extensionLifecycleBuilder(config.applicationGroupIdentifier)
+        callRepository = environment.callRepositoryBuilder(apiClient)
+        channelRepository = environment.channelRepositoryBuilder(
+            databaseContainer,
+            apiClient
+        )
+
+        authRepository.delegate = self
```
martinmitrevski, line 204: The additional setup delegate stuff could be also moved into separate methods, since this one is becoming too big.

Sources/StreamChat/ChatClient.swift
```diff
@@ -516,24 +446,197 @@ extension ChatClient: ConnectionDetailsProviderDelegate {
...
+            """
+                Missing token refresh provider to get a new token
+                When using expiring tokens you need to provide a way to refresh it by passing `tokenProvider` when \
+                calling `ChatClient.connectUser()`.
+            """
+        }
+    }
+}
+
+class ChatClientFactory {
```
martinmitrevski, line 504: This is nice 👍 Maybe move it in a separate file?

Sources/StreamChat/WebSocketClient/EventMiddlewares/TypingStartCleanupMiddleware.swift
```diff
@@ -14,8 +14,6 @@ class TypingStartCleanupMiddleware: EventMiddleware {
     /// A closure that will be invoked with `stop typing` event when the `incomingTypingStartEventTimeout` has passed
     /// after `start typing` event.
     let emitEvent: (Event) -> Void
-    /// A closure to get a list of user ids to skip typing events for them.
-    let excludedUserIds: () -> Set<UserId>
```
martinmitrevski, line 18: Why do we remove this?

martinmitrevski (commented):
Good that we tackle the lazy vars 👍  We should revisit few things:
- some properties are `var` and others `let`, we should have some consistency (preferably all are constants setup on `init`)
- optionality is introduced in few places, do we need that? 
- we should split up the init method, tends to be hard to follow

Sources/StreamChat/APIClient/RequestEncoder.swift
nuno-vieira: We need the reference semantics. Because not this is created in the init without keeping any reference to the encoder. From what I remember, all requests were failing if we kept them as structs.

Sources/StreamChat/WebSocketClient/EventMiddlewares/TypingStartCleanupMiddleware.swift
nuno-vieira: This was on of the responsible for the circular dependency. We were using the `authenticationRepository` to get the current user, and populate here the `excludedUserIds`. This is over complicated for something that is really simple todo. We have access to the current user in the `DatabaseContainer` inside this `Middleware`. So we have simplified the code, and fixed the circular dependency 👍

Sources/StreamChat/ChatClient.swift
nuno-vieira: Yes this can be all `let` 👍 They were `var` becasue they were lazy before, now we can change them to `let`. Will do 👍

Sources/StreamChat/ChatClient.swift
nuno-vieira: The setter was not private before, I will need to investigate if this is user for testing or not 🤔

Sources/StreamChat/ChatClient.swift
nuno-vieira: Yes I can move it to a separate file 👍

Sources/StreamChat/ChatClient.swift
nuno-vieira: I think it is useful that everything is in the same place. If I move this to a different method, I will also need to pass all the objects required for the setup 🤔 I will see if I can improve this either way

nuno-vieira:
> Good that we tackle the lazy vars 👍 We should revisit few things:
> 
> * some properties are `var` and others `let`, we should have some consistency (preferably all are constants setup on `init`)
> * optionality is introduced in few places, do we need that?
> * we should split up the init method, tends to be hard to follow

> optionality is introduced in few places, do we need that?

Do you mean some closures? Yes we need those because we need to populate them in the init of `ChatClient` and this implementations require the reference of ` ChatClient`, so we need to initialise all the dependencies first, and only then we can populate this closures

Sources/StreamChat/APIClient/APIClient.swift
nuno-vieira: The explaination is here: https://github.com/GetStream/stream-chat-swift/pull/2762#issuecomment-1703066227

Sources/StreamChat/ChatClient.swift
nuno-vieira: ✅ I've also moved the Environemnt struct to a new file as well, to make it consistent.

Sources/StreamChat/ChatClient.swift
nuno-vieira: Improved these ones a bit 👍

Sources/StreamChat/ChatClient.swift
nuno-vieira: This one was not possible to extract to a separate function because it needs to be configured before passing it to the `APIClient`. At this time `self` is not available yet.

PR #2772, opened 2023-09-07
Make `ChannelListLoadingView` Customizable + Fix UI Glitch

nuno-vieira:
### 🔗 Issue Links
Resolves https://github.com/GetStream/ios-issues-tracking/issues/512
Resolves https://github.com/GetStream/ios-issues-tracking/issues/537

### 🎯 Goal
Make `ChannelListLoadingView` Customizable.
Fix a UI Glitch that sometimes would appear as dummy text for a split second.

### 📝 Summary
- Add customization support for `ChannelListLoadingView`
- Fix dummy texts UI Glitch in `ChannelListLoadingView`
- Rename`Components.chatChannelListLoadingView` -> `Components.channelListLoadingView`

### 🛠 Implementation
Currently, our Skeleton View is not very flexible, and ideally, we should have a generic Skeleton View, which would adapt independently of the Channel List Cell layout. Something similar to what we had in a previous WIP PR: https://github.com/GetStream/stream-chat-swift/blob/CIS-1010-Add_States_to_ChannelList/Sources/StreamChatUI/Utils/LoaderView.swift

Either way, some customers complained that the current Channel List Loading View is not customizable at all, and it is not possible to override views or colours. So this PR fixes that and makes sure that this view is now overridable.

In terms of the glitch, it was a quick fix. The layout subviews of the loading cell were being called in the incorrect lifecycle method, and so the bounds of the views at that stage were empty/zero. 

### Glitch Fix Showcase 
| Before | After |
| ------ | ----- |
|  <video src="https://github.com/GetStream/stream-chat-swift/assets/12814114/a72e9f4d-f0e4-4864-8b6b-51b389cd5972"/>   |  <video src="https://github.com/GetStream/stream-chat-swift/assets/12814114/e427ff32-b221-4543-80e9-c7492250c86a"/>  |

### 🧪 Manual Testing Notes
This is easier to reproduce by launching the app in Debug mode, and with a real device. By just launching the app, the glitch would always appear. 

### ☑️ Contributor Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,12 +11,16 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Fix connecting user from background thread [#2762](https://github.com/GetStream/stream-chat-swift/pull/2762)
 
 ## StreamChatUI
+### ✅ Added
+- Add customization support for `ChannelListLoadingView` [#2772](https://github.com/GetStream/stream-chat-swift/pull/2772)
 ### 🐞 Fixed
 - Fix tapping on the status bar scrolling to the bottom instead of the top [#2763](https://github.com/GetStream/stream-chat-swift/pull/2763)
 - Fix empty channel header view for new DM Channels [#2764](https://github.com/GetStream/stream-chat-swift/pull/2764)
 - Fix showing copy message action when text is empty [#2765](https://github.com/GetStream/stream-chat-swift/pull/2765)
+- Fix dummy texts UI Glitch in `ChannelListLoadingView` [#2772](https://github.com/GetStream/stream-chat-swift/pull/2772)
 ### 🔄 Changed
 - Make record button in composer, visible depending on the channel's capabilities. [#2758](https://github.com/GetStream/stream-chat-swift/pull/2758)
+- Rename`Components.chatChannelListLoadingView` -> `Components.channelListLoadingView` [#2772](https://github.com/GetStream/stream-chat-swift/pull/2772)
 
 # [4.36.0](https://github.com/GetStream/stream-chat-swift/releases/tag/4.36.0)
 _August 28, 2023_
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -11,12 +11,16 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
+### ✅ Added
+- Add customization support for `ChannelListLoadingView` [#2772](https://github.com/GetStream/stream-chat-swift/pull/2772)
 ### 🐞 Fixed
 - Fix tapping on the status bar scrolling to the bottom instead of the top [#2763](https://github.com/GetStream/stream-chat-swift/pull/2763)
 - Fix empty channel header view for new DM Channels [#2764](https://github.com/GetStream/stream-chat-swift/pull/2764)
 - Fix showing copy message action when text is empty [#2765](https://github.com/GetStream/stream-chat-swift/pull/2765)
+- Fix dummy texts UI Glitch in `ChannelListLoadingView` [#2772](https://github.com/GetStream/stream-chat-swift/pull/2772)
 ### 🔄 Changed
 - Make record button in composer, visible depending on the channel's capabilities. [#2758](https://github.com/GetStream/stream-chat-swift/pull/2758)
+- Rename`Components.chatChannelListLoadingView` -> `Components.channelListLoadingView` [#2772](https://github.com/GetStream/stream-chat-swift/pull/2772)
```
martinmitrevski, line 23: I don't think this rename is worth it. We would need to manage a deprecation, and the current name is not that bad tbh.

martinmitrevski (commented):
lgtm, just not sure if the rename is worth it.

CHANGELOG.md
nuno-vieira: The name is not consistent with the other view states: 

image

It does not make sense to have `channelListErrorView`, `channelListEmptyView` and then `chatChannelListLoadingView`.

This is a really simple deprecation, with one click it is fixed on the app side, since we use the "renamed" deprecation.

martinmitrevski: We also have `isChatChannelListStatesEnabled`, and most of our types have the prefix `ChatChannelList`, so I don't think it's worth it. However, it's up you in this case.

nuno-vieira: That is why I only changed in this place. The inconsistency is only in the `Components` config. The properties are together, so it makes sense for them to have the same pattern. 
 
Spending time reverting this now is not worth IMO, for such a simple change 🤔

martinmitrevski (approved):

PR #2775, opened 2023-09-13
Make the Logger thread-safe to avoid crashes

nuno-vieira:
### 🔗 Issue Links
Related to https://github.com/GetStream/ios-issues-tracking/issues/270

### 🎯 Goal
Fix crashes related to the Logger not being Thread-safe.

### 🧪 Manual Testing Notes
N/A

### ☑️ Contributor Checklist


Sources/StreamChat/Utils/Logger/Logger.swift
```diff
@@ -178,20 +178,26 @@ public enum LogConfig {
     /// Underlying logger instance to control singleton.
     private static var _logger: Logger?
 
+    private static var queue = DispatchQueue(label: "io.getstream.logconfig")
```
martinmitrevski, line 181: Isn't it better for it to be concurrent with a barrier? (`attributes: .concurrent`)

Sources/StreamChat/Utils/Logger/Logger.swift
```diff
@@ -178,20 +178,26 @@ public enum LogConfig {
...
+                    return logger
+                } else {
+                    _logger = Logger(identifier: identifier, destinations: destinations)
+                    return _logger!
+                }
             }
         }
         set {
-            _logger = newValue
+            queue.async {
```
martinmitrevski, line 198: `async(flags: .barrier)`

Sources/StreamChat/Utils/Logger/Logger.swift
```diff
@@ -209,14 +215,27 @@ public class Logger {
...
+            loggerQueue.async {
+                self._destinations = newValue
+            }
+        }
+    }
+
+    private var _destinations: [LogDestination]
 
-    private let loggerQueue = DispatchQueue(label: "LoggerQueue \(UUID())")
+    private let loggerQueue = DispatchQueue(label: "io.getstream.logger")
```
martinmitrevski, line 233: Same stuff here as well

martinmitrevski (commented):
Looks good, just few comments about the thread safety mechanism

Sources/StreamChat/Utils/Logger/Logger.swift
ipavlidakis: In a serial queue there isn't a point to use async with .barrier i believe. if we change the queue to be concurrent then this suggestion can be applied

martinmitrevski: https://github.com/GetStream/stream-chat-swift/pull/2775#discussion_r1325540582

Sources/StreamChat/Utils/Logger/Logger.swift
nuno-vieira: That is for performance reasons. I think that, in this case, the optimization is not necessary at all 🤔

Sources/StreamChat/Utils/Logger/Logger.swift
nuno-vieira: same as above

Sources/StreamChat/Utils/Logger/Logger.swift
nuno-vieira: same as above

Sources/StreamChat/Utils/Logger/Logger.swift
nuno-vieira: I've made it concurrent 👍 Since this is a logger, it probably makes sense to make sure it does not cause performance issues. Although it would be very unlikely

martinmitrevski, line 181: You forgot to make this one concurrent

martinmitrevski (commented):
One of the queues is still serial

Sources/StreamChat/Utils/Logger/Logger.swift
nuno-vieira: This one is the LogConfig,it is not worth making it concurrent, this is called only once when the logger is creatted

nuno-vieira: Besides that we don't use async here

martinmitrevski: are you sure? We have this code:
```swift
queue.async(flags: .barrier) {
       _logger = newValue
}
```

Tests/StreamChatTests/Utils/Logger_Tests.swift
```diff
@@ -0,0 +1,31 @@
...
+
+import Foundation
+@testable import StreamChat
+import XCTest
+
+final class Logger_Tests: XCTestCase {
+    func test_log_isThreadSafe() {
+        LogConfig.destinationTypes = [ConsoleLogDestination.self]
+
+        DispatchQueue.concurrentPerform(iterations: 100) { _ in
```
ipavlidakis, line 15: Without your changes, is those tests failing?

Sources/StreamChat/Utils/Logger/Logger.swift
nuno-vieira: Ups! This was a mistake. I've changed this queue to concurrent as well, since we can benefit from concurrent reads here.

Tests/StreamChatTests/Utils/Logger_Tests.swift
nuno-vieira: Yes 👍

martinmitrevski (approved):
lgtm ✅

PR #2777, opened 2023-09-14
Fix memory leak caused by `Timer.addTimeout()`

nuno-vieira:
### 🔗 Issue Links
Related to https://github.com/GetStream/ios-issues-tracking/issues/270
Resolves https://github.com/GetStream/ios-issues-tracking/issues/552

### 🎯 Goal
Fix the memory leaks caused by `Timer.addTimeout()`, which was causing the `ChatClient` to leak.

### 🛠 Implementation
The `Timer.addTimeout()` function was really hard to understand. But one thing that is evident is that there is a reference cycle between the `TimerControl` and the `completionCancellingTimer` since both need each other, as we can see here:
```swift
        var timer: TimerControl?
        let completionCancellingTimer: (Result<T, Error>) -> Void = { result in
            timer?.cancel()
            block(result)
        }

        timer = schedule(timeInterval: timeout, queue: .global()) {
            onTimeout()
            completionCancellingTimer(.failure(ClientError.WaiterTimeout()))
        }
```

The function was rewritten and it seems like the logic it is the same. At least all the unit tests are passing and after a smoke test, everything is working fine. But to make sure, it would be nice to get a review from @polqf.

I've also tried making the `var timer: TimerControl?` a weak reference, but it seems it gets instantly deallocated and the code does not work as intended.

### 🧪 Manual Testing Notes
- Open the Demo App with the Debugger ON
- Login and Logout multiple times
- Open the Memory Graph Debug tool
- Only one instance of ChatClient should be alive

### ☑️ Contributor Checklist


martinmitrevski (commented):
Does this one help in solving the crash? If not, would be better if we postpone this one for the next release.

nuno-vieira:
> Does this one help in solving the crash? If not, would be better if we postpone this one for the next release.

Yes this is actually the most crucial one

martinmitrevski (approved):
the change looks ok. Let's have 2 approvals before merging (cc @ipavlidakis) and some smoke QA by @testableapple.

polqf (approved):
Looks good!

PR #2780, opened 2023-09-15
Added threshold for queued messages before being sent

martinmitrevski:
### 🔗 Issue Links

### 🎯 Goal

### 📝 Summary

### 🛠 Implementation

### 🎨 Showcase

| Before | After |
| ------ | ----- |
|  img   |  img  |

### 🧪 Manual Testing Notes

### ☑️ Contributor Checklist

### 🎁 Meme


Sources/StreamChat/Repositories/OfflineRequestsRepository.swift
```diff
@@ -23,6 +23,10 @@ extension Endpoint {
 /// OfflineRequestsRepository handles both the enqueuing and the execution of offline requests when needed.
 /// When running the queued requests, it basically passes the requests on to the APIClient, and waits for its result.
 class OfflineRequestsRepository {
+    enum Constants {
+        static let retryHoursThreshold: CGFloat = 12
+        static let secondsInHour: CGFloat = 3600
+    }
```
nuno-vieira, line 29: Could we move the `retryHoursThreshold` to the actual repository and maybe make it configurable in the `ChatClientConfig`?

martinmitrevski: done

Sources/StreamChat/ChatClient+Environment.swift
```diff
@@ -165,12 +165,14 @@ extension ChatClient {
...
-            _ apiClient: APIClient
+            _ apiClient: APIClient,
+            _ retryHoursThreshold: CGFloat
         ) -> OfflineRequestsRepository = {
             OfflineRequestsRepository(
                 messageRepository: $0,
                 database: $1,
-                apiClient: $2
+                apiClient: $2,
+                retryHoursThreshold: $3
```
ipavlidakis, line 175: Why not simply `retryThreshold` and allow the pass in seconds (like most Foundation APIs)?

martinmitrevski: thought about it, but this one doesn't require such precision and hours should be easier to set for customers. I'm open to changing it if everyone thinks it's better the other way.

nuno-vieira: IMO I think in this one, it makes sense to be in hours because it should be easier to use for customers. But I think it should be an `Int` or, at most `Float` and not `CGFloat`, since the latter is for CoreGraphics, and this is a config for LLC, not UI Kit. `Int` should be enough tho, since it does not make sense to select 3 and a half hours for example...

Sources/StreamChat/Config/ChatClientConfig.swift
```diff
@@ -190,6 +190,9 @@ public struct ChatClientConfig {
     /// The `URLSessionConfiguration` being used as default configuration for the `APIClient` and
     /// `WebSocketClient`
     public var urlSessionConfiguration: URLSessionConfiguration = .default
+    
+    /// How many hours the unsent messages should be queued for sending when the internet connection is available.
+    public var unsentMessagesRetryHoursThreshold: CGFloat = 12
```
nuno-vieira, line 195: What about using `Max` instead of `Retry`? I think "Max" it is more clear 🤔 Retry it feels like it will retry every 12 hours 🤔 
```suggestion
    public var unsentMessagesMaxHoursThreshold: CGFloat = 12
```

Sources/StreamChat/ChatClient+Environment.swift
martinmitrevski: makes sense, changed this

nuno-vieira (approved):
LGTM! ✅

Sources/StreamChat/Config/ChatClientConfig.swift
ipavlidakis: `unsentMessagesMaxHoursThreshold` doesn't really mean anything TBH. What is the it's the max threshold (expressed in hours) of what exactly?

Sources/StreamChat/ChatClient+Environment.swift
ipavlidakis: I would argue that the precision isn't needed as we don't know all the requirements people may have. Making it in seconds it's flexible while maintaining simplicity. The type would also need to be TimeInterval. Not a blocker ofc just stating my side of thinking.

Sources/StreamChat/Config/ChatClientConfig.swift
martinmitrevski: It was renamed into `queuedActionsMaxHoursThreshold`. Open to other suggestions as well.

ipavlidakis (approved):

PR #2785, opened 2023-09-15
Updated the docs for logout and disconnect

martinmitrevski:
### 🎯 Goal

Docs update about logout and disconnect.

### DIFF
--- a/docusaurus/docs/iOS/uikit/getting-started.md
+++ b/docusaurus/docs/iOS/uikit/getting-started.md
@@ -81,6 +81,7 @@ Whenever your users leave the chat component, you should use disconnect to stop 
 
 ```swift
 chatClient.disconnect {
+    // dismiss the current screen or go to another screen
     print("disconnect completed")
 }
 ```
@@ -89,6 +90,7 @@ If your users logout form their account you should use logout instead for comple
 
 ```swift
 chatClient.logout {
+    // dismiss the current screen or go to another screen
     print("logout completed")
 }
 ```
--- a/docusaurus/docs/iOS/uikit/getting-started.md
+++ b/docusaurus/docs/iOS/uikit/getting-started.md
@@ -80,14 +80,20 @@ ChatClient.shared.connectUser(userInfo: userInfo, tokenProvider: tokenProvider) 
 Whenever your users leave the chat component, you should use disconnect to stop receiving chat updates and events while using other features of your app. You disconnect by calling:
 
 ```swift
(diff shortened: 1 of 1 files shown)


docusaurus/docs/iOS/uikit/getting-started.md
```diff
@@ -80,15 +80,21 @@ ChatClient.shared.connectUser(userInfo: userInfo, tokenProvider: tokenProvider)
 Whenever your users leave the chat component, you should use disconnect to stop receiving chat updates and events while using other features of your app. You disconnect by calling:
 
 ```swift
-chatClient.disconnect()
+chatClient.disconnect {
+    print("disconnect completed")
+}
```
nuno-vieira, line 85: Maybe better to show a real example, WDYT?
```suggestion
chatClient.disconnect {
    dismiss(animated: true)
}
```

docusaurus/docs/iOS/uikit/getting-started.md
```diff
@@ -80,15 +80,21 @@ ChatClient.shared.connectUser(userInfo: userInfo, tokenProvider: tokenProvider)
...
 ```
 
 If your users logout form their account you should use logout instead for completely logging out from the session. You logout by calling:
 
 ```swift
-chatClient.logout()
+chatClient.logout {
+    print("logout completed")
+}
```
nuno-vieira, line 93: Same here:
```suggestion
chatClient.logout {
    dismiss(animated: true)
}
```

docusaurus/docs/iOS/uikit/getting-started.md
martinmitrevski: I think it's better to print, since any other action requires some context. Also, `dismiss` is UIKit action, so better to keep it relevant for both UI SDKs. 
If you still think it's better, I can change it, no worries.

nuno-vieira: I think it would be important clarify that they need to wait for the completion block before leaving the screen or going to another screen, it can also be just a comment, like `// dismiss the current screen or go to another screen` 🤔 Otherwise customers might not fully understand what do we mean by "you should wait for the completion block". This is a bit vague, wait for what? They may ask 🤔

martinmitrevski: I like the comment approach more, just updated it 👍

nuno-vieira (approved):

PR #2793, opened 2023-09-19
Fix unread messages banner showing for system messages

nuno-vieira:
### 🔗 Issue Links
None

### 🎯 Goal
Fix regressions introduced by https://github.com/GetStream/stream-chat-swift/pull/2592

### 📝 Summary
Fix the unread messages banner showing on system messages.

### 🧪 Manual Testing Notes
**Fix the unread messages banner showing on system messages**:
1. Truncate a channel w/message and that has unread messages
2. Open this channel
3. The unread banner should not show on the truncated message.

### ☑️ Contributor Checklist


Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
```diff
@@ -1458,7 +1458,15 @@ private extension ChatChannelController {
...
 
     private func getFirstUnreadMessageId() -> MessageId? {
-        guard let lastReadMessageId = channel?.reads.first(where: { $0.user.id == client.currentUserId })?.lastReadMessageId else {
+        let messages = self.messages.filter {
+            $0.type == .regular || $0.type == .reply || $0.type == .deleted
+        }
+
+        let currentUserUnread = channel?.reads.first(where: {
+            $0.unreadMessagesCount > 0 && $0.user.id == client.currentUserId
+        })
```
nuno-vieira, line 1467: When getting the `lastReadMessageId`, we actually need to make sure the message is currenlty unread, otherwise we need to ignore it

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
```diff
@@ -1458,7 +1458,15 @@ private extension ChatChannelController {
     }
 
     private func getFirstUnreadMessageId() -> MessageId? {
-        guard let lastReadMessageId = channel?.reads.first(where: { $0.user.id == client.currentUserId })?.lastReadMessageId else {
+        let messages = self.messages.filter {
+            $0.type == .regular || $0.type == .reply || $0.type == .deleted
+        }
```
nuno-vieira, line 1463: When getting the first unread message, we need to skip system, ephemeral and error messages

Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
```diff
@@ -342,7 +342,7 @@ open class ChatChannelVC: _ViewController,
         willDisplayMessageAt indexPath: IndexPath
     ) {
         let message = chatMessageListVC(vc, messageAt: indexPath)
-        if message?.id == firstUnreadMessageId {
+        if message?.id == firstUnreadMessageId || firstUnreadMessageId == nil {
```
nuno-vieira, line 345: If there is no firstUnreadMessageId, then it means we have seen all messages as well

polqf, line 345: firstUnreadMessageId is initially nil in most of the cases, because of the lifecycle, so this might not be enough

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
```diff
@@ -1458,7 +1458,15 @@ private extension ChatChannelController {
     }
 
     private func getFirstUnreadMessageId() -> MessageId? {
-        guard let lastReadMessageId = channel?.reads.first(where: { $0.user.id == client.currentUserId })?.lastReadMessageId else {
+        let messages = self.messages.filter {
+            $0.type == .regular || $0.type == .reply || $0.type == .deleted
```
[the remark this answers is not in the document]
nuno-vieira: > It would be better not to filter the whole messages list but instead do something like
Makes sense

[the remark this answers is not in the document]
nuno-vieira: > I don't think we want .deleted
It is questionable. I'll check if we have a unread deleted message, if that showsUp in the unread count

Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
nuno-vieira: This seems to be working fine 🤔

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
```diff
@@ -1458,16 +1458,25 @@ private extension ChatChannelController {
     }
 
     private func getFirstUnreadMessageId() -> MessageId? {
-        guard let lastReadMessageId = channel?.reads.first(where: { $0.user.id == client.currentUserId })?.lastReadMessageId else {
-            // We default to the oldest message in the history
-            return messages.last?.id
```
nuno-vieira, line 1463: Why do we return this here? @polqf 

I don't get the logic behind these two tests:
- test_firstUnreadMessageId_whenReadsDoesNotContainCurrentUserId()
- test_firstUnreadMessageId_whenReadsContainsCurrentUserId_whenLastReadMessageIdIsNil()

For me, it looks like it should return `nil`. If we can't get the `lastReadMessageId()` it means there are no unreads, right?

Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
```diff
@@ -370,9 +368,7 @@ open class ChatChannelVC: _ViewController,
             }
         case is MarkUnreadActionItem:
             dismiss(animated: true) { [weak self] in
-                self?.channelController.markUnread(from: message.id) { _ in
-                    self?.updateUnreadMessagesRelatedComponents()
-                }
```
nuno-vieira, line 375: @polqf This is not necessary. Whenever the reads are updated in the channel, the `didUpdateChannel` event is called so we can centralize updating the UI in this delegate event.

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
polqf: > If we can't get the lastReadMessageId it means there are no unreads

It is the other way around. If there are no "reads", it means the first unread message is the first one in the history

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
```diff
@@ -1458,16 +1458,25 @@ private extension ChatChannelController {
...
+        let messages = self.messages.filter {
```
polqf, line 1461: Why is this still here?

Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
```diff
@@ -110,7 +110,9 @@ open class ChatChannelVC: _ViewController,
     private var hasSeenLastMessage: Bool = false
 
     /// The id of the first unread message
-    private var firstUnreadMessageId: MessageId?
+    private var firstUnreadMessageId: MessageId? {
+        channelController.firstUnreadMessageId
```
polqf, line 114: This can be problematic. We are switching back to not having local state. This interferes with what I am working on. We might need alignment.

Plus getting `channelController .firstUnreadMessageId` is not cheap

Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
```diff
@@ -463,6 +459,8 @@ open class ChatChannelVC: _ViewController,
         }
 
         channelAvatarView.content = (channelController.channel, client.currentUserId)
+
+        updateUnreadMessagesRelatedComponents()
```
polqf, line 463: This will now produce unnecessary overhead and extra updates

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
nuno-vieira: How so? We now filter the lastReadMessageId based on the unreadCount. If there are no unread counts and no lastReadMessageId, it means there are no unreads. So how do we default to the first message in history?

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
nuno-vieira: The messages are used in multiple places, it is not enough, change it it only in `messages.last?.id`. Since we use `firstIndex` as well, we can't assume we will only need the first or last message 🤔

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
```diff
@@ -1458,16 +1458,31 @@ private extension ChatChannelController {
...
+
+        // If there are no unreads, then return nil.
+        if unreads.isEmpty {
+            return nil
+        }
+
+        let currentUserUnread = unreads.first(where: {
+            $0.user.id == client.currentUserId
+        })
+
```
polqf, line 1471: The idea would be that you first get the user's "read", and then you apply the rest of the logic.

```
guard let currentUserRead = channel?.reads.first(where: { $0.user.id == client.currentUserId }) else { 
    return .last(where: { $0.type == .regular || $0.type == .reply })?.id
}

guard currentUserUnread.unreadMessagesCount > 0 else { return nil }

guard let lastReadMessageId = currentUserUnread.lastReadMessageId else { 
    return .last(where: { $0.type == .regular || $0.type == .reply })?.id
}
```

Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
polqf: Have you tried after the latest changes? (moving back to keeping the state on the UI)

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
nuno-vieira: Makes sense 👍

Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
nuno-vieira: Yes it is working fine 🤔

nuno-vieira: This should be fine, because whenever we mark the channel read, the `firstUnreadMessageId` will be `nil` and this is what we need to make sure the unread badge or banner does not appear for every new message

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
```diff
@@ -1458,18 +1458,24 @@ private extension ChatChannelController {
     }
 
     private func getFirstUnreadMessageId() -> MessageId? {
-        let messages = self.messages.filter {
-            $0.type == .regular || $0.type == .reply
+        guard let currentUserRead = channel?.reads.first(where: {
+            $0.user.id == client.currentUserId
+        }) else {
+            return nil
```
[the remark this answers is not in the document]
nuno-vieira: No, because if there are no reads from us, it means there are no unreads as well. We only must return `messages.last(where: { $0.type == .regular || $0.type == .reply })?.id` if `unreadCount` is more than zero. So we need to check first the `messagesUnreadCount` value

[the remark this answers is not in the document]
polqf: It could also happen that no reads have yet been created -> no message read -> all of them are unread

polqf (approved):

PR #2794, opened 2023-09-20
[BW] Remove flaky E2E test and make it manual

testableapple:


testableapple:
> Is that an allure scenario now?

Yeah, https://streamio.testops.cloud/project/1/test-cases/2723

PR #2805, opened 2023-09-26
Release fix: Do not mark as read new unseen messages

polqf:
### 🔗 Issue Links

### 🎯 Goal

Do not mark unseen new messages as read. Wait until they are seen

### 📝 Summary

When a user was in a chat, but scrolled up, new messages were being marked as read, even though those were not seen

### 🛠 Implementation

Since 4.37.0, we have been using local flags to keep track of whether the first unread message and the newest message in the history were seen. For the latter, we need to make sure we reset the local flag to false after marking the channel as read, so that we don't mistakenly mark messages as read when they are not seen.

### 🧪 Manual Testing Notes

1. Open a channel
2. Make sure it is marked as read
3. Scroll up so the newest message is not visible on screen
4. Send a message from another device

Expected result:
- The new message should not be marked as read until it is seen

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
+++ b/Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
@@ -345,12 +345,11 @@ open class ChatChannelVC: _ViewController,
         _ vc: ChatMessageListVC,
         willDisplayMessageAt indexPath: IndexPath
     ) {
+        guard !hasSeenFirstUnreadMessage else { return }
+
         let message = chatMessageListVC(vc, messageAt: indexPath)
         if message?.id == firstUnreadMessageId {
             hasSeenFirstUnreadMessage = true
-        }
-        if isLastMessageFullyVisible {
-            hasSeenLastMessage = true
         }
     }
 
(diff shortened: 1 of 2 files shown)


Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
```diff
@@ -254,7 +254,10 @@ open class ChatChannelVC: _ViewController,
 
     /// Marks the channel read and updates the UI optimistically.
     public func markRead() {
-        channelController.markRead { [weak self] _ in
+        channelController.markRead { [weak self] error in
+            if error == nil {
+                self?.hasSeenLastMessage = false
+            }
```
polqf, line 260: This is the actual fix.
After marking the channel as read, we want to reset whether the last cell has been seen or not. 
When it is seen again, this value will be marked as true

Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
```diff
@@ -103,8 +103,8 @@ open class ChatChannelVC: _ViewController,
         channelController.isMarkedAsUnread
     }
 
-    /// Determines whether all unread messages have been seen
-    private var hasSeenAllUnreadMessages: Bool = false
+    /// Determines whether first unread message has been seen
+    private var hasSeenFirstUnreadMessage: Bool = false
```
polqf, line 107: This is just a rename, to be more explicit

nuno-vieira (approved):
LGTM! ✅

PR #2808, opened 2023-09-28
Perf: Add throttling to mark as read

polqf:


Sources/StreamChatUI/Utils/Throttler.swift
```diff
@@ -0,0 +1,54 @@
...
+        workItem?.cancel()
+
+        let workItem = DispatchWorkItem { [weak self] in
+            guard let workItem = self?.workItem, !workItem.isCancelled else { return }
+            action()
+            self?.previousRun = Date()
+            self?.workItem = nil
+        }
+
+        self.workItem = workItem
```
ipavlidakis, line 38: That seems more like a debouncer rather a throttle as you always have one item that is being executed after some interval (and the ones in the middle are being discarded).
I'm using [this](https://web.archive.org/web/20180324022838/http://demo.nimius.net/debounce_throttle/) visualisation to separate those 2.

If you end up needing a debouncer instead of a throttler we already have an implementation in LLC. If you continue with the throttler we will need some tests for each functionality.

polqf: On the tests side of things, we have decided not to go with them for now, until all the assumptions are validated.

On the throttling, it looks quite similar to the debouncing, but this changes it all: https://github.com/GetStream/stream-chat-swift/pull/2808/files#diff-b726cf389a7cfaa48b83a83145188d57562aff5f1eaa6bc8dcb98e1fdc513eedR41

ipavlidakis (approved):
LGTM

Sources/StreamChatUI/Utils/Throttler.swift
nuno-vieira: This one is an awesome article to understand better the difference between Throttling and Debouncing: https://redd.one/blog/debounce-vs-throttle

This class was done based on that as well

PR #2809, opened 2023-09-28
Fix skipping messages logic + performance improvements

nuno-vieira:
### 🎯 Goal
Fixes the logic around skipping messages when receiving messages and scrolling at the top.
It also improves the efficiency of skipping messages by transforming the `LazyCachedMapCollection` to `Array` only when there are message list UI changes.

### 🎨 Showcase
To be filled when merged to develop

### 🧪 Manual Testing Notes
Run the performance load script while scrolling the message list to the top. The message list should not jump + performance should be good.

### ☑️ Contributor Checklist
To be filled when merged to develop

### DIFF
--- a/Sources/StreamChat/Utils/LazyCachedMapCollection.swift
+++ b/Sources/StreamChat/Utils/LazyCachedMapCollection.swift
@@ -70,6 +70,10 @@ public struct LazyCachedMapCollection<Element>: RandomAccessCollection {
             return value
         }
     }
+
+    public func append(_ element: Element) {
+        cache.values.append(element)
+    }
 }
 
 extension LazyCachedMapCollection: ExpressibleByArrayLiteral {
--- a/Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
+++ b/Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
@@ -1059,13 +1059,13 @@ private extension ChatMessageListVC {
         let isNewestChangeInsertion = newestChange?.isInsertion == true
         let isNewestChangeNotByCurrentUser = newestChange?.item.isSentByCurrentUser == false
         let isNewestChangeNotVisible = !listView.isLastCellFullyVisible && !listView.previousMessagesSnapshot.isEmpty
-        let isNewPageLoading = insertions.count > 1 && insertions.count == changes.count
+        let isLoadingNewPage = insertions.count > 1 && insertions.count == changes.count
         let shouldSkipMessages =
             isFirstPageLoaded
                 && isNewestChangeNotVisible
                 && isNewestChangeInsertion
                 && isNewestChangeNotByCurrentUser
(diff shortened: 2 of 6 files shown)


Sources/StreamChat/Utils/LazyCachedMapCollection.swift
```diff
@@ -70,6 +70,10 @@ public struct LazyCachedMapCollection<Element>: RandomAccessCollection {
             return value
         }
     }
+
+    public func append(_ element: Element) {
```
martinmitrevski, line 74: Where do you use this one?

martinmitrevski (commented):
looks good, just one question. @testableapple, please start with the QA as well.

Sources/StreamChat/Utils/LazyCachedMapCollection.swift
nuno-vieira: It is not shown in the changes, but we are appending the Root of the thread in the `getReplies()` function at the end. Previously we were using the `Array`. Now we need for `LazyCachedMapCollection` to support this as well.

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -1052,31 +1054,28 @@ private extension ChatMessageListVC {
...
     }
 
-    func addSkippedMessagesIfNeeded(with changes: [ListChange<ChatMessage>], newestChange: ListChange<ChatMessage>?) {
+    func shouldSkipMessages(with changes: [ListChange<ChatMessage>], newestChange: ListChange<ChatMessage>?) -> Bool {
         let insertions = changes.filter(\.isInsertion)
         let isNewestChangeInsertion = newestChange?.isInsertion == true
         let isNewestChangeNotByCurrentUser = newestChange?.item.isSentByCurrentUser == false
         let isNewestChangeNotVisible = !listView.isLastCellFullyVisible && !listView.previousMessagesSnapshot.isEmpty
-        let hasMultipleInsertions = insertions.count > 1
+        let isNewPageLoading = insertions.count == changes.count
```
polqf, line 1062:
```suggestion
        let isLoadingNewPage = insertions.count == changes.count
```

Sources/StreamChatUI/ChatThread/ChatThreadVC.swift
```diff
@@ -125,9 +125,6 @@ open class ChatThreadVC: _ViewController,
             queueAudioPlayer.dataSource = self
         }
 
-        // Set the initial data
-        messages = getReplies(from: messageController)
```
polqf, line 129: Not setting initial data anymore?

nuno-vieira: This was reverted actually 👍

martinmitrevski (approved):

PR #2813, opened 2023-09-28
Performance Improvements

polqf:
### 🎯 Goal

Aggregator of all the performance improvements added in the `performance-head` branch


Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
```diff
@@ -258,12 +260,14 @@ open class ChatChannelVC: _ViewController,
 
     /// Marks the channel read and updates the UI optimistically.
     public func markRead() {
-        channelController.markRead { [weak self] error in
-            if error == nil {
-                self?.hasSeenLastMessage = false
+        throttler.throttle { [weak self] in
```
nuno-vieira, line 263: Nit: This will make it more consistent with our `Debouncer`
```suggestion
        throttler.execute { [weak self] in
```

nuno-vieira: We will handle this later 👍

martinmitrevski (approved):
LGTM 🚀

PR #2821, opened 2023-10-03
Fix channel list states view when creating a channel and loading cached channels

nuno-vieira:
### 🔗 Issue Links
Resolves https://github.com/GetStream/ios-issues-tracking/issues/546
Resolves https://github.com/GetStream/ios-issues-tracking/issues/576

### 🎯 Goal
- Fix showing empty view when creating a new channel
- Fix showing loading view for cached channels
- Bonus: Fixed some swiftlint warnings

Both are bundled in this PR since they are related.

### 🛠 Implementation
**Fix showing empty view when creating a new channel:**
This was happening because whenever the channels data was changed, we did not re-evaluate the state views. Now we also handle the state view changes whenever there are data changes.

**Fix showing loading view for cached channels:**
Because of the recent introduction of DIffKit in the Channel List, we had a race condition when determining if we should show the Loading View. The problem is that `ChatChannelListVC.channels` will be empty even if we have cached items when opening the channel. The reason is that the item observers are only started when we call `ChannelListController.channels`. So the `didChangeState` event will be called before the `ChatChannelListVC.channels` being populated, so the fix for now is to make sure we use the `controller.channels` directly.

### 🧪 Manual Testing Notes
**Fix showing empty view when creating a new channel:**
1. Open Demo App
2. Login with Jar Jar Binks (It has not channels already)
3. Create a new Channel
4. The channel should appear in the Channel List
5. Remove that same Channel
6. The empty view should reappear

**Fix showing loading view for cached channels:**
1. Open Demo App
2. Login with any user that has channels already
3. Kill the app
4. Open the App
5. Loading View should not appear

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
+++ b/Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
@@ -430,10 +430,10 @@ open class ChatChannelListVC: _ViewController,
 
             switch controller.state {
             case .initialized, .localDataFetched:
-                isLoading = channels.isEmpty
+                isLoading = controller.channels.isEmpty
             case .remoteDataFetched:
                 isLoading = false
-                shouldHideEmptyView = !channels.isEmpty
+                shouldHideEmptyView = !controller.channels.isEmpty
             case .localDataFetchFailed, .remoteDataFetchFailed:
                 shouldHideEmptyView = emptyView.isHidden
                 isLoading = false
@@ -445,7 +445,7 @@ open class ChatChannelListVC: _ViewController,
         } else {
             switch controller.state {
             case .initialized, .localDataFetched:
-                if channels.isEmpty {
+                if controller.channels.isEmpty {
                     loadingIndicator.startAnimating()
                 } else {
                     loadingIndicator.stopAnimating()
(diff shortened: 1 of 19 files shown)


martinmitrevski (approved):
lgtm ✅

Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
```diff
@@ -417,16 +418,22 @@ open class ChatChannelListVC: _ViewController,
     // MARK: - DataControllerStateDelegate
 
     open func controller(_ controller: DataController, didChangeState state: DataController.State) {
+        handleStateChanges(state)
+    }
+
+    /// Called whenever the channels data changes or the controller.state changes.
+    /// It controls the visibility of the channel list state views.
+    open func handleStateChanges(_ newState: DataController.State) {
```
polqf, line 426: I am not sure this one should be public. Any reason I am missing here?

nuno-vieira: The reason is that before the customer could override the `didCHangeState` delegate event and it was enough to customize the channel state view changes. Now we call this in 2 places. So we need to extract this to a new method so that the customer can customize the state changes only in one place 👍

nuno-vieira: Resolving since we discussed it internally ✅

polqf (approved):

PR #2830, opened 2023-10-05
Add DemoShare to Testflight build

testableapple:
Test: https://github.com/GetStream/stream-chat-swift/actions/runs/6420318119


StreamChat.xcodeproj/project.pbxproj
```diff
@@ -12060,7 +12060,7 @@
 				DYLIB_COMPATIBILITY_VERSION = 1;
 				DYLIB_CURRENT_VERSION = 1;
 				DYLIB_INSTALL_NAME_BASE = "@rpath";
-				GENERATE_INFOPLIST_FILE = YES;
+				GENERATE_INFOPLIST_FILE = NO;
```
polqf, line 12063: Could we revert the ones that apply to the framework targets? I don't think this change should be harmful, but I would not like to introduce these changes just before the release, if possible

polqf (approved):

PR #2831, opened 2023-10-05
Background Mapping - Main implementation

polqf:
### 🔗 Issue Links

https://github.com/GetStream/ios-issues-tracking/issues/283

### 🎯 Goal

Make Background Mapping production ready

### 📝 Summary

Background Mapping has been there for a while, but we never got the chance to ship it. This PR intends to make it production ready.

### 🧪 Manual Testing Notes

Things to test in both Channel List and Chat:
- Pagination
- Jump to messages
- Initial load
- Performance

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/Controllers/BackgroundListDatabaseObserver.swift
+++ b/Sources/StreamChat/Controllers/BackgroundListDatabaseObserver.swift
@@ -125,6 +125,14 @@ class BackgroundListDatabaseObserver<Item, DTO: NSManagedObject> {
         }
     }
 
+    private lazy var processingQueue: OperationQueue = {
+        let operationQueue = OperationQueue()
+        operationQueue.underlyingQueue = queue
+        operationQueue.name = "com.stream.background-list-database-observer"
+        operationQueue.maxConcurrentOperationCount = 1
+        return operationQueue
+    }()
+
     deinit {
         releaseNotificationObservers?()
     }
@@ -236,11 +244,26 @@ class BackgroundListDatabaseObserver<Item, DTO: NSManagedObject> {
     }
 
     private func processItems(_ changes: [ListChange<Item>]?) {
-        mapItems { [weak self] items in
-            self?._items = items
-            let changes = changes ?? items.enumerated().map { .insert($1, index: IndexPath(item: $0, section: 0)) }
-            self?.notifyDidChange(changes: changes)
+        let operation = AsyncOperation { [weak self] _, done in
+            guard let self = self else {
+                done(.continue)
+                return
+            }
+
+            self.mapItems { [weak self] items in
+                self?._items = items
+                let returnedChanges: [ListChange<Item>]
+                if let existingChanges = changes {
+                    returnedChanges = existingChanges
+                } else {
+                    returnedChanges = items.enumerated().map { .insert($1, index: IndexPath(item: $0, section: 0)) }
+                }
+                self?.notifyDidChange(changes: returnedChanges)
+                done(.continue)
+            }
         }
+
+        processingQueue.addOperation(operation)
     }
 
     /// Listens for `Will/DidRemoveAllData` notifications from the context and simulates the callback when the notifications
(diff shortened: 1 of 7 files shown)


Sources/StreamChat/Controllers/BackgroundListDatabaseObserver.swift
```diff
@@ -158,12 +166,14 @@ class BackgroundListDatabaseObserver<Item, DTO: NSManagedObject> {
             cacheName: nil
         )
 
+        _ = processingQueue
```
martinmitrevski, line 169: what's this?

Sources/StreamChat/Controllers/BackgroundListDatabaseObserver.swift
```diff
@@ -177,8 +187,6 @@ class BackgroundListDatabaseObserver<Item, DTO: NSManagedObject> {
         guard !isInitialized else { return }
         isInitialized = true
 
-        onWillChange?()
```
martinmitrevski, line 180: why is this not needed anymore?

Sources/StreamChat/Controllers/BackgroundListDatabaseObserver.swift
nuno-vieira: Can't we simply pass the `processingQueue` in the constructor? This solution is a bit confusing. Besides that, I would try to avoid lazy initializations since we had concurrency issues already with this approach.

Sources/StreamChat/Controllers/BackgroundListDatabaseObserver.swift
nuno-vieira: I'm a bit concerned about this one 🤔 Won't this have any breaking behaviour?

Sources/StreamChat/Controllers/BackgroundListDatabaseObserver.swift
```diff
@@ -231,15 +235,43 @@ class BackgroundListDatabaseObserver<Item, DTO: NSManagedObject> {
...
+
+            self.queue.async(flags: .barrier) {
+                self._items = items
+            }
+
+            let returnedChanges = changes ?? items.enumerated().map { .insert($1, index: IndexPath(item: $0, section: 0)) }
+            if notify {
+                self.notifyDidChange(changes: returnedChanges)
+            }
+            onCompletion()
```
nuno-vieira, line 274: Can we add comments on what is `processItems` + `mapItems` + `getInitialItems` doing, please? 🙏 

Another thing that helps reviewing the code is to follow the top-down approach, where high-level functions are at the top, and low-level at the bottom. It is a bit hard reading the functions right now because we need to constantly scroll again to the top to understand the private low-level helpers.

Sources/StreamChat/Controllers/BackgroundListDatabaseObserver.swift
polqf: `processingQueue` is internal details, so it should not be passed from the outside ideally. This approach makes the definition of the queue be in one place, whilst making it thread-safe.

I can move its definition to the init, but that will pollute it a bit.

Sources/StreamChat/Controllers/BackgroundListDatabaseObserver.swift
polqf: It was a mistake on the previous implementation of [BackgroundListDatabaseObserver.swift](https://github.com/GetStream/stream-chat-swift/pull/2831/files/54cb84326cd4a4bcd9c1a2a79babdd5c133ac9b2#diff-f87ca7837e4104c46b1b50f6256905f9120d826b2962fa91e6d4a0a8b7557e03)

Sources/StreamChat/Controllers/BackgroundListDatabaseObserver.swift
polqf: Sure, will add them

Sources/StreamChat/Controllers/BackgroundListDatabaseObserver.swift
nuno-vieira: The problem here is that at first sight `_ = processingQueue` is not clear why this is needed. I would rather configure the queue in the init then if we don't want to pass it from the outside.

nuno-vieira (approved):
LGTM! ✅

PR #2837, opened 2023-10-10
Background Mapping - Entity Observer

polqf:
### 🔗 Issue Links

https://github.com/GetStream/ios-issues-tracking/issues/283

### 🎯 Goal

This is, theoretically, the last big chunk of Background Mapping before we do a bigger testing.
The goal is to move to background the current Entity Observers, using the same approach that we used for the List Observer.

### 📝 Summary

This PR introduces a wrapper for Entity Observers, that based on a flag would use the new background implementation, or the existing foreground one.

### 🛠 Implementation

- New `BackgroundDatabaseObserver` that unifies all the logic for both List and Entity database observers.
- New `EntityDatabaseObserverWrapper` that arbitrates on using one implementation or another based on an `isBackground` flag.
- Tests were added before unifying the logic under `BackgroundDatabaseObserver` to ensure the refactor was a success.

### ☑️ Contributor Checklist


Sources/StreamChat/Controllers/BackgroundDatabaseObserver.swift
```diff
@@ -0,0 +1,277 @@
+//
+// Copyright © 2023 Stream.io Inc. All rights reserved.
+//
+
+import CoreData
+import Foundation
+
+class BackgroundDatabaseObserver<Item, DTO: NSManagedObject> {
```
polqf, line 8: It looks like there are a lot of changes, but it is just the previous implementation of `BackgroundListDatabaseObserver` extracted in another place so it can be reused for Entity Observers

Sources/StreamChat/Controllers/EntityDatabaseObserver.swift
```diff
@@ -284,3 +363,87 @@ private extension ListChangeAggregator {
...
+    /// Called with the aggregated changes after the internal `NSFetchResultsController` calls `controllerDidChangeContent`
+    /// on its delegate.
+    private var listeners: [(EntityChange<Item>) -> Void] = []
+
+    /// Creates a new `ListObserver`.
+    ///
+    /// Please note that no updates are reported until you call `startUpdating`.
+    ///
+    /// - Important: ⚠️ Because the observer uses `NSFetchedResultsController` to observe the entity in the DB, it's required
+    /// that the provided `fetchRequest` has at lease one `NSSortDescriptor` specified.
```
martinmitrevski, line 382: *at least

Sources/StreamChat/Controllers/EntityDatabaseObserver.swift
```diff
@@ -284,3 +363,87 @@ private extension ListChangeAggregator {
...
+    ///
+    /// - Parameters:
+    ///   - context: The `NSManagedObjectContext` the observer observes.
+    ///   - fetchRequest: The `NSFetchRequest` that specifies the elements of the list.
+    ///   - itemCreator: A close the observe uses to convert DTO objects into Model objects.
```
martinmitrevski, line 387: This is a bit unclear, probably "a closure the observer"?

Sources/StreamChat/Controllers/EntityDatabaseObserver.swift
```diff
@@ -284,3 +363,87 @@ private extension ListChangeAggregator {
...
+
+    /// Called with the aggregated changes after the internal `NSFetchResultsController` calls `controllerDidChangeContent`
+    /// on its delegate.
+    private var listeners: [(EntityChange<Item>) -> Void] = []
+
+    /// Creates a new `ListObserver`.
+    ///
+    /// Please note that no updates are reported until you call `startUpdating`.
+    ///
+    /// - Important: ⚠️ Because the observer uses `NSFetchedResultsController` to observe the entity in the DB, it's required
```
martinmitrevski, line 381: What happens if there's no sort descriptor?

TestTools/StreamChatTestTools/Mocks/StreamChat/Controllers/EntityDatabaseObserverWrapper_Mock.swift
```diff
@@ -0,0 +1,30 @@
+//
```
martinmitrevski, line 1: Wrong header here

martinmitrevski (approved):
LGTM! ✅ Added few small notes about wrong comments, but code-wise looks great 👌

Sources/StreamChat/Controllers/EntityDatabaseObserver.swift
polqf: Copy paste 😂

Sources/StreamChat/Controllers/EntityDatabaseObserver.swift
polqf: Copy paste 😂

Sources/StreamChat/Controllers/EntityDatabaseObserver.swift
polqf: A nice crash 😄 

```
*** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: 'An instance of NSFetchedResultsController requires a fetch request with sort descriptors'
```

DatabaseObservers are internal classes, and therefore it is us controlling its usage.

martinmitrevski (approved):

PR #2839, opened 2023-10-11
Background Mapping - Offline Fixes

polqf:
### 🔗 Issue Links

Continuation of https://github.com/GetStream/stream-chat-swift/pull/2831

### 🎯 Goal

The goal of this PR was to fix some deficiencies BG Mapping had when being Offline

### 📝 Summary

There was an assumption that for the initial fetch we would not need to notify the listeners. This was not a problem when being online, given that any interaction with the network would trigger updates, and therefore the listeners would be notified.

This is, though, a big problem when in offline mode, where we depend on the initial data.

### 🛠 Implementation

This PR makes sure the listener is always notified with the initial fetch

### 🧪 Manual Testing Notes

1. Open the app
2. Close the app
3. Disable internet
4. Open the app while offline (should have some content fetched in another session)

Expected result:
- The data in the channel list should be displayed
- Opening a channel results in a message list filled with the messages fetched in a previous session

### ☑️ Contributor Checklist

### DIFF
--- a/Tests/StreamChatTests/BackgroundListDatabaseObserver_Tests.swift
+++ b/Tests/StreamChatTests/BackgroundListDatabaseObserver_Tests.swift
@@ -54,20 +54,18 @@ final class BackgroundListDatabaseObserver_Tests: XCTestCase {
     }
 
     func test_changeAggregatorSetup() throws {
-        // Start observing to ensure everything is set up
-        try observer.startObserving()
-
-        let onDidChangeExpectation = expectation(description: "onDidChange")
-        observer.onDidChange = { _ in
-            onDidChangeExpectation.fulfill()
+        let expectation1 = expectation(description: "onWillChange is called")
+        observer.onWillChange = {
+            expectation1.fulfill()
         }
 
-        let onWillChangeExpectation = expectation(description: "onWillChange")
-        observer.onWillChange = { onWillChangeExpectation.fulfill() }
+        let expectation2 = expectation(description: "onDidChange is called")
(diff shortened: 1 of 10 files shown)


Tests/StreamChatTests/Query/Sorting/ListDatabaseObserver+Sorting.swift
```diff
@@ -371,6 +384,20 @@ final class ListDatabaseObserver_Sorting: XCTestCase {
             ($0, $1, $1)
         })
     }
+
+    private func startObservingAndWaitForInitialUpdate(isBackground: Bool, file: StaticString = #file, line: UInt = #line) throws {
```
martinmitrevski, line 388: This one has the same logic as `startObservingWaitingForInitialResults`, maybe we can reuse it?

Tests/StreamChatTests/BackgroundListDatabaseObserver_Tests.swift
```diff
@@ -115,49 +113,6 @@ final class BackgroundListDatabaseObserver_Tests: XCTestCase {
         XCTAssertFalse(testFRC.test_performFetchCalled)
     }
 
-    func test_updateStillReported_whenSamePropertyAssigned() throws {
```
martinmitrevski, line 118: Why is this one deleted?

martinmitrevski (approved):
LGTM ✅  Added few questions around the tests, but no blockers for merging.

nuno-vieira (approved):

PR #2852, opened 2023-10-25
Reuse logic to get message text content

nuno-vieira:
### 🔗 Issue Links
None

### 🎯 Goal
Reuse the logic to get the message content of a message. 

### 📝 Summary
Right now, this logic is spread across multiple places in the UI SDK. We should aim to make this logic reusable so that it is easier to maintain. For example, when implementing auto translations, it was pretty difficult to track where we should update this logic. 

### 🛠 Implementation
The proposed solution is to extract this logic to a component like `ChatMessageTextRenderer` or `ChatMessageTextPresenter`, which will basically reuse the UI logic of rendering the message text content. Since we don't have presenters or view models, this is a simple solution that could help us for now.

This component replaces the `message.textContent` that had a similar goal, but it lacked information in order to handle all scenarios.

For now, we could make this component internal, and then if we see customers finding it useful to expose it, we can expose it and make it customizable.

### 🎨 Showcase
N/A

### 🧪 Manual Testing Notes
N/A

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
+++ b/Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
@@ -69,6 +69,9 @@ open class ChatMessageContentView: _View, ThemeProvider, UITextViewDelegate {
     /// The component responsible to get the tapped mentioned user in a UITextView
     var textViewUserMentionsHandler = TextViewMentionedUsersHandler()
 
+    /// The component responsible for getting the text content of a message.
+    let messageTextRenderer = ChatMessageTextRenderer()
+
     // MARK: Content && Actions
 
     /// The provider of cell index path which displays the current content view.
@@ -577,19 +580,10 @@ open class ChatMessageContentView: _View, ThemeProvider, UITextViewDelegate {
             setNeedsLayout()
         }
 
-        var text = content?.textContent ?? ""
-
-        // Translated text
-        if layoutOptions?.contains(.translation) == true,
-           let currentUserLang = channel?.membership?.language,
-           let translatedText = content?.translatedText(for: currentUserLang) {
-            text = translatedText
-
-            if let languageText = Locale.current.localizedString(forLanguageCode: currentUserLang.languageCode) {
-                translationLabel?.text = L10n.Message.translatedTo(languageText)
-            }
+        var text = content?.text ?? ""
+        if let message = content, let channel = channel {
+            text = messageTextRenderer.text(for: message, channel: channel)
         }
-
         let attributedText = NSAttributedString(
             string: text,
             attributes: [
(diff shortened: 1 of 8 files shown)


Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageTextRenderer.swift
```diff
@@ -0,0 +1,26 @@
+//
+// Copyright © 2023 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+import StreamChat
+
+/// The component responsible for getting the text content of a message text.
+struct ChatMessageTextRenderer {
+    func text(for message: ChatMessage, channel: ChatChannel) -> String {
```
martinmitrevski, line 10: Do we need the `channel` here? Another alternative could be to provide optional language (and default to the system language if not provided). That way we could use it in `textContent`, without deprecating. Wdyt?

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
```diff
@@ -577,19 +580,10 @@ open class ChatMessageContentView: _View, ThemeProvider, UITextViewDelegate {
...
-        if layoutOptions?.contains(.translation) == true,
-           let currentUserLang = channel?.membership?.language,
-           let translatedText = content?.translatedText(for: currentUserLang) {
-            text = translatedText
-
-            if let languageText = Locale.current.localizedString(forLanguageCode: currentUserLang.languageCode) {
-                translationLabel?.text = L10n.Message.translatedTo(languageText)
-            }
+        var text = content?.text ?? ""
+        if let message = content, let channel = channel {
```
martinmitrevski, line 584: Is there a chance the message is there, while the channel is not available?

martinmitrevski (commented):
nice idea, we definitely need to unify the logic here 👍 
I added a suggestion in the comments, let me know what you think.
In any case, tests for the new text renderer would be needed, as well as CHANGELOG update.

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageTextRenderer.swift
nuno-vieira: Yeah that might be a good idea 👍 I'll look into this

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageContentView.swift
nuno-vieira: In this case, it is very unlikely. But it is probably a good idea to make it future-proof and make the `ChatChannel` optional like you said above 👍

Sources/StreamChatUI/ChatMessageList/ChatMessage/ChatMessageTextRenderer.swift
ipavlidakis: Shall we rename this one to `ChatMessageTextFormatter`?

nuno-vieira:
Closing for now. I'll re-open once I comeback to this.

PR #2878, opened 2023-11-10
[Fix]ComposerView input text repositioning while typing

ipavlidakis:
### 🔗 Issue Links

Resolves #https://github.com/GetStream/ios-issues-tracking/issues/627

### 🎯 Goal

Avoid unnecessary button visibility changes.

### 📝 Summary

Height calculation and scroll position for the textview are causing weird UI glitches

### 🛠 Implementation

Move calculations in one place (layoutSubviews) and allow every flow to go through it. Used https://raw.githubusercontent.com/KennethTsang/GrowingTextView/master/Sources/GrowingTextView/GrowingTextView.swift as guide. Due to this fix we can now safely deprecate the setTextViewHeight method which contained a workaround for one of the UI Issues. One thing to note is that the placeholder label **should not** be embedded in the view as its constraints were conflicting with the text size calculations.

### 🧪 Manual Testing Notes

- Run the app and try the following scenarios
- - Type many lines and then tap the shrink/unshrink button
- - - text position should remain intact
- - Type many lines and paste images while having the caret in top, middle and bottom positions
- - - Images should be pasted as expected and text position should be maintained in every case
- - Paste a very long text
- - - The caret should move at the bottom

### ☑️ Contributor Checklist


martinmitrevski (approved):
LGTM! An approval from @testableapple would also be needed before merging.

nuno-vieira (approved):
LGTM! ✅

testableapple (approved):
LGTM 🙌

opfeffer:
Excited for this 🎉 Does this also work when `shrinkInputButton` and `commandsButton` are hidden?

nuno-vieira:
> Excited for this 🎉 Does this also work when `shrinkInputButton` and `commandsButton` are hidden?

@opfeffer Yes 👍

Sources/StreamChatUI/CommonViews/InputTextView/InputTextView.swift
```diff
@@ -36,6 +36,8 @@ open class InputTextView: UITextView, AppearanceProvider {
         }
     }
 
+    open var placeholderContainer: UIStackView = .init()
```
nuno-vieira, line 39: Can we add a description of the goal of this view? 🙏

martinmitrevski: Important not to slow down the CI today just for this though.

PR #2881, opened 2023-11-10
Move all remaining usage of ListDatabaseObserver to the wrapper

polqf:
### 🔗 Issue Links
N/A

### 🎯 Goal

Enable the SDK to be able to use Background Mapping for all the Database Observers

### 📝 Summary

When Backgorund Mapping was initiated, its focus was on Channel List and Messages List. Now that we have verified it there, we are expanding it to all the other Database Observers

### 🛠 Implementation

Replaces ListDatabaseObserver with ListDatabaseObserverWrapper

### ☑️ Contributor Checklist


TestTools/StreamChatTestTools/Mocks/StreamChat/Controllers/ListDatabaseObserverWrapper_Mock.swift
```diff
@@ -0,0 +1,28 @@
+//
```
martinmitrevski, line 1: Stream header.

Sources/StreamChat/Controllers/MessageController/MessageController.swift
```diff
@@ -80,7 +80,7 @@ public class ChatMessageController: DataController, DelegateCallable, DataStoreP
     public var listOrdering: MessageOrdering = .topToBottom {
         didSet {
             if state != .initialized {
-                _repliesObserver.reset()
```
martinmitrevski, line 83: This is removed because we now `reset` before removing the store coordinator, right?

martinmitrevski (approved):
LGTM ✅ There's on header leftover and one question. Would be good if @testableapple  runs a QA here as well.
Do you think we should include it in the next release?

polqf:
> Do you think we should include it in the next release?

Yes, ideally should be in the following one IMO

PR #2885, opened 2023-11-13
Add new `NewMessageErrorEvent` when observing `EventsController`

nuno-vieira:
### 🔗 Issue Links
Resolves https://github.com/GetStream/ios-issues-tracking/issues/628

### 🎯 Goal
To be able to observe when a message failed to be sent.

### 📝 Summary
A new `NewMessageErrorEvent` can now be observed through `EventsController` which will be triggered when a message failed to be sent. 

### 🛠 Implementation
Because of the async nature of how creating a new message works in the SDK, we need to use the `EventsNotificationCenter` in the `MessageSender` to notify when a message fails to be sent. 

The customer can observe this error using the `EventsController`. We could make the event internal, and expose it only on `ChannelControllerDelegate`, the reason I did not opt for this solution was because it adds additional complexity and makes it less flexible for the customer to where and how it can listen for this event. But let me know your thoughts on this one. 

### 🎨 Showcase
N/A

### 🧪 Manual Testing Notes
1. Send a message which fails (Ex: Without network)
2. It should print on Xcode console the error with `New Message Error:`

### ☑️ Contributor Checklist


Sources/StreamChat/Repositories/MessageRepository.swift
```diff
@@ -71,11 +71,11 @@ class MessageRepository {
...
                     case let .success(payload):
-                        self?.saveSuccessfullySentMessage(cid: cid, message: payload.message) { message in
-                            if let message = message {
+                        self?.saveSuccessfullySentMessage(cid: cid, message: payload.message) { result in
+                            if let message = result.value {
                                 completion(.success(message))
-                            } else {
-                                completion(.failure(.failedToSendMessage))
+                            } else if let error = result.error {
+                                completion(.failure(.failedToSendMessage(error)))
```
martinmitrevski, line 78: we miss a fallback else here or use a switch maybe.

Sources/StreamChat/Repositories/MessageRepository.swift
```diff
@@ -90,22 +90,22 @@ class MessageRepository {
...
         database.write({
             let messageDTO = try $0.saveMessage(payload: message, for: cid, syncOwnReactions: false, cache: nil)
             if messageDTO.localMessageState == .sending || messageDTO.localMessageState == .sendingFailed {
                 messageDTO.locallyCreatedAt = nil
                 messageDTO.localMessageState = nil
             }
 
-            messageModel = try? messageDTO.asModel()
+            let messageModel = try messageDTO.asModel()
+            completion(.success(messageModel))
```
martinmitrevski, line 103: why is it moved here?

Sources/StreamChat/WebSocketClient/Events/MessageEvents.swift
```diff
@@ -224,3 +224,9 @@ class MessageReadEventDTO: EventDTO {
 public struct NewMessagePendingEvent: Event {
     public var message: ChatMessage
 }
+
+// Triggered when a message failed being sent.
+public struct NewMessageErrorEvent: Event {
+    public var messageId: MessageId
```
martinmitrevski, line 230: why not `let`?

martinmitrevski (approved):
LGTM! ✅ Just added few questions. Also, the branch should be updated before merging (watch the changelog).

Sources/StreamChat/Repositories/MessageRepository.swift
nuno-vieira: Nice catch!

Sources/StreamChat/Repositories/MessageRepository.swift
nuno-vieira: Here we garantee the message is always present 👍

nuno-vieira: Because we now have a success and error branches, there's no need to keep a reference of this model, and return an optional value.

Sources/StreamChat/WebSocketClient/Events/MessageEvents.swift
nuno-vieira: Good catch!

PR #2897, opened 2023-11-17
Get rid of the dependency between TestTools and TestHelpers

testableapple:
### 🔗 Issue Links

Resolve https://github.com/GetStream/ios-issues-tracking/issues/644

### 🎯 Goal

- To improve developer experience, move required stuff from TestHelpers to TestTools

### 📝 Summary

- Just moved stuff required by Unit tests from TestHelpers to TestTools

### ☑️ Contributor Checklist

### 🎁 Meme

gif


Package.swift
```diff
@@ -27,6 +27,10 @@ let package = Package(
             targets: ["StreamChatTestMockServer"]
         ),
     ],
+    dependencies: [
```
martinmitrevski, line 30: Why not embed these in the test tools / mock server?

martinmitrevski (commented):
I think we can also address the Difference/Swifter dependency here.
Apart from that, looks good. I assume customers won't have any build errors while updating to this, right?

Package.swift
testableapple: Great idea, let's try that. This will increase the PR grammatically, but that will be just embedded libs.

martinmitrevski (approved):
LGTM ✅

PR #2906, opened 2023-11-21
[Background Mapping] Fix Mark as unread

polqf:
### 🔗 Issue Links

Closes https://github.com/GetStream/ios-issues-tracking/issues/652

### 🎯 Goal

Fix Mark as unread when Background Mapping is enabled

### 📝 Summary

We need to use the data that the API Request returns to us, instead of waiting for the delegate calls.
The most important thing here is that we cannot use the delegate calls because we cannot know if the update is from a “mark as unread” action or a different thing. If we just update the unread banner on the delegate call, we lose all the context as the channel will be marked as read, and the views would then disappear.

### 🛠 Implementation

- The completion of `markUnread(from:completion:)` moves from just returning the `Error` to returning a `Result` object with the data.
- `ChannelController.getFirstUnreadMessageId(for:)` is now public
- `Channel` exposes now a convenience function to get the lastReadMessageId based on a user id (`lastReadMessageId(userId:)`)

### 🧪 Manual Testing Notes

Precondition: BG Mapping Enabled

1. Open a channel with many messages
2. Mark a message as unread

Expected result:
- It should update the "Unread messages" banner to the right location

### ☑️ Contributor Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,8 +3,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 # Upcoming
 
+## StreamChatUI
 ### 🐞 Fixed
 - Fix skip slow mode capability not handled [#2904](https://github.com/GetStream/stream-chat-swift/pull/2904)
+
+### 🔄 Changed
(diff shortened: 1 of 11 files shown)


martinmitrevski (commented):
LGTM codewise ✅ Only spotted one issue while testing (shared on Slack), not sure if it's related to BM. @testableapple should run a QA as well 🙏

PR #2910, opened 2023-11-24
[Background Mapping] Ensure new pages are not requested until the current page has been returned

polqf:
### 🔗 Issue Links

https://github.com/GetStream/ios-issues-tracking/issues/637

### 🎯 Goal

We should not be requesting more than a page at a time when paginating in a Channel

### 📝 Summary

There was no way for us to guarantee that we would not request a new page when already getting one.

### 🛠 Implementation

StatefulScrollViewPaginationHandler  now keeps track of whether there is any active page being fetched, and would not allow a new one until the completion of `onNewTopPage`/`onNewBottomPage` is called.

### 🧪 Manual Testing Notes
1. Open a channel with many pages
2. Scroll fast to the top

Expected result:
The same page should not be requested twice

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChatUI/Utils/ViewPaginationHandling/StatefulScrollViewPaginationHandler.swift
+++ b/Sources/StreamChatUI/Utils/ViewPaginationHandling/StatefulScrollViewPaginationHandler.swift
@@ -1,0 +1,74 @@
+//
+// Copyright © 2023 Stream.io Inc. All rights reserved.
+//
+
+import UIKit
+
+/// A component responsible to handle when to load new pages in a scrollView holding state associated to the content view
+final class StatefulScrollViewPaginationHandler: StatefulViewPaginationHandling {
+    private var bottomPageRequestItemCount: Int?
+    private var topPageRequestItemCount: Int?
+    private var canRequestNewTopPage: Bool {
+        topPageRequestItemCount == nil
+    }
+
+    private var canRequestNewBottomPage: Bool {
+        bottomPageRequestItemCount == nil
+    }
+
+    private weak var scrollView: UIScrollView?
+    private var observation: NSKeyValueObservation?
+    private var previousPosition: CGFloat = 0.0
+
+    var topThreshold: Int = 100
+    var bottomThreshold: Int = 400
+
+    var onNewTopPage: StatefulViewPaginationHandlingBlock?
+    var onNewBottomPage: StatefulViewPaginationHandlingBlock?
+
+    init(scrollView: UIScrollView) {
+        self.scrollView = scrollView
+
+        observation = self.scrollView?.observe(\.contentOffset, changeHandler: { [weak self] scrollView, _ in
+            self?.onChanged(scrollView)
+        })
+    }
+
+    private func onChanged(_ scrollView: UIScrollView) {
+        guard scrollView.isTrackingOrDecelerating && scrollView.contentSize.height > 0 else {
+            return
+        }
+
+        let bottomThreshold = CGFloat(bottomThreshold)
+        let position = scrollView.contentOffset.y
+        if canRequestNewBottomPage, position > scrollView.contentSize.height - bottomThreshold - scrollView.frame.size.height {
+            onNewBottomPage?({ bottomPageRequestItemCount = $0 }, { [weak self] error in
+                if error != nil {
+                    self?.bottomPageRequestItemCount = nil
+                }
+            })
+        }
+
+        let topThreshold = CGFloat(topThreshold)
+        if canRequestNewTopPage, position >= 0 && position <= topThreshold && position <= max(0, previousPosition) {
+            onNewTopPage?({ topPageRequestItemCount = $0 }, { [weak self] error in
+                if error != nil {
+                    self?.topPageRequestItemCount = nil
+                }
+            })
+        }
+
+        previousPosition = position
+    }
+
+    func updateElementsCount(with newCount: Int) {
+        guard bottomPageRequestItemCount != newCount || topPageRequestItemCount != newCount else { return }
+
+        topPageRequestItemCount = nil
+        bottomPageRequestItemCount = nil
+    }
+
+    deinit {
+        observation?.invalidate()
+    }
+}
--- a/Sources/StreamChatUI/Utils/ViewPaginationHandling/StatefulScrollViewPaginationHandler.swift
+++ b/Sources/StreamChatUI/Utils/ViewPaginationHandling/StatefulScrollViewPaginationHandler.swift
@@ -62,7 +62,13 @@ final class StatefulScrollViewPaginationHandler: StatefulViewPaginationHandling 
     }
 
     func updateElementsCount(with newCount: Int) {
-        guard bottomPageRequestItemCount != newCount || topPageRequestItemCount != newCount else { return }
+        var shouldReset = false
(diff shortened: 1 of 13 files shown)


martinmitrevski (approved):

Sources/StreamChatUI/Utils/ViewPaginationHandling/StatefulScrollViewPaginationHandler.swift
```diff
@@ -0,0 +1,75 @@
...
+    var bottomThreshold: Int = 400
+
+    var onNewTopPage: StatefulViewPaginationHandlingBlock?
+    var onNewBottomPage: StatefulViewPaginationHandlingBlock?
+
+    init(scrollView: UIScrollView) {
+        self.scrollView = scrollView
+
+        observation = self.scrollView?.observe(\.contentOffset, changeHandler: { [weak self] scrollView, _ in
+            self?.onChanged(scrollView)
```
martinmitrevski, line 34: this will be called a lot. Do you think it might have impact on performance?

martinmitrevski (commented):
looks good, added couple of questions 👍 
I'm getting one request now as expected. Would love to hear @nuno-vieira's feedback when he's back tomorrow as well.

Sources/StreamChatUI/Utils/ViewPaginationHandling/StatefulScrollViewPaginationHandler.swift
ipavlidakis: It's a shame we can't use `scrollView.publisher(for: \.contentOffset)` and debounce it :/

Sources/StreamChatUI/Utils/ViewPaginationHandling/StatefulScrollViewPaginationHandler.swift
```diff
@@ -0,0 +1,75 @@
...
+
+    var canRequestNewBottomPage: Bool {
+        bottomPageRequestItemCount == nil
+    }
+
+    private weak var scrollView: UIScrollView?
+    private var observation: NSKeyValueObservation?
+    private var previousPosition: CGFloat = 0.0
+
+    var topThreshold: Int = 100
```
ipavlidakis, line 24: it's not clear to me if those values are always expected to be accessed from the same thread. If that's not the case, is it worth protect them behind a serial access?

Sources/StreamChatUI/Utils/ViewPaginationHandling/ViewPaginationHandling.swift
```diff
@@ -11,3 +11,14 @@ protocol ViewPaginationHandling {
     var onNewTopPage: (() -> Void)? { get set }
     var onNewBottomPage: (() -> Void)? { get set }
 }
+
+typealias StatefulViewPaginationHandlingBlock = (((Int) -> Void, @escaping (Error?) -> Void) -> Void)
+protocol StatefulViewPaginationHandling {
```
ipavlidakis, line 16: Is this protocol really necessary? We don't seem to have tests for it (it would be great if we do btw) and this doesn't seem something that will have variations.

Sources/StreamChatUI/Utils/ViewPaginationHandling/StatefulScrollViewPaginationHandler.swift
martinmitrevski: why not use it though? We could put iOS 13 check, and use either the observation or the combine subscription. We anyway might get rid of iOS 11 (we're waiting on the usage data from product).

Sources/StreamChatUI/Utils/ViewPaginationHandling/StatefulScrollViewPaginationHandler.swift
polqf: This is internal only, and the expectation is that it is accessed always during the set up/layout of the view, so would not complicate much here

Sources/StreamChatUI/Utils/ViewPaginationHandling/StatefulScrollViewPaginationHandler.swift
polqf: This is the same code we have in the current implementation of ScrollViewPaginationHandler 😅 . Did not want to create yet another intermediate class given it is not much code that can be shared.

Sources/StreamChatUI/Utils/ViewPaginationHandling/ViewPaginationHandling.swift
polqf: Which kind of tests would you add for the protocol? Tests for its concrete implementation are there

Sources/StreamChatUI/Utils/ViewPaginationHandling/StatefulScrollViewPaginationHandler.swift
polqf: > why not use it though? We could put iOS 13 check, and use either the observation or the combine subscription. We anyway might get rid of iOS 11 (we're waiting on the usage data from product).

It is not only about adding an availability check for iOS 13, you need conditional compiler annotations to allow the usage of AnyCancellable. It can be done, but 🤷

Sources/StreamChatUI/Utils/ViewPaginationHandling/ViewPaginationHandling.swift
ipavlidakis: The reason i'm not sure about the protocol's purpose, is that in the way it's setup now, you can swap out the StatefulViewPaginationHandler inside InvertedScrollViewPaginationHandler and replace it with an instance of another InvertedScrollViewPaginationHandler. It seems to me that the StatefulViewPaginationHandler is more of a store for the ViewPaginationHandlers rather than a ViewPaginationHandler itself.

Sources/StreamChatUI/Utils/ViewPaginationHandling/StatefulScrollViewPaginationHandler.swift
martinmitrevski: The cancellable can be declared as `Any`, but it all depends if there's possible performance impact here. If there isn't, then it's not worth it imo.

polqf: If we are sure that Combine was released initially with all those APIs, we could. I am just not entirely sure about it

polqf: The documentation states iOS 13.0+ for both, so it should be the case 🤔 But `import Combine` does not raise any warnings, while it does when using its APIs

martinmitrevski: I think the main question here is: do you think there's a performance impact? Maybe just a quick instruments run with / without this call can verify this. 
If there isn't, then I don't see value in debouncing and changing the implementation.

polqf: No big performance hit, but debouncer would be preferrable

ipavlidakis (approved):

PR #2917, opened 2023-12-04
Fix `JumpToUnreadMessagesButton` not localizable

nuno-vieira:
### 🔗 Issue Links
Resolves https://github.com/GetStream/ios-issues-tracking/issues/661

### 🎯 Goal
Fixes the button using a hardcoded String instead of being localized.

### 🧪 Manual Testing Notes
- The button should show "X unread".

### ☑️ Contributor Checklist


martinmitrevski (approved):

Sources/StreamChatUI/Resources/en.lproj/Localizable.strings
```diff
@@ -46,6 +46,7 @@
 "message.moderation.delete" = "Delete Message";
 
 "messageList.typingIndicator.typing-unknown" = "Someone is typing";
+"messageList.jump-to-unread-button" = "%d unread";
```
opfeffer, line 49: sorry for the unsolicited feedback but I think ideally this would be in the `Localizable.stringsdict` file instead to allow for proper localization of pluralization 🙏

martinmitrevski: @nuno-vieira the suggestion makes sense (e.g. if you add "message/s" at the end).

opfeffer: yea sorry for not being clearer in my initial comment. its probably ok for English but may be needed for other languages.

PR #2919, opened 2023-12-05
[Enhancement]Expose audioSessionConfiguration for the AudioPlayer

ipavlidakis:
### 🎯 Goal

Allow VoiceRecording feature to work aside with any other stack that also uses/manipulates the shared AVAudioSession instace.

### 📝 Summary

In a WebRTC stack (which controls the AVAudioSession) if the user tries to send a VoiceRecording (chat during a voice/video call), once the message recording finishes, our AudioPlayer & AudioRecorder will try to mark the shared AVAudioSession as deactivated. That causes a conflict with the WebRTC stack which still requires the session to be activated.

### 🛠 Implementation

The AudioSession handling as an implementation of our StreamAudioPlayer and StreamAudioRecorder (and not of them conforming protocols, AudioPlaying and AudioRecording). For this reason, we added a public method which can be used to update the underline `audioSessionConfigurator` on both classes. 

### 🎨 Showcase

An example of usage is the following:
#### Step 1. Create our custom AudioSessionConfigurator
```swift
final class DemoAudioSessionConfigurator: AudioSessionConfiguring {

    private let audioSessionConfigurator: StreamAudioSessionConfigurator

    init() {
        self.audioSessionConfigurator = StreamAudioSessionConfigurator()
    }

    func activateRecordingSession() throws {
        try audioSessionConfigurator.activateRecordingSession()
    }
    
    func deactivateRecordingSession() throws {
//        guard !WebRTCSessionIsActive else {
//            return
//        }
        try audioSessionConfigurator.deactivateRecordingSession()
    }
    
    func activatePlaybackSession() throws {
        try audioSessionConfigurator.activatePlaybackSession()
    }
    
    func deactivatePlaybackSession() throws {
//        guard !WebRTCSessionIsActive else {
//            return
//        }
        try audioSessionConfigurator.deactivatePlaybackSession()
    }
    
    func requestRecordPermission(_ completionHandler: @escaping (Bool) -> Void) {
        audioSessionConfigurator.requestRecordPermission(completionHandler)
    }
}
```

#### Step 2. Update the AudioPlayer instance with our AudioSessionConfigurator
For this step we need to add the following line in our subclass of `ChatChannelVC`
```swift
final class MyChatChannelVC: ChatChannelVC {
...

override func viewDidLoad() {
    ...
    let audioSessionConfigurator = DemoAudioSessionConfigurator()
    // Update the audioPlayer
    (audioPlayer as? StreamAudioPlayer)?.configure(audioSessionConfigurator)
    // Update the audioRecorder
    (messageComposerVC.voiceRecordingVC.audioRecorder as? StreamAudioRecorder).configure(audioSessionConfigurator)
}

...
}
```

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/Audio/AudioPlayer/AudioPlaying.swift
+++ b/Sources/StreamChat/Audio/AudioPlayer/AudioPlaying.swift
@@ -61,7 +61,7 @@ open class StreamAudioPlayer: AudioPlaying, AppStateObserverDelegate {
     /// The player that will be used for the playback of the audio files
     fileprivate let player: AVPlayer
 
-    private let audioSessionConfigurator: AudioSessionConfiguring
+    private var audioSessionConfigurator: AudioSessionConfiguring
 
     /// The assetPropertyLoader is being used during the loading of an asset with non-nil URL, to provide
     /// async information about the asset's properties. Currently, we are only loading the `duration`
@@ -78,6 +78,7 @@ open class StreamAudioPlayer: AudioPlaying, AppStateObserverDelegate {
 
     // MARK: - Lifecycle
 
+    @available(*, deprecated, message: "You can use the Builder API to create a new instance of the player.")
     public required convenience init() {
         self.init(
             assetPropertyLoader: StreamAssetPropertyLoader(),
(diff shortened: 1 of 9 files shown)


Sources/StreamChat/Audio/AudioPlayer/AudioPlaying.swift
```diff
@@ -78,6 +78,7 @@ open class StreamAudioPlayer: AudioPlaying, AppStateObserverDelegate {
 
     // MARK: - Lifecycle
 
+    @available(*, deprecated, message: "You can use the Builder API to create a new instance of the player.")
```
martinmitrevski, line 81: Which builder API? It's not clear atm.

martinmitrevski (commented):
Makes sense to me 👍 Good if we explain in the docs the use-case and how to get around it.

martinmitrevski:
> > Makes sense to me 👍 Good if we explain in the docs the use-case and how to get around it.
> 
> Do you mean, to create a new tutorial or update the existing ones?

Just update the existing docs (not the tutorial), explaining why would you need this (e.g. you have an active audio session going on).

PR #2928, opened 2023-12-07
Fix user mention not tappable when contains "@" character

nuno-vieira:
### 🔗 Issue Links
Resolves https://github.com/GetStream/ios-issues-tracking/issues/665

### 🎯 Goal

Fix user mention not tappable when contains "@" character.
**Bonus**: Fix user mention not tappable if user does not have a name

### 🧪 Manual Testing Notes
1. Mention a user that contains a special "@" character
2. Tap on that user
3. It should log that "showUser" was called

### ☑️ Contributor Checklist


Sources/StreamChatUI/Utils/TextViewMentionedUsersHandler.swift
```diff
@@ -23,7 +23,7 @@ class TextViewMentionedUsersHandler {
         else {
             return nil
         }
-        let name = String(text[range].replacingOccurrences(of: "@", with: ""))
+        let name = String(text[range].dropFirst())
```
opfeffer, line 26: i know this is an implementation detail on our end-- but we actually override and hide the @ symbol in the message. by calling `dropFirst()` youll remove whatever the first letter is... Could you maybe only dropFirst if its the @mention delimiter (@)?

nuno-vieira: Nice catch 👍 I've opted for another approach which should work in every case 👍

martinmitrevski (approved):
LGTM ✅  Changelog needs updating though

PR #2938, opened 2023-12-14
Fix `_dispatch_lane_resume` crash in RepeatingTimer

nuno-vieira:
### 🔗 Issue Links
Resolves https://github.com/GetStream/ios-issues-tracking/issues/676

### 🎯 Goal
Fixes a crash in `RepeatingTimer`. The root cause is because of a double `resume()` call. Although we had protections against this, the reason this still happens very rarely is because this class is called from multiple threads, and so the way to fix it is to make sure the state is accessed in a serial queue.

### 🧪 Manual Testing Notes
N/A

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/Utils/Timers.swift
+++ b/Sources/StreamChat/Utils/Timers.swift
@@ -98,7 +98,9 @@ private class RepeatingTimer: RepeatingTimerControl {
         timer.cancel()
         // If the timer is suspended, calling cancel without resuming
         // triggers a crash. This is documented here https://forums.developer.apple.com/thread/15902
-        timer.resume()
+        if state == .suspended {
+            timer.resume()
+        }
     }
 
     func resume() {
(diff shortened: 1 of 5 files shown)


Sources/StreamChat/Utils/Timers.swift
```diff
@@ -97,24 +98,30 @@ private class RepeatingTimer: RepeatingTimerControl {
         timer.cancel()
         // If the timer is suspended, calling cancel without resuming
         // triggers a crash. This is documented here https://forums.developer.apple.com/thread/15902
-        resume()
+        if state == .suspended {
```
martinmitrevski, line 101: Didn't know you have to resume an already cancelled timer, that's counterintuitive 😅

martinmitrevski (approved):
LGTM ✅  We should also update the changelog.

Sources/StreamChat/Utils/Timers.swift
nuno-vieira: Yup, really weird 🤷‍♂️ This a really old API, we probably should move to something more modern, but for now it should work. If it doesn't we can reconsider it.

PR #2949, opened 2023-12-21
Fix messages failing to be deleted when pending updates to the server

nuno-vieira:
### 🔗 Issue Links
Resolves https://github.com/GetStream/ios-issues-tracking/issues/685

### 🎯 Goal
Fix messages failing to be deleted when pending updates to the server.

### 📝 Summary
- Fix messages failing to be deleted when pending updates to the server
- Makes sure message actions always show at least Edit and Delete action
- Fixes actions not appearing for local-only messages with BG Mapping enabled

### 🛠 Implementation
The root cause of the problem is that we were incorrectly calculating the logic for `isWaitingToBeSentToTheServer`, which has now been renamed to `isLocalOnly` to better describe the goal of the property. Before, if a message was being edited, but not yet published, we would assume it was local only. This is incorrect since the message is already in the server, since it was published once, even tho it does not have the most updated data. 

While debugging this issue, I also fixed a issue that was happening with background mapping. The issue is similar to this one: https://github.com/GetStream/stream-chat-swift/pull/2922. Because the `ChatMessageController.message` is not available on the main thread with background mapping, sometimes it was returning `nil`. 

### 🧪 Manual Testing Notes

**Fix messages failing to be deleted when pending updates to the server:**
N/A
Right now this not manually testable in our Demo app, since we do not allow doing any message action while the message is being edited. 

**Fixes actions not appearing for local-only messages with BG Mapping enabled**
1. Turn off Wi-Fi
2. Send a message
3. Long press the message
4. Message Actions should appear

### ☑️ Contributor Checklist


Sources/StreamChatUI/MessageActionsPopup/ChatMessageActionsVC.swift
```diff
@@ -87,9 +87,10 @@ open class ChatMessageActionsVC: _ViewController, ThemeProvider {
     open var messageActions: [ChatMessageActionItem] {
         guard
             let currentUser = messageController.dataStore.currentUser(),
-            let message = message,
-            message.isDeleted == false
```
martinmitrevski, line 91: why do we remove this condition?

martinmitrevski (approved):
LGTM ✅  Just added one question.

Sources/StreamChatUI/MessageActionsPopup/ChatMessageActionsVC.swift
nuno-vieira: I've reverted this. This is a bit redundant, but better to keep it for now 👍

PR #2951, opened 2023-12-21
Add custom location attachment to Demo App + Mixed attachments improvements

nuno-vieira:
### 🔗 Issue Links
Resolves https://github.com/GetStream/ios-issues-tracking/issues/684

### 🎯 Goal
Adds a custom attachment example to the Demo App. This will make it easier to test features for custom attachments. It is also helpful for customers to see how a custom attachment can be implemented. In the future, we should update the documentation to use this example instead of the dummy "workout" example we currently have.

While working on this, I realised it was not easy for custom attachments to support the mixed attachments rendering, so I made some improvements to make it easier. 

### 📝 Summary
- Adds a custom location attachment to the Demo App
- Makes it easier for custom attachments to support mixed attachment rendering
- Adds the possibility to enable/disable background mapping in the demo app.
- Fixes deleted messages, always showing custom attachments
- Fix blocked messages showing attachments

#### Adds a custom location attachment to the Demo App
For this attachment, we are just saving the coordinates of a location. Then, the app is responsible for showing that location on a map. For better performance, an image snapshot of the map is generated in the message list, and only when tapping the snapshot we open the actual map. We could also save the generated snapshots upload it to the CDN, and save them in the message. This way, Web could use the generated snapshots from iOS or Android, but this is out of scope right now, especially for a Demo App example.

#### Makes it easier for custom attachments to support mixed attachment rendering
Previously, in order to support mixed attachments for our custom attachment was quite complex and not intuitive. It was required to override the `MixedAttachmentInjector`, override the `injectors` property, make sure we did it lazily and make sure we defined the correct order. 

Now, we just need to register our custom attachment to the `MixedAttachmentInjector` registry, and then when overriding the Catalog, make sure to return the `MixedAttachmentInjector` when there are multiples attachment types.

**Example:**
```swift
Components.default.mixedAttachmentInjector.register(.location, with: LocationAttachmentViewInjector.self)

class DemoAttachmentViewCatalog: AttachmentViewCatalog {
    override class func attachmentViewInjectorClassFor(message: ChatMessage, components: Components) -> AttachmentViewInjector.Type? {
        let hasMultipleAttachmentTypes = message.attachmentCounts.keys.count > 1
        let hasLocationAttachment = message.attachmentCounts.keys.contains(.location)
        if hasLocationAttachment {
            if hasMultipleAttachmentTypes {
                return MixedAttachmentViewInjector.self
            }
            return LocationAttachmentViewInjector.self
        }

        return super.attachmentViewInjectorClassFor(message: message, components: components)
    }
}
```

#### ⚠️ Breaking changes
In order to properly support mixed attachments, a minor breaking change was introduced. The `ChatMessageListView` now needs a `ChatMessage` to calculate the `reuseIdentifier` of the cell. The reason is that before we were using only the `MixedAttachmentInjector` type as an identifier. This is incorrect, because depending on the attachments of the message, the underlying injectors will be different, and so the ID must be calculated based on the underlying injectors of the mixed injector. We also don't want to render an attachment view, if the message does not have that attachment, so we need to check the actual attachments from the message.   

### 🎨 Showcase

https://github.com/GetStream/stream-chat-swift/assets/12814114/3e4bb115-52d1-4fab-b721-be3c2974be35

#### Precondition
Location Attachments should be enabled in the Demo App Configuration. 

#### Adds a custom location attachment to the Demo App
Please check the video above

#### Makes it easier for custom attachments to support mixed attachments rendering
1. Add a location with another attachment
2. Both attachments should render in the composer
3. Both attachments should render in the message list

#### Fixes deleted messages, always showing custom attachments
1. Add a location attachment
2. Delete a message with a location attachment
3. "Message deleted" text should appear.

#### Fix blocked messages showing attachments
1. Add a message with an image and a file and type cursing words (so that the message is blocked by moderation)
2. The message should render as system message and not include any attachments.

### ☑️ Contributor Checklist


DemoApp/DemoAppConfiguration.swift
```diff
@@ -19,24 +18,15 @@ enum DemoAppConfiguration {
...
 
-    // MARK: Internal configuration
-
-    static var isStreamInternalConfiguration: Bool {
-        ProcessInfo.processInfo.environment["STREAM_DEV"] != nil
-    }
-
     // This function is called from `DemoAppCoordinator` before the Chat UI is created
     static func setInternalConfiguration() {
-        StreamRuntimeCheck.assertionsEnabled = isStreamInternalConfiguration
```
martinmitrevski, line 30: Maybe we could move everything in one place? At the moment, we set `assertionsEnabled` here, while we have all the other setup in the demo app config.

DemoApp/StreamChat/Components/CustomAttachments/LocationAttachment/LocationAttachmentSnapshotView.swift
```diff
@@ -0,0 +1,131 @@
...
+            activityIndicatorView.centerXAnchor.constraint(equalTo: imageView.centerXAnchor),
+            activityIndicatorView.centerYAnchor.constraint(equalTo: imageView.centerYAnchor)
+        ])
+    }
+
+    @objc func handleTapOnWorkoutAttachment() {
+        didTapOnLocation?()
+    }
+
+    override func updateContent() {
```
martinmitrevski, line 71: this method is a bit long and hard to follow. Would be nice to split it in smaller methods, like setting region, calculating the point of the pin etc.

DemoApp/StreamChat/Components/CustomAttachments/DemoQuotedChatMessageView.swift
```diff
@@ -0,0 +1,28 @@
+//
+// Copyright © 2023 Stream.io Inc. All rights reserved.
```
martinmitrevski, line 2: Make sure to update the new files with the new year before merging.

DemoApp/StreamChat/StreamChatWrapper+DemoApp.swift
```diff
@@ -45,12 +46,17 @@ extension StreamChatWrapper {
         Components.default.channelListSearchStrategy = .messages
 
         // Customize UI components
+        Components.default.attachmentViewCatalog = DemoAttachmentViewCatalog.self
```
martinmitrevski, line 49: Not in the scope of this PR, but I think the demo app becomes very complex in terms of different configurations. Might be hard for integrators to use it as an example for implementation.

Sources/StreamChatUI/ChatMessageList/Attachments/AttachmentViewCatalog.swift
```diff
@@ -15,8 +15,6 @@ open class AttachmentViewCatalog {
         message: ChatMessage,
         components: Components
     ) -> AttachmentViewInjector.Type? {
-        if message.isDeleted { return nil }
```
martinmitrevski, line 18: Why was this check removed?

Sources/StreamChatUI/ChatMessageList/ChatMessageListView.swift
```diff
@@ -80,18 +80,28 @@ open class ChatMessageListView: UITableView, Customizable, ComponentsProvider {
...
     ///   - attachmentViewInjectorType: The type of attachment injector.
     ///   - layoutOptions: The message content view layout options.
+    ///   - message: The message data.
     /// - Returns: The cell reuse identifier.
     open func reuseIdentifier(
         contentViewClass: ChatMessageContentView.Type,
         attachmentViewInjectorType: AttachmentViewInjector.Type?,
-        layoutOptions: ChatMessageLayoutOptions
+        layoutOptions: ChatMessageLayoutOptions,
+        message: ChatMessage?
```
martinmitrevski, line 89: I assume customers didn't override this frequently, right?

Sources/StreamChatUI/ChatMessageList/Attachments/MixedAttachmentViewInjector.swift
```diff
@@ -3,15 +3,66 @@
...
+    /// **Advanced:** You can use the `injectorsRegistry` property directly in case you want to change the default order
+    /// of how different types of attachments are rendered.
+    public static func register(_ type: AttachmentType, with injector: AttachmentViewInjector.Type) {
+        injectorsRegistry.append((type, injector))
+    }
+
+    /// The mixed injectors for the given message.
+    ///
+    /// Given a message, it determines which injectors it should use to render the attachments.
+    public static func injectors(for message: ChatMessage?) -> [AttachmentViewInjector.Type] {
```
martinmitrevski, line 47: Maybe we could write some tests for this method?

Sources/StreamChatUI/ChatMessageList/ChatMessageListView.swift
martinmitrevski: Also, it seems we don't have it mentioned in the public docs.

docusaurus/docs/iOS/uikit/guides/working-with-attachments.md
```diff
@@ -270,7 +270,30 @@ Components.default.attachmentViewCatalog = MyAttachmentViewCatalog.self
 
 The `AttachmentViewCatalog` class is used to pick the `AttachmentViewInjector` for a message, in our case we only need to check if the message contains a workout attachment.
 
-If needed you can also send a message with a workout attachment directly from Swift just to test that everything works correctly:
+In case you want your attachments to be rendered along with other types of attachments, you can do this by adhering your custom injector to the `MixedAttachmentViewInjector`. For this, first, you need to register your custom injector to the `MixedAttachmentViewInjector` registry like so:
```
martinmitrevski, line 273: maybe rephrase this: "by adhering your custom injector to the `MixedAttachmentViewInjector`". When I read it, I get the impression that I should implement a protocol, which is not the case.

martinmitrevski (commented):
Looks great 👌 Added few small things, but nothing major.

DemoApp/StreamChat/Components/CustomAttachments/DemoQuotedChatMessageView.swift
nuno-vieira: Done 👍 I've also updated the Xcode template for upcoming files

DemoApp/StreamChat/StreamChatWrapper+DemoApp.swift
nuno-vieira: Yeah, something we might need to discuss in our tech forum

Sources/StreamChatUI/ChatMessageList/Attachments/AttachmentViewCatalog.swift
nuno-vieira: You also commented on this in the other PR 😂 This logic, was moved to a higher level so that this is not overridden by customers. This logic should apply to every injector (including custom ones), and not only our injectors.

Sources/StreamChatUI/ChatMessageList/ChatMessageListView.swift
nuno-vieira: Yeah this one should be fine. It will be very rare if someone overriden these one.

Sources/StreamChatUI/ChatMessageList/Attachments/MixedAttachmentViewInjector.swift
nuno-vieira: Yeah makes sense!

docusaurus/docs/iOS/uikit/guides/working-with-attachments.md
nuno-vieira: I see, makes sense, I'll simplify it 👍

DemoApp/StreamChat/Components/CustomAttachments/LocationAttachment/LocationAttachmentSnapshotView.swift
nuno-vieira: Yeah, I did not focus much on it because it is Demo code, but I'll improve the readability

DemoApp/DemoAppConfiguration.swift
nuno-vieira: Done ✅

DemoApp/StreamChat/Components/CustomAttachments/LocationAttachment/LocationAttachmentSnapshotView.swift
nuno-vieira: Done ✅

docusaurus/docs/iOS/uikit/guides/working-with-attachments.md
nuno-vieira: Done ✅

Sources/StreamChatUI/ChatMessageList/Attachments/MixedAttachmentViewInjector.swift
nuno-vieira: Done ✅

DemoApp/StreamChat/Components/CustomAttachments/LocationAttachment/LocationAttachmentSnapshotView.swift
martinmitrevski: much better 👍

Sources/StreamChatUI/ChatMessageList/Attachments/AttachmentViewCatalog.swift
martinmitrevski: 1. That was last year
2. I'm old 🤣

martinmitrevski (approved):
LGTM! ✅

PR #2964, opened 2024-01-08
Replace usages of `split` with `components` for Channel ID initializers

bdotdub:
### 🎯 Goal

This makes the `id` computer property on `ChannelId` safer by using `components(separatedBy:)` rather than `split(separator:)`.

### 📝 Summary

We're seeing a case where we're accidentally passing an empty string into `ChannelId(type:id:)`. Because `id` on `ChannelId` blindly does a `split` (ie. the `rawValue` is `messaging:`, we get an index out of bounds exceptions and the app crashes.

This PR uses `components(separatedBy:)` to ensure we have two parts coming back.

> [!NOTE]
> I think `ChannelId(type:id:)` should also throw if `id` is empty, but that seems like a much bigger (and breaking) change so will leave that to you all

### 🎨 Showcase

Tested via `test_channelId_malformed_init_with_type`

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/Models/ChannelId.swift
+++ b/Sources/StreamChat/Models/ChannelId.swift
@@ -10,6 +10,7 @@ import Foundation
 ///
 public struct ChannelId: Hashable, CustomStringConvertible {
     private static let separator: Character = ":"
+    private static let separatorCharSet = CharacterSet(charactersIn: String(separator))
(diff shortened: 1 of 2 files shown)


nuno-vieira (approved):
LGTM! ✅

PR #2966, opened 2024-01-09
Allow resending messages with failed attachments

nuno-vieira:
### 🔗 Issue Links
Resolves https://github.com/GetStream/ios-issues-tracking/issues/697

### 🎯 Goal
Allow resending messages with failed attachments.

### 📝 Summary

Right now, due to some limitations and tech debt, we do not allow editing or resending a message that is currently being sent. This is unfortunate and bad UX, but it is a serious issue if a message with attachments fails to upload the attachments. In this case, the message is still marked as pending, and so it is not possible to edit or delete it, including the attachments. It would be required to retry all the attachments until all of them are successful. 

So in order to fix this bug, there are 3 things we need to do:
1. Mark a message as failed if an attachment fails uploading
2. Re-upload failed attachments when resending a message
3. When all attachments finish uploading, mark message as pending send
     - The user can also retry the attachment directly, instead of the whole message. When doing so, if the attachment is re-sent successfully, we need to resend the message again if all attachments were successfully uploaded. 

#### Precondition
Setup Proxyman to fail to upload file requests

Rule: 
[proxyman_map_local_tree_rules.config.zip](https://github.com/GetStream/stream-chat-swift/files/13908007/proxyman_map_local_tree_rules.config.zip)

#### Scenario 1 - Resend message with a failed attachment
1. Upload a message with an attachment or multiple attachments
2. Attachment should fail
3. Message is marked as failed (Has error indicator)
4. Long press the message
5. Resend message
6. The attachments and the message should be resent successfully

#### Scenario 2 - Retring a failed attachment should resend the message
1. Upload a message with 2 attachments
2. Attachments should fail
3. Message is marked as failed (Has error indicator)
4. Retry attachment 1
5. Attachment 1 is successfully sent
6. Retry attachment 2
7. Attachment 2 is successfully sent, as well as the message

#### Scenario 3 - Deleting a message with failed attachments
1. Upload a message with an attachment or multiple attachments
2. Attachment should fail
3. Message is marked as failed (Has error indicator)
4. Long press the message
5. Delete the message

#### Scenario 4 - Editing a message with failed attachments
1. Upload a message with an attachment or multiple attachments
2. Attachment should fail
3. Message is marked as failed (Has error indicator)
4. Long press the message
5. Edit the message
6. Update the attachments
7. Send the edited message

### 🎨 Showcase
| Scenario 1 | Scenario 2 |
|--------|--------|
| <video src="https://github.com/GetStream/stream-chat-swift/assets/12814114/6da7b4a6-17c2-448f-8c39-66225811da01"/> | <video src="https://github.com/GetStream/stream-chat-swift/assets/12814114/8386d759-e185-480c-a9e5-ba705bd49d72"/> |

| Scenario 3 | Scenario 4 |
|--------|--------|
| <video src="https://github.com/GetStream/stream-chat-swift/assets/12814114/9ceb9939-65c2-4e2f-9861-6092641913b5"/> | <video src="https://github.com/GetStream/stream-chat-swift/assets/12814114/a6d6e84a-0248-48af-be92-b9fbfa421cbb"/> |

### ☑️ Contributor Checklist


Sources/StreamChat/ChatClientFactory.swift
```diff
@@ -43,6 +43,7 @@ class ChatClientFactory {
 
     func makeApiClient(
         encoder: RequestEncoder,
+        databaseContainer: DatabaseContainer,
```
martinmitrevski, line 46: We should avoid adding dependency to the db in the api client. Could this be a leftover actually?

Sources/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -531,7 +531,12 @@ class MessageDTO: NSManagedObject {
 extension MessageDTO {
     /// A possible additional local state of the message. Applies only for the messages of the current user.
     var localMessageState: LocalMessageState? {
-        get { localMessageStateRaw.flatMap(LocalMessageState.init(rawValue:)) }
+        get {
+            if attachments.filter({ $0.localState == .uploadingFailed }).isEmpty == false {
```
martinmitrevski, line 535: Iterating through the attachments on each access to the local message state - could this have performance implications?

Sources/StreamChat/APIClient/AttachmentUploader/AttachmentUploader.swift
```diff
@@ -20,25 +20,34 @@ public protocol AttachmentUploader {
...
                     thumbnailURL: file.thumbnailURL
-                )
-                return uploadedAttachment
-            })
+                )))
+            case let .failure(error):
+                let messageId = attachment.id.messageId
+                self?.database.write { session in
+                    let messageDTO = session.message(id: messageId)
+                    messageDTO?.localMessageState = .sendingFailed
```
martinmitrevski, line 47: I think it's better we keep this one actually? Maybe just move the additional check for all attachments here - in case one after the failed one succeeds. Wdyt?

martinmitrevski (commented):
Looks good, although I think we should do small modifications (see comments). Also, please check why the tests are failing.

Sources/StreamChat/ChatClientFactory.swift
nuno-vieira: The DB is not part of the APIClient, actually, it is part of the StreamAttachmentUploader, which is a dependency of the APIClient. But the APIClient does not have a direct reference to the DB

nuno-vieira:
@martinmitrevski Resolved all comments since the PR was still in Draft, and not ready for review. Now it is 👍

martinmitrevski (approved):
great work 👏  Before merging - It seems something is broken with vale, it breaks the test builds.

testableapple (approved):
LGTM 🙌

nuno-vieira:
LLC Tests and UI tests are passing locally. Will merge it and tackle the vale issue on another PR.

PR #2967, opened 2024-01-12
Fix mentioned users without name and if not in local device

nuno-vieira:
### 🔗 Issue Links
Resolves https://getstream.slack.com/archives/C04E70P9HJP/p1704309904112239

### 📝 Summary
- Fix not mentioning users if they are not fetched in the local device
- Fix mentioned users without names not highlighted

### 🛠 Implementation
We were using the whole `ChatUser` object in order to mention a user when creating a message. This is not needed, since we only need the user ids. Besides this, with our current architecture, it means that if the `ChatUser` is not available in the local DB, the user would not be mentioned.

#### Fix not mentioning users if they are not fetched in the local device:
Make sure there are no regressions when mentioning a user.

#### Fix mentioned users without names not highlighted
1. Go to a channel with users without names (ex: 4.47.0, `tomasso-XXX`) 
2. Write "@.." to show the mention users suggestions
3. Tap on `tomasso-XXXX`
4. Send the message
5. The message should highlight the user id

### ☑️ Contributor Checklist


Sources/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -607,13 +612,7 @@ extension NSManagedObjectContext: MessageDatabaseSession {
             }
         )
 
-        // If a user is able to mention someone,
-        // most probably we have that user already saved in DB.
-        // Ideally, this should be `loadOrCreate` but then
-        // we miss non-optional fields of DTO and fail to save.
-        message.mentionedUsers = Set(
```
martinmitrevski, line 614: What's not clear to me is: when would the `mentionedUsers` value be set, now that we delete this code? We mention in a comment above that this should still be used for reading mentioned users.

martinmitrevski (commented):
Just have one question, otherwise looks good.

Sources/StreamChat/Database/DTOs/MessageDTO.swift
nuno-vieira: This is set on the `saveMessage()` whenever we fetch channels, messages from the servers

martinmitrevski (approved):
LGTM ✅

testableapple (approved):
LGTM 🚢

PR #2972, opened 2024-01-16
Remove data instead of deleting store when logging out

nuno-vieira:
### 🔗 Issue Links
Resolves https://github.com/GetStream/ios-issues-tracking/issues/625

### 🎯 Goal
Fix crashes when logging and logging out. 

### 📝 Summary
We reached a point at which it is very difficult to avoid crashes the way we are deleting CoreData data when logging out. Deleting the persistence store is very risky because it is really hard to make sure there are no references to old managed object contexts, especially with `CoreDataLazy` and `LazyCachedMapCollection`. 

### 🛠 Implementation
This PR changes how we delete the whole data from CoreData. Instead of deleting the persistence store, we delete all data from CoreData using delete batch requests. 

### 🧪 Manual Testing Notes
Overall Smoke testing.

### ☑️ Contributor Checklist


Sources/StreamChat/Database/DatabaseContainer.swift
```diff
@@ -257,31 +239,20 @@ class DatabaseContainer: NSPersistentContainer {
         }
```
martinmitrevski, line 239: There's a fatal error here, probably good if we get rid of it.

Sources/StreamChat/Database/DTOs/AttachmentDTO.swift
```diff
@@ -9,9 +9,9 @@ import Foundation
 class AttachmentDTO: NSManagedObject {
     /// An attachment id.
     @NSManaged private var id: String
-    var attachmentID: AttachmentId {
-        get { AttachmentId(rawValue: id)! }
-        set { id = newValue.rawValue }
+    var attachmentID: AttachmentId? {
```
martinmitrevski, line 12: Why do we change this?

Sources/StreamChat/Database/DatabaseContainer.swift
```diff
@@ -257,31 +239,20 @@ class DatabaseContainer: NSPersistentContainer {
...
+            let entityNames = self?.managedObjectModel.entities.compactMap(\.name)
+            var deleteError: Error?
+            entityNames?.forEach { [weak self] entityName in
+                let deleteFetch = NSFetchRequest<NSFetchRequestResult>(entityName: entityName)
+                let deleteRequest = NSBatchDeleteRequest(fetchRequest: deleteFetch)
+                do {
+                    try self?.writableContext.execute(deleteRequest)
+                    try self?.writableContext.save()
+                } catch {
+                    log.error("Batch delete request failed with error \(error)")
```
martinmitrevski, line 251: Since there's a possibility that this operation can fail, we should also do what you suggested before - on login we check if we have existing data from another user and delete it, which should cover this failure. Wdyt?

Sources/StreamChat/Database/DatabaseContainer.swift
```diff
@@ -257,31 +239,20 @@ class DatabaseContainer: NSPersistentContainer {
...
+                    deleteError = error
```
martinmitrevski, line 252: Can we simulate one failure and inspect the db file that remains?

Sources/StreamChat/Database/StreamChatModel.xcdatamodeld/StreamChatModel.xcdatamodel/contents
```diff
@@ -140,7 +140,7 @@
         <relationship name="devices" optional="YES" toMany="YES" deletionRule="Nullify" destinationEntity="DeviceDTO" inverseName="user" inverseEntity="DeviceDTO"/>
         <relationship name="flaggedMessages" optional="YES" toMany="YES" deletionRule="Nullify" destinationEntity="MessageDTO" inverseName="flaggedBy" inverseEntity="MessageDTO"/>
         <relationship name="flaggedUsers" optional="YES" toMany="YES" deletionRule="Nullify" destinationEntity="UserDTO" inverseName="flaggedBy" inverseEntity="UserDTO"/>
-        <relationship name="mutedUsers" toMany="YES" deletionRule="Nullify" destinationEntity="UserDTO" inverseName="mutedBy" inverseEntity="UserDTO"/>
+        <relationship name="mutedUsers" toMany="YES" deletionRule="Cascade" destinationEntity="UserDTO" inverseName="mutedBy" inverseEntity="UserDTO"/>
```
martinmitrevski, line 143: Do we really need cascade in this case, can you clarify?

Sources/StreamChat/Database/DTOs/CurrentUserDTO.swift
```diff
@@ -139,35 +139,8 @@ extension NSManagedObjectContext: CurrentUserDatabaseSession {
...
-        // we already have cached value in `userInfo` so all setup is complete
-        // so we can just return cached value
-        if let currentUser = userInfo[Self.currentUserKey] as? CurrentUserDTO {
-            return currentUser
-        }
-
-        // we do not have cached value in `userInfo` so we try to load current user from DB
-        if let currentUser = CurrentUserDTO.load(context: self) {
-            // if we have current user we save it to `userInfo` so we do not have to load it again
-            userInfo[Self.currentUserKey] = currentUser
```
martinmitrevski, line 155: Why do we remove this caching? Don't think it has an impact on the crash, while it might impact performance.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### 🐞 Fixed
 - Fix not possible to resend messages with failed attachments [#2966](https://github.com/GetStream/stream-chat-swift/pull/2966)
 - Fix not mentioning users if they are not fetched in the local device [#2967](https://github.com/GetStream/stream-chat-swift/pull/2967)
+- Fix `startObserver()` crash in login/logout flow [#2972](https://github.com/GetStream/stream-chat-swift/pull/2972)
```
martinmitrevski, line 10: Can we reword this, describing it's an edge case? Reading it like this, customers will get the impression this happens frequently, which is not the case.

Sources/StreamChat/Database/StreamChatModel.xcdatamodeld/StreamChatModel.xcdatamodel/contents
martinmitrevski: Clarified on Slack

Sources/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -1183,7 +1183,7 @@ private extension ChatMessage {
         $_author = ({ user }, nil)
         $_attachments = ({
             dto.attachments
-                .map { $0.asAnyModel() }
+                .compactMap { $0.asAnyModel() }
```
ipavlidakis, line 1186: This compactMap here will completely absorb the case where attachments can't be converted. Shall we surface the error before doing the compactMap?

Sources/StreamChat/Database/DatabaseContainer.swift
```diff
@@ -257,31 +239,20 @@ class DatabaseContainer: NSPersistentContainer {
...
         writableContext.perform { [weak self] in
-            self?.sendNotificationForAllContexts(name: Self.WillRemoveAllDataNotification)
-
-            // Because this is a critical part of Stream, we don't want to make such changes unless it is in a controlled environment
-            if StreamRuntimeCheck._isBackgroundMappingEnabled {
-                // Before we remove the persistent store, we need to reset the state of the `NSManagedObjectContext`s to
-                // avoid a context from trying to access data from a previous store
-                self?.allContext.forEach { context in
-                    context.performAndWait { context.reset() }
+            let entityNames = self?.managedObjectModel.entities.compactMap(\.name)
```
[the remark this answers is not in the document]
ipavlidakis: I imagine, inside the writable's context

Sources/StreamChat/Database/DatabaseContainer.swift
```diff
@@ -257,31 +239,20 @@ class DatabaseContainer: NSPersistentContainer {
         }
 
         writableContext.perform { [weak self] in
```
ipavlidakis, line 241: does it makes sense to have it as a performAndWait?

Sources/StreamChat/Database/StreamChatModel.xcdatamodeld/StreamChatModel.xcdatamodel/contents
ipavlidakis, line 143: Does this mean that we delete all those users from the DB, right only because we deleted the CurrentUser? I just wanted to confirm that's indeed expected.

Sources/StreamChat/Database/DatabaseContainer.swift
```diff
@@ -257,31 +239,20 @@ class DatabaseContainer: NSPersistentContainer {
...
-                self?.allContext.forEach { context in
-                    context.performAndWait { context.reset() }
+            let entityNames = self?.managedObjectModel.entities.compactMap(\.name)
+            var deleteError: Error?
+            entityNames?.forEach { [weak self] entityName in
+                let deleteFetch = NSFetchRequest<NSFetchRequestResult>(entityName: entityName)
+                let deleteRequest = NSBatchDeleteRequest(fetchRequest: deleteFetch)
+                do {
+                    try self?.writableContext.execute(deleteRequest)
+                    try self?.writableContext.save()
```
ipavlidakis, line 249: Does it make sense to only save once all entities have been completed? In that way we can at least have a state where we can clearly interact when we try to login.

Sources/StreamChat/Database/DatabaseContainer.swift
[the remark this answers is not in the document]
nuno-vieira: Yes exactly, like @ipavlidakis said

Sources/StreamChat/Database/DTOs/AttachmentDTO.swift
nuno-vieira: This was dangerous already, and in the past I did get a crash on this sometimes. With the current setup now I was able to reproduce it, so I fixed it. Most likely it is because we can have a corrupted attachment when logging out or someting. So to be safe, it is better to avoid the force unwrap

Sources/StreamChat/Database/DTOs/CurrentUserDTO.swift
nuno-vieira: This was also a source of potential crashes. The `userInfo` is not thread-safe so we should not be using it for caching. Besides that, we do not have that observer anymore, so there would be no way to erase this. I think for now this should have a big impact, but if we want to have caching, it needs to be done in higher-level, not in the DTO directly

Sources/StreamChat/Database/DTOs/MessageDTO.swift
nuno-vieira: There's no reason for it to fail unless there is a logout process going on or something. The logging will make the code a bit harder to read, so I'm not sure it is worth it 🤔

Sources/StreamChat/Database/DTOs/CurrentUserDTO.swift
martinmitrevski: For the current user, it makes sense to have caching - it's frequently being accessed. Yes, please try to keep the caching, on a higher-level is also fine.

Sources/StreamChat/Database/StreamChatModel.xcdatamodeld/StreamChatModel.xcdatamodel/contents
martinmitrevski: I would say that's the expected behaviour. When you delete the current user, all the related data should be gone too.

Sources/StreamChat/Database/DatabaseContainer.swift
martinmitrevski: Hmm, what if things fail, or the app is killed by the user? We won't have anything deleted. At least in this way we delete step by step (we could also put the higher prio ones first?)

Sources/StreamChat/Database/DatabaseContainer.swift
nuno-vieira: Done ✅

Sources/StreamChat/Database/DTOs/MessageDTO.swift
ipavlidakis: It will make it easier to debug and will provide integrators with some logs to communicate with us. Readability won't be affected as the one liner is actually dangerous on hide the error here.

Sources/StreamChat/Database/DatabaseContainer.swift
nuno-vieira: Yeah I thought about this as well 🤔 It could help make sure this is executed 👍

Sources/StreamChat/Database/DatabaseContainer.swift
ipavlidakis: If things fail (I'm guessing that this if the app gets killed before it completes) that's ok because then we come back, on login(or even on startup) we can erase everything.

Sources/StreamChat/Database/DatabaseContainer.swift
nuno-vieira: Changed it 👍

Sources/StreamChat/Database/DatabaseContainer.swift
nuno-vieira: The performAndWait will probably make sure we are safer against not finishing the operation. But we might be overthinking here, we clear the data again whenever we login with a different user, so it is fine.

CHANGELOG.md
nuno-vieira: Done ✅

Sources/StreamChat/Database/DTOs/MessageDTO.swift
nuno-vieira: Done 👍

Sources/StreamChat/Database/DatabaseContainer.swift
martinmitrevski: maybe just put an if statement for a particular fetch request and skip the delete? Find the db file from the file system on a simulator, and let's inspect it.

Sources/StreamChat/Database/DTOs/CurrentUserDTO.swift
nuno-vieira: Moving it to a higher-level actually required a lot of work, so for now I'm leaving it like it was. But now we properly invalidate the cache. Before, there could even be an issue where we did not properly invalidate the current user since we only deleted the cache on logout. But if the current user changes, we also need to invalidate the cache.

Sources/StreamChat/Database/DatabaseContainer.swift
nuno-vieira: ### Removing all data

Before:
image

After:
image

### Removing all data with mute fails
Before:
image

After:
image

We can see in both cases, there is a signifcant drop of MB's in the `8br4watad788` db file. But ofc, we can't really read it to check what is happening since the file is not readable.

martinmitrevski (approved):
✅

ipavlidakis (approved):

testableapple (approved):
❤️

PR #2978, opened 2024-01-22
XCMetrics

testableapple:
### 🔗 Issue Links

### 🎯 Goal

### 📝 Summary

### 🛠 Implementation

### 🎨 Showcase

| Before | After |
| ------ | ----- |
|  img   |  img  |

### 🧪 Manual Testing Notes

### ☑️ Contributor Checklist

### 🎁 Meme


StreamChatUITestsAppUITests/Tests/Performance/LoadTime.swift
```diff
@@ -0,0 +1,42 @@
...
+                MessageListPage.cells.waitCount(1)
+                stopMeasuring()
+                userRobot.tapOnBackButton().logout().login()
+            }
+        }
+    }
+    
+    func testChannelList() {
+        WHEN("channels are generated") {
+            backendRobot.generateChannels(count: 1, messagesCount: 200)
```
martinmitrevski, line 29: 1 channel is not enough, we should test with the default of 25. Parsing larger channels is one of the things we should track. Messages count can also be 25, doesn't need to be 200 for the channel list.

StreamChatUITestsAppUITests/Tests/Performance/ScrollTime.swift
```diff
@@ -0,0 +1,33 @@
...
+//
+
+import XCTest
+
+@available(iOS 15.0, *)
+class ScrollTime: StreamTestCase {
+    
+    func testMessageList() {
+        WHEN("user opens the message list") {
+            backendRobot.generateChannels(count: 1, messagesCount: 200)
```
martinmitrevski, line 12: Not sure if it's possible with the tooling, but here's an idea to test: Generate first 25 messages, scroll up few times, and then the backend robot generates 25 more. We will measure appending those messages while scrolling, which can also cause a frame drop.

StreamChatUITestsAppUITests/Tests/Performance/ScrollTime.swift
```diff
@@ -0,0 +1,33 @@
...
+        THEN("user scrolls the message list") {
+            measure(metrics: [XCTOSSignpostMetric.scrollingAndDecelerationMetric]) {
+                userRobot.scrollMessageListUp()
+            }
+        }
+    }
+    
+    func testChannelList() {
+        WHEN("user opens the channel list") {
+            backendRobot.generateChannels(count: 200)
```
martinmitrevski, line 24: Same here, 200 is not realistic to have, since we also have limit on the backend of 25 (I think). We could try the same pagination thing, if possible.

StreamChatUITestsAppUITests/Tests/Performance/LoadTime.swift
```diff
@@ -0,0 +1,42 @@
...
+//
+
+import XCTest
+
+@available(iOS 15.0, *)
+class LoadTime: StreamTestCase {
+    
+    func testMessageList() {
+        WHEN("user logs in") {
+            backendRobot.generateChannels(count: 1, messagesCount: 200)
```
martinmitrevski, line 12: What do the robots generate? Is it objects directly in memory or is decoding included? Since decoding takes the most of the time when loading new data.

martinmitrevski (commented):
This is looking really good 👏  Before merging, let's see if it's possible to do some of the suggestions (see comments).

StreamChatUITestsAppUITests/Tests/Performance/LoadTime.swift
testableapple: Nice catch, copy-pasted from `testMessageList` and forgot to update.

StreamChatUITestsAppUITests/Tests/Performance/ScrollTime.swift
testableapple: Hmm, I don't get it. In this test we generate a channel with a history of 200 messages. When user opens the channel and scrolls it, paginations goes into play and loads new pages after every 25 messages.

StreamChatUITestsAppUITests/Tests/Performance/LoadTime.swift
testableapple: The BackendRobot generates a json that will be returned to the SDK on `/channels/*` call

StreamChatUITestsAppUITests/Tests/Performance/LoadTime.swift
testableapple: Updated 👍

StreamChatUITestsAppUITests/Tests/Performance/ScrollTime.swift
testableapple: You described the pagination logic, didn't you? If so, this is how it works at the moment 🤔

StreamChatUITestsAppUITests/Tests/Performance/LoadTime.swift
martinmitrevski: great, that's what we need

StreamChatUITestsAppUITests/Tests/Performance/ScrollTime.swift
martinmitrevski: discussed over Slack

nuno-vieira:
> XCMetrics are executed on every PR (including release) and on every push to develop

@testableapple do we really need to have it on every push to develop? We need to merge develop to the PR before merging it, so this is a bit redudant, I don't think we need to run this on every push to develop 🤔

fastlane/Fastfile
```diff
@@ -344,12 +345,94 @@ lane :build_test_app_and_frameworks do
...
+    develop_value = develop_performance[metric.to_s]
+    branch_value = branch_performance[metric]
+
+    status = if branch_value > develop_value && branch_value < develop_value * 1.5
+               '🟡' # Warning if new value is 50% slower than develop
+             elsif branch_value > develop_value
+               '🔴' # Failure if new value is more than 50% slower than develop
+             else
+               '🟢' # Success if new value is faster than or equal to develop
+             end
```
nuno-vieira, line 387: I'm not sure I'm following this 🤔 

What is the difference between this two?
- Warning if new value is 50% slower than develop
- Failure if new value is more than 50% slower than develop

Another thing is that 50% is way too much, something half slower is not a good 😄 We should maybe put 5% as warning, and 10% as a failure. This follows the default values of XCMeasure in Xcode.

fastlane/Fastfile
```diff
@@ -344,12 +345,94 @@ lane :build_test_app_and_frameworks do
...
+  branch_performance = xcmetrics_log_parser(log: action_log)
+
+  sh("git clone git@github.com:GetStream/stream-swift-performance-benchmarks.git #{File.dirname(performance_path)}")
+  previous_release_branch = "release/#{last_git_tag}"
+  performance_benchmarks = JSON.parse(File.read(performance_path))
+  develop_performance = performance_benchmarks['develop']
+  previous_release_performance = performance_benchmarks[previous_release_branch]
+
+  github_comment = "## Performance Metrics\n| `metric` | `#{current_branch}` | `#{previous_release_branch}` | `develop` | `status` |\n| - | - | - | - | - |\n"
+  [:startup_time, :message_list_load_time, :channel_list_load_time, :message_list_scroll_time, :channel_list_scroll_time].each do |metric|
```
nuno-vieira, line 376: Another nice to have would be to show the amount of "%" that it was better or worse 💡 Example:

| `metric` | `ci/xcmetrics` | `release/4.47.1` | `develop` | `status` | `diff` |
| - | - | - | - | - | - |
| Startup time | 1.4s | 1.5s | 42s | 🟢 | 2% ⬆️ |
| Message list load time | 2.7s | 2.2s | 2s | 🟡 | -6% ⬇️  |
| Channel list load time | 2.1s | 1s | 1s | 🔴 | -11% ⬇️  |
| Message list scroll time | 2.6s | 42s | 42s | 🟢 | 5% ⬆️ |
| Channel list scroll time | 2.2s | 42s | 42s | 🟢 | -1% ⬇️  |

fastlane/Fastfile
nuno-vieira: Yes if it is random, is not useful at all, so we should remove it 👍

fastlane/Fastfile
```diff
@@ -344,12 +345,102 @@ lane :build_test_app_and_frameworks do
...
+
+    status_emoji = if branch_value > develop_value && branch_value < develop_value / 0.95
+                     '🟡' # Warning if new value is less than 5% slower than develop
+                   elsif branch_value > develop_value
+                     '🔴' # Failure if new value is more than 5% slower than develop
+                   else
+                     '🟢' # Success if new value is faster than or equal to develop
+                   end
+
+    diff = ((branch_value - develop_value) / develop_value.to_f * 100).to_f.round(1)
```
nuno-vieira, line 390: I think this is the correct formula: 
```suggestion
    diff = ((develop_value - branch_value) * 100 / develop_value).to_f.round(1)
```

testableapple: Updated 👍

ipavlidakis (approved):

ipavlidakis (approved):

ipavlidakis (approved):

ipavlidakis (approved):

ipavlidakis (approved):

fastlane/Fastfile
```diff
@@ -345,6 +349,155 @@ lane :build_test_app_and_frameworks do
...
+
+      max_stddev = 0.1 # Default Xcode Max STDDEV is 10%
+      status_emoji =
+        if branch_value > release_value && branch_value < release_value + (release_value * max_stddev)
+          '🟡' # Warning if a branch is 10% less performant than the previous release
+        elsif branch_value > release_value
+          '🔴' # Failure if a branch is more than 10% less performant than the previous release
+        else
+          '🟢' # Success if a branch is more performant or equals to the previous release
+        end
```
testableapple, line 410: Take a closer look at this. Currently, there is no constant baseline. Technically, the previous release is a baseline for every new PR.

If you think we need a constant baseline, it should not be a problem to introduce one.

.github/workflows/xcmetrics.yml
```diff
@@ -0,0 +1,76 @@
...
+          ssh-private-key: ${{ secrets.BOT_SSH_PRIVATE_KEY }}
+
+      - name: Check for PR approvals
+        if: ${{ github.event_name == 'pull_request_review' && github.event.review.state == 'approved' }}
+        env:
+          GITHUB_TOKEN: '${{ secrets.CI_BOT_GITHUB_TOKEN }}'
+        run: |
+          current_branch=${{ github.event.pull_request.head.ref }}
+          echo "Current branch: $current_branch"
+          approvals_count=$(gh api "https://api.github.com/repos/GetStream/stream-chat-swift/pulls/${{ github.event.pull_request.number }}/reviews" | jq '[.[] | select(.state == "APPROVED")] | length')
```
laevandus, line 34: Thinking if it would be nicer if we pull out these URLs and have them as ENV variables or some sort of a constants. But on the other hand, does not really matter much since we do not expect these to change anyway.

martinmitrevski (commented):
Great work, this is amazing 👏  I have mostly questions around the report. Since I'm using the showcase as a reference, some of these questions might not be valid.
- What does duration mean? The duration of the test? If yes, why is it different on a branch?
- Frame count is the total across all the duration? If yes, I think we are not interested in this value - not sure if it gives something meaningful.
- Frame rate is definitely an important one. But the values (if correct) are a bit strange. We would need to know the max frame rate on the device to compare? E.g. on iPhones it's 60 fps, on iPad it's 120 fps. We probably need to display that (100 / 120 fps).
- Can you clarify what does performance percentage mean? Increase of the performance of the new branch that's being compared?

.github/workflows/xcmetrics.yml
testableapple: Yeah, don't see a reason to extract them into a separate env variable as they will only be used in this place. But we could extract the whole step into an independent action. It will be just a bit more readable 🤔

Will think about it 👍

testableapple: @laevandus, I extracted it into a separate action. Wdyt?

testableapple: The only thing is that to test this, we need to approve the PR 😄
Will test it then when everything else is reviewed/approved.

laevandus (dismissed):

.github/workflows/xcmetrics.yml
testableapple: @laevandus, I need another round of testing, could you approve one more time please?

martinmitrevski:
> Nah, it's the duration of the scroll, excluding hangs. The difference between tests should not be huge if any.
Ok, since you mentioned the release column was added by yourself, it's not an issue.
> Yeah, I believe it's just the amount of frames during scrolling. I agree that it's not that useful for us, just wanted to showcase everything that Apple offers so we can decide what to leverage. I can remove this metric if everyone agrees.
Yes, I would vote for removing - it doesn't bring any value.
> The branch column is the original one and 78.4 fps sounds reasonable.
78.4 feels weird - which device is that? It should be either around 60 fps or around 120 fps. How do you calculate this?

laevandus (approved):

.github/workflows/xcmetrics.yml
testableapple: Done ✅

nuno-vieira:
> @martinmitrevski,
> 
> > What does duration mean? The duration of the test? If yes, why is it different on a branch?
> 
> * Nah, it's the duration of the scroll, excluding hangs. The difference between tests should not be huge if any.
> 
> > Frame count is the total across all the duration? If yes, I think we are not interested in this value - not sure if it gives something meaningful.
> 
> * Yeah, I believe it's just the amount of frames during scrolling. I agree that it's not that useful for us, just wanted to showcase everything that Apple offers so we can decide what to leverage. I can remove this metric if everyone agrees.
> 
> > Frame rate is definitely an important one. But the values (if correct) are a bit strange. We would need to know the max frame rate on the device to compare? E.g. on iPhones it's 60 fps, on iPad it's 120 fps. We probably need to display that (100 / 120 fps).
> 
> * The data under the `release/` column is a sampler that I filled in manually a.k.a randomly and it has nothing to do with real data. Let me know if something else is required here. The `branch` column is the original one and `78.4 fps` sounds reasonable.
> 
> > Can you clarify what does performance percentage mean? Increase of the performance of the new branch that's being compared?
> 
> It is an increase (🔼) or a decrease (🔽) in app's performance. We discussed this internally with @nuno-vieira and, after playing with some different options, agreed that this should look less confusing. But please let me know your thoughts if it does not make sense or you have a better option 🙂

@martinmitrevski @testableapple Yes, the goal is to use the same thing as Xcode does. When there is a performance increase it says: "5% better", or when it is worse "-5% performance". And the baseline is usually 10%

TestTools/StreamChatTestMockServer/MockServer/ChannelResponses.swift
```diff
@@ -350,16 +357,19 @@ public extension StreamMockServer {
...
         parentId: String? = nil,
         channelId: String?,
         author: [String: Any]?,
-        replyCount: Int? = 0
+        replyCount: Int? = 0,
+        withAttachments: Bool = false,
+        overallMessagesCount: Int = 1
     ) -> [String : Any]? {
-        let timeInterval = TimeInterval(index * 1000 - 123_456_789)
+        let timeInterval = TimeInterval(index + channelIndex * 1000 - 123_456_789)
```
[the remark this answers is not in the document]
testableapple: Just a hardcoded timestamp to go back in time, so all messages will have this gap between each other.

.github/actions/pr-approvals/action.yml
```diff
@@ -0,0 +1,16 @@
...
+      shell: bash
+      run: |
+        if [ "${{ github.event.review.state }}" == "approved" ]; then
+          pr_approvals=$(gh api "${GITHUB_API_URL}/repos/${GITHUB_REPOSITORY}/pulls/${{ github.event.pull_request.number }}/reviews" | jq '[.[] | select(.state == "APPROVED")] | length')
+          echo "PR approvals count: $pr_approvals"
+          echo "pr_approvals=$pr_approvals" >> $GITHUB_ENV
+        else
+          echo "There are no PR approvals"
+          echo 'pr_approvals=0' >> $GITHUB_ENV
+        fi
```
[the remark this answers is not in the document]
testableapple: It will be run on every approval, but only the first approval will run xcmetrics. Not ideal, but I tried to limit the runs as much as possible. Let me know wdyt.

.github/workflows/xcmetrics.yml
```diff
@@ -0,0 +1,65 @@
...
+
+      - uses: ./.github/actions/pr-approvals
+        if: ${{ github.event_name == 'pull_request_review' }}
+
+      - name: Check if the PR was already measured
+        if: ${{ github.event_name == 'pull_request_review' }}
+        run: |
+          xcmetrics=$(gh api "${GITHUB_API_URL}/repos/GetStream/stream-swift-performance-benchmarks/contents/stream-chat-swift.json?ref=main")
+          already_measured=$(printf "%s" "$xcmetrics" | jq -r '.content' | base64 -d | jq "has(\"${{ github.event.pull_request.head.ref }}\")")
+          echo "already_measured=$already_measured" >> $GITHUB_ENV
```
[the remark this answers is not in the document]
martinmitrevski: Yes, we will never miss a regression - we still have the nightly checks, the manual option and the release checks. So once is enough imo.

.github/actions/pr-approvals/action.yml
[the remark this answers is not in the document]
martinmitrevski: Do we need to wait for the approvals though? Might slow down the flow a bit. My proposal is: 
- once when the PR is raised 
- if there are major changes in the PR - we should do manual invocation
- we have nightly checks
- we have release checks
Although, since we usually do manual QA after approval, maybe this won't have impact - so both options are ok imo.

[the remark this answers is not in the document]
nuno-vieira: @testableapple Yeah I agree with @martinmitrevski, probably it is better to just do it when the PR is open 🤔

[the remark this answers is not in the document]
testableapple: Updated 👍

So now the workflow will be triggered once:
- Nightly
- On any non-draft PR

nuno-vieira (approved):
LGTM!! ✅ 🚢 🎉

martinmitrevski (approved):
🚢

laevandus (approved):

PR #2979, opened 2024-01-23
Fix broken link in the UIKit docs

testableapple:
## Then

- https://getstream.io/chat/docs/sdk/ios/uikit/uikit/custom-components/#setuplayout

## Now

- https://getstream.io/chat/docs/sdk/ios/uikit/custom-components/#setuplayout

PS:
- Also updated links to apple dev


laevandus (approved):
Verified that updated links lead to correct pages.

martinmitrevski (approved):

Issue #1816, opened 2022-02-28
link message flickers when you open a ChatView


linhewinner:
https://user-images.githubusercontent.com/97132583/156061270-42d1d1b2-c8c5-4a18-b48f-b37054ebc2ce.MOV

See attached video.
It look as though the links were rendered as normal text initially for a split second before it is turned into a link

## GetStream Environment
**GetStream Chat version:** 4.10.0
**GetStream Chat frameworks:** StreamChat, StreamChatUI
**iOS version:**
**Swift version:**
**Xcode version:**
**Device:**

b-onc:
Hello @linhewinner 

Thanks for the report. We're looking into it and will get back to you.

adamrushy:
Hey, @linhewinner 

I tried reproducing this recently on the latest version of the SDK and on iOS 15.4 on both light/dark modes and is working as expected. Are you still able to reproduce this on the latest version? If so, can you provide any further details of your integration? Any custom UI etc...?

Thank you,
Adam

b-onc:
Issue closed due to inactivity. Please reopen if you can still reproduce!

nuno-vieira:
This issue is reproducible on our side, so we are re-opening it.

nuno-vieira:
Although this is reproducible, in the more recent versions it is much harder to reproduce and the issue is also very subtle. We for now lowering the priority of this one and focusing on more important things that we have on the roadmap. 

Hope you understand! 

Best,
Nuno

nuno-vieira:
Hi @linhewinner,

This fix will be available on 4.48.0.

Thank you for your patience!

Best,
Nuno

PR #2984, opened 2024-01-25
Add Composer Link Preview

nuno-vieira:
### 🔗 Issue Links
Resolves https://github.com/GetStream/ios-issues-tracking/issues/523
Related: https://github.com/GetStream/stream-chat-swift/issues/1816

### 🎯 Goal
Add composer link preview when the user is typing links.

### 📝 Summary
- Adds new `ChatClient.enrichUrl()` function to get metadata from URL
- Adds new `TextLinkDetector` for parsing and detecting links given a string
- Fixes message link preview showing empty space when no metadata is available 
- Adds `Components.default.isComposerLinkPreviewEnabled` flag to enable composer link previews
- Adds support for showing link previews in the composer
- Allow to change skipEnrichUrl when editing a message
- Only show url preview if `urlEnrichmentEnabled`
- Fixes link flickering when opening a channel

### 🛠 Implementation
- The `TextLinkDetector` was added to `StreamChat` so that it can be used in the SwiftUI SDK as well. It is just a easier wrapper of `NSDataDetector`. 
- The feature is disabled by default because enabling it can break the customer's Composer. Especially if a customer is upgrading from an old version to a newer one, they might not notice that this new component was added and ship the app without realising this was added.
- Since we currently do not support cancelling requests, a debouncer was added so that the `enrichUrl` request is not spammed, and we ensure only the last enrichment is used. But even in this scenario, if a failing enrich takes too much time, it could override a valid enrichment.
- The `ComposerLinkPreviewView` was added as part of the `ComposerView` and not above the `ComposerView` since the view should also work while editing a message. In this case, the preview should appear below the edit message header.

### 🎨 Showcase

https://github.com/GetStream/stream-chat-swift/assets/12814114/2040d6af-d2ae-4b77-95a6-6b7575fb90e4

### 🧪 Manual Testing Notes

⚠️  **Known backend Issues:**
- The backend should show always the first link of the message in the link preview, but right now, this is random. So the link preview of the composer might not match the one from the actual message.
- Typing links with capitalized "Www" sometimes fails the enrichment.
- When editing a message, the `skipEnrichUrl` won't have any effect on the already enriched links.

#### Showing link preview view

GIVEN I'm in the a channel
WHEN a user types a valid link in the composer
THEN the link should be highlighted
AND a view showing the link preview should appear

GIVEN I'm in the a channel
WHEN a user types a multiple valid links in the composer
THEN only the first link should be highlighted
AND a view showing the link preview should appear

GIVEN I'm in the a channel
AND a user types a valid link in the composer
AND the link preview is shown
WHEN the user sends the message
THEN the message should show the preview in the message cell

#### Closing the link preview view

GIVEN I'm in the a channel
AND a user types a valid link in the composer
AND the link preview is shown
WHEN the user taps the close button of the preview
AND the user sends the message
THEN the message does not show any preview in the message cell

GIVEN I'm in the a channel
AND a user types a valid link in the composer
AND the link preview is shown
WHEN the user taps the close button of the preview
AND the user changes the links in the composer
THEN the link preview should re-appear

#### Editing message (Not working yet 100% because of lack of Backend support)

GIVEN I'm in the a channel
WHEN a user edits a message that contains a link
THEN the link should be highlighted
AND a view showing the link preview should appear

GIVEN I'm in the a channel
AND a user edits a message that contains a link
WHEN the user closes the link preview
AND updates the message
THEN the messages should not include any preview

GIVEN I'm in the a channel
AND a user edits a message that contains a link but without preview
WHEN the user does not close the link preview
AND updates the message
THEN the messages should include the link preview

### ☑️ Contributor Checklist


Sources/StreamChat/ChatClient.swift
```diff
@@ -378,6 +378,21 @@ public class ChatClient {
         }
     }
 
+    /// Get the link attachment preview data from the provided url.
+    ///
+    /// This will return the data present in the OG Metadata.
+    public func enrichUrl(_ url: URL, completion: @escaping (Result<LinkAttachmentPayload, Error>) -> Void) {
```
martinmitrevski, line 384: I don't think this should be in the `ChatClient` - we mostly have `connect` / `disconnect` methods here. Maybe in the channel controller?

Sources/StreamChat/Utils/TextLinkDetector.swift
```diff
@@ -0,0 +1,57 @@
...
+        do {
+            detector = try NSDataDetector(types: types.rawValue)
+        } catch {
+            detector = nil
+            assertionFailure("Data detector failed to be created in \(self)")
+        }
+    }
+
+    /// Checks if the provided text contains links or not.
+    public func hasLinks(in text: String) -> Bool {
```
martinmitrevski, line 28: Since these are public methods, we should document them properly (add `Parameters` and `Returns`).

Sources/StreamChat/Utils/TextLinkDetector.swift
```diff
@@ -0,0 +1,57 @@
...
+        firstLink(in: text) != nil
+    }
+
+    /// Parses the first link of the provided text.
+    public func firstLink(in text: String) -> TextLink? {
+        let fullRange = NSRange(location: 0, length: text.utf16.count)
```
martinmitrevski, line 34: A short comment why we use `utf16.count` would be nice here.

Sources/StreamChatUI/CommonViews/InputTextView/InputTextView.swift
```diff
@@ -214,6 +235,23 @@ open class InputTextView: UITextView, AppearanceProvider {
         isScrollEnabled = true
     }
 
+    // MARK: - Link Detection
+
+    /// Resets the text attributes whenever the text changes.
+    open func resetAttributes() {
```
martinmitrevski, line 241: Why do we need to do this?

Sources/StreamChatUI/Composer/ComposerVC.swift
```diff
@@ -946,6 +970,49 @@ open class ComposerVC: _ViewController,
...
+    /// - Parameter links: The new parsed links from the input text.
+    open func didChangeLinks(_ links: [TextLink]) {
+        /// We only try to display the link preview of the first link.
+        guard let link = links.first else {
+            dismissLinkPreview()
+            return
+        }
+
+        enrichUrlDebouncer.execute { [weak self] in
+            self?.channelController?.client.enrichUrl(link.url) { [weak self] result in
```
martinmitrevski, line 983: This is also an argument to use the `channelController` for enriching urls, instead of the `client`.

martinmitrevski (commented):
Great work 👏  Few small things, and I think we should move the `enrichUrl` out of the client.

Sources/StreamChat/ChatClient.swift
nuno-vieira: This is to be aligned with the other SDKs, they also have `enrichUrl` in the `ChatClient`. That is why I opted for putting it in the `ChatClient`. I guess we can also move it to the `ChatChannelController`, but enriching a URL is not dependent on a Channel, so if we move the `enrichUrl` to the `CHatChannelController` it means the customer needs to create a controller or load a channel just to enrich a url

Sources/StreamChat/Utils/TextLinkDetector.swift
nuno-vieira: In this case they are self explanatory, that is why I didn't add them 🤔

Sources/StreamChatUI/CommonViews/InputTextView/InputTextView.swift
nuno-vieira: We are changing the color of the text, whenever we have a link in the input. When thera are new changes, the previous range of the text, will have the same blue color, if we do not reset it. I'll make sure I add a comment explaining why we need this 👍

Sources/StreamChatUI/Composer/ComposerVC.swift
nuno-vieira: Some as above 👍

Sources/StreamChat/ChatClient.swift
martinmitrevski: hmm, do all SDKs have it in their client or some specific one? In general, the preview is a composer feature, so enriching URLs by only being part of a channel is not an issue. However, I think the decision tree should be as follows - if more than 1 SDK has it in the client, then leave it as it is for consistency, otherwise move it to the channel controller. What do you think?

Sources/StreamChat/Utils/TextLinkDetector.swift
martinmitrevski: I know, but It's good for our DocC docs, I think they are still autogenerated on the Swift Package Index pages.

Sources/StreamChat/Utils/TextLinkDetector.swift
laevandus: There is also this NSRange init available: `NSRange(text.startIndex..<text.endIndex, in: text)`

Sources/StreamChat/ChatClient.swift
nuno-vieira: I can see that `uploadAttachment` is also part of `ChatClient` in other SDKs. All SDKs have this. But we have `uploadAttachment` in `ChannelController` as well. So maybe moving it to `ChannelController` makes it more consistent on our SDK, it is also just more straightforward to use I guess. But this will require some changes in tests etc...

Sources/StreamChat/Utils/TextLinkDetector.swift
nuno-vieira: I'll update this then 👍

Sources/StreamChatUI/CommonViews/InputTextView/InputTextView.swift
nuno-vieira: @laevandus Good point, the function is named `resetAttributes` because initially I was actually resetting all `textStorage` attributes, but that did not work out very well, so I changed it 😄 Either way, I'm going to test a different approach which I'm testing out and I'll as for a re-review then 👍 Will keep in mind your suggestions as wel 👍

nuno-vieira: This has now been simplified 👍

Sources/StreamChat/ChatClient.swift
nuno-vieira: @martinmitrevski Moved to the ChannelController 👍

Sources/StreamChatUI/Composer/ComposerVC.swift
```diff
@@ -106,7 +108,8 @@ open class ComposerVC: _ViewController,
             mentionedUsers: Set<ChatUser>,
             command: Command?,
             extraData: [String: RawJSON] = [:],
-            cooldownTime: Int = 0
+            cooldownTime: Int = 0,
+            skipEnrichUrl: Bool = false
```
martinmitrevski, line 112: Shouldn't this be controlled from the dashboard setting, coming in the channel config as `urlEnrichmentEnabled`?

martinmitrevski (commented):
Looks great 👌 Before merging:
- check my comment about controlling url enrichment from the dashboard
- let's fix the failing snapshot tests

Sources/StreamChat/Controllers/MessageController/MessageController.swift
```diff
@@ -243,20 +243,22 @@ public class ChatMessageController: DataController, DelegateCallable, DataStoreP
...
     ///   - text: The updated message text.
+    ///   - skipEnrichUrl: If true, skips url enriching.
     ///   - attachments: An array of the attachments for the message.
     ///   - extraData: Custom extra data. When `nil` is passed the message custom fields stay the same. Equals `nil` by default.
     ///   - completion: The completion. Will be called on a **callbackQueue** when the network request is finished.
     ///                 If request fails, the completion will be called with an error.
-    ///
     public func editMessage(
         text: String,
+        skipEnrichUrl: Bool = false,
```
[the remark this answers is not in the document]
martinmitrevski: About the second part - we should try to follow the REST API as closely as possible. It helps with consistency with the dashboard and other SDKs. 
In general, I think we don't even need a param here, it should all be controlled from the config on the dashboard.

Sources/StreamChatUI/Composer/ComposerVC.swift
nuno-vieira: No, the `skipEnrichUrl` is only used when `urlEnrichmentEnabled` is true. If you send `skipEnrichUrl = true`, and `urlEnrichmentEnabled` is `false`, it won't enrich any url.

Sources/StreamChat/Controllers/MessageController/MessageController.swift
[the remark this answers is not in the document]
nuno-vieira: I agree @laevandus, but we already have other params in the function, so it would be weird to have some "options" outside of the options structure. And since we can't do breaking changes, I think in this case it makes sense to be consistant and not introduce an `options` structure, especially because `createNewMessage` also has multiple params instead of the `options`approach. I also prefer the `options` struct, and we are using it already in newer APIs. But for this one, to keep consistent we probably should stick with regular params 👍

Sources/StreamChat/Controllers/MessageController/MessageController.swift
```diff
@@ -243,20 +243,22 @@ public class ChatMessageController: DataController, DelegateCallable, DataStoreP
     ///
     /// - Parameters:
     ///   - text: The updated message text.
+    ///   - skipEnrichUrl: If true, skips url enriching.
```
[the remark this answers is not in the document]
nuno-vieira: Yes, makes sense, I just copied this description from the `createNewMessage` but it is not clear at all

Sources/StreamChat/Utils/TextLinkDetector.swift
```diff
@@ -0,0 +1,69 @@
...
+        // Example: João
+        //    utf16.count == 4
+        //    utf8.count == 5
+        NSRange(location: 0, length: text.utf16.count)
+    }
+}
+
+private extension NSTextCheckingResult {
+    /// Maps a match to `TextLink`.
+    func toTextLink(with text: String) -> TextLink? {
```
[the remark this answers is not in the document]
nuno-vieira: Makes sense 👍

Sources/StreamChatUI/CommonViews/QuotedChatMessageView/QuotedChatMessageView.swift
```diff
@@ -184,6 +187,26 @@ open class QuotedChatMessageView: _View, ThemeProvider, SwiftUIRepresentable {
...
+        let attributedText = NSMutableAttributedString(
+            string: text,
+            attributes: [
+                .foregroundColor: appearance.colorPalette.text,
+                .font: appearance.fonts.subheadline
+            ]
+        )
+
+        linkDetector.links(in: text).forEach { textLink in
+            attributedText.addAttribute(.link, value: textLink.url.absoluteString, range: textLink.range)
```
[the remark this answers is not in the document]
nuno-vieira: Did not notice that, nice catch 👍

Sources/StreamChatUI/CommonViews/InputTextView/InputTextView.swift
```diff
@@ -214,6 +237,16 @@ open class InputTextView: UITextView, AppearanceProvider {
         isScrollEnabled = true
     }
 
+    // MARK: - Link Detection
+
+    /// Highlights the links in the input text view.
+    open func highlightLinks(_ links: [TextLink]) {
+        links.forEach { link in
+            let linkColor = appearance.colorPalette.textLinkColor
+            textStorage.addAttribute(.foregroundColor, value: linkColor, range: link.range)
```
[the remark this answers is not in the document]
nuno-vieira: In this case it is not, because we don't want the link to be interactable. And with `link` it will be interactable. I will add a comment explaining this 👍

Sources/StreamChatUI/Composer/ComposerVC.swift
nuno-vieira: We only use it to decide if we should show the link preview or not. If `urlEnrichmentEnabled` is `false` we never show the link preview in the composer. Since when sending a message, the url won't be enriched

martinmitrevski (approved):

laevandus (approved):

testableapple (approved):
Looks great to me 🚢

Issue #2568, opened 2023-04-13
channelListController.synchronize multiple completion callbacks


AndrewSB:
## What did you do?

i'm trying to wrap `ChatChannelListController.synchronize` in a continuation, so i can make it async

```swift
        return await withCheckedContinuation { continuation in
            channelListController.synchronize { error in
                if let error = error {
                    CrashReporter.logNonFatal(error)
                } else {
                    self.channels = Array(channelListController.channels).filter { !$0.isGroupChat }
                }
                continuation.resume(returning: self.channels)
            }
        }
```

## What did you expect to happen?

i expected the callback to be called once, and only once, with either the channels or with an error 

## What happened instead?

when there's an error, the callback is called with an error first, then again after retry. this causes a swift runtime crash, 
`Fatal error: SWIFT TASK CONTINUATION MISUSE: fetchServerActiveChatChannels(currentUserId:) tried to resume its continuation more than once`

it looks like your `APIClient.operation` internally retries up to 3 times, calling completion upto 3 times. i'd expect not to receive an error in the completion of `synchronize(_ completion:)` if you're retrying.

thoughts?

testableapple:
Hi @AndrewSB,
Thanks for reporting the issue. 

We are going to investigate this on our side and will get back to you as soon as possible.
In the meantime, you may try to erase the continuation once it is called (e.g.: https://forums.swift.org/t/how-to-prevent-swift-task-continuation-misuse/57581/2). Hope this will unblock you while we are investigating this.

Best,
Alexey

testableapple:
@AndrewSB, could you share the SDK version, please? We fixed a couple of issues around this lately, so if you are not on the latest version, would be nice to try it out.

Best,
Alexey

AndrewSB:
the bug seems really straightforward to me, your networking code calls a completion handler multiple times, one for each retry, if theres a network error.

if thats expected, it shouldn't be called a "completion" handler and satisfy the contract in some other way, i'd recommend just not calling the handler for any errors until the retries have been spent

polqf:
Hi @AndrewSB, 
If you have a close look at the implementation, you would see how we are not calling the completion block when we are retrying.
Would be nice to understand the whole flow you execute for that to happen.

Is that consistently happening? Please do share the integration code, as I don't really know where this execution falls based on the snippet you shared some days ago ([this one](https://gist.github.com/AndrewSB/d1062b1912e6ea7d1db0147fd40a7dc7))

[...]

PR #2985, opened 2024-01-29
Remove connectionIDWaiters property with internal synchronization sin…

laevandus:
…ce it can cause threading issues when reading and mutating the dictionary

### 🔗 Issue Links

Resolves [GetStream/ios-issues-tracking/709](https://github.com/GetStream/ios-issues-tracking/issues/709)

Related [GetStream/stream-chat-swift/2568](https://github.com/GetStream/stream-chat-swift/issues/2568)

### 🎯 Goal

Clean up unused code and remove a way of causing threading issues with mutating `_connectionIdWaiters` through the separate property

### 📝 Summary

Removing a way for accidentally creating a timing issue which can cause hard to find issues.

### 🛠 Implementation

We have had issues in the past where we create a property which wraps another property with a dispatch queue. If the value is, for example, dictionary, then code like `connectionIdWaiters["key"] = value` causes timing issues since the subscript first triggers the getter on the queue, then modifies the value, and then uses the setter on the queue again. If someone else reads or writes between that get and set cycle, we start dropping changes to the dictionary (e.g. thread B reads the state before thread A called set with a new key, thread B happily commits its updates and thread A's change gets lost). 

### ☑️ Contributor Checklist

### 🎁 Meme

### DIFF
--- a/Sources/StreamChat/Repositories/ConnectionRepository.swift
+++ b/Sources/StreamChat/Repositories/ConnectionRepository.swift
@@ -10,11 +10,6 @@ class ConnectionRepository {
     private var _connectionId: ConnectionId?
     private var _connectionStatus: ConnectionStatus = .initialized
 
-    private(set) var connectionIdWaiters: [String: (Result<ConnectionId, Error>) -> Void] {
-        get { connectionQueue.sync { _connectionIdWaiters } }
-        set { connectionQueue.async(flags: .barrier) { self._connectionIdWaiters = newValue }}
-    }
-
     /// The current connection status of the client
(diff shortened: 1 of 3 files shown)


Sources/StreamChat/Repositories/ConnectionRepository.swift
```diff
@@ -180,7 +175,9 @@ class ConnectionRepository {
         }
 
         let waiterToken = String.newUniqueId
-        connectionIdWaiters[waiterToken] = completion
```
laevandus, line 183: The change I did in the unit-test below captured the issue on this line. Unit-test would fail and says that many of the completion blocks were never called.

PR #2996, opened 2024-02-01
Reconfigure cells in the channel list when handling StagedChangeset.elementUpdated

laevandus:
### 🔗 Issue Links

Resolves [ios-issues-tracking/716](https://github.com/GetStream/ios-issues-tracking/issues/716)

### 🎯 Goal

Increase list performance and reduce UI glitches

### 📝 Summary

- Avatars flash when opening the channel list

### 🛠 Implementation

Avatars flash in the channel list when we apply `StagedChangeset.elementUpdated` changes. I observed that we call`reload(using:interrupt:setData:)` (applying data changes to the view) on the channel view 5 times after opening it (demo app) where one of the reload call contains 20 (in my case) `elementUpdated` changes. These, currently trigger cell reloads which cause the UI glitch (avatars flashing) and also reduces the performance since all the cells need to be created (it was the 4th reload call)

The proposed fix is to use Apple provided `UICollectionView.reconfigureItems(at:)` and `UITableView.reconfigureRows(at:)`. Since the channel list is a collection view and we only use a single cell type, then we can turn on cell reusing there. Message list is a table view and has more sophisticated reuse identifier handling, so currently do not use reconfiguring there (needs a separate PR since it is more complex).

I also did some micro-measurements with time profiler and iPhone 12 Pro Max.

#### Precondition
R2-D2 account is logged in.

#### Steps

1. Build the app for profiling
2. Launch the app on a device with time profiler running
3. Wait a little bit when things settle down and check what was the overall main thread usage without diving into details.

#### Results

5 consecutive runs (with 3 warmup runs before)

| Run | reloadItems (ms)  | reconfigureItems (ms) |
| --- | ------------- | ------------- |
| 1 | 484  | 381  |
| 2 | 479  | 398  |
| 3 | 499  | 387  |
| 4 | 491  | 386  |
| 5 | 494  | 393  |
| Avg | 489  | 389 |

Around 20% less CPU time on the main thread.

### 🎨 Showcase

| Before | After |
| ------ | ----- |
|  |  |

### 🧪 Manual Testing Notes

We need to make sure that reconfiguring items does not have any side-effects. We might have relied to the fact that cells are reloaded.

### ☑️ Contributor Checklist

### 🎁 Meme

gif

### DIFF
--- a/Scripts/updateDependency.sh
+++ b/Scripts/updateDependency.sh
@@ -67,3 +67,8 @@ do
 done
 
 rm -rf $dependency_directory
+
+if [[ $dependency_directory == *"DifferenceKit"* ]]; then
+    # We currently use customized UIKit extensions in Utils/DifferenceKit+Stream.swift
+    rm $output_directory/Extensions/UIKitExtension.swift
+fi
--- a/Sources/StreamChatUI/ChatMessageList/ChatMessageListView+DiffKit.swift
+++ b/Sources/StreamChatUI/ChatMessageList/ChatMessageListView+DiffKit.swift
@@ -22,7 +22,8 @@ extension ChatMessageListView {
         CATransaction.setCompletionBlock(completion)
         reload(
             using: changeset,
-            with: animation()
+            with: animation(),
+            reconfigure: { _ in false }
         ) { [weak self] newMessages in
             self?.onNewDataSource?(newMessages)
         }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Fix link flickering when opening a channel [#2984](https://github.com/GetStream/stream-chat-swift/pull/2984)
 - Fix link flickering when quoting a message with a link [#2984](https://github.com/GetStream/stream-chat-swift/pull/2984)
 - Fix channel item actions gesture overriding native swipe go-back gesture [#3000](https://github.com/GetStream/stream-chat-swift/pull/3000)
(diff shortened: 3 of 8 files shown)


Scripts/updateDependency.sh
```diff
@@ -67,3 +67,8 @@ do
 done
 
 rm -rf $dependency_directory
+s
+if [[ $dependency_directory == *"DifferenceKit"* ]]; then
+    # We currently use customized UIKit extensions in Utils/DifferenceKit+Stream.swift
+    rm $output_directory/Extensions/UIKitExtension.swift
+fi
```
laevandus, line 74: I added this for making sure we keep using the customised extension which is a copy of this file. Can be cleaned up if we manage to get the change into `DifferenceKit` itself

nuno-vieira (commented):
LGTM 👍 We are just missing the Changelog and then 🚢  (We can mention that flashing avatars in the channel list were fixed)

laevandus:
Seems like some tests are not happy. I will keep it in draft until I have investigated it.

Sources/StreamChatUI/ChatMessageList/ChatMessageListView+DiffKit.swift
```diff
@@ -22,7 +22,8 @@ extension ChatMessageListView {
         CATransaction.setCompletionBlock(completion)
         reload(
             using: changeset,
-            with: animation()
+            with: animation(),
+            reconfigure: { _ in false }
```
laevandus, line 26: This is a table view and will need more work to efficiently using reconfigure (we have some reuse identifier logic for smart reusing so this needs to take this into an account). Needs to be a separate PR.

nuno-vieira (approved):
LGTM! ✅

PR #3000, opened 2024-02-02
Fix channel item actions gesture overriding native swipe go-back gesture

pm-dev:
### 🔗 Issue Links

Resolves https://github.com/GetStream/ios-issues-tracking/issues/645

### 🎯 Goal

When showing the ChannelListVC in a UINavigationController's stack, I noticed the swipe-to-go-back interaction wasn't working correctly. Upon investigation, I found the SwipeableView steals the gesture recognition from the navigation controller's interactivePoopGestureRecognizer

### 📝 Summary

I've changed this view to ignore horizontal swipe gesture's when the action items are not expanded (i.e. view is not "open") _and_ the gesture is left-to-right. This SDK doesn't currently support action items on the left side, so this change should be safe.

### 🛠 Implementation

- I've also made the delegate functions open. Making them public doesn't make sense because panGestureRecognizer is public and subclass or client of this view can override the delegate themselves.
- An alternative to this solution is to leave the delegate functions open and require subclassing to get this behavior, but I believe this behavior should be the default.
- I believe this logic is valid even for right to left languages, but might want to check me on that.

### 🧪 Manual Testing Notes
1. Swipe left the channel item
2. The actions should appear at the right
3. Try closing the actions by swiping to the right, close to the left edge, like you would navigate back
4. It should not move the channel list item view

(In the Demo App right now we don't have a way to swipe back, so this is the best way to test it.)

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChatUI/ChatChannelList/SwipeableView.swift
+++ b/Sources/StreamChatUI/ChatChannelList/SwipeableView.swift
@@ -157,13 +157,16 @@ open class SwipeableView: _View, ComponentsProvider, UIGestureRecognizerDelegate
             return false
         }
 
-        // Additionally, if the horizontal swipe is left-to-right and action items
-        // are not expanded, we'll deny this recognizer, so that it doesn't steal
-        // the gesture from another recognizer, such as a UINavigationController's
-        // interactivePopGestureRecognizer
-        if !isOpen, translation.x > 0 {
+        // Additionally, if the gesture recognizer started from the left edge of the screen,
+        // ignore this recognizer so that it doesn't steal the gesture from UINavigationController's
+        // interactivePopGestureRecognizer.
+        let edgeThreshold: CGFloat = 30
+        let location = recognizer.location(in: self)
+        if location.x < edgeThreshold {
             return false
         }
+
+
         return true
     }
 
--- a/Sources/StreamChatUI/ChatChannelList/SwipeableView.swift
+++ b/Sources/StreamChatUI/ChatChannelList/SwipeableView.swift
@@ -142,7 +142,7 @@ open class SwipeableView: _View, ComponentsProvider, UIGestureRecognizerDelegate
         }
     }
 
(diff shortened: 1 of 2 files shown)


nuno-vieira (dismissed):
LGTM ✅ We just need to update the changelog to the correct section 👍

Sources/StreamChatUI/ChatChannelList/SwipeableView.swift
```diff
@@ -153,11 +153,21 @@ open class SwipeableView: _View, ComponentsProvider, UIGestureRecognizerDelegate
...
+        }
 
-        return abs(translation.x) > abs(translation.y)
+        // Additionally, if the horizontal swipe is left-to-right and action items
+        // are not expanded, we'll deny this recognizer, so that it doesn't steal
+        // the gesture from another recognizer, such as a UINavigationController's
+        // interactivePopGestureRecognizer
+        if !isOpen, translation.x > 0 {
+            return false
+        }
```
nuno-vieira, line 166: In order to support swipe-left-to-right actions, we can actually define a threshold for the origin of the swipe 🤔 If the swipe is too close to the edge, we return false, but if it is not, we don't ignore it. (This is how I believe for example how the Mail app from Apple works)

The thing is that, we (The SDK) by default do not support swipe-left-to-right actions, but another customer might i've added swipe-left-to-right actions, and this would prevent it to work. 

If you do not have the time to add this improvement, I can take it from here ofc 🙂 This already very helpful, Thank you! 🙏

pm-dev: Thanks @nuno-vieira! I wasn't considering the fact that a client may have added left-to-right actions.
I'll let you take it from here. We've implemented a work-around, so prioritize this as you wish.

martinmitrevski (approved):
LGTM! ✅

nuno-vieira (approved):

pm-dev:
🎉

PR #3005, opened 2024-02-06
Do not call Cached.computeValue from the mutate function

laevandus:
### 🔗 Issue Links

Resolves [ios-issues-tracking/718](https://github.com/GetStream/ios-issues-tracking/issues/718)

### 🎯 Goal

Fixing a crash on logout

### 📝 Summary

Do not call computeValue from the mutate since, in rare cases it can trigger reentrance which leads to a crash due to exclusive access requirement in the Swift programming language (note that mutate takes inout argument).

### 🛠 Implementation

Created a unit-test to simulate the crash we were seeing. Applied a fix by moving computeValue outside of the mutate. This is acceptable because `@Cached` is thread-safe but not concurrency safe.

Note: CoreDataLazy has pretty similar implementation but this PR is not going to change that.

### 🧪 Manual Testing Notes

We should try login and logout in the demo app. It should never crash.

### ☑️ Contributor Checklist


martinmitrevski (commented):
Let's discuss this one first on the tech discussion, please don't merge it for the next release 🙏

laevandus:
I'll keep this in draft since I will revisit the implementation to see if we could make a fix one level higher.

PR #3008, opened 2024-02-06
Fix attachments being sent with local url paths

nuno-vieira:
### 🔗 Issue Links
Resolves https://github.com/GetStream/ios-issues-tracking/issues/715

### 🎯 Goal
Fixes attachments being uploaded with local URL paths.

### 📝 Summary
**Main improvements:**
- Fixes `MessageSender` sending messages that did not have all attachments uploaded
- Fixes wrong construction of attachment's sandbox URL, which would cause URLs from different sessions to fail
- Automatically restart attachments that were in progress when the app was killed

**Bonus:**
- Fixes rare crash in `AttachmentQueueUploader.pendingAttachmentIDs`
- Protection against sending messages to the server with local URL paths
    - This is just an additional protection, since after the fixes this would not be needed, but I think is good we make sure we do not allow this to happen in any scenario. Especially since our solution at the moment is a bit error-prone, because we use the same url properties in the attachment payloads for both local and remote urls. 

####  `Issue 1: allAttachmentsAreUploadedOrEmptyPredicate()`
**Problem**: This predicate was the root cause of the original issue. It was returning attachments that were still in progress and not only the ones that were fully uploaded. This predicate is used to resend messages when the SDK is initialised and there are pending messages to be sent to the server. So this caused messages with local url paths to be sent to the server.

**Solution:** In theory, the old predicate was correct, but after some investigation, it seems that there is a bug in CoreData, and the old predicate does not work in 100% of cases. The SUBQUERY approach works flawlessly as it is shown by the tests. 

#### `Issue 2: Automatically restart attachments that were in progress when the app is killed`
**Problem:** After fixing issue 1, there was still a problem. The message would be in `.sending` state and with attachments stuck in progress. Since we do not allow to do any action in `.sending`state messages, we could not manually restart the attachments, and so the message would be forever in the sending state. 

**Solution:** In order to fix this problem, when the SDK is initialised and we resend the messages that are in a pending state, we also restart the attachments that are in progress.

#### `Issue 3: Wrong sandbox URL`
**Problem:** After fixing Issue 1 and Issue 2, there was still one more problem. 😅 The attachment local file paths were invalid after killing and opening the app again, so restarting the attachments would always fail, marking the message as failed to send.

**Solution:** At first, I thought about saving the files as binary data in CoreData but then after some deeper investigation, I found that the `storeAttachment()` function had an issue, and the sandboxed url was not properly constructed. This issue was actually introduced recently when we fixed attachments not having unique names. The problem was that we're using the `fileName` as part of the attachment's final name. The `fileName` was picked up from the `temporaryURL`, this works fine the first time, but when relaunching the app, the `temporaryURL` is already the sandbox URL, and so, this would cause the name of the final file to be incorrect, and it would say that the file did not exist, although it was there indeed.

### 🧪 Manual Testing Notes
1. Open a channel
2. Send a message with multiple videos (Easier to test with longer videos)
3. Kill the app while the attachments are uploading
4. Open the app
5. All attachments which were suspended, should be uploaded and the message published to the server with the remote urls.

### ☑️ Contributor Checklist


Sources/StreamChat/Database/DTOs/AttachmentDTO.swift
```diff
@@ -8,9 +8,12 @@ import Foundation
...
     /// An attachment id.
-    @NSManaged private var id: String
+    @NSManaged private var id: String?
     var attachmentID: AttachmentId? {
-        get { AttachmentId(rawValue: id) }
+        get {
+            guard let id = self.id else { return nil }
+            return AttachmentId(rawValue: id)
+        }
         set { id = newValue?.rawValue ?? "" }
```
ipavlidakis, line 17: why not simply?
```suggestion
        set { id = newValue?.rawValue }
```

Sources/StreamChat/Database/DTOs/AttachmentDTO.swift
```diff
@@ -79,6 +82,13 @@ class AttachmentDTO: NSManagedObject {
         request.predicate = NSPredicate(format: "localStateRaw == %@", LocalAttachmentState.pendingUpload.rawValue)
         return request
     }
+    
+    static func loadInProgressAttachments(context: NSManagedObjectContext) -> [AttachmentDTO] {
+        let request = NSFetchRequest<AttachmentDTO>(entityName: AttachmentDTO.entityName)
+        request.sortDescriptors = [NSSortDescriptor(keyPath: \AttachmentDTO.id, ascending: true)]
+        request.predicate = NSPredicate(format: "localStateRaw == %@", LocalAttachmentState.uploading(progress: 0).rawValue)
```
ipavlidakis, line 89: Is there any chance we may get attachments with uploading progress different to 0 and less than 100?

Sources/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -158,11 +158,11 @@ class MessageDTO: NSManagedObject {
...
     }
 
-    private static func allAttachmentsAreUploadedOrEmptyPredicate() -> NSCompoundPredicate {
-        NSCompoundPredicate(orPredicateWithSubpredicates: [
-            .init(format: "NOT (ANY attachments.localStateRaw != %@)", LocalAttachmentState.uploaded.rawValue),
-            .init(format: "attachments.@count == 0")
-        ])
+    static func allAttachmentsAreUploadedOrEmptyPredicate() -> NSCompoundPredicate {
+        .init(
+            format: "SUBQUERY(attachments, $a, $a.localStateRaw == %@).@count == attachments.@count",
```
ipavlidakis, line 163: Awesome! (Maybe it's helpful to start thinking adding FetchIndexes for fields like this one)

ipavlidakis (approved):

Sources/StreamChat/Database/DTOs/AttachmentDTO.swift
nuno-vieira: Forgot to update this from old code 👍

Sources/StreamChat/Database/DTOs/AttachmentDTO.swift
nuno-vieira: No, the rawValue is just "uploading", the progress is not accounted for the rawValue 👍

Sources/StreamChat/Database/DTOs/MessageDTO.swift
nuno-vieira: I think in this case this will be a small and simple query, but if we see that adding fetch indexes would be benificial down the road we can add them 🙂

laevandus (approved):

martinmitrevski (approved):
Amazing work 🥇

testableapple (approved):
LGTM 🚢

PR #3010, opened 2024-02-07
Fix `ChatMessageListVC` swipe-to-reply pan gesture recognizer overriding native swipe go-back gesture

pm-dev:
### 🔗 Issue Links

N/A

### 🎯 Goal

Similar to https://github.com/GetStream/stream-chat-swift/pull/3000, the `ChatMessageListVC` is disabling the native swipe to go back functionality.

### 📝 Summary

The `ChatMessageListVC` uses a pan gesture recognizer for its swipe-to-reply functionality, but this pan gesture recognizer disables the navigation controller's interactive pop functionality.

### 🛠 Implementation

This PR prevents the pan gesture recognizer from being added to its view if `components.messageSwipeToReplyEnabled` is turned off. An alternative implementation would have made the pan gesture recognizer a stored property on the VC and still added it to the view, so that clients could control it's `enabled` property.

We've implemented a workaround for the time being, which is to subclass ChatMessageListVC and implement `UIGestureRecognizerDelegate`'s `shouldRequireFailureOf` function, so that the pan gesture recognizer waits for the native interactive pop gesture to fail.

### ☑️ Contributor Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,10 +17,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### ✅ Added
 - Add `Components.default.isComposerLinkPreviewEnabled` flag to enable composer link previews [#2984](https://github.com/GetStream/stream-chat-swift/pull/2984)
 - Add support for showing link previews in the composer [#2984](https://github.com/GetStream/stream-chat-swift/pull/2984)
-- Fix issue where an interactive pop gesture from a navigation controller would not work from the channel list [#3000](https://github.com/GetStream/stream-chat-swift/pull/3000)
 ### 🐞 Fixed
 - Fix link flickering when opening a channel [#2984](https://github.com/GetStream/stream-chat-swift/pull/2984)
 - Fix link flickering when quoting a message with a link [#2984](https://github.com/GetStream/stream-chat-swift/pull/2984)
+- Fix channel item actions gesture overriding native swipe go-back gesture [#3000](https://github.com/GetStream/stream-chat-swift/pull/3000)
 
 # [4.47.1](https://github.com/GetStream/stream-chat-swift/releases/tag/4.47.1)
 _January 24, 2024_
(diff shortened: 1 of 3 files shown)


nuno-vieira:
Hi @pm-dev, is this maybe an issue on your app only? At least with our Demo App, I can easily swipe back, with both the swipe to reply enabled and disabled 🤔

pm-dev:
@nuno-vieira 
The native swipe-to-go-back functionality works out of the box in the demo app because it uses the default `backBarButtonItem`. However if anyone wants a custom back button icon, they'll need to set `leftBarButtonItem = UIBarButtonItem(...)`. When a vc has its `navigationItem.leftBarButtonItem` set, this disables the interactive pop gesture for that view controller, and the only way to re-enable it, is to implement the navigation controller's interactive pop gesture recognizer's delegate yourself.

Most production apps will not use Apple's back icon, meaning it's likely they're already explicitly setting `interactivePopGestureRecognizer.delegate` and implementing `gestureRecognizerShouldBegin`. However to make this work with `ChatChannelVC` they'll also need to implement `gestureRecognizer(_:shouldBeRequiredToFailBy:)` and return true. We've done this, and now everything works as expected.

Perhaps this is the correct fix, but I would argue it's also unnecessary and asking for problems to add an enabled gesture recognizer to a view when it's not being used. At the very least I would expect it's enabled/disabled to be based on `components.messageSwipeToReplyEnabled`

To see the swipe-to-go-back behavior of native vs custom back buttons:
[Tester.zip](https://github.com/GetStream/stream-chat-swift/files/14199653/Tester.zip)

nuno-vieira:
> @nuno-vieira The native swipe-to-go-back functionality works out of the box in the demo app because it uses the default `backBarButtonItem`. However if anyone wants a custom back button icon, they'll need to set `leftBarButtonItem = UIBarButtonItem(...)`. When a vc has its `navigationItem.leftBarButtonItem` set, this disables the interactive pop gesture for that view controller, and the only way to re-enable it, is to implement the navigation controller's interactive pop gesture recognizer's delegate yourself.
> 
> Most production apps will not use Apple's back icon, meaning it's likely they're already explicitly setting `interactivePopGestureRecognizer.delegate` and implementing `gestureRecognizerShouldBegin`. However to make this work with `ChatChannelVC` they'll also need to implement `gestureRecognizer(_:shouldBeRequiredToFailBy:)` and return true. We've done this, and now everything works as expected.
> 
> Perhaps this is the correct fix, but I would argue it's also unnecessary and asking for problems to add an enabled gesture recognizer to a view when it's not being used. At the very least I would expect it's enabled/disabled to be based on `components.messageSwipeToReplyEnabled`
> 
> To see the swipe-to-go-back behavior of native vs custom back buttons: [Tester.zip](https://github.com/GetStream/stream-chat-swift/files/14199653/Tester.zip)

Thanks for the explanation 👍 We will improve this and take it from here as well. Thank you once again 🙏

nuno-vieira:
Closing this one in favor of this: https://github.com/GetStream/stream-chat-swift/pull/3029

PR #3014, opened 2024-02-07
Do not check token expiration client side

nuno-vieira:
### 🔗 Issue Links
Resolves https://github.com/GetStream/ios-issues-tracking/issues/719

### 🎯 Goal

Do not check token expiration client side. This is a bad practice since the user can change the device's time. 

### 📝 Summary
- Stops checking expiration client side
- Fixes duplicated refresh tokens
- Fixes token refreshing not working when done server-side
- Adds the possibility to set refresh tokens in the Demo App

### 🛠 Implementation
The main reason why server side refresh token was not working properly it was because `isGettingToken = false` should be set on `tokenReceived` as well, and not only when the waiting completion blocks are executed. The refresh token wouldn't be executed because `isGettingToken` was `true`, so it would ignore the refresh.

### 🧪 Manual Testing Notes
1. Open Demo App Configuration
2. Tap on "tokenRefreshDetails"
3. Input the app secret from the dashboard and the desired duration
4. Connect a user
5. The app should work normally. The console log will output this whenever a token is refreshed:
    `> Demo App Token Refreshing: New token generated.`

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/Repositories/AuthenticationRepository.swift
+++ b/Sources/StreamChat/Repositories/AuthenticationRepository.swift
@@ -338,6 +338,7 @@ class AuthenticationRepository {
         }
 
         let onTokenReceived: (Token) -> Void = { [weak self, weak connectionRepository] token in
+            self?.isGettingToken = false
             self?.prepareEnvironment(userInfo: userInfo, newToken: token)
             // We manually change the `connectionStatus` for passive client
             // to `disconnected` when environment was prepared correctly
--- a/Sources/StreamChat/Repositories/AuthenticationRepository.swift
+++ b/Sources/StreamChat/Repositories/AuthenticationRepository.swift
@@ -363,13 +363,11 @@ class AuthenticationRepository {
         log.debug("Requesting a new token", subsystems: .authentication)
         tokenProvider { [weak self] result in
             switch result {
-            case let .success(newToken) where !newToken.isExpired:
+            case let .success(newToken):
                 onTokenReceived(newToken)
                 self?.tokenQueue.sync(flags: .barrier) {
                     self?._tokenExpirationRetryStrategy.resetConsecutiveFailures()
                 }
-            case .success:
-                retryFetchIfPossible(nil)
             case let .failure(error):
                 log.info("Failed fetching token with error: \(error)")
                 retryFetchIfPossible(error)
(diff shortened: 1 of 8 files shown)


Sources/StreamChat/Repositories/AuthenticationRepository.swift
```diff
@@ -365,11 +365,10 @@ class AuthenticationRepository {
             switch result {
             case let .success(newToken) where !newToken.isExpired:
```
martinmitrevski, line 366: why do we still keep the check? This can also have opposite effect - if you put the time forward.

Sources/StreamChat/Repositories/ConnectionRepository.swift
```diff
@@ -141,8 +141,7 @@ class ConnectionRepository {
             shouldNotifyConnectionIdWaiters = true
             connectionId = id
 
-        case let .disconnecting(source) where source.serverError?.isInvalidTokenError == true,
-             let .disconnected(source) where source.serverError?.isInvalidTokenError == true:
+        case let .disconnected(source) where source.serverError?.isInvalidTokenError == true:
```
martinmitrevski, line 144: why are we removing the `disconnecting` case? Too many token provider calls?

Sources/StreamChat/Repositories/AuthenticationRepository.swift
```diff
@@ -200,7 +200,7 @@ class AuthenticationRepository {
             return
         }
 
-        scheduleTokenFetch(isRetry: false, userInfo: nil, tokenProvider: tokenProvider, completion: completion)
+        scheduleTokenFetch(isRetry: true, userInfo: nil, tokenProvider: tokenProvider, completion: completion)
```
martinmitrevski, line 203: Would be nice to provide more details on the PR, to understand how you came up with the fix 🙏

Sources/StreamChat/Repositories/ConnectionRepository.swift
nuno-vieira: Yes, we are only interested when the websocket finishes disconnecting. The way it was before it would trigger two refresh calls always, which does not make sense

martinmitrevski (approved):
Great job! Really like how you solved the testing part, useful for us, but still frictionless for anyone trying out the demo app 👏  Codewise, looks good, but extra testing would be needed by @testableapple.

testableapple (approved):
LGTM 🚀

PR #3016, opened 2024-02-08
Make reconfigure closure optional when reloading with staged changesets

laevandus:
### 🔗 Issue Links

Related to [PR:2996](https://github.com/GetStream/stream-chat-swift/pull/2996)

### 🎯 Goal

Make call sites nicer when we actually do not want to use reconfigure support (example: message list). Also optional closure allows skipping the partition call if we want to skip reconfigure.
This is a cleanup to the existing PR.

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChatUI/ChatMessageList/ChatMessageListView+DiffKit.swift
+++ b/Sources/StreamChatUI/ChatMessageList/ChatMessageListView+DiffKit.swift
@@ -22,8 +22,7 @@ extension ChatMessageListView {
         CATransaction.setCompletionBlock(completion)
         reload(
             using: changeset,
-            with: animation(),
-            reconfigure: { _ in false }
+            with: animation()
         ) { [weak self] newMessages in
             self?.onNewDataSource?(newMessages)
         }
--- a/Sources/StreamChatUI/Utils/Extensions/DifferenceKit+Stream.swift
+++ b/Sources/StreamChatUI/Utils/Extensions/DifferenceKit+Stream.swift
@@ -29,7 +29,7 @@ extension UITableView {
     func reload<C>(
         using stagedChangeset: StagedChangeset<C>,
         with animation: @autoclosure () -> RowAnimation,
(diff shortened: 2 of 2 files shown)


Sources/StreamChatUI/ChatMessageList/ChatMessageListView+DiffKit.swift
```diff
@@ -22,8 +22,7 @@ extension ChatMessageListView {
         CATransaction.setCompletionBlock(completion)
         reload(
             using: changeset,
-            with: animation(),
-            reconfigure: { _ in false }
```
nuno-vieira, line 26: Btw, why don't we need this for the Message List ? 🤔

laevandus: I did some testing with message list and did not see noticeable improvement (in time profiler). There are a couple of reasons. 
1. `Stagedchangeset.elementUpdated` does not happen often
2. Message list uses many different reuse identifiers, even when we get `elementUpdated` it is highly probable that the computed reuse identifier changes and in that case we must call reload row (table view throws an exception if we call reconfigure but reuse identifier changes)
3. For actually making this work, this closure needs to compare the current reuse identifier by accessing the current cell's reuseIdentifier and also computing the new reuseIdentifier. This gets a bit complex since this code is in ChatMessageListView but it needs to access some functions from ChatMessageListVC for being able to compute the reuse identifier.
```
let reuseIdentifier = self.reuseIdentifier(
            contentViewClass: contentViewClass, <- function on ChatMessageListVC
            attachmentViewInjectorType: attachmentViewInjectorType, <- function on ChatMessageListVC
            layoutOptions: layoutOptions, <- function on ChatMessageListVC
            message: message <- function on ChatMessageListVC
        )
```

After taking all of this into account it felt like it is actually not worth it but I can be convinced otherwise (I did some hack implementation while I was measuring).

Instruments runs (reconfigure rows did show 2 less hitches :))
messagelist

PR #3021, opened 2024-02-09
Skipping channel updates did not work when displaying message list

laevandus:
### 🔗 Issue Links

Resolves [ios-issues-tracking/42](https://github.com/GetStream/ios-issues-tracking/issues/42)

Related [ios-issues-tracking/669](https://github.com/GetStream/ios-issues-tracking/issues/669)

### 🎯 Goal

Fix channel list update skip logic when we are displaying messages instead

### 📝 Summary

Track channel list's visibility with `viewWillAppear` and `viewWillDisappear`

### 🛠 Implementation

We have a logic in place which tries to skip channel updates when user is in the message list instead. Since we use navigation controller, then after pushing the message list to the stack, channel list stays in the view hierarchy. Therefore, we can't check if channel list's window is nil. Instead, we should track it with `viewWillAppear` and `viewWillDisappear`.

Did some extra testing:
- UINavigationController: skipping happens
- UISplitViewController: skipping does not happen (e.g. channel list and message list side by side)
- UITabViewController: skipping happens

### ☑️ Contributor Checklist


nuno-vieira (approved):
LGTM! ✅ Just make sure to update the changelog position 👍

PR #3025, opened 2024-02-12
Fix composer link preview overridden by previous enrichment

nuno-vieira:
### 🔗 Issue Links
None

### 🎯 Goal

Fix composer link preview overriding actual current link preview

### 📝 Summary

- Increases the debouncer to 0.4s to be more aligned with how usually requests take to complete
- Fix composer link preview overriding actual current link preview

### 🛠 Implementation

The proper way to do this is to cancel the previous request whenever we send a new request to enrich the url. The problem is that right now we do not have support for cancellable requests in our SDK, mostly because of tech debt issues. So right now, the best way to improve this is when we show the link preview we need to make sure that url is still the one in the composer input.

### 🧪 Manual Testing Notes

**Link preview should not show after sending the message**
1. Type a valid link in the composer
2. Very quickly after typing it, send the message
3. After the message was sent, it should not show the link preview

**Valid old link preview should not override current link preview**
1. Type a link in the composer
2. Very quickly type a new one
3. Only the new one should be shown

**Incorrect old link preview should not override current link preview**
1. Type an invalid link in the composer
2. Very quickly type a new valid one
3. Only the new one should be shown (The preview should not be dismissed)

### ☑️ Contributor Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 # Upcoming
 
-### 🔄 Changed
+## StreamChatUI
+### 🐞 Fixed
+- Fix overriding of actual current composer link preview [#3025](https://github.com/GetStream/stream-chat-swift/pull/3025)
 
 # [4.48.1](https://github.com/GetStream/stream-chat-swift/releases/tag/4.48.1)
 _February 09, 2024_
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## StreamChatUI
 ### 🐞 Fixed
-- Fix overriding of actual current composer link preview [#3025](https://github.com/GetStream/stream-chat-swift/pull/3025)
+- Fix composer link preview overridden by previous enrichment [#3025](https://github.com/GetStream/stream-chat-swift/pull/3025)
 
 # [4.48.1](https://github.com/GetStream/stream-chat-swift/releases/tag/4.48.1)
 _February 09, 2024_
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -3,7 +3,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 # Upcoming
 
-### 🔄 Changed
+## StreamChatUI
+### 🐞 Fixed
+- Fix overriding of actual current composer link preview [#3025](https://github.com/GetStream/stream-chat-swift/pull/3025)
```
martinmitrevski, line 8: Maybe change the wording a bit here to be more clear? ("actual current" is not entirely clear imo)

Sources/StreamChatUI/Composer/ComposerVC.swift
```diff
@@ -289,7 +289,7 @@ open class ComposerVC: _ViewController,
 
     open var cooldownTracker: CooldownTracker = CooldownTracker(timer: ScheduledStreamTimer(interval: 1))
 
-    public var enrichUrlDebouncer = Debouncer(0.3, queue: .main)
+    public var enrichUrlDebouncer = Debouncer(0.4, queue: .main)
```
martinmitrevski, line 292: Is there any visible delay because of this change?

martinmitrevski (approved):
LGTM! ✅  Just please update the changelog entry a bit.

Sources/StreamChatUI/Composer/ComposerVC.swift
nuno-vieira: There was before the actual fix. But it is better to put 0.4 since it is actually the value proposed in the Figma designs. I initially put it shorter because it was better, but decided to put the actual value from Figma

CHANGELOG.md
nuno-vieira: WDYT of this one:
- Fix composer link preview overridden by previous enrichment

martinmitrevski: much better 👍

testableapple (approved):
🚢

PR #3039, opened 2024-02-20
Fix controllers having empty data when state changed to `.remoteDataFetched` with background mapping enabled

nuno-vieira:
### 🔗 Issue Links
Resolves https://github.com/GetStream/ios-issues-tracking/issues/675

### 🎯 Goal
- Fix search showing empty data with background mapping enabled
- Fix channel list showing empty data for a split second with background mapping enabled

#### Problem
The root cause of Search not working correctly with background mapping is that when we call `state = .remoteDataFetched` after the HTTP Request, at this time, the `viewContext` does not know about the changes because of background mapping. So, the `controller.messages` has empty data when the delegate event `didChangeState` is called. 

#### Solution
The only solution I found without changing the codebase too much is to change how we report the `.remoteDataFetched` state. Now, we will only report this state when the `FRC` reports changes, so that when `didChangeState` is called, we have all the data available.

### 🧪 Manual Testing Notes
**Scenario 1**
1. Logout
2. Login (Test both with Background Mapping enabled and disabled)
3. The first load of the Channel List should not show an empty view

**Scenario 2**
1. Logout
2. Login (Test both with Background Mapping enabled and disabled)
3. Search a message "Hey"
4. Results should show

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift
+++ b/Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift
@@ -89,6 +89,7 @@ public class ChatChannelListController: DataController, DelegateCallable, DataSt
                 }
                 log.debug("didChangeChannels: \(changes.map(\.debugDescription))")
                 $0.controller(self, didChangeChannels: changes)
+                $0.controller(self, didChangeState: state)
             }
         }
 
--- a/Sources/StreamChat/Controllers/DataController.swift
+++ b/Sources/StreamChat/Controllers/DataController.swift
@@ -24,7 +24,9 @@ public class DataController: Controller {
     public internal(set) var state: State = .initialized {
         didSet {
             callback {
-                self.stateMulticastDelegate.invoke { $0.controller(self, didChangeState: self.state) }
+                if self.state != .remoteDataFetched {
+                    self.stateMulticastDelegate.invoke { $0.controller(self, didChangeState: self.state) }
+                }
             }
         }
     }
--- a/Tests/StreamChatTests/Controllers/ChannelWatcherListController/ChatChannelWatcherListController_Tests.swift
+++ b/Tests/StreamChatTests/Controllers/ChannelWatcherListController/ChatChannelWatcherListController_Tests.swift
@@ -266,6 +266,8 @@ final class ChatChannelWatcherListController_Tests: XCTestCase {
 
         // Simulate network call response
         env.watcherListUpdater!.channelWatchers_completion!(nil)
+        // State change is only notified when observer is called with data changes.
+        controller.watchersObserver.onDidChange?([])
 
         // Assert delegate is notified about state changes
         AssertAsync.willBeEqual(delegate.state, .remoteDataFetched)
--- a/Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
+++ b/Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
@@ -156,8 +156,8 @@ public class ChatChannelController: DataController, DelegateCallable, DataStoreP
 
     /// Database observers.
     /// Will be `nil` when observing channel with backend generated `id` is not yet created.
-    private var channelObserver: EntityDatabaseObserverWrapper<ChatChannel, ChannelDTO>?
-    private var messagesObserver: ListDatabaseObserverWrapper<ChatMessage, MessageDTO>?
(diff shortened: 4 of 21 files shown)


Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift
```diff
@@ -89,6 +89,7 @@ public class ChatChannelListController: DataController, DelegateCallable, DataSt
                 }
                 log.debug("didChangeChannels: \(changes.map(\.debugDescription))")
                 $0.controller(self, didChangeChannels: changes)
+                $0.controller(self, didChangeState: state)
```
nuno-vieira, line 92: The only potential risk here is that this changes how the SDK works at the moment 🤔 The `didChangeState` will get called everytime whenever there are data updates, which did not happen before 🤔 So it might be worth having some logic to call this only on the first page when remoteDataFetched 🤔

martinmitrevski: Theoretically, it shouldn't change much, I think. We will just have more state updates with the same state. However, to be safe, maybe we can add more conditions here, like first page + background mapping enabled.

Sources/StreamChat/Controllers/DataController.swift
```diff
@@ -24,7 +24,14 @@ public class DataController: Controller {
     public internal(set) var state: State = .initialized {
         didSet {
             callback {
-                self.stateMulticastDelegate.invoke { $0.controller(self, didChangeState: self.state) }
+                // We only report state changes if not `.remoteDataFetched` because the data
+                // coming from the server will only be available once `FRC` reports changes,
+                // so when that happens, only then we report `didChangeState`. This is required
+                // mostly because of background mapping.
+                // More context here: https://github.com/GetStream/stream-chat-swift/pull/3039
+                if self.state != .remoteDataFetched {
```
martinmitrevski, line 32: Also here, I would prefer if we limit this state check to BM enabled only, wdyt?

Tests/StreamChatTests/Controllers/ChannelWatcherListController/ChatChannelWatcherListController_Tests.swift
```diff
@@ -266,6 +266,8 @@ final class ChatChannelWatcherListController_Tests: XCTestCase {
 
         // Simulate network call response
         env.watcherListUpdater!.channelWatchers_completion!(nil)
+        // State change is only notified when observer is called with data changes.
+        controller.watchersObserver.onDidChange?([])
```
martinmitrevski, line 270: Not sure if we should expose the observers to the tests. If you want to simulate sth, maybe it's better to send an event via the event notification center? Same applies for the other tests.

Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift
laevandus: I have been looking at this for some time and the only thought is that since all of this is related to background mapping then should we only do this when background mapping is enabled. If background mapping gets enabled for everyone at some point then we could just clean up that check. Less risk that additional `didChangeState` would degrade customer's implementation somehow. 🤔

martinmitrevski (commented):
Great that you found the root cause 👍  I think we can make it a bit stricter though, there's no need to affect the non-BM implementation for now.

Tests/StreamChatTests/Controllers/ChannelWatcherListController/ChatChannelWatcherListController_Tests.swift
nuno-vieira: It is fine to expose the observers internally, actually the tests should be like this in the first place, but they would need to be injected in the initialiser, which at the moment is not possible. Either way, since were going to do this only in the BG Mapping, I will remove these tests

Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift
nuno-vieira: We agreed on doing this change only with BG Mapping enabled 👍

Sources/StreamChat/Controllers/DataController.swift
nuno-vieira: We agreed on doing this change only with BG Mapping enabled 👍

martinmitrevski (approved):
LGTM ✅

laevandus (approved):

PR #3042, opened 2024-02-22
Fix controllers having empty data when state changed to .remoteDataFetched with background mapping enabled

nuno-vieira:
### 🔗 Issue Links
Resolves https://github.com/GetStream/ios-issues-tracking/issues/675
Resolves https://github.com/GetStream/ios-issues-tracking/issues/626

### 🎯 Goal
- Fix search showing empty data with background mapping enabled
- Fix channel list showing empty data for a split second with background mapping enabled

#### Problem
The root cause of Search not working correctly with background mapping is that when we call `state = .remoteDataFetched` after the HTTP Request, at this time, the items from the background observer are not yet populated. They are only populated when the FRC Notifies of changes, which because it is in a background thread, is an async process.  So, the `controller.messages` has empty data when the delegate event `didChangeState` is called. 

#### Solution
After trying multiple solutions, I think this one is the one which requires the least changes and has less impact on the customer's app. The solution involves in adding a function to the background observer to "force" refreshing the items of the observer and provide a completion block to notify when they are updated. This way, we can call this refresh function and only call `didChangeState` event when the items are updated. 

### 🧪 Manual Testing Notes
**Scenario 1**
1. Logout
2. Login (Test both with Background Mapping enabled and disabled)
3. The first load of the Channel List should not show an empty view

**Scenario 2**
1. Logout
2. Login (Test both with Background Mapping enabled and disabled)
3. Search a message "Hey"
4. Results should show

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/Controllers/DatabaseObserver/BackgroundDatabaseObserver.swift
+++ b/Sources/StreamChat/Controllers/DatabaseObserver/BackgroundDatabaseObserver.swift
@@ -143,16 +143,18 @@ class BackgroundDatabaseObserver<Item, DTO: NSManagedObject> {
 
     /// This method will add a new operation to the `processingQueue`, where operations are executed one-by-one.
     /// The operation added to the queue will start the process of getting new results for the observer.
-    private func updateItems(changes: [ListChange<Item>]?) {
+    func updateItems(changes: [ListChange<Item>]?, completion: (() -> Void)? = nil) {
         let operation = AsyncOperation { [weak self] _, done in
             guard let self = self else {
                 done(.continue)
+                completion?()
                 return
             }
 
             self.frc.managedObjectContext.perform {
                 self.processItems(changes) {
                     done(.continue)
+                    completion?()
                 }
             }
         }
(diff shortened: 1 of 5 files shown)


nuno-vieira:
Closing in favor of: https://github.com/GetStream/stream-chat-swift/pull/3042

Sources/StreamChat/Controllers/DatabaseObserver/BackgroundDatabaseObserver.swift
```diff
@@ -143,16 +143,18 @@ class BackgroundDatabaseObserver<Item, DTO: NSManagedObject> {
 
     /// This method will add a new operation to the `processingQueue`, where operations are executed one-by-one.
     /// The operation added to the queue will start the process of getting new results for the observer.
-    private func updateItems(changes: [ListChange<Item>]?) {
+    func updateItems(changes: [ListChange<Item>]?, completion: (() -> Void)? = nil) {
         let operation = AsyncOperation { [weak self] _, done in
```
laevandus, line 147: I see, since the processing queue is serial we are guaranteed that it runs as a last thing.

laevandus (approved):

martinmitrevski (approved):

testableapple (approved):

PR #3043, opened 2024-02-23
Add support for user mentions in channels with 100+ members

nuno-vieira:
### 🔗 Issue Links
Resolves https://github.com/GetStream/ios-issues-tracking/issues/535

### 🎯 Goal
Add support for user mentions in channels with 100+ members.

### 🛠 Implementation
Previously we were only searching channel members using the local data for user mentions. This meant that if a channel had 100+ members it would not be possible to search for all members in the channel.

Now, when the members of the channel are bigger than the current members in the local DB, we perform an actual channel member query request.

### 🧪 Manual Testing Notes
**Scenario 1 - Search All Users**
1. Enable mention all app users
2. Open a channel
3. Type "@"
4. It should show users from all channels

**Scenario 1 - Search Remote Channel Members**
1. Disable mention of all app users (Default)
2. Open a channel with 100+ Members (ex: Release 4.46.0)
3. Type "@Lei"
4. It should show users in mentions suggestions
**NOTE:** The `Release 4.46.0` channel contains users without names, searching channel members is only supported with names.

**Scenario 1 - Search Local Channel Members**
1. Disable mention of all app users (Default)
2. Open a channel with 100- Members
3. Type "@"
4. It should show users in mentions suggestions

### ☑️ Contributor Checklist


Sources/StreamChatUI/Composer/ComposerVC.swift
```diff
@@ -893,44 +903,49 @@ open class ComposerVC: _ViewController,
...
+            filter: .autocomplete(.name, text: term),
+            sort: [.init(key: .name, isAscending: true)]
+        )
+    }
+
+    /// Returns the member list controller to be used for searching members inside a channel.
+    ///
+    /// - Parameter term: The potential user mention the current user is typing.
+    /// - Returns: `ChatChannelMemberListController` instance that will be used for searching members in a channel.
+    open func makeMemberListControllerForMemberSuggestions(typingMention term: String) -> ChatChannelMemberListController? {
```
laevandus, line 929: Would it be better to take a ChannelMemberListQuery as an argument here for forcing subclassers to override once function for query and the other for UI? Currently I could just override this function and skip the `queryForChannelMentionSuggestionsSearch` all together. WDYT? Then subclassers have less ways for making a mistake and getting confused (someone overrides makeMember, other dev overrides query, but later it turns out that the makeMember subclasser never calls queryFor anymore, these kinds of simple issues).

martinmitrevski: that's a good point

martinmitrevski (approved):
LGTM ✅  Please consider @laevandus suggestion about the public API.

Sources/StreamChatUI/Composer/ComposerVC.swift
nuno-vieira: Yeah, I think it makes sense, although it removes some flexibility, in case we have other ways of initialising the controller without a member query 🤔 But in this case it should be unlikely tho

nuno-vieira: Actually, I would prefer to have only `makeMemberListControllerForMemberSuggestions(typingMention:)` and not `queryForChannelMentionSuggestionsSearch`, the only reason I created the `queryForChannelMentionSuggestionsSearch` was because we already had `queryForMentionSuggestionsSearch`, so I wanted to have consistency. But having only `makeMemberListControllerForMemberSuggestions(typingMention:)` would be enough and the most flexible approach. 

The disadvantage of your suggestion is that it will always require doing 2 function calls to pass to create the memberListController, one for creating the member list query and another for creating the actual controller. 

Right now, we just do this:
```swift
memberListController = makeMemberListControllerForMemberSuggestions(typingMention: typingMention)
```

With your suggestion, we would need to do this:
```swift
let query = queryForChannelMentionSuggestionsSearch(typingMention: typingMention)
memberListController = makeMemberListControllerForMemberSuggestions(query: query)
```

The 2nd approach has some disadvantages:
1- If a customer wants to create the member list controller in another place and wants to make sure that it uses the same query as in the other places, they will always need to follow this pattern. They might forget to call `queryForChannelMentionSuggestionsSearch(typingMention: typingMention)` or know that it actually exists, and so they create a MemberListQuery from scratch. 
2- It forces having a MemberListQuery to create the memberListController, which right now is a good thing, but it removes some flexibility which is something that is usually a bad practice in the UI SDK since customers usually have a lot of different use cases.

DemoApp/StreamChat/Components/DemoChatChannelListRouter.swift
```diff
@@ -409,7 +409,7 @@ final class DemoChatChannelListRouter: ChatChannelListRouter {
                 guard let cid = channelController.channel?.cid else { return }
                 let client = channelController.client
                 self.rootViewController.present(MembersViewController(
-                    membersController: client.memberListController(query: .init(cid: cid))
+                    membersController: client.memberListController(query: .init(cid: cid, pageSize: 105))
```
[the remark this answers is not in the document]
nuno-vieira: It was just a random value, to be able to see the member list for channels above 100+

laevandus (approved):

Sources/StreamChatUI/Composer/ComposerVC.swift
nuno-vieira: Discussed on Slack ✅

testableapple (approved):

Issue #3049, opened 2024-02-28
App crash on creating a new 1:1 chat, found nil when unwrapping optional in ListDataBaseObserver


kirsanov-dev:
## What did you do?
Create new 1:1 message with a new user with empty list of chats (may also happen with a user with some other chats in the list)

## What did you expect to happen?
New chat appear on the Chats screen.

## What happened instead?
App crash with an error "Unexpectedly found nil when implicitly unwrapping an Optional value" in ListDataBaseObserver class in StreamChat source code. The app is crashing on both phone A (the creator of new chat) and phone B (the user who is chat created with).

<img width="719" alt="Screenshot at Feb 28 10-17-34" src="https://github.com/GetStream/stream-chat-swift/assets/80982911/347f950a-fde0-417c-9595-12dbf23b2334">

## GetStream Environment
**GetStream Chat version: 4.48.0**
**GetStream Chat frameworks:** StreamChat, StreamChatUI
**iOS version: 17.2**
**Swift version: 5.8.1**
**Xcode version: 15.2**
**Device: phone A (creator of the chat) iPhone 8, iOS 16.7.2, phone B (target user) iPhone 15 iOS 17.2**

kirsanov-dev:
Video of the issue
https://github.com/GetStream/stream-chat-swift/assets/80982911/4b7818a5-4c05-49bd-88b5-d2901c5221cf

testableapple:
Hi @kirsanov-dev,
Thanks for opening the issue.

Unfortunately, I was not able to reproduce it on our side. Could you clarify something, please:
- Could you share the crash log?
- Could you also share the code snippets on how you're doing the logout? In particular I'm interested if you're waiting for the completion block of the logout function.
- Could you also share how you are creating the ChatClient? Do you destroy the ChatClient at any point? 
- Do you have Background Mapping enabled?
- Could you check out 4.49.0? We had some regression in 4.48.0. Even though that one is not related, it's worth trying the latest release, just in case.

Best,
Alexey

kirsanov-dev:
Hi @testableapple 
1. Couldn't find crash logs.
2.     
```
// Close the WebSocket connection when component dismounts
    func disconnectUser(completion: (() -> Void)? = nil) {
        chatClient.disconnect {
            completion?()
        }
    }

```
```
    func logout() {
        // Since we don't need the result of logout() response we call subscribe() right here
        _ = authRepository.logout().subscribe()

[...]

PR #3053, opened 2024-02-29
Stop throwing InvalidModel when dto is invalid

nuno-vieira:
### 🔗 Issue Links
Related to https://github.com/GetStream/stream-chat-swift/issues/3049

### 🎯 Goal
Fix rare crash in `startObserver()` on logout when `DTO` is not valid

### 📝 Summary
Since nowadays we just delete the whole data when logging out and do not delete the Core Data store, we don't need this workaround anymore of checking if a DTO is valid or not. In fact, this was causing more problems now. Since we want to avoid as much as possible throwing `InvalidModel` since this will cause a crash when converting DTO's to Models. 

### 🧪 Manual Testing Notes
Requires smoke testing when logging in and logging out. 

### ☑️ Contributor Checklist


Sources/StreamChat/Database/DTOs/ChannelConfigDTO.swift
```diff
@@ -24,8 +24,7 @@ final class ChannelConfigDTO: NSManagedObject {
     @NSManaged var commands: NSOrderedSet
 
     func asModel() throws -> ChannelConfig {
-        guard isValid else { throw InvalidModel(self) }
```
laevandus, line 27: Can we also delete `isValid` from our extension of `NSManagedObject`?

laevandus (approved):

Sources/StreamChat/Database/DTOs/ChannelConfigDTO.swift
martinmitrevski: Trying to understand what's the impact of this change. I see `isValid` just checks if there's an existing object with that identifier. Now, even if the object was potentially deleted, the data in memory is still somehow available, and the corresponding model would just be populated, right?

nuno-vieira: Yes. The thing is that, worse it can't be than this. If isValid is false and it reaches this point, it crashes. So if it does not crash at this point, at most it can crash later, either way having this is the worst scenario, so better to remove it

nuno-vieira: @laevandus We cant remove it, since it is used in other places

PR #3055, opened 2024-03-01
Fix current user cache not deleted on logout causing unread count issues after switching users

nuno-vieira:
### 🔗 Issue Links

Resolves https://github.com/GetStream/ios-issues-tracking/issues/755
Related to https://github.com/GetStream/stream-chat-swift/issues/3049

### 🎯 Goal

Fix current user cache not deleted on logout causing unread count issues after switching users

### 🛠 Implementation
We only found this issue after switching our DemoApp to also use a ChatClient singleton. Since most customers integrate the SDK like that, we should replicate it on the DemoApp too. The issue here was that we were invalidating the `currentUser` cache, BUT only the writableContext, and we need to do it on all of our managed object contexts.

### 🧪 Manual Testing Notes
1. Login with Han Solo
2. Check the unread counts
3. Logout
4. Login with Leia Organa
5. Unread counts should be different and not the same from 1.
6. Send a message from another device
7. Unread Count should increment on that channel

### ☑️ Contributor Checklist


Sources/StreamChat/Database/DatabaseContainer.swift
```diff
@@ -224,8 +224,14 @@ class DatabaseContainer: NSPersistentContainer {
 
     /// Removes all data from the local storage.
     func removeAllData(completion: ((Error?) -> Void)? = nil) {
+        /// Cleanup the current user cache for all manage object contexts.
+        allContext.forEach { context in
+            context.perform {
```
laevandus, line 229: Should we block here with `performAndWait` like we do with writableContext just below? Or should we chain performs (writableContext below + allContext here) and then calling the completion handler?

Sources/StreamChat/Database/DatabaseContainer.swift
```diff
@@ -59,7 +59,7 @@ class DatabaseContainer: NSPersistentContainer {
     static var cachedModel: NSManagedObjectModel?
 
     /// All `NSManagedObjectContext`s this container owns.
-    private lazy var allContext: [NSManagedObjectContext] = [viewContext, backgroundReadOnlyContext, writableContext]
+    private(set) lazy var allContext: [NSManagedObjectContext] = [viewContext, backgroundReadOnlyContext, writableContext]
```
laevandus, line 62: Adding `s` to the `allContext` property would be nice but not an important issue here.

martinmitrevski (approved):

Sources/StreamChat/Database/DatabaseContainer.swift
nuno-vieira: @laevandus It should be ok to not wait here since it is a quite fast operation, and the other one which takes more time we will wait for it 👍

Sources/StreamChat/Database/DatabaseContainer.swift
nuno-vieira: In order to not re-run the CI, I'll leave this one for now 👍

PR #3057, opened 2024-03-05
[Chat state layer] channel lists

laevandus:
### 🔗 Issue Links

Related to [#728](https://github.com/GetStream/ios-issues-tracking/issues/728)

### 🎯 Goal

**Merges to another feature branch**

Add async-await supported `ChatList` and its supporting observable state object `ChatListState` (replacement for `ChatChannelListController`).

### 📝 Summary

- ChatList: contains actions for paginated loading
- ChatListState: observable object which observes DB changes and manages a sorted list of channels matching the query
- No unit-tests is expected since the final interface is WIP

### ☑️ Contributor Checklist


Sources/StreamChat/ChatClient.swift
```diff
@@ -74,6 +74,8 @@ public class ChatClient {
     let callRepository: CallRepository
 
     let channelRepository: ChannelRepository
+    
+    let channelListUpdater: ChannelListUpdater
```
laevandus, line 78: Added it here since it does not contain any state and therefore can be reused.

Sources/StreamChat/ChatClient.swift
```diff
@@ -355,8 +361,7 @@ public class ChatClient {
         authenticationRepository.logOutUser()
 
         // Stop tracking active components
-        activeChannelControllers.removeAllObjects()
-        activeChannelListControllers.removeAllObjects()
+        syncRepository.resetTracking()
```
laevandus, line 364: I needed the similar tracking for the new type. `SyncRepository` also calls these remove calls within `resetTracking`.

Sources/StreamChat/Extensions/BidirectionalCollection+Extensions.swift
```diff
@@ -47,6 +47,76 @@ extension BidirectionalCollection {
 
 extension BidirectionalCollection where Element == ChatMessage {
     func uniquelyMerged(_ insertedSortedElements: [Element]) -> [Element] {
-        uniquelyMerged(insertedSortedElements, areInIncreasingOrder: { first, second in first.createdAt < second.createdAt }, unique: { first, second in first.id != second.id })
+        uniquelyMerged(insertedSortedElements, areInIncreasingOrder: { first, second in first.createdAt < second.createdAt }, dropsExisting: { first, second in first.id == second.id })
+    }
+    
+    func uniquelyApplied(_ changes: [ListChange<Element>]) -> [Element] {
```
laevandus, line 53: It is going to need a lot of tests but currently I am planning to first create the public API layer first. No worries, a lot of tests coming after that.

In short, this is used for mutating an (native) array of channels/messages (not backed by FRC like we have in controllers where: `public var channels: LazyCachedMapCollection<ChatChannel> {`).

Sources/StreamChat/Repositories/SyncOperations.swift
```diff
@@ -104,23 +104,44 @@ final class RefetchChannelListQueryOperation: AsyncOperation {
...
-                        subsystems: .offlineSupport
-                    )
-                    done(.retry)
-                }
+                Self.handleReset(query, with: result, context: context, done: done)
             }
         }
     }
+    
+    init(queryProvider: @escaping () -> ChannelListQuery?, updater: ChannelListUpdater, context: SyncContext) {
```
laevandus, line 112: Since we have a new type then I had to make sure that if someone uses that, then the sync repo resets channels in that case as well. The new type, ChatList is a struct, so that is why I did this closure based approach. Also, ChatList is 13.0 or later, I can't use the type in this code without a lot of available checks.

Sources/StreamChat/StateLayer/ChatClient+ChatList.swift
```diff
@@ -0,0 +1,31 @@
...
+    ///   - dynamicFilter: A filter block for filtering by channel's extra data fields or as a manual filter when ``ChatClientConfig.isChannelAutomaticFilteringEnabled`` is false ([read more](https://getstream.io/chat/docs/sdk/ios/client/controllers/channels/)).
+    ///
+    /// - Throws: An error while communicating with the Stream API.
+    /// - Returns: An instance of ``ChatList`` which represents actions and the current state of the list.
+    public func makeChatList(with query: ChannelListQuery, dynamicFilter: ((ChatChannel) -> Bool)? = nil) async throws -> ChatList {
+        let channels = try await channelListUpdater.update(channelListQuery: query)
+        let list = ChatList(channels: channels, query: query, dynamicFilter: dynamicFilter, channelListUpdater: channelListUpdater, client: self)
+        let state = list.state
+        syncRepository.trackChannelListQuery { [weak state] in state?.query }
+        return list
```
laevandus, line 29: Not the most happy with this but I feel like it needs to be tackled later. The core idea is that SyncRepository wants to know which queries are active, active are the ones we have a ChatChannelListController instance or a ChatList + ChatListState combo. Whenever ChatListState gets deallocated, the query closure returns nil.

I added a future TODO to rethink this since it is not so nice.

Sources/StreamChat/StateLayer/ChatList.swift
```diff
@@ -0,0 +1,99 @@
...
+            query,
+            client.config,
+            channelListUpdater,
+            client.channelRepository
+        )
+        self.state = state
+    }
+    
+    /// An observable object representing the current state of the channel list.
+    public let state: ChatListState
```
laevandus, line 45: I have been thinking about using `@dynamicMemberLookup` for this type which would make it easier to access properties in the state type. Currently I left it out but it is part of my notes for future iterations.

Sources/StreamChat/StateLayer/ChatListState.swift
```diff
@@ -0,0 +1,53 @@
...
+        )
+    }
+    
+    /// An array of channels for the specified ``ChannelListQuery``.
+    @Published public private(set) var channels = [ChatChannel]()
+    
+    /// True, if all the channels were loaded with paginated loading.
+    @Published public private(set) var hasLoadedAllChannels = false
+    
+    // MARK: - Mutating the State
```
laevandus, line 35: Since this is ObservableState then here are MainActor annotated functions which will force mutations on the main thread and therefore, should cause less issues with threading and using it in SwiftUI views directly.

But this is not safe from cases where someone reads properties from background threads and at the same time keeps mutating the state on the main actor/thread.

Sources/StreamChat/StateLayer/Coordinators/ChannelListQueryLinkingCoordinator.swift
```diff
@@ -0,0 +1,100 @@
+//
+// Copyright © 2024 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+/// Listens to local events and adds or removes channels from ``ChannelListQueryDTO.channels``.
+@available(iOS 13.0, *)
+struct ChannelListQueryLinkingCoordinator {
```
laevandus, line 9: This is the same linking/unlinking logic we have in the controller. Needs some rethinking in next iterations.

Sources/StreamChat/StateLayer/General/OrderedChannels.swift
```diff
@@ -0,0 +1,21 @@
+//
+// Copyright © 2024 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+struct OrderedChannels {
```
laevandus, line 7: Just a convenience for updating ChatListState.channels

Sources/StreamChat/StateLayer/Pagination/PaginatedChannelsLoader.swift
```diff
@@ -0,0 +1,43 @@
...
+        // Optimization: remove the database fetch and instead replicate filtering and sorting defined by ChannelDTO.channelListFetchRequest(query:chatClientConfig:). Currently we can't ensure that payloadChannels is sorted the same way.
+        let newSortedChannels = try await channelRepository.channels(
+            for: payloadChannels.map(\.cid),
+            query: query,
+            config: chatClientConfig
+        )
+        let result = await state.orderedChannels.withInsertingPaginated(newSortedChannels)
+        await state.setSortedChannels(result, canLoadMore: payloadChannels.count < pagination.pageSize)
+        
+        return payloadChannels
```
laevandus, line 26: If we move away from FRC driven data source then here we can optimise pagination to only use API requests and its payloads and skipping some of the DB writes and reads. This is for future iterations.

Sources/StreamChat/StateLayer/ChatClient+ChatList.swift
laevandus: While I was looking into this then there seems to be these options:

1. Follow what ChatChannelListController is doing: keep track of allocated instances weakly, when syncing occurs, reset `ChannelListQueryDTO` channels. e.g. `queryDTO.channels.removeAll()`. Con: is ChatClient+SyncRepository needs to track instances. Pro: less updating, no need to figure out how to store `ChannelListQuery` which is required for resetting/reloading.
2. Remove tracking of active controllers/queries  from `SyncRepository` and instead, `SyncRepository` could fetch all the `ChannelListQueryDTO`s and reset `channels` properties. Pro: no coupling Con: controller/ChatList need to figure out when resetting happens and then triggering reload themselves. The challenge is that we need `ChannelListQuery` for loading and only controller and ChatList knows that.
3. SyncRepository does not reset any channels for queries and instead sends some event which ChatChannelListController and ChatList listen to. When this happens, they call `resetChannelsQuery` themselves. Pro: no weakly tracking Con: reset requires `watchedAndSynchedChannelIds`, `synchedChannelIds` arguments which are managed by SyncRepository which makes it more challenging.

I am happy to try 2 or 3 but I feel like it should then be a separate PR for the feature branch.

Sources/StreamChat/StateLayer/ChatListState.swift
laevandus: If merging and sorting lists does not end up on the main thread, it should be fine, that is a goal. If we are talking about thousands of channels per query, then we probably need to be smarter with paginated loading and unload some of them. 

I need review entity observers since these are calling `onDidChange` on the main thread. Probably a separate PR to move onDidChange to background thread (only for new code). Added a note for myself to notion.

Sources/StreamChat/StateLayer/General/OrderedChannels.swift
laevandus: I'll remove this one since it does not have much code anyway.

Sources/StreamChat/Repositories/SyncRepository.swift
```diff
@@ -68,6 +72,20 @@ class SyncRepository {
     deinit {
         cancelRecoveryFlow()
     }
+    
+    // MARK: - Tracking Active
```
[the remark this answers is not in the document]
laevandus: Let's use [this comment](https://github.com/GetStream/stream-chat-swift/pull/3057#discussion_r1512691224) for sync repo discussion.

Sources/StreamChat/Repositories/SyncRepository.swift
```diff
@@ -134,6 +152,16 @@ class SyncRepository {
                 )
             }
         operations.append(contentsOf: refetchChannelListQueryOperations)
+        operations.append(contentsOf: propertyQueue.sync {
```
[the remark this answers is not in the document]
laevandus: Let's use [this comment](https://github.com/GetStream/stream-chat-swift/pull/3057#discussion_r1512691224) for sync repo discussion.

Sources/StreamChat/StateLayer/ChatList.swift
```diff
@@ -0,0 +1,99 @@
...
+
+/// An object which represents a list of `ChatChannel`.
+@available(iOS 13.0, *)
+public struct ChatList {
+    /// The query specifying and filtering the list of channels.
+    public let query: ChannelListQuery
+    
+    private let channelListQueryLinkingCoordinator: ChannelListQueryLinkingCoordinator
+    private let channelListUpdater: ChannelListUpdater
+    private let paginatedChannelsLoader: PaginatedChannelsLoader
```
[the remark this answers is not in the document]
laevandus: Some of these could be nested types instead (the ones I added since these are only meant for `ChatList` or `Chat` and nowhere else).

Sources/StreamChat/StateLayer/ChatList.swift
```diff
@@ -0,0 +1,99 @@
...
+    }
+    
+    /// An observable object representing the current state of the channel list.
+    public let state: ChatListState
+    
+    // MARK: - Channel List Pagination
+    
+    /// Loads channels for the specified pagination parameters and updates ``ChatListState.channels``.
+    ///
+    /// - Parameters:
```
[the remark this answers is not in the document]
laevandus: That is true, I ended up manually changing it since then it is consistent with other parameter formats. If you prefer that, I have no problem for changing the format to that. There are probably many other cases as well. Adding an item to the todo list in Notion.

Sources/StreamChat/StateLayer/Coordinators/ChannelListQueryLinkingCoordinator.swift
```diff
@@ -0,0 +1,100 @@
...
+        /// When we receive events, we need to check if a channel should be added or removed from
+        /// the current query depending on the following events:
+        /// - Channel created: We analyse if the channel should be added to the current query.
+        /// - New message sent: This means the channel will reorder and appear on first position,
+        ///   so we also analyse if it should be added to the current query.
+        /// - Channel is updated: We only check if we should remove it from the current query.
+        ///   We don't try to add it to the current query to not mess with pagination.
+        func eventsController(_ controller: EventsController, didReceiveEvent event: Event) {
+            switch event {
+            case let channelAdded as NotificationAddedToChannelEvent:
```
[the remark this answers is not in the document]
laevandus: To be honest, the whole type is an ouch and screams for something better. It is part of my todo list on Notion for future iterations. currently just mimics what we had in the controller.

Sources/StreamChat/StateLayer/ChatList.swift
```diff
@@ -0,0 +1,99 @@
...
+
+import Foundation
+
+/// An object which represents a list of `ChatChannel`.
+@available(iOS 13.0, *)
+public struct ChatList {
+    /// The query specifying and filtering the list of channels.
+    public let query: ChannelListQuery
+    
+    private let channelListQueryLinkingCoordinator: ChannelListQueryLinkingCoordinator
```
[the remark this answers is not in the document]
laevandus: Added if for tech discussion

Sources/StreamChat/Extensions/BidirectionalCollection+Extensions.swift
laevandus: We should probably review this during the next tech discussion as well. Adding it as a topic.

Sources/StreamChat/StateLayer/General/OrderedChannels.swift
laevandus: Removed

Sources/StreamChat/StateLayer/ChatList.swift
```diff
@@ -0,0 +1,99 @@
+//
+// Copyright © 2024 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+/// An object which represents a list of `ChatChannel`.
+@available(iOS 13.0, *)
+public struct ChatList {
```
laevandus, line 9: Maybe `ChannelList` is a better name 🤔

Sources/StreamChat/StateLayer/ChatClient+ChatList.swift
```diff
@@ -0,0 +1,31 @@
...
+// Copyright © 2024 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+// MARK: - Factory Methods for Creating Chat Lists
+
+@available(iOS 13.0, *)
+extension ChatClient {
+    /// Creates an instance of ``ChatList`` which represents an array of channels matching to the specified ``ChannelListQuery``.
```
nuno-vieira, line 11: We are talking about channels, but it creates a "ChatList". It is probably better to rename this to `ChannelList` and maybe even rename `Chat` to `Channel` 🤔 It will probably be confusing to old customers who migrate and we have all these new names 🤔 

We need to clarify the namings since it can be confusing:
- Channel List or Chat List
- Channel or Chat or MessageList
- Thread or MessageList

One idea for UI:
- MessageList: The tables of messages.
- Chat: A message list and a composer tied together.
- Channel: A Chat in a channel.
- Thread: A Chat in a thread.

So in terms of LLC, a Chat, could be a Low level component that is shared between the Channel and the Thread.

Sources/StreamChat/StateLayer/ChatList.swift
```diff
@@ -0,0 +1,99 @@
...
+    /// The query specifying and filtering the list of channels.
+    public let query: ChannelListQuery
```
nuno-vieira, line 11: Like I said above, it feels were we have **Chat**List, but then we provic a **Channel**ListQuery
```suggestion
public struct ChannelList {
    /// The query specifying and filtering the list of channels.
    public let query: ChannelListQuery
```

Sources/StreamChat/StateLayer/Coordinators/ChannelListQueryLinkingCoordinator.swift
```diff
@@ -0,0 +1,100 @@
...
+struct ChannelListQueryLinkingCoordinator {
+    private let eventsController: EventsController
+    private let eventsDelegate: EventsDelegate
+    
+    init(state: ChatListState, query: ChannelListQuery, dynamicFilter: ((ChatChannel) -> Bool)?, channelListUpdater: ChannelListUpdater, eventsController: EventsController, database: DatabaseContainer, clientConfig: ChatClientConfig) {
+        self.eventsController = eventsController
+        eventsDelegate = EventsDelegate(database: database)
+        
+        eventsController.delegate = eventsDelegate
+        eventsDelegate.linkingHandler = { [weak state] channel, shouldLink in
```
nuno-vieira, line 18: I think it would make sense to extract this as a class or struct component. Otherwise, it will be harder to debug being a closure., since stack traces don't give much info with closures and this function is quite large and important

Sources/StreamChat/StateLayer/ChatListState.swift
```diff
@@ -0,0 +1,50 @@
...
+                await self.setValue(channels.uniquelyApplied(changes, sorting: query.sort), for: \.channels)
+            })
+        )
+    }
+    
+    /// An array of channels for the specified ``ChannelListQuery``.
+    @Published public private(set) var channels = [ChatChannel]()
+    
+    /// True, if all the channels were loaded with paginated loading.
+    @Published public private(set) var hasLoadedAllChannels = false
```
nuno-vieira, line 34: How do we distinguish between old channels and new channels?

For example, when jumping to a mid page, we want to know when we finish loading all new channels but also all old channels

Sources/StreamChat/StateLayer/Coordinators/ChannelListQueryLinkingCoordinator.swift
```diff
@@ -0,0 +1,100 @@
...
+            let isQueryMatching: (ChatChannel) -> Bool = { channel in
+                if let dynamicFilter {
+                    return dynamicFilter(channel)
+                }
+                if clientConfig.isChannelAutomaticFilteringEnabled {
+                    // When auto-filtering is enabled the channel will appear or not automatically if the
+                    // query matches the DB Predicate. So here we default to saying it always belong to the current query.
+                    return true
+                }
+                return channel.membership != nil
```
nuno-vieira, line 33: I guess this was the old fallback. But in reality, this should be false, if automatic filtering is disabled and no filter block is provided, we should not make assumptions, and the channel should not be included ever.
```suggestion
                return false
```

Sources/StreamChat/StateLayer/Coordinators/ChannelListQueryLinkingCoordinator.swift
nuno-vieira, line 18:
```suggestion
        eventsDelegate.linkingHandler = { [weak state] channel, **shouldLink** in
```
Where does the shouldLink come from?

Sources/StreamChat/StateLayer/ChatClient+ChatList.swift
martinmitrevski: yes, I also think `ChannelList` makes more sense. Not sure I agree about `Chat`, as an interactor object that is used in the whole chat view, might make sense to keep it (let's discuss it).

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
```diff
@@ -372,6 +372,17 @@ extension ChannelDTO {
         request.fetchBatchSize = query.pagination.pageSize
         return request
     }
+    
+    static func loadChannelList(query: ChannelListQuery, chatClientConfig: ChatClientConfig, ids: Set<ChannelId>, context: NSManagedObjectContext) throws -> [ChannelDTO] {
```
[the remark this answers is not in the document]
laevandus: I might be able to skip this since ChannelListQuery requires to sort outside of CoreData fetch request. We do that also on the entity observer level.

laevandus:
There are going to be larger changes around this. Taking account all the comments so no feedback is lost. Thank you.

PR #3063, opened 2024-03-07
Add a static benchmark for xcmetrics

testableapple:
### 📝 Summary

- Create a static benchmark
- Convert the number of hitches to float instead of int
- Fix an issue with zero division
- Bump timeout on GitHub Actions (sometimes there is a queue on Firebase)


nuno-vieira (approved):
LGTM! ✅

PR #3064, opened 2024-03-08
[ChatStateLayer] StreamCollection as opaque base collection wrapper

laevandus:
### 🔗 Issue Links

### 🎯 Goal

Add `StreamCollection` which allows dynamically to change the internal base collection (to different types).

### 🛠 Implementation

The new state layer uses `LazyCachedMapCollection` as the backing storage for items fetched from the database. Later, we would like to change it to a native Swift array. `StreamCollection` acts as the middle layer until this has happened allowing to change the base collection without requiring to make breaking changes to the public API layer. Ideally, the custom collection type is removed with one API breaking change, but this depends on when the new data flow lands.

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/Utils/StreamCollection.swift
+++ b/Sources/StreamChat/Utils/StreamCollection.swift
@@ -1,0 +1,41 @@
+//
+// Copyright © 2024 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+/// A collection wrapping the base collection which allows base collections of different types.
+///
+/// - Note: The type of the base collection can change in the future.
+public struct StreamCollection<Element>: RandomAccessCollection {
+    public typealias Index = Int
+
+    private let _endIndex: () -> Index
+    private let _position: (Index) -> Element
+    private let _startIndex: () -> Index
+
+    /// Creates an instance of the collection using the base collection as the data source.
+    init<BaseCollection>(_ baseCollection: BaseCollection) where BaseCollection: RandomAccessCollection, BaseCollection.Element == Element, BaseCollection.Index == Index {
+        _endIndex = { baseCollection.endIndex }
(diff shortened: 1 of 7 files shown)


Sources/StreamChat/Utils/StreamCollection.swift
```diff
@@ -0,0 +1,41 @@
...
+
+/// A collection wrapping the base collection which allows base collections of different types.
+///
+/// - Note: The type of the base collection can change in the future.
+public struct StreamCollection<Element>: RandomAccessCollection {
+    public typealias Index = Int
+
+    private let _endIndex: () -> Index
+    private let _position: (Index) -> Element
+    private let _startIndex: () -> Index
```
laevandus, line 15: Using closures allows hiding the type of the base collection since for now we want to use `LazyCachedMapCollection`, later `Array`.

martinmitrevski (approved):

Sources/StreamChat/Utils/StreamCollection.swift
```diff
@@ -0,0 +1,41 @@
...
+    /// Accesses the element at the specified position.
+    public subscript(position: Index) -> Element {
+        _position(position)
+    }
+}
+
+extension StreamCollection: CustomStringConvertible {
+    public var description: String {
+        let contents = map { String(describing: $0) }.joined(separator: ", ")
+        return "\(Self.self)(\(contents))"
```
ipavlidakis, line 39: nit: `String(describing: type(of: self))`

Sources/StreamChat/Utils/StreamCollection.swift
ipavlidakis: I'm not sure I understand the benefit of using closures here. Do you mind elaborating?

PR #3066, opened 2024-03-08
[ChatStateLayer] ChannelList and ChannelListState

laevandus:
### 🔗 Issue Links

### 🎯 Goal

Add ChannelList and ChannelListState for actions and state when loading lists of channels with `ChannelListQuery`

### 📝 Summary

* Adds actions and observable state support
* Does not include channel linking (will be a separate follow-up PR)
* Does not include tracking logic with SyncRepository (I have some ideas around it, will be a separate follow-up PR)

### 🛠 Implementation

ChannelListUpdater makes the queries. ChatClient provides the factory method for creating the ChannelList objects. ChannelListState holds the state. 

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/ChatClient.swift
+++ b/Sources/StreamChat/ChatClient.swift
@@ -74,6 +74,8 @@ public class ChatClient {
     let callRepository: CallRepository
 
     let channelRepository: ChannelRepository
+    
+    let channelListUpdater: ChannelListUpdater
 
     func makeMessagesPaginationStateHandler() -> MessagesPaginationStateHandling {
         MessagesPaginationStateHandler()
@@ -195,6 +197,10 @@ public class ChatClient {
         extensionLifecycle = environment.extensionLifecycleBuilder(config.applicationGroupIdentifier)
         callRepository = environment.callRepositoryBuilder(apiClient)
         channelRepository = environment.channelRepositoryBuilder(
+            databaseContainer,
+            apiClient
+        )
+        channelListUpdater = environment.channelListUpdaterBuilder(
             databaseContainer,
             apiClient
         )
--- a/Sources/StreamChat/StateLayer/ChannelList.swift
+++ b/Sources/StreamChat/StateLayer/ChannelList.swift
@@ -42,9 +42,7 @@ public struct ChannelList {
     /// - Returns: An array of channels for the pagination.
     @discardableResult public func loadChannels(with pagination: Pagination) async throws -> [ChatChannel] {
         guard pagination.pageSize > 0 else { return [] }
-        let channels = try await channelListUpdater.loadChannels(query: query, pagination: pagination)
-        await state.setLoadedAllChannels(channels.count < pagination.pageSize)
-        return channels
+        return try await channelListUpdater.loadChannels(query: query, pagination: pagination)
     }
     
     /// Loads more channels and updates ``ChatListState.channels``.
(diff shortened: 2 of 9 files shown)


Sources/StreamChat/ChatClient.swift
```diff
@@ -74,6 +74,8 @@ public class ChatClient {
     let callRepository: CallRepository
 
     let channelRepository: ChannelRepository
+    
+    let channelListUpdater: ChannelListUpdater
```
laevandus, line 78: Does not hold state internally, therefore reusable for all the ChannelLists.

Sources/StreamChat/StateLayer/ChannelList.swift
```diff
@@ -0,0 +1,76 @@
...
+            channels,
+            query,
+            client.config,
+            client.databaseContainer
+        )
+        self.state = state
+        
+        // These are currently not implemented compared to ChannelListController:
+        #warning("Implement query reset (e.g. SyncRepository)")
+        #warning("Implement linking and unlinking based on EventController callbacks")
```
laevandus, line 28: Separate PRs for reducing the scope

Sources/StreamChat/StateLayer/ChannelListState.swift
```diff
@@ -0,0 +1,45 @@
...
+        observer.start(
+            with: .init(channelsDidChange: { [weak self] channels in await self?.setValue(channels, for: \.channels) })
+        )
+    }
+    
+    /// An array of channels for the specified ``ChannelListQuery``.
+    @Published public private(set) var channels = [ChatChannel]()
+    
+    /// True, if all the channels were loaded with paginated loading.
+    @Published public private(set) var hasLoadedAllChannels = false
```
laevandus, line 30: @nuno-vieira was asking about jumping to mid page in [here](https://github.com/GetStream/stream-chat-swift/pull/3057/files#r1514736176). 
Since pagination for channels only supports offset/cursor based pagination then I think it is OK if we go for just one boolean state (although I could misuse the API when I start using the cursor based pagination).

For messages pagination we also provided an is loading property, not certain if we should do the same here. `ChatChannelListController` also did not have is loading property.

Sources/StreamChat/StateLayer/ChatClient+Factory.swift
```diff
@@ -195,23 +218,3 @@ extension ChatClient {
         )
     }
 }
-
-// MARK: - Compatibility Methods
-
-@available(iOS 13.0, *)
```
laevandus, line 201: Some hack I used for testing channels but does not really need to be here.

Sources/StreamChat/StateLayer/ChannelListState.swift
laevandus: We did discuss something similar with @martinmitrevski when adding watchers property and the decision was to skip such property for now. Makes sense if I remove it from here as well and let the UI side to handle that state instead.

Sources/StreamChat/StateLayer/ChannelList.swift
```diff
@@ -0,0 +1,76 @@
...
+    // MARK: - Channel List Pagination
+    
+    /// Loads channels for the specified pagination parameters and updates ``ChatListState.channels``.
+    ///
+    /// - Parameters:
+    ///   - pagination: The pagination configuration which includes limit and cursor.
+    ///
+    /// - Throws: An error while communicating with the Stream API.
+    /// - Returns: An array of channels for the pagination.
+    @discardableResult public func loadChannels(with pagination: Pagination) async throws -> [ChatChannel] {
```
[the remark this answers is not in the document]
martinmitrevski: Sounds good. Just one question: can this mess up the state if you also use `loadNextChannels`?

martinmitrevski (approved):

PR #3068, opened 2024-03-11
[ChatStateLayer] Channel list query reset in SyncRepository

laevandus:
### 🔗 Issue Links

*Merges to feature/chat-state-layer-channel-list-query*

### 🎯 Goal

Refetch channel list queries when syncing is triggered for the new chat state layer.

### 📝 Summary

SyncRepository is keeping track of active ChannelListController objects (see [#1](https://github.com/GetStream/stream-chat-swift/blob/develop/Sources/StreamChat/ChatClient.swift#L45), [#2](https://github.com/GetStream/stream-chat-swift/blob/develop/Sources/StreamChat/Repositories/SyncRepository.swift#L129), [#3](https://github.com/GetStream/stream-chat-swift/blob/develop/Sources/StreamChat/Repositories/SyncOperations.swift#L103)). New ChannelList and ChannelListState objects should support the same functionality (resetting ends up fetching channels for the query and updating CoreData store).

This PR proposes to do this differently in a less coupled fashion for the new state layer. 

### 🛠 Implementation

Proposal is to use a notification, which contains an object. If ChannelState observers the notification and adds the query to the shared object, then we could skip having references to ChannelList/ChannelListState types in SyncRepository.

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/ChatClient+Environment.swift
+++ b/Sources/StreamChat/ChatClient+Environment.swift
@@ -127,7 +127,8 @@ extension ChatClient {
             _ offlineRequestsRepository: OfflineRequestsRepository,
             _ eventNotificationCenter: EventNotificationCenter,
             _ database: DatabaseContainer,
-            _ apiClient: APIClient
+            _ apiClient: APIClient,
+            _ channelListUpater: ChannelListUpdater
         ) -> SyncRepository = {
             SyncRepository(
                 config: $0,
@@ -136,7 +137,8 @@ extension ChatClient {
                 offlineRequestsRepository: $3,
                 eventNotificationCenter: $4,
                 database: $5,
-                apiClient: $6
+                apiClient: $6,
+                channelListUpdater: $7
             )
         }
 
--- a/Sources/StreamChat/Repositories/SyncRepository.swift
+++ b/Sources/StreamChat/Repositories/SyncRepository.swift
@@ -138,8 +138,7 @@ class SyncRepository {
             }
         operations.append(contentsOf: refetchChannelListQueryOperations)
         let channelListRegistry = ChannelListRegistry()
-        let notification = Notification(name: SyncRepository.syncRepositoryChannelListQueryRegistrationNotification, object: channelListRegistry)
-        NotificationCenter.default.post(notification)
+        NotificationCenter.default.post(Notification(name: .syncRepositoryChannelListQueryRegistration, object: channelListRegistry))
         operations.append(contentsOf: channelListRegistry.registeredChannelListQueries.map { query in
             RefetchChannelListQueryOperation(query: query, channelListUpdater: channelListUpdater, context: context)
         })
@@ -364,7 +363,9 @@ extension SyncRepository {
             }
         }
     }
-    
-    /// A notification which contains ChannelListRegistry as the notification's object.
-    static let syncRepositoryChannelListQueryRegistrationNotification = Notification.Name("syncRepositoryChannelListQueryRegistrationNotification")
+}
+
+extension Notification.Name {
+    /// A notification which contains ``ChannelListRegistry`` as the notification's object.
+    static let syncRepositoryChannelListQueryRegistration = Notification.Name("io.getstream.StreamChat.channelListQueryRegistration")
 }
--- a/Sources/StreamChat/Repositories/SyncRepository.swift
+++ b/Sources/StreamChat/Repositories/SyncRepository.swift
@@ -350,20 +350,18 @@ private extension Date {
 extension SyncRepository {
     final class ChannelListRegistry {
         private var channelListQueries = [ChannelListQuery]()
-        private let lock = NSLock()
+        private let queue = DispatchQueue(label: "io.getstream.sync-repository.channel-list-registry")
         
         var registeredChannelListQueries: [ChannelListQuery] {
-            let queries: [ChannelListQuery]
-            lock.lock()
-            queries = channelListQueries
-            lock.unlock()
-            return queries
+            queue.sync {
+                channelListQueries
+            }
         }
         
         func register(query: ChannelListQuery) {
-            lock.lock()
-            channelListQueries.append(query)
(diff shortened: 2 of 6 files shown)


Sources/StreamChat/ChatClient+Environment.swift
```diff
@@ -127,7 +127,8 @@ extension ChatClient {
             _ offlineRequestsRepository: OfflineRequestsRepository,
             _ eventNotificationCenter: EventNotificationCenter,
             _ database: DatabaseContainer,
-            _ apiClient: APIClient
+            _ apiClient: APIClient,
+            _ channelListUpater: ChannelListUpdater
```
laevandus, line 131: Injecting `ChannelListUpdater` to `SyncRepository` which allows refetching channels for the query. Old way is leveraging ChannelListController directly.

Sources/StreamChat/Repositories/SyncRepository.swift
```diff
@@ -337,3 +346,27 @@ private extension Date {
         Calendar.current.dateComponents([.day], from: self, to: Date()).day ?? 0
     }
 }
+
+extension SyncRepository {
+    final class ChannelListRegistry {
+        private var channelListQueries = [ChannelListQuery]()
+        let lock = NSLock()
```
laevandus, line 353: NSRecursiveLock and DispatchQueue would allow shorter implementation.

Sources/StreamChat/Repositories/SyncRepository.swift
```diff
@@ -134,6 +137,12 @@ class SyncRepository {
                 )
             }
         operations.append(contentsOf: refetchChannelListQueryOperations)
+        let channelListRegistry = ChannelListRegistry()
+        let notification = Notification(name: SyncRepository.syncRepositoryChannelListQueryRegistrationNotification, object: channelListRegistry)
+        NotificationCenter.default.post(notification)
+        operations.append(contentsOf: channelListRegistry.registeredChannelListQueries.map { query in
+            RefetchChannelListQueryOperation(query: query, channelListUpdater: channelListUpdater, context: context)
+        })
```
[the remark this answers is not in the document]
martinmitrevski: hm, can you elaborate a bit more on the issue? What do you mean by "no receive(on:) when using Combine"?

Sources/StreamChat/Repositories/SyncRepository.swift
```diff
@@ -134,6 +137,12 @@ class SyncRepository {
                 )
             }
         operations.append(contentsOf: refetchChannelListQueryOperations)
```
martinmitrevski, line 139: would we still need this line?

laevandus: I did not touch the controller related implementation at the moment. If we go for notification center based approach, the controller related implementation can be cleaned up as well and we would end up deleting this part. That should be a separate small PR then.

Sources/StreamChat/Repositories/SyncRepository.swift
```diff
@@ -337,3 +346,27 @@ private extension Date {
...
 }
+
+extension SyncRepository {
+    final class ChannelListRegistry {
+        private var channelListQueries = [ChannelListQuery]()
+        private let lock = NSLock()
+        
+        var registeredChannelListQueries: [ChannelListQuery] {
+            let queries: [ChannelListQuery]
+            lock.lock()
```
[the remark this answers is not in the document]
laevandus: I will change it to use a queue. Locks are faster than queues but it is not a high performance related code path anyway. Will change it.

Sources/StreamChat/Repositories/SyncRepository.swift
laevandus: I will change it.

Sources/StreamChat/Repositories/SyncRepository.swift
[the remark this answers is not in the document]
laevandus: I was thinking about this flow:

SyncRepository posts a notification on a background thread:
```swift
NotificationCenter.default.post(notification)
```
and someone is using such code for receiving the notification
```swift
… = NotificationCenter.default
        .publisher(for: SyncRepository.syncRepositoryChannelListQueryRegistrationNotification)
        .compactMap { $0.object as? SyncRepository.ChannelListRegistry }
        .receive(on: DispatchQueue.main)
        .sink(receiveValue: { [query] registry in registry.register(query: query) })
```
Then, when `post(notification)` has returned, the `sink` has not run yet because it was scheduled to run on a different thread. Meaning the query has not been added to the registry class and when SyncRepository accesses registered queries, it would miss that query (timing issue). As long as observers run observing code synchronously, it would be OK and query is added to the query before `post()` returns. 

But since it is a internal implementation then this downside should be OK.

Sources/StreamChat/Repositories/SyncRepository.swift
laevandus: I will change it to use DispatchQueue.

Sources/StreamChat/Repositories/SyncRepository.swift
martinmitrevski: I think that would be a duplicated logic, right? Good to remove it, so that we are sure that things work well with the new approach.

Sources/StreamChat/Repositories/SyncRepository.swift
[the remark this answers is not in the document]
martinmitrevski: Alright, thanks for the explanation. Yes, probably better to use our wrapper for consistency.

martinmitrevski (approved):
LGTM ✅  Would be nice to address the comments before merging

Sources/StreamChat/Repositories/SyncRepository.swift
laevandus: I will create a separate follow-up PR for addressing it

PR #3071, opened 2024-03-11
[ChatStateLayer] Loading watchers for a channel

laevandus:
### 🔗 Issue Links

### 🎯 Goal

Support loading watchers for channels.

### 📝 Summary

* Add `Chat.loadWatchers(with:)` (argument is `Pagination`, aligns with other platforms what support general pagination method)
* Add `Chat.loadNextWatchers(with:)` (convenience which uses pagination with an offset)
* Add `ChatState.watchers`
* Add `ChatState.hasLoadedAllWatchers` (`ChatChannelWatcherListController` does not have it but added it so that out pagination implementations have similar interfaces)

### 🛠 Implementation

Follows along what we have in `ChatChannelWatcherListController` (pagination + watchers). 

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/Controllers/ChannelWatcherListController/ChatChannelWatcherListController.swift
+++ b/Sources/StreamChat/Controllers/ChannelWatcherListController/ChatChannelWatcherListController.swift
@@ -92,7 +92,8 @@ public class ChatChannelWatcherListController: DataController, DelegateCallable,
             return
         }
 
-        updater.channelWatchers(query: query) { error in
+        updater.channelWatchers(query: query) { result in
+            let error = result.error
             self.state = error == nil ? .remoteDataFetched : .remoteDataFetchFailed(ClientError(with: error))
             self.callback { completion?(error) }
         }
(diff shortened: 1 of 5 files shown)


Sources/StreamChat/Controllers/ChannelWatcherListController/ChatChannelWatcherListController.swift
```diff
@@ -92,7 +92,8 @@ public class ChatChannelWatcherListController: DataController, DelegateCallable,
             return
         }
 
-        updater.channelWatchers(query: query) { error in
+        updater.channelWatchers(query: query) { result in
```
laevandus, line 95: Async await wrapper needs to access the payload

Sources/StreamChat/StateLayer/ChatState.swift
```diff
@@ -96,6 +97,18 @@ public final class ChatState: ObservableObject {
...
+    
+    /// An array of users who are currently watching the channel.
+    ///
+    /// Use load watchers method in ``Chat`` for populating this array.
+    @Published public private(set) var watchers = StreamCollection<ChatUser>([])
+    
+    /// True, if all the watchers were loaded with paginated loading.
+    ///
+    /// The value is set to true when the pagination request returns less than the pagination page size.
+    @Published public private(set) var hasLoadedAllWatchers = false
```
laevandus, line 110: Debatable if we should provide this 🤔

laevandus: Let me take it away since I also felt like it was too much. If our UI code would benefit from this, we can always add it or add something similar on the UI level instead.

PR #3069, opened 2024-03-11
[ChatStateLayer] Channel list linking and unlinking

laevandus:
### 🔗 Issue Links

Related: [#728](https://github.com/GetStream/ios-issues-tracking/issues/728)

*Merges to feature/chat-state-layer-channel-list-query*

### 🎯 Goal

Add the linking and unlinking logic from ChannelListController to the new state layer.

### 📝 Summary

* Listen to events and trigger the same linking and unlinking code paths

### ☑️ Contributor Checklist

- [x] I have signed the [Stream CLA](https://docs.google.com/forms/d/e/1FAIpQLScFKsKkAJI7mhCr7K9rEIOpqIDThrWxuvxnwUq2XkHyG154vQ/viewform) (required)
- [x] This change follows zero ⚠️ policy (required)
- [ ] This change should be manually QAed
- [ ] Changelog is updated with client-facing changes
- [ ] New code is covered by unit tests
- [ ] Comparison screenshots added for visual changes
- [ ] Affected documentation updated (docusaurus, tutorial, CMS)

PR #3077, opened 2024-03-12
[ChatStateLayer] Use fetch request for channel list contains channel check

laevandus:
### 🔗 Issue Links

### 🎯 Goal

Use fetch request for checking the changed channel is part of the channel list query

### 📝 Summary

Fetch request avoids looping over the collection which is currently a `LazyCachedMapCollection` and looping would trigger loading each of the channel which can be expensive and wasteful.

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/StateLayer/ChannelListState+Observer.swift
+++ b/Sources/StreamChat/StateLayer/ChannelListState+Observer.swift
@@ -116,15 +116,24 @@ extension ChannelListState {
             return false
         }
         
+        private func isChannelInList(_ cid: ChannelId) async throws -> Bool {
+            try await database.backgroundRead { [query] context in
+                guard let (channelDTO, queryDTO) = context.getChannelWithQuery(cid: cid, query: query) else { return false }
+                return queryDTO.channels.contains(channelDTO)
+            }
+        }
+        
         private func linkChannelIfNeeded(_ channel: ChatChannel) async throws {
-            guard !channelListObserver.items.contains(where: { $0.cid == channel.cid }) else { return }
+            let listContainsChannel = try await isChannelInList(channel.cid)
+            guard !listContainsChannel else { return }
             guard isBelongingToChannelListQuery(channel: channel) else { return }
             try await channelListUpdater.link(channel: channel, with: query)
             try await channelListUpdater.startWatchingChannels(withIds: [channel.cid])
         }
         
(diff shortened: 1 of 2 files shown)


Sources/StreamChat/Database/DTOs/ChannelDTO.swift
```diff
@@ -349,7 +349,29 @@ extension ChannelDTO {
...
+    }
+    
+    static func channelListContainsChannelFetchRequest(
+        cid: ChannelId,
+        query: ChannelListQuery,
+        chatClientConfig: ChatClientConfig
+    ) -> NSFetchRequest<ChannelDTO> {
+        let request = channelListFetchRequest(query: query, chatClientConfig: chatClientConfig)
+        let channelPredicate = NSPredicate(format: "cid == %@", cid.rawValue)
+        let channelListPredicates = channelListPredicates(query: query, chatClientConfig: chatClientConfig)
```
laevandus, line 365: Diff is slightly hard to read but I extracted existing predicates into a separate function `channelListPredicate` which is used in two places. Here I add the cid equals check.

Sources/StreamChat/StateLayer/ChannelListState+Observer.swift
```diff
@@ -117,14 +117,20 @@ extension ChannelListState {
         }
         
         private func linkChannelIfNeeded(_ channel: ChatChannel) async throws {
-            guard !channelListObserver.items.contains(where: { $0.cid == channel.cid }) else { return }
```
laevandus, line 120: The reasoning is that looping `LazyCachedMapCollection` triggers loading each DTO separately which can get expensive and loses the point of using cached collection.
Related to PR comment [here](https://github.com/GetStream/stream-chat-swift/pull/3069#discussion_r1521072803)

martinmitrevski (commented):
can't we use the existing `dataStore.channel(cid: ChannelId)` instead?

laevandus:
> can't we use the existing `dataStore.channel(cid: ChannelId)` instead?

Since channel lists are tied to `ChannelListQueryDTO` then in this case the linking logic needs to follow that. But that said, I just found `session.getChannelWithQuery(cid: channel.cid, query: query)` which seems to be doing what I need.

martinmitrevski (approved):

PR #3079, opened 2024-03-13
[ChatStateLayer] Member banning and member lists

laevandus:
### 🔗 Issue Links

### 🎯 Goal

Add functionality in ChatChannelMemberController to the chat state layer.

### 📝 Summary

* Add `banUser`, `shadowBanUser`, `unbanUser`
* Add `MemberState` which is an observable object

### ☑️ Contributor Checklist


Sources/StreamChat/ChatClient.swift
```diff
@@ -76,6 +76,10 @@ public class ChatClient {
     let channelRepository: ChannelRepository
     
     let channelListUpdater: ChannelListUpdater
+    
+    let memberUpdater: ChannelMemberUpdater
+    
+    let memberListUpdater: ChannelMemberListUpdater
```
laevandus, line 82: These should be converted to MemberRepository in the future. Also captured in Notion.

Sources/StreamChat/StateLayer/Chat.swift
```diff
@@ -606,6 +624,57 @@ public final class Chat {
...
+    ///  - Note: Channel watchers cannot be banned.
+    ///  - Note: In most cases, only admins or moderators are allowed to ban other users from the channel.
+    ///
+    /// - Parameters:
+    ///   - userId: The user id of the channel member.
+    ///   - reason: The reason that the ban was created.
+    ///   - timeoutInMinutes: The number of minutes the user should be banned for. Nil means that the user is banned forever or until the user is unbanned explicitly.
+    ///
+    /// - Throws: An error while communicating with the Stream API or missing required capabilities.
+    public func banUser(_ userId: UserId, reason: String? = nil, timeout timeoutInMinutes: Int? = nil) async throws {
```
laevandus, line 645: All the other platforms use the term `User` here. Therefore, I added `banUser` and not `banMember`.

Sources/StreamChat/Workers/ChannelMemberListUpdater.swift
```diff
@@ -11,32 +11,59 @@ class ChannelMemberListUpdater: Worker {
     ///   - query: The query used in the request.
     ///   - completion: Called when the API call is finished. Called with `Error` if the remote update fails.
     func load(_ query: ChannelMemberListQuery, completion: ((Error?) -> Void)? = nil) {
+        load(query, includeMembersResponse: false, completion: { result in
+            completion?(result.error)
+        })
+    }
+    
+    private func load(_ query: ChannelMemberListQuery, includeMembersResponse: Bool, completion: ((Result<[ChatChannelMember], Error>) -> Void)? = nil) {
```
laevandus, line 19: The current load method only returned error, I wanted to return the update ChatChannelMember so that I can use it for MemberState. Therefore, I added a private method with a boolean for not requiring existing code to initialise ChatChannelMember when it does not need it.

Sources/StreamChat/ChatClient.swift
```diff
@@ -206,6 +210,14 @@ public class ChatClient {
             databaseContainer,
             apiClient
         )
+        memberUpdater = environment.memberUpdaterBuilder(
```
martinmitrevski, line 213: Feels a bit strange to have member updaters in the chat client. Although they don't depend on anything `Chat` specific, I believe we can instantiate them on demand - no need to move them here. Wdyt?

Sources/StreamChat/StateLayer/Chat.swift
martinmitrevski: It depends on what is the scope of the ban - if it's only the channel, then a member is banned. If the user is banned throughout the chats, then user should be used. I think in this case member is correct.

Sources/StreamChat/StateLayer/MemberState.swift
```diff
@@ -0,0 +1,73 @@
...
+//
+
+import Foundation
+
+/// An observable object for the specified channel member.
+///
+/// - Note: All the properties of the represented ``ChatChannelMember`` can be accessed directly on the ``MemberState`` object (e.g. `state.name` returns ``ChatChannelMember.name``).
+@available(iOS 13.0, *)
+@dynamicMemberLookup
+public final class MemberState: ObservableObject {
```
martinmitrevski, line 12: Why do we need this one?

Sources/StreamChat/StateLayer/Chat.swift
```diff
@@ -168,6 +172,20 @@ public final class Chat {
...
+    // MARK: Member State Observing
+    
+    /// Returns an observable member state for the specified user id.
+    ///
+    /// The member state is refreshed before returning the observable state.
+    ///
+    /// - Parameter userId: The user id of the channel member.
+    ///
+    /// - Returns: An instance of `MemberState` which conforms to the `ObservableObject`.
+    public func makeMemberState(for userId: UserId) async throws -> MemberState {
```
martinmitrevski, line 184: how would this work? You need to explicitly call this to listen to a specific member? I would expect to listen to a list of members and their updates in the chat state.

laevandus: That is a good point, it is kind of pointless. I added it since we had MemberController. Let me get rid of it. I have another PR for member list.

Sources/StreamChat/ChatClient.swift
laevandus: I'll get rid of these. These should be converted to a repository class anyway at some point.

Sources/StreamChat/StateLayer/Chat.swift
laevandus: Changed it to "Member"

Sources/StreamChat/StateLayer/MemberState.swift
laevandus: On the second thought, not really needed, removed it.

Sources/StreamChat/StateLayer/Chat.swift
```diff
@@ -168,6 +178,89 @@ public final class Chat {
...
+    ///
+    /// Calling ``Chat.loadMembers(with:sort:pagination:)`` automatically updates ``MemberListState.members`` for the same query.
+    ///
+    /// - Parameters:
+    ///   - filter: The filter to use for fetching channel members.
+    ///   - sort: An array of sorting parameters.
+    ///   - pagination: The pagination configuration which includes a limit and an offset or a cursor.
+    ///
+    /// - Returns: An instance of `MemberListState` which conforms to the `ObservableObject`.
+    public func makeMemberListState(for filter: Filter<MemberListFilterScope>, sort: [Sorting<ChannelMemberListSortingKey>] = [], pagination: Pagination) async throws -> MemberListState {
```
martinmitrevski, line 206: Another idea: why don't we simplify even further - this can be a member list array in the chat state. It also fits conceptually, removes some boilerplate code and it's actually easier for customers. Wdyt?

Sources/StreamChat/Workers/ChannelMemberListUpdater.swift
nuno-vieira: @laevandus I don't it is necessary to maintain both functions. From now one, we should always prefer to return the responses 👍

Sources/StreamChat/StateLayer/Chat.swift
nuno-vieira: @martinmitrevski I don't think that will work because you might have multiple members list, and not just one. What I think it makes sense is to move this to the `ChatClient` level, and create the member list state there. AFAIK you could create a member list query here that has nothing to do with this Chat. Since the customer can query a member list of any channel, and not only this one, I think it makes sense to move this one level higher

Sources/StreamChat/StateLayer/Chat.swift
```diff
@@ -168,6 +178,89 @@ public final class Chat {
...
+    ///   - sort: An array of sorting parameters.
+    ///   - pagination: The pagination configuration which includes a limit and an offset or a cursor.
+    ///
+    /// - Throws: An error while communicating with the Stream API.
+    /// - Returns: An array of paginated channel members for the query.
+    public func loadMembers(with filter: Filter<MemberListFilterScope>, sort: [Sorting<ChannelMemberListSortingKey>] = [], pagination: Pagination) async throws -> [ChatChannelMember] {
+        var query = ChannelMemberListQuery(cid: cid, filter: filter, sort: sort)
+        query.pagination = pagination
+        return try await memberListUpdater.load(query)
+    }
```
nuno-vieira, line 194: I think this one makes sense to be here in the Chat object, and maybe for this one, we can follow what @martinmitrevski said, and have a `members` property in the state.

nuno-vieira (commented):
Overall looks good, although I think we should distinguish the `Chat.loadMembers` vs `ChatClient.makeMemberList()` since the latter should be more generic and allow loading the members of any channel the customer provides, so it should not be dependent on `Chat`.

Sources/StreamChat/StateLayer/Chat.swift
martinmitrevski: Those are good points, but we need to think this through a bit more. I mean the same thing could be said about e.g. watchers as well - you might want to access them without creating a chat. But on the other hand, that will put too much responsibility to the client. 

The question is - how expensive would it be to create a chat, since naturally/logically the members should be part of the chat.
The other question - how frequently would customers want to do that? If we optimize for a small chunk, it's probably not worth it.

Sources/StreamChat/Workers/ChannelMemberListUpdater.swift
laevandus: Yes, I will change this like I did in the other PR where you suggested it. Thanks!

Sources/StreamChat/StateLayer/Chat.swift
laevandus: I'll think about all the things said here. It kind of sounds nice if `ChatState` has `members` but at the same time there is still a way to create member lists by not creating the Chat instance.

Sources/StreamChat/StateLayer/Chat.swift
laevandus: Thanks, I'll try out a hybrid approach where `ChatState` has members and there is still a way to create member lists without `Chat`.

Sources/StreamChat/StateLayer/Chat.swift
```diff
@@ -168,6 +179,70 @@ public final class Chat {
...
+    /// - Note: Channel member sorting keys are set when creating the ``Chat`` instance.
+    /// It is also possible to create separate ``MemberList`` objects if needed with different filtering options. See ``ChatClient.makeMemberList(with:)``.
+    ///
+    /// - Parameter pagination: The pagination configuration which includes a limit and an offset or a cursor.
+    ///
+    /// - Throws: An error while communicating with the Stream API.
+    /// - Returns: An array of paginated channel members for the query.
+    public func loadMembers(with pagination: Pagination) async throws -> [ChatChannelMember] {
+        try await memberList.loadMembers(with: pagination)
+    }
```
laevandus, line 193: I tried out what we discussed which is Chat having a way to have channel members as part of the state but still having a way to create separate MemberLists. My proposal is that the `Chat` one only supports setting sorting options (like messages) but if one creates a separate MemberList, then all the filtering and sorting is available.

Sources/StreamChat/StateLayer/Chat.swift
laevandus: It is implemented now

Sources/StreamChat/StateLayer/Chat.swift
laevandus: This is implemented

Sources/StreamChat/StateLayer/Chat.swift
```diff
@@ -170,6 +182,70 @@ public final class Chat {
...
+    ///  - Note: In most cases, only admins or moderators are allowed to ban other users from the channel.
+    ///
+    /// - Parameters:
+    ///   - userId: The user id of the channel member.
+    ///   - reason: The reason that the ban was created.
+    ///   - timeoutInMinutes: The number of minutes the user should be banned for. Nil means that the user is banned forever or until the user is unbanned explicitly.
+    ///
+    /// - Throws: An error while communicating with the Stream API or missing required capabilities.
+    public func banMember(_ userId: UserId, reason: String? = nil, timeout timeoutInMinutes: Int? = nil) async throws {
+        try state.channel?.requireCapability(of: .banChannelMembers)
```
nuno-vieira, line 217: I'm not sure this should be part of the LLC SDK 🤔 This is more like a UI level flag, and not for LLC, since the backend will respond with a failure if the capability is not there 🤔 This was not in the previous request right?

Sources/StreamChat/StateLayer/Chat.swift
```diff
@@ -170,6 +182,70 @@ public final class Chat {
...
+    /// Learn more about [shadow banning and moderation tools](https://getstream.io/chat/docs/ios-swift/moderation/?language=swift#shadow-ban).
+    ///
+    /// - Parameters:
+    ///   - userId: The user id of the channel member.
+    ///   - reason: The reason that the ban was created.
+    ///   - timeoutInMinutes: The number of minutes the user should be banned for. Nil means that the user is banned forever or until the user is unbanned explicitly.
+    ///
+    /// - Throws: An error while communicating with the Stream API or missing required capabilities.
+    public func shadowBanMember(_ userId: UserId, reason: String? = nil, timeout timeoutInMinutes: Int? = nil) async throws {
+        try state.channel?.requireCapability(of: .banChannelMembers)
```
nuno-vieira, line 235: Again, these require capibility should not be part of the actual request, but meant to be used in the UI only

Sources/StreamChat/StateLayer/Chat.swift
laevandus: It is in a bit mixed state. Sometime we have been checking for capabilities before. I guess, better not to do it on the client level and let the request to fail. I'll remove it (it was not there for controllers).

Sources/StreamChat/StateLayer/Chat.swift
laevandus: Removed

laevandus:
> Overall looks good, although I think we should distinguish the `Chat.loadMembers` vs `ChatClient.makeMemberList()` since the latter should be more generic and allow loading the members of any channel the customer provides, so it should not be dependent on `Chat`.

This is also addressed. Now we have separate member lists (filtering, sorting) and also Chat+ChatState having a way to load all the members.

Sources/StreamChat/StateLayer/Chat.swift
martinmitrevski: just seeing this. @nuno-vieira why do we need this only in the UI? We know that we don't have the capability, so we can just throw the error. If we don't, we put too much responsibility on the caller - which doesn't necessarily are our UI components, and customers can forget to do the capability check.

martinmitrevski: yes, just write it down as an open issue and we will discuss is on Thursday, feel free to merge now.

martinmitrevski (approved):

Sources/StreamChat/StateLayer/Chat.swift
nuno-vieira: The thing is that if the customer forgets to check the capability, the server will Error either way, so it is kind the same thing, although we save 1 request 🤔 The reason I thought that it would be better to skip this check in the LLC is because of the following pros/cons:

Pros of checking capibility in the LLC:
- 1 less request
Cons of checking capibility in the LLC:
1- Maintanance (We need to do this for every action)
2- Error-prone (We can forget to add this for some othe actions)
3- There could be a scenario where the capiblity is updated on the server, and we might block a valid request

nuno-vieira (approved):
LGTM! ✅

Sources/StreamChat/StateLayer/Chat.swift
laevandus: Added it as a discussion topic for our next meeting

laevandus: Discussion result: let's not check capabilities

Issue #2807, opened 2023-09-27
Setting a Mute Duration 


miacando:
I noticed in the [documentations for other languages](https://getstream.io/chat/docs/react/muting_channels/?language=javascript) that they have a parameter for optionally setting an expiration time, but none for Swift. How do I do it in Swift?

laevandus:
Hi @miacando, we have released a new version 4.51.0 which addresses this request. Please refer to the `muteChannel()` function and its `expiration` argument.

PR #3083, opened 2024-03-13
Add mute expiration support when muting a channel

testableapple:
### 🔗 Issue Links

Resolves https://github.com/GetStream/stream-chat-swift/issues/2807
Relates https://github.com/GetStream/ios-issues-tracking/issues/762

### 🎯 Goal

Add mute expiration support when muting a channel

### 🧪 Manual Testing Notes

- Open a ChannelList
- Swipe a channel
- Tap the Debug Button
- Tap "Mute channel with expiration" button
- Set up an expiration duration (e.g.: 10000)
- Write a message to this channel from a different account
- Unread count bubble does not appear
- Wait expiration duration
- Write one more message
- Unread count bubble should appear

### ☑️ Contributor Checklist

### 🎁 Meme

[!gif](https://media.giphy.com/media/v1.Y2lkPTc5MGI3NjExa2lxcGIzMnJlMWZudzUxZmRscDRtN2x2bGo0cjhpZmFudml1cnlzMCZlcD12MV9pbnRlcm5hbF9naWZfYnlfaWQmY3Q9Zw/swtiK9jRfE0zS/giphy.gif)


laevandus (approved):
I have nothing to add, looks nice

martinmitrevski (approved):
Amazing work 👏

DemoApp/StreamChat/Components/DemoChatChannelListRouter.swift
```diff
@@ -298,6 +298,19 @@ final class DemoChatChannelListRouter: ChatChannelListRouter {
                     }
                 }
             }),
+            .init(title: "Mute channel with expiration", isEnabled: canMuteChannel, handler: { [unowned self] _ in
+                self.rootViewController.presentAlert(title: "Enter expiration", textFieldPlaceholder: "Milliseconds") { expiration in
```
ipavlidakis, line 302: nit: It may be better to use seconds as they are more human readable format for time

Sources/StreamChat/APIClient/Endpoints/ChannelEndpoints.swift
```diff
@@ -58,14 +58,20 @@ extension Endpoint {
             body: body
         )
     }
-
-    static func muteChannel(cid: ChannelId, mute: Bool) -> Endpoint<EmptyResponse> {
-        .init(
+    
+    static func muteChannel(cid: ChannelId, mute: Bool, expiration: Int? = nil) -> Endpoint<EmptyResponse> {
+        var body: [String: AnyEncodable] = ["channel_cid": AnyEncodable(cid)]
```
ipavlidakis, line 63: ChannelId is already Codable. i don't think there is a need for the wrap to AnyEncodable.

Tests/StreamChatTests/APIClient/Endpoints/ChannelEndpoints_Tests.swift
```diff
@@ -212,6 +212,35 @@ final class ChannelEndpoints_Tests: XCTestCase {
             XCTAssertEqual(mute ? "moderation/mute/channel" : "moderation/unmute/channel", endpoint.path.value)
         }
     }
+    
+    func test_muteChannelWithExpiration_buildsCorrectly() {
```
ipavlidakis, line 216: Personally, i prefer to have seperate test methods for each case. In that way test results are clear when they are failing. 

One way you can approach that is to create a private assert method and pass the expected result. Based on the result you passed in the assert will verify true or false. Then you can have two test cases one for failing one for succeeding

DemoApp/StreamChat/Components/DemoChatChannelListRouter.swift
testableapple: Makes sense, updated 👍

laevandus:
I tried mute/unmute with expiration and without expiration. Worked correctly. But to be fair, millisecond alert panel in the test app looks slightly odd.

Sources/StreamChat/APIClient/Endpoints/ChannelEndpoints.swift
testableapple: Hmm, you sure? 🤔
Screenshot 2024-03-14 at 11 23 13 AM

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
```diff
@@ -306,17 +306,19 @@ public class ChatChannelController: DataController, DelegateCallable, DataStoreP
...
     ///
-    /// - Parameter completion: The completion. Will be called on a **callbackQueue** when the network request is finished.
-    ///                         If request fails, the completion will be called with an error.
+    /// - Parameters:
+    ///   - expiration: Duration of mute in milliseconds.
+    ///   - completion: The completion. Will be called on a **callbackQueue** when the network request is finished.
+    ///                 If request fails, the completion will be called with an error.
     ///
-    public func muteChannel(completion: ((Error?) -> Void)? = nil) {
+    public func muteChannel(expiration: Int? = nil, completion: ((Error?) -> Void)? = nil) {
```
nuno-vieira, line 314: All our expirations are in seconds (TimeInterval). I think we really should try to keep it consitant with the rest of the SDK as well as adhering to iOS guidelines. Example, Slow Mode, Pin Expiration, are all in seconds and use TimeInterval as the type

nuno-vieira: Actually, in slow mode, the cooldown is an `Int` as well, but it is seconds. 😞 Ideally, all time-related stuff should be `TimeInterval` which represents seconds.

testableapple: The problem is that according to our API, both [Pin Expiration](https://getstream.io/chat/docs/react/pinned_messages/?language=javascript) and [Slow Mode](https://getstream.io/chat/docs/react/slow_mode/?language=javascript&q=slow%20mode) use `seconds` when [Mute Expiration](https://getstream.io/chat/docs/react/muting_channels/?language=javascript) uses `milliseconds`. I agree that it's not convenient, but I still think that on this, we should align with our backend and other SDKs that leverage milliseconds in this case.

nuno-vieira: I can see that Android also uses milliseconds. Then lets keep it like this 👍

Tests/StreamChatTests/APIClient/Endpoints/ChannelEndpoints_Tests.swift
testableapple: Tbh I just copy-pasted `test_muteChannel_buildsCorrectly` to `test_muteChannelWithExpiration_buildsCorrectly` and added an expiration. Not sure it makes sense to rewrite both of them, also considering this feature will be replaced by OpenAPI, but let me know what you think.

Sources/StreamChat/APIClient/Endpoints/ChannelEndpoints.swift
ipavlidakis: Yeah this is because the expiration and the channelId have different types so the dictionary doesn't know how to setup. You could use `any Encodable` potentially (which is Swift's way to solve the issue that our AnyEncodable tries to fix) but that may require further changes. Sorry for confusing you here. You can ignore me.

Tests/StreamChatTests/APIClient/Endpoints/ChannelEndpoints_Tests.swift
ipavlidakis: Definitely not a reason there. Thanks for clarifying it :)

ipavlidakis (approved):
Great work! 🚢

PR #3086, opened 2024-03-14
[ChatStateLayer] ConnectedUser and ConnectedUserState

laevandus:
### 🔗 Issue Links

### 🎯 Goal

Add `CurrentChatUserController` functionality to the new state layer (devices, observing the current user's data)

### 📝 Summary

* `CurrentUser`: loadDevices, addDevice, removeDevice, markAllChannelsRead
* `CurrentUserState`: CurrentUserDTO changes

### ☑️ Contributor Checklist


Sources/StreamChat/Database/DTOs/CurrentUserDTO.swift
```diff
@@ -66,6 +66,11 @@ extension CurrentUserDTO {
         let new = NSEntityDescription.insertNewObject(into: context, for: request)
         return new
     }
+    
+    static func load(context: NSManagedObjectContext) throws -> CurrentChatUser {
+        guard let dto = load(context: context) else { throw ClientError.CurrentUserDoesNotExist() }
+        return try dto.asModel()
+    }
```
laevandus, line 73: Just found about `try? database.viewContext.currentUser?.asModel()`

Sources/StreamChat/StateLayer/CurrentUser.swift
```diff
@@ -0,0 +1,105 @@
...
+    ///
+    /// - Note: Setting any arguments to nil will keep the existing value.
+    ///
+    /// - Parameters:
+    ///   - name: The name to be set to the user.
+    ///   - imageURL: The URL of the avatar image.
+    ///   - extraData: Additional data associated with the user.
+    ///
+    /// - Throws: An error while communicating with the Stream API or when user is not logged in.
+    public func updateUserData(name: String? = nil, imageURL: URL? = nil, extraData: [String: RawJSON] = [:]) async throws {
```
nuno-vieira, line 38: The previous name is a bit redudant, maybe it is a good opportunity to change it?
```suggestion
    public func update(name: String? = nil, imageURL: URL? = nil, extraData: [String: RawJSON] = [:]) async throws {
```

Sources/StreamChat/StateLayer/CurrentUser.swift
```diff
@@ -0,0 +1,105 @@
...
+    /// Unregistering a device removes the device from the user and stops further new message notifications.
+    ///
+    /// - Parameter deviceId: The id of the device to unregister from push notifications.
+    ///
+    /// - Throws: An error while communicating with the Stream API or when user is not logged in.
+    public func removeDevice(_ deviceId: DeviceId) async throws {
+        try await currentUserUpdater.removeDevice(id: deviceId, currentUserId: try currentUserId())
+    }
+    
+    // MARK: -
```
nuno-vieira, line 87: Typo?
```suggestion
    // MARK: - Helpers
```

Sources/StreamChat/StateLayer/CurrentUserState.swift
```diff
@@ -0,0 +1,36 @@
...
+    /// Provides data about the represented user.
+    public subscript<T>(dynamicMember keyPath: KeyPath<CurrentChatUser, T>) -> T {
+        user[keyPath: keyPath]
+    }
+    
+    // MARK: - Mutating the State
+    
+    @MainActor func setValue<Value>(_ value: Value, for keyPath: ReferenceWritableKeyPath<CurrentUserState, Value>) {
+        self[keyPath: keyPath] = value
+    }
```
nuno-vieira, line 35: I'm not sure I fully understood why do we need this 🤔 Why can't we simple update the state without the keypaths?

Sources/StreamChat/Workers/CurrentUserUpdater.swift
```diff
@@ -119,16 +119,30 @@ class CurrentUserUpdater: Worker {
     ///     - currentUserId: The current user identifier.
     ///     - completion: Called when request is successfully completed, or with error.
     func fetchDevices(currentUserId: UserId, completion: ((Error?) -> Void)? = nil) {
+        fetchDevices(currentUserId: currentUserId, includeResponse: false, completion: { completion?($0.error) })
+    }
+    
+    private func fetchDevices(currentUserId: UserId, includeResponse: Bool, completion: ((Result<[Device], Error>) -> Void)? = nil) {
```
nuno-vieira, line 125: Why don't we always include the device response? Since this is a new major version, we can change this. IMO we should change it. From now on, all responses should return the actual payload 🤔 
```suggestion
    private func fetchDevices(currentUserId: UserId, completion: ((Result<[Device], Error>) -> Void)? = nil) {
```

nuno-vieira (commented):
Added a couple of comments, but overall looks good 👍

Sources/StreamChat/Workers/CurrentUserUpdater.swift
martinmitrevski: +1 on this

Sources/StreamChat/StateLayer/CurrentUser.swift
laevandus: Great suggestion, also aligns with what `Chat` has for updating channel data.
```swift
// Chat.swift
public func update(
        name: String?,
        imageURL: URL?,
        team: String?,
        members: Set<UserId> = [],
        invites: Set<UserId> = [],
        extraData: [String: RawJSON] = [:]
    ) async throws {
```

Sources/StreamChat/StateLayer/CurrentUser.swift
laevandus: Actually it wasn't, just used it for creating a line without text. It was kind of `// MARK: - Private` or helper like you suggested. Does not hurt adding a text to the separator as well.

Sources/StreamChat/StateLayer/CurrentUserState.swift
laevandus: It is used for forcing mutations on the main thread because it is an ObservableObject and therefore highly likely it is used by SDK consumers directly in SwiftUI views. In that case we do not want to cause runtime issues for them that something is triggering SwiftUI view update from a background thread.

When I was adding this, I also considered:

**Option 1**
Make `XYZState: ObservableObject` to require `@MainActor`: `@MainActor XYZState: ObservableObject`
Pro: Everything is main actor, therefore mutation is always happen on the main actor
Con: Harder to use this state type since existing code needs to adopt MainActor as well (our SwiftUI view models do not use `@MainActor`)

**Option 2**
Add a convenience @MainActor forced key-path based function for mutating stuff. 
Pro: Easy to adopt the state object since its public interface does not require @MainActor isolation
Con: It is not guarded against data corruption. Think about background threads reading public properties and at the same time main thread mutating the state. 

To be fair, we never had a real discussion between these options. Do you happen to have a preference on this matter? Other State objects currently follow the same pattern. If we decide to change it, it should be done for all the `XYZState` objects. I am happy to discuss this.

Sources/StreamChat/Workers/CurrentUserUpdater.swift
laevandus: I can change it on the updater level, but not breaking the existing public interface since it is used by controller

Sources/StreamChat/Database/DTOs/CurrentUserDTO.swift
laevandus: After reviewing the caching part, I'll better keep this as is.

Sources/StreamChat/StateLayer/CurrentUserState.swift
laevandus: Slack discussion was concluded to not expose main actor on the public interface level. Therefore, let's just use it for internal implementation and forcing mutation on the main thread.

Sources/StreamChat/Workers/CurrentUserUpdater.swift
laevandus: Changed it on the updater level. Done.

Sources/StreamChat/StateLayer/CurrentUser.swift
```diff
@@ -0,0 +1,105 @@
+//
+// Copyright © 2024 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+/// An object which represents the currently logged in user.
+@available(iOS 13.0, *)
+public struct CurrentUser {
```
martinmitrevski, line 9: We also have the model `CurrentChatUser` - this would be a bid confusing.

laevandus: Yeah, that is a slight problem with all of this. Need to think if I can come up with a new name.

laevandus: The best I could come up with is `ConnectedUser` since it is related to the `ChatClient.connectUser()`

nuno-vieira: I think it is a good idea 👍

Sources/StreamChat/StateLayer/ChatClient+Factory.swift
```diff
@@ -4,7 +4,20 @@
...
+
+@available(iOS 13.0, *)
+extension ChatClient {
+    /// Creates an instance of ``ConnectedUser`` which represents the logged-in user state and its actions.
+    ///
+    /// - Throws: An error if no user is currently logged-in.
+    func makeConnectedUser() async throws -> ConnectedUser {
+        let user = try await databaseContainer.backgroundRead { try CurrentUserDTO.load(context: $0) }
+        return ConnectedUser(user: user, client: self)
+    }
```
nuno-vieira, line 17: Could we also maybe return a "ConnectedUser" once the connectUser call is finished? 🤔

nuno-vieira (approved):
LGTM! ✅ Just added a suggestion, but not blocking for sure 👍

Sources/StreamChat/StateLayer/ChatClient+Factory.swift
nuno-vieira: Btw, instead of exposing the error from `try CurrentUserDTO.load(context: $0)`, it is probably better to expose a clearer error that the user is not connected, and so there is not current user available 🤔

laevandus: Returning `ConnectedUser` from the `connectUser` call makes a lot of sense. The existing connect user method needs to stay the same, therefore, I could add a new async one which does that. I'll try it out but maybe a separate PR for that.

Load error thingy makes sense, I'll change that.

laevandus: I ended up adding it to this PR. I left the `makeConnectedUser` available as well although I have mixed feelings about it. Maybe it would be better to hide it and to require users to get it through the connect call. But then, it gives extra flexibility for users.
There is also a question about logout handling but I want to think about that in a separate PR. Added a note to my TODO list.

Sources/StreamChat/ChatClient.swift
```diff
@@ -284,6 +284,24 @@ public class ChatClient {
...
+    /// Connects the client with the given user.
+    ///
+    /// - Parameters:
+    ///   - userInfo: The user info passed to `connect` endpoint.
+    ///   - tokenProvider: The closure used to retreive a token. Token provider will be used to establish the initial connection and also to obtain the new token when the previous one expires.
+    ///
+    /// - Throws: An error while communicating with the Stream API.
+    /// - Returns: A type representing the connected user and its state.
+    @available(iOS 13.0, *)
+    public func connectUser(userInfo: UserInfo, tokenProvider: @escaping TokenProvider) async throws -> ConnectedUser {
```
nuno-vieira, line 297: The return should probably be discardable here, since customers might not want to do anything with it when connecting

PR #3087, opened 2024-03-14
[ChatStateLayer] Local message lookup in chat

laevandus:
### 🔗 Issue Links

### 📝 Summary

Controllers conform to `DataStoreProvider` which requires that controller has `ChatClient` getter. New state layer does not capture `ChatClient` instances. Some of our UI code need these quick lookup methods the `DataStore` type defines.

Since the `DataStore` type itself if tied to `DatabaseContainer` and we are considering the possibility of not using CD as the only data source then it feels like it is better if the new layer does not expose `DataStore` directly. Maybe we just want to use in-memory array as a data source, then using this type makes it more complicated (I would need to inject data to it).

Therefore, proposing to just adding lookup methods on the new state layer level where needed.  

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/StateLayer/ChatState.swift
+++ b/Sources/StreamChat/StateLayer/ChatState.swift
@@ -10,12 +10,14 @@ public final class ChatState: ObservableObject {
     private let authenticationRepository: AuthenticationRepository
     private let cid: ChannelId
     private var channelObserver: EntityDatabaseObserverWrapper<ChatChannel, ChannelDTO>?
+    private let dataStore: DataStore
     private let paginationState: MessagesPaginationState
     private let observer: Observer
     
     init(cid: ChannelId, channelQuery: ChannelQuery, messageOrder: MessageOrdering, authenticationRepository: AuthenticationRepository, database: DatabaseContainer, eventNotificationCenter: EventNotificationCenter, paginationState: MessagesPaginationState) {
         self.authenticationRepository = authenticationRepository
         self.cid = cid
+        dataStore = DataStore(database: database)
         self.messageOrder = messageOrder
         self.paginationState = paginationState
         observer = Observer(cid: cid, channelQuery: channelQuery, messageOrder: messageOrder, database: database, eventNotificationCenter: eventNotificationCenter)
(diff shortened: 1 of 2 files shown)


Sources/StreamChat/StateLayer/ChatState.swift
```diff
@@ -47,6 +49,17 @@ public final class ChatState: ObservableObject {
...
+    /// Access a message available locally by its id.
+    ///
+    /// - Note: This method does a local lookup of the message.
+    ///
+    /// - Parameter messageId: The id of the message available locally.
+    ///
+    /// - Returns: An instance of the locally available chat message
+    public func message(for messageId: MessageId) -> ChatMessage? {
+        dataStore.message(id: messageId)
+    }
```
nuno-vieira, line 61: Can we just get the message in the local array for now? 🤔 Why do we need this again, I did not fully understood?

Most of our customers are not using this `DataStoreProvider`, and ideally it would be nice to deprecate it. 

If we do want to expose a way for customers to query data in the local db only, it would be nice to createa new entity for that, maybe `MessagesStore` or something like that. Exposing local db only methods in the State Layer can be confusing IMO 🤔

martinmitrevski: This is used in the logic around jumping to a message in SwiftUI, maybe also in UIKit. In general, would be nice to have a way to access local message.
I don't think we need a new entity for that. But I agree with this "Exposing local db-only methods in the State Layer can be confusing IMO". My suggestion is to update the method to sth like `localMessage` or `cachedMessage`.

laevandus: The general summary of the DataStore is:
UIKit
- [message] jump to message 
- [current user] creating message actions
- [channel] message search

SwiftUI
- [message] jump to message

We use it pretty rarely which is nice. The less we expose it the better for future.

Sounds like the options are:
a) Rename it to `localMessage` for emphasising that it only looks up locally available messages (It is still slightly confusing since it currently looks up any message, not messages part of the channel. If we later move to non FRC driven message pagination and use API responses directly within ChatMessage, then this method would change to be channel specific internally. All in all, I think it is fine. I could add a console warning if we want to limit the current function to channel specific lookup for future sake).

`
let message = dataStore.message(id: messageId)
if message.cid != cid {
  log.error("Trying to access message not part of the channel")
  return nil
}
return message
`
Just to guard against misuse.

b) UI components take the hit and do the lookup using the `.messages` which internally will trigger MessageDTO>ChatMessage conversions since it is a LazyCachedMapCollection. That is not so great.

martinmitrevski: yes, b) is the best option

martinmitrevski (approved):

PR #3091, opened 2024-03-15
Add `ChatClient.loadAppSettings` and `ChatClient.appSettings`

nuno-vieira:
### 🔗 Issue Links
Relates to https://github.com/GetStream/ios-issues-tracking/issues/757

### 🎯 Goal
Add a way to fetch the app settings. When connecting the user, it loads the app settings.

### 📝 Summary
The app settings are useful to get file upload configurations, and we were missing it in the iOS SDK. This work is also going to be useful for implementing the new size limit in https://github.com/GetStream/ios-issues-tracking/issues/757.

### 🛠 Implementation
Besides making the app settings request available to be called manually, we also load them whenever the user is connected, so that file upload limitations are available as soon as the user is connected.

It would be simpler if the backend just retrieved the app settings on the `connectUser` response since this would simplify the integration a lot. If the app settings call fails, we won't have the app settings available. We could add a retry mechanism for this call, but we agreed with Android team and the rest of the SDKs to simplify this for now, since even if the appSettings request failed, the server will always protect against these checks.

### 🧪 Manual Testing Notes
1. Open Proxyman
2. Verify that the app settings request is called when connecting the user.

### ☑️ Contributor Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## StreamChat
 ### ✅ Added
 - Add mute expiration support when muting a channel [#3083](https://github.com/GetStream/stream-chat-swift/pull/3083)
+- Add `ChatClient.loadAppSettings` and `ChatClient.appSettings` [#3091](https://github.com/GetStream/stream-chat-swift/pull/3091)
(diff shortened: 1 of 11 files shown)


martinmitrevski (approved):
LGTM ✅  Hurts seeing I have to delete most of this code with open api 😀

laevandus:
Agree that it would be nice if app settings would be part of the connect call. I was also thinking about if we should chain it with the connect call but that does not sound right as well.

laevandus (approved):

PR #3092, opened 2024-03-15
Add attachment queue uploader tests to flaky plan

nuno-vieira:
### 🔗 Issue Links

None

### 🎯 Goal

Add attachment queue uploader flaky tests to the flaky plan. As well as one test from SyncOperation.


nuno-vieira:
@martinmitrevski The existing tests are not flaky, I just added them as well to keep it simple. The new tests added after the sandbox url fixing were the ones which sometimes are flaky. This is related to AssertAsync. This class is not very well designed, so it is hard to test. So we need AssertAsync to test it, which sometimes the timeout is more than one defined.

martinmitrevski:
> @martinmitrevski The existing tests are not flaky, I just added them as well to keep it simple. The new tests added after the sandbox url fixing were the ones which sometimes are flaky. This class is not very well designed, so it is hard to test. So we need AssertAsync to test it, which sometimes the timeout is more than the one defined.
> 
> The test that is flaky is this one: `test_attachmentsAreCopiedToSandbox_beforeBeingSent`, but the whole test cal can be flaky because of the usage of AssertAsync, so better to just put everything of this class as flaky

I think it's better to put only the one that fails though. We can prevent the case where sth was stable, but became flaky over time. Wdyt?

nuno-vieira:
> > @martinmitrevski The existing tests are not flaky, I just added them as well to keep it simple. The new tests added after the sandbox url fixing were the ones which sometimes are flaky. This class is not very well designed, so it is hard to test. So we need AssertAsync to test it, which sometimes the timeout is more than the one defined.
> > The test that is flaky is this one: `test_attachmentsAreCopiedToSandbox_beforeBeingSent`, but the whole test cal can be flaky because of the usage of AssertAsync, so better to just put everything of this class as flaky
> 
> I think it's better to put only the one that fails though. We can prevent the case where sth was stable, but became flaky over time. Wdyt?

@martinmitrevski Fixed it then 👍 I've only added 2 tests from the Queue uploader, which were the ones constantly failing

martinmitrevski (approved):

PR #3097, opened 2024-03-18
[ChatStateLayer] UserSearch and UserSearchState

laevandus:
### 🔗 Issue Links

### 🎯 Goal

Adds `UserSearchController` functionality to the new chat state layer.

### 📝 Summary

- Add `UserSearch` and `UserSearchState` for representing the functionality in `ChatUserSearchController`

### 🛠 Implementation

`ChatUserSearchController` does not really need CoreData support and it currently uses the store only for converting `UserPayload` > `ChatUser` (`UserPayload` is saved to CD as `UserDTO` and the DTO is converted to `ChatUser`). There is also complexity around `UserDTO` and `UserListQueryDTO`. Since we do not really need any CoreData storage here, there are no CoreDataLazy in `ChatUser`, it is possible map the `UserPayload` to `ChatUser` with one exception (`isFlaggedByCurrentUser`), I am proposing to skip CoreData in the `UserSearch`.

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/StateLayer/UserSearch.swift
+++ b/Sources/StreamChat/StateLayer/UserSearch.swift
@@ -1,0 +1,84 @@
+//
+// Copyright © 2024 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+/// An object which represents a list of `ChatUser` for the specified search query.
+@available(iOS 13.0, *)
+public struct UserSearch {
+    private let userListUpdater: UserListUpdater
+    
+    init(client: ChatClient, environment: Environment = .init()) {
+        userListUpdater = environment.userListUpdaterBuilder(
+            client.databaseContainer,
+            client.apiClient
+        )
+        state = UserSearchState()
+    }
+    
+    /// An observable object representing the current state of the search.
+    public let state: UserSearchState
+    
+    /// Searches for users with the specified search term and updates ``UserSearchState.users``.
+    ///
+    /// - Parameter term: The search term for searching users.
+    ///
+    /// - Throws: An error while communicating with the Stream API.
+    /// - Returns: An array of users for the search term.
+    @discardableResult public func search(term: String) async throws -> [ChatUser] {
+        try await search(query: .search(term: term))
+    }
+    
+    /// Searches for users with the specified query and updates ``UserSearchState.users``.
+    ///
+    /// - Parameters:
+    ///   - query: The user list query used for searching.
+    ///
+    /// - Throws: An error while communicating with the Stream API.
+    /// - Returns: An array of users for the query.
+    @discardableResult public func search(query: UserListQuery) async throws -> [ChatUser] {
+        let limit = query.pagination?.pageSize ?? .usersPageSize
+        let pagination = Pagination(pageSize: limit, offset: 0)
+        return try await search(query: query, pagination: pagination)
+    }
+    
+    /// Searches for more users with the specified query and updates ``UserSearchState.users``.
+    ///
+    /// - Parameters
+    ///   - limit: The limit for the page size. The default limit is 30.
+    ///
+    /// - Throws: An error while communicating with the Stream API.
+    /// - Returns: An array of loaded channels.
+    @discardableResult public func loadNextUsers(limit: Int? = nil) async throws -> [ChatUser] {
+        guard let query = await state.value(forKeyPath: \.query) else { throw ClientError("Call search() before calling for next page") }
+        let limit = (limit ?? query.pagination?.pageSize) ?? .usersPageSize
+        let offset = await state.value(forKeyPath: \.users.count)
+        let pagination = Pagination(pageSize: limit, offset: offset)
+        return try await search(query: query, pagination: pagination)
+    }
+    
+    // MARK: - Private
+    
+    private func search(query: UserListQuery, pagination: Pagination) async throws -> [ChatUser] {
+        let task = Task {
+            let users = try await userListUpdater.fetch(userListQuery: query, pagination: pagination)
+            try Task.checkCancellation()
+            return users
+        }
+        await state.setActiveTask(task, query: query)
+        let users = try await task.value
+        await state.setUsers(users, for: query, pagination: pagination)
+        return users
+    }
+}
+
+@available(iOS 13.0, *)
+extension UserSearch {
+    struct Environment {
+        var userListUpdaterBuilder: (
+            _ database: DatabaseContainer,
+            _ apiClient: APIClient
+        ) -> UserListUpdater = UserListUpdater.init
+    }
+}
--- a/Sources/StreamChat/StateLayer/UserSearchState.swift
+++ b/Sources/StreamChat/StateLayer/UserSearchState.swift
@@ -1,0 +1,47 @@
+//
+// Copyright © 2024 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+/// Represents a list of user search results.
+@available(iOS 13.0, *)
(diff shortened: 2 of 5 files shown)


Sources/StreamChat/StateLayer/UserSearch.swift
```diff
@@ -0,0 +1,67 @@
+//
+// Copyright © 2024 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+/// An object which represents a list of `ChatUser` for the specified search query.
+@available(iOS 13.0, *)
+public struct UserSearch {
```
laevandus, line 9: It is pretty much the same as `UserList` with the two exceptions:
- `UserList` has static query, whereas here the query can change
- `UserList` is backed by CoreData observing, here we skip CoreData

Sources/StreamChat/StateLayer/UserSearchState.swift
```diff
@@ -0,0 +1,31 @@
...
+import Foundation
+
+/// Represents a list of user search results.
+@available(iOS 13.0, *)
+public final class UserSearchState: ObservableObject {
+    /// The search query the results corresponds to.
+    @Published public private(set) var searchQuery: UserListQuery?
+    
+    /// An array of search results for the specified query and pagination state.
+    @Published public private(set) var users = StreamCollection<ChatUser>([])
```
laevandus, line 14: I also considered:
- renaming it to `results`
- having native Swift array type (but then it is not aligned with other `XYZState` types)
WDTY?

Sources/StreamChat/StateLayer/UserSearch.swift
```diff
@@ -0,0 +1,67 @@
...
+            client.databaseContainer,
+            client.apiClient
+        )
+        state = UserSearchState()
+    }
+    
+    /// An observable object representing the current state of the search.
+    public let state: UserSearchState
+    
+    @discardableResult public func search(term: String) async throws -> [ChatUser] {
```
laevandus, line 23: Add documentation

Sources/StreamChat/StateLayer/UserSearchState.swift
martinmitrevski: I think you made good decisions, both of them are consistent with the rest of the codebase.

Sources/StreamChat/StateLayer/UserSearchState.swift
```diff
@@ -0,0 +1,31 @@
...
+    
+    @MainActor func setUsers(_ newUsers: [ChatUser], for pagination: Pagination) {
+        if pagination.offset > 0 {
+            var result = Array(users[..<pagination.offset])
+            result.append(contentsOf: newUsers)
+            users = StreamCollection(result)
+        } else {
+            users = StreamCollection(newUsers)
+        }
+    }
```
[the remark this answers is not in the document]
martinmitrevski: hmm, this is strange - can you please explain why we need this? I don't think we had such logic before.

Sources/StreamChat/Workers/UserListUpdater.swift
```diff
@@ -66,9 +72,11 @@ enum UpdatePolicy {
     case merge
     /// The resulting user set of the query will replace the existing user set.
     case replace
+    /// The resulting user set of the query is not stored locally.
+    case ephermal
```
martinmitrevski, line 76: Not sure if `UpdatePolicy` would be still the correct wording here. Basically, `ephermal`, as an update policy doesn't sound correct, wdyt? 
Another point, there's a typo here, should be `ephemeral` 😁

Sources/StreamChat/Workers/UserListUpdater.swift
```diff
@@ -17,26 +17,32 @@ class UserListUpdater: Worker {
...
             switch result {
             case let .success(userListPayload):
-                var users = [ChatUser]()
-                self?.database.write { session in
-                    if case .replace = policy {
-                        let dto = try session.saveQuery(query: userListQuery)
-                        dto?.users.removeAll()
+                if case .ephermal = policy {
+                    if let completion {
+                        let users = userListPayload.users.map { $0.asModal() }
```
nuno-vieira, line 22: Typo: ?
```suggestion
                        let users = userListPayload.users.map { $0.asModel() }
```

Sources/StreamChat/Workers/UserListUpdater.swift
```diff
@@ -17,26 +17,32 @@ class UserListUpdater: Worker {
...
+                        completion(.success(users))
```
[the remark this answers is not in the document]
nuno-vieira: TBH, I would not even introduce this policy. We should never hit the DB, if we can

Sources/StreamChat/Workers/UserListUpdater.swift
nuno-vieira: I would remove the UpdatePolicy all thogether. Search should hit the server, and never go through the DB. The results should always reflect the server

Sources/StreamChat/Workers/UserListUpdater.swift
```diff
@@ -94,3 +102,24 @@ private extension UserListQuery {
         return query
     }
 }
+
+private extension UserPayload {
+    func asModal() -> ChatUser {
```
nuno-vieira, line 107: Typo:
```suggestion
    func asModel() -> ChatUser {
```

nuno-vieira (commented):
Looks good overall 👍 It would be nice though that we remove the Update Policy, and never hit DB for v5

Sources/StreamChat/StateLayer/UserSearch.swift
laevandus: Let's keep this PR open until the cancellation part is also properly implemented.

Sources/StreamChat/StateLayer/UserSearchState.swift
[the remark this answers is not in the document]
laevandus: `ChatUserSearchController` uses CoreData for driving the results array like all the other controllers. Since in this PR I went for skipping the CoreData part. The downside in that case is that then I need to take care of managing the users array myself since any of the fetched results are never put into the DB (which makes sense for searching).
I'll think a little bit more about this, sure I can make it nicer and more reliable.

Sources/StreamChat/Workers/UserListUpdater.swift
laevandus: Thank you

Sources/StreamChat/Workers/UserListUpdater.swift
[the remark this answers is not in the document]
laevandus: I'll consider separating the fetch part. Currently we do not hit the DB as well but the end result is that the updated is messier. I'll clean it up

Sources/StreamChat/StateLayer/UserSearch.swift
martinmitrevski: Don't think this PR should wait on the cancellation part. We need to handle that after we merge the open api stuff, otherwise we might need to redo it again.

Sources/StreamChat/Workers/UserListUpdater.swift
martinmitrevski: logically, that makes sense. But, we need to check if it's used somewhere else or breaks some existing logic.

laevandus: Removed the added policy

nuno-vieira (approved):

PR #3100, opened 2024-03-19
[ChatStateLayer] MessageSearch and MessageSearchState

laevandus:
### 🔗 Issue Links

### 🎯 Goal

Represent functionality in the `ChatMessageSearchController` in the new state layer. 

### 📝 Summary

* Add `MessageSearch`
* Add `MessageSearchState`

### 🛠 Implementation

Keep the internal setup the same where CoreData drives results although it is not ideal (ideally we want to drive it with API request only but we not currently ready for it)

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/StateLayer/MessageSearch.swift
+++ b/Sources/StreamChat/StateLayer/MessageSearch.swift
@@ -56,7 +56,7 @@ public struct MessageSearch {
         query.filterHash = state.explicitFilterHash
         let result = try await messageUpdater.search(query: query, policy: .replace)
         await state.set(query: query, cursor: result.payload.next)
-        return result.messages
+        return result.models
     }
     
     /// Searches for more messages matching with the last search query.
@@ -70,14 +70,14 @@ public struct MessageSearch {
         let limit = (limit ?? query.pagination?.pageSize) ?? Int.messagesPageSize
         let pagination: Pagination = await {
             if !query.sort.isEmpty, let nextPageCursor = await state.value(forKeyPath: \.nextPageCursor) {
-                Pagination(pageSize: limit, cursor: nextPageCursor)
+                return Pagination(pageSize: limit, cursor: nextPageCursor)
             } else {
-                Pagination(pageSize: limit, offset: await state.value(forKeyPath: \.messages.count))
+                return Pagination(pageSize: limit, offset: await state.value(forKeyPath: \.messages.count))
             }
         }()
         let result = try await messageUpdater.search(query: query.withPagination(pagination), policy: .merge)
         await state.set(query: query, cursor: result.payload.next)
-        return result.messages
+        return result.models
     }
     
     // MARK: - Private
--- a/Sources/StreamChat/StateLayer/MessageSearch.swift
+++ b/Sources/StreamChat/StateLayer/MessageSearch.swift
@@ -1,0 +1,113 @@
+//
+// Copyright © 2024 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+/// An object which represents a list of ``ChatMessage`` for the specified search query.
+@available(iOS 13.0, *)
+public struct MessageSearch {
+    private let authenticationRepository: AuthenticationRepository
+    private let messageUpdater: MessageUpdater
+    
+    init(client: ChatClient, environment: Environment = .init()) {
+        authenticationRepository = client.authenticationRepository
+        messageUpdater = environment.messageUpdaterBuilder(
+            client.config.isLocalStorageEnabled,
+            client.messageRepository,
+            client.databaseContainer,
+            client.apiClient
+        )
+        state = environment.stateBuilder(client.databaseContainer)
+    }
+    
+    /// An observable object representing the current state of the search.
+    public let state: MessageSearchState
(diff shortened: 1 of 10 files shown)


Sources/StreamChat/Workers/MessageUpdater.swift
```diff
@@ -610,22 +610,26 @@ class MessageUpdater: Worker {
         })
     }
 
-    func search(query: MessageSearchQuery, policy: UpdatePolicy = .merge, completion: ((Result<MessageSearchResultsPayload, Error>) -> Void)? = nil) {
+    func search(query: MessageSearchQuery, policy: UpdatePolicy = .merge, completion: ((Result<(payload: MessageSearchResultsPayload, messages: [ChatMessage]), Error>) -> Void)? = nil) {
```
laevandus, line 613: I'll change it a bit since it is too messy in unit-tests otherwise.

Sources/StreamChat/StateLayer/MessageSearch.swift
```diff
@@ -0,0 +1,113 @@
...
+    
+    /// Searches for more messages matching with the last search query.
+    ///
+    /// - Parameter limit: The limit for the page size. The default limit is 25.
+    ///
+    /// - Throws: An error while communicating with the Stream API.
+    /// - Returns: A next page of chat messages matching to the last query.
+    @discardableResult public func loadNextMessages(limit: Int? = nil) async throws -> [ChatMessage] {
+        guard let query = state.query else { throw ClientError("Call search() before calling for next page") }
+        let limit = (limit ?? query.pagination?.pageSize) ?? Int.messagesPageSize
```
nuno-vieira, line 70: Probably out of scope, but ideally we want to remove these `Int.messagesPageSize` public extensions for v5. We should provide a config in `ChatClientConfig` with the default messages pagination size.

Sources/StreamChat/StateLayer/MessageSearch.swift
```diff
@@ -0,0 +1,113 @@
...
+        let pagination: Pagination = await {
+            if !query.sort.isEmpty, let nextPageCursor = await state.value(forKeyPath: \.nextPageCursor) {
+                Pagination(pageSize: limit, cursor: nextPageCursor)
```
nuno-vieira, line 73: To fix Xcode 14 compilation:
```suggestion
               return Pagination(pageSize: limit, cursor: nextPageCursor)
```

Sources/StreamChat/StateLayer/MessageSearch.swift
```diff
@@ -0,0 +1,113 @@
...
+            } else {
+                Pagination(pageSize: limit, offset: await state.value(forKeyPath: \.messages.count))
```
nuno-vieira, line 75: Same here

Sources/StreamChat/StateLayer/MessageSearchState.swift
```diff
@@ -0,0 +1,46 @@
...
+// Copyright © 2024 Stream.io Inc. All rights reserved.
+//
+
+import Combine
+import Foundation
+
+/// Represents a list of message search results.
+@available(iOS 13.0, *)
+public final class MessageSearchState: ObservableObject {
+    private var cancellables = Set<AnyCancellable>()
```
nuno-vieira, line 11: Can you clarify how are these cancelables working? 🤔 Since they are private, how do we cancel them?

Sources/StreamChat/StateLayer/MessageSearch.swift
laevandus: Added a TODO item since I have used these in many other places as well. Needs to be changed with one go.

Sources/StreamChat/StateLayer/MessageSearchState.swift
laevandus: This is only for internal observation for forwarding the query to the `Observer` object.

```swift
 $query
            .assign(to: \.query, on: observer)
            .store(in: &cancellables)
```

Sources/StreamChat/StateLayer/MessageSearchState+Observer.swift
```diff
@@ -0,0 +1,62 @@
...
+        }
+        
+        private func reset(to query: MessageSearchQuery?) {
+            if let query {
+                messagesObserver = BackgroundListDatabaseObserver(
+                    context: database.backgroundReadOnlyContext,
+                    fetchRequest: MessageDTO.messagesFetchRequest(for: query),
+                    itemCreator: { try $0.asModel() as ChatMessage },
+                    sorting: []
+                )
```
laevandus, line 45: In general I am not happy about this part but we need to do the work of supporting MessagePayload>ChatMessage conversion first. Slightly silly since results are API driven and we do not really use the CD storage for search. But that is the current situation.

Sources/StreamChat/StateLayer/MessageSearch.swift
martinmitrevski: good point, but yes, out of scope for this one.

martinmitrevski (approved):
Looks great 👌

PR #3102, opened 2024-03-19
[ChatStateLayer] Client, connection and channel events

laevandus:
### 🔗 Issue Links

### 🎯 Goal

Add functionality in `ChannelEventsController`, `EventsController`, and `ChatConnectionController ` to the new state layer.

### 📝 Summary

* Add subscribe methods for channel events to `Chat`(related to `ChannelEventsController`)
* Add send custom event to `Chat` (related to `ChannelEventsController`)
* Add subscribe methods for client events to `ChatClient` (related to `EventsController`)
* Connection events use the same subscribe method in `ChatClient` (related to `ChatConnectionController`)

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/ChatClient.swift
+++ b/Sources/StreamChat/ChatClient.swift
@@ -2,6 +2,7 @@
 // Copyright © 2024 Stream.io Inc. All rights reserved.
 //
 
+import Combine
 import CoreData
 import Foundation
 
@@ -495,7 +496,47 @@ public class ChatClient {
         }
     }
     
-    // MARK: - Internal
+    // MARK: - Listening for Client Events
+    
+    /// Subscribes to web-socket events of the specified event type.
+    ///
+    /// - Note: The handler is always called on the main thread.
+    ///
+    /// An example of observing connection status changes:
+    /// ```swift
+    /// client.subscribe(to: ConnectionStatusUpdated.self) { connectionEvent in
+    ///     switch connectionEvent.connectionStatus {
+    ///         case .connected:
+    ///           …
+    ///     }
+    /// }
+    /// ```
+    ///
+    /// - SeeAlso: ``Chat.subscribe(to:handler:)`` for subscribing to channel specific events.
+    ///
+    /// - Parameters:
+    ///   - event: The event type to subscribe to (e.g. ``ConnectionStatusUpdated``).
+    ///   - handler: The handler closure which is called when the event happens.
+    ///
+    /// - Returns: A cancellable instance, which you use when you end the subscription. Deallocation of the result will tear down the subscription stream.
+    @available(iOS 13.0, *)
+    public func subscribe<E>(to event: E.Type, handler: @escaping (E) -> Void) -> AnyCancellable where E: Event {
+        eventNotificationCenter.subscribe(to: E.self, handler: handler)
+    }
+
+    /// Subscribes to all the web-socket events.
+    ///
+    /// - SeeAlso: ``Chat.subscribe(handler:)`` for subscribing to channel specific events.
+    ///
+    /// - Parameter handler: The handler closure which is called when the event happens.
+    ///
+    /// - Returns: A cancellable instance, which you use when you end the subscription. Deallocation of the result will tear down the subscription stream.
+    @available(iOS 13.0, *)
+    public func subscribe(_ handler: @escaping (Event) -> Void) -> AnyCancellable {
+        eventNotificationCenter.subscribe(handler: handler)
+    }
+    
+    // MARK: -
 
     /// Fetches the app settings and updates the `ChatClient.appSettings`.
     /// - Parameter completion: The completion block once the app settings has finished fetching.
@@ -513,6 +554,8 @@ public class ChatClient {
             }
         }
     }
+
+    // MARK: - Internal
 
     func createBackgroundWorkers() {
         guard config.isClientInActiveMode else { return }
--- a/Sources/StreamChat/ChatClient.swift
+++ b/Sources/StreamChat/ChatClient.swift
@@ -504,7 +504,7 @@ public class ChatClient {
     ///
     /// An example of observing connection status changes:
     /// ```swift
-    /// client.subscribe(to: ConnectionStatusUpdated.self) { connectionEvent in
+    /// client.subscribe(toEvent: ConnectionStatusUpdated.self) { connectionEvent in
     ///     switch connectionEvent.connectionStatus {
     ///         case .connected:
     ///           …
@@ -512,7 +512,7 @@ public class ChatClient {
     /// }
     /// ```
     ///
-    /// - SeeAlso: ``Chat.subscribe(to:handler:)`` for subscribing to channel specific events.
+    /// - SeeAlso: ``Chat.subscribe(toEvent:handler:)`` for subscribing to channel specific events.
     ///
     /// - Parameters:
     ///   - event: The event type to subscribe to (e.g. ``ConnectionStatusUpdated``).
@@ -520,7 +520,7 @@ public class ChatClient {
     ///
     /// - Returns: A cancellable instance, which you use when you end the subscription. Deallocation of the result will tear down the subscription stream.
     @available(iOS 13.0, *)
-    public func subscribe<E>(to event: E.Type, handler: @escaping (E) -> Void) -> AnyCancellable where E: Event {
+    public func subscribe<E>(toEvent event: E.Type, handler: @escaping (E) -> Void) -> AnyCancellable where E: Event {
         eventNotificationCenter.subscribe(to: E.self, handler: handler)
     }
 
--- a/Sources/StreamChat/StateLayer/Chat.swift
+++ b/Sources/StreamChat/StateLayer/Chat.swift
@@ -953,14 +953,14 @@ public final class Chat {
     ///
     /// - Note: The handler is always called on the main thread.
     /// - Important: Subscribing to events not related to this channel, like ``ConnectionStatusUpdated``, does not trigger the handler.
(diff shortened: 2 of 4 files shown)


Sources/StreamChat/StateLayer/Chat.swift
```diff
@@ -936,6 +942,73 @@ public final class Chat {
...
+            handler: handler
+        )
+    }
+    
+    /// Subscribes to web-socket events of a single type in this channel.
+    ///
+    /// - Parameter event: The event type to subscribe to (e.g. ``MessageNewEvent``).
+    ///
+    /// - Returns: Web-socket events of this channel, as an asynchronous stream of events.
+    public func subscribe<T>(to event: T.Type) -> AsyncStream<T> where T: Event {
```
martinmitrevski, line 969: Topic for discussion, but I would actually prefer using only Combine cancellables - the room for error is too big with async streams. Curious to hear the others opinion.

Sources/StreamChat/Workers/EventNotificationCenter.swift
```diff
@@ -73,3 +74,47 @@ extension EventNotificationCenter {
         process([event], postNotifications: postNotification, completion: completion)
     }
 }
+
+@available(iOS 13.0, *)
+extension EventNotificationCenter {
```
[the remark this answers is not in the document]
martinmitrevski: Same here - I think we should reduce all the options we offer. It becomes too complex to manage for us, and can be confusing for customers.

martinmitrevski (commented):
looks good 👍  Would be nice to simplify it - pick one direction to go with for event listening.

Sources/StreamChat/StateLayer/Chat.swift
laevandus: Yeah, probably makes sense to get rid of the stream variant (at least for now). One can always create streams later with the handler (if someone really wants to use streams).
@nuno-vieira please also share your opinion.

laevandus: Let me remove it for now and we can consider adding it back if we decide to do so.

Sources/StreamChat/ChatClient.swift
```diff
@@ -404,6 +405,48 @@ public class ChatClient {
...
+    ///
+    /// - Parameters:
+    ///   - event: The event type to subscribe to (e.g. ``ConnectionStatusUpdated``).
+    ///   - handler: The handler closure which is called when the event happens.
+    ///
+    /// - Returns: A cancellable instance, which you use when you end the subscription. Deallocation of the result will tear down the subscription stream.
+    @available(iOS 13.0, *)
+    public func subscribe<T>(to eventType: T.Type, handler: @escaping (T) -> Void) -> AnyCancellable where T: Event {
+        eventNotificationCenter.subscribe(to: T.self, handler: handler)
+    }
```
laevandus, line 435: `ChatConnectionController` is currently the way of observing connection events. But since observing connection events is same as observing any events, there is no point of having separate subscribe methods. Instead, I just added a snippet for that to the documentation (since people who migrate might get confused about it).

nuno-vieira: Yeah the connection controller has always been very confusing, so Ideally we should not have it in the new state layer 👍

Sources/StreamChat/ChatClient.swift
```diff
@@ -404,6 +405,48 @@ public class ChatClient {
...
+    /// - Note: The handler is always called on the main thread.
+    ///
+    /// An example of observing connection status changes:
+    /// ```swift
+    /// client.subscribe(to: ConnectionStatusUpdated.self) { connectionEvent in
+    ///     switch connectionEvent.connectionStatus {
+    ///         case .connected:
+    ///           …
+    ///     }
+    /// }
```
nuno-vieira, line 422: Nice 👍 Initially, I thought about creating a separate function per event, to avoid all the casting we do right now, but this is actually even better 👍

Sources/StreamChat/ChatClient.swift
```diff
@@ -404,6 +405,48 @@ public class ChatClient {
...
+    ///
+    /// - SeeAlso: ``Chat.subscribe(to:handler:)`` for subscribing to channel specific events.
+    ///
+    /// - Parameters:
+    ///   - event: The event type to subscribe to (e.g. ``ConnectionStatusUpdated``).
+    ///   - handler: The handler closure which is called when the event happens.
+    ///
+    /// - Returns: A cancellable instance, which you use when you end the subscription. Deallocation of the result will tear down the subscription stream.
+    @available(iOS 13.0, *)
+    public func subscribe<T>(to eventType: T.Type, handler: @escaping (T) -> Void) -> AnyCancellable where T: Event {
```
nuno-vieira, line 433: I think it makes sense to add a little more context in this case 🤔 
```suggestion
    public func subscribe<T>(toEvent eventType: T.Type, handler: @escaping (T) -> Void) -> AnyCancellable where T: Event {
```

Sources/StreamChat/ChatClient.swift
```diff
@@ -404,6 +405,48 @@ public class ChatClient {
...
+    /// ```swift
+    /// client.subscribe(to: ConnectionStatusUpdated.self) { connectionEvent in
+    ///     switch connectionEvent.connectionStatus {
+    ///         case .connected:
+    ///           …
+    ///     }
+    /// }
+    /// ```
+    ///
+    /// - SeeAlso: ``Chat.subscribe(to:handler:)`` for subscribing to channel specific events.
```
nuno-vieira, line 425: TBH I don't think we need this 🤔 No other SDK has a component to observer only Channel events. The new `ChatClient.subscribe(toEvent:)` is more than enough IMO

Sources/StreamChat/ChatClient.swift
```diff
@@ -404,6 +405,48 @@ public class ChatClient {
...
+
+    /// Subscribes to all the web-socket events.
+    ///
+    /// - SeeAlso: ``Chat.subscribe(handler:)`` for subscribing to channel specific events.
+    ///
+    /// - Parameter handler: The handler closure which is called when the event happens.
+    ///
+    /// - Returns: A cancellable instance, which you use when you end the subscription. Deallocation of the result will tear down the subscription stream.
+    @available(iOS 13.0, *)
+    public func subscribe(_ handler: @escaping (Event) -> Void) -> AnyCancellable {
```
nuno-vieira, line 445: I don't think we need this either 🤔 If a customer creates a custom event type, they will also be able to observe it with the other one right? If that is true (I think it is), then I would rather not have this one. It just causes more confusion to the customers, having 2 ways to do the same thing

Sources/StreamChat/StateLayer/Chat.swift
```diff
@@ -936,6 +942,51 @@ public final class Chat {
...
+    /// - Returns: A cancellable instance, which you use when you end the subscription. Deallocation of the result will tear down the subscription stream.
+    public func subscribe<T>(to event: T.Type, handler: @escaping (T) -> Void) -> AnyCancellable where T: Event {
+        eventNotificationCenter.subscribe(
+            to: event,
+            filter: { [cid] in
+                EventNotificationCenter.channelFilter(cid: cid, event: $0)
+            },
+            handler: handler
+        )
+    }
```
nuno-vieira, line 963: Like I said above, I think we should actually remove this 👍 Unless I'm missing something? What the customer can do here, is the same of using directly the `ChatClient.subscribe(to: ChannelEvent.Type)` right?

nuno-vieira (commented):
Looks really nice overall 👌 I just think we can slim it down and make it more simpler, and more aligned with the other SDKs (Just 1 API to observe events instead of multiple ways)

Sources/StreamChat/ChatClient.swift
laevandus: You are right, this one receives every single channel event as well.

Sources/StreamChat/ChatClient.swift
laevandus: Currently I was following what Apple is doing:
```swift
func subscribe<E>(to event: E.Type, …
```
Well, I should use `E` instead of `T` :)

Sources/StreamChat/ChatClient.swift
laevandus: client.subscribe { event in
  switch event {
    case  let connectionEvent as ConnectionStatusUpdated:
      ...
  }
}

client.subscribe { event in
  Logger().debug(type(of: event))
}

Sources/StreamChat/StateLayer/Chat.swift
```diff
@@ -947,6 +953,51 @@ public final class Chat {
...
+    ///
+    /// - Note: The handler is always called on the main thread.
+    /// - SeeAlso: ``ChatClient.subscribe(to:handler:)`` for subscribing to client events.
+    ///
+    /// - Parameters:
+    ///   - event: The event type to subscribe to (e.g. ``MessageNewEvent``).
+    ///   - handler: The handler closure which is called when the event happens.
+    ///
+    /// - Returns: A cancellable instance, which you use when you end the subscription. Deallocation of the result will tear down the subscription stream.
+    public func subscribe<E>(to event: E.Type, handler: @escaping (E) -> Void) -> AnyCancellable where E: Event {
```
laevandus, line 966: We can limit the E: to `where E: ChannelSpecificEvent`

Sources/StreamChat/ChatClient.swift
laevandus: We had a longer discussion on this in the tech meeting and the end result was that let's keep it as is and revisit it when we can use the new enum based events.

PR #3104, opened 2024-03-20
[ChatStateLayer] Sync repository channel list query tracking

laevandus:
### 🔗 Issue Links

### 🎯 Goal

Clean up the notification approach from the `SyncRepository`

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/Repositories/SyncRepository.swift
+++ b/Sources/StreamChat/Repositories/SyncRepository.swift
@@ -40,6 +40,9 @@ class SyncRepository {
     let activeChannelListControllers: ThreadSafeWeakCollection<ChatChannelListController>
     let offlineRequestsRepository: OfflineRequestsRepository
     let eventNotificationCenter: EventNotificationCenter
+    
+    private var _activeChannelListQueryProviders = [() -> ChannelListQuery?]()
+    private let syncQueue = DispatchQueue(label: "io.getstream.sync-repository-queue")
 
     private lazy var operationQueue: OperationQueue = {
         let operationQueue = OperationQueue()
@@ -71,6 +74,22 @@ class SyncRepository {
     deinit {
         cancelRecoveryFlow()
     }
+    
+    // MARK: - Tracking Active
+    
+    func trackChannelListQuery(_ provider: @escaping () -> ChannelListQuery?) {
+        syncQueue.sync {
+            _activeChannelListQueryProviders.append(provider)
+        }
+    }
+    
+    func removeAllTracked() {
+        syncQueue.sync {
+            _activeChannelListQueryProviders.removeAll()
+        }
+    }
+    
+    // MARK: -
 
     func syncLocalState(completion: @escaping () -> Void) {
         cancelRecoveryFlow()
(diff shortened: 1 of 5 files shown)


Sources/StreamChat/StateLayer/ChatClient+Factory.swift
```diff
@@ -36,7 +36,10 @@ extension ChatClient {
     /// - Returns: An instance of ``ChannelList`` which represents actions and the current state of the list.
     public func makeChannelList(with query: ChannelListQuery, dynamicFilter: ((ChatChannel) -> Bool)? = nil) async throws -> ChannelList {
         let channels = try await channelListUpdater.update(channelListQuery: query)
-        return ChannelList(channels: channels, query: query, dynamicFilter: dynamicFilter, channelListUpdater: channelListUpdater, client: self)
+        let channelList = ChannelList(channels: channels, query: query, dynamicFilter: dynamicFilter, channelListUpdater: channelListUpdater, client: self)
+        let state = channelList.state
+        syncRepository.trackChannelListQuery { [weak state] in state?.query }
+        return channelList
```
laevandus, line 42: `ChannelList` is a struct, that is why we use state instead which is a class.

nuno-vieira: Should it be a struct, though? 🤔  Maybe a point to add to the Tech Discussion, but I don't think that Chat, ChannelList, MemberList, etc... should be structs. For these reasons:
- They contain reference types, so there's not much value on them to be value types
- Since there's no value on them being a struct (Unless I'm missing something), it is not worth the additional binary size that Structs produce, more details here: https://forums.swift.org/t/reducing-value-witness-size-in-struct-composition/52172
- It will be difficult to mock for testing (For us and for the customers), which would require exposing protocols instead of the structs directly.

Sources/StreamChat/Repositories/SyncRepository.swift
```diff
@@ -71,6 +74,22 @@ class SyncRepository {
     deinit {
         cancelRecoveryFlow()
     }
+    
+    // MARK: - Tracking Active
+    
+    func trackChannelListQuery(_ provider: @escaping () -> ChannelListQuery?) {
```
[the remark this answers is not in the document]
nuno-vieira: I think in this case, we can move to the new provider, since there's not much knowledge of this in the current team either way. WDYT @martinmitrevski ?

Sources/StreamChat/StateLayer/ChatClient+Factory.swift
laevandus: Currently `Chat` and all the `XYZState` are classes, other types are structs. `Chat` could also be a struct. These are currently structs since they do not hold any mutable state. 
Currently there is no strong reason of `non-State` types must be structs. Let's discuss this in the tech meeting.

Sources/StreamChat/Repositories/SyncRepository.swift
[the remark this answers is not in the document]
laevandus: If I want to replicate this exactly, then I need to add some sort of callback as well since currently SyncRepository calls into `ChatChannelListController` and updates the `hasLoadedAllPreviousChannels`.
Then it should look like: `provider: @escaping () -> (query: ChannelListQuery, callback: ([ChatChannel]) -> Void)` (or a separate type for making this more readable).

https://github.com/GetStream/stream-chat-swift/blob/develop/Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift#L195C5-L215C6
```swift
func resetQuery(
        watchedAndSynchedChannelIds: Set<ChannelId>,
        synchedChannelIds: Set<ChannelId>,
        completion: @escaping (Result<(synchedAndWatched: [ChatChannel], unwanted: Set<ChannelId>), Error>) -> Void
    ) {
        let pageSize = query.pagination.pageSize
        worker.resetChannelsQuery(
            for: query,
            pageSize: pageSize,
            watchedAndSynchedChannelIds: watchedAndSynchedChannelIds,
            synchedChannelIds: synchedChannelIds
        ) { [weak self] result in
            switch result {
            case let .success((newChannels, unwantedCids)):
                self?.hasLoadedAllPreviousChannels = newChannels.count < pageSize
                completion(.success((newChannels, unwantedCids)))
            case let .failure(error):
                completion(.failure(error))
            }
        }
    }
```

Sources/StreamChat/StateLayer/ChatClient+Factory.swift
laevandus: Discussion result: let's change them to classes in a separate commit/PR

laevandus: This is done in the base branch.

Sources/StreamChat/Repositories/SyncRepository.swift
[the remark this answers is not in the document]
laevandus: I played around with it a bit but still feels like it is better to do such cleanup later for the controller. It is already captured in the Notion page as well.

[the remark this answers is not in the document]
martinmitrevski: I think it's best if we don't change this in the first iteration. Sync is very complex and tricky to test, so best to do as little changes as possible while we do the bigger stuff. It's anyway internal, we can change it later.

PR #3107, opened 2024-03-20
Add all tests from attachment queue uploader to flaky plan

nuno-vieira:
### 🔗 Issue Links
None

### 🎯 Goal

It seems like the tests from the attachment queue uploader keep failing randomly. So it is better we add all of them to the flaky plan. Since these tests rely a lot on AssertAsync, it is better they are part of the flaky plan for now.


laevandus (approved):
Yeah, keep seeing some of these as well

PR #3113, opened 2024-03-25
[ChatStateLayer] Unit-test setup for Chat

laevandus:
### 🔗 Issue Links

### 🎯 Goal

Unit-test setup for `Chat`

### 📝 Summary

- How to handle accessing background workers
- Context and `deletedMessagesVisibility` threading 

### ☑️ Contributor Checklist


Sources/StreamChat/StateLayer/Chat.swift
```diff
@@ -36,16 +38,14 @@ public final class Chat {
         client: ChatClient,
         environment: Environment = .init()
     ) {
-        attachmentQueueUploader = client.attachmentQueueUploader
```
laevandus, line 39: This is accessing background workers in `ChatClient`. These are setup later when connecting a user, so I figured it is better to access these later in `Chat` as well but this requires that I store the `ChatClient` instance. Controllers do the same and I guess we need to have this shortcut available as a public API anyway.
From the design point of view it feels a little bit weird that a child object keeps a strong reference to the creator object but I guess it is fine.

Sources/StreamChat/StateLayer/Chat.swift
```diff
@@ -269,6 +270,7 @@ public final class Chat {
     ///
     /// - Throws: An error while sending a message to the Stream API.
     public func resendMessage(_ messageId: MessageId) async throws {
+        let messageEditor = try client.backgroundWorker(of: MessageEditor.self)
```
laevandus, line 273: After going back and forth I ended up with this solution which is accessing background workers only when needed and throwing an error if it is missing (See `ChatClient.createBackgroundWorkers`).

Sources/StreamChat/StateLayer/Chat.swift
```diff
@@ -1012,17 +1020,17 @@ extension Chat {
...
+        if let worker = backgroundWorkers.compactMap({ $0 as? T }).first {
+            return worker
+        }
+        if currentUserId == nil {
+            throw ClientError.CurrentUserDoesNotExist()
+        }
+        if !config.isClientInActiveMode {
+            throw ClientError.ClientIsNotInActiveMode()
+        }
+        throw ClientError("Background worker of type \(T.self) is not set up")
```
laevandus, line 1034: This is trying to follow the flow in `ChatClient` and giving the most accurate error message. Maybe, this should be moved to close to `createBackgroundWorkers`.

Sources/StreamChat/StateLayer/ChatState+Observer.swift
```diff
@@ -38,8 +39,8 @@ extension ChatState {
                     for: cid,
                     pageSize: channelQuery.pagination?.pageSize ?? .messagesPageSize,
                     sortAscending: messageOrder.isAscending,
-                    deletedMessagesVisibility: context.deletedMessagesVisibility ?? .visibleForCurrentUser,
-                    shouldShowShadowedMessages: context.shouldShowShadowedMessages ?? false
+                    deletedMessagesVisibility: clientConfig.deletedMessagesVisibility,
+                    shouldShowShadowedMessages: clientConfig.shouldShowShadowedMessages
```
laevandus, line 43: This was throwing an assert in Apple's code since I was accessing `NSManagedObjectContext`'s `userInfo` from the main thread when running tests. Realised that it is nicer to use the config directly since then I do not need to think about the default value.

TestTools/StreamChatTestTools/Mocks/StreamChat/Workers/ChannelUpdater_Mock.swift
```diff
@@ -250,6 +252,12 @@ final class ChannelUpdater_Mock: ChannelUpdater {
     override func deleteChannel(cid: ChannelId, completion: ((Error?) -> Void)? = nil) {
         deleteChannel_cid = cid
         deleteChannel_completion = completion
+        if let result = deleteChannel_completion_next_result {
```
laevandus, line 255: For async functions I want to feed in the next result before calling the function. Our current tests with completion handler call the function first, and then access the deleteChannel_completion later. In summary, the order changes with async function testing.

Tests/StreamChatTests/StateLayer/Chat_Tests.swift
```diff
@@ -0,0 +1,110 @@
...
+            memberUpdaterBuilder: { [unowned self] in
+                self.memberUpdater = ChannelMemberUpdater_Mock(database: $0, apiClient: $1)
+                return self.memberUpdater!
+            },
+            messageUpdaterBuilder: { [unowned self] in
+                self.messageUpdater = MessageUpdater_Mock(isLocalStorageEnabled: $0, messageRepository: $1, database: $2, apiClient: $3)
+                return self.messageUpdater!
+            },
+            readStateSenderBuilder: { [unowned self] in
+                self.readStateSender = Chat.ReadStateSender(cid: $0, channelUpdater: $1, authenticationRepository: $2, messageRepository: $3)
```
laevandus, line 101: New type, will get a mock version later

TestTools/StreamChatTestTools/Mocks/StreamChat/Workers/ChannelUpdater_Mock.swift
```diff
@@ -26,6 +26,7 @@ final class ChannelUpdater_Mock: ChannelUpdater {
 
     @Atomic var deleteChannel_cid: ChannelId?
     @Atomic var deleteChannel_completion: ((Error?) -> Void)?
+    @Atomic var deleteChannel_completion_next_result: Result<Void, Error>?
```
laevandus, line 29: These are `Atomic` but the mock actually never uses the `mutate` function so all in all, we never use the internal locking what `Atomic` provides. Probably makes sense to delete all of these `Atomic` property wrappers.

Tests/StreamChatTests/StreamChatFlakyTests.xctestplan
```diff
@@ -9,7 +9,8 @@
     }
   ],
   "defaultOptions" : {
-
+    "defaultTestExecutionTimeAllowance" : 60,
+    "testTimeoutsEnabled" : true
```
laevandus, line 13: 60 is the lowest value Xcode allows to set. This is used for failing async test cases which never return

PR #3121, opened 2024-04-02
[ChatStateLayer] ChannelList tests

laevandus:
### 🔗 Issue Links

### 🎯 Goal

* Add basic test coverage to `ChannelList`, `ChannelListState`, and `ChannelListState.Observer`

### 📝 Summary

- New CoreDate observer which does not lag (otherwise `loadChannel` finished, but `ChannelListState.channels` were not updated)
- Various tests covering main uses-cases

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/Controllers/DatabaseObserver/StateLayerListDatabaseObserver.swift
+++ b/Sources/StreamChat/Controllers/DatabaseObserver/StateLayerListDatabaseObserver.swift
@@ -1,0 +1,130 @@
+//
+// Copyright © 2024 Stream.io Inc. All rights reserved.
+//
+
+import CoreData
+import Foundation
+
+/// A CoreDate store observer which immediately reports changes as soon as the store has been changed.
+///
+/// Skips granual list changes what would require to convert DTO to a model object. It reports only when the list has changed and returns a lazy loaded collection.
+@available(iOS 13.0, *)
+final class StateLayerListDatabaseObserver<Item, DTO: NSManagedObject> {
+    private let frc: NSFetchedResultsController<DTO>
+    private(set) var resultsDelegate: FetchedResultsDelegate?
+    private let queue = DispatchQueue(label: "io.getstream.streamchat.statelayerdatabaseobserver")
+
+    let itemCreator: (DTO) throws -> Item
+    let sorting: [SortValue<Item>]
+    let request: NSFetchRequest<DTO>
+    let context: NSManagedObjectContext
+    
+    init(
+        context: NSManagedObjectContext,
+        fetchRequest: NSFetchRequest<DTO>,
+        itemCreator: @escaping (DTO) throws -> Item,
+        sorting: [SortValue<Item>] = [],
+        fetchedResultsControllerType: NSFetchedResultsController<DTO>.Type = NSFetchedResultsController<DTO>.self
+    ) {
+        self.context = context
+        request = fetchRequest
+        self.itemCreator = itemCreator
+        self.sorting = sorting
+        frc = fetchedResultsControllerType.init(
+            fetchRequest: request,
+            managedObjectContext: context,
+            sectionNameKeyPath: nil,
+            cacheName: nil
+        )
+    }
+    
+    convenience init(
+        databaseContainer: DatabaseContainer,
+        fetchRequest: NSFetchRequest<DTO>,
+        itemCreator: @escaping (DTO) throws -> Item,
+        sorting: [SortValue<Item>]
+    ) {
+        // We must use the writableContext since state layer needs to react to the change immediately.
+        // Otherwise async functions updating the CoreData store can return before we have updated the respective state object.
+        self.init(
+            context: databaseContainer.writableContext,
+            fetchRequest: fetchRequest,
+            itemCreator: itemCreator,
+            sorting: sorting
+        )
+    }
+    
+    var items: StreamCollection<Item> {
+        queue.sync { _items }
+    }
+
+    private var _items = StreamCollection<Item>([])
+    
+    func startObserving(didChange: ((StreamCollection<Item>) async -> Void)? = nil) throws {
+        resultsDelegate = FetchedResultsDelegate(onDidChange: { [weak self] in
+            guard let self else { return }
+            // Runs on the NSManagedObjectContext's queue
+            let collection = Self.makeCollection(
+                frc: self.frc,
+                context: self.context,
+                itemCreator: self.itemCreator,
+                sorting: self.sorting
+            )
+            queue.sync {
+                self._items = collection
+            }
+            Task { await didChange?(collection) }
+        })
+        frc.delegate = resultsDelegate
+        try frc.performFetch()
+        // Set initial state on the calling thread, which requires to use performAndWait
+        queue.async { [frc, context, itemCreator, sorting, weak self] in
+            var collection: StreamCollection<Item>!
+            context.performAndWait {
+                collection = Self.makeCollection(frc: frc, context: context, itemCreator: itemCreator, sorting: sorting)
+            }
+            self?._items = collection
+        }
+    }
+    
+    static func makeCollection(
+        frc: NSFetchedResultsController<DTO>,
+        context: NSManagedObjectContext,
+        itemCreator: @escaping (DTO) throws -> Item,
+        sorting: [SortValue<Item>]
+    ) -> StreamCollection<Item> {
+        var result = LazyCachedMapCollection<Item>(
+            source: frc.fetchedObjects ?? [],
+            map: { dto in
+                var resultItem: Item!
+                do {
+                    resultItem = try itemCreator(dto)
+                } catch {
+                    log.assertionFailure("Unable to convert a DB entity to model: \(error.localizedDescription)")
+                }
+                return resultItem
+            },
+            context: context
+        )
+        if !sorting.isEmpty {
+            let sorted = Array(result).sort(using: sorting)
+            result = LazyCachedMapCollection(source: sorted, map: { $0 }, context: context)
+        }
+        return StreamCollection(result)
+    }
+}
+
+@available(iOS 13.0, *)
+extension StateLayerListDatabaseObserver {
+    final class FetchedResultsDelegate: NSObject, NSFetchedResultsControllerDelegate {
+        let onDidChange: (() -> Void)?
+        
+        init(onDidChange: (() -> Void)?) {
+            self.onDidChange = onDidChange
+        }
+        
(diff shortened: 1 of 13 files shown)


Sources/StreamChat/Controllers/DatabaseObserver/StateLayerListDatabaseObserver.swift
```diff
@@ -0,0 +1,130 @@
...
+//
+
+import CoreData
+import Foundation
+
+/// A CoreDate store observer which immediately reports changes as soon as the store has been changed.
+///
+/// Skips granual list changes what would require to convert DTO to a model object. It reports only when the list has changed and returns a lazy loaded collection.
+@available(iOS 13.0, *)
+final class StateLayerListDatabaseObserver<Item, DTO: NSManagedObject> {
```
laevandus, line 12: While writing tests I encountered an issue where this did not work:
```swift
try await channelList.loadChannels(with: pagination)
let channels = channelList.state.channels // <-- new channels were not returned
```

The core issue was that if I change the local store using `writableContext` but I observe the store using the `backgroundContext`, then there is a slight delay before `backgroundContext` is notified about the store change. Moreover, `BackgroundListDatabaseObserver` what I was using before has internal operation queue which delays the change even more. I also looked into using `ListDatabaseObserver` but it does some unnecessary model conversions the state layer does not need (see ChangeAggregator which converts DTOs to models). Also there was an opportunity to skip `performAndWait` and `@Atomic` locking when FRC reported changes. Thirdly, I was able to hide bring in async callback to this level and do the StreamCollection conversion here instead.

Sources/StreamChat/Controllers/DatabaseObserver/StateLayerListDatabaseObserver.swift
```diff
@@ -0,0 +1,130 @@
...
+        frc.delegate = resultsDelegate
+        try frc.performFetch()
+        // Set initial state on the calling thread, which requires to use performAndWait
+        queue.async { [frc, context, itemCreator, sorting, weak self] in
+            var collection: StreamCollection<Item>!
+            context.performAndWait {
+                collection = Self.makeCollection(frc: frc, context: context, itemCreator: itemCreator, sorting: sorting)
+            }
+            self?._items = collection
+        }
```
laevandus, line 87: On the second look, this is not strictly required every single time so I should make it configurable.

Sources/StreamChat/Controllers/DatabaseObserver/StateLayerListDatabaseObserver.swift
```diff
@@ -0,0 +1,130 @@
...
+        )
+    }
+    
+    var items: StreamCollection<Item> {
+        queue.sync { _items }
+    }
+
+    private var _items = StreamCollection<Item>([])
+    
+    func startObserving(didChange: ((StreamCollection<Item>) async -> Void)? = nil) throws {
```
laevandus, line 63: State layer needs this every single time, will remove the optional

Sources/StreamChat/Controllers/DatabaseObserver/StateLayerListDatabaseObserver.swift
laevandus: I need to add `ListChange` as well since too much depends on it. I will play around with this a bit.

Sources/StreamChat/Controllers/DatabaseObserver/StateLayerListDatabaseObserver.swift
```diff
@@ -0,0 +1,111 @@
...
+//
+
+import CoreData
+import Foundation
+
+/// A CoreDate store observer which immediately reports changes as soon as the store has been changed.
+///
+/// List changes are reported using item ids to reduce the burden of converting DTOs to models.
+@available(iOS 13.0, *)
+final class StateLayerListDatabaseObserver<Item, ItemID, DTO: NSManagedObject> {
```
laevandus, line 12: While writing tests I encountered an issue where this did not work:
```swift
try await channelList.loadChannels(with: pagination)
let channels = channelList.state.channels // <-- new channels were not returned
```

The core issue was that if I change the local store using `writableContext` but I observe the store using the `backgroundContext`, then there is a slight delay before `backgroundContext` is notified about the store change. Moreover, `BackgroundListDatabaseObserver` what I was using before has internal operation queue which delays the change even more. I also looked into using `ListDatabaseObserver` but it does some unnecessary model conversions the state layer does not need (see ChangeAggregator which converts DTOs to models). Also there was an opportunity to skip `performAndWait` and `@Atomic` locking when FRC reported changes. Thirdly, I was able to hide bring in async callback to this level and do the StreamCollection conversion here instead.
The cost is that we have another database observer.

Tests/StreamChatTests/StateLayer/ChannelList_Tests.swift
```diff
@@ -0,0 +1,293 @@
...
+        let cancellable = channelList.state.$channels
+            .dropFirst() // ignore initial
+            .sink { channels in
+                XCTAssertEqualIgnoringOrder(incomingChannelListPayload.channels.map(\.channel.cid.rawValue), channels.map(\.cid.rawValue))
+                expectation.fulfill()
+            }
+        try await env.client.mockDatabaseContainer.write { session in
+            session.saveChannelList(payload: incomingChannelListPayload, query: self.channelList.query)
+        }
+        #if swift(>=5.8)
```
martinmitrevski, line 176: I think it's best if we just use the old `wait(for: expectation)` API. We shouldn't bring complexity to the tests with if statements, wdyt?

TestTools/StreamChatTestTools/Extensions/XCTest+Helpers.swift
```diff
@@ -23,6 +23,14 @@ public func XCTAssertEqual<T: Equatable>(_ expected: T,
     }
 }
 
+/// Asserts diff between the expected and received collections after sorting both collections
+public func XCTAssertEqualIgnoringOrder<T: RandomAccessCollection>(_ expected: T,
```
martinmitrevski, line 27: Why do we need this? Shouldn't the order of the collections be the same?

Sources/StreamChat/StateLayer/ChannelListState.swift
```diff
@@ -30,20 +30,31 @@ public final class ChannelListState: ObservableObject {
...
+    
     // MARK: - Mutating the State
     
     @MainActor func value<Value>(forKeyPath keyPath: KeyPath<ChannelListState, Value>) -> Value {
         self[keyPath: keyPath]
     }
     
-    @MainActor func setValue<Value>(_ value: Value, for keyPath: ReferenceWritableKeyPath<ChannelListState, Value>) {
-        self[keyPath: keyPath] = value
+    @MainActor private func noteChannelsDidChange(_ channels: StreamCollection<ChatChannel>, _ changes: [ListChange<ChannelId>]) {
```
martinmitrevski, line 56: What does "note" mean in this context? Maybe "handle" is more appropriate?

Tests/StreamChatTests/StateLayer/ChannelList_Tests.swift
martinmitrevski: discussed internally, disregard it for now

Sources/StreamChat/Controllers/DatabaseObserver/StateLayerListDatabaseObserver.swift
```diff
@@ -0,0 +1,111 @@
...
+                    resultItem = try itemCreator(dto)
+                } catch {
+                    log.assertionFailure("Unable to convert a DB entity to model: \(error.localizedDescription)")
+                }
+                return resultItem
+            },
+            context: context
+        )
+        if !sorting.isEmpty {
+            let sorted = Array(result).sort(using: sorting)
```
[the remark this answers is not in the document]
laevandus: This is copy-paste from the `ListDatabaseObserver`. We should tackle this separately. I think only `ChannelListQuery` uses the runtime sorting outside of FRC.

Sources/StreamChat/StateLayer/ChannelList.swift
```diff
@@ -13,7 +13,7 @@ public class ChannelList {
     private let channelListUpdater: ChannelListUpdater
     
     init(
-        channels: [ChatChannel],
+        initialChannels: [ChatChannel]?,
```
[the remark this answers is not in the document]
laevandus: Both are fine with me, I'll change it back.

TestTools/StreamChatTestTools/Extensions/XCTest+Helpers.swift
laevandus: I can remove it and add more pre-defined dummy data which makes sure the final result is correctly sorted

Sources/StreamChat/StateLayer/ChannelListState.swift
laevandus: Renamed

Sources/StreamChat/Controllers/DatabaseObserver/StateLayerListDatabaseObserver.swift
martinmitrevski: ok, let's merge it in this state, but add it to the list. I think we should try a separate context.

Sources/StreamChat/Controllers/DatabaseObserver/StateLayerListDatabaseObserver.swift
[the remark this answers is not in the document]
martinmitrevski: probably we should also re-use the same logic, instead of copy-pasting it. Maintaining it could prove difficult.

martinmitrevski (approved):
Let's merge it, but pls keep track of the open points.

Sources/StreamChat/Controllers/DatabaseObserver/StateLayerListDatabaseObserver.swift
laevandus: I'll add it as a separate TODO and will try it in a separate PR. If it does not work, tests will fail immediately.

Sources/StreamChat/Controllers/DatabaseObserver/StateLayerListDatabaseObserver.swift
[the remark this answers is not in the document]
laevandus: Hmm, I could turn it into a single convenience init and reuse it. I'll add it as a TODO in notion and try it out in a separate PR.

PR #3126, opened 2024-04-04
[ChatStateLayer] Factory methods

laevandus:
### 🔗 Issue Links

### 🎯 Goal

Make it easier to use ConnectedUser and ChannelList factory methods.

### 📝 Summary

- ConnectedUser: MainActor + throws
- ChannelList: no forced server sync + no async + no throwing

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/StateLayer/ChatClient+Factory.swift
+++ b/Sources/StreamChat/StateLayer/ChatClient+Factory.swift
@@ -4,6 +4,12 @@
 
 import Foundation
 
+/*
+ The naming for methods here follow rules:
+ - `make`: creates an instance which represents a local state (server state needs to be explicitly fetched)
+ - `retrieve`: creates an instance which syncs with server and returns the latest state while also creating an entity on the server if needed
+ */
+
 // MARK: - Factory Methods for Currently Logged In User
 
 @available(iOS 13.0, *)
@@ -11,8 +17,8 @@ extension ChatClient {
     /// Creates an instance of ``ConnectedUser`` which represents the logged-in user state and its actions.
     ///
     /// - Throws: An error if no user is currently logged-in.
-    public func makeConnectedUser() async throws -> ConnectedUser {
-        let user = try await databaseContainer.read { try CurrentUserDTO.load(context: $0) }
+    @MainActor public func makeConnectedUser() throws -> ConnectedUser {
+        let user = try CurrentUserDTO.load(context: databaseContainer.viewContext)
         return ConnectedUser(user: user, client: self)
     }
 }
@@ -27,6 +33,7 @@ extension ChatClient {
     /// Refer to [querying channels in Stream documentation](https://getstream.io/chat/docs/ios-swift/query_channels/?language=swift) for additional details.
     ///
     /// - Note: Only channels that the user can read are returned, therefore, make sure that the query uses a filter that includes such logic. It is recommended to include a members filter which includes the currently logged in user (e.g. `.containMembers(userIds: ["thierry"])`).
+    /// - Important: This returns a channel list with the latest state from the server.
     ///
     /// - Parameters:
     ///   - query: The query specifies which channels are part of the list and how channels are sorted.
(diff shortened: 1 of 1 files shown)


Sources/StreamChat/StateLayer/ChatClient+Factory.swift
```diff
@@ -34,9 +34,8 @@ extension ChatClient {
...
     /// - Throws: An error while communicating with the Stream API.
     /// - Returns: An instance of ``ChannelList`` which represents actions and the current state of the list.
-    public func makeChannelList(with query: ChannelListQuery, dynamicFilter: ((ChatChannel) -> Bool)? = nil) async throws -> ChannelList {
-        let channels = try await channelListUpdater.update(channelListQuery: query)
-        let channelList = ChannelList(initialChannels: channels, query: query, dynamicFilter: dynamicFilter, channelListUpdater: channelListUpdater, client: self)
+    public func makeChannelList(with query: ChannelListQuery, dynamicFilter: ((ChatChannel) -> Bool)? = nil) -> ChannelList {
+        let channelList = ChannelList(query: query, dynamicFilter: dynamicFilter, channelListUpdater: channelListUpdater, client: self)
         syncRepository.trackChannelListQuery { [weak channelList] in channelList?.state.query }
         return channelList
     }
```
laevandus, line 41: - Drops `async` and `throws` 
- Server sync is not forced, one needs to use `loadChannels()` function for fetching server state

Sources/StreamChat/StateLayer/ChatClient+Factory.swift
```diff
@@ -11,8 +11,8 @@ extension ChatClient {
     /// Creates an instance of ``ConnectedUser`` which represents the logged-in user state and its actions.
     ///
     /// - Throws: An error if no user is currently logged-in.
-    public func makeConnectedUser() async throws -> ConnectedUser {
-        let user = try await databaseContainer.backgroundRead { try CurrentUserDTO.load(context: $0) }
+    @MainActor public func makeConnectedUser() throws -> ConnectedUser {
```
laevandus, line 14: - Drops `async`, but keeps `throws`
- Requires `@MainActor` because of `viewContext`

Sources/StreamChat/StateLayer/ChatClient+Factory.swift
martinmitrevski: We need to think this one through. Maybe we keep both options? I can image customers forgetting to call `loadChannels` and complaining about why they don't get the latest data.

laevandus: Could happen. I'll think about naming here.

laevandus: Not sure if there is better naming, open for ideas.
```
let a = makeChannelList(with: ChannelListQuery(filter: .noTeam))
let b = try await makeChannelList(refreshing: ChannelListQuery(filter: .noTeam))
```

nuno-vieira: The `make` functions should never make any calls. If they do, it will be misleading. Factory methods should not have any side effects. Just create a function.

Usually, you don't want it to perform any calls initially. You want to create it, put it in a view, and only when the view is loaded do you perform a call. 

If for some reason, we want to provide a way to create a channel list, and at the same time initialise the loading, we should name it something else then `make`. Although I don't see the need having this TBH.

nuno-vieira: For me, this is what makes more sense:

> Drops async and throws
Server sync is not forced, one needs to use loadChannels() function for fetching server state

laevandus: Sounds like I should take a step back, look at the ChatClient+Factory as a whole, and come back with a proposal for all the methods we have there.

Channel related methods do require sync with a server in some cases so I need to think about it. Currently these are also named as `makeXYZ`.

Something along the lines:
- `makeXYZ` creates the object, does not do any refresh calls
- `somethingSomething` syncs with server and then returns the object

I'll look around in Apple APIs etc for inspiration for the second case.

laevandus: Seems like there are 3 cases:
1. Give me an instance which represents the local state (sync)
2. Give me an instance which represents the server state (async, throws)
3. Create something new on the server if needed (async, throws)

martinmitrevski: For me `load` makes most sense. In any case, we should also expose a `watch` method and be consistent with the other SDKs: https://getstream.io/chat/docs/android/watch_channel/?language=kotlin&q=watch

laevandus: `Chat.watch()` is there. 
Naming is hard since some of the methods actually create a new thing on the backend side (direct channel/chat case).

martinmitrevski: Actually, we can use `get` and `getOrCreate`. We use both on video, and also the API is using this convention.

`chat.get` and `chat.getOrCreate` - I think this is the best one.

laevandus: Concrete examples:

**make+retrieve**
```swift
// Give me channel list with the local state
func makeChannelList(…) -> ChannelList

// Give me channel list with the server state
func retrieveChannelList(…) async throws -> ChannelList

// Create a new direct message channel or if it exists, give me the one we already have
func retrieveDirectMessageChat(…) async throws -> Chat
```

**make+fetch**
```swift
// Give me channel list with the local state
func makeChannelList(…) -> ChannelList

// Give me channel list with the server state
func fetchChannelList(…) async throws -> ChannelList

// Create a new direct message channel or if it exists, give me the one we already have
func fetchDirectMessageChat(…) async throws -> Chat
```

**get+getOrCreate**
```swift
// Give me channel list with the local state
func getChannelList(…) -> ChannelList

// Give me channel list with the server state
func getOrCreateChannelList(…) async throws -> ChannelList

// Create a new direct message channel or if it exists, give me the one we already have
func getOrCreateDirectMessageChat(…) async throws -> Chat
```

martinmitrevski: no, it should be sth like this:
```
let chat = client.makeChat(...) // sync
chat.get()
chat.getOrCreate()
```
Check the call docs in video: https://getstream.io/video/docs/ios/guides/joining-creating-calls/

laevandus: This would be going back to `synchronize()` and having channel id check in every method call.

nuno-vieira: other SDKs seem to use `queryChannelList()` and `queryChannel()` right? There seems to exist a lot of different patterns in the SDK and we should start to align this now. 

For Chat:
- makeChat()
- query() / fetch() / get()
- watch()

For ChannelList:
- makeChannelList()
- query() / fetch() / get()
- watch()
- queryOrCreate() / fetchOrCreate() / getOrCreate()

But since @martinmitrevski is saying that `get` is what it is used on video, then the get is not the preferrable way?

laevandus:
I am going to close this since we will discuss this in a separate document before implementing it.

PR #3127, opened 2024-04-04
Fix thread parent message not updated when inside a thread

nuno-vieira:
### 🔗 Issue Links
Resolves https://github.com/GetStream/ios-issues-tracking/issues/784

### 🎯 Goal
It fixes the Thread Parent Message not being updated while in the Thread Screen.

### 📝 Summary
The issue happens because, weirdly, the `MessageController.repliesObserver` is reporting the parent message inside the replies, even though the `threadRepliesFetchRequest()` discards the parent message. Writing the predicate `id != parentMessageId` also does not fix it; it still returns the parent message inside the replies, especially when the parent message is updated.

This is why the `ChatThreadVC.getReplies()` has the logic that it has. If it includes the parent message (which should never happen in the first place), we update it, if not, we don't do anything. Before, we were not updating the parent message, so that is why we had this bug. But this should not be required to do, in SwiftUI it works, and it uses the same LLC SDK version. So, something is missing here or it is just a random bug from CoreData. 

### 🧪 Manual Testing Notes

1. Disable Background Mapping
2. Open a thread
3. Add a reaction to the root parent message
4. It should update

### ☑️ Contributor Checklist


Sources/StreamChatUI/ChatThread/ChatThreadVC.swift
```diff
@@ -445,12 +445,16 @@ open class ChatThreadVC: _ViewController,
...
+        var messages = messageController.replies
         let isFirstPage = messages.count < messageController.repliesPageSize
         let shouldAddRootMessageAtTheTop = isFirstPage || messageController.hasLoadedAllPreviousReplies
         if shouldAddRootMessageAtTheTop, let threadRootMessage = messageController.message {
             if messages.last?.id != threadRootMessage.id {
                 messages.append(threadRootMessage)
+            } else if messages.last?.id == threadRootMessage.id {
+                // When the last message is already the thread root message, update it.
+                let lastIndex = messages.count - 1
+                messages[lastIndex] = threadRootMessage
```
nuno-vieira, line 457: This does not make much sense, we are checking if the `messages.last?.id` is the same as the `threadRootMessage.id`, but in fact, `messages.last?.id` in this case is the same as doing `messageController.replies.last?.id`. The replies should never include the parent message according to `MessageDTO.threadRepliesPredicate()`, but somehow, with BG Mapping disabled, the thread parent message is included in the replies when `didUpdateMessage` delegate is reported.

martinmitrevski: weird 🤔 Anyway, no harm in adding this

martinmitrevski (approved):

Sources/StreamChatUI/ChatThread/ChatThreadVC.swift
laevandus: This seem to happen because we are inserting the parent message to messageController.replies here on this line
```if messages.last?.id != threadRootMessage.id {
    messages.append(threadRootMessage)
}```
messages here is `LazyCachedMapCollection<ChatMessage>`  and turns out that `LazyCachedMapCollection` has an append method which updates cached values.

```public func append(_ element: Element) {
  cache.values.append(element)
}```
`LazyCachedMapCollection`  in turn uses that cache property to return what it represents,  in this case: what FRC returned + the value we added ourselves.

nuno-vieira: OMG, did not expect `cache` was a reference type... 😭 Will close this ticket, since the solution does not make sense anymore

PR #3129, opened 2024-04-05
Fix thread parent message not updated when inside a thread

nuno-vieira:
### 🔗 Issue Links
Resolves https://github.com/GetStream/ios-issues-tracking/issues/784

### 🎯 Goal
It fixes the Thread Parent Message not being updated while in the Thread Screen.

### 📝 Summary
In the UIKit SDK, we were `appending` the parent message to the replies in `ChatThreadVC.getReplies()`, the problem was that the `LazyCachedMapCollection.append()` function was mutating the `cache` instance directly. So even if we did this:
```swift
var messages = messageController.replies
```
It would not create a copy, because even though `LazyCachedMapCollection` is a value type, the `LazyCachedMapCollection.cache` is a reference type. So this would append the parent message to the `ChatMessageController` directly.

**Solution:**
The previous `LazyCachedMapCollection.append()`implementation was dangerous and misleading, so this PR makes sure that we actually mutate and create a new copy of `LazyCachedMapCollection` when appending a new element.

### 🧪 Manual Testing Notes

1. Disable Background Mapping
2. Open a thread
3. Add a reaction to the root parent message
4. It should update

### ☑️ Contributor Checklist


Sources/StreamChat/Utils/LazyCachedMapCollection.swift
```diff
@@ -71,7 +82,11 @@ public struct LazyCachedMapCollection<Element>: RandomAccessCollection {
         }
     }
 
-    public func append(_ element: Element) {
+    public mutating func append(_ element: Element) {
```
nuno-vieira, line 85: In theory this is a breaking change, but I'm 100% sure nobody is using this, if they do, it is a bug, so they need to be notified that appending a new element, is a mutating func and it will create a new copy.

laevandus (approved):
Great change! We definitely want to add `mutating` to the append method since it was incorrect before.

martinmitrevski (approved):

PR #3140, opened 2024-04-10
[ChatStateLayer] DB observing improvements & tests for MessageState

laevandus:
### 🔗 Issue Links

### 🎯 Goal

- Basic test coverage for MessageState

### 📝 Summary

- Tests for MessageState
- Changed database observer to react more quickly (I saw that it switched to global executor and then to main when it was not needed)
- Changed the DB read method to use the same state layer context

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/Controllers/DatabaseObserver/StateLayerDatabaseObserver.swift
+++ b/Sources/StreamChat/Controllers/DatabaseObserver/StateLayerDatabaseObserver.swift
@@ -60,16 +60,23 @@ extension StateLayerDatabaseObserver where ResultType == EntityResult {
         return item
     }
     
+    /// Starts observing the database and dispatches changes on the MainActor.
+    ///
+    /// - Important: Does not send the initial value, use the ``item`` for setting the initial value.
     func startObserving(didChange: @escaping (Item?) async -> Void) throws {
-        try startObserving(didChange: { item in Task(priority: .high) { await didChange(item) } })
+        try startObserving(onContextDidChange: { item in Task.mainActor { await didChange(item) } })
     }
     
-    func startObserving(didChange: @escaping (Item?) -> Void) throws {
+    /// Starts observing the database and dispatches changes on the NSManagedObjectContext's queue.
+    ///
+    /// - Note: Use it if we need to do additional processing on the context's queue.
+    /// - Important: Does not send the initial value, use the ``item`` for setting the initial value.
+    func startObserving(onContextDidChange: @escaping (Item?) -> Void) throws {
         resultsDelegate = FetchedResultsDelegate(onDidChange: { [weak self] in
             guard let self else { return }
             // Runs on the NSManagedObjectContext's queue, therefore skip performAndWait
             let item = Self.makeEntity(frc: self.frc, context: self.context, itemCreator: self.itemCreator, sorting: self.sorting)
-            didChange(item)
+            onContextDidChange(item)
         })
         frc.delegate = resultsDelegate
         try frc.performFetch()
@@ -107,16 +114,23 @@ extension StateLayerDatabaseObserver where ResultType == ListResult {
         return collection
     }
     
+    /// Starts observing the database and dispatches changes on the MainActor.
+    ///
+    /// - Important: Does not send the initial value, use the ``item`` for setting the initial value.
     func startObserving(didChange: @escaping (StreamCollection<Item>) async -> Void) throws {
-        try startObserving(didChange: { items in Task(priority: .high) { await didChange(items) } })
+        try startObserving(onContextDidChange: { items in Task.mainActor { await didChange(items) } })
     }
     
-    func startObserving(didChange: @escaping (StreamCollection<Item>) -> Void) throws {
+    /// Starts observing the database and dispatches changes on the NSManagedObjectContext's queue.
+    ///
+    /// - Note: Use it if we need to do additional processing on the context's queue.
+    /// - Important: Does not send the initial value, use the ``item`` for setting the initial value.
+    func startObserving(onContextDidChange: @escaping (StreamCollection<Item>) -> Void) throws {
         resultsDelegate = FetchedResultsDelegate(onDidChange: { [weak self] in
             guard let self else { return }
             // Runs on the NSManagedObjectContext's queue, therefore skip performAndWait
             let collection = Self.makeCollection(frc: self.frc, context: self.context, itemCreator: self.itemCreator, sorting: self.sorting)
-            didChange(collection)
+            onContextDidChange(collection)
         })
         frc.delegate = resultsDelegate
         try frc.performFetch()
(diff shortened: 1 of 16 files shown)


Sources/StreamChat/Controllers/DatabaseObserver/StateLayerDatabaseObserver.swift
```diff
@@ -107,16 +114,23 @@ extension StateLayerDatabaseObserver where ResultType == ListResult {
         return collection
     }
     
+    /// Starts observing the database and dispatches changes on the MainActor.
+    ///
+    /// - Important: Does not send the initial value, use the ``item`` for setting the initial value.
     func startObserving(didChange: @escaping (StreamCollection<Item>) async -> Void) throws {
-        try startObserving(didChange: { items in Task(priority: .high) { await didChange(items) } })
+        try startObserving(onContextDidChange: { items in Task.mainActor { await didChange(items) } })
```
martinmitrevski, line 121: can you expand a bit more on this change?

Sources/StreamChat/StateLayer/MessageState+Observer.swift
```diff
@@ -27,29 +27,31 @@ extension MessageState {
...
+                var lastSortedReactions: [ChatMessageReaction]?
+                try messageObserver.startObserving(onContextDidChange: { message in
                     guard let message else { return }
-                    let currentReactions = message.latestReactions
-                    if lastReactions != currentReactions {
-                        lastReactions = currentReactions
-                        let sortedReactions = currentReactions.sorted(by: { $0.updatedAt > $1.updatedAt })
-                        await handlers.reactionsDidChange(sortedReactions)
+                    let changedReactions: [ChatMessageReaction]?
+                    let currentReactions = message.latestReactions.sorted(by: MessageState.reactionsSorting)
```
martinmitrevski, line 45: Reactions should also be sorted by the customers. But I assume we do this in the UI components?
This will also change a bit when we have reactions v2.

Sources/StreamChat/StateLayer/MessageState.swift
```diff
@@ -7,31 +7,26 @@ import Foundation
...
 public final class MessageState: ObservableObject {
-    private let chat: Chat
     private let messageOrder: MessageOrdering
     private let observer: Observer
 
     let replyPaginationHandler: MessagesPaginationStateHandling
     
-    init(message: ChatMessage, chat: Chat, messageOrder: MessageOrdering, database: DatabaseContainer, clientConfig: ChatClientConfig, replyPaginationHandler: MessagesPaginationStateHandling) {
-        self.chat = chat
+    init(message: ChatMessage, messageOrder: MessageOrdering, database: DatabaseContainer, clientConfig: ChatClientConfig, replyPaginationHandler: MessagesPaginationStateHandling) {
```
martinmitrevski, line 15: The linter should be complaining on this formatting, I think

Tests/StreamChatTests/StateLayer/MessageState_Tests.swift
```diff
@@ -0,0 +1,223 @@
...
+    }
+    
+    // MARK: - Observing Reactions
+    
+    func test_restoringReactions_whenReactionsStored_thenInitialStateIsSet() async throws {
+        let messagePayload = makeMessagePayload(reactionCount: 3, messageId: messageId)
+        try await env.client.databaseContainer.write { session in
+            try session.saveMessage(payload: messagePayload, for: self.channelId, syncOwnReactions: true, cache: nil)
+            // Unrelated
+            try session.saveMessage(payload: self.makeMessagePayload(reactionCount: 3, messageId: self.unrelatedMessageId), for: self.channelId, syncOwnReactions: true, cache: nil)
```
martinmitrevski, line 59: for methods with many params, we should move the params in a separate line:
```swift
saveMessage(
    payload: messagePayload,
    ....
)
```

Tests/StreamChatTests/StateLayer/MessageState_Tests.swift
```diff
@@ -0,0 +1,223 @@
...
+        XCTAssertEqual(LocalMessageState.sendingFailed, await messageState.message.localState)
+    }
+    
+    // MARK: - Observing Reactions
+    
+    func test_restoringReactions_whenReactionsStored_thenInitialStateIsSet() async throws {
+        let messagePayload = makeMessagePayload(reactionCount: 3, messageId: messageId)
+        try await env.client.databaseContainer.write { session in
+            try session.saveMessage(payload: messagePayload, for: self.channelId, syncOwnReactions: true, cache: nil)
+            // Unrelated
```
martinmitrevski, line 58: A bit of explanation here would be nice (also in other places).

martinmitrevski (commented):
Looks good 👍  Added few comments and questions

Sources/StreamChat/StateLayer/MessageState+Observer.swift
laevandus: For now I just preserved what we are doing in the `UserSearchController`

Sources/StreamChat/Controllers/DatabaseObserver/StateLayerDatabaseObserver.swift
laevandus: 95% of cases the observer's callback is forwarded to `XYZState` which runs the mutation on the main actor. With the old code what happened was:

1. We are on the NSManagedObjectContext's thread, we run the closure
2. The closure creates a Task and starts running it on the global executor aka background thread
3. didChange is called on the global executor which ends up calling some function in the `XYZState` which requires main actor and the handling jumps to the main thread

With this change, we skip the global executor context switch. Just an optimization.

In rare cases, I want to run additional code in the didChange before forwarding the data to `XYZState`. For these cases one can use `func startObserving(onContextDidChange` which is synchronous and runs on the NSManagedObjectContext's queue (e.g. I might want to do some additional processing. Example is reading reactions and sorting them on the NSManageObjectContext's queue). The other case is, unit-tests.

Sources/StreamChat/StateLayer/MessageState.swift
laevandus: Done

Tests/StreamChatTests/StateLayer/MessageState_Tests.swift
laevandus: Done, reviewed other cases as well in the changed code

martinmitrevski (approved):

Sources/StreamChat/Controllers/DatabaseObserver/StateLayerDatabaseObserver.swift
laevandus: If it goes through and we introduce `@MainActor` state objects then this actually opens up an easy way to use initial argument in the observer.

PR #3141, opened 2024-04-10
[ChatStateLayer] User search results were incorrect

laevandus:
### 🔗 Issue Links

### 🎯 Goal

User search did not work at all

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/StateLayer/UserSearch.swift
+++ b/Sources/StreamChat/StateLayer/UserSearch.swift
@@ -61,14 +61,9 @@ public class UserSearch {
     
     private func search(query: UserListQuery, pagination: Pagination) async throws -> [ChatUser] {
         let query = query.withPagination(pagination)
-        let task = Task {
-            let users = try await userListUpdater.fetch(userListQuery: query, pagination: pagination)
-            try Task.checkCancellation()
-            return users
-        }
-        await state.setActiveTask(task, query: query)
-        let users = try await task.value
-        await state.setUsers(users, for: query, pagination: pagination)
+        await state.handleStartingFetchingQuery(query)
+        let users = try await userListUpdater.fetch(userListQuery: query, pagination: pagination)
+        await state.handleFinishedFetchingQuery(query, users: users)
         return users
     }
 }
--- a/Sources/StreamChat/StateLayer/UserSearch.swift
+++ b/Sources/StreamChat/StateLayer/UserSearch.swift
@@ -61,9 +61,9 @@ public class UserSearch {
     
     private func search(query: UserListQuery, pagination: Pagination) async throws -> [ChatUser] {
         let query = query.withPagination(pagination)
-        await state.handleStartingFetchingQuery(query)
+        await state.setQuery(query)
         let users = try await userListUpdater.fetch(userListQuery: query, pagination: pagination)
-        await state.handleFinishedFetchingQuery(query, users: users)
+        await state.handleDidFetchQuery(query, users: users)
         return users
     }
 }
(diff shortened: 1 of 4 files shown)


Sources/StreamChat/StateLayer/UserSearchState.swift
```diff
@@ -27,23 +23,40 @@ extension UserSearchState {
...
+            result = StreamCollection(incomingUsers)
         } else {
-            var result = Array(users[..<pagination.offset])
-            result.append(contentsOf: newUsers)
-            users = StreamCollection(result)
+            // Filter and sorting are the same but incoming users might contain duplicates (depends how pagination is used)
+            let incomingIds = Set(incomingUsers.map(\.id))
+            let incomingRemovedUsers = users.filter { !incomingIds.contains($0.id) }
+            let sortValues = completedQuery.sort.map(\.sortValue)
+            result = StreamCollection((incomingRemovedUsers + incomingUsers).sort(using: sortValues))
```
nuno-vieira, line 49: How can they contain duplicates? 🤔

Sources/StreamChat/StateLayer/UserSearch.swift
```diff
@@ -61,14 +61,9 @@ public class UserSearch {
...
     private func search(query: UserListQuery, pagination: Pagination) async throws -> [ChatUser] {
         let query = query.withPagination(pagination)
-        let task = Task {
-            let users = try await userListUpdater.fetch(userListQuery: query, pagination: pagination)
-            try Task.checkCancellation()
-            return users
-        }
-        await state.setActiveTask(task, query: query)
-        let users = try await task.value
-        await state.setUsers(users, for: query, pagination: pagination)
```
nuno-vieira, line 71: Can you clarify in the PR description why this did not work? 🙏

laevandus: The problem with this one is if you write code like this in the UI:
```swift
state = .loading
do {
  try await userSearch.search(term: "name")
  state = .idle
} catch {
  state = .failure(error)
}
```
Then `Task.checkCancellation()` throws an error when the previous request has been cancelled. Then one needs to explicitly ignore the cancellation which feels like might cause too many support headache.

So for now, I just removed the cancellation part. Downside is that, we do not cancel requests what are irrelevant.

Since the search method also returns `[ChatUser]` then if felt incorrect to internally cancel and return an empty array.

Open for ideas. Some options:
1. We could add documentation about the cancellation and still cancel old requests
2. No cancellation, just discarding irrelevant query results

Sources/StreamChat/StateLayer/UserSearchState.swift
laevandus: If you call `loadNextUsers` multiple times.

Sources/StreamChat/StateLayer/UserSearchState.swift
```diff
@@ -27,23 +23,40 @@ extension UserSearchState {
...
+    @MainActor private func setValue<Value>(_ value: Value, for keyPath: ReferenceWritableKeyPath<UserSearchState, Value>) {
+        self[keyPath: keyPath] = value
+    }
         
-    @MainActor func setActiveTask(_ task: Task<[ChatUser], Error>, query: UserListQuery) {
-        if let activeTask {
-            activeTask.cancel()
-        }
-        activeTask = task
+    @MainActor func handleStartingFetchingQuery(_ query: UserListQuery) {
```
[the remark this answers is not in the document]
laevandus: 1. Setting `state.query` as soon as we start the query
2. Using `state.query` for filtering out results for API requests which come in later than new requests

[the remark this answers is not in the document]
laevandus: I renamed it

Sources/StreamChat/StateLayer/UserSearch.swift
martinmitrevski: As discussed on the tech meeting - I would go with 3) for now - request cancellation is a big topic, that ideally we would like to handle on the API client level.

PR #3144, opened 2024-04-11
[ChatStateLayer] Require main actor in all the state objects

laevandus:
### 🔗 Issue Links

### 🎯 Goal

Use @MainActor for all the `XYZState` objects but keep their owning objects without main actor isolation.

### ☑️ Contributor Checklist


Sources/StreamChat/Utils/StateBuilder.swift
```diff
@@ -0,0 +1,24 @@
+//
+// Copyright © 2024 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+/// A builder for objects requiring @MainActor.
+final class StateBuilder<State> {
```
laevandus, line 8: I added this helper class for enabling creating state objects on the main actor.

The flow is that you create this in non-isolated context:
```swift
stateBuilder = StateBuilder {
  environment.blabla
```

And it is triggered when we access `.state` for the first time and `.state` requires @MainActor. 
`@MainActor public lazy var state: MessageSearchState = stateBuilder.build()`

This was the only good way to workaround the case of, for example,
ChatState is @MainActor but its owning type, Chat, is non-isolated.

Sources/StreamChat/StateLayer/ChannelList.swift
```diff
@@ -22,19 +20,24 @@ public class ChannelList {
...
+            )
+        }
     }
     
+    /// The query specifying and filtering the list of channels.
+    public let query: ChannelListQuery
+    
     /// An observable object representing the current state of the channel list.
-    public let state: ChannelListState
+    @MainActor public lazy var state: ChannelListState = stateBuilder.build()
```
martinmitrevski, line 40: can we avoid it being lazy?

Sources/StreamChat/StateLayer/ChannelListState.swift
```diff
@@ -41,14 +39,4 @@ public final class ChannelListState: ObservableObject {
     
     /// An array of channels for the specified ``ChannelListQuery``.
     @Published public internal(set) var channels = StreamCollection<ChatChannel>([])
-    
-    // MARK: - Mutating the State
-    
-    @MainActor func value<Value>(forKeyPath keyPath: KeyPath<ChannelListState, Value>) -> Value {
```
martinmitrevski, line 47: Removing these is a big benefit 👍

Sources/StreamChat/Utils/StateBuilder.swift
```diff
@@ -0,0 +1,24 @@
...
+    private var builder: (@MainActor() -> State)?
+    
+    init(builder: (@escaping @MainActor() -> State)) {
+        self.builder = builder
+    }
+    
+    @MainActor func build() -> State {
+        guard let builder else {
+            fatalError("Calling build multiple times for \(State.self)")
```
martinmitrevski, line 17: No fatal errors please 😅  We avoid crashing the SDK by ourselves.

martinmitrevski (commented):
Looks great 👍  It would be great if we can get rid of the laziness of the state. Additionally, we shouldn't use fatal errors.

Sources/StreamChat/StateLayer/ChannelListState.swift
laevandus: Yeah, that is a really nice change

Sources/StreamChat/Utils/StateBuilder.swift
laevandus: I will need to think about this one a bit. I only added the part of setting it to nil since then the closure can release captured values. Hmm.

laevandus: I found an alternative solution as well. Would avoid having this type at all.

laevandus: Alternative did not work, I deleted the optional part from there since most of the time we capture things which are alive all the time anyway.

Sources/StreamChat/StateLayer/ChannelList.swift
laevandus: I had an [alternative idea](https://getstream.slack.com/archives/CRB9P6L03/p1712903579989979?thread_ts=1712901708.377809&cid=CRB9P6L03) as well but it did not work for all the cases. So at the moment, I don't have ideas which would avoid using lazy var for state. `@MainActor` and lazy var are currently just tools for pushing the state creation to the main actor. Tricky

martinmitrevski (approved):

PR #3151, opened 2024-04-16
[ChatStateLayer] Add test coverage for ConnectedUser

laevandus:
### 🔗 Issue Links

### 🎯 Goal

Tests for ConnectedUser

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/StateLayer/ConnectedUserState.swift
+++ b/Sources/StreamChat/StateLayer/ConnectedUserState.swift
@@ -17,6 +17,9 @@ import Foundation
             with: .init(
                 userDidChange: { [weak self] in self?.user = $0 })
         )
+        if let user = observer.userObserver.item {
+            self.user = user
+        }
     }
     
     /// The represented user.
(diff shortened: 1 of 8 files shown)


Sources/StreamChat/StateLayer/ConnectedUserState.swift
```diff
@@ -17,6 +17,9 @@ import Foundation
             with: .init(
                 userDidChange: { [weak self] in self?.user = $0 })
         )
+        if let user = observer.userObserver.item {
+            self.user = user
+        }
```
laevandus, line 22: Planning to do a separate PR for cleaning up some of the initial value part. Since I recently made state properties lazy then now it is important to always check the observer's state and not trust the injected user value which is part of the `init` (we might already have changed the user before state has been loaded, I caught this while writing tests).

Sources/StreamChat/Workers/CurrentUserUpdater.swift
```diff
@@ -66,28 +66,32 @@ class CurrentUserUpdater: Worker {
         currentUserId: UserId,
         completion: ((Error?) -> Void)? = nil
     ) {
-        database.write { (session) in
+        database.write({ session in
```
laevandus, line 69: My tests failed because I was relying on database to update before this function returns. But here we did not use DB write's completion handler.

Sources/StreamChat/StateLayer/ConnectedUserState.swift
martinmitrevski: probably we should also check other places

martinmitrevski (approved):
LGTM ✅

PR #3155, opened 2024-04-18
[ChatStateLayer] Rename pagination methods and add new convenience methods

laevandus:
### 🔗 Issue Links

### 🎯 Goal

Rename pagination methods based on what we discussed:

- Use `more` instead of `next`
- Use `older` and `newer` instead of `previous` and `next` for messages

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/StateLayer/Chat.swift
+++ b/Sources/StreamChat/StateLayer/Chat.swift
@@ -373,7 +373,7 @@ public class Chat {
     
     /// Loads messages for the specified pagination parameters and updates ``ChatState/messages``.
     ///
-    /// - Important: Loading messages for the pagination index 0 resets ``ChatState/messages``.
+    /// - Important: Loading messages for the pagination with parameter nil resets ``ChatState/messages``. Example `chat.loadMessages(with: MessagesPagination(pageSize: 30, parameter: nil))`.
     ///
     /// - Parameters:
     ///   - message: The parent message id which has replies.
@@ -685,9 +685,9 @@ public class Chat {
     ///
     /// - Throws: An error while communicating with the Stream API.
     /// - Returns: An array of messages for the pagination.
-    @discardableResult public func loadReplies(of messageId: MessageId, pagination: MessagesPagination) async throws -> [ChatMessage] {
+    @discardableResult public func loadReplies(for messageId: MessageId, pagination: MessagesPagination) async throws -> [ChatMessage] {
         let messageState = try await messageState(for: messageId)
-        return try await messageUpdater.loadReplies(of: messageId, pagination: pagination, cid: cid, paginationStateHandler: messageState.replyPaginationHandler)
+        return try await messageUpdater.loadReplies(for: messageId, pagination: pagination, cid: cid, paginationStateHandler: messageState.replyPaginationHandler)
     }
 
     // MARK: -
@@ -700,9 +700,9 @@ public class Chat {
     ///   - limit: The limit for the page size. The default limit is 25.
     ///
     /// - Throws: An error while communicating with the Stream API.
-    public func loadReplies(before replyId: MessageId?, of parentMessageId: MessageId, limit: Int? = nil) async throws {
+    public func loadReplies(before replyId: MessageId?, for parentMessageId: MessageId, limit: Int? = nil) async throws {
         let messageState = try await messageState(for: parentMessageId)
-        return try await messageUpdater.loadReplies(of: parentMessageId, before: replyId, limit: limit, cid: cid, paginationStateHandler: messageState.replyPaginationHandler)
+        return try await messageUpdater.loadReplies(for: parentMessageId, before: replyId, limit: limit, cid: cid, paginationStateHandler: messageState.replyPaginationHandler)
     }
     
     /// Loads more replies after the specified reply id and updates ``MessageState/replies``.
@@ -713,9 +713,9 @@ public class Chat {
     ///   - limit: The limit for the page size. The default limit is 25.
     ///
     /// - Throws: An error while communicating with the Stream API.
-    public func loadReplies(after replyId: MessageId?, of parentMessageId: MessageId, limit: Int? = nil) async throws {
+    public func loadReplies(after replyId: MessageId?, for parentMessageId: MessageId, limit: Int? = nil) async throws {
         let messageState = try await messageState(for: parentMessageId)
-        return try await messageUpdater.loadReplies(of: parentMessageId, after: replyId, limit: limit, cid: cid, paginationStateHandler: messageState.replyPaginationHandler)
+        return try await messageUpdater.loadReplies(for: parentMessageId, after: replyId, limit: limit, cid: cid, paginationStateHandler: messageState.replyPaginationHandler)
     }
     
     /// Loads replies around the specified reply id to ``MessageState/replies``.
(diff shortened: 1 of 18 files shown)


Sources/StreamChat/StateLayer/Chat.swift
```diff
@@ -518,7 +539,11 @@ public class Chat {
...
     /// - Returns: An array of pinned messages for the specified pagination.
-    public func loadPinnedMessages(for pagination: PinnedMessagesPagination? = nil, sort: [Sorting<PinnedMessagesSortingKey>] = [], limit: Int = .messagesPageSize) async throws -> [ChatMessage] {
+    public func loadPinnedMessages(
+        with pagination: PinnedMessagesPagination? = nil,
+        sort: [Sorting<PinnedMessagesSortingKey>] = [],
+        limit: Int = .messagesPageSize
+    ) async throws -> [ChatMessage] {
         let query = PinnedMessagesQuery(pageSize: limit, sorting: sort, pagination: pagination)
         return try await channelUpdater.loadPinnedMessages(in: cid, query: query)
     }
```
laevandus, line 549: `PinnedMessagesPagination` has sophisticated options which are not currently represented with convenience methods. I am not sure if it is needed at this point.
```swift
public enum PinnedMessagesPagination: Hashable {
    /// When used, the backend returns messages around the message with the given id.
    case aroundMessage(_ messageId: MessageId)

    /// When used, the backend returns messages pinned earlier then the message with the given id.
    ///
    /// When `inclusive == true` the results include the message with the given id.
    case before(_ messageId: MessageId, inclusive: Bool)

    /// When used, the backend returns messages pinned earlier then the message with the given id.
    ///
    /// When `inclusive == true` the results include the message with the given id.
    case after(_ messageId: MessageId, inclusive: Bool)

    /// When used, the backend returns messages pinned at around the given timestamp.
    case aroundTimestamp(_ timestamp: Date)

    /// When used, the backend returns messages pinned earlier than the given timestamp.
    ///
    /// When `inclusive == true` the results include the message pinned at the given timestamp.
    case earlier(_ timestamp: Date, inclusive: Bool)

    /// When used, the backend returns messages pinned later than the given timestamp.
    ///
    /// When `inclusive == true` the results include the message pinned at the given timestamp.
    case later(_ timestamp: Date, inclusive: Bool)
}
```

Sources/StreamChat/StateLayer/Chat.swift
```diff
@@ -675,57 +700,67 @@ public class Chat {
...
+    /// - Parameters:
+    ///   - replyId: The message id of the reply from which older messages are loaded. If nil, the oldest currently loaded message id in ``MessageState/replies`` is used.
+    ///   - parentMessageId: The parent message id which has replies.
+    ///   - limit: The limit for the page size. The default limit is 25.
+    ///
+    /// - Throws: An error while communicating with the Stream API.
+    public func loadReplies(before replyId: MessageId?, of parentMessageId: MessageId, limit: Int? = nil) async throws {
+        let messageState = try await makeMessageState(for: parentMessageId)
+        return try await messageUpdater.loadReplies(of: parentMessageId, before: replyId, limit: limit, cid: cid, paginationStateHandler: messageState.replyPaginationHandler)
+    }
```
laevandus, line 714: Also considered 
`public func loadReplies(of parentMessageId: MessageId, before replyId: MessageId?, limit: Int? = nil) async throws` but it did not read so well.

Sources/StreamChat/StateLayer/Chat.swift
```diff
@@ -376,54 +386,47 @@ public class Chat {
...
     }
     
     // MARK: -
     
-    /// Loads messages for the first page and updates ``ChatState/messages``.
-    ///
-    /// - Note: Loading the first page resets the ``ChatState/messages``.
-    ///
-    /// - Throws: An error while communicating with the Stream API.
-    public func loadMessagesFirstPage() async throws {
```
martinmitrevski, line 396: Question: What would be the alternative of this one? What did we decide on?

Sources/StreamChat/StateLayer/Chat.swift
martinmitrevski: Best to keep it as it is now - no new convenience methods.

Sources/StreamChat/StateLayer/Chat.swift
```diff
@@ -571,7 +596,7 @@ public class Chat {
     ///
     /// - Throws: An error while communicating with the Stream API.
     /// - Returns: An array of reactions for the next page.
-    @discardableResult public func loadNextReactions(of messageId: MessageId, limit: Int? = nil) async throws -> [ChatMessageReaction] {
+    @discardableResult public func loadMoreReactions(of messageId: MessageId, limit: Int? = nil) async throws -> [ChatMessageReaction] {
```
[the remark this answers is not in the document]
martinmitrevski: Yes, `for` feels better imo.

martinmitrevski (approved):
LGTM 👍

Sources/StreamChat/StateLayer/Chat.swift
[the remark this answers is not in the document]
laevandus: I will change it around, thanks!

Sources/StreamChat/StateLayer/Chat.swift
laevandus: `chat.loadMessages(with: .init(pageSize: n, offset: 0))`

I added a commend to docs in `loadMessages`: `/// - Important: Loading messages for the pagination index 0 resets ``ChatState/messages``.` This is what the first page was doing.

laevandus: Oh, not offset, `parameter: nil`

laevandus: Fixed it:
`/// - Important: Loading messages for the pagination with parameter nil resets ``ChatState/messages``. Example `chat.loadMessages(with: MessagesPagination(pageSize: 30, parameter: nil))`.`

nuno-vieira: What about `get()`? Shouldn't get do this as well?

Sources/StreamChat/StateLayer/Chat.swift
```diff
@@ -660,72 +685,82 @@ public class Chat {
...
-    ///   - message: The parent message id which has replies.
+    ///   - replyId: The message id of the reply from which newer messages are loaded. If nil, the newest currently loaded message id in ``MessageState/replies`` is used.
+    ///   - parentMessageId: The parent message id which has replies.
     ///   - limit: The limit for the page size. The default limit is 25.
     ///
     /// - Throws: An error while communicating with the Stream API.
-    public func loadRepliesFirstPage(of messageId: MessageId, limit: Int? = nil) async throws {
-        let messageState = try await makeMessageState(for: messageId)
-        return try await messageUpdater.loadRepliesFirstPage(of: messageId, limit: limit, cid: cid, paginationStateHandler: messageState.replyPaginationHandler)
+    public func loadReplies(after replyId: MessageId?, for parentMessageId: MessageId, limit: Int? = nil) async throws {
```
nuno-vieira, line 724: I think we might need to discuss if we want replies to be in the `Chat` object 🤔 Especially with Threads v2, it could make sense to create a `Thread` and `ThreadList` objects. If we have `Chat.loadReplies` and `Thread.loadReplies` will be confusing I guess 🤔

Sources/StreamChat/StateLayer/Chat.swift
laevandus: Uh-oh, yeah, `get` is a better answer, I will change it.

Sources/StreamChat/StateLayer/Chat.swift
laevandus: That is a fair question. The current setup involves using `MessageState` with parent message id and reading replies from there.

PR #3157, opened 2024-04-19
[ChatState] Reset watchers and members in Chat when triggering get()

laevandus:
### 🔗 Issue Links

### 🎯 Goal

After we added `get()` to Chat which resets `ChatState.messages` I realised that since we have `ChatState.members` and `ChatState.watchers` then it should also reset these as well. Otherwise users load 10000 members and there is no way to reset the whole state.

### 📝 Summary

* `ChatState.members` is internally driven by `MemberList`, therefore I need to make sure we associate channel payload members with `ChannelMemberListQuery`. 
* Watchers needs to be reset before channel payload is saved

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/Workers/ChannelUpdater.swift
+++ b/Sources/StreamChat/Workers/ChannelUpdater.swift
@@ -37,6 +37,7 @@ class ChannelUpdater: Worker {
     ///   - isInRecoveryMode: Determines whether the SDK is in offline recovery mode
     ///   - onChannelCreated: For some type of channels we need to obtain id from backend.
     ///     This callback is called with the obtained `cid` before the channel payload is saved to the DB.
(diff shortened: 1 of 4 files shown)


Sources/StreamChat/Workers/ChannelUpdater.swift
```diff
@@ -46,6 +47,7 @@ class ChannelUpdater: Worker {
         channelQuery: ChannelQuery,
         isInRecoveryMode: Bool,
         onChannelCreated: ((ChannelId) -> Void)? = nil,
+        actions: ChannelUpdateActions? = nil,
```
laevandus, line 50: Resetting members and watchers needs to happen within the same save closure for avoiding multiple DB observer changes (aka better performance). The cost is having more logic in the main channel update method.

PR #3165, opened 2024-04-24
[ChatStateLayer] Reset contexts when logging out

laevandus:
### 🔗 Issue Links

### 🎯 Goal

Fix logout issue with duplicate channels in DB.

### 📝 Summary

While testing the new state in the SwiftUI SDK we found an issue:

1. Log in with A
2. Open channel with A and B
3. Logout and log in with B
4. Open channel with A and B → issue happens (assert telling multiple ChannelDTOs for the same cid)

Simulator Screen Recording - iPhone 15 Pro - 2024-04-24 at 15 44 39

### 🛠 Implementation

It seems like when we flush the database on logout, open channel is not flushed. Not sure, but I think it is related to state layer's context which immediately reacts to changes.

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/Database/DatabaseContainer.swift
+++ b/Sources/StreamChat/Database/DatabaseContainer.swift
@@ -289,6 +289,7 @@ class DatabaseContainer: NSPersistentContainer {
         allContext.forEach { context in
             context.perform {
                 context.invalidateCurrentUserCache()
+                context.reset()
             }
         }
 


Sources/StreamChat/Database/DatabaseContainer.swift
```diff
@@ -289,6 +289,7 @@ class DatabaseContainer: NSPersistentContainer {
         allContext.forEach { context in
             context.perform {
                 context.invalidateCurrentUserCache()
+                context.reset()
```
martinmitrevski, line 292: hmm, this doesn't seem like it should be fixed here, but in `develop`. @nuno-vieira  thoughts?

laevandus:
I am going to move this PR to develop. Closing!

PR #3166, opened 2024-04-24
Allow to pin any message

testableapple:
### 🔗 Issue Links

Resolve https://github.com/GetStream/ios-issues-tracking/issues/782

### 🎯 Goal

Allow to pin any message

### 📝 Summary

- Replace `update` with `partial update` on pinning the message
- Update DemoApp pinning logic

### 🧪 Manual Testing Notes

1. Pin the current user's message
2. Pin the participant's message
3. Login as a participant
=> Verify that messages are pinned with correct labels

1. Unpin the current user's message
2. Unpin the participant's message
3. Login as a participant
=> Verify that messages are unpinned

### ☑️ Contributor Checklist

### 🎁 Meme

gif


Sources/StreamChat/Workers/MessageUpdater.swift
```diff
@@ -465,21 +465,18 @@ class MessageUpdater: Worker {
                 throw ClientError.MessageDoesNotExist(messageId: messageId)
             }
 
-            switch messageDTO.localMessageState {
-            case nil, .pendingSync, .syncingFailed, .deletingFailed:
-                try session.pin(message: messageDTO, pinning: pinning)
-                messageDTO.localMessageState = .pendingSync
```
nuno-vieira, line 471: I remember now why this was useful 😅 If the pinning failed, we would be able to retry it since editing a message is retryable. 

We can still use the new approach; the only thing we need to add is that if the request fails, we need to unpin it again.

Sources/StreamChat/Workers/MessageUpdater.swift
```diff
@@ -465,21 +465,18 @@ class MessageUpdater: Worker {
...
+            } catch {
+                log.warning("Failed to optimistically add the message pinning to the database: \(error)")
             }
-        }, completion: {
-            completion?($0)
+        }, completion: { error in
+            let endpoint: Endpoint<EmptyResponse> = .pinMessage(messageId: messageId, request: .init(set: .init(pinned: true)))
+            self.apiClient.request(endpoint: endpoint) { result in
+                guard let error = result.error else { return }
+                completion?(error)
```
nuno-vieira, line 477: So here, we probably should unpin if it fails

Sources/StreamChat/Workers/MessageUpdater.swift
```diff
@@ -493,21 +490,14 @@ class MessageUpdater: Worker {
...
-                throw ClientError.MessageEditing(
-                    messageId: messageId,
-                    reason: "message is in `\(messageDTO.localMessageState!)` state"
-                )
+            session.unpin(message: messageDTO)
+        }, completion: { error in
+            let endpoint: Endpoint<EmptyResponse> = .pinMessage(messageId: messageId, request: .init(set: .init(pinned: false)))
+            self.apiClient.request(endpoint: endpoint) { result in
+                guard let error = result.error else { return }
+                completion?(error)
```
nuno-vieira, line 498: And here should pin it back, if unpinning fails

Sources/StreamChat/Workers/MessageUpdater.swift
martinmitrevski: why do we remove this code, can you explain?

Sources/StreamChat/Workers/MessageUpdater.swift
martinmitrevski: yes, that makes sense.

martinmitrevski (commented):
Looks good in general, but would be nice to share more details about the db related code we remove.

Sources/StreamChat/Workers/MessageUpdater.swift
laevandus: Noting that depending on the localMessageState we use different mechanisms for making the API request. Pending sync == MessageEditor, pendingSend == MessageSender.
These types use different endpoints.

Sources/StreamChat/Workers/MessageUpdater.swift
```diff
@@ -465,21 +465,18 @@ class MessageUpdater: Worker {
...
-                )
+            } catch {
+                log.warning("Failed to optimistically add the message pinning to the database: \(error)")
             }
-        }, completion: {
-            completion?($0)
+        }, completion: { error in
+            let endpoint: Endpoint<EmptyResponse> = .pinMessage(messageId: messageId, request: .init(set: .init(pinned: true)))
+            self.apiClient.request(endpoint: endpoint) { result in
+                guard let error = result.error else { return }
```
laevandus, line 476: Completion handler is skipped here which is unexpected.

Sources/StreamChat/Workers/MessageUpdater.swift
testableapple: Fair enough, good point 👍

Sources/StreamChat/Workers/MessageUpdater.swift
testableapple: Updated 👍

Sources/StreamChat/Workers/MessageUpdater.swift
testableapple: Updated 👍

testableapple:
@martinmitrevski,
> why do we remove this code, can you explain?

This removes the `editing flow` for `pinning a message` and uses the same flow as `adding a reaction` with the `new partial update` endpoint. This is because the `MessageEditor` has an observer that sends a message update all the time the local state changes.

Sources/StreamChat/Workers/MessageUpdater.swift
testableapple: @laevandus, could you elaborate, please?

Sources/StreamChat/Workers/MessageUpdater.swift
testableapple: @martinmitrevski, this removes the editing flow for pinning a message and uses the same flow as adding a reaction with the new partial update endpoint. This is because the MessageEditor has an observer that sends a message update all the time the local state changes.

Sources/StreamChat/Workers/MessageUpdater.swift
laevandus: How it looks currently:
```swift
func pinMessage(messageId: MessageId, pinning: MessagePinning, completion: ((Error?) -> Void)? = nil) {
        database.write({ session in
            guard let messageDTO = session.message(id: messageId) else {
                throw ClientError.MessageDoesNotExist(messageId: messageId)
            }

            try session.pin(message: messageDTO, pinning: pinning)
        }, completion: { [weak self] error in
            let endpoint: Endpoint<EmptyResponse> = .pinMessage(
                messageId: messageId,
                request: .init(set: .init(pinned: true))
            )

            self?.apiClient.request(endpoint: endpoint) { result in
                guard let error = result.error else { return }
                
                    if let messageDTO = session.message(id: messageId) {
                        session.unpin(message: messageDTO)
                    }
                    completion?(error)
                }
            }
            completion?(error)
        })
    }
```

* If I call the pinMessage function it must trigger completion only once and never ignore calling it
* We trigger completion after write + after (apiRequest and another write) = 2 times
* The second write calls the completion from the write closure which is not correct, we should call it from the write's completion. If we call it from within the write closure, DB has not saved the change yet

testableapple: But we don't want to call the second database "write" on success 🤔

Please check out the latest commit, I copy-pasted the logic from `addReaction` and `deleteReaction` functions.

laevandus: I mixed up the previous snippet since on error we want to revert the state in the DB.

With the current implementation I am never receiving the error if the API request fails and the message stays as unpinned.
```swift
updater.pinMessage(messageId: "a", pinning: .xcz, completion: { error in
  if let error {
    // …
  } else {
    // … I expect that message is pinned here but it is not since API request failed
  }
}
```

Sources/StreamChat/Repositories/MessageRepository.swift
```diff
@@ -257,4 +257,37 @@ class MessageRepository {
...
+    }
+    
+    func undoMessageUnpinning(
+        on messageId: MessageId,
+        completion: (() -> Void)? = nil
+    ) {
+        database.write { session in
+            if let messageDTO = session.message(id: messageId) {
+                let pinning: MessagePinning = .init(expirationDate: messageDTO.pinExpires?.bridgeDate)
+                try session.pin(message: messageDTO, pinning: pinning)
```
laevandus, line 284: Does it really work like this?
If we call `session.unpin(…)` then this sets `pinExpires` to nil therefore we lose this information and calling `undoMessageUnpinning` can't recover this information anymore.

Wouldn't it be better not to have undo methods here and do everything within the `MessageUpdater`?

Sources/StreamChat/Workers/MessageUpdater.swift
testableapple: Updated, please check it out 🙂

Sources/StreamChat/Repositories/MessageRepository.swift
testableapple: Oh, you're right.

I updated the function and now pass `pinning` as an argument. Check it out 🙂

Sources/StreamChat/Workers/MessageUpdater.swift
testableapple: @laevandus, I can't understand how to bypass this swiftlint warning 🤔
`MessageUpdater.swift:484:1: warning: (hoistPatternLet) Reposition let or var bindings within pattern.`

This is about `case .failure(let apiError):`

Sources/StreamChat/Workers/MessageUpdater.swift
```diff
@@ -488,26 +495,36 @@ class MessageUpdater: Worker {
...
+            )
+            
+            self?.apiClient.request(endpoint: endpoint) { result in
+                switch result {
+                case .success:
+                    completion?(nil)
+                case .failure(let apiError):
+                    repository?.undoMessageUnpinning(on: messageId, pinning: pinning) { _ in
+                        completion?(apiError)
+                    }
```
laevandus, line 525: Feels like it would be clearer if we just have a single unpin and pin method. Therefore, I would like to propose removing the `MessageRepository` change and having a private unpin and pin here which only changes database. e.g. something like `pinLocalMessage`. Then we can use that within internal method here which:
1. modifies database state (through `unpinLocalMessage`)
2. makes an api request
3. reverts the local change if the API request fails (`pinLocalMessage` with the existing pinning state)

WDYT?

Sources/StreamChat/Workers/MessageUpdater.swift
testableapple: Will disable the rule `hoistPatternLet`

Sources/StreamChat/Workers/MessageUpdater.swift
testableapple: Could you check out the last commit? I think I got your point, but I might have missed something.

Tests/StreamChatTests/Workers/MessageUpdater_Tests.swift
```diff
@@ -1802,95 +1802,6 @@ final class MessageUpdater_Tests: XCTestCase {
         XCTAssertTrue(completionError is ClientError.MessageDoesNotExist)
     }
 
-    func test_pinMessage_updatesLocalMessageCorrectly() throws {
```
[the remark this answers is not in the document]
testableapple: Added new tests 👍

laevandus (approved):
Looks nice!

martinmitrevski (approved):
LGTM! ✅

PR #3167, opened 2024-04-24
[Reactions v2] New reactions query endpoint

nuno-vieira:
### 🔗 Issue Links

Resolves https://github.com/GetStream/ios-issues-tracking/issues/787

### 🎯 Goal

Adds the new reactions v2 query endpoint to be able to filter reactions based on type and author.

### 📝 Summary
- Adds new `ChatReactionListController`
- Adds new Slack Reactions List View to the Slack Example app

### 🛠 Implementation
Creates a new `ChatReactionListController` to follow the SDK pattern, where each query endpoint represents a controller. 

TODO:
- [ ] Everything is tested besides the `ReactionListController`. 
- [ ] Website docs for the new reaction list controller

### 🎨 Showcase

https://github.com/GetStream/stream-chat-swift/assets/12814114/42954fea-95ac-45cf-8046-0291d94e1d9d

### 🧪 Manual Testing Notes
Note: Backend user enrichment is not working on the new query reactions endpoint. So this is not testable until it is fixed on the backend.

1. Open the Slack Example App
2. Go to a channel
3. Long press on a reaction
4. It should show only the reactions of that type.

### ☑️ Contributor Checklist


Sources/StreamChat/APIClient/Endpoints/MessageEndpoints.swift
```diff
@@ -126,3 +136,39 @@ extension Endpoint {
...
+    }
+
+    public var type: MessageReactionType?
+    public var authorId: UserId?
+
+    public init(type: MessageReactionType? = nil, authorId: UserId? = nil) {
+        self.type = type
+        self.authorId = authorId
+    }
+}
```
nuno-vieira, line 153: The backend also accepts a mongo db query, but I think in this case this is enough? Because really the only supported queries are `and` and `$eq`. So using our existing query filters for these seems to be an overkill

Sources/StreamChat/Controllers/MessageController/MessageController.swift
```diff
@@ -55,6 +55,7 @@ public class ChatMessageController: DataController, DelegateCallable, DataStoreP
     /// To observe changes of the reactions, set your class as a delegate of this controller or use the provided
     /// `Combine` publishers.
     ///
+    // TODO: Why is this not an observer?
     public var reactions: [ChatMessageReaction] = [] {
```
nuno-vieira, line 59: Our reactions in `ChatMessageController` are not using an observer. Which means, it is not really offline-first. But for now, it is okay. It makes things simpler and it does not seem to require any sorting on CoreData, nor in the local memory data.

Sources/StreamChat/Controllers/MessageController/MessageController.swift
```diff
@@ -228,6 +229,8 @@ public class ChatMessageController: DataController, DelegateCallable, DataStoreP
         do {
             try messageObserver.startObserving()
             try repliesObserver?.startObserving()
+            // TODO: Not good, this impacts the reaction query, since the initial results won't be filtered with the given query,
+            // because of this is most likely better to just create a new `ReactionListController` so that we don't break the existing message controller.
             reactions = Array(messageObserver.item?.latestReactions.sorted(by: { $0.updatedAt > $1.updatedAt }) ?? [])
```
nuno-vieira, line 234: This is not good. The initial data, comes from latestReactions, and at this point, we don't have a reaction filter as well. In order to not break things, I prefer to create a new `ReactionListController` to make sure the new endpoint is totally isolated, and also easier to maintain than this one.

Examples/SlackClone/SlackChatMessageListViewController.swift
```diff
@@ -10,4 +10,106 @@ final class SlackChatMessageListViewController: ChatMessageListVC {
...
+    }
+
+    func openReactionsSheet(for message: ChatMessage, with reactionType: MessageReactionType) {
+        let messageController = client.messageController(cid: message.cid!, messageId: message.id)
+        let slackReactionListView = SlackReactionListViewController(
+            messageController: messageController,
+            reactionType: reactionType
+        )
+        present(slackReactionListView, animated: true)
+    }
```
nuno-vieira, line 40: When long pressing a reaction, it shows all the reactions only for that type.

Examples/SlackClone/SlackChatMessageListViewController.swift
```diff
@@ -10,4 +10,106 @@ final class SlackChatMessageListViewController: ChatMessageListVC {
...
+}
+
+/// Displays the reaction authors for a specific reaction type.
+class SlackReactionListViewController: UITableViewController, ChatMessageControllerDelegate {
+    let messageController: ChatMessageController
+    let reactionType: MessageReactionType
+
+    init(messageController: ChatMessageController, reactionType: MessageReactionType) {
```
nuno-vieira, line 48: Feels weird needing to have a `ChatMessageController` per reaction type to only display reactions. IMO having a `ReactionListController` makes more sense.

Sources/StreamChat/APIClient/Endpoints/MessageEndpoints.swift
nuno-vieira, line 153: This needs to be changed to our `Filter` type to make it consistent with the rest of the queries.

Sources/StreamChat/Controllers/MessageController/MessageController.swift
martinmitrevski: It would be confusing for customers - on one hand you have some reactions being loaded with a message controller, on another with a new one.

nuno-vieira: The `ChatMessageController` was designed with Message Actions UI, which also includes the reactions. But the reactions should have never been part of the message controller in the first place IMO. In Slack like UIs, you have lists of reactions. Using a `ChatMessageController` just for displaying reactions is probably even more confusing. 

If we want to make this work on `ChatMessageController` than it will require us to pass the reactions query in the initialised of the message controller in order to properly implement this, which is even more confusing for customers. Where having a `ReactionListController` is already a common pattern in the SDK. We can even deprecated the `loadReactions` from `ChatMessageController` and mention that there is this new `ReactionListController`. Or keep both, but mention that there is a new one.

nuno-vieira: Another 2 points why using `MessageController` is not a good idea:
1. It does not follow the common pattern in the SDK, usually whenever we have a query endpoint, we have a controller.
2. When calling `MessageController.synchronize()`, it will make a request to fetch the message and start the message observers. This is actually why we set the initial reactions in the observers, and not in the `synchronize()`. We use the message controller in the `ReactionsView` without calling synchronize to avoid the additional fetch. So it is actually counter-intuitive at the moment because we always tell customers to call sync in order to get updates, for when using reactions in the Message Controller, we do not call it.

Sources/StreamChat/Workers/ReactionListUpdater.swift
```diff
@@ -0,0 +1,32 @@
+//
+// Copyright © 2024 Stream.io Inc. All rights reserved.
+//
+
+import CoreData
+
+class ReactionListUpdater: Worker {
```
laevandus, line 7: Thinking if we should keep it next to the old load reactions method in `MessageUpdater`.

laevandus (approved):

martinmitrevski (approved):

PR #3174, opened 2024-05-02
Fix crash in @Cached property value when computeValue was called

laevandus:
### 🔗 Issue Links

Resolves [ios-issues-tracking/805](https://github.com/GetStream/ios-issues-tracking/issues/805)

### 🎯 Goal

Fix issues related to `computeValue` callback in `@Cached`

### 📝 Summary

* Do not call `computeValue` from the closure passed into `mutate` since it can trigger side-effects leading to crashes
* If multiple threads are accessing the property at the same time, and the cached value is nil, we might run `computeValue` multiple times. Old: `computeValue` was always called once, but in rare cases crashed.
* If cached value is nil, we now need to acquire a lock twice (first: reading the current value, second: setting a new value fetched from `computeValue`). Old: lock was acquired once, but in rare cases crashed.

### 🛠 Implementation

The `computeValue` callback which is passed into mutate's closure (locked region) can trigger side-effects. An example is CoreData context change which leads to reentrancy (exclusive access assert in Swift runtime). Another case we have seen is a deadlock.

### 🧪 Manual Testing Notes

No specific steps rather than just playing around with the demo app a bit. Logging out, logging in etc.

### ☑️ Contributor Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Add `reason` parameter to `flagMessage()` [#3173](https://github.com/GetStream/stream-chat-swift/issues/3173)
 ### 🐞 Fixed
 - Reset managed object contexts before removing all the data on logout [#3170](https://github.com/GetStream/stream-chat-swift/pull/3170)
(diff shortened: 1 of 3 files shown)


martinmitrevski (approved):
Makes sense to me 👍 @testableapple please do some testing on this one before pto.

nuno-vieira (approved):
LGTM! ✅  Nice Catch! 👌

testableapple (approved):
LGTM 🙌

PR #3175, opened 2024-05-02
[ChatStateLayer] Add ReactionList

laevandus:
### 🔗 Issue Links

### 🎯 Goal

Add `ReactionList` for querying reactions using a filter.

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/StateLayer/MemberList.swift
+++ b/Sources/StreamChat/StateLayer/MemberList.swift
@@ -30,8 +30,6 @@ public final class MemberList {
     
     /// Fetches the most recent state from the server and updates the local store.
     ///
-    /// - Important: Loaded members in ``MemberListState/members`` are reset.
-    ///
     /// - Throws: An error while communicating with the Stream API.
     public func get() async throws {
         let pagination = Pagination(pageSize: .channelMembersPageSize)
@@ -39,8 +37,6 @@ public final class MemberList {
     }
     
     /// Loads channel members for the specified pagination parameters and updates ``MemberListState/members``.
-    ///
-    /// - Important: If pagination offset is 0 and cursor is nil, then loaded members are reset.
     ///
     /// - Parameter pagination: The pagination configuration which includes a limit and an offset or a cursor.
     ///
--- a/Sources/StreamChat/StateLayer/ReactionList.swift
+++ b/Sources/StreamChat/StateLayer/ReactionList.swift
@@ -1,0 +1,87 @@
+//
+// Copyright © 2024 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+/// An object which represents a list of `ChatMessageReaction` for the specified query.
+@available(iOS 13.0, *)
+public final class ReactionList {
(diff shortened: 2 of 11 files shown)


Sources/StreamChat/StateLayer/MemberList.swift
```diff
@@ -40,8 +38,6 @@ public final class MemberList {
     
     /// Loads channel members for the specified pagination parameters and updates ``MemberListState/members``.
     ///
-    /// - Important: If pagination offset is 0 and cursor is nil, then loaded members are reset.
```
laevandus, line 43: Removed it because this is not implemented. Same with `ReactionList`. We can add it back when we have added support for this.

martinmitrevski (approved):

Sources/StreamChat/StateLayer/ReactionList.swift
```diff
@@ -0,0 +1,87 @@
...
+    }
+    
+    /// Loads more message reactions and updates ``MemberListState/members``.
+    ///
+    /// - Parameters
+    ///   - limit: The limit for the page size. The default limit is 25.
+    ///
+    /// - Throws: An error while communicating with the Stream API.
+    /// - Returns: An array of message reactions.
+    @discardableResult public func loadMoreReactions(limit: Int? = nil) async throws -> [ChatMessageReaction] {
```
nuno-vieira, line 57: Maybe there is no point on limit being nil? 🤔 
```suggestion
    @discardableResult public func loadMoreReactions(limit: Int = 25) async throws -> [ChatMessageReaction] {
```

nuno-vieira (commented):
LGTM Overall, but added some comments for discussion as well 👍

Sources/StreamChat/StateLayer/ReactionList.swift
nuno-vieira, line 57: We have these 2 functions in `Chat` object as well. Probably it makes sense to remove them since they are now a bit redudant with the new `ReactionList` ?

nuno-vieira: There is also the possiblity of just replacing the old `Chat.loadReactions` with the new query endpoints API. IF we think it does not make sense to introduce a new object just for reactions 🤔

laevandus: Currently it just matches with other methods. But it is a fair point, optional Int looks a bit funny. If we go for it, we should change it in other types as well (separate PR). I am up for it changing it everywhere to non-optionals.

laevandus: I am gonna create another PR which changes all the call sites.

laevandus: We have the same case with channel members what you were mentioning in another PR. I am thinking about taking the discussion of having `MessageState.reactions` and `ChatState.members` as a separate discussion point.

laevandus:
Captured the questions, will be treated separately.

PR #3177, opened 2024-05-03
Modern state layer

laevandus:
### 🔗 Issue Links

Resolves [ios-issues-tracking/728](https://github.com/GetStream/ios-issues-tracking/issues/728)

### 🎯 Goal

Add a modern state layer with async-await and observable state objects

### 📝 Summary

A new set of types which represent a new way how to interact with Stream API (alternative way to controllers).
* Leverages async-await 
* Method calls wait for API requests to finish
* Main actor guarded state objects for easy way for observing data changes

New types:
* ChannelList & ChannelListState
* Chat & ChatState + MessageState
* ConnectedUser + ConnectedUserState
* MemberList + MemberListState
* MessageSearch + MessageSearchState
* ReactionList + ReactionListState
* UserList + UserListState
* UserSearch + UserSearchState

### ☑️ Contributor Checklist


Sources/StreamChat/StateLayer/ChatClient+Factory.swift
```diff
@@ -0,0 +1,283 @@
...
+    public func makeDirectMessageChat(
+        with members: [UserId],
+        type: ChannelType = .messaging,
+        isCurrentUserMember: Bool = true,
+        name: String? = nil,
+        imageURL: URL? = nil,
+        team: String? = nil,
+        messageOrdering: MessageOrdering = .topToBottom,
+        memberSorting: [Sorting<ChannelMemberListSortingKey>] = [],
+        channelListQuery: ChannelListQuery? = nil,
```
nuno-vieira, line 185: Same here

Sources/StreamChat/StateLayer/ChatState+Internal.swift
```diff
@@ -0,0 +1,101 @@
...
+                    // Everything has been read
+                    return nil
+                } else {
+                    return lookupUnreadMessageId(after: lastReadIndex)
+                }
+            } else {
+                // Can't reach the last read message (if all have been loaded then the channel might have been truncated or hidden, in that case, use the oldest message)
+                return hasLoadedAllOldestMessages ? oldestRegularMessageId : nil
+            }
+        }
```
nuno-vieira, line 52: This is copy pasted from `ChatChannelController` right? 🤔 The problem here is that we now need 2 maintain the same logic twice which is a bit concerning

Sources/StreamChat/StateLayer/ConnectedUser.swift
```diff
@@ -0,0 +1,169 @@
...
+    // MARK: - Moderating Users
+    
+    /// Mutes the user in all the channels.
+    ///
+    /// - Note: Messages from muted users are not delivered via push notifications.
+    ///
+    /// - Parameter userId: The id of the user to mute.
+    ///
+    /// - Throws: An error while communicating with the Stream API.
+    public func muteUser(_ userId: UserId) async throws {
```
nuno-vieira, line 113: nit: We have this pattern in other places as well, but the omitting of the label would make sense if we passed the whole `ChatUser`. If we are only passing the ID, then it kind makes sense to still be a named label to make it clear we need to pass an ID 
```suggestion
    public func muteUser(id: UserId) async throws {
```

Sources/StreamChat/StateLayer/MemberList.swift
```diff
@@ -0,0 +1,86 @@
...
+    }
+    
+    /// An observable object representing the current state of the member list.
+    @MainActor public lazy var state: MemberListState = stateBuilder.build()
+    
+    /// Fetches the most recent state from the server and updates the local store.
+    ///
+    /// - Throws: An error while communicating with the Stream API.
+    public func get() async throws {
+        let pagination = Pagination(pageSize: .channelMembersPageSize)
```
nuno-vieira, line 35: We should use the query pagination size here as well

Sources/StreamChat/StateLayer/MemberList.swift
```diff
@@ -0,0 +1,86 @@
...
+    
+    /// Loads more channel members and updates ``MemberListState/members``.
+    ///
+    /// - Parameters
+    ///   - limit: The limit for the page size. The default limit is 30.
+    ///
+    /// - Throws: An error while communicating with the Stream API.
+    /// - Returns: An array of channel members.
+    @discardableResult public func loadMoreMembers(limit: Int? = nil) async throws -> [ChatChannelMember] {
+        let pageSize = limit ?? Int.channelMembersPageSize
```
nuno-vieira, line 57: We should also use the query pagination size here too if the limit is nil

Sources/StreamChat/StateLayer/ReactionList.swift
```diff
@@ -0,0 +1,87 @@
...
+    }
+    
+    /// An observable object representing the current state of the reaction list.
+    @MainActor public lazy var state: ReactionListState = stateBuilder.build()
+    
+    /// Fetches the most recent state from the server and updates the local store.
+    ///
+    /// - Throws: An error while communicating with the Stream API.
+    public func get() async throws {
+        let pagination = Pagination(pageSize: 25)
```
nuno-vieira, line 35: Same here we should use the page size from the initial query

Sources/StreamChat/StateLayer/UserList.swift
```diff
@@ -0,0 +1,81 @@
...
+    /// An observable object representing the current state of the users list.
+    @MainActor public lazy var state: UserListState = stateBuilder.build()
+    
+    /// Fetches the most recent state from the server and updates the local store.
+    ///
+    /// - Important: Loaded users in ``UserListState/users`` are reset.
+    ///
+    /// - Throws: An error while communicating with the Stream API.
+    public func get() async throws {
+        let pagination = Pagination(pageSize: .usersPageSize)
```
nuno-vieira, line 35: Same here, we need to use the pageSize from the query

docusaurus/docs/iOS/client/state-layer/state-layer-overview.md
```diff
@@ -0,0 +1,233 @@
+---
+title: State Layer
+---
+
+:::note
+Since `StreamChat` version 4.55.0.
```
nuno-vieira, line 6: Maybe a little bit more clear? WDYT? 🤔 
```suggestion
Available from `StreamChat` version 4.55.0.
```

docusaurus/docs/iOS/client/state-layer/state-layer-overview.md
```diff
@@ -0,0 +1,233 @@
...
+:::
+
+The `StreamChat` framework represents the Stream API with [controllers](../controllers/controllers-overview.md) which use completion handlers and delegates for communicating about state changes. In addition to controllers, we provide a new and modern way of interacting with the Stream API. It follows an architecture where we have objects interacting with the Stream API through async functions. These objects are accompanied with state objects that hold the current state. This architecture follows similar patterns as our [Video SDK for iOS](https://getstream.io/video/sdk/ios/).
```
nuno-vieira, line 9: It could be interesting to add, that his new state layer is the future, and on v5, it will be the only way to interact with the SDK, and the controller will be deprecated 🤔

nuno-vieira (commented):
The new state layer is looking promising 🙂 This PR is also helpful get an overview of the general API. I added some comments since there some small bugs here and there, but also added points of discussion 👍

Sources/StreamChat/StateLayer/Chat.swift
```diff
@@ -0,0 +1,1439 @@
...
+        showReplyInChannel: Bool = false,
+        attachments: [AnyAttachmentPayload] = [],
+        quote quotedMessageId: MessageId? = nil,
+        mentions: [UserId] = [],
+        pinning: MessagePinning? = nil,
+        extraData: [String: RawJSON] = [:],
+        silent: Bool = false,
+        skipPushNotification: Bool = false,
+        skipEnrichURL: Bool = false,
+        messageId: MessageId? = nil
```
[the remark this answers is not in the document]
martinmitrevski: I would prefer consistency with other platforms tbh. Introducing new structs can lead to more differences. They are all optional anyway.

Sources/StreamChat/StateLayer/ChatState+Internal.swift
martinmitrevski: I agree, we can probably introduce some helper class. We need to update the logic anyway soon, so having it in 2 places is a bit error prone.

docusaurus/docs/iOS/client/state-layer/state-layer-overview.md
martinmitrevski: We don't know that 100% to write it publicly, it depends on the context when v5 is released and would probably need an approval from the business side.

Sources/StreamChat/Repositories/SyncRepository.swift
```diff
@@ -134,6 +156,20 @@ class SyncRepository {
                 )
             }
         operations.append(contentsOf: refetchChannelListQueryOperations)
+        
+        let channelListQueries = syncQueue.sync {
```
martinmitrevski, line 160: We should properly test the sync process, cc @testableapple

Sources/StreamChat/StateLayer/ChannelListState+Observer.swift
```diff
@@ -0,0 +1,146 @@
...
+        
+        private func isChannelInList(_ cid: ChannelId) async throws -> Bool {
+            try await database.read { [query] context in
+                guard let (channelDTO, queryDTO) = context.getChannelWithQuery(cid: cid, query: query) else { return false }
+                return queryDTO.channels.contains(channelDTO)
+            }
+        }
+        
+        private func linkChannelIfNeeded(_ channel: ChatChannel) async throws {
+            let listContainsChannel = try await isChannelInList(channel.cid)
```
martinmitrevski, line 132: this code is also duplicated, right? Let's discuss if/how should we solve this.

Sources/StreamChat/StateLayer/Chat.swift
```diff
@@ -0,0 +1,1439 @@
...
+    ///
+    /// - Parameters:
+    ///   - messageId: The id of the message to send the reaction.
+    ///   - type: The type that describes a message reaction. Common examples are: “like”, “love”, “smile”, etc. An user can have only 1 reaction of each type per message.
+    ///   - score: The score of the reaction for cumulative reactions (example: n number of claps).
+    ///   - enforceUnique: If `true`, the added reaction will replace all reactions the user has (if any) on this message.
+    ///   - extraData: The reaction's extra data.
+    ///
+    /// - Throws: An error while communicating with the Stream API.
+    public func sendReaction(
```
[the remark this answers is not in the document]
martinmitrevski: Yes, add is better.

Sources/StreamChat/StateLayer/Chat.swift
```diff
@@ -0,0 +1,1439 @@
...
+    
+    /// Loads more reactions and updates ``MessageState/reactions``.
+    ///
+    /// - Parameters:
+    ///   - messageId: The id of the message to load reactions.
+    ///   - limit: The limit for the page size. The default limit is 25.
+    ///
+    /// - Throws: An error while communicating with the Stream API.
+    /// - Returns: An array of reactions for the next page.
+    @discardableResult public func loadMoreReactions(
```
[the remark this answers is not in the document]
martinmitrevski: It kind of depends. It's always a tradeoff between ease of use for customers and maintenance. When I think about how customers would use this, I can imagine more people needing to load members from the chat, so that one should stay. About this one, I'm not entirely sure.

Sources/StreamChat/Workers/ChannelUpdater.swift
```diff
@@ -658,3 +718,304 @@ class ChannelUpdater: Worker {
...
+extension ChannelUpdater {
+    func acceptInvite(cid: ChannelId, message: String?) async throws {
+        try await withCheckedThrowingContinuation { continuation in
+            acceptInvite(cid: cid, message: message) { error in
+                continuation.resume(with: error)
+            }
+        }
+    }
+    
+    func addMembers(currentUserId: UserId? = nil, cid: ChannelId, userIds: Set<UserId>, message: String? = nil, hideHistory: Bool) async throws {
```
martinmitrevski, line 734: formatting here should be improved

Sources/StreamChat/Workers/ChannelUpdater.swift
```diff
@@ -658,3 +718,304 @@ class ChannelUpdater: Worker {
...
+        try await withCheckedThrowingContinuation { continuation in
+            uploadFile(type: type, localFileURL: localFileURL, cid: cid, progress: progress) { result in
+                continuation.resume(with: result)
+            }
+        }
+    }
+    
+    // MARK: -
+    
+    func loadMessages(with channelQuery: ChannelQuery, pagination: MessagesPagination) async throws -> [ChatMessage] {
```
martinmitrevski, line 947: can you please remind me why we needed these 3 methods below?

Sources/StreamChat/Workers/MessageUpdater.swift
```diff
@@ -770,3 +806,212 @@ private extension DatabaseSession {
...
+        try await withCheckedThrowingContinuation { continuation in
+            unpinMessage(messageId: messageId) { result in
+                continuation.resume(with: result)
+            }
+        }
+    }
+    
+    // MARK: -
+    
+    func loadReplies(for parentMessageId: MessageId, pagination: MessagesPagination, cid: ChannelId, paginationStateHandler: MessagesPaginationStateHandling) async throws -> [ChatMessage] {
```
martinmitrevski, line 982: Same question here - probably the answer would be the same as the channel updater one.

Sources/StreamChat/Workers/UserListUpdater.swift
```diff
@@ -63,3 +66,61 @@ enum UpdatePolicy {
...
+
+private extension UserPayload {
+    func asModel() -> ChatUser {
+        ChatUser(
+            id: id,
+            name: name,
+            imageURL: imageURL,
+            isOnline: isOnline,
+            isBanned: isBanned,
+            isFlaggedByCurrentUser: false, // TODO: isFlaggedByCurrentUser is not set
```
martinmitrevski, line 115: We should check this.

martinmitrevski (commented):
First, we should acknowledge the outstanding work you did with this feature 👏 
Looks great, added some questions and comments, also the ones from Nuno are very useful (one thing I disagree with him is introducing structs for many params, but let's discuss it).
This is my first review iteration, would do another one soon.
Important to test again the uikit app, as well as the `new-state` branch in SwiftUI.

docusaurus/docs/iOS/client/state-layer/state-layer-overview.md
nuno-vieira: I think we should at least say that the State layer should be the preferrable way, and that we recommend using first the state layer

docusaurus/docs/iOS/client/state-layer/state-layer-overview.md
nuno-vieira: Btw it actually might be from 4.56.0 🤔

Sources/StreamChat/StateLayer/Chat.swift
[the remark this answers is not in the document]
laevandus: Other platforms use `sendReaction`, that is why I changed it. iOS was the only one using `addReaction`.
Example: https://getstream.io/chat/docs/ios-swift/send_reaction/?language=javascript

Sources/StreamChat/Repositories/SyncRepository.swift
```diff
@@ -68,6 +74,22 @@ class SyncRepository {
     deinit {
         cancelRecoveryFlow()
     }
+    
+    // MARK: - Tracking Active
+    
+    func trackChannelListQuery(_ provider: @escaping () -> ChannelListQuery?) {
```
laevandus, line 80: Needs the same fix what @nuno-vieira did for controllers.

Sources/StreamChat/StateLayer/Chat.swift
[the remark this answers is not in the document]
laevandus: I did consider something like this weeks ago:

```
func reply(
        to parentMessageId: MessageId,
        text: String,
        options: [SendReplyOption] 
```
where 
```
enum SendReplyOption {
  case showReplyInChannel
  case attachments([AnyattachmentPayload])
…
}
```
which gives you
```
chat.reply(to: "xyz", text: "hi", options: [.showReplyInChannel, .quotedMessage("123")])
```

But all in all, pretty similar to default argument usage.

Sources/StreamChat/StateLayer/Chat.swift
```diff
@@ -0,0 +1,1439 @@
...
+        // Note that watching is started in ChatClient+Chat when channel updater's update is called.
+        try await channelUpdater.startWatching(cid: cid, isInRecoveryMode: false)
+    }
+    
+    /// Stop watching the channel which disables server-side events.
+    ///
+    /// Please refer to [Watching a Channel](https://getstream.io/chat/docs/ios-swift/watch_channel/?language=swift) for additional information.
+    ///
+    /// - Throws: An error while communicating with the Stream API.
+    public func stopWatching() async throws {
```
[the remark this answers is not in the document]
laevandus: Yeah, why not. Makes it easier to check what the whole watching is all about.

Sources/StreamChat/StateLayer/MemberList.swift
laevandus: I'll go over all the cases with one commit. Will review them all.

Sources/StreamChat/Workers/UserListUpdater.swift
laevandus: That is problematic since the property is local only, server does not have that. Yep, needs to be reviewed again.

Sources/StreamChat/Workers/ChannelUpdater.swift
laevandus: No other reason than keeping the `Chat.swift` has clean as possible

Sources/StreamChat/Workers/MessageUpdater.swift
laevandus: No other reason than keeping the `Chat.swift` has clean as possible

Sources/StreamChat/StateLayer/Chat.swift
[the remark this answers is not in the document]
nuno-vieira: I see 👍 Then it makes sense to be sendReaction 👍 But we should update our comments

Sources/StreamChat/StateLayer/Chat.swift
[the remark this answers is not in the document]
nuno-vieira: Discuss on tech discussion, for now we will keep it like this and have both ways

Sources/StreamChat/StateLayer/Chat.swift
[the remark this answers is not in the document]
martinmitrevski: We shouldn't introduce this, React is also not doing that: https://getstream.io/chat/docs/react/threads/?language=javascript

Sources/StreamChat/StateLayer/ChannelListState+Observer.swift
laevandus: https://github.com/GetStream/stream-chat-swift/pull/3190

Sources/StreamChat/StateLayer/Chat.swift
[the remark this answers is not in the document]
nuno-vieira: @martinmitrevski JS is using a whole object to provide the parameters, also Kotlin. You need to provide a "Message" object, which is kinda confusing as well. 

IMO, these are related to Swift, JS and Kotlin guidelines, and not really about consistency. 

For example, on JS, you can see that sometimes they also pass Options arguments, like when getting replies:
```js
await channel.getReplies(parentMessageId, {limit: 20, id_lte: "42"});
```

For sending a message, they only use 1 parameter, which is kinda like our "Options" argument:
```js
const reply = await channel.sendMessage({
    text: 'Hey, I am replying to a message!',
    parent_id: parentID,
    show_in_channel: false,
});
```

Having so many functions arguments makes things harder to maintain, and test. We need to always provide optional params, and constantly changing our mocks to support the new parameters

Sources/StreamChat/Workers/ChannelUpdater.swift
laevandus: Done

Sources/StreamChat/Repositories/SyncRepository.swift
laevandus: Done in SyncController

Sources/StreamChat/StateLayer/ChannelListState+Observer.swift
laevandus: Done

Sources/StreamChat/Repositories/SyncRepository.swift
laevandus: We tested this quite a bit with @testableapple and found one issue which is fixed now.

nuno-vieira (approved):
LGTM! ✅

Sources/StreamChat/Workers/UserListUpdater.swift
laevandus: Only way to fix it is to make additional DB reads when converting API response to the model object. Currently this particular `asModel()` function is used for user search where this property is not used.

martinmitrevski: I think it's not relevant for `ChatUser`, only for the `CurrentUser`. Let's just remove the TODO.

martinmitrevski (approved):
LGTM! ✅

Sources/StreamChat/StateLayer/Chat.swift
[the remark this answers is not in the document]
laevandus: 12 arguments is a pretty long list and most of them are not used by customers anyway. At the same default values help to hide it. I did not find a good reference what other companies are doing in cases like these (Apple, Google)

PR #3178, opened 2024-05-03
[ChatStateLayer] Use non-optional limit arguments in pagination methods

laevandus:
### 🔗 Issue Links

### 🎯 Goal

A proposal to use non-optionals for `limit`

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/StateLayer/ChannelList.swift
+++ b/Sources/StreamChat/StateLayer/ChannelList.swift
@@ -68,13 +68,11 @@ public class ChannelList {
     ///
     /// - Throws: An error while communicating with the Stream API.
     /// - Returns: An array of loaded channels.
-    @discardableResult public func loadMoreChannels(limit: Int? = nil) async throws -> [ChatChannel] {
-        let limit = limit ?? query.pagination.pageSize
-        let count = await state.channels.count
-        return try await channelListUpdater.loadNextChannels(
+    @discardableResult public func loadMoreChannels(limit: Int = 20) async throws -> [ChatChannel] {
+        try await channelListUpdater.loadNextChannels(
             query: query,
             limit: limit,
-            loadedChannelsCount: count
+            loadedChannelsCount: state.channels.count
         )
     }
 }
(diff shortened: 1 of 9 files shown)


Sources/StreamChat/StateLayer/Chat.swift
```diff
@@ -223,7 +223,7 @@ public class Chat {
     ///
     /// - Throws: An error while communicating with the Stream API.
     /// - Returns: An array of channel members for the pagination request.
-    @discardableResult public func loadMoreMembers(limit: Int? = nil) async throws -> [ChatChannelMember] {
+    @discardableResult public func loadMoreMembers(limit: Int = 30) async throws -> [ChatChannelMember] {
```
martinmitrevski, line 226: how did you come up with the values? Did you take into account the backend defaults?

laevandus: These are just copy-pasted values from our `Int` extension. 

I put up this PR but I am actually not convinced anymore if it is a good idea.

martinmitrevski: I agree, it's not. Better to pass nil and let the backend return its default.

Sources/StreamChat/StateLayer/ChannelList.swift
```diff
@@ -68,13 +68,11 @@ public class ChannelList {
     ///
     /// - Throws: An error while communicating with the Stream API.
     /// - Returns: An array of loaded channels.
-    @discardableResult public func loadMoreChannels(limit: Int? = nil) async throws -> [ChatChannel] {
-        let limit = limit ?? query.pagination.pageSize
-        let count = await state.channels.count
-        return try await channelListUpdater.loadNextChannels(
+    @discardableResult public func loadMoreChannels(limit: Int = 20) async throws -> [ChatChannel] {
```
nuno-vieira, line 71: Btw, I know I added this comment previously, but actually, the `limit: Int?` being optional has a valid reason. 

If the limit is `nil` which should be the default, then we should use the pagination size of the `self.query`.

This is so that, whenever the customer initialises the `ChannelList` with a given query, and sets the pagination size there, it should use by default the pagination size of the initial query. So all limits, should be like so:
`let limit = limit ?? self.query.pagination.pageSize ?? defaultValue`

nuno-vieira (changes requested):
Requesting changes, because I actually just remebered why `limit` should indeed be optional. Sorry for the confusion 😄

laevandus:
Seems like we all agreed to not change it

PR #3181, opened 2024-05-06
Add fastlane lane to get the frameworks size and update the README

nuno-vieira:
### 🔗 Issue Links
Relates to https://github.com/GetStream/ios-issues-tracking/issues/448

### 🎯 Goal

Add an easy way to get the SDK sizes. For now, it just shows the sizes in Megabytes. In the future, we can use [bloaty](https://github.com/google/bloaty) to show the diffing between 2 binaries and show which files increased the most. We would also need to store the previous binary somewhere. Either way, it is out of the scope of this PR. This one is for us to have a way to manually check the SDK size more easily.

UPDATE:
It also updates the Labels with the most updated sizes.


fastlane/Fastfile
```diff
@@ -771,6 +771,25 @@ lane :copyright do
...
+  stream_chat_size = File.size('../Build/DemoApp.xcarchive/Products/Library/Frameworks/ChatSample.app/Frameworks/StreamChat.framework/StreamChat')
+  stream_chat_ui_size = File.size('../Build/DemoApp.xcarchive/Products/Library/Frameworks/ChatSample.app/Frameworks/StreamChatUI.framework/StreamChatUI')
+
+  # Converting to MB, not MiB (1024 bytes), since macOS finder uses MB
+  stream_chat_size_in_mb = (stream_chat_size.to_f / 1000 / 1000).round(2)
+  stream_chat_ui_size_in_mb = (stream_chat_ui_size.to_f / 1000 / 1000).round(2)
+  
+  UI.success("StreamChat: #{stream_chat_size_in_mb}MB")
+  UI.success("StreamChatUI: #{stream_chat_ui_size_in_mb}MB")
+end
```
laevandus, line 791: I would be so nice to post results to PR. Otherwise it is too hidden.

nuno-vieira: Yes, right now, this is a manual process. Eventually, we would like to do one or all of these, probably a task for @testableapple:
- Run this for every release and update the README size labels
- Run this for every PR (3-4min increase on CI because of the archive, but maybe there's way to optimize) and show the difference from the previous release/commit (The diff can be a simple subtract of the new and old size)
- The latest stage would be to run it for every PR and show the difference at a file/source code level using [bloaty](https://github.com/google/bloaty)

fastlane/Fastfile
```diff
@@ -771,6 +771,44 @@ lane :copyright do
...
+    archive_path: archive_dir,
+    export_options_plist: "Scripts/SDKSizeExportOptions.plist",
+    export_archive: "#{root_dir}/Export",
+    export_path: "#{root_dir}/Export"
+  )
+
+  packaging_url = "../#{root_dir}/Export/Packaging.log"
+  packaging_file_content = File.read(packaging_url)
+  match = packaging_file_content.match(/\b(\d+)\sbytes\sfor\s\.\/Payload\/ChatSample\.app\/Frameworks\/StreamChatUI\.framework\/Assets\.car\b/)
+  assets_thinned_size = match[1].to_i
```
nuno-vieira, line 798: At the moment I did not find a better way to get the Assets.car with the correct size. Using it from the Products folder directly, it is not the correct size since the assets are not thinned. So we need to simulate an ad-hoc export with App thinning enabled. Then, the real size is located in the Packaging.log, but I'm not sure if there's an easier way to do this TBH.

For now, this seems to work, and we have the size reports similar to what Emerge Tools report.

fastlane/Fastfile
laevandus: Yep, sounds like a plan. Let's get the informative piece in now.

laevandus (approved):

fastlane/Fastfile
ipavlidakis: Wow that's awesome! It will be great it bring it over also to video. Nice job!

ipavlidakis (approved):

PR #3184, opened 2024-05-07
Fix triggering user query calls whenever a new user was added in `UserListController`

nuno-vieira:
### 🔗 Issue Links
Relates to https://github.com/GetStream/ios-issues-tracking/issues/800

### 🎯 Goal
Fixes adding new users causing additional HTTP calls just to check if the new user is part of the query.

### 📝 Summary
The responsible for this was `NewUserQueryUpdater`. This is a really old worker that we had, and it seems like we had one of these for channel lists as well, and then we had to remove it because of problems. This approach is for sure a no go, creating an HTTP call for every new user just to check if it stills matches the query is not recommended by the backend team, and it can be a pretty fragile solution. Observing this in the `UserListController` also seems to be an overkill. Only for the `MemberListController` this is useful. So for the time being, we will just remove this. If customers want `UserListController` to be reactive, we will need to implement the automatic query filter and/or the filter block approach that we have for channel lists. 

References:
- Old [NewQueryChannelUpdater](https://github.com/GetStream/stream-chat-swift/blob/a53c876f806c61e5cd63db0a7cf9d82ce2643f75/Sources_v3/Workers/Background/NewChannelQueryUpdater.swift)
- Introduction of the new Channel List linking approach: https://github.com/GetStream/stream-chat-swift/pull/1438
- Removal of [NewQueryChannelUpdater](https://github.com/GetStream/stream-chat-swift/pull/1460)

### 🧪 Manual Testing Notes

N/A

### ☑️ Contributor Checklist


CHANGELOG.md
```diff
@@ -4,6 +4,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 # Upcoming
 
 ## StreamChat
+### 🐞 Fixed
+- Fix triggering user query calls whenever a new user was added in `UserListController` #[3184](https://github.com/GetStream/stream-chat-swift/pull/3184)
```
martinmitrevski, line 8: Thinking if we should mention that the `UserListController` is no longer reactive, wdyt?

martinmitrevski (approved):
LGTM ✅

CHANGELOG.md
nuno-vieira: Yeah, I was thinking about that too, but I was afraid it might warn customers for no important reason, since most of them won't care about this, but maybe better to be transparent, so I'll add this 👍

nuno-vieira: Done ✅

PR #3186, opened 2024-05-08
Improve channel list performance by enabling reconfigure cells

laevandus:
### 🔗 Issue Links

Resolves [ios-issues-tracking/790](https://github.com/GetStream/ios-issues-tracking/issues/790)

### 🎯 Goal

Enable cell reconfiguring and guard against reentrancy which led to an assertion in Apple's code/

### 📝 Summary

* Enable cell reconfiguring in channel list which makes it much faster
* Guard against reentrancy which can happen if reconfigure cells call ends up triggering Core Data store change which is then picked up by database observers (and leads to calling reloadChannels again)

### 🧪 Manual Testing Notes

Exploratory testing:
Open channel list and trigger channel list changes (e.g. new messages arriving)

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
+++ b/Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
@@ -84,6 +84,9 @@ open class ChatChannelListVC: _ViewController,
     /// updates when the channel list is not visible in the window.
     private(set) var skippedRendering = false
     private(set) var skipChannelUpdates = true
+    
+    /// Loading cells can trigger persistent store changes which leads to reloadChannels re-entrancy.
+    private var isApplyingChangesets = false
 
     /// A component responsible to handle when to load new channels.
     private lazy var viewPaginationHandler: ViewPaginationHandling = {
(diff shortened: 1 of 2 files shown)


Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
```diff
@@ -248,15 +251,24 @@ open class ChatChannelListVC: _ViewController,
...
+        isApplyingChangesets = true
         let previousChannels = channels
         let newChannels = Array(controller.channels)
         let stagedChangeset = StagedChangeset(source: previousChannels, target: newChannels)
-        collectionView.reload(using: stagedChangeset, reconfigure: { _ in false }) { [weak self] newChannels in
+        collectionView.reload(using: stagedChangeset, reconfigure: { _ in true }) { [weak self] newChannels in
             self?.channels = newChannels
         }
+        isApplyingChangesets = false
     }
```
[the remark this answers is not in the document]
nuno-vieira: Probably not worth to to test it 🤔

Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
```diff
@@ -248,15 +251,24 @@ open class ChatChannelListVC: _ViewController,
...
+        
+        isApplyingChangesets = true
         let previousChannels = channels
         let newChannels = Array(controller.channels)
         let stagedChangeset = StagedChangeset(source: previousChannels, target: newChannels)
-        collectionView.reload(using: stagedChangeset, reconfigure: { _ in false }) { [weak self] newChannels in
+        collectionView.reload(using: stagedChangeset, reconfigure: { _ in true }) { [weak self] newChannels in
             self?.channels = newChannels
         }
+        isApplyingChangesets = false
```
nuno-vieira, line 271: Shouldn't this be set inside the reload completion block? 🤔

martinmitrevski: I also think that

laevandus: This is actually not a completion but it is the `setData` closure which is called before `performBatchUpdates` (too early).

```swift
 func reload<C>(
        using stagedChangeset: StagedChangeset<C>,
        reconfigure: ((IndexPath) -> Bool)? = nil,
        interrupt: ((Changeset<C>) -> Bool)? = nil,
        setData: (C) -> Void
    ) {
```

where `setData` is called after each changeset

```swift
for changeset in stagedChangeset {
            if let interrupt = interrupt, interrupt(changeset), let data = stagedChangeset.last?.data {
                setData(data)
                return reloadData()
            }
            
            performBatchUpdates({
                setData(changeset.data)
                
                if !changeset.sectionDeleted.isEmpty {
                    deleteSections(IndexSet(changeset.sectionDeleted))
                }
                // later we call reconfigureItems which can trigger re-entrancy
```

PR #3187, opened 2024-05-08
Fix duplicated watching channel calls when reconnecting

nuno-vieira:
### 🔗 Issue Links
Resolves https://github.com/GetStream/ios-issues-tracking/issues/800

### 🎯 Goal

Fix having duplicated watching channel calls after reconnection logic.

Context: Whenever we reconnect, if the user is watching channels, we track these channels and re-watch them. 

### 📝 Summary
This was because we were incorrectly adding a channel controller to `activeChannelControllers` whenever we created a controller. This is wrong; we should only add it when we call synchronize or start watching a channel. Only in this scenario do we need to track a channel. 

Besides this, we did not control whether we already had a channel in the `activeChannelControllers`. So, there were cases where we had duplicated active channel controllers, and whenever a reconnection happened, it would trigger duplicated watch calls.

Lastly, even if a channel is open, if the customer stops watching the channel, it should not be in the `activeChannelControllers`. This was also missing.

### 🧪 Manual Testing Notes
1- Open Proxyman
2- Open Demo App Configuration and setup token refresh with 15s
3- Open the Channel List
4- Wait for reconnection
5- No channel query call should be made (Do not confuse with channel list call)
6- Open and close a few channels
7- 4. and 5.
8- Open a channel, and remain open
9- When reconnecting the a watch call should be made for the open channel

### ☑️ Contributor Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## StreamChat
 ### 🐞 Fixed
 - Fix triggering user query calls whenever a new user was added in `UserListController` [#3184](https://github.com/GetStream/stream-chat-swift/pull/3184)
+- Fix duplicated watching channel calls when reconnecting [#3187](https://github.com/GetStream/stream-chat-swift/pull/3187)
 ### 🔄 Changed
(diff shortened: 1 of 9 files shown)


laevandus (approved):

martinmitrevski (approved):
Great find 👌

laevandus:
I manually verified

PR #3188, opened 2024-05-09
Only sync at most 250 channels at the same time

nuno-vieira:
### 🔗 Issue Links

Resolves https://github.com/GetStream/ios-issues-tracking/issues/792

### 🎯 Goal
Fixes trying to sync more than 255 channels at the same time. (Backend limit)

### 📝 Summary
Changes the limit to 250, and it should be more than enough. At most, customers will have about 100 active channels at the same time, and even then, it wouldn't be a big deal if all channels where not synced. Only the most recents are important, and eventually, they will be refetched anyways. 

The reason we could hit 255+ channels syncing is that we never erase the channels from the DB; we just unlink them from the queries. So, there could be a case where a user could have more than 255 channels in their local DB, with really old channels. 

### 🧪 Manual Testing Notes
N/A

### ☑️ Contributor Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### 🐞 Fixed
 - Fix triggering user query calls whenever a new user was added in `UserListController` [#3184](https://github.com/GetStream/stream-chat-swift/pull/3184)
 - Fix duplicated watching channel calls when reconnecting [#3187](https://github.com/GetStream/stream-chat-swift/pull/3187)
+- Fix `/sync` call failing because of surpassing the 255 channels limit [#3188](https://github.com/GetStream/stream-chat-swift/pull/3188)
 ### 🔄 Changed
 - Do not retry rate-limited requests [#3182](https://github.com/GetStream/stream-chat-swift/pull/3182)
 - `UserListController` won't update automatically when a new user is added [#3184](https://github.com/GetStream/stream-chat-swift/pull/3184)
--- a/Sources/StreamChat/Repositories/SyncRepository.swift
+++ b/Sources/StreamChat/Repositories/SyncRepository.swift
@@ -191,7 +191,7 @@ class SyncRepository {
     private func getChannelIds(completion: @escaping ([ChannelId]) -> Void) {
         database.backgroundReadOnlyContext.perform {
             let request = ChannelDTO.allChannelsFetchRequest
-            request.fetchLimit = 1000
+            request.fetchLimit = 250
             request.propertiesToFetch = ["cid"]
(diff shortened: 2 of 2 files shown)


Sources/StreamChat/Repositories/SyncRepository.swift
```diff
@@ -191,7 +191,7 @@ class SyncRepository {
     private func getChannelIds(completion: @escaping ([ChannelId]) -> Void) {
         database.backgroundReadOnlyContext.perform {
             let request = ChannelDTO.allChannelsFetchRequest
-            request.fetchLimit = 1000
+            request.fetchLimit = 250
```
laevandus, line 194: When looking at the code it seems like, if there are 500 channels, then we would only get 250 of them and the other 250 is never synced, only when opening it manually with a channel controller.

Another interesting thing with syncing is that it pauses all the other API requests until sync repo is done. Not sure why it is a requirement.

I guess what I am saying is that sync repo should need a review and the current change is nice!

laevandus (approved):

Sources/StreamChat/Repositories/SyncRepository.swift
martinmitrevski: Didn't we say we will use 100? https://getstream.slack.com/archives/CRB9P6L03/p1712674377370179?thread_ts=1712674016.356819&cid=CRB9P6L03

Also, shouldn't we also cleanup/delete the rest of the channels?

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### 🐞 Fixed
 - Fix triggering user query calls whenever a new user was added in `UserListController` [#3184](https://github.com/GetStream/stream-chat-swift/pull/3184)
 - Fix duplicated watching channel calls when reconnecting [#3187](https://github.com/GetStream/stream-chat-swift/pull/3187)
+- Fix `/sync` call failing because of surpassing the 255 channels limit [#3188](https://github.com/GetStream/stream-chat-swift/pull/3188)
```
martinmitrevski, line 10: 250 channels limit

Sources/StreamChat/Repositories/SyncRepository.swift
nuno-vieira: I did not notice that comment sorry. I also had Auto merge enabled, so the PR got merged when @laevandus approved 😄

CHANGELOG.md
nuno-vieira: The limit is actually 255, not 250

Sources/StreamChat/Repositories/SyncRepository.swift
nuno-vieira: The cleanup maybe better to make it part of a different PR, since it needs a strategy where we will do this

PR #3190, opened 2024-05-10
[ChatStateLayer] Use a single channel list linker for linking and unlinking channels f…

laevandus:
…rom query

### 🔗 Issue Links

### 🎯 Goal

Use a single entity for linking and unlinking channels.

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift
+++ b/Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift
@@ -121,6 +121,10 @@ public class ChatChannelListController: DataController, DelegateCallable, DataSt
 
     private let filter: ((ChatChannel) -> Bool)?
     private let environment: Environment
+    private lazy var channelListLinker: ChannelListLinker = self.environment
+        .channelListLinkerBuilder(
+            query, filter, client.config, client.databaseContainer, worker
+        )
 
     /// Creates a new `ChannelListController`.
     ///
@@ -145,6 +149,7 @@ public class ChatChannelListController: DataController, DelegateCallable, DataSt
 
     override public func synchronize(_ completion: ((_ error: Error?) -> Void)? = nil) {
         startChannelListObserverIfNeeded()
+        channelListLinker.start(with: client.eventNotificationCenter)
         client.startTrackingChannelListController(self)
         updateChannelList(completion)
     }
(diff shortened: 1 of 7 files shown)


Sources/StreamChat/Workers/ChannelListLinker.swift
```diff
@@ -0,0 +1,128 @@
...
+            return filter(channel)
+        }
+
+        if clientConfig.isChannelAutomaticFilteringEnabled {
+            // When auto-filtering is enabled the channel will appear or not automatically if the
+            // query matches the DB Predicate. So here we default to saying it always belong to the current query.
+            return true
+        }
+
+        return false
```
laevandus, line 126: I went for the new implementation for this. The old was `deprecatedFallback() ?? true`.

Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift
```diff
@@ -255,78 +260,8 @@ public class ChatChannelListController: DataController, DelegateCallable, DataSt
...
-
-            self?.worker.startWatchingChannels(withIds: [channel.cid]) { error in
-                guard let error = error else { return }
-                log.warning(
-                    "Failed to start watching linked channel: \(channel.cid), error: \(error.localizedDescription)"
-                )
-            }
-        }
-    }
+    public func eventsController(_ controller: EventsController, didReceiveEvent event: Event) {}
```
martinmitrevski, line 264: do we still need this?

laevandus: No, but since it is public, I did not touch. If OK, I can throw it away.

laevandus: Do you think it is fine?

nuno-vieira: Yes we can removove this 👍 It is public, but not overridable. I think it was public because of the requirements of the Delegate being public. But no one will really call this function.

Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift
```diff
@@ -255,78 +260,8 @@ public class ChatChannelListController: DataController, DelegateCallable, DataSt
...
-
-        guard !client.config.isChannelAutomaticFilteringEnabled else {
-            // When auto-filtering is enabled the channel will appear or not automatically if the
-            // query matches the DB Predicate. So here we default to saying it always belong to the current query.
-            return true
-        }
-
-        // Fallback to legacy of checking if a channel should be linked to the current channel list query.
-        let deprecatedFallback: () -> Bool? = {
-            self.delegate?.controller(self, shouldAddNewChannelToList: channel)
```
nuno-vieira, line 308: This is really old, and most likely nobody is using this anymore. If they are, they shouldn't really. Better to remove this function all together

Sources/StreamChat/Workers/ChannelListLinker.swift
nuno-vieira: Yes, I guess we can remove the deprecatedFallback, but better if we also remove the old functions 👍

Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift
laevandus: I will add an item about removing deprecated methods to v5.

Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift
laevandus: Removed

martinmitrevski (approved):

nuno-vieira (approved):

Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift
nuno-vieira: The thing is that if we are not using them already, not even has a fallback, then there's no point in waiting for v5 🤔

laevandus: Let me do another PR after this to delete some unused APIs.

PR #3192, opened 2024-05-10
Change sync channels limit to 100

nuno-vieira:
### 🎯 Goal

Updates the channel limit updated here: https://github.com/GetStream/stream-chat-swift/pull/3188


martinmitrevski (commented):
Should we also update the changelog?

nuno-vieira:
@martinmitrevski The change log is already changed, and describes the error message, so it is not dependent on the solution (whatever is the new limit)

martinmitrevski (approved):

PR #3194, opened 2024-05-13
[ChatStateLayer] Share mark and mark unread message

laevandus:
### 🔗 Issue Links

### 🎯 Goal

Share the logic used for marking channel read.

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/Workers/ReadStateSender.swift
+++ b/Sources/StreamChat/Workers/ReadStateSender.swift
@@ -4,74 +4,96 @@
 
 import Foundation
 
-@available(iOS 13.0, *)
-extension Chat {
-    actor ReadStateSender {
-        private let cid: ChannelId
-        private let authenticationRepository: AuthenticationRepository
-        private let messageRepository: MessageRepository
-        private let channelUpdater: ChannelUpdater
-        
-        init(
-            cid: ChannelId,
-            channelUpdater: ChannelUpdater,
-            authenticationRepository: AuthenticationRepository,
-            messageRepository: MessageRepository
-        ) {
-            self.authenticationRepository = authenticationRepository
-            self.cid = cid
-            self.channelUpdater = channelUpdater
-            self.messageRepository = messageRepository
+final class ReadStateSender {
+    private let authenticationRepository: AuthenticationRepository
+    private let channelUpdater: ChannelUpdater
+    private let messageRepository: MessageRepository
+    @Atomic private var markingRead = false
+    @Atomic private(set) var isMarkedAsUnread = false
+    
+    init(
+        authenticationRepository: AuthenticationRepository,
+        channelUpdater: ChannelUpdater,
+        messageRepository: MessageRepository
+    ) {
+        self.authenticationRepository = authenticationRepository
+        self.channelUpdater = channelUpdater
+        self.messageRepository = messageRepository
+    }
+    
+    func markRead(_ channel: ChatChannel, completion: @escaping (Error?) -> Void) {
+        guard
+            !markingRead,
+            let currentUserId = authenticationRepository.currentUserId,
+            let currentUserRead = channel.reads.first(where: { $0.user.id == currentUserId }),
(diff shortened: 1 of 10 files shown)


Sources/StreamChat/Workers/ReadStateSender.swift
```diff
@@ -4,74 +4,96 @@
...
-            cid: ChannelId,
-            channelUpdater: ChannelUpdater,
-            authenticationRepository: AuthenticationRepository,
-            messageRepository: MessageRepository
-        ) {
-            self.authenticationRepository = authenticationRepository
-            self.cid = cid
-            self.channelUpdater = channelUpdater
-            self.messageRepository = messageRepository
+final class ReadStateSender {
```
laevandus, line 7: Moved controller's logic to this type which I created for the state layer. Now both use the same logic.

Sources/StreamChat/Workers/ReadStateSender.swift
```diff
@@ -4,74 +4,96 @@
...
-                    try await channelUpdater.markRead(cid: cid, userId: currentUserId)
+        }
+    }
+    
+    func markUnread(
+        from messageId: MessageId,
+        in channel: ChatChannel,
+        completion: @escaping (Result<ChatChannel, Error>) -> Void
+    ) {
+        guard !markingRead,
```
martinmitrevski, line 57: I guess it's ok we use the same bool for marking as unread, do you see any issues with it? (seems like it was like that before too)

Tests/StreamChatTests/Controllers/ChannelController/ChannelController_Tests.swift
```diff
@@ -4182,16 +4182,16 @@ final class ChannelController_Tests: XCTestCase {
         }
         client.setToken(token: .unique(userId: .unique))
 
+        let updater = try XCTUnwrap(env.channelUpdater)
```
martinmitrevski, line 4185: why was this needed?

martinmitrevski (approved):

Sources/StreamChat/Workers/ReadStateSender.swift
nuno-vieira, line 7: I would rename it to something like this:
```suggestion
final class ReadStateHandler {
```

The "Sender" postfix is used as a background worker, like the "MessageSender", so it is used fro components which are observing data changes and then queue and send requests async. 

This component the goal is just to reuse the State logic, kinda like we have the PaginationStateHandler. So I guess it makes sense to have consistancy and also have the "Handler" postfix

Sources/StreamChat/Workers/ReadStateSender.swift
nuno-vieira: I think it is important we use the same bool. We don't want to mark **unread**, while we are marking **read**, otherwise we will have race conditions. The problem is that it seems that we maybe lose some requests here. If it happens a user marks unread/read while one of the requests is being executed we don't do anything

nuno-vieira: But I would say that is not a big deal at the moment

laevandus: Currently I just replicated what controller was doing and did not want to change any logic there. But that said, it does not seem like 100% reliable implementation when many unread and read calls happen.

Sources/StreamChat/Workers/ReadStateSender.swift
laevandus: Sure, let's do that. I'll do another commit and fix all the call sites.

Tests/StreamChatTests/Controllers/ChannelController/ChannelController_Tests.swift
laevandus: This was because I needed to move the logic how we fetch the last read message. Before it uses observer, now it fetches from the DB which caused a delay and I needed to change tests to take this into account.

PR #3202, opened 2024-05-16
Keep the message order when sending multiple messages with failures

laevandus:
### 🔗 Issue Links

Resolves [ios-issues-tracking/819](https://github.com/GetStream/ios-issues-tracking/issues/819)

### 🎯 Goal

Fix the issue of incorrect order of messages when multiple messages are in the pending send state 

### 📝 Summary

**Before**
`MessageSender` tries to send queued messages one by one and even if one fails, it just tries to send the next one. This could change the order since the next one is delivered but the previous one is in failed state still and if I resend it, it gets ordered after the last succeeded message.
**After**
`MessageSender` fails all the messages in the queue after encountering a failure. User can then try to resend them explicitly.

### 🛠 Implementation

Fail all the messages in the queue when encountering a failure.

### ☑️ Contributor Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## StreamChat
 ### ✅ Added
 - Add a new state layer with async-await and observable state objects ([learn more](https://getstream.io/chat/docs/sdk/ios/state-layer/state-layer-overview/)) [#3177](https://github.com/GetStream/stream-chat-swift/pull/3177)
+### 🐞 Fixed
+- Keep the message order when sending multiple messages and one of them fails [#3202](https://github.com/GetStream/stream-chat-swift/pull/3202)
 
 # [4.55.0](https://github.com/GetStream/stream-chat-swift/releases/tag/4.55.0)
 _May 13, 2024_
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### ✅ Added
 - Add a new state layer with async-await and observable state objects ([learn more](https://getstream.io/chat/docs/sdk/ios/state-layer/state-layer-overview/)) [#3177](https://github.com/GetStream/stream-chat-swift/pull/3177)
 ### 🐞 Fixed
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## StreamChat
 ### ✅ Added
 - Add a new state layer with async-await and observable state objects ([learn more](https://getstream.io/chat/docs/sdk/ios/state-layer/state-layer-overview/)) [#3177](https://github.com/GetStream/stream-chat-swift/pull/3177)
+### 🐞 Fixed
+- Keep the message order when sending multiple messages and one of them fails [#3202](https://github.com/GetStream/stream-chat-swift/pull/3202)
```
nuno-vieira, line 10: Typo: ?
```suggestion
- Keep the message order when sending multiple messages when one of them fails [#3202](https://github.com/GetStream/stream-chat-swift/pull/3202)
```

Sources/StreamChat/Workers/Background/MessageSender.swift
```diff
@@ -195,31 +207,36 @@ private class MessageSendingQueue {
...
-                        self.eventsNotificationCenter.process(event)
-                    case let .failedToSendMessage(error):
-                        let event = NewMessageErrorEvent(messageId: request.messageId, error: error)
-                        self.eventsNotificationCenter.process(event)
-                    }
+                // If we failed to send a message we must fail all the next messages as well for preserving the message order
+                if let error = result.error, case .failedToSendMessage = error {
+                    sortedRequests.forEach { self?.handleSendRequestResult(request: $0, result: result) }
+                } else {
+                    self?.handleSendRequestResult(request: request, result: result)
```
nuno-vieira, line 218: I wonder if this could be configurable 🤔 Some customers might not want this behaviour. For some apps it could be more important if the delivery is done, and have as less errors as possible. 

The current approach has the con that if you send 10 messages, and the 3rd one fails, you will need to manually retry the other 7 messages :/

laevandus: I was also thinking about this again. Sounds like we should go for a full solution instead which is when failure is encountered because of internet connection, we fail all the messages in the queue. As soon as websocket connects, we automatically retry all the failed messages.

laevandus:
I think we need to have in-depth plan on how to handle all the different use-cases first if we change this. I'll close it for now.

PR #3203, opened 2024-05-16
[Threads v2] Add `ChatThreadListController`

nuno-vieira:
### 🔗 Issue Links

### 🎯 Goal
Adds the implementaton of `ChatThreadListController` to be able to fetch the threads of the current user.

### 📝 Summary
- Implementation of `ChatThreadListController` to fetch the threads of the current user (Internal for now, until UI Components are exposed)
- Add default arguments to `ThreadListQuery`, since the optional arguments were confusing and unclear. It is better to have explicit default arguments on the client side.
- Make `watch` a required argument of `ThreadListQuery` so that it is more aligned to the future APIs (like the modern state layer). The goal is for the developer to make a decision on whether he wants to observe changes or not and not make that decision ourselves.

### 🧪 Manual Testing Notes
N/A

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/Controllers/ThreadListController/ThreadListController.swift
+++ b/Sources/StreamChat/Controllers/ThreadListController/ThreadListController.swift
@@ -5,12 +5,17 @@
 import CoreData
 import Foundation
 
-extension ChatClient {
-    /// Creates a new `ThreadListController` with the provided thread query.
-    /// - Returns: A new instance of `ChatThreadListController`.
-    internal func threadListController(query: ThreadListQuery) -> ChatThreadListController {
-        .init(query: query, client: self)
-    }
+/// `ChatThreadListController` uses this protocol to communicate changes to its delegate.
+internal protocol ChatThreadListControllerDelegate: DataControllerStateDelegate {
+    /// The controller changed the list of observed threads.
+    ///
+    /// - Parameters:
+    ///   - controller: The controller emitting the change callback.
+    ///   - changes: The change to the list of threads.
+    func controller(
+        _ controller: ChatThreadListController,
+        didChangeThreads changes: [ListChange<ChatThread>]
+    )
 }
 
 /// `ChatThreadListController` is a controller class which allows querying and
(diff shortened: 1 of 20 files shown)


Sources/StreamChat/Workers/ThreadListUpdater.swift
```diff
@@ -0,0 +1,44 @@
+//
+// Copyright © 2024 Stream.io Inc. All rights reserved.
+//
+
+import CoreData
+
+/// Makes a thread list query call to the backend and updates the local storage with the results.
+class ThreadListUpdater: Worker {
+    /// The cursor to be used for fetching a new page of threads.
+    var nextCursor: String?
```
laevandus, line 10: Ideally updater would be without any state but then the completion needs to change the type to something which contains the array of threads and the cursor. But on the other-hand, this updater is not meant to be shared anyway so it is all right.

Sources/StreamChat/Database/DTOs/ThreadDTO.swift
```diff
@@ -174,4 +210,10 @@ extension NSManagedObjectContext {
 
         return threadDTO
     }
+
+    func deleteAllThreads() throws {
+        let fetchRequest: NSFetchRequest<ThreadDTO> = NSFetchRequest(entityName: ThreadDTO.entityName)
```
laevandus, line 215: `NSBatchDeleteRequest` could be an option here, but since we do not expect to have so many threads, so this is also OK. Also batch delete needs extra code for contexts.

laevandus (approved):

Sources/StreamChat/Workers/ThreadListUpdater.swift
```diff
@@ -0,0 +1,44 @@
+//
+// Copyright © 2024 Stream.io Inc. All rights reserved.
+//
+
+import CoreData
+
+/// Makes a thread list query call to the backend and updates the local storage with the results.
+class ThreadListUpdater: Worker {
```
martinmitrevski, line 8: I would prefer if we use repositories for the new features. We anyway want to migrate away from the workers.

Sources/StreamChat/Workers/ThreadListUpdater.swift
martinmitrevski: I agree, it should be stateless. And actually I don't see why it shouldn't be reused, especially if we make it a repository.

nuno-vieira: The reason I put here the cursor it was because of those reasons you stated, It would require exposing the Cursor, and besides that, we use the Updaters in the State Layer, so it would be one less thing to be duplicated. But I also agree, this should not be part of the Updater

Sources/StreamChat/Controllers/ThreadListController/ThreadListController.swift
```diff
@@ -0,0 +1,198 @@
...
+
+    /// The updater used to fetch the remote data and communicate with servers.
+    private lazy var updater: ThreadListUpdater = self.environment
+        .threadListUpdaterBuilder(
+            client.databaseContainer,
+            client.apiClient
+        )
+
+    /// A Boolean value that returns whether pagination is finished.
+    internal private(set) var hasLoadedAllOlderThreads: Bool = false
```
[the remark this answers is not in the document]
nuno-vieira: Yes that is why I keep it like this. Even tho it is not thread-safe, we never had issues with it at the moment, so I left it the same 👍

Sources/StreamChat/Database/DTOs/ThreadDTO.swift
nuno-vieira: Yeah, my first approach was with `NSBatchDeleteRequest`, and even used mergeChanges etc.. but I whenever I opened the app (Not fresh login) I would get duplicated threads, and I could not figure out why

Sources/StreamChat/Workers/ThreadListUpdater.swift
nuno-vieira: With our current architecture, our Updaters are already our Repositories, IMO introducing the same thing but with a different name just causes incosistencies (like we have the moment). If you check the `MessageRepository`, it is pretty much the same thing as the Updaters, but with a different name. The state layer is also using the Updaters for now. So I would only change this when we start working on v5. For now I would favor consistency

nuno-vieira: We also just add the `ReactionListController` which was also created with `ReactionListUpdater`, so it does not make sense for this one to be different IMO

martinmitrevski: I would prefer to use Repositories tbh, also for the reactions, which is probably something I missed in the review (probably because of the urgency of the feature). Now we're in a state where we migrated some stuff, and then we keep going back to the old names again.

PR #3207, opened 2024-05-17
[Threads v2] Add `ChatThreadListController`

nuno-vieira:
### 🔗 Issue Links

### 🎯 Goal
Adds the implementaton of `ChatThreadListController` to be able to fetch the threads of the current user.

### 📝 Summary
- Implementation of `ChatThreadListController` to fetch the threads of the current user (Internal for now, until UI Components are exposed)
- Add default arguments to `ThreadListQuery`, since the optional arguments were confusing and unclear. It is better to have explicit default arguments on the client side.
- Make `watch` a required argument of `ThreadListQuery` so that it is more aligned to the future APIs (like the modern state layer). The goal is for the developer to make a decision on whether he wants to observe changes or not and not make that decision ourselves.

### 🧪 Manual Testing Notes
N/A

### ☑️ Contributor Checklist


Sources/StreamChat/Controllers/ThreadListController/ThreadListController.swift
```diff
@@ -0,0 +1,211 @@
...
+        updater.loadThreads(
+            query: query
+        ) { [weak self] result in
+            switch result {
+            case let .success(threadListResponse):
+                self?.threadListObserver.refreshItems { [weak self] in
+                    self?.callback {
+                        let threads = threadListResponse.threads
+                        self?.nextCursor = threadListResponse.next
+                        self?.state = .remoteDataFetched
```
nuno-vieira, line 123: @laevandus I move this to inside the callback function, so for 99% of cases it will be thread safe, since these functions are called from the MainThread, and the callback will be also on the main thread (unless customers change this, which usually is not the case)

martinmitrevski (approved):
LGTM! ✅

PR #3209, opened 2024-05-17
Fetch tags for release workflow

testableapple:


.github/workflows/start-new-release.yml
```diff
@@ -3,10 +3,6 @@ name: "Start new release"
 on:
   workflow_dispatch:
     inputs:
-      branch:
-        description: 'Start release from this branch'
-        default: 'develop'
-        required: true
```
testableapple, line 9: `develop` is the default branch out of the box, so there is no need for the additional input field.

.github/workflows/start-new-release.yml
```diff
@@ -19,7 +15,7 @@ jobs:
     steps:
     - uses: actions/checkout@v4.1.1
       with:
-          ref: ${{ github.event.inputs.branch }}
+        fetch-depth: 0 # to fetch git tags
```
testableapple, line 18: This is the actual fix.

nuno-vieira (approved):

PR #3220, opened 2024-05-23
Low-level client support for Polls

martinmitrevski:
### 🔗 Issue Links

Resolves https://github.com/GetStream/ios-issues-tracking/issues/795.

### 🎯 Goal

Add support in the LLC for polls.

### 📝 Summary

- Exposed a `PollController` for dealing with a poll
- Exposed a `PollVoteListController` for dealing with vote querying
- Exposed new models and events

### 🛠 Implementation

Similar to how we do things in other places.

### 🎨 Showcase

| Before | After |
| ------ | ----- |
|  img   |  img  |

### 🧪 Manual Testing Notes

The UI implementation is part of the `polls` branch in the SwiftUI SDK. It should be used for testing there.

### ☑️ Contributor Checklist

### 🎁 Meme

Alt Text


Sources/StreamChat/Controllers/PollController/PollController.swift
```diff
@@ -0,0 +1,283 @@
...
+    ///
+    /// - Parameters:
+    ///   - text: The text of the suggested poll option.
+    ///   - position: An optional position for the suggested option.
+    ///   - custom: An optional dictionary of custom data for the suggested option.
+    ///   - completion: A closure to be called upon completion, with an optional `Error` if something went wrong.
+    public func suggestPollOption(
+        text: String,
+        position: Int? = nil,
+        custom: [String: RawJSON]? = nil,
```
laevandus, line 188: We use `userData` quite a bit for cases like this. Would it make sense to unify the naming here?

Sources/StreamChat/Controllers/PollController/PollController.swift
```diff
@@ -0,0 +1,283 @@
...
+    internal var multicastDelegate: MulticastDelegate<PollControllerDelegate> = .init() {
+        didSet {
+            stateMulticastDelegate.set(mainDelegate: multicastDelegate.mainDelegate)
+            stateMulticastDelegate.set(additionalDelegates: multicastDelegate.additionalDelegates)
+        }
+    }
+    
+    let ownVotesQuery: PollVoteListQuery
+    
+    private var pollObserver: EntityDatabaseObserverWrapper<Poll, PollDTO>?
```
laevandus, line 57: Is there any specific reason why both observers are not created with `lazy var` properties?

Sources/StreamChat/Controllers/PollController/PollVoteListController.swift
```diff
@@ -0,0 +1,203 @@
...
+        }
+        _basePublishers = BasePublishers(controller: self)
+        return _basePublishers as? BasePublishers ?? .init(controller: self)
+    }
+    
+    private let eventsController: EventsController
+    private let pollsRepository: PollsRepository
+    private let environment: Environment
+
+    /// Creates a new `UserListController`.
```
laevandus, line 109: > `PollVoteListController`

Sources/StreamChat/Database/DTOs/PollDTO.swift
```diff
@@ -0,0 +1,177 @@
...
+    @NSManaged var pollDescription: String?
+    @NSManaged var enforceUniqueVote: Bool
+    @NSManaged var id: String
+    @NSManaged var name: String
+    @NSManaged var updatedAt: DBDate?
+    @NSManaged var voteCount: Int
+    @NSManaged var custom: Data?
+    @NSManaged var voteCountsByOption: [String: Int]?
+    @NSManaged var isClosed: Bool
+    @NSManaged var maxVotesAllowed: NSNumber?
```
laevandus, line 23: Int?

Sources/StreamChat/Database/DTOs/PollVoteDTO.swift
```diff
@@ -0,0 +1,226 @@
...
+    }
+    
+    @discardableResult
+    func savePollVote(
+        payload: PollVotePayload,
+        query: PollVoteListQuery?,
+        cache: PreWarmedCache?
+    ) throws -> PollVoteDTO {
+        guard let poll = try poll(id: payload.pollId) else {
+            throw ClientError.Unexpected()
```
laevandus, line 99: `MessageDoesNotExist` but for polls would be nice since then it prints out the id at least, e.g.`PollDoesNotExist(pollId:)`

Sources/StreamChat/Database/DTOs/PollVoteDTO.swift
```diff
@@ -0,0 +1,226 @@
...
+            option: option,
+            user: userDto,
+            context: self,
+            cache: nil
+        )
+        
+        dto.createdAt = Date().bridgeDate
+        dto.updatedAt = Date().bridgeDate
+        dto.pollId = pollId
+        dto.isAnswer = true // TODO: fix
```
laevandus, line 164: Highlighting the `TODO`

Sources/StreamChat/Database/DTOs/PollVoteDTO.swift
```diff
@@ -0,0 +1,226 @@
...
+    }
+    
+    func pollVote(id: String, pollId: String) throws -> PollVoteDTO? {
+        PollVoteDTO.load(voteId: id, pollId: pollId, context: self)
+    }
+    
+    func removePollVote(with id: String, pollId: String) throws {
+        guard let dto = try pollVote(id: id, pollId: pollId) else {
+            // TODO: error
+            throw ClientError.Unknown()
```
laevandus, line 192: Highlighting the `TODO` e.g. `PollDoesNotExist(pollId:)`

Sources/StreamChat/Database/DatabaseSession.swift
```diff
@@ -516,6 +562,64 @@ extension DatabaseSession {
...
+        if let vote = payload.vote {
+            if payload.eventType == .pollVoteRemoved {
+                if let dto = try? pollVote(id: vote.id, pollId: vote.pollId) {
+                    delete(pollVote: dto)
+                }
+            } else if payload.eventType == .pollVoteChanged {
+                var voteUpdated = false
+                let userId = vote.userId ?? "anon"
+                if let optionId = vote.optionId, !optionId.isEmpty {
+                    let id = "\(optionId)-\(vote.pollId)-\(userId)"
```
laevandus, line 575: This appeared already once before, wondering if it makes sense to put this into a separate `static func` for making sure we never have issues with `id` creation.

Sources/StreamChat/Models/ChatMessage.swift
```diff
@@ -202,6 +202,8 @@ public struct ChatMessage {
     public var readByCount: Int { _readByCount }
 
     @CoreDataLazy internal var _readByCount: Int
+    
+    public let poll: Poll?
```
laevandus, line 206: Since it is public and most of the public properties have documentation then I guess it would be nice to have something here as well although it is pretty self-explanatory.

Sources/StreamChat/Controllers/PollController/PollController.swift
martinmitrevski: It's actually `extraData` that we use currently, but we should use `custom` to align with other platforms. Now, the question is - do we want to add the new stuff the correct way, or we keep consistency with the rest of the codebase. Not sure, wdyt?

Sources/StreamChat/Controllers/PollController/PollController.swift
martinmitrevski: Nope, I will make them lazy

Sources/StreamChat/Database/DTOs/PollDTO.swift
martinmitrevski: Wasn't able to make it optional, it defaults to 0 if it's an Int in Core Data. And we need this one to be optional.

Sources/StreamChat/Database/DTOs/PollVoteDTO.swift
martinmitrevski: knew I forgot sth 😅 Yeah, new errors need to be introduced.

Sources/StreamChat/Database/DatabaseSession.swift
martinmitrevski: yes, good point, will add it

Sources/StreamChat/Models/Poll.swift
```diff
@@ -0,0 +1,71 @@
...
+    
+    /// The date and time when the poll was last updated.
+    /// This property is optional and may be `nil`.
+    public let updatedAt: Date?
+    
+    /// The count of votes received for the poll.
+    public let voteCount: Int
+    
+    /// A dictionary containing custom fields associated with the poll.
+    public let custom: [String: RawJSON]
```
nuno-vieira, line 42: Ideally this should be extraData, since we use it everywhere. IMO we should only rename it to custom in v5
```suggestion
    public let extraData: [String: RawJSON]
```

Sources/StreamChat/Models/Poll.swift
```diff
@@ -0,0 +1,71 @@
...
+    /// A boolean indicating whether the poll is closed.
+    public let isClosed: Bool
+    
+    /// The maximum number of votes allowed per user.
+    /// This property is optional and may be `nil`.
+    public let maxVotesAllowed: Int?
+    
+    /// A string representing the visibility of the voting process.
+    /// This property is optional and may be `nil`.
+    public let votingVisibility: String?
```
nuno-vieira, line 57: Don't we have a data structure to represent the `VotingVisiblity`? 🤔

Sources/StreamChat/Models/Poll.swift
```diff
@@ -0,0 +1,71 @@
...
+    
+    /// The user who created the poll.
+    /// This property is optional and may be `nil`.
+    public let createdBy: ChatUser?
```
nuno-vieira, line 61: How can this be optional? 🤔 Could the Poll be created by the system somehow?

Sources/StreamChat/Query/PollVoteListQuery.swift
```diff
@@ -0,0 +1,65 @@
...
+
+import Foundation
+
+/// A query used for querying specific reactions from a message.
+public struct PollVoteListQuery: Encodable {
+    public var pollId: String
+    public var optionId: String?
+    /// The pagination information to query the reactions.
+    public var pagination: Pagination
+    /// The filter details to query the reactions.
```
nuno-vieira, line 13:
```suggestion
    /// The filter details to query the votes.
```

Sources/StreamChat/Query/PollVoteListQuery.swift
```diff
@@ -0,0 +1,65 @@
...
+/// Non extra-data-specific filer keys for reaction list.
+public extension FilterKey where Scope: AnyVoteListFilterScope {
+    /// A filter key for matching the reaction type
+    static var optionId: FilterKey<Scope, String> { "option_id" }
+
+    /// A filter key for matching the user id of the reaction's author.
+    static var userId: FilterKey<Scope, UserId> { "user_id" }
+    
+    static var pollId: FilterKey<Scope, String> { "poll_id" }
+}
```
nuno-vieira, line 55: Copy pasting consequences here 😄 We should rename where Reaction is present

Sources/StreamChat/Repositories/PollsRepository.swift
```diff
@@ -0,0 +1,269 @@
...
+    func createPoll(
+        name: String,
+        allowAnswers: Bool? = nil,
+        allowUserSuggestedOptions: Bool? = nil,
+        description: String? = nil,
+        enforceUniqueVote: Bool? = nil,
+        maxVotesAllowed: Int? = nil,
+        votingVisibility: String? = nil,
+        options: [PollOption]? = nil,
+        custom: [String: RawJSON]? = nil,
```
nuno-vieira, line 25: nit: Since the repository/updater is internal, I find it more useful not to make these properties optional, so that when we use it in the controllers, we don't forget to reason about all the available properties

Sources/StreamChat/Repositories/PollsRepository.swift
```diff
@@ -0,0 +1,269 @@
...
+        } completion: { [weak self] error in
+            if error == nil {
+                self?.apiClient.request(
+                    endpoint: .removePollVote(
+                        messageId: messageId,
+                        pollId: pollId,
+                        voteId: voteId
+                    )
+                ) {
+                    completion?($0.error)
```
nuno-vieira, line 132: If the request fails, shouldn't we add back the vote?

Sources/StreamChat/Repositories/PollsRepository.swift
```diff
@@ -0,0 +1,269 @@
...
+            sort: sort,
+            filter: filter
+        )
+        apiClient.request(
+            endpoint: .queryPollVotes(pollId: pollId, queryPollVotesRequest: request),
+            completion: { [weak self] result in
+                guard let self else { return }
+                switch result {
+                case let .success(response):
+                    database.write { session in
```
nuno-vieira, line 224: Fix compilation issue:
```suggestion
                    self.database.write { session in
```

Tests/StreamChatTests/Controllers/PollsControllers/PollController_Tests.swift
```diff
@@ -0,0 +1,369 @@
...
+        let query = controller.ownVotesQuery
+        try client.databaseContainer.writeSynchronously { session in
+            try session.savePollVotes(payload: response, query: query, cache: nil)
+        }
+
+        XCTAssertEqual(controller.poll?.id, pollId)
+        XCTAssertEqual(controller.ownVotes.count, ownVotes.count)
+        
+        controller = nil
+        client = nil
```
nuno-vieira, line 195: Why is this here? This looks a bit suspicious 🤔

Tests/StreamChatTests/Controllers/PollsControllers/PollController_Tests.swift
```diff
@@ -0,0 +1,369 @@
...
+        controller.castPollVote(answerText: nil, optionId: "123") {
+            completionError = $0
+        }
+
+        // Simulate network response with the error.
+        let networkError = TestError()
+        client.mockPollsRepository.castPollVote_completion?(networkError)
+
+        // Assert error is propagated.
+        AssertAsync.willBeEqual(completionError as? TestError, networkError)
```
nuno-vieira, line 212: nit: Whenever we can, it would be nice not to use AssertAsync, and just use regular expectations. We agreed on moving out from custom frameworks in tests.

Tests/StreamChatTests/Repositories/PollsRepository_Tests.swift
```diff
@@ -0,0 +1,479 @@
...
+        let referenceEndpoint: Endpoint<PollPayloadResponse> = .createPoll(
+            createPollRequest: .init(name: pollName)
+        )
+        XCTAssertEqual(apiClient.request_endpoint, AnyEndpoint(referenceEndpoint))
+    }
+    
+    // MARK: - Closing poll
+    
+    func test_closePoll_whenSuccessfull() {
+        let completionCalled = expectation(description: "completion called")
```
nuno-vieira, line 77: Using expectations is much better and familiar to everyone 👍 It would be nice that we kept consistency and not use AssertAsync in someplaces and others expecations (unless it is necessary to use AssertAsync because of bad design)

nuno-vieira (commented):
Looks great overall !👌  Added a couple of suggestions and questions 👍 

Btw, test_removingAllData() is failing because you need to provide there the new entities.

It would also be better if next time we split the PR maybe into the Poll List, and the Poll only. Otherwise it is a bit harder to revie the 4k lines of code, and Github does not handle well just a big PR 😄

Sources/StreamChat/Controllers/PollController/PollController.swift
```diff
@@ -0,0 +1,277 @@
...
+        self.pollId = pollId
+        self.environment = environment
+        ownVotesQuery = PollVoteListQuery(
+            pollId: pollId,
+            optionId: nil,
+            pagination: .init(pageSize: 25, cursor: nil),
+            filter: .and(
+                [.equal(.userId, to: client.currentUserId ?? ""), .equal(.pollId, to: pollId)]
+            )
+        )
```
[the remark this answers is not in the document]
martinmitrevski: This query fetches the current user's votes. The backend allows max of 10 votes per user, but I put it to the max of the query just in case. Having this configurable shouldn't be possible for customers.

Sources/StreamChat/Controllers/PollController/PollController.swift
```diff
@@ -0,0 +1,277 @@
...
+    
+    /// Casts a vote for a poll.
+    ///
+    /// - Parameters:
+    ///   - answerText: An optional text answer for the poll.
+    ///   - optionId: An optional identifier for the poll option.
+    ///   - completion: A closure to be called upon completion, with an optional `Error` if something went wrong.
+    public func castPollVote(
+        answerText: String?,
+        optionId: String?,
```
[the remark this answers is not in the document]
martinmitrevski: Good point, let me throw an error here

Sources/StreamChat/Controllers/PollController/PollController.swift
```diff
@@ -0,0 +1,277 @@
...
+    }
+}
+
+/// Represents the visibility of votes in a poll.
+public enum VotingVisibility: String {
+    /// Votes are public and can be seen by everyone.
+    case `public`
+    /// Votes are anonymous and cannot be attributed to individual users.
+    case anonymous
+}
```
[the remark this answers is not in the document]
martinmitrevski: good point, will update just in case

Sources/StreamChat/Query/PollVoteListQuery.swift
laevandus: One more above here

Sources/StreamChat/Models/Poll.swift
martinmitrevski: yep, forgot to change it, thx

Sources/StreamChat/Models/Poll.swift
martinmitrevski: it's optional in the spec. My assumption is for polls created anonymously or by the backend.

Sources/StreamChat/Query/PollVoteListQuery.swift
martinmitrevski: hahah 🤦 fixed them, thx

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
```diff
@@ -715,37 +718,64 @@ public class ChatChannelController: DataController, DelegateCallable, DataStoreP
...
+    ///   - votingVisibility: The visibility setting for the poll.
+    ///   - options: Optional array of predefined poll options.
+    ///   - custom: Optional dictionary for any custom data associated with the poll.
+    ///   - completion: A closure to be executed once the poll is created, returning either a `MessageId` on success or an `Error` on failure.
+    public func createPoll(
+        name: String,
+        allowAnswers: Bool? = nil,
+        allowUserSuggestedOptions: Bool? = nil,
+        description: String? = nil,
+        enforceUniqueVote: Bool? = nil,
```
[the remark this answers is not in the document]
nuno-vieira: The problem with this is that it is not explicit, so the default value is not transparent to the developer. We at least need to say it in the code comments. 

For example, in `ThreadListQuery` all the values are also nullable. But if I did not put the values from the backend on the SDK, then the developer would need to go to the Backend API to check what the defaults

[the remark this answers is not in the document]
martinmitrevski: I documented the defaults in the method - safer to pass nil over a default value.

Sources/StreamChat/Controllers/PollController/PollController.swift
```diff
@@ -0,0 +1,291 @@
...
+
+extension PollController {
+    struct Environment {
+        var pollObserverBuilder: (
+            _ isBackgroundMappingEnabled: Bool,
+            _ database: DatabaseContainer,
+            _ fetchRequest: NSFetchRequest<PollDTO>,
+            _ itemCreator: @escaping (PollDTO) throws -> Poll,
+            _ fetchedResultsControllerType: NSFetchedResultsController<PollDTO>.Type
+        ) -> EntityDatabaseObserverWrapper<Poll, PollDTO> = {
```
laevandus, line 267: Shorter `= EntityDatabaseObserverWrapper.init` should also work here but so important. Same with the other observer.

Sources/StreamChat/Controllers/PollController/PollVoteListController.swift
```diff
@@ -0,0 +1,204 @@
...
+            _ apiClient: APIClient
+        ) -> PollsRepository = PollsRepository.init
+
+        var createPollListDatabaseObserver: (
+            _ isBackgroundMappingEnabled: Bool,
+            _ database: DatabaseContainer,
+            _ fetchRequest: NSFetchRequest<PollVoteDTO>,
+            _ itemCreator: @escaping (PollVoteDTO) throws -> PollVote
+        )
+            -> ListDatabaseObserverWrapper<PollVote, PollVoteDTO> = {
```
laevandus, line 189: Alternative: `=  ListDatabaseObserverWrapper.init` (not important)

laevandus:
I guess state layer support comes with a separate PR?

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
```diff
@@ -715,37 +718,64 @@ public class ChatChannelController: DataController, DelegateCallable, DataStoreP
...
+            votingVisibility: votingVisibility?.rawValue,
+            options: options,
+            custom: extraData
+        ) { [weak self] result in
+            guard let self else { return }
+            switch result {
+            case let .success(poll):
+                self.createNewMessage(text: "", poll: poll, completion: completion)
+            case let .failure(error):
+                completion(.failure(error))
```
laevandus, line 778: We should use `callback {}` here for calling the completion handler like controllers do.

Sources/StreamChat/Controllers/PollController/PollController.swift
```diff
@@ -0,0 +1,291 @@
...
+        }
+        
+        pollsRepository.castPollVote(
+            messageId: messageId,
+            pollId: pollId,
+            answerText: answerText,
+            optionId: optionId,
+            currentUserId: client.currentUserId,
+            query: ownVotesQuery,
+            completion: completion
```
laevandus, line 179: We should use `callback {}` here for completion handler.

Sources/StreamChat/Controllers/PollController/PollController.swift
```diff
@@ -0,0 +1,291 @@
...
+    ///   - completion: A closure to be called upon completion, with an optional `Error` if something went wrong.
+    public func removePollVote(
+        voteId: String,
+        completion: ((Error?) -> Void)? = nil
+    ) {
+        pollsRepository.removePollVote(
+            messageId: messageId,
+            pollId: pollId,
+            voteId: voteId,
+            completion: completion
```
laevandus, line 196: We should use `callback {}` here for completion handler.

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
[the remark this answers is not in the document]
nuno-vieira: IMO, I think it is actually safer to have the default values on the SDK side. Because if the customer does not pass anything, and if the backend changes the values (hopefully this does not happen) it could be a silent breaking change for the customer.

Sources/StreamChat/Database/DTOs/PollVoteDTO.swift
```diff
@@ -46,7 +46,14 @@ class PollVoteDTO: NSManagedObject {
...
     static func load(voteId: String, pollId: String, context: NSManagedObjectContext) -> PollVoteDTO? {
         let request = fetchRequest(for: voteId, pollId: pollId)
-        return load(by: request, context: context).first
+        let results = load(by: request, context: context)
+        if results.count > 1 {
+            for i in 1..<results.count {
+                let dto = results[i]
+                context.delete(dto)
+            }
+        }
```
nuno-vieira, line 55: Can you explain this one? Maybe add a comment? I get the feeling this should not be part of the `load` function, since it is doing more than just loading, should this logic be outside of the load function?

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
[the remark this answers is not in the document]
martinmitrevski: I disagree with that - if the backend says a value is optional, then we shouldn't be making assumptions by sending default values. We already had few issues because of this, last example was in the video SDK last week.
As for your concern - the no-breaking changes policy applies for both frontend and backend, so as long as we respect the contract it's gonna be ok.

Sources/StreamChat/Database/DTOs/PollVoteDTO.swift
martinmitrevski: actually removed it, found a better way

Sources/StreamChat/Controllers/PollController/PollController.swift
martinmitrevski: fixed

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
martinmitrevski: fixed

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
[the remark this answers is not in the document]
nuno-vieira: It depends on the use case ofc. In some scenarios, making the default values explicit on the SDK is easier for the customer, because it does not require the customer to read the API documentation. Even if we add a comment saying what is the default, it can easily be out of sync with the backend. For example, queries, it is useful to see the default values in the SDK

Sources/StreamChat/Database/DTOs/PollDTO.swift
```diff
@@ -0,0 +1,180 @@
...
+        pollDto.voteCount = payload.voteCount
+        pollDto.custom = try JSONEncoder.default.encode(payload.custom)
+        pollDto.voteCountsByOption = payload.voteCountsByOption
+        pollDto.isClosed = payload.isClosed ?? false
+        if let maxVotesAllowed = payload.maxVotesAllowed {
+            pollDto.maxVotesAllowed = NSNumber(value: maxVotesAllowed)
+        }
+        pollDto.votingVisibility = payload.votingVisibility
+        
+        pollDto.createdBy = UserDTO.loadOrCreate(id: payload.createdById, context: self, cache: cache)
```
laevandus, line 125: Seems like we missing a line here which stores the user. `payload.createdBy` is `UserPayload` and we are not saving it here so this can return nil.

nuno-vieira (approved):

laevandus (approved):

PR #3222, opened 2024-05-24
[Threads v2] Add mark read and mark unread for Threads

nuno-vieira:
### 🔗 Issue Links

### 🎯 Goal
Adds mark read and mark unread functionality to Threads V2.

### 📝 Summary
- Adds mark read and mark unread functionality to Threads V2.
- Fixes `notification.mark_unread` not parsing if `lastReadMessageId` was not available

### 🛠 Implementation
- Adds `ChatMessageController.markThreadRead()`
- Adds `ChatMessageController.markThreadUnread()`
- Adds threads unread count to `UnreadCount`
- Handles thread unread events
- Handles thread read events

Note: For now, introducing a new `ThreadController` does not make sense, since our Thread Views use the `MessageController` for handling threads. 

Some things are missing, which will be added in the next and final PR:
- Optimistic unread/read updates
- New Unread Endpoint

### 🧪 Manual Testing Notes
N/A

### ☑️ Contributor Checklist


Sources/StreamChat/Controllers/MessageController/MessageController.swift
```diff
@@ -718,6 +718,23 @@ public class ChatMessageController: DataController, DelegateCallable, DataStoreP
...
+    /// Marks the thread read if this message is the root of a thread.
+    public func markThreadRead(completion: ((Error?) -> Void)? = nil) {
+        messageUpdater.markThreadRead(cid: cid, threadId: messageId) { error in
+            completion?(error)
+        }
+    }
+
+    /// Marks the thread unread if this message is the root of a thread.
+    public func markThreadUnread(completion: ((Error?) -> Void)? = nil) {
+        messageUpdater.markThreadUnread(
```
nuno-vieira, line 731: TODO: Forgot to set this to internal

Sources/StreamChat/Database/DatabaseSession.swift
```diff
@@ -411,6 +411,10 @@ protocol ThreadDatabaseSession {
         cache: PreWarmedCache?
     ) throws -> ThreadDTO
 
+    /// Updates the a thread with details from a thread event.
```
martinmitrevski, line 414: "the a"

Sources/StreamChat/Models/UnreadCount.swift
```diff
@@ -11,20 +11,25 @@ public struct UnreadCount: Decodable, Equatable {
         case userUnreadCount = "user"
         case channels = "unread_channels"
         case messages = "total_unread_count"
+        case threads = "unread_threads"
     }
 
-    /// The default value representing no unread channels and messages.
-    public static let noUnread = UnreadCount(channels: 0, messages: 0)
+    /// The default value representing no unread channels, messages and threads.
+    public static let noUnread = UnreadCount(channels: 0, messages: 0, threads: nil)
```
martinmitrevski, line 18: why not 0?

martinmitrevski (approved):
LGTM ✅  Just 2 very small things

Sources/StreamChat/Models/UnreadCount.swift
nuno-vieira: I initially had 0, then switched to `nil` because I was thinking about customers using this to check if there are unreads or not. Hopefully they are not, and this should only be used for initializations. Because if a customer was previously using this to check if there are reads or not, this is breaking. But it is breaking either if we set it to `nil` or `0`. So I switched back to 0 because it makes more sense.

nuno-vieira: What I mean by breaking is that, if somehow a customer is doing something like:
```
if unreadCount == .noUnread {
   // do something
} 
```

This can stop working, because most likely the customer will have now unread threads, and some customers won't care about the threads for now

nuno-vieira: Maybe we could even deprecated this `noUnread` ? 🤔 It is not a good idea at all

nuno-vieira: Yes, but the property is misleading, it is not clear what "noUnread" means: Do you count all the messages? Do you count channels and messages? Do you also count threads? It is just better to access the properties directly IMO

nuno-vieira: For now I'll keep it like this, but on the next PR I'll revisit this

PR #3223, opened 2024-05-27
Implement user block feature

testableapple:
### 🔗 Issue Links

Resolve https://github.com/GetStream/ios-issues-tracking/issues/788

### 🎯 Goal

Implement user blocking feature.

### 🧪 Manual Testing Notes

1. On first entry
- Enable blocking feature in configuration
- Open channel long-press participant's message
- Verify action item (`Block user` / `Unblock user`)

2. On reconnect
- Enable blocking feature in configuration
- Open channel
- Block user
- Log out and log in as the same user
- Open channel long-press participant's message
- Verify action item (`Block user` / `Unblock user`)

3. Get the list of blocked users
- Enable blocking feature in configuration
- Open channel
- Block user
- Open channel debug menu
- Open blocked users view
- Verify that blocked user is in the list

4. 1:1 channel
- Enable blocking feature in configuration
- Open a 1:1 channel
- Block user
- Send a message as a blocked user
- There should be no messages from the blocked user

5. Group channel
- Enable blocking feature in configuration
- Open a group channel
- Block some user
- Send a message as a blocked user
- The message should be visible to the main user

6. The visibility of the 1:1 channel with a blocked user
- Enable blocking feature in configuration
- Open a 1:1 channel
- Block user
- Go back to channel list
- The 1:1 channel with the blocked user should be hidden
- Open a group channel
- Unblock that user
- Go back to channel list
- The 1:1 channel with the unblocked user should be visible

### ☑️ Contributor Checklist

### 🎁 Meme

gif


Sources/StreamChat/APIClient/Endpoints/ModerationEndpoints.swift
```diff
@@ -16,6 +16,44 @@ extension Endpoint {
...
+    
+    static func loadBlockedUsers() -> Endpoint<BlockedUsersResponse> {
+        .init(
+            path: .blockUser,
+            method: .get,
+            queryItems: nil,
+            requiresConnectionId: false
+        )
+    }
+}
```
testableapple, line 55: Do we need something like this? If so, what should it look like?
We don't have `loadMutedUsers()` nor `loadFlaggedUsers()`.

nuno-vieira: This should be a new action in `CurrentUserController`, something like `getBlockedUsers()`.

Sources/StreamChat/Models/BlockedUser.swift
```diff
@@ -0,0 +1,36 @@
+//
+// Copyright © 2024 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+/// A type representing a blocked user. `BlockedUser` is an immutable snapshot of a blocked user entity at the given time.
+///
+public class BlockedUser {
```
nuno-vieira, line 9: It is better that our models are value types 
```suggestion
public struct BlockedUser {
```

Sources/StreamChat/APIClient/Endpoints/ModerationEndpoints.swift
```diff
@@ -16,6 +16,44 @@ extension Endpoint {
     }
 }
 
+// MARK: - User blocking
+
+extension Endpoint {
+    static func blockUser(_ userId: UserId, keepChannelsVisible: Bool = false) -> Endpoint<BlockedUsersResponse> {
```
nuno-vieira, line 22: Preferrable, we only set default values in the controller level. Not on the endpoints or updaters. This is to make sure that we don't forget to fill in all the arguments. Only the in the controller level we should provide default arguments to simplify the customers API 👍 
```suggestion
    static func blockUser(_ userId: UserId, keepChannelsVisible: Bool) -> Endpoint<BlockedUsersResponse> {
```

Sources/StreamChat/APIClient/Endpoints/Payloads/CurrentUserPayloads.swift
```diff
@@ -67,6 +71,7 @@ class CurrentUserPayload: UserPayload {
         devices = try container.decodeIfPresent([DevicePayload].self, forKey: .devices) ?? []
         mutedUsers = try container.decodeIfPresent([MutedUserPayload].self, forKey: .mutedUsers) ?? []
         mutedChannels = try container.decodeIfPresent([MutedChannelPayload].self, forKey: .mutedChannels) ?? []
+        blockedUsers = try container.decodeIfPresent([BlockedUserPayload].self, forKey: .blockedUsers) ?? []
```
nuno-vieira, line 74: According to the spec, we will also need to include the blocked users in the `Member` models

Sources/StreamChat/Models/BlockedUser.swift
```diff
@@ -0,0 +1,36 @@
...
+        hasher.combine(blockedUserId)
+    }
+}
+
+extension BlockedUser: Equatable {
+    public static func == (lhs: BlockedUser, rhs: BlockedUser) -> Bool {
+        lhs.blockedUserId == rhs.blockedUserId
+            && lhs.blockedAt == rhs.blockedAt
+    }
+}
```
nuno-vieira, line 36: Why do we need to it be Equatable for now? 🤔

Sources/StreamChat/Models/BlockedUser.swift
testableapple: Updated 👍

Sources/StreamChat/APIClient/Endpoints/ModerationEndpoints.swift
testableapple: Updated 👍

Sources/StreamChat/Models/BlockedUser.swift
testableapple: We don't, removed 👍

Sources/StreamChat/APIClient/Endpoints/ModerationEndpoints.swift
testableapple: Created:
```swift
    /// Get all blocked users.
    ///
    /// - Parameter completion: Called when the API call is finished. Called with `Error` if the remote update fails.
    ///
    func getBlockedUsers(completion: ((Error?) -> Void)? = nil) {
        currentUserUpdater.apiClient.request(endpoint: .loadBlockedUsers()) {
            completion?($0.error)
        }
    }
```

Sources/StreamChat/Controllers/CurrentUserController/CurrentUserController.swift
```diff
@@ -252,6 +252,16 @@ public extension CurrentChatUserController {
             }
         }
     }
+    
+    /// Get all blocked users.
+    ///
+    /// - Parameter completion: Called when the API call is finished. Called with `Error` if the remote update fails.
+    ///
+    func getBlockedUsers(completion: ((Error?) -> Void)? = nil) {
```
nuno-vieira, line 260: We need to return the users, otherwise, the developer can't do nothing with this 😛

testableapple: Updated 👍

Sources/StreamChat/APIClient/Endpoints/Payloads/CurrentUserPayloads.swift
testableapple: Done 👍

Sources/StreamChat/APIClient/Endpoints/Payloads/CurrentUserPayloads.swift
```diff
@@ -105,3 +110,32 @@ struct MutedUsersResponse: Decodable {
...
 }
+
+/// A blocked users response.
+struct BlockedUsersResponse: Decodable {
+    private enum CodingKeys: String, CodingKey {
+        case blockedUsers = "blocked_users"
+    }
+
+    /// The blocked users.
+    public let blockedUsers: [BlockedUserPayload]
```
laevandus, line 121: Delete `public`?

Sources/StreamChat/APIClient/Endpoints/Payloads/UserPayloads.swift
```diff
@@ -26,6 +26,8 @@ enum UserPayloadsCodingKeys: String, CodingKey, CaseIterable {
     case devices
     case unreadCount = "unread_count"
     case language
+    case blockedUsers = "blocked_users"
```
laevandus, line 29: Unused?

Sources/StreamChat/Database/DTOs/CurrentUserDTO.swift
```diff
@@ -106,6 +109,19 @@ extension NSManagedObjectContext: CurrentUserDatabaseSession {
 
         return dto
     }
+    
+    func saveCurrentUserBlockedUsers(_ blockedUsers: [BlockedUserPayload]) throws -> [BlockedUserDTO] {
+        invalidateCurrentUserCache()
```
laevandus, line 114: Feels like it shouldn't be here since this is logout related

Sources/StreamChat/Models/BlockedUser.swift
```diff
@@ -0,0 +1,29 @@
...
+// Copyright © 2024 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+/// A type representing a blocked user. `BlockedUser` is an immutable snapshot of a blocked user entity at the given time.
+///
+public struct BlockedUser {
+    /// The unique identifier of the blocked user.
+    public let blockedUserId: UserId
```
laevandus, line 11: just `userId`?

Sources/StreamChat/Workers/CurrentUserUpdater.swift
```diff
@@ -161,6 +161,33 @@ class CurrentUserUpdater: Worker {
...
+    
+    /// Get all blocked users.
+    ///
+    /// - Parameter completion: Called when the API call is finished. Called with `Error` if the remote update fails.
+    ///
+    func getBlockedUsers(completion: @escaping (Result<[BlockedUser], Error>) -> Void) {
+        apiClient.request(endpoint: .loadBlockedUsers()) {
+            switch $0 {
+            case let .success(payload):
+                var blockedUsers: Set<BlockedUserDTO> = []
```
laevandus, line 173: Safer if this is `Set<BlockedUser>` then DTOs are only accessed in the write closure (and it is sure that correct thread accesses it).

CHANGELOG.md
```diff
@@ -3,6 +3,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 # Upcoming
 
+## StreamChat
+### ✅ Added
+- Add a user block feature) [#3223](https://github.com/GetStream/stream-chat-swift/pull/3223)
```
laevandus, line 8: Extra `)`

Sources/StreamChat/APIClient/Endpoints/Payloads/UserPayloads.swift
testableapple: It's used here: https://github.com/GetStream/stream-chat-swift/blob/feature/block/Sources/StreamChat/APIClient/Endpoints/Payloads/CurrentUserPayloads.swift#L74

Sources/StreamChat/Workers/CurrentUserUpdater.swift
testableapple: Interesting, but how will we update the DTO?:
```
session.currentUser?.blockedUsers = blockedUsers
```
Left side is `Set<BlockedUserDTO>` when the right side, in this case, is `Set<BlockedUser>` 🤔

Sources/StreamChat/APIClient/Endpoints/Payloads/CurrentUserPayloads.swift
testableapple: Updated 👍

Sources/StreamChat/Models/BlockedUser.swift
testableapple: Updated 👍

CHANGELOG.md
testableapple: Updated 👍

CHANGELOG.md
```diff
@@ -3,6 +3,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 # Upcoming
 
+## StreamChat
+### ✅ Added
+- Add a user block feature [#3223](https://github.com/GetStream/stream-chat-swift/pull/3223)
```
nuno-vieira, line 8:
```suggestion
- Add support for user blocking [#3223](https://github.com/GetStream/stream-chat-swift/pull/3223)
```

Sources/StreamChat/APIClient/Endpoints/Payloads/CurrentUserPayloads.swift
```diff
@@ -105,3 +110,32 @@ struct MutedUsersResponse: Decodable {
...
+    }
+
+    let blockedUserId: String
+    let blockedByUserId: String
+    let createdAt: Date
+}
+
+extension BlockedUserPayload: Equatable {
+    static func == (lhs: BlockedUserPayload, rhs: BlockedUserPayload) -> Bool {
+        lhs.blockedUserId == rhs.blockedUserId && lhs.createdAt == rhs.createdAt
```
nuno-vieira, line 139: Shouldn't we use the blockedByUserId as well here?

Sources/StreamChat/Models/BlockedUser.swift
```diff
@@ -0,0 +1,29 @@
+//
+// Copyright © 2024 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+/// A type representing a blocked user. `BlockedUser` is an immutable snapshot of a blocked user entity at the given time.
+///
```
nuno-vieira, line 8: This should be enough 👍 
```suggestion
/// A type representing a blocked user.
```

Sources/StreamChat/Database/DTOs/BlockedUserDTO.swift
```diff
@@ -0,0 +1,28 @@
...
+
+@objc(BlockedUserDTO)
+class BlockedUserDTO: NSManagedObject {
+    @NSManaged var blockedUserId: String
+    @NSManaged var blockedAt: DBDate?
+    
+    func asModel() throws -> BlockedUser {
+        .init(
+            userId: blockedUserId,
+            blockedAt: blockedAt?.bridgeDate ?? .init()
```
nuno-vieira, line 16: The blockedAt is nullable right?
```suggestion
            blockedAt: blockedAt?.bridgeDate
```

Sources/StreamChat/StateLayer/ConnectedUser.swift
```diff
@@ -127,6 +127,28 @@ public final class ConnectedUser {
         try await userUpdater.unmuteUser(userId)
     }
     
+    /// Blocks the user in all the channels.
+    ///
+    /// - Note: Messages from muted users are not delivered via push notifications.
```
nuno-vieira, line 132: Is this a typo?
```suggestion
    /// - Note: Messages from blocked users are not delivered via push notifications.
```

Sources/StreamChat/Workers/CurrentUserUpdater.swift
nuno-vieira, line 173: This needs to be model, not DTO
```suggestion
                var blockedUsers: Set<BlockedUser> = []
```

Sources/StreamChat/Workers/CurrentUserUpdater.swift
```diff
@@ -161,6 +161,33 @@ class CurrentUserUpdater: Worker {
...
+                self.database.write({ session in
+                    for blockedUser in payload.blockedUsers {
+                        blockedUsers.insert(blockedUser.asDTO(context: self.database.writableContext))
+                    }
+                    session.currentUser?.blockedUsers = blockedUsers
+                    session.currentUser?.user.blockedUserIds = blockedUsers.map { $0.blockedUserId }
```
nuno-vieira, line 179: You need to map the DTOs here to the models and set the property
```suggestion
                    session.currentUser?.user.blockedUserIds = blockedUsers.map { $0.blockedUserId }
                    blockedUsers = blockedUsers.compactMap { try? $0.asModel() }
```

Sources/StreamChat/Workers/CurrentUserUpdater.swift
```diff
@@ -161,6 +161,33 @@ class CurrentUserUpdater: Worker {
...
+                }, completion: {
+                    if let error = $0 {
+                        log.error("Failed to save blocked users to the database. Error: \(error)")
+                    }
+                    completion(.success(blockedUsers.compactMap { try? $0.asModel() }))
```
nuno-vieira, line 184: Then here, you just pass the property
```suggestion
                    completion(.success(blockedUsers))
```

Sources/StreamChat/Workers/UserUpdater.swift
```diff
@@ -27,6 +27,62 @@ class UserUpdater: Worker {
...
+                    var blockedUsers: Set<BlockedUserDTO> = []
+                    for blockedUser in payload.blockedUsers {
+                        blockedUsers.insert(blockedUser.asDTO(context: self.database.writableContext))
+                    }
+                    session.currentUser?.blockedUsers = blockedUsers
+                    session.currentUser?.user.blockedUserIds = blockedUsers.map { $0.blockedUserId }
+                }, completion: {
+                    if let error = $0 {
+                        log.error("Failed to save blocked user with id: <\(userId)> to the database. Error: \(error)")
+                    }
```
nuno-vieira, line 51: Ideally if the request fails, we should remove the user from the blocked users

Sources/StreamChat/Workers/UserUpdater.swift
```diff
@@ -27,6 +27,62 @@ class UserUpdater: Worker {
...
+                self.database.write({ session in
+                    let currentUserDTO = session.currentUser
+                    if let userDTO = currentUserDTO?.blockedUsers.first(where: { $0.blockedUserId == userId }) {
+                        currentUserDTO?.blockedUsers.remove(userDTO)
+                        currentUserDTO?.user.blockedUserIds.removeAll { $0 == userId }
+                    }
+                }, completion: {
+                    if let error = $0 {
+                        log.error("Failed to remove blocked user with id: <\(userId)> from the database. Error: \(error)")
+                    }
```
nuno-vieira, line 78: The same here, if the request fails, we need to add back the user to the blocked users

Sources/StreamChatUI/MessageActionsPopup/ChatMessageActionsVC.swift
```diff
@@ -131,18 +131,23 @@ open class ChatMessageActionsVC: _ViewController, ThemeProvider {
...
-            if channelConfig.mutesEnabled && !message.isSentByCurrentUser {
+            if channelConfig.mutesEnabled && !isSentByCurrentUser {
                 let isMuted = currentUser.mutedUsers.map(\.id).contains(message.author.id)
                 actions.append(isMuted ? unmuteActionItem() : muteActionItem())
             }
+            
+            if !isSentByCurrentUser {
+                let isBlocked = currentUser.blockedUsers.map(\.userId).contains(message.author.id)
+                actions.append(isBlocked ? unblockActionItem() : blockActionItem())
+            }
```
nuno-vieira, line 150: We should not introduce this by default because it will break the customer's UI, and some customers might not even want this feature. So we need to introduce a Components flag, like `isBlockingUsersEnabled`. Or we don't have put this in our SDK, and only in the DemoApp 🤔 Depends on the requirements

Sources/StreamChat/Workers/CurrentUserUpdater.swift
```diff
@@ -161,6 +161,33 @@ class CurrentUserUpdater: Worker {
             completion?($0.error)
         }
     }
+    
+    /// Get all blocked users.
+    ///
+    /// - Parameter completion: Called when the API call is finished. Called with `Error` if the remote update fails.
+    ///
+    func getBlockedUsers(completion: @escaping (Result<[BlockedUser], Error>) -> Void) {
```
nuno-vieira, line 169: I don't see this being user anywhere? 🤔 It is not used in the controller nor the new state layer. 

Also, it would be nice to add to the Demo App, a way to show the blocked users, like we have with "Show Members"

Sources/StreamChat/StateLayer/ConnectedUser.swift
nuno-vieira: Or should we delete this Note? Isn't this from muted users?

Sources/StreamChat/Workers/CurrentUserUpdater.swift
testableapple: I was able to bypass the compiler errors only by making it an Array 🤔
```swift
var blockedUsers: [BlockedUser] = []
```

Sources/StreamChat/Workers/UserUpdater.swift
testableapple: I added this logic, but I'm not sure I've chosen the best approach.

Sources/StreamChat/Workers/UserUpdater.swift
testableapple: Same as above.

Sources/StreamChatUI/MessageActionsPopup/ChatMessageActionsVC.swift
testableapple: Added `isBlockingUsersEnabled` 👍

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
```diff
@@ -224,6 +226,13 @@ extension NSManagedObjectContext {
         }
 
         dto.isFrozen = payload.isFrozen
+        
+
+        // Backend only returns a boolean for blocked 1:1 channels
+        // on channel list query
+        if let isBlocked = payload.isBlocked {
+            dto.isBlocked = isBlocked
+        }
```
laevandus, line 235: Thinking if we should have 
```
} else {
  dto.isBlocked = false
}
```
for cases where one has 1:1 channel with a blocked user but then adds more members.

Sources/StreamChat/Database/DTOs/CurrentUserDTO.swift
```diff
@@ -14,7 +14,8 @@ class CurrentUserDTO: NSManagedObject {
     /// Contains the timestamp when last sync process was finished.
     /// The date later serves as reference date for the last event synced using `/sync` endpoint
     @NSManaged var lastSynchedEventDate: DBDate?
-
+    
+    @NSManaged var blockedUserIds: [String]
```
laevandus, line 18: Should we do `Set<String>` here since the order is not important and Set avoids possible duplicates (although it should never happen)

Sources/StreamChat/Database/DTOs/CurrentUserDTO.swift
```diff
@@ -91,6 +92,8 @@ extension NSManagedObjectContext: CurrentUserDatabaseSession {
 
         let mutedUsers = try payload.mutedUsers.map { try saveUser(payload: $0.mutedUser) }
         dto.mutedUsers = Set(mutedUsers)
+        
+        dto.blockedUserIds = currentUser?.blockedUserIds ?? []
```
laevandus, line 96: `payload.blockedUserIds`?

Sources/StreamChat/Database/StreamChatModel.xcdatamodeld/StreamChatModel.xcdatamodel/contents
```diff
@@ -133,6 +134,7 @@
         <relationship name="channelConfig" optional="YES" maxCount="1" deletionRule="Nullify" destinationEntity="ChannelConfigDTO" inverseName="commands" inverseEntity="ChannelConfigDTO"/>
     </entity>
     <entity name="CurrentUserDTO" representedClassName="CurrentUserDTO" syncable="YES">
+        <attribute name="blockedUserIds" optional="YES" attributeType="Transformable"/>
```
laevandus, line 137: Out other transformable types have value transformer set to `NSSecureUnarchiveFromData`

Sources/StreamChat/Models/Channel.swift
```diff
@@ -49,6 +49,12 @@ public struct ChatChannel {
     /// It's not possible to send new messages to a frozen channel.
     ///
     public let isFrozen: Bool
+    
+    /// Returns `true` if the channel is blocked.
+    ///
+    /// It's not possible to send new messages to a blocked channel.
+    ///
+    public let isBlocked: Bool?
```
laevandus, line 57: Optional bool is so weird, should we go for Bool and false as default? Another option is enum but that sounds like too much.

Sources/StreamChat/Models/BlockedUser.swift
```diff
@@ -0,0 +1,28 @@
+//
+// Copyright © 2024 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+/// A type representing a blocked user.
+public struct BlockedUser {
```
laevandus, line 8: Since we have `ChatUser` and `CurrentUser` with full user information available in these types then the `BlockedUser` is a bit different since it only includes id and blocked date. Should we rename it to `BlockedUserInfo` instead? @nuno-vieira

Sources/StreamChat/Workers/CurrentUserUpdater.swift
testableapple: Updated 👍

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
testableapple: Sounds reasonable. Updated 👍

Sources/StreamChat/Database/DTOs/CurrentUserDTO.swift
testableapple: Updated 👍

Sources/StreamChat/Database/DTOs/CurrentUserDTO.swift
testableapple: Yeah, also noticed this as well. Already fixed 🤠

Sources/StreamChat/Models/Channel.swift
testableapple: Made it `false` by default 👍

Sources/StreamChat/APIClient/Endpoints/BlockedUserPayload.swift
```diff
@@ -0,0 +1,97 @@
...
+    let teams: [TeamId]
+    let isOnline: Bool
+}
+
+extension BlockedUserPayload: Equatable {
+    static func == (lhs: BlockedUserPayload, rhs: BlockedUserPayload) -> Bool {
+        lhs.id == rhs.id &&
+        lhs.createdAt == rhs.createdAt
+    }
+}
```
laevandus, line 97: Do we needed it and if yes, just wondering why only `id` and `createdAt`?

Sources/StreamChat/Controllers/CurrentUserController/CurrentUserController.swift
```diff
@@ -252,6 +252,25 @@ public extension CurrentChatUserController {
             }
         }
     }
+    
+    /// Get all blocked users.
+    ///
+    /// - Parameter completion: Called when the API call is finished. Called with `Error` if the remote update fails.
+    ///
+    func getBlockedUsers(completion: @escaping (Result<[BlockedUser], Error>) -> Void) {
```
laevandus, line 260: In other places we use `loadXYZ`, what if we use the same term here as well for consistency?

nuno-vieira: I think `get` here kinda makes sense, because we are just getting the blocked users, and not setting it anywhere 🤔 When we use `load` we usually assign the data into a property of the controller. But maybe for consistency, it is better we just name `load` 🤔

CHANGELOG.md
```diff
@@ -10,6 +10,9 @@ _June 06, 2024_
 
 ## StreamChat
 ### ✅ Added
+- Add support for user blocking [#3223](https://github.com/GetStream/stream-chat-swift/pull/3223)
+
+### 🔄 Changed
```
nuno-vieira, line 15: This is incorrectly added in 4.57.0, instead of upcoming

DemoApp/StreamChat/Components/DemoChatChannelListRouter.swift
```diff
@@ -447,6 +447,22 @@ final class DemoChatChannelListRouter: ChatChannelListRouter {
...
+                    if let blockedUsers = try? result.get() {
+                        self.rootViewController.present(MembersViewController(
+                            membersController: client.memberListController(
+                                query: .init(
+                                    cid: cid,
+                                    filter: .in(.id, values: blockedUsers.map({ $0.userId }))
+                                )
+                            )
+                        ), animated: true)
+                    }
```
nuno-vieira, line 463: Nit: a little be easier to read if we use guard here. Otherwise, it is too much nested blocks
```suggestion
                    guard let blockedUsers = try? result.get() else { return }
                    self.rootViewController.present(MembersViewController(
                        membersController: client.memberListController(
                            query: .init(
                                cid: cid,
                                filter: .in(.id, values: blockedUsers.map({ $0.userId }))
                            )
                        )
                    ), animated: true)
```

Sources/StreamChat/APIClient/Endpoints/BlockedUserPayload.swift
```diff
@@ -0,0 +1,97 @@
...
+        case isOnline = "online"
+    }
+
+    let id: String
+    let name: String?
+    let createdAt: Date
+    let updatedAt: Date
+    let isBanned: Bool
+    let lastActiveAt: Date?
+    let privacySettings: UserPrivacySettingsPayload?
```
nuno-vieira, line 84: Are we sure this is correct? We should not have access to privacy settings from other users 🤔

Sources/StreamChat/APIClient/Endpoints/BlockedUserPayload.swift
nuno-vieira: Yes same question 🤔

Sources/StreamChat/Models/BlockedUser.swift
nuno-vieira: Maybe `BlockedUserDetails` ? We also have `MessagePinDetails` and `MessageModerationDetails`. So maybe `Details` postfix has more consistancy, WDYT?

Sources/StreamChat/Models/UserInfo.swift
```diff
@@ -18,6 +18,8 @@ public struct UserInfo {
     public let language: TranslationLanguage?
     /// The privacy settings of the user. Example: If the user does not want to expose typing events or read events.
     public let privacySettings: UserPrivacySettings?
+    /// The privacy settings of the user. Example: If the user does not want to expose typing events or read events.
+    public let blockedUserIds: [UserId]?
```
nuno-vieira, line 22: Copy pasted comment 😛

nuno-vieira (commented):
Looking great 👌 Added a couple of comments only 👍

Sources/StreamChat/Workers/UserUpdater.swift
```diff
@@ -27,6 +27,60 @@ class UserUpdater: Worker {
...
+                        
+                        self.database.write({ session in
+                            session.currentUser?.blockedUserIds.remove(userId)
+                        })
+                    }
+                    completion?($0)
+                })
+            case let .failure(error):
+                completion?(error)
+            }
```
nuno-vieira, line 54: The remove of the userId, should be done in the `failure` case, which is where the http request fails. Currently is being done when the DB write fails.

Sources/StreamChat/Workers/UserUpdater.swift
```diff
@@ -27,6 +27,60 @@ class UserUpdater: Worker {
...
+                        
+                        self.database.write({ session in
+                            session.currentUser?.blockedUserIds.insert(userId)
+                        })
+                    }
+                    completion?($0)
+                })
+            case let .failure(error):
+                completion?(error)
+            }
```
nuno-vieira, line 81: Same here

CHANGELOG.md
testableapple: That's weird. I definitely did move it up to the upcoming one. Fixed again now.

Sources/StreamChat/Models/UserInfo.swift
testableapple: Fixed 🤠

Sources/StreamChat/APIClient/Endpoints/BlockedUserPayload.swift
testableapple: Can't say I fully understand when we need to implement Equatable, so just added it just in case. Removed now 👍

Sources/StreamChat/APIClient/Endpoints/BlockedUserPayload.swift
testableapple: Ugh, API is now different. Updated.

Sources/StreamChat/Workers/UserUpdater.swift
testableapple: Updated 👍

Sources/StreamChat/Workers/UserUpdater.swift
testableapple: Updated 👍

Sources/StreamChat/Controllers/CurrentUserController/CurrentUserController.swift
testableapple: Updated to `load` 👍

CHANGELOG.md
```diff
@@ -3,13 +3,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
-### 🔄 Changed
+## StreamChat
+### ✅ Added
+- Add support for user blocking [#3223](https://github.com/GetStream/stream-chat-swift/pull/3223)
 
 # [4.57.0](https://github.com/GetStream/stream-chat-swift/releases/tag/4.57.0)
 _June 06, 2024_
 
-## StreamChat
-### ✅ Added
```
nuno-vieira, line 12: ⚠️ We are removing this and adding "Changed". Please revert this 👍

testableapple: Oops, updated 👍

nuno-vieira (approved):
LGTM! ✅

Sources/StreamChat/Workers/CurrentUserUpdater.swift
```diff
@@ -202,6 +227,14 @@ extension CurrentUserUpdater {
         }
     }
     
+    func loadBlockedUsers() async throws -> [BlockedUserDetails] {
+        try await withCheckedThrowingContinuation { continuation in
+            loadBlockedUsers { error in
```
laevandus, line 232: Naming: it is actually `result`, not `error` here.

laevandus (approved):

Sources/StreamChat/Workers/CurrentUserUpdater.swift
testableapple: Updated 👍

Tests/StreamChatTests/Controllers/SearchControllers/UserController/UserController_Tests.swift
```diff
@@ -491,28 +491,28 @@ final class UserController_Tests: XCTestCase {
         AssertAsync.staysTrue(weakController != nil)
     }
 
-    // MARK: - Unlag user
+    // MARK: - Block user
 
-    func test_unflagUser_propagatesError() {
```
laevandus, line 496: Did you delete flagging related tests?

laevandus (changes requested):
The icons looks with lighter color and the show blocked users debug command reported JSON parsing issue

docusaurus/docs/iOS/common-content/reference-docs/stream-chat/api-client/endpoints/payloads/channel-config.md
```diff
@@ -133,11 +133,3 @@ A channel updated date.
 ``` swift
 public let updatedAt: Date
 ```
-
-### `flagsEnabled`
-
-Determines if users are able to flag messages. Enabled by default.
-
-``` swift
-public var flagsEnabled: Bool
```
laevandus, line 142: Should it be removed?

testableapple: So weird. Returned 👍

laevandus (approved):

laevandus:
I tried QA verification steps and these worked correctly.

Sources/StreamChat/Workers/UserUpdater.swift
```diff
@@ -27,6 +28,70 @@ class UserUpdater: Worker {
...
+            switch $0 {
+            case .success:
+                self.database.write({ session in
+                    session.currentUser?.blockedUserIds.insert(userId)
+                    if let channel = ChannelDTO.directMessageChannel(
+                        participantId: userId,
+                        context: self.database.writableContext
+                    ) {
+                        try session.channel(cid: .init(cid: channel.cid))?.isHidden = true
+                    }
```
nuno-vieira, line 48: This should be enough: 🤔 
```suggestion
                    if let channel = ChannelDTO.directMessageChannel(
                        participantId: userId,
                        context: self.database.writableContext
                    ) {
                       channel.isHidden = true
                    }
```

Sources/StreamChat/Workers/UserUpdater.swift
```diff
@@ -27,6 +28,70 @@ class UserUpdater: Worker {
...
+            switch $0 {
+            case .success:
+                self.database.write({ session in
+                    session.currentUser?.blockedUserIds.remove(userId)
+                    if let channel = ChannelDTO.directMessageChannel(
+                        participantId: userId,
+                        context: self.database.writableContext
+                    ) {
+                        try session.channel(cid: .init(cid: channel.cid))?.isHidden = false
+                    }
```
nuno-vieira, line 80: same here

Sources/StreamChat/Workers/UserUpdater.swift
testableapple: Nice! Updated 🤠

nuno-vieira (approved):
LGTM! ✅

Sources/StreamChat/APIClient/Endpoints/BlockedUserPayload.swift
```diff
@@ -0,0 +1,78 @@
...
+    let blockedByUserId: String
+    let createdAt: Date
+}
+
+extension BlockingUserPayload: Equatable {
+    static func == (lhs: BlockingUserPayload, rhs: BlockingUserPayload) -> Bool {
+        lhs.blockedUserId == rhs.blockedUserId
+            && lhs.blockedByUserId == rhs.blockedByUserId
+            && lhs.createdAt == rhs.createdAt
+    }
```
laevandus, line 25: That is not needed since Swift will synthesise this automatically. We sometimes have custom equality implementations when we want to exclude some properties from comparison.

Sources/StreamChat/APIClient/Endpoints/BlockedUserPayload.swift
```diff
@@ -0,0 +1,78 @@
...
+    let createdAt: Date
+    let blockedUser: BlockedUserPayload
+}
+
+extension BlockPayload: Equatable {
+    static func == (lhs: BlockPayload, rhs: BlockPayload) -> Bool {
+        lhs.blockedUserId == rhs.blockedUserId
+            && lhs.userId == rhs.userId
+            && lhs.createdAt == rhs.createdAt
+    }
```
laevandus, line 58: Not sure why we exclude blockedUser here. Should we just delete this add add `Equatable` to line 39 and let Swift to compare all the properties automatically.

Sources/StreamChat/Controllers/CurrentUserController/CurrentUserController.swift
```diff
@@ -257,6 +257,25 @@ public extension CurrentChatUserController {
...
+            switch $0 {
+            case let .success(blockedUsers):
+                self.callback {
+                    completion(.success(blockedUsers))
+                }
+            case let .failure(error):
+                self.callback {
+                    completion(.failure(error))
+                }
+            }
```
laevandus, line 276: Seems like this could be just
```
self.callback {
completion($0)
}
```
since Result types match

Sources/StreamChat/APIClient/Endpoints/BlockedUserPayload.swift
testableapple: To line `31`? It throws: `Type 'BlockPayload' does not conform to protocol 'Equatable'`

Sources/StreamChat/APIClient/Endpoints/BlockedUserPayload.swift
testableapple: Done 👍

Sources/StreamChat/APIClient/Endpoints/BlockedUserPayload.swift
laevandus: Ah I see, we skipped BlockedUserPayload. Let's keep it was what you have

Sources/StreamChat/APIClient/Endpoints/BlockedUserPayload.swift
testableapple: Done 👍

Sources/StreamChat/Controllers/CurrentUserController/CurrentUserController.swift
testableapple: Done 👍

laevandus:
Verified! Worked as expected

PR #3226, opened 2024-05-31
Use cursor when fetching more votes

laevandus:
### 🎯 Goal

Vote list endpoint only supports cursors.

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/Controllers/PollController/PollVoteListController.swift
+++ b/Sources/StreamChat/Controllers/PollController/PollVoteListController.swift
@@ -45,7 +45,7 @@ public class PollVoteListController: DataController, DelegateCallable, DataStore
         return pollVotesObserver.items
     }
     
-    public private(set) var hasLoadedAllVotes: Bool = false
+    public var hasLoadedAllVotes: Bool { votesCursor == nil }
 
     /// Set the delegate of `PollVoteListController` to observe the changes in the system.
     public weak var delegate: PollVoteListControllerDelegate? {
@@ -105,6 +105,7 @@ public class PollVoteListController: DataController, DelegateCallable, DataStore
     private let eventsController: EventsController
     private let pollsRepository: PollsRepository
     private let environment: Environment
+    private var votesCursor: String?
 
     /// Creates a new `PollVoteListController`.
     ///
@@ -126,6 +127,7 @@ public class PollVoteListController: DataController, DelegateCallable, DataStore
 
         pollsRepository.queryPollVotes(query: query) { [weak self] result in
             guard let self else { return }
+            self.votesCursor = result.value?.next
             if let error = result.error {
                 self.state = .remoteDataFetchFailed(ClientError(with: error))
             } else {
@@ -164,10 +166,10 @@ public extension PollVoteListController {
     ) {
         let limit = limit ?? query.pagination.pageSize
         var updatedQuery = query
-        updatedQuery.pagination = Pagination(pageSize: limit, offset: votes.count)
+        updatedQuery.pagination = Pagination(pageSize: limit, cursor: votesCursor)
         pollsRepository.queryPollVotes(query: updatedQuery) { [weak self] result in
             guard let self else { return }
-            self.hasLoadedAllVotes = (result.value?.count ?? 0) < limit
+            self.votesCursor = result.value?.next
             self.callback { completion?(result.error) }
         }
     }
--- a/Sources/StreamChat/Controllers/PollController/PollVoteListController.swift
+++ b/Sources/StreamChat/Controllers/PollController/PollVoteListController.swift
@@ -45,7 +45,8 @@ public class PollVoteListController: DataController, DelegateCallable, DataStore
         return pollVotesObserver.items
     }
     
-    public var hasLoadedAllVotes: Bool { votesCursor == nil }
+    /// A Boolean value that returns whether pagination is finished.
+    public private(set) var hasLoadedAllVotes: Bool = false
 
     /// Set the delegate of `PollVoteListController` to observe the changes in the system.
     public weak var delegate: PollVoteListControllerDelegate? {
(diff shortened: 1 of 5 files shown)


Sources/StreamChat/Repositories/PollsRepository.swift
```diff
@@ -288,6 +288,11 @@ class PollsRepository {
             )
         }
     }
+    
+    struct VotePaginationResponse {
```
martinmitrevski, line 292: Why not use `PollVoteListResponse`?

Sources/StreamChat/Controllers/PollController/PollVoteListController.swift
```diff
@@ -45,7 +45,7 @@ public class PollVoteListController: DataController, DelegateCallable, DataStore
         return pollVotesObserver.items
     }
     
-    public private(set) var hasLoadedAllVotes: Bool = false
+    public var hasLoadedAllVotes: Bool { votesCursor == nil }
```
nuno-vieira, line 48: I think we should not do this.  I also had this bug on the Thread v2. Initially, the cursor will always be `nil`. So, until we fetch the first page, this will be misleading and say that we loaded all votes .
```suggestion
    public var hasLoadedAllVotes: Bool = false
```

Sources/StreamChat/Controllers/PollController/PollVoteListController.swift
```diff
@@ -126,6 +127,7 @@ public class PollVoteListController: DataController, DelegateCallable, DataStore
 
         pollsRepository.queryPollVotes(query: query) { [weak self] result in
             guard let self else { return }
+            self.votesCursor = result.value?.next
```
nuno-vieira, line 130: This is also why the `hasLoadedAllVotes: Bool { votesCursor == nil }` is a bad idea. If the response is an error, there will be no cursor, but it will say we loaded all votes already

Sources/StreamChat/Repositories/PollsRepository.swift
nuno-vieira: Yes seems like we could just reuse `PollVoteListResponse` 🤔

nuno-vieira (changes requested):
If it only works with cursor, how was it working before? Or we did not test the pagination before? 😬 

I'm requesting changes because I believe the implementation is not correct, we need to make this happens:
1. hasLoadedAllVotes is set only when the response is succesful and the cursor is nil
2. The cursor should not be set when the response fails
3. For the reasons above, hasLoadedAllVotes can't be computed property

Sources/StreamChat/Repositories/PollsRepository.swift
laevandus: I added a separate one since `PollVoteListResponse` is API payload and this returns what DB has after the whole conversion cycle.

martinmitrevski:
> If it only works with cursor, how was it working before? Or we did not test the pagination before? 😬
> 
> I'm requesting changes because I believe the implementation is not correct, we need to make sure this happens:
> 
> 1. hasLoadedAllVotes is set only when the response is succesful and the cursor is nil
> 2. The cursor should not be set when the response fails
> 3. For the reasons above, hasLoadedAllVotes can't be computed property

@nuno-vieira it didn't work only with a specific filter for answers, otherwise it worked for regular votes querying.

Sources/StreamChat/Controllers/PollController/PollVoteListController.swift
```diff
@@ -126,6 +128,10 @@ public class PollVoteListController: DataController, DelegateCallable, DataStore
 
         pollsRepository.queryPollVotes(query: query) { [weak self] result in
             guard let self else { return }
+            if let value = result.value {
+                self.nextCursor = value.next
+                self.hasLoadedAllVotes = value.votes.count < query.pagination.pageSize
```
nuno-vieira, line 133: I actually had a bug on Threads v2 where because the current user was removed from those threads, the page did not return the `limit` that it should have. So it is actually safer we check the `next` value to determine if there are more data or not. (This is a bug on the backend, but better to be safer)
```suggestion
                self.hasLoadedAllVotes = value.next == nil
```

Sources/StreamChat/Controllers/PollController/PollVoteListController.swift
```diff
@@ -164,10 +170,13 @@ public extension PollVoteListController {
...
         let limit = limit ?? query.pagination.pageSize
         var updatedQuery = query
-        updatedQuery.pagination = Pagination(pageSize: limit, offset: votes.count)
+        updatedQuery.pagination = Pagination(pageSize: limit, cursor: nextCursor)
         pollsRepository.queryPollVotes(query: updatedQuery) { [weak self] result in
             guard let self else { return }
-            self.hasLoadedAllVotes = (result.value?.count ?? 0) < limit
+            if let value = result.value {
+                self.nextCursor = value.next
+                self.hasLoadedAllVotes = value.votes.count < limit
```
nuno-vieira, line 178: Same here:
```suggestion
                self.hasLoadedAllVotes = value.next == nil
```

Sources/StreamChat/Repositories/PollsRepository.swift
nuno-vieira: Ah I see 👍

Sources/StreamChat/Controllers/PollController/PollVoteListController.swift
laevandus: Done

Sources/StreamChat/Controllers/PollController/PollVoteListController.swift
laevandus: Thank you

martinmitrevski (approved):

nuno-vieira (approved):
LGTM! ✅

PR #3245, opened 2024-06-12
Reset channel members and watchers states when fetching the initial state

laevandus:
### 🔗 Issue Links

Resolves [ios-issues-tracking/issues/866](https://github.com/GetStream/ios-issues-tracking/issues/866)

### 🎯 Goal

* Reset members and watches when resetting the channel data to the initial state

### 📝 Summary

* When we run synchronize then we should reset member data since members might have been removed
* It was fixed for chat state layer and state layer has tests for this already.

### 🧪 Manual Testing Notes

- Log in with Luke on device 1
- Log in with Han on device 2
- Luke creates a channel with Han
- Han opens the channel and sees that they are both part of the channel
- Han quits the demo app on device 2
- Luke leaves the channel
- Han launches the app again and navigates to the channel

What happens: Luke is shown as a member
Expected: Luke is not shown as a member for Han 

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/Workers/ChannelUpdater.swift
+++ b/Sources/StreamChat/Workers/ChannelUpdater.swift
@@ -56,6 +56,9 @@ class ChannelUpdater: Worker {
         
         let didLoadFirstPage = channelQuery.pagination?.parameter == nil
         let didJumpToMessage: Bool = channelQuery.pagination?.parameter?.isJumpingToMessage == true
+        let resetMembers = didLoadFirstPage
+        let resetMessages = didLoadFirstPage || didJumpToMessage
+        let resetWatchers = didLoadFirstPage
         let isChannelCreate = onChannelCreated != nil
 
         let completion: (Result<ChannelPayload, Error>) -> Void = { [weak database] result in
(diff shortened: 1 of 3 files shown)


Sources/StreamChat/Workers/ChannelUpdater.swift
```diff
@@ -918,19 +921,12 @@ extension ChannelUpdater {
     ) async throws -> ChannelPayload {
         // Just populate the closure since we select the endpoint based on it.
         let useCreateEndpoint: ((ChannelId) -> Void)? = channelQuery.cid == nil ? { _ in } : nil
-        let parameter = channelQuery.pagination?.parameter
-        let reset = parameter == nil || parameter?.isJumpingToMessage == true
-        let actions = ChannelUpdateActions(
-            memberListSorting: memberSorting,
-            resetMembers: reset,
-            resetWatchers: reset
-        )
```
laevandus, line 927: Moved the fix I used for state layer to the completion handler based update method instead. State layer already has tests for this case.

nuno-vieira (approved):

PR #3250, opened 2024-06-17
Turn on background mapping

laevandus:
### 🔗 Issue Links

Resolves [ios-issues-tracking/issues/863](https://github.com/GetStream/ios-issues-tracking/issues/863)

### 🎯 Goal

Set background mapping default to true.

### ☑️ Contributor Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Fix notifications muted state for the current user in channel members [#3236](https://github.com/GetStream/stream-chat-swift/pull/3236)
 - Reset channel members and watchers state when fetching the initial state of the channel [#3245](https://github.com/GetStream/stream-chat-swift/pull/3245)
 - Fix inconsistent message text when extremely quickly updating it [#3242](https://github.com/GetStream/stream-chat-swift/pull/3242)
+### 🔄 Changed
+- Set background mapping enabled (prepare data on a separate queue when database changes) [#3250](https://github.com/GetStream/stream-chat-swift/pull/3250)
 
 ## StreamChatUI
 ### 🐞 Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Reset channel members and watchers state when fetching the initial state of the channel [#3245](https://github.com/GetStream/stream-chat-swift/pull/3245)
 - Fix inconsistent message text when extremely quickly updating it [#3242](https://github.com/GetStream/stream-chat-swift/pull/3242)
 ### 🔄 Changed
-- Set background mapping enabled (prepare data on a separate queue when database changes) [#3250](https://github.com/GetStream/stream-chat-swift/pull/3250)
+- Enable background mapping by default, which improves performance overall [#3250](https://github.com/GetStream/stream-chat-swift/pull/3250)
 
 ## StreamChatUI
 ### 🐞 Fixed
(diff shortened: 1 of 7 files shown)


martinmitrevski (approved):

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Fix notifications muted state for the current user in channel members [#3236](https://github.com/GetStream/stream-chat-swift/pull/3236)
 - Reset channel members and watchers state when fetching the initial state of the channel [#3245](https://github.com/GetStream/stream-chat-swift/pull/3245)
 - Fix inconsistent message text when extremely quickly updating it [#3242](https://github.com/GetStream/stream-chat-swift/pull/3242)
+### 🔄 Changed
+- Set background mapping enabled (prepare data on a separate queue when database changes) [#3250](https://github.com/GetStream/stream-chat-swift/pull/3250)
```
nuno-vieira, line 13: We should try to avoid adding implementation details in the Changelog, it is better just to describe the end goal:
```suggestion
- Enable background mapping by default, which improves performance overall [#3250](https://github.com/GetStream/stream-chat-swift/pull/3250)
```

Sources/StreamChat/Config/StreamRuntimeCheck.swift
```diff
@@ -13,7 +13,7 @@ public enum StreamRuntimeCheck {
     /// For *internal use* only
     ///
     ///  Enables background mapping of DB models
-    public static var _isBackgroundMappingEnabled = false
+    public static var _isBackgroundMappingEnabled = true
```
nuno-vieira, line 16: Now we can remove the `_` since this won't be a "private" feature anymore:
```suggestion
    public static var isBackgroundMappingEnabled = true
```

martinmitrevski: Would be a breaking change, also let's have the flexibility to turn it off if something goes wrong. Therefore, would prefer if we keep it.

nuno-vieira: That is why `_` exists. It is saying that this is for now a private API, and once it is public, it ofc will break

Sources/StreamChat/Config/StreamRuntimeCheck.swift
```diff
@@ -13,7 +13,7 @@ public enum StreamRuntimeCheck {
     /// For *internal use* only
```
nuno-vieira, line 13: We need to remove this as well, since it is not internal anymore
```suggestion
```

Sources/StreamChat/Config/StreamRuntimeCheck.swift
martinmitrevski: But what changed for it to be "public" API? It was always public, we are just turning this feature flag on by default.

nuno-vieira: It was not public. It was internal, and only some customers were test-driving this. That is why we have this:
` For *internal use* only`.

The `underscore` is also a way to comunicate that it is a "internal" property and a breaking change can happen. This pattern is normally used across APIs and SDKs, even Apple uses this as well.

nuno-vieira: Otherwise, what is the point of having the underscore in the first place?

martinmitrevski: I think we have a misunderstanding about this one - we said that we **will turn it on by default**, not that we will make it public API so customers can change this as they see fit. That's what the PR title also says.

laevandus: I expect that we delete this key in coming months if it is stable for everyone. Renaming the key would trigger build errors for ones who were test-driving it first.

laevandus:
Ran unit-tests and UI-tests locally.

PR #3257, opened 2024-06-18
[Demo app] Fallback to jpeg path extension if the downloaded file does not specify it

laevandus:
### 🔗 Issue Links

Related [ios-issues-tracking/issues/872](https://github.com/GetStream/ios-issues-tracking/issues/872)

### 🎯 Goal

When notification extension downloads an image without path extension, UNNotificationAttachment creation fails. Just fallback to jpeg if not set.

Example is google maps and static links to maps.

### ☑️ Contributor Checklist

### DIFF
--- a/DemoAppPush/NotificationService.swift
+++ b/DemoAppPush/NotificationService.swift
@@ -28,7 +28,11 @@ class NotificationService: UNNotificationServiceExtension {
                 return
             }
 
-            let localURL = URL(fileURLWithPath: path).appendingPathComponent(url.lastPathComponent)
+            // UNNotificationAttachment requires path extension to be set (fallback to jpeg if not set)
+            var localURL = URL(fileURLWithPath: path).appendingPathComponent(UUID().uuidString + url.lastPathComponent)
+            if localURL.pathExtension.isEmpty {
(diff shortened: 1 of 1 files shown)


DemoAppPush/NotificationService.swift
```diff
@@ -28,7 +28,11 @@ class NotificationService: UNNotificationServiceExtension {
                 return
             }
 
-            let localURL = URL(fileURLWithPath: path).appendingPathComponent(url.lastPathComponent)
+            // UNNotificationAttachment requires path extension to be set (fallback to jpeg if not set)
+            var localURL = URL(fileURLWithPath: path).appendingPathComponent(UUID().uuidString + url.lastPathComponent)
+            if localURL.pathExtension.isEmpty {
+                localURL.appendPathExtension("jpeg")
+            }
```
laevandus, line 35: Simplest fix to the missing path extension triggering UNNotificationAttachment failure. Not sure if we should put more effort into it. I tried `resourceValues` as well for downloaded URL but it fails in the extension.

testableapple (approved):
LGTM 👍

PR #3260, opened 2024-06-20
Significantly improve performance of database observers

laevandus:
### 🔗 Issue Links

Related [ios-issues-tracking/issues/870](https://github.com/GetStream/ios-issues-tracking/issues/870)

### 🎯 Goal

* Improved performance of database observers by reusing unchanged items

### 📝 Summary

* Reuse existing model instances what did not change when NSFetchedResultsController reports list changes
* Reuse newly created items stored in `ListChange` when did change content callback is triggered
* Affects `BackgroundDatabaseObserver` and `StateLayerDatabaseObserver` (list type)

### 🛠 Implementation

`DTO` and immutable model types are matched using a single `id` value. Since `ListChange.item` contains a newly converted item, instead of recreating it again, we reuse the value.

### 🎨 Showcase

When comparing how much CPU time does the `BackgroundDatabaseObserver.updateItems` takes when opening the demo app, navigating to a channel, loading several pages of messages, then we can see significant improvement.

| Before | After |
| ------ | ----- |
|  ~1100ms  |  ~200ms  |

result

### 🧪 Manual Testing Notes

Since we reuse existing items and do not convert the whole list, we need to be sure, that NSFetchedResultsController does not miss any data changes in more complex relationships. Therefore, we need to do a manual testing round covering many use-cases.

### ☑️ Contributor Checklist

### 🎁 Meme

gif

### DIFF
--- a/Sources/StreamChat/Config/StreamRuntimeCheck.swift
+++ b/Sources/StreamChat/Config/StreamRuntimeCheck.swift
@@ -39,6 +39,8 @@ public enum StreamRuntimeCheck {
     ///  Enables using our legacy web socket connection.
     public static var _useLegacyWebSocketConnection = false
     
+    /// For *internal use* only
+    ///
     /// Enables reusing unchanged converted items in database observers.
-    public static var isDatabaseObserverItemReusingEnabled = true
+    public static var _isDatabaseObserverItemReusingEnabled = true
 }
--- a/Sources/StreamChat/Config/StreamRuntimeCheck.swift
+++ b/Sources/StreamChat/Config/StreamRuntimeCheck.swift
@@ -38,4 +38,7 @@ public enum StreamRuntimeCheck {
     ///
     ///  Enables using our legacy web socket connection.
     public static var _useLegacyWebSocketConnection = false
+    
+    /// Enables reusing unchanged converted items in database observers.
+    public static var isDatabaseObserverItemReusingEnabled = true
 }
--- a/Sources/StreamChat/Controllers/DatabaseObserver/BackgroundDatabaseObserver.swift
+++ b/Sources/StreamChat/Controllers/DatabaseObserver/BackgroundDatabaseObserver.swift
@@ -16,6 +16,7 @@ class BackgroundDatabaseObserver<Item, DTO: NSManagedObject> {
 
     /// Used to convert the `DTO`s to the resulting `Item`s.
     private let itemCreator: (DTO) throws -> Item
+    private let itemReuseKeyPaths: (item: KeyPath<Item, String>, dto: KeyPath<DTO, String>)?
     private let sorting: [SortValue<Item>]
 
     /// Used to observe the changes in the DB.
@@ -64,6 +65,7 @@ class BackgroundDatabaseObserver<Item, DTO: NSManagedObject> {
     ///   - fetchRequest: The `NSFetchRequest` that specifies the elements of the list.
     ///   - context: The `NSManagedObjectContext` the observer observes.
     ///   - itemCreator: A closure the observer uses to convert DTO objects into Model objects.
+    ///   - itemReuseKeyPaths: A pair of keypaths used for reusing existing items if they have not changed
     ///   - sorting: An array of SortValue that define the order of the elements in the list.
     ///   - fetchedResultsControllerType: The `NSFetchedResultsController` subclass the observer uses to create its FRC. You can
     ///    inject your custom subclass if needed, i.e. when testing.
@@ -71,10 +73,12 @@ class BackgroundDatabaseObserver<Item, DTO: NSManagedObject> {
         context: NSManagedObjectContext,
         fetchRequest: NSFetchRequest<DTO>,
         itemCreator: @escaping (DTO) throws -> Item,
+        itemReuseKeyPaths: (item: KeyPath<Item, String>, dto: KeyPath<DTO, String>)? = nil,
         sorting: [SortValue<Item>],
         fetchedResultsControllerType: NSFetchedResultsController<DTO>.Type
     ) {
         self.itemCreator = itemCreator
+        self.itemReuseKeyPaths = itemReuseKeyPaths
         self.sorting = sorting
         changeAggregator = ListChangeAggregator<DTO, Item>(itemCreator: itemCreator)
         frc = fetchedResultsControllerType.init(
@@ -166,7 +170,7 @@ class BackgroundDatabaseObserver<Item, DTO: NSManagedObject> {
     /// When the process is done, it also updates the `_items`, which is the locally cached list of mapped items
     /// This method will be called through an operation on `processingQueue`, which will serialize the execution until `onCompletion` is called.
     private func processItems(_ changes: [ListChange<Item>]?, onCompletion: @escaping () -> Void) {
-        mapItems { [weak self] items in
+        mapItems(changes) { [weak self] items in
             guard let self = self else {
                 onCompletion()
                 return
@@ -186,22 +190,17 @@ class BackgroundDatabaseObserver<Item, DTO: NSManagedObject> {
     /// This method is intended to be called from the `managedObjectContext` that is publishing the changes (The one linked to the `NSFetchedResultsController`
     /// in this case).
     /// Once the objects are mapped, those are sorted based on `sorting`
-    private func mapItems(completion: @escaping ([Item]) -> Void) {
+    private func mapItems(_ changes: [ListChange<Item>]?, completion: @escaping ([Item]) -> Void) {
         let objects = frc.fetchedObjects ?? []
-
-        var items: [Item?] = []
-        items = objects.map { [weak self] in
-            guard self?.isDeletingDatabase == false else { return nil }
-            return try? self?.itemCreator($0)
-        }
-
-        let sorting = self.sorting
-        queue.async {
-            var result = items.compactMap { $0 }
-            if !sorting.isEmpty {
-                result = result.sort(using: sorting)
-            }
-            completion(result)
-        }
+        let items = DatabaseItemConverter.convert(
+            dtos: objects,
+            existing: rawItems,
+            changes: changes,
+            itemCreator: itemCreator,
+            itemReuseKeyPaths: itemReuseKeyPaths,
+            sorting: sorting,
+            checkCancellation: { [weak self] in self?.isDeletingDatabase == true }
+        )
+        completion(items)
     }
 }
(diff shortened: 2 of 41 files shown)


Sources/StreamChat/Controllers/ReactionListController/ReactionListController.swift
```diff
@@ -172,7 +172,13 @@ extension ChatReactionListController {
             _ itemCreator: @escaping (MessageReactionDTO) throws -> ChatMessageReaction
         )
             -> ListDatabaseObserverWrapper<ChatMessageReaction, MessageReactionDTO> = {
-                ListDatabaseObserverWrapper(isBackground: $0, database: $1, fetchRequest: $2, itemCreator: $3)
+                ListDatabaseObserverWrapper(
+                    isBackground: $0,
+                    database: $1,
+                    fetchRequest: $2,
+                    itemCreator: $3,
+                    itemReuseKeyPaths: nil // no id
```
martinmitrevski, line 180: What would be the behaviour for reactions in this case? All of the models would be re-created, right?

Sources/StreamChat/Utils/Database/DatabaseItemConverter.swift
```diff
@@ -0,0 +1,57 @@
...
+        changes: [ListChange<Item>]?,
+        itemCreator: (DTO) throws -> Item,
+        itemReuseKeyPaths: (item: KeyPath<Item, String>, dto: KeyPath<DTO, String>)?,
+        sorting: [SortValue<Item>],
+        checkCancellation: () -> Bool
+    ) -> [Item] where DTO: NSManagedObject {
+        let items: [Item]
+        
+        // Reuse converted items by id
+        if let itemReuseKeyPaths, !existing.isEmpty {
```
martinmitrevski, line 32: thinking if we should feature flag it (by default enabled though), so we can easily disable it if things go wrong. Wdyt?

Tests/StreamChatTests/StateLayer/StateLayerDatabaseObserver_Tests.swift
```diff
@@ -203,6 +204,35 @@ final class StateLayerDatabaseObserver_Tests: XCTestCase {
         XCTAssertEqual(2, changeCount)
     }
     
+    // MARK: - Reusing Existing Items
+    
+    func test_reuseItems_whenSomeItemsChange_thenOthersAreReused() async throws {
```
martinmitrevski, line 209: maybe some more tests would be nice, covering different models

martinmitrevski (commented):
great work 👏  Added few small comments. Would be great if @testableapple tests this a bit, first here, then in the SwiftUI branch.

Sources/StreamChat/Controllers/ReactionListController/ReactionListController.swift
laevandus: Yes, everything in FRC is recreated. 
Did not have a good idea how to create a reliable reuse key here, therefore skipped for now.

Sources/StreamChat/StateLayer/DatabaseObserver/StateLayerDatabaseObserver.swift
```diff
@@ -52,14 +57,27 @@ final class StateLayerDatabaseObserver<ResultType: DatabaseObserverType, Item, D
 
 @available(iOS 13.0, *)
 extension StateLayerDatabaseObserver where ResultType == EntityResult {
+    convenience init(
+        databaseContainer: DatabaseContainer,
+        fetchRequest: NSFetchRequest<DTO>,
+        itemCreator: @escaping (DTO) throws -> Item
+    ) {
```
laevandus, line 64: Created separate init methods since list vs entity observing uses different arguments. Sorting does not make sense for single entity observing.

Sources/StreamChat/StateLayer/MessageState+Observer.swift
```diff
@@ -30,7 +30,8 @@ extension MessageState {
                     for: messageId,
                     sort: ChatMessageReaction.defaultSortingDescriptors()
                 ),
-                itemCreator: { try $0.asModel() }
+                itemCreator: { try $0.asModel() },
+                itemReuseKeyPaths: nil // no id
```
laevandus, line 34: Review after we have discussed custom id for reactions

Sources/StreamChat/StateLayer/ReactionListState+Observer.swift
```diff
@@ -13,7 +13,8 @@ extension ReactionListState {
             reactionListObserver = StateLayerDatabaseObserver(
                 databaseContainer: database,
                 fetchRequest: MessageReactionDTO.reactionListFetchRequest(query: query),
-                itemCreator: { try $0.asModel() }
+                itemCreator: { try $0.asModel() },
+                itemReuseKeyPaths: nil // no id
```
laevandus, line 17: Review after we have discussed custom id for reactions

Sources/StreamChat/Utils/Database/DatabaseItemConverter.swift
laevandus: I think it makes sense. I will add one.

Sources/StreamChat/Controllers/ReactionListController/ReactionListController.swift
laevandus: MessageReactionDTO has `id`, I'll expose it on the ChatMessageReaction level and then I can use it for reusing.

laevandus: Done

Tests/StreamChatTests/StateLayer/StateLayerDatabaseObserver_Tests.swift
laevandus: I'll add one for each of the model.

Sources/StreamChat/Config/StreamRuntimeCheck.swift
```diff
@@ -38,4 +38,7 @@ public enum StreamRuntimeCheck {
     ///
     ///  Enables using our legacy web socket connection.
     public static var _useLegacyWebSocketConnection = false
+    
+    /// Enables reusing unchanged converted items in database observers.
+    public static var isDatabaseObserverItemReusingEnabled = true
```
martinmitrevski, line 43: can we add _ as a prefix, like the other ones?

martinmitrevski (approved):
Looks great 👌  We just need to use _ for the runtime check. @testableapple can you do a QA?

Sources/StreamChat/Controllers/DatabaseObserver/BackgroundDatabaseObserver.swift
```diff
@@ -186,22 +190,17 @@ class BackgroundDatabaseObserver<Item, DTO: NSManagedObject> {
...
-
-        let sorting = self.sorting
-        queue.async {
-            var result = items.compactMap { $0 }
-            if !sorting.isEmpty {
-                result = result.sort(using: sorting)
-            }
-            completion(result)
-        }
+        let items = DatabaseItemConverter.convert(
```
nuno-vieira, line 195: Can't we support this on the regular Database observers as well? 🤔

Sources/StreamChat/Utils/Database/DatabaseItemConverter.swift
```diff
@@ -0,0 +1,57 @@
+//
+// Copyright © 2024 Stream.io Inc. All rights reserved.
+//
+
+import CoreData
+import Foundation
+
+/// Converts database models to immutable value types by reusing existing items.
+enum DatabaseItemConverter {
```
nuno-vieira, line 9: Should we add some test coverage to this component To make sure the logic is correct?

nuno-vieira (commented):
Looks great 🙌 Just added a few questions!

Sources/StreamChat/Config/StreamRuntimeCheck.swift
```diff
@@ -38,4 +38,9 @@ public enum StreamRuntimeCheck {
     ///
     ///  Enables using our legacy web socket connection.
     public static var _useLegacyWebSocketConnection = false
+    
+    /// For *internal use* only
+    ///
+    /// Enables reusing unchanged converted items in database observers.
+    public static var _isDatabaseObserverItemReusingEnabled = true
```
nuno-vieira, line 45: Additionally, we probably could set it to false by default. If we fingd any issue with this, we will need to tell every customer to disable this 🤔

Sources/StreamChat/Config/StreamRuntimeCheck.swift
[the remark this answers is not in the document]
nuno-vieira: Additionally, we probably could set it to false by default. If we fingd any issue with this, we will need to tell every customer to disable this 🤔

Sources/StreamChat/Controllers/DatabaseObserver/BackgroundDatabaseObserver.swift
laevandus: Since we turned on background mapping, then I did not bother update those (we use `BackgroundDatabaseObserver` over `EntityDatabaseObserver`).
Other ones we'll delete at some point.

Sources/StreamChat/Utils/Database/DatabaseItemConverter.swift
laevandus: I added tests one layer above since I needed DTOs and models. Therefore, this is covered indirectly.

nuno-vieira: I see; I was thinking if this was possible to test as Unit Test only, but thanks for clarifying 👍

Sources/StreamChat/Controllers/DatabaseObserver/BackgroundDatabaseObserver.swift
nuno-vieira: I see, makes sense 👍

Sources/StreamChat/Config/StreamRuntimeCheck.swift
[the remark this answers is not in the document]
nuno-vieira: Discussed internally, and the majority wants to enable this by default

nuno-vieira (approved):

testableapple (approved):
Looks awesome 🤠

PR #3268, opened 2024-06-26
[CI] Update frame rate row in xcmetrics report

testableapple:
### 📝 Summary

In our xcmetrics report we assumed that the benchmark should always be lower than the branch value. This is wrong for the `frame_rate` metric. It turned out that the higher fps, the better 🙂

### ☑️ Contributor Checklist


fastlane/Fastfile
```diff
@@ -400,23 +400,38 @@ lane :xcmetrics do |options|
...
+            fail_status
+          else
+            success_status
+          end
         end
 
       benchmark_value_avoids_zero_division = benchmark_value == 0 ? 1 : benchmark_value
       diff = ((benchmark_value - branch_value) * 100.0 / benchmark_value_avoids_zero_division).round(2)
-      diff_emoji = diff > 0 ? '🔼' : '🔽'
+      diff_emoji = (is_frame_rate && diff < 0) || (!is_frame_rate && diff > 0) ? '🔼' : '🔽'
```
laevandus, line 434: Can we add equal sign here as well when there is no difference

laevandus (approved):

fastlane/Fastfile
testableapple: Then share the emoji you'd like to see in this case 😄

laevandus: 🟰:)

nuno-vieira (approved):

fastlane/Fastfile
testableapple: done

PR #3277, opened 2024-07-04
Add additional fields to Equatable conformance in ChatChannel and ChatMessage

laevandus:
### 🔗 Issue Links

Related [PBE-4291](https://stream-io.atlassian.net/browse/PBE-4291)

### 🎯 Goal

Add more fields to equality checks for ChatChannel and ChatMessage for SwiftUI.

### 📝 Summary

SwiftUI SDK switched from custom view ids to relying on model diffing. Therefore, equality checks need to return false if the model has changed.

### ☑️ Contributor Checklist


Sources/StreamChat/Models/ChatMessage.swift
```diff
@@ -413,7 +413,12 @@ public extension ChatMessage {
 
 extension ChatMessage: Hashable {
     public static func == (lhs: Self, rhs: Self) -> Bool {
-        lhs.id == rhs.id
+        lhs.id == rhs.id &&
```
martinmitrevski, line 416: How did you come up with these properties? Maybe we can think about unifying it with this one: https://github.com/GetStream/stream-chat-swift/blob/081601c1e9cb925e8aa0f6470f8e3120add5ca54/Sources/StreamChatUI/ChatMessageList/ChatMessageListView%2BDiffKit.swift#L34.

laevandus: Some trial and error but I realised now that I need to add a lot more due to the offline support where the updatedAt won't get changed if I am offline. Sounds like it makes sense to copy what we do for UIKit list updates.

laevandus: I unified them with what we have for DiffKit.

martinmitrevski: do you think it might have performance impact?

Sources/StreamChat/Models/Channel.swift
```diff
@@ -274,7 +274,27 @@ extension ChatChannel: AnyChannel {}
...
+            lhs.memberCount == rhs.memberCount &&
+            lhs.membership == rhs.membership &&
+            lhs.muteDetails == rhs.muteDetails &&
+            lhs.name == rhs.name &&
+            lhs.ownCapabilities == rhs.ownCapabilities &&
+            lhs.previewMessage == rhs.previewMessage &&
+            lhs.reads == rhs.reads &&
+            lhs.team == rhs.team &&
+            lhs.truncatedAt == rhs.truncatedAt &&
+            lhs.watcherCount == rhs.watcherCount
```
nuno-vieira, line 297: Better to reuse the DiffKit implementation 🤔 Or we can copy the DiffKit implementation here, and then change the DiffKit's `isContentEqual` to just `lhs == rhs`

Sources/StreamChatUI/ChatChannelList/ChatChannelListVC.swift
```diff
@@ -474,26 +474,6 @@ open class ChatChannelListVC: _ViewController,
...
-            memberCount == source.memberCount &&
-            membership == source.membership &&
-            watcherCount == source.watcherCount &&
-            team == source.team &&
-            reads == source.reads &&
-            muteDetails == source.muteDetails &&
-            cooldownDuration == source.cooldownDuration &&
-            extraData == source.extraData &&
-            previewMessage == source.previewMessage
+        self == source
```
nuno-vieira, line 477: Yes exactly 👍

CHANGELOG.md
```diff
@@ -10,6 +10,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### 🐞 Fixed
 - Fix an edge case with a semicolon in user name [#3275](https://github.com/GetStream/stream-chat-swift/pull/3275)
 
+### 🔄 Changed
```
nuno-vieira, line 13: Isn't this a fix? 🤔

nuno-vieira (approved):
LGTM! ✅

Sources/StreamChat/Models/Channel.swift
martinmitrevski: this is better actually, since DiffKit is in UIKit and not needed in SwiftUI

CHANGELOG.md
laevandus: From SwiftUI's point of view it is a fix. I'll change it around.

martinmitrevski: Not really, more like a change we're doing. It was never a bug.

Sources/StreamChat/Models/Channel.swift
laevandus: Copied and sorted but moved updatedAt to the top since it is the most probable one to change

CHANGELOG.md
martinmitrevski: The context here is LLC - and we change the equatable conformance. The edit message bug will be put in the SwiftUI Fixed changelog accordingly.

Sources/StreamChat/Models/ChatMessage.swift
laevandus: Need to profile but since we compare more properties then some impact indeed. Don't think it is going to be much :)

CHANGELOG.md
nuno-vieira: Makes sense, I'm fine with either Change or a fix in this case 👍

Sources/StreamChat/Models/ChatMessage.swift
laevandus: Profiled it a bit and seems like the longer == implementation triggers Swift to copy the passed in values.
Screenshot 2024-07-04 at 13 37 40
OR
the shorter == implementation is inlined and I never see it in the profiler. I think it is OK, in the future we probably want to go for making these types non-copiable anyway.

martinmitrevski (approved):
✅  Check why the tests are failing before merging

PR #3285, opened 2024-07-08
Drop iOS 12 support

martinmitrevski:
### 🔗 Issue Links

Resolves https://stream-io.atlassian.net/browse/PBE-4993.

### 🎯 Goal

Drop iOS 12 support, since its usage is pretty low and it makes it harder to maintain it. It will also facilitate the adoption of Swift 6.

### 📝 Summary

Removed starscream, and some iOS < 13 code.

### 🛠 Implementation

Mostly removing code.

### 🎨 Showcase

| Before | After |
| ------ | ----- |
|  img   |  img  |

### 🧪 Manual Testing Notes

### ☑️ Contributor Checklist

### 🎁 Meme

Alt Text


laevandus:
There are some comments like "Any because CheckedContinuation<Void, Error> requires iOS 13" what we can fix and clean up as well.

martinmitrevski:
> There are some comments like "Any because CheckedContinuation<Void, Error> requires iOS 13" what we can fix and clean up as well.

Good point, fixed those as well.

laevandus (approved):

nuno-vieira:
@martinmitrevski We still have a couple of places where we say "Remove this when dropping iOS 12 support" can we tackle those as well?

Package.swift
```diff
@@ -7,7 +7,7 @@ let package = Package(
     name: "StreamChat",
     defaultLocalization: "en",
     platforms: [
-        .iOS(.v12), .macOS(.v10_15)
+        .iOS(.v13), .macOS(.v11)
```
nuno-vieira, line 10: - We need to update the README.me iOS bagde to iOS 13+ (it is still on 11 actually)
- We also need to update the Documentation.docc which we say iOS 12+

martinmitrevski: done

martinmitrevski:
> @martinmitrevski We still have a couple of places where we say "Remove this when dropping iOS 12 support" can we tackle those as well?
> 
> ChannelMemberListQueryDTO.Line:61, we can remove the additional nil check

Good point, removed those as well.

nuno-vieira (approved):

PR #3299, opened 2024-07-12
Background mapping always enabled cleanup

laevandus:
### 🔗 Issue Links

Resolves [PBE-4853](https://stream-io.atlassian.net/browse/PBE-4853)

### 🎯 Goal

Remove the background mapping flag and cleanup unneeded code.

### 📝 Summary

- Always use background mapping by deleting the runtime check
- Remove `CoreDataLazy` which not needed anymore in `ChatMessage`, `ChatChannel`, and `CurrentChatUser`
- Remove laziness from the `LazyCachedMapCollection`
- Remove `EntityDatabaseObserverWrapper` and `ListDatabaseObserverWrapper` and use background variants directly

Note that `ChatMessageController.message` and `ChatCurrentUser.user` does not use background DB observers.

### 🧪 Manual Testing Notes

Exploratory testing.

### ☑️ Contributor Checklist

### 🎁 Meme

[PBE-4853]: https://stream-io.atlassian.net/browse/PBE-4853?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Sources/StreamChat/Controllers/DatabaseObserver/BackgroundEntityDatabaseObserver.swift
```diff
@@ -0,0 +1,91 @@
+//
+// Copyright © 2024 Stream.io Inc. All rights reserved.
+//
+
+import CoreData
+import Foundation
+
+/// Observes changes of a single entity specified using an `NSFetchRequest`in the provided `NSManagedObjectContext`.
+/// This observation is performed on the background
+class BackgroundEntityDatabaseObserver<Item, DTO: NSManagedObject>: BackgroundDatabaseObserver<Item, DTO> {
```
laevandus, line 10: Moved it to a separate file, no new code.

Sources/StreamChat/Controllers/DatabaseObserver/ListDatabaseObserver.swift
```diff
@@ -252,12 +252,12 @@ class ListDatabaseObserver<Item, DTO: NSManagedObject> {
...
             var result: LazyCachedMapCollection<Item>!
             context.performAndWait {
-                result = LazyCachedMapCollection(
-                    source: frc.fetchedObjects ?? [],
-                    itemCreator: itemCreator,
-                    sorting: sorting,
-                    context: context
-                )
+                let dtos = frc.fetchedObjects ?? []
+                var mapped = dtos.compactMapLoggingError { try itemCreator($0) }
```
laevandus, line 256: Old code had the assert which caused the crash (see the LazyCachedMapCollection init for more information), new code logs the error and skips the element.

Sources/StreamChat/Controllers/DatabaseObserver/BackgroundListDatabaseObserver.swift
```diff
@@ -5,119 +5,32 @@
...
             sorting: sorting,
             fetchedResultsControllerType: fetchedResultsControllerType
         )
     }
+    
+    /// Since DB updates now happen in a background thread, sometimes we need to
+    /// wait for the updates to do some action, so this function is useful for that.
+    func refreshItems(completion: @escaping () -> Void) {
+        updateItems(changes: nil, completion: completion)
+    }
```
laevandus, line 35: Old code, moved from the existing wrapper

Sources/StreamChat/Models/ChatMessage.swift
```diff
@@ -61,8 +61,8 @@ public struct ChatMessage {
     ///
     /// If message is inline reply this property will contain the message quoted by this reply.
     ///
-    public var quotedMessage: ChatMessage? { _quotedMessage }
-    @CoreDataLazy internal var _quotedMessage: ChatMessage?
+    public var quotedMessage: ChatMessage? { _quotedMessage() }
+    let _quotedMessage: () -> ChatMessage?
```
laevandus, line 65: `public let quotedMessage: ChatMessage?` fails to compile due to the error of recursive reference of the type.

nuno-vieira, line 65: Why is this still a closure? Can't this be static data as well?

laevandus: `Value type 'ChatMessage' cannot have a stored property that recursively contains it`
I am not sure if there is way. indirect was enums but for properties, hmm

nuno-vieira (commented):
Really exciting for this finally to become a reality 🙌

Sources/StreamChat/Models/ChatMessage.swift
laevandus: [Here](https://forums.swift.org/t/value-type-state-cannot-have-a-stored-property-that-recursively-contains-it/63622/6) is some discussion about it. Hmm

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
```diff
@@ -437,12 +437,8 @@ extension ChatChannel {
 
         let reads: [ChatChannelRead] = try dto.reads.map { try $0.asModel() }
 
-        let unreadCount: () -> ChannelUnreadCount = {
```
martinmitrevski, line 440: we don't need the `isValid` checks anymore?

Sources/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -1309,72 +1309,43 @@ private extension ChatMessage {
 
         if let currentUser = context.currentUser {
             isSentByCurrentUser = currentUser.user.id == dto.user.id
-            $_currentUserReactions = ({
```
martinmitrevski, line 1312: what's happening here, can you share more details? :)

Sources/StreamChat/Models/Channel.swift
```diff
@@ -68,12 +68,10 @@ public struct ChatChannel {
     ///
     /// - Note: This property will contain no more than `ChatClientConfig.channel.lastActiveMembersLimit` members.
     ///
-    public var lastActiveMembers: [ChatChannelMember] { _lastActiveMembers }
-    @CoreDataLazy private var _lastActiveMembers: [ChatChannelMember]
```
martinmitrevski, line 72: really nice to finally see this 😄

martinmitrevski (commented):
Great stuff 👍  Before merging, we should test and profile extensively, also on SwiftUI. Alexey should also do a round here.

Sources/StreamChat/Models/ChatMessage.swift
laevandus: Seems like the workaround I have at the moment is the most simple one. But otherwise it is not possible at the moment to have it like other properties.

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
laevandus: Because it runs immediately then I don't see any reason why we would need it. Made sense when this was called at a random point of time in the future.

Sources/StreamChat/Database/DTOs/MessageDTO.swift
laevandus: Diff looks pretty unreadable here.
Old approach captured a closure which was fed into CoreDataLazy. CoreDataLazy then ran this closure. Since it is a property wrapper then the call site looks a bit weird. But this line triggers CoreDataLazy's projectedValue.

Now we run the code immediately here instead.

Sources/StreamChat/Config/StreamRuntimeCheck.swift
```diff
@@ -29,9 +24,7 @@ public enum StreamRuntimeCheck {
     ///
     ///  Returns true if the maximum depth of relationships to fetch when performing a mapping is not yet met
     static func _canFetchRelationship(currentDepth: Int) -> Bool {
-        guard _isBackgroundMappingEnabled else { return true }
-
-        return currentDepth <= _backgroundMappingRelationshipsMaxDepth
+        currentDepth <= _backgroundMappingRelationshipsMaxDepth
```
nuno-vieira, line 27: Do we still need this? 🤔

DemoApp/StreamChat/Components/DemoChatChannelListVC.swift
```diff
@@ -199,8 +199,6 @@ final class DemoChatChannelListVC: ChatChannelListVC, EventsControllerDelegate {
 
     func eventsController(_ controller: EventsController, didReceiveEvent event: Event) {
         if let newMessageEvent = event as? MessageNewEvent {
-            // This is a DemoApp integration test to make sure there are no deadlocks when
```
nuno-vieira, line 202: We probably can remove this now, no need to leave this here 🤔

Sources/StreamChat/Controllers/MemberListController/MemberListController.swift
```diff
@@ -95,9 +95,8 @@ public class ChatChannelMemberListController: DataController, DelegateCallable,
         )
     }
 
-    private func createMemberListObserver() -> ListDatabaseObserverWrapper<ChatChannelMember, MemberDTO> {
+    private func createMemberListObserver() -> BackgroundListDatabaseObserver<ChatChannelMember, MemberDTO> {
```
nuno-vieira, line 98: ⚠️  This actually needs to the the Main thread Database observer, otherwise, user mentions won't work properly

Sources/StreamChat/Controllers/MemberListController/MemberListController.swift
```diff
@@ -95,9 +95,8 @@ public class ChatChannelMemberListController: DataController, DelegateCallable,
...
         let observer = environment.memberListObserverBuilder(
-            false,
```
nuno-vieira, line 100: You can see here, that the value of `isBackgroundMapping` is hardcoded to `false`

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
nuno-vieira: Yeah in theory this should not be needed anymore

nuno-vieira: Either way, as next step we should try to avoid these DB calls, and use the payloads data directly

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
```diff
@@ -499,30 +492,27 @@ extension ChatChannel {
...
-        let fetchMembers: () -> [ChatChannelMember] = {
-            MemberDTO
-                .loadLastActiveMembers(cid: cid, context: context)
-                .compactMap { try? $0.asModel() }
-        }
+        let watchers = UserDTO.loadLastActiveWatchers(cid: cid, context: context)
+            .compactMap { try? $0.asModel() }
+        
+        let members = MemberDTO.loadLastActiveMembers(cid: cid, context: context)
+            .compactMap { try? $0.asModel() }
```
nuno-vieira, line 501: We should do what we do on `ThreadList` with the `latestReplies`, to avoid DB queries. Just a note, since this PR is out of scope

Sources/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -1309,72 +1309,43 @@ private extension ChatMessage {
...
-            Set(dto.reads.compactMap { try? $0.user.asModel() })
-        }
+        latestReplies = {
+            guard !dto.replies.isEmpty else { return [] }
+            return MessageDTO.loadReplies(for: dto.id, limit: 5, context: context)
+                .compactMap { try? ChatMessage(fromDTO: $0, depth: depth) }
+        }()
+
+        let message = try? dto.quotedMessage?.relationshipAsModel(depth: depth)
+        _quotedMessage = { message }
```
nuno-vieira, line 1346: Can't this just be `quotedMessage = try? dto.quotedMessage?.relationshipAsModel(depth: depth)`?

It would be nice to get rid of this relationship depth. :/

Sources/StreamChat/Models/ChatMessage.swift
nuno-vieira: What I mean, is remove the stored property all together, we shouldn't need to have a stored property here 🤔

Sources/StreamChat/Utils/LazyCachedMapCollection.swift
```diff
@@ -5,98 +5,60 @@
...
 
     public mutating func append(_ element: Element) {
-        // Even though `LazyCachedMapCollection` is a value type,
-        // the `cache` is not, so whenever we append a new value,
-        // we need to make sure we are creating a copy.
-        self = LazyCachedMapCollection(self)
-        cache.values.append(element)
+        var elements = mappedElements
+        elements.append(element)
+        self = LazyCachedMapCollection(elements: elements)
```
nuno-vieira, line 53: Can we just do this? 🤔 
```suggestion
        mappedElements.append(element)
```

nuno-vieira (commented):
Looks really good overall 👍 Just did request changes, because of `MemberListController`

nuno-vieira (dismissed):
Forgot to request changes

Sources/StreamChat/Controllers/MemberListController/MemberListController.swift
laevandus: I'll change it back but we should tackle this in following PRs, I bet it is related to the issue of accessing initial items.

Sources/StreamChat/Database/DTOs/MessageDTO.swift
laevandus: Unfortunately this is not possible because the type of the property is ChatMessage (the recursive type error discussed below)

Sources/StreamChat/Models/ChatMessage.swift
laevandus: Can you give me a rough example please. Because it is in the public interface then I can't change it much and since there is the compiler error about recursive types then I am not sure how to refactor it.

Sources/StreamChat/Config/StreamRuntimeCheck.swift
laevandus: I don't know the background of it but feels like it is just limiting how much work the DB fetch is doing. Not sure about the side-effects. I think I would like to do this in a separate PR and then spend some time testing it.
Because we throw `RecursionError` then I would like to take it separately.

DemoApp/StreamChat/Components/DemoChatChannelListVC.swift
laevandus: Done

Sources/StreamChat/Controllers/MemberListController/MemberListController.swift
laevandus: Changed it back

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
laevandus: That would be the next logical step, I also have some ideas on how to do it.

laevandus: For now I kept it as simple as possible and just converted these closures to run immediately.

Sources/StreamChat/Utils/LazyCachedMapCollection.swift
laevandus: Yes, if I make it var. Changed it

Sources/StreamChat/Models/ChatMessage.swift
martinmitrevski: We shouldn't remove the property btw, it's a public API that's used a lot. As for the implementation details, up to you, the closure looks ok to me.

Sources/StreamChat/Config/StreamRuntimeCheck.swift
nuno-vieira: Yes lets handle it in another PR 👍

Sources/StreamChat/Models/ChatMessage.swift
nuno-vieira: This is what I mean:
```suggestion
    public var quotedMessage: ChatMessage?
```

When creating the ChatMessage, we just do this:
`.init(quotedMessage: try? dto.quotedMessage?.relationshipAsModel(depth: depth))`

But I think we could just have `dto.quotedMessage` only maybe? 🤔

Sources/StreamChat/Database/DTOs/MessageDTO.swift
nuno-vieira: Ah I got it, it is a limitation of value types

Sources/StreamChat/Models/ChatMessage.swift
nuno-vieira: Ah I got it, it is a limitation of value types

nuno-vieira (approved):
LGTM! ✅

laevandus:
I did another round of profiling against UIKit and SwiftUI demo apps. Something to consider is to think about how to reduce running multiple fetch requests when creating ChatChannel but that is not new in this PR.

martinmitrevski (approved):

PR #3303, opened 2024-07-12
Add timeout support when connecting user

nuno-vieira:
### 🔗 Issue Links
https://stream-io.atlassian.net/browse/PBE-4780

### 🎯 Goal

Add support for timing out reconnection.

### 📝 Summary

SDK:
- Add an option to configure reconnection timeout
- Fix Channel List not hiding error view when data is available
- Fix not trying to connect if the network is not available (This can cause edge cases, according to Apple)
- Ignore network status since it is not reliable, even on the device sometimes network status is not returned correctly
- Fix invalid token errors considered as possible to reconnect

DemoApp:
- Fix the demo app not showing the connection bar
- Fixes fake token failures in the demo app not working
- Fix Demo App channel list error view animation

### 🛠 Implementation

An alternative approach to https://github.com/GetStream/stream-chat-swift/pull/3302 uses a separate entity to control the timeout, independent of the exponential backoff mechanism.

It uses a separate timeout handler to timeout the reconnection logic instead of using a number of retries of the existing backoff retry mechanism. The 2 main reasons for using a separate timeout handler are the following:
- It is clearer for the customer to expose a maximum timeout time than a maximum number of retries.
- It has less impact in the implementation of `RetryStrategy`, which at the moment has some flaws, described in this document: https://www.notion.so/stream-wiki/Reconnection-Problems-63eeaaab19d94d50b9922b76ebb414ab

### 🧪 Manual Testing Notes

**With timeout**
1. Go To Demo App Configuration Screen
2. Set reconnection timeout, ex: 15s
3. Disable Network on your device
4. Login
5. Should show "Connecting..." banner
6. After 15s
7. Should show "Disconnect" banner
8. Enable Network on your device
9. Should show "Connecting.." and then "Connected" and then disappear
10. Tap on Error View to reload the channel list
11. Should load channel list

**Without timeout**
1. Go To Demo App Configuration Screen
2. Disable timeout (By default)
3. Disable Network on your device
4. Login
5. Should show "Connecting..." banner
6. Wait 60s
7. Should still show "Connecting..." banner
8. Enable Network on your device
9. Should show "Connected" and then disappear
10. Tap on Error View to reload the channel list
11. Should load channel list

**Retest token refresh**:
1. Open Demo App Configuration
2. Edit tokenRefreshDetails
3. Add a fake App Secret
4. Set the duration, ex: 10
5. Set the number of refreshes, ex: 3
6. Login
7. Should see "Demo App Token Refreshing: Token refresh failed." 3x in Xcode Console
8. Should see "Disconnected"
9. Logout
10. Put a valid App Secret from Stream's Dashboard
11. Should see "Demo App Token Refreshing: Token refresh failed." 3x in Xcode Console
12. Then you should see "Connected"
13. For each 10 seconds, the banner should show "Connecting" -> "Connected"

### ☑️ Contributor Checklist

### DIFF
--- a/DemoApp/Shared/Token+Development.swift
+++ b/DemoApp/Shared/Token+Development.swift
@@ -26,7 +26,7 @@ extension StreamChatWrapper {
 
                 let numberOfSuccessfulRefreshes = refreshDetails.numberOfSuccessfulRefreshesBeforeFailing
                 let shouldNotFail = numberOfSuccessfulRefreshes == 0
-                if shouldNotFail || self.numberOfRefreshTokens <= numberOfSuccessfulRefreshes {
+                if shouldNotFail || self.numberOfRefreshTokens >= numberOfSuccessfulRefreshes {
                     print("Demo App Token Refreshing: New token generated successfully.")
                     let newToken = generatedToken ?? initialToken
                     completion(.success(newToken))
--- a/Sources/StreamChat/ChatClient.swift
+++ b/Sources/StreamChat/ChatClient.swift
@@ -258,7 +258,7 @@ public class ChatClient {
             environment.backgroundTaskSchedulerBuilder(),
             environment.internetConnection(eventNotificationCenter, environment.internetMonitor),
             config.staysConnectedInBackground,
-            config.reconnectionTimeout.map { ScheduledStreamTimer(interval: $0, fireOnStart: false) }
+            config.reconnectionTimeout.map { ScheduledStreamTimer(interval: $0, fireOnStart: false, repeats: false) }
         )
     }
 
--- a/Sources/StreamChat/ChatClient.swift
+++ b/Sources/StreamChat/ChatClient.swift
@@ -288,6 +288,8 @@ public class ChatClient {
         tokenProvider: @escaping TokenProvider,
         completion: ((Error?) -> Void)? = nil
     ) {
+        connectionRecoveryHandler?.start()
+
         authenticationRepository.connectUser(
             userInfo: userInfo,
             tokenProvider: tokenProvider,
@@ -437,6 +439,7 @@ public class ChatClient {
     /// Disconnects the chat client from the chat servers. No further updates from the servers
     /// are received.
     public func disconnect(completion: @escaping () -> Void) {
+        connectionRecoveryHandler?.stop()
         connectionRepository.disconnect(source: .userInitiated) {
             log.info("The `ChatClient` has been disconnected.", subsystems: .webSocket)
             completion()
--- a/Sources/StreamChat/ChatClient.swift
+++ b/Sources/StreamChat/ChatClient.swift
@@ -288,8 +288,6 @@ public class ChatClient {
         tokenProvider: @escaping TokenProvider,
         completion: ((Error?) -> Void)? = nil
     ) {
-        connectionRecoveryHandler?.start()
-
         authenticationRepository.connectUser(
             userInfo: userInfo,
             tokenProvider: tokenProvider,
@@ -439,7 +437,6 @@ public class ChatClient {
     /// Disconnects the chat client from the chat servers. No further updates from the servers
     /// are received.
     public func disconnect(completion: @escaping () -> Void) {
-        connectionRecoveryHandler?.stop()
         connectionRepository.disconnect(source: .userInitiated) {
             log.info("The `ChatClient` has been disconnected.", subsystems: .webSocket)
             completion()
--- a/Sources/StreamChat/ChatClient.swift
+++ b/Sources/StreamChat/ChatClient.swift
@@ -257,7 +257,8 @@ public class ChatClient {
             extensionLifecycle,
             environment.backgroundTaskSchedulerBuilder(),
             environment.internetConnection(eventNotificationCenter, environment.internetMonitor),
-            config.staysConnectedInBackground
+            config.staysConnectedInBackground,
+            config.reconnectionTimeout.map { ScheduledStreamTimer(interval: $0, fireOnStart: false) }
         )
     }
 
--- a/Sources/StreamChat/ChatClient.swift
+++ b/Sources/StreamChat/ChatClient.swift
@@ -289,6 +289,8 @@ public class ChatClient {
         tokenProvider: @escaping TokenProvider,
         completion: ((Error?) -> Void)? = nil
     ) {
+        connectionRecoveryHandler?.start()
+
         authenticationRepository.connectUser(
             userInfo: userInfo,
             tokenProvider: tokenProvider,
@@ -380,6 +382,7 @@ public class ChatClient {
         userInfo: UserInfo,
         completion: ((Error?) -> Void)? = nil
     ) {
+        connectionRecoveryHandler?.start()
         authenticationRepository.connectGuestUser(userInfo: userInfo, completion: { completion?($0) })
     }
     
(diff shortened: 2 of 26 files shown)


DemoApp/Shared/Token+Development.swift
```diff
@@ -26,7 +26,7 @@ extension StreamChatWrapper {
 
                 let numberOfSuccessfulRefreshes = refreshDetails.numberOfSuccessfulRefreshesBeforeFailing
                 let shouldNotFail = numberOfSuccessfulRefreshes == 0
-                if shouldNotFail || self.numberOfRefreshTokens <= numberOfSuccessfulRefreshes {
+                if shouldNotFail || self.numberOfRefreshTokens >= numberOfSuccessfulRefreshes {
```
martinmitrevski, line 29: why has this changed?

Sources/StreamChat/WebSocketClient/ConnectionStatus.swift
```diff
@@ -42,9 +42,7 @@ extension ConnectionStatus {
             self = .disconnecting
 
         case let .disconnected(source):
-            let isWaitingForReconnect = webSocketConnectionState.isAutomaticReconnectionEnabled || source.serverError?
-                .isInvalidTokenError == true
```
martinmitrevski, line 46: why do we remove the invalid token condition?

martinmitrevski (commented):
Looks good, just added few questions. In any case, @testableapple  can start testing.

Sources/StreamChat/Workers/Background/ConnectionRecoveryHandler.swift
```diff
@@ -136,14 +148,12 @@ extension DefaultConnectionRecoveryHandler {
     }
 
     @objc private func internetConnectionAvailabilityDidChange(_ notification: Notification) {
-        guard let isAvailable = notification.internetConnectionStatus?.isAvailable else { return }
-
-        log.debug("Internet -> \(isAvailable ? "✅" : "❌")", subsystems: .webSocket)
+        if let isAvailable = notification.internetConnectionStatus?.isAvailable {
```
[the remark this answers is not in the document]
nuno-vieira: Yes, for now it is only used only for logging, we don't care about the state that it is reported. If the connection status changes, and we are not connectted or disconnected on purpose, we try to connect

DemoApp/Shared/Token+Development.swift
nuno-vieira: This was not working at all, I don't remember when I changed this, but I broke it. 

Glad you made this comment, because I forgot about adding this scenario to the QA Scenarios.

Sources/StreamChat/WebSocketClient/ConnectionStatus.swift
nuno-vieira: Because it does not make sense, if the token is invalid, it should not mean we are reconnecting, it should stop trying to automatically reconnect. In fact, inside `isAutomaticReconnectionEnabled` this error makes it `false`. But then here we set it to `true`

martinmitrevski: the var name is `isWaitingForReconnect` though - it determines whether the state should be connecting or disconnected. If it's an invalid token error, doesn't that mean that maybe a token refresh is in progress?

nuno-vieira: Inside the `isAutomaticReconnectionEnabled` we have this:

image

The problem, it seems, is that we have this:

image

And it looks like this includes expired tokens:

image

So I guess this should be updated

martinmitrevski: yes, this doesn't make much sense - but this was added explicitly 2 years ago it seems, so we should be careful with this change.

nuno-vieira: Done ✅ Ive made some changes and I think now it makes much more sense

martinmitrevski (commented):
Some risky changes here and there. Can't we limit this PR to the timeout support only? This whole reconnection thing was very fragile already, not sure if we should do all these changes.

DemoApp/Shared/Token+Development.swift
nuno-vieira: Done

nuno-vieira:
> Some risky changes here and there. Can't we limit this PR to the timeout support only? This whole reconnection thing was very fragile already, not sure if we should do all these changes.

@martinmitrevski All changes are required to pass all QA Scenarios. I tried to make the changes as minimal as possible, but even with a simple change, there was always something conflicting. 

Also, we need to slowly and progressively improve and simplify this part of the codebase, and this PR is already a nice step in that direction.

Sources/StreamChat/Repositories/ConnectionRepository.swift
```diff
@@ -140,10 +140,8 @@ class ConnectionRepository {
         case let .connected(connectionId: id):
             shouldNotifyConnectionIdWaiters = true
             connectionId = id
-        case let .disconnected(source) where source.serverError?.isInvalidTokenError == true:
-            if source.serverError?.isExpiredTokenError == true {
-                onExpiredToken()
-            }
+        case let .disconnected(source) where source.serverError?.isExpiredTokenError == true:
```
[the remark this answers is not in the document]
nuno-vieira: Yes exactly, and this is the desired behaviour, according to this comment:

> // We should notify waiters if connectionId was obtained (i.e. state is .connected)
> // or for .disconnected state except for disconnect caused by an expired token

image

So the previous code was not respecting the desired logic according to the comment

Sources/StreamChat/Errors/ErrorPayload.swift
```diff
@@ -57,7 +57,7 @@ extension ErrorPayload {
 
 extension ClosedRange where Bound == Int {
     /// The error codes for token-related errors. Typically, a refreshed token is required to recover.
-    static let tokenInvalidErrorCodes: Self = StreamErrorCode.expiredToken...StreamErrorCode.invalidTokenSignature
+    static let tokenInvalidErrorCodes: Self = StreamErrorCode.notYetValidToken...StreamErrorCode.invalidTokenSignature
```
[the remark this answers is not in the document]
nuno-vieira: I've check all places where it is used and added test coverage

martinmitrevski (approved):
Let's merge it then. Please check the failing e2e tests first - something around the offline mode, not sure if it's connected to your changes.
Afterwards, @testableapple should do a thorough QA around reconnection and token expiry.

testableapple (approved):
LGTM ✅
Just one tiny issue, discussed in slack.

Sources/StreamChat/ChatClient.swift
```diff
@@ -379,6 +382,7 @@ public class ChatClient {
         userInfo: UserInfo,
         completion: ((Error?) -> Void)? = nil
     ) {
+        connectionRecoveryHandler?.start()
```
masarusanjp, line 385: @nuno-vieira @martinmitrevski Hello, I found that this code is causing an issue, so I'm leaving a comment.

`connectGuestUser` calls logOut first. As a result, ConnectionRecoveryHandler.stop is invoked, causing the monitoring of appDidBecomeActive and appDidEnterBackground to stop. This leads to the issue where "the WebSocket does not reconnect when the app resumes from the background."

masarusanjp: I posted issue https://github.com/GetStream/stream-chat-swift/issues/3338

nuno-vieira: Hi @masarusanjp, thanks for spotting this issue!

The actual issue is that when reconnecting, logOut should not be happening, so the logic should be changed here. Stopping the recovery handler on logout is something that we want, but it caused this unwanted side effect due to the fact that logOutFirst in guest users is not properly implemented, it seems.

PR #3305, opened 2024-07-15
 Reliable access to database models while observer is processing database changes

laevandus:
### 🔗 Issue Links

Related [PBE-4853](https://stream-io.atlassian.net/browse/PBE-4853)

### 🎯 Goal

Return items reflecting the current DB state when accessing observer's items while background mapping is still running.

### 📝 Summary

- Synchronously fetch items (but reuse existing) if items are accessed while FRC's change is being processed.
- Remove `refreshItems` which was used a workaround for waiting background observer to finish processing

### 🛠 Implementation

Background observers fetch initial items and process FRC changes using a queue. It is good for pushing the model mapping to a background thread but has a shortcoming if we need to access updated items before the queue has finished processing it.
Example is that I create a controller, call synchronize, and then in the completion want to access updated items. At that time the observer is actually in the middle of processing update items and would return items before the last DB change. This makes many things hard with controllers. For fixing this, this PR starts tracking the processing state. If we access items while processing is in progress, we do a synchronous access to the DB (BUT while reusing items which does not make it slow). 

### 🧪 Manual Testing Notes

Exploratory testing

#### Case 1
1. Log out
2. Log in
3. Observe that channel list goes from loading state to list without showing empty view for a split secod

### ☑️ Contributor Checklist

[PBE-4853]: https://stream-io.atlassian.net/browse/PBE-4853?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### 🔄 Changed
 - Dropped iOS 12 support [#3285](https://github.com/GetStream/stream-chat-swift/pull/3285)
 - Increase QoS for `Throttler` and `Debouncer` to `utility` [#3295](https://github.com/GetStream/stream-chat-swift/issues/3295)
-- Fetch database items immediately if the initial fetch has not finished in background database observers [#3304](https://github.com/GetStream/stream-chat-swift/issues/3304)
+- Provide immediate access to controller managed database items [#3304](https://github.com/GetStream/stream-chat-swift/issues/3304)
 
 # [4.59.0](https://github.com/GetStream/stream-chat-swift/releases/tag/4.59.0)
 _July 10, 2024_
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### 🔄 Changed
 - Dropped iOS 12 support [#3285](https://github.com/GetStream/stream-chat-swift/pull/3285)
 - Increase QoS for `Throttler` and `Debouncer` to `utility` [#3295](https://github.com/GetStream/stream-chat-swift/issues/3295)
+- Fetch database items immediately if the initial fetch has not finished in background database observers [#3304](https://github.com/GetStream/stream-chat-swift/issues/3304)
 
 # [4.59.0](https://github.com/GetStream/stream-chat-swift/releases/tag/4.59.0)
 _July 10, 2024_
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### 🔄 Changed
 - Dropped iOS 12 support [#3285](https://github.com/GetStream/stream-chat-swift/pull/3285)
 - Increase QoS for `Throttler` and `Debouncer` to `utility` [#3295](https://github.com/GetStream/stream-chat-swift/issues/3295)
-- Reliable access to database models while observer is processing database changes [#3304](https://github.com/GetStream/stream-chat-swift/issues/3304)
+- Improve reliability of accessing database models in controller provided completion handlers [#3304](https://github.com/GetStream/stream-chat-swift/issues/3304)
 
 # [4.59.0](https://github.com/GetStream/stream-chat-swift/releases/tag/4.59.0)
 _July 10, 2024_
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### 🔄 Changed
 - Dropped iOS 12 support [#3285](https://github.com/GetStream/stream-chat-swift/pull/3285)
 - Increase QoS for `Throttler` and `Debouncer` to `utility` [#3295](https://github.com/GetStream/stream-chat-swift/issues/3295)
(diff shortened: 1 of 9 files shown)


Sources/StreamChat/Controllers/DatabaseObserver/BackgroundDatabaseObserver.swift
```diff
@@ -34,22 +34,33 @@ class BackgroundDatabaseObserver<Item, DTO: NSManagedObject> {
...
         get { queue.sync { _isInitialized } }
         set { queue.async(flags: .barrier) { self._isInitialized = newValue } }
     }
 
-    private var _isDeletingDatabase: Bool = false
-    private var isDeletingDatabase: Bool {
-        get { queue.sync { _isDeletingDatabase }}
-        set { queue.async(flags: .barrier) { self._isDeletingDatabase = newValue } }
-    }
-
```
laevandus, line 52: Unused, therefore deleted it

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### 🔄 Changed
 - Dropped iOS 12 support [#3285](https://github.com/GetStream/stream-chat-swift/pull/3285)
 - Increase QoS for `Throttler` and `Debouncer` to `utility` [#3295](https://github.com/GetStream/stream-chat-swift/issues/3295)
+- Fetch database items immediately if the initial fetch has not finished in background database observers [#3304](https://github.com/GetStream/stream-chat-swift/issues/3304)
```
nuno-vieira, line 10: It probably does not make sense to provide a changelog entry 🤔 This is just implementation details. What does the customer gain with this?

Sources/StreamChat/Controllers/DatabaseObserver/BackgroundDatabaseObserver.swift
```diff
@@ -34,22 +34,33 @@ class BackgroundDatabaseObserver<Item, DTO: NSManagedObject> {
...
+        // Otherwise load items synchronously if the initial fetch has not finished
+        return queue.sync(flags: .barrier) {
+            var items = [Item]()
+            frc.managedObjectContext.performAndWait {
+                items = mapItems(changes: nil, reusableItems: [])
+            }
+            _items = items
+            _hasFetchedInitialItems = true
+            return _items
+        }
```
nuno-vieira, line 53: We should be able to remove this now `BackgroundListDatabaseObserver.refreshItems()`, I think it makes sense to be part of this PR, since the `refreshItems()` was introduced because of this limitation. After that, we should check that the Channel List and Thread List states work correctly. (Loading View, Empty View and Error View)

CHANGELOG.md
laevandus: Probably makes sense, all of this is internal details. I'll remove it.

martinmitrevski: It can actually be useful - but re-worded in a way that now you can reliably access database items immediately.

Sources/StreamChat/Controllers/DatabaseObserver/BackgroundDatabaseObserver.swift
laevandus: I'll review it

laevandus: Done, removed

CHANGELOG.md
laevandus: Tried to reword it

Sources/StreamChat/Controllers/ThreadListController/ThreadListController.swift
```diff
@@ -112,15 +112,11 @@ public class ChatThreadListController: DataController, DelegateCallable, DataSto
...
-                            self?.nextCursor = threadListResponse.next
-                            self?.hasLoadedAllThreads = threadListResponse.next == nil
-                            completion?(nil)
-                        }
-                    }
+                self?.callback {
+                    self?.state = .remoteDataFetched
+                    self?.nextCursor = threadListResponse.next
+                    self?.hasLoadedAllThreads = threadListResponse.next == nil
+                    completion?(nil)
```
nuno-vieira, line 119: @laevandus Unforuntaly it does not work :/

CHANGELOG.md
laevandus: Reworded it again. WDYT?

martinmitrevski (approved):
This has worked great in my tests, excellent work 👏  I think it even feels a bit faster.

laevandus:
Tried the branch with SwiftUI SDK and did not find problems.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### 🔄 Changed
 - Dropped iOS 12 support [#3285](https://github.com/GetStream/stream-chat-swift/pull/3285)
 - Increase QoS for `Throttler` and `Debouncer` to `utility` [#3295](https://github.com/GetStream/stream-chat-swift/issues/3295)
+- Reliable access to database models while observer is processing database changes [#3304](https://github.com/GetStream/stream-chat-swift/issues/3304)
```
nuno-vieira, line 10: WDYT about this:
```suggestion
- Reliable access to data when accessing it in controllers completion calls [#3304](https://github.com/GetStream/stream-chat-swift/issues/3304)
```

nuno-vieira (approved):
LGTM! ✅ 👌 Just did a changelog suggestion

CHANGELOG.md
laevandus: The most generic version I could think of:
`Improve reliability of accessing database models`

PR #3314, opened 2024-07-18
Add support for enabling message list view animations

nuno-vieira:
### 🔗 Issue Links

Resolves https://stream-io.atlassian.net/browse/PBE-5174

### 🎯 Goal

Add support for enabling message list animations as requested by some customers.

### 🧪 Manual Testing Notes

1. Login
2. Open a Channel
3. Send a message
4. Load more messages
5. No animations should be present
6. Logout
7. Go to configration screen
8. Enable animations
9. Repeat 2, 3, 4
10. Should have animations

### ☑️ Contributor Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Increase QoS for `Throttler` and `Debouncer` to `utility` [#3295](https://github.com/GetStream/stream-chat-swift/issues/3295)
 - Improve reliability of accessing database models in controller provided completion handlers [#3304](https://github.com/GetStream/stream-chat-swift/issues/3304)
 
+## StreamChatUI
+### ✅ Added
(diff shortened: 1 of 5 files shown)


laevandus:
Tried it with animations enabled but seems like it is not very smooth because the timestamp appears for a split second. Should we fix that separately?
https://github.com/user-attachments/assets/9727115e-419b-4205-bf3e-9191ced5d893

martinmitrevski (approved):
Should we put some disclaimer that if you enable this mode, you're in control of the animations (and the possible glitches)?

nuno-vieira:
> Tried it with animations enabled but seems like it is not very smooth because the timestamp appears for a split second. Should we fix that separately? https://github.com/user-attachments/assets/9727115e-419b-4205-bf3e-9191ced5d893

Yes, this is one of the cons of having this enabled. The customer is fine with this, so at least we give this flexibility to customers in case they want to customize it. And maybe for their UI is not that bad. The biggest problem is really loading previous messages, sometimes the messages float around.

PR #3316, opened 2024-07-18
Fix a rare issue with incorrect message order when sending messages offline

laevandus:
### 🔗 Issue Links

Resolves [PBE-4319](https://stream-io.atlassian.net/browse/PBE-4319)

### 🎯 Goal

Fix the issue of incorrect order of messages when multiple messages are in the pending send state and connection comes back.

### 📝 Summary

- Forward queued messages to `OfflineRequestsRepository` for retrying on the next web-socket connect event (otherwise we have a race between it and `MessageSender`). Message statuses are set to `.sendingFailed`.
- When `MessageSender` is waiting for connection, it sends all the requests to the offline repository until it is notified connection status change (ensures that `MessageSender` and `OfflineRequestsRepository` are not sending messages for the came channel at the same time).
- `OfflineRequestsRepository` removes duplicate send message requests (in case or manual retries while offline)

### 🛠 Implementation

`APIClient` sends failed API requests to `SyncRepository` [here](https://github.com/GetStream/stream-chat-swift/blob/8a9d94f7ed7b9b25577a9f181c4ab881bc64fbc2/Sources/StreamChat/APIClient/APIClient.swift#L171). This will then be passed to `OfflineRequestsRepository` which uses the [shouldBeQueuedOffline for deciding if it should retry](https://github.com/GetStream/stream-chat-swift/blob/develop/Sources/StreamChat/Repositories/OfflineRequestsRepository.swift#L164) the failed request.
When this happens, we have `MessageSender` and `OfflineRequestsRepository` both trying to send messages on connect. This PR eliminates this race and makes `OfflineRequestsRepository` to be solely responsible of sending messages when connection status changes. When we have connected, `MessageSender` is allowed to send messages again if user does so. But all the previous ones, are handled by `OfflineRequestsRepository`.

### 🎨 Showcase

Recorded from the point where the first request has failed with a connection issue.

| Before | After |
| ------ | ----- |
|  img   |  img  |

#### Test case 1:
1. Log in and open a channel
2. Network link conditioner to 100% loss
3. Send 5 messages
4. Wait until the failure happens (takes a while, but all 5 will get failure badge) 
5. Manually trigger retries for failed messages
6. Turn off network link conditioner in settings (I set it back to Wi-Fi, reconnection takes a bit of time as well)
7. Observe the message list > messages are sent one by one and keep the order and there is no duplicate message error in the console
8. Send some new messages > messages are sent

#### Test case 2
1. Log in and open a channel
2. Network link conditioner to 100% loss
3. Send 5 messages
4. Wait until the failure happens (takes a while, but all 5 will get failure badge) 
5. Turn off network link conditioner in settings (I set it back to Wi-Fi, reconnection takes a bit of time as well)
6. Immediately try to send some more messages

Messages are sent one by one. Currently order changes because server gives updated timestamps. Can be improved if we would like to, but I did not want to include more changes to this PR. Solution would be to update local timestamps after each message send.

### ☑️ Contributor Checklist

[PBE-4319]: https://stream-io.atlassian.net/browse/PBE-4319?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 # Upcoming
 
-### 🔄 Changed
+## StreamChat
+### 🐞 Fixed
+- Fix a rare issue with incorrect message order when sending multiple messages while offline [#3316](https://github.com/GetStream/stream-chat-swift/issues/3316)
 
 # [4.60.0](https://github.com/GetStream/stream-chat-swift/releases/tag/4.60.0)
 _July 18, 2024_
--- a/Sources/StreamChat/ChatClient.swift
+++ b/Sources/StreamChat/ChatClient.swift
@@ -660,6 +660,7 @@ extension ChatClient: ConnectionStateDelegate {
             }
         )
         connectionRecoveryHandler?.webSocketClient(client, didUpdateConnectionState: state)
+        try? backgroundWorker(of: MessageSender.self).didUpdateConnectionState(state)
     }
 }
 
@@ -671,6 +672,21 @@ extension ChatClient: ConnectionDetailsProviderDelegate {
 
     func provideConnectionId(timeout: TimeInterval = 10, completion: @escaping (Result<ConnectionId, Error>) -> Void) {
         connectionRepository.provideConnectionId(timeout: timeout, completion: completion)
+    }
+}
+
+extension ChatClient {
+    func backgroundWorker<T>(of type: T.Type) throws -> T {
+        if let worker = backgroundWorkers.compactMap({ $0 as? T }).first {
+            return worker
+        }
+        if currentUserId == nil {
+            throw ClientError.CurrentUserDoesNotExist()
+        }
+        if !config.isClientInActiveMode {
+            throw ClientError.ClientIsNotInActiveMode()
+        }
+        throw ClientError("Background worker of type \(T.self) is not set up")
     }
 }
 
--- a/Sources/StreamChat/Repositories/MessageRepository.swift
+++ b/Sources/StreamChat/Repositories/MessageRepository.swift
@@ -88,6 +88,47 @@ class MessageRepository {
             })
         }
     }
+    
+    /// Marks the message's local status to failed and adds it to the offline retry which sends the message when connection comes back.
+    func scheduleOfflineRetry(for messageId: MessageId, completion: @escaping (Result<ChatMessage, MessageRepositoryError>) -> Void) {
+        var dataEndpoint: DataEndpoint!
+        var messageModel: ChatMessage!
+        database.write { session in
+            guard let dto = session.message(id: messageId) else {
+                throw MessageRepositoryError.messageDoesNotExist
+            }
+            guard let channelDTO = dto.channel, let cid = try? ChannelId(cid: channelDTO.cid) else {
+                throw MessageRepositoryError.messageDoesNotHaveValidChannel
+            }
+            
+            // Send the message to offline handling
+            let requestBody = dto.asRequestBody() as MessageRequestBody
+            let endpoint: Endpoint<MessagePayload.Boxed> = .sendMessage(
+                cid: cid,
+                messagePayload: requestBody,
+                skipPush: dto.skipPush,
+                skipEnrichUrl: dto.skipEnrichUrl
+            )
+            dataEndpoint = endpoint.withDataResponse
+            
+            // Mark it as failed
+            dto.localMessageState = .sendingFailed
+            messageModel = try dto.asModel()
+        } completion: { [weak self] writeError in
+            if let writeError {
+                switch writeError {
+                case let repositoryError as MessageRepositoryError:
+                    completion(.failure(repositoryError))
+                default:
+                    completion(.failure(.failedToSendMessage(writeError)))
+                }
+                return
+            }
+            // Offline repository will send it when connection comes back on, until then we show the message as failed
+            self?.apiClient.queueOfflineRequest?(dataEndpoint.withDataResponse)
+            completion(.success(messageModel))
+        }
+    }
 
     func saveSuccessfullySentMessage(
         cid: ChannelId,
@@ -126,11 +167,12 @@ class MessageRepository {
             // error code for duplicated messages.
             let isDuplicatedMessageError = errorPayload.code == 4 && errorPayload.message.contains("already exists")
             if isDuplicatedMessageError {
-                database.write {
+                database.write({
                     let messageDTO = $0.message(id: messageId)
                     messageDTO?.markMessageAsSent()
+                }, completion: { _ in
                     completion(.failure(.failedToSendMessage(error)))
-                }
+                })
                 return
             }
         }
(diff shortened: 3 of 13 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### 🐞 Fixed
 - Improve the stability of the reconnection process [#3303](https://github.com/GetStream/stream-chat-swift/pull/3303)
 - Fix invalid token errors considered as recoverable errors [#3303](https://github.com/GetStream/stream-chat-swift/pull/3303)
+- Fix a rare issue with incorrect message order when sending multiple messages while offline [#3311](https://github.com/GetStream/stream-chat-swift/issues/3311)
```
martinmitrevski, line 12: This would go to the next release, so the entry would have to be moved from here.

Sources/StreamChat/APIClient/Endpoints/EndpointPath+OfflineRequest.swift
```diff
@@ -7,14 +7,15 @@ import Foundation
...
+        case .editMessage, .deleteMessage, .pinMessage, .unpinMessage, .addReaction, .deleteReaction:
             return true
         case .createChannel, .connect, .sync, .users, .guest, .members, .search, .devices, .channels, .updateChannel,
              .deleteChannel, .channelUpdate, .muteChannel, .showChannel, .truncateChannel, .markChannelRead, .markChannelUnread,
              .markAllChannelsRead, .channelEvent, .stopWatchingChannel, .pinnedMessages, .uploadAttachment, .message,
              .replies, .reactions, .messageAction, .banMember, .flagUser, .flagMessage, .muteUser, .translateMessage,
              .callToken, .createCall, .deleteFile, .deleteImage, .og, .appSettings, .threads, .thread, .markThreadRead, .markThreadUnread, .polls, .pollsQuery,
-             .poll, .pollOption, .pollOptions, .pollVotes, .pollVoteInMessage, .pollVote, .unread, .blockUser, .unblockUser:
+             .poll, .pollOption, .pollOptions, .pollVotes, .pollVoteInMessage, .pollVote, .unread, .blockUser, .unblockUser,
+             .sendMessage:
```
martinmitrevski, line 18: why is this?

martinmitrevski (commented):
Looks good 👍  Added few questions, and also e2e tests need to be checked before merging.

Sources/StreamChat/APIClient/Endpoints/EndpointPath+OfflineRequest.swift
laevandus: APIClient sends failed API requests to SyncRepository [here](https://github.com/GetStream/stream-chat-swift/blob/8a9d94f7ed7b9b25577a9f181c4ab881bc64fbc2/Sources/StreamChat/APIClient/APIClient.swift#L171). This will then be passed to `OfflineRequestsRepository` which uses this property [for deciding if it should retry](https://github.com/GetStream/stream-chat-swift/blob/develop/Sources/StreamChat/Repositories/OfflineRequestsRepository.swift#L164) the failed request.

If we allow this and send message fails while there are multiple messages queued for sending, then it causes the race between offline repo retrying a message and message sender trying to send a next message which leads to incorrect message order.

I'll add it to the PR description.

laevandus:
I'll need to do some work on updating UI-tests.

laevandus:
After a longer discussion we'll change the direction a bit.

Sources/StreamChat/ChatClient.swift
```diff
@@ -660,6 +660,7 @@ extension ChatClient: ConnectionStateDelegate {
             }
         )
         connectionRecoveryHandler?.webSocketClient(client, didUpdateConnectionState: state)
+        try? backgroundWorker(of: MessageSender.self).didUpdateConnectionState(state)
```
nuno-vieira, line 663: Can you explain what is this doing? 🤔

Sources/StreamChat/Workers/Background/MessageSender.swift
```diff
@@ -117,6 +114,15 @@ class MessageSender: Worker {
             }
         }
     }
+    
+    func didUpdateConnectionState(_ state: WebSocketConnectionState) {
+        guard case WebSocketConnectionState.connected = state else { return }
```
nuno-vieira, line 119: When the state is disconnected, dont we need to notify the message sender that it was disconnect?

Sources/StreamChat/Workers/Background/MessageSender.swift
```diff
@@ -184,38 +191,66 @@ private class MessageSendingQueue {
...
+            case .messageDoesNotExist, .messageNotPendingSend, .messageDoesNotHaveValidChannel:
+                let event = NewMessageErrorEvent(messageId: request.messageId, error: repositoryError)
+                eventsNotificationCenter.process(event)
+            case .failedToSendMessage(let clientError):
+                let event = NewMessageErrorEvent(messageId: request.messageId, error: clientError)
+                eventsNotificationCenter.process(event)
+                
+                if ClientError.isEphemeral(error: clientError) {
+                    // We hit a connection error, therefore all the remaining and upcoming requests should be scheduled for keeping the order
+                    isWaitingForConnection = true
```
nuno-vieira, line 239: I see we only set this value to true here, shouldn't we always set it to true the `didUpdateConnectionStatus` is called? 🤔

laevandus: Because we have automatic retries in APIClient then it is better to give APIClient a chance to try to send it. Message sender sends messages serially so it is not a problem to do it later here when the request actually failed because of connection.

nuno-vieira (changes requested):
Code wise for now it looks good, but I'm going to request changes, since while doing QA it is introducing some regressions 👍

Sources/StreamChat/Repositories/MessageRepository.swift
```diff
@@ -88,6 +88,47 @@ class MessageRepository {
             })
         }
     }
+    
+    /// Marks the message's local status to failed and adds it to the offline retry which sends the message when connection comes back.
+    func scheduleOfflineRetry(for messageId: MessageId, completion: @escaping (Result<ChatMessage, MessageRepositoryError>) -> Void) {
```
laevandus, line 93: MessageSender uses this to send queued messages directly to offline repo when it sees that there is connection issue.

Sources/StreamChat/Repositories/SyncRepository.swift
```diff
@@ -132,6 +132,11 @@ class SyncRepository {
         // Enter recovery mode so no other requests are triggered.
         apiClient.enterRecoveryMode()
 
+        // Run offline actions requests as the first thing
+        if config.isLocalStorageEnabled {
+            operations.append(ExecutePendingOfflineActions(offlineRequestsRepository: offlineRequestsRepository))
+        }
+
```
laevandus, line 139: Moved it to the first item on the syncing list so that we retry more quickly rather than waiting after other stuff.

nuno-vieira (approved):
LGTM! ✅

martinmitrevski (approved):

testableapple (approved):
LGTM 👍

PR #3325, opened 2024-07-22
Improve performance of `ChatChannel` database model conversions ~7 times

laevandus:
### 🔗 Issue Links

Related to [PBE-5004](https://stream-io.atlassian.net/browse/PBE-5004)

### 🎯 Goal

Improve performance of `ChatChannel` database model conversions by removing additional fetch requests

### 📝 Summary

- Roughly 7x CPU usage reduction when converting channel database models to immutable `ChatChannel`
- CPU time is saved by reducing additional fetch requests and using the DTOs data directly

### 🛠 Implementation

In #3299 we removed CoreDataLazy and this opened up an opportunity to optimise `asModel()` functions (these convert local database data to immutable models like `ChatChannel`). This PR replaces additional DB calls with accessing the NSManagedObject's data directly.

### 🎨 Showcase

Profiling steps: log out Chewbacca, log in with Chewbacca, wait until CPU usage settles. Observe the time spent in converting `ChannelDTO` to `ChatChannel` when the user is opening the channel list.

Measured: iPhone 12 Pro Max
279 ms > 38 ms 7x improvement

| Before | After |
| ------ | ----- |
|  Baseline   | Final |

### 🧪 Manual Testing Notes

Exploratory testing around variety of channel actions, e.g. send a message > does everything look allright.

### ☑️ Contributor Checklist

[PBE-5004]: https://stream-io.atlassian.net/browse/PBE-5004?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Sources/StreamChat/Database/DTOs/ChannelDTO.swift
```diff
@@ -446,6 +446,11 @@ extension ChatChannel {
 
             let allUnreadMessages = currentUserRead?.unreadMessagesCount ?? 0
 
+            // Therefore, no unread messages with mentions and we can skip the fetch
+            if allUnreadMessages == 0 {
+                return .noUnread
+            }
```
laevandus, line 452: Below we used to do an additional fetch request for unread messages mentioning the current user. That is not needed if there are no unread messages anyway.

Sources/StreamChat/Utils/Database/NSManagedObject+Extensions.swift
```diff
@@ -163,3 +163,25 @@ extension NSManagedObjectContext {
...
+        if let sortDescriptors = request.sortDescriptors, !sortDescriptors.isEmpty {
+            result.sort(using: sortDescriptors)
+        }
+        if request.fetchLimit > 0 {
+            let removeCount = result.count - request.fetchLimit
+            if removeCount > 0 {
+                result.removeObjects(in: NSRange(location: result.count - removeCount, length: removeCount))
+            }
+        }
+        return result as? [Element] ?? []
```
laevandus, line 185: Optimisation opportunity: first sort, then run filtering in batches until we have expected amount of objects.

Sources/StreamChat/Utils/Database/NSManagedObject+Extensions.swift
```diff
@@ -163,3 +163,25 @@ extension NSManagedObjectContext {
...
+    }
```
nuno-vieira, line 186: I think we can probably simplify this, and use just regular siltering and sorting like we did with Threads here: https://github.com/GetStream/stream-chat-swift/blob/develop/Sources/StreamChat/Database/DTOs/ThreadDTO.swift#L150-L152

This way it is also consistent with Threads. From now on, this kind of `lastReplies`, `lastMessages` etc... should be filtered and sorted at runtime when converting DTOs to modes as these values usually are very low and it has better performance than performing DB queries.

laevandus: For some of the things we have really complicated filtering. For example, channel messages has 10 predicates. Therefore, I went to reusing the NSPredicate approach because it needs to match with NSFetchRequests passed into DB observers.
But that said, some of the predicates can be skipped, for example, cid predicate because when accessing messages from the ChannelDTO, then I already know these belong to the current channel.

Sources/StreamChat/Utils/Database/NSManagedObject+Extensions.swift
laevandus: This gave a boost from 105 ms to 71 ms

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
```diff
@@ -471,35 +476,51 @@ extension ChatChannel {
...
-                )?
+            guard !dto.messages.isEmpty else { return nil }
+            let request = MessageDTO.lastMessageRequest(
+                from: currentUser.user.id,
+                in: dto.cid,
+                context: context
+            )
+            return try? dto.messages
+                .filtered(using: request)
+                .first?
```
nuno-vieira, line 501: We probably can simplify this. The backend returns the latestMessages for us. It looks like we are ignoring it and doing unnecessary query logic for this. Ideally, we just need to store the latestMessages in a new `MessageDTO.latestMessages` and then map it directly into `ChatChannel.latestMessages` but sort it by `createdAt`

Sources/StreamChat/Utils/Database/NSManagedObject+Extensions.swift
nuno-vieira: https://github.com/GetStream/stream-chat-swift/pull/3325/files#r1686787091

☝️ We should not need this complex predicate here, the backend already returns us the `latestMessages` we don't need to compute it locally. (We did it before but it is not really needed, it is just overengineered)

laevandus: Thank you for the excellent suggestion, this gave another boost, so now we are down to 55 ms from the original 266 ms

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
laevandus: Removed the request part and asking for this information from the `dto.messages` directly.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 # Upcoming
 
 ### 🔄 Changed
+- Improve performance of `ChatChannel` database model conversions more than 4 times [#3322](https://github.com/GetStream/stream-chat-swift/pull/3322)
```
nuno-vieira, line 7: You forgot to put the StreamChat header and also we can use the "Performance" subheader for performance improvements. We have done this in the past as well 👍 
```suggestion
## StreamChat
### ⚡ Performance
- Improve performance of `ChatChannel` database model conversions more than 4 times [#3322](https://github.com/GetStream/stream-chat-swift/pull/3322)
```

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
```diff
@@ -436,70 +436,76 @@ extension ChatChannel {
...
-                )
+            guard !dto.messages.isEmpty else { return [] }
+            let request = MessageDTO.channelMessagesRequest(
+                for: dto.cid,
+                limit: dto.managedObjectContext?.localCachingSettings?.chatChannel.latestMessagesLimit ?? 25,
+                deletedMessagesVisibility: dto.managedObjectContext?.deletedMessagesVisibility ?? .visibleForCurrentUser,
+                shouldShowShadowedMessages: dto.managedObjectContext?.shouldShowShadowedMessages ?? false
+            )
+            return dto.messages
+                .filtered(using: request)
```
nuno-vieira, line 469: I think this should be enough 🤔 THe logic of messages visibility etc is already applied on the `FetchResultsController`, the `ChatChannel.messages` AKA `latestMessages` should just return whatever the backend returns
```suggestion
            guard !dto.messages.isEmpty else { return [] }
            return dto.messages
                .sorted(by: { $0.createdAt.bridgeDate > $1.createdAt.bridgeDate })
```

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
```diff
@@ -436,70 +436,76 @@ extension ChatChannel {
...
-
-            return try? MessageDTO
-                .loadLastMessage(
-                    from: currentUser.user.id,
-                    in: dto.cid,
-                    context: context
-                )?
+            guard let currentUserId = context.currentUser?.user.id else { return nil }
+            return try? dto.messages
+                .sorted(by: { $0.createdAt.bridgeDate > $1.createdAt.bridgeDate })
```
nuno-vieira, line 476: We are doing this sorting in multiple places, we probably can just do it once and reuse it. Something like:
`let latestMessages = dto.messages.sorted(by: { $0.createdAt.bridgeDate > $1.createdAt.bridgeDate })`

And then use `latestMessages` everywhere

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
```diff
@@ -436,70 +436,76 @@ extension ChatChannel {
...
+        let watchers = dto.watchers
+            .sorted { lhs, rhs in
+                let lhsActivity = lhs.lastActivityAt?.bridgeDate ?? .distantPast
+                let rhsActivity = rhs.lastActivityAt?.bridgeDate ?? .distantPast
+                if lhsActivity == rhsActivity {
+                    return lhs.id > rhs.id
+                }
+                return lhsActivity > rhsActivity
+            }
+            .prefix(context.localCachingSettings?.chatChannel.lastActiveWatchersLimit ?? 100)
```
nuno-vieira, line 494: Nice one 👍

Sources/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -1309,21 +1323,28 @@ private extension ChatMessage {
...
-                MessageReactionDTO
-                    .loadReactions(ids: dto.ownReactions, context: context)
-                    .compactMap { try? $0.asModel() }
-            )
+            if !dto.ownReactions.isEmpty {
+                currentUserReactions = Set(
+                    MessageReactionDTO
+                        .loadReactions(ids: dto.ownReactions, context: context)
+                        .compactMap { try? $0.asModel() }
+                )
```
nuno-vieira, line 1331: Can't we simply do this?
```suggestion
                currentUserReactions = dto.ownReactions..compactMap { try? $0.asModel() }
```

Sources/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -1309,21 +1323,28 @@ private extension ChatMessage {
...
+            } else {
+                currentUserReactions = []
+            }
```
nuno-vieira, line 1334: Why was this introduced? 🤔

Sources/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -1309,21 +1323,28 @@ private extension ChatMessage {
...
-                .loadReactions(ids: dto.latestReactions, context: context)
-                .compactMap { try? $0.asModel() }
-        )
+        latestReactions = {
+            guard !dto.latestReactions.isEmpty else { return Set() }
+            return Set(
+                MessageReactionDTO
+                    .loadReactions(ids: dto.latestReactions, context: context)
+                    .compactMap { try? $0.asModel() }
+            )
```
nuno-vieira, line 1346: Same here?
```suggestion
            return dto.latestReactions.compactMap { try? $0.asModel() }
```

Sources/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -1337,7 +1358,7 @@ private extension ChatMessage {
             .sorted { $0.id.index < $1.id.index }
 
         latestReplies = {
-            guard !dto.replies.isEmpty else { return [] }
+            guard dto.replyCount > 0 else { return [] }
```
nuno-vieira, line 1361: We can also do the same optimization of the latest messages, with latestReplies:
```suggestion
            guard dto.replyCount > 0 else { return [] }
            return dto.replies.sorted.map()
```

Sources/StreamChat/Utils/Database/NSManagedObject+Extensions.swift
```diff
@@ -163,3 +163,42 @@ extension NSManagedObjectContext {
...
+            }
+            allObjects = result
+        } else {
+            if let predicate = request.predicate {
+                allObjects.filter(using: predicate)
+            }
+        }
+        
+        return allObjects as? [Element] ?? []
+    }
```
nuno-vieira, line 203: I would really like to remove this, all the fetch request logic should only be used on the `FetchResultsController`, the Immutable Models, should just get data from the backend, and not have predicates

nuno-vieira (commented):
It is looking good, but just added a couple more suggestions 👍

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
laevandus: If we would be OK with this returning slightly different messages sometimes then I am happy to remove the fetch request from here (e.g contains deleted message when message list would filter them out). 
Currently it is 5 last messages.
From other SDK's point of view it makes sense as well.

Sources/StreamChat/Database/DTOs/MessageDTO.swift
laevandus: Unfortunately these are ids, not DTOs so the fetch request needs to stay for now.

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
nuno-vieira: Yes, no other SDK is doing this "Smart" logic on the `ChatChannel.messages` property. This data should be dummy and just represent what the server gives 👍 We had this for ages now, but it is about time to change this, since we are now doing all these improvements 🚀

Sources/StreamChat/Utils/Database/NSManagedObject+Extensions.swift
laevandus: If we go for removing the messages fetch, then I am happy to delete this. Will look into this tomorrow, but I feel like it makes sense.

Sources/StreamChat/Database/DTOs/MessageDTO.swift
nuno-vieira: Ah I see 👍

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
laevandus: Sounds good to me. It will probably give another boost.

Sources/StreamChat/Database/DTOs/MessageDTO.swift
nuno-vieira: I guess same problem here with the Ids right?

Sources/StreamChat/Database/DTOs/MessageDTO.swift
laevandus: Skipping the fetch request when no ownReaction ids are present

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
```diff
@@ -434,72 +434,77 @@ extension ChatChannel {
...
+            var messages = sortedMessageDTOs
+                .prefix(dto.managedObjectContext?.localCachingSettings?.chatChannel.latestMessagesLimit ?? 25)
                 .compactMap { try? $0.relationshipAsModel(depth: depth) }
+            if let oldest = dto.oldestMessageAt?.bridgeDate {
+                messages = messages.filter { $0.createdAt >= oldest }
+            }
+            if let truncated = dto.truncatedAt?.bridgeDate {
+                messages = messages.filter { $0.createdAt >= truncated }
+            }
+            return messages
```
laevandus, line 470: We have some tests for these, so I kept this part to match what tests are expecting. Ideally we should add latestMessages to the ChannelDTO but I would prefer to do so in a separate PR.

nuno-vieira (approved):
LGTM! ✅ 🚀 Really happy to see we are cleaning up these things 🚀

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 # Upcoming
 
 ## StreamChat
+### ⚡ Performance
+- Improve performance of `ChatChannel` database model conversions more than 4 times [#3322](https://github.com/GetStream/stream-chat-swift/pull/3322)
```
nuno-vieira, line 8: Now it is 7 times 😋 
```suggestion
- Improve performance of `ChatChannel` database model conversions more than 7 times [#3322](https://github.com/GetStream/stream-chat-swift/pull/3322)
```

martinmitrevski (approved):
Great work 👏

PR #3339, opened 2024-07-29
[CI] Comment SDK size on every commit

testableapple:
### 🔗 Issue Links

Resolve https://stream-io.atlassian.net/browse/PBE-4314

### 📺 Showcase

Screenshot 2024-07-29 at 1 22 06 PM

### DIFF
--- a/.github/workflows/sdk-performance-metrics.yml
+++ b/.github/workflows/sdk-performance-metrics.yml
@@ -1,4 +1,4 @@
-name: Metrics
+name: SDK Performance
 
 on:
   schedule:
@@ -17,7 +17,7 @@ env:
 
 jobs:
   performance:
-    name: Performance
+    name: Metrics
     runs-on: macos-14
     env:
       GITHUB_TOKEN: '${{ secrets.CI_BOT_GITHUB_TOKEN }}'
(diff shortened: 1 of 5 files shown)


.github/workflows/sdk-performance-metrics.yml
```diff
@@ -1,4 +1,4 @@
-name: Metrics
+name: SDK Performance
```
testableapple, line 1: That's because on the `Actions` tab it was not possible to determine where was the `Performance` and where was the `Size` metrics.

fastlane/Fastfile
```diff
@@ -479,26 +480,9 @@ lane :xcmetrics do |options|
...
-
-    Dir.chdir(File.dirname(xcmetrics_path)) do
-      if sh('git status -s', log: false).to_s.empty?
-        pr_comment_required = false
-        UI.important('No changes in performance benchmarks. Skipping commit and comment.')
-      else
-        sh('git add -A')
-        sh("git commit -m 'Update #{xcmetrics_path}: #{current_branch}'")
-        sh('git push')
-      end
```
testableapple, line 495: This is to eliminate the useless branch metrics in the baseline json file. Basically now performance metrics will keep only the benchmark, similar to size metrics.

nuno-vieira (approved):
Great! 👌

PR #3345, opened 2024-07-30
Fix message search not showing results for new search terms

laevandus:
### 🎯 Goal

Message search results were empty after logging in and quickly typing a search term 

### 📝 Summary

- Reload collection view on state change and update UI state on messages array change

### 🧪 Manual Testing Notes

1. Logout and login for clearing local DB
2. Quickly type a search term like "tes"

### ☑️ Contributor Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Fix a rare issue with incorrect message order when sending multiple messages while offline [#3316](https://github.com/GetStream/stream-chat-swift/issues/3316)
 - Fix sorting channel list by unread count [#3340](https://github.com/GetStream/stream-chat-swift/pull/3340)
 
+## StreamChatUI
+### 🐞 Fixed
+- Fix message search not showing results for new search terms [#3345](https://github.com/GetStream/stream-chat-swift/pull/3345)
+
 # [4.60.0](https://github.com/GetStream/stream-chat-swift/releases/tag/4.60.0)
 _July 18, 2024_
 
--- a/Sources/StreamChatUI/ChatChannelList/Search/ChatMessageSearchVC.swift
+++ b/Sources/StreamChatUI/ChatChannelList/Search/ChatMessageSearchVC.swift
@@ -31,7 +31,7 @@ open class ChatMessageSearchVC: ChatChannelListSearchVC, ChatMessageSearchContro
     // MARK: - ChatChannelListSearchVC Abstract Implementations
 
     override open var hasEmptyResults: Bool {
-        messages.isEmpty
+        messageSearchController.messages.isEmpty
     }
 
     override open func loadSearchResults(with text: String) {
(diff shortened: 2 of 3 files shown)


Sources/StreamChatUI/ChatChannelList/Search/ChatMessageSearchVC.swift
```diff
@@ -10,7 +10,11 @@ import UIKit
 @available(iOSApplicationExtension, unavailable)
 open class ChatMessageSearchVC: ChatChannelListSearchVC, ChatMessageSearchControllerDelegate {
     /// The data of the message list.
-    public private(set) var messages: [ChatMessage] = []
+    public private(set) var messages: [ChatMessage] = [] {
+        didSet {
+            handleStateChanges(messageSearchController.state)
+        }
+    }
```
laevandus, line 17: Internally it needs messages to be set before

Sources/StreamChatUI/ChatChannelList/Search/ChatMessageSearchVC.swift
```diff
@@ -105,6 +109,10 @@ open class ChatMessageSearchVC: ChatChannelListSearchVC, ChatMessageSearchContro
 
     // MARK: - ChatMessageSearchControllerDelegate
 
+    override open func controller(_ controller: DataController, didChangeState state: DataController.State) {
+        reloadMessages()
+    }
```
laevandus, line 114: It is not super great because this class is a subclass of a channel list VC but here we are dealing with messages instead. Therefore, we need to override superclass implementation which does not know about triggering reloading messages.

testableapple (approved):

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### 🐞 Fixed
 - Fix a rare issue with incorrect message order when sending multiple messages while offline [#3316](https://github.com/GetStream/stream-chat-swift/issues/3316)
 - Fix sorting channel list by unread count [#3340](https://github.com/GetStream/stream-chat-swift/pull/3340)
+- Fix message search not showing results for new search terms [#3345](https://github.com/GetStream/stream-chat-swift/pull/3345)
```
nuno-vieira, line 16: This should be in the # StreamChatUI section

Sources/StreamChatUI/ChatChannelList/Search/ChatMessageSearchVC.swift
nuno-vieira: How is this not a problem on the Channel Search? 🤔

nuno-vieira, line 114: wait, this will break the state logic right? So, the loading views etc won't appear?

laevandus: I added `didSet` to `messages` which then calls the validation.

PR #3355, opened 2024-08-01
Fix rare crashes when deleting local database content on logout

laevandus:
### 🔗 Issue Links

Resolves [PBE-4309](https://stream-io.atlassian.net/browse/PBE-4309)

### 🎯 Goal

Fix crashes on logout related to local database access

### 📝 Summary

- Notify `NSManagedObjectContext` that data was [deleted on the SQL level](https://developer.apple.com/documentation/coredata/nsbatchdeleterequest)
- Discard any `DatabaseContainer` write requests while clearing local data

### 🛠 Implementation

Batch delete removes data on the SQL level and does not notify contexts. Contexts can be left into invalid state if we do not manually notify them.

### 🧪 Manual Testing Notes

Repeat
1. Log in
2. Do something
3. Log out

No crashes or any other odd things should appear.

### ☑️ Contributor Checklist

[PBE-4309]: https://stream-io.atlassian.net/browse/PBE-4309?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/StreamChat/Database/DatabaseContainer.swift
+++ b/Sources/StreamChat/Database/DatabaseContainer.swift
@@ -77,6 +77,7 @@ class DatabaseContainer: NSPersistentContainer {
         return context
     }()
 
+    private var canWriteData = true
     private var stateLayerContextRefreshObservers = [NSObjectProtocol]()
     private var loggerNotificationObserver: NSObjectProtocol?
     private let localCachingSettings: ChatClientConfig.LocalCaching?
@@ -217,6 +218,12 @@ class DatabaseContainer: NSPersistentContainer {
     func write(_ actions: @escaping (DatabaseSession) throws -> Void, completion: @escaping (Error?) -> Void) {
         writableContext.perform {
             log.debug("Starting a database session.", subsystems: .database)
+            guard self.canWriteData else {
+                log.debug("Discarding write attempt.", subsystems: .database)
+                completion(nil)
+                return
+            }
+            
             do {
                 FetchCache.clear()
                 try actions(self.writableContext)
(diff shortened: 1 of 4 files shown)


Sources/StreamChat/Database/DatabaseContainer.swift
```diff
@@ -217,6 +218,12 @@ class DatabaseContainer: NSPersistentContainer {
     func write(_ actions: @escaping (DatabaseSession) throws -> Void, completion: @escaping (Error?) -> Void) {
         writableContext.perform {
             log.debug("Starting a database session.", subsystems: .database)
+            guard self.canWriteData else {
+                log.debug("Discarding write attempt.", subsystems: .database)
+                completion(nil)
+                return
+            }
```
laevandus, line 225: Added a safeguard here for cases where batch delete is in process and some delayed action triggers database write.

Tests/StreamChatTests/Database/DatabaseContainer_Tests.swift
```diff
@@ -193,6 +119,45 @@ final class DatabaseContainer_Tests: XCTestCase {
...
+        let expectation = XCTestExpectation(description: "Remove")
+        container.removeAllData { error in
+            XCTAssertNil(error)
+            expectation.fulfill()
+        }
+        
+        // Save just after triggering remove all
+        container.write { session in
+            try session.saveChannel(payload: self.dummyPayload(with: .unique), query: nil, cache: nil)
+        }
```
laevandus, line 141: Without `canWriteData` this would cause a test failure in this test (but sometimes, because this depends on the timing. I got test failures when I repeated the test 100 times)

Tests/StreamChatTests/Database/DatabaseContainer_Tests.swift
```diff
@@ -365,4 +330,84 @@ final class DatabaseContainer_Tests: XCTestCase {
             XCTAssertEqual(database.backgroundReadOnlyContext.shouldShowShadowedMessages, shouldShowShadowedMessages)
         }
     }
+    
+    // MARK: -
+    
+    private func writeALotOfData(to container: DatabaseContainer) throws {
```
nuno-vieira, line 336: Maybe this name is more clear, because it is not just a lot of data, it needs to be all entities
```suggestion
    private func writeDataForAllEntities(to container: DatabaseContainer) throws {
```

Tests/StreamChatTests/Database/DatabaseContainer_Tests.swift
```diff
@@ -193,6 +119,45 @@ final class DatabaseContainer_Tests: XCTestCase {
             }
         }
     }
+    
+    func test_removingAllData_whileAnotherWrite() throws {
```
nuno-vieira, line 123: Can we maybe add a test or try in our Demo App to logout while there is another controller reference alive and then access that controller's data? 🤔 It would be interesting to see if it previously crashed, and now it does not. Although it might be tricky to actually reproduce this 🤔

nuno-vieira (changes requested):
LGTM ✅ Just added a couple suggestions, but nothing blocking 👍

nuno-vieira (approved):

Tests/StreamChatTests/Database/DatabaseContainer_Tests.swift
laevandus: I think I should be able to create a test for this.

1. Create controller
2. Write some data
3. Call remove all
4. When remove is done, accessing controller content

laevandus: Added a test what fails without current changes.

testableapple (approved):

PR #3365, opened 2024-08-07
Improve database container read and reset

laevandus:
### 🔗 Issue Links

Resolves [PBE-5422](https://stream-io.atlassian.net/browse/PBE-5422)

### 🎯 Goal

Fix issues around database read wrappers and context resets

### 📝 Summary

- One DatabaseContainer read uses DatabaseSession, the async one NSManagedObjectContext > unify to use DatabaseSession
- Fix and issue where after logout contexts are not reset (keep referencing deleted objects)
- Change the read assert to trigger only based on counts (did not find a way how to fix the issue of it keeping tracking deleted objects)

### 🧪 Manual Testing Notes

Case 1:
Log in and log out

Case 2:
1. Open Demo Configuration Screen
2. Enable Hard Delete Message
3. Open a channel
4. Hard delete a message
5. Go to background
6. Wait 1min ( or as soon as web-socket's disconnect is logged)
7. Open the app again
8. :boom: 

### ☑️ Contributor Checklist

[PBE-5422]: https://stream-io.atlassian.net/browse/PBE-5422?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/StreamChat/Database/DTOs/CurrentUserDTO.swift
+++ b/Sources/StreamChat/Database/DTOs/CurrentUserDTO.swift
@@ -203,15 +203,16 @@ extension CurrentChatUser {
     fileprivate static func create(fromDTO dto: CurrentUserDTO) throws -> CurrentChatUser {
         let user = dto.user
 
-        let extraData: [String: RawJSON]
-        do {
-            extraData = try JSONDecoder.default.decode([String: RawJSON].self, from: dto.user.extraData)
-        } catch {
-            log.error(
-                "Failed to decode extra data for user with id: <\(dto.user.id)>, using default value instead. "
-                    + "Error: \(error)"
-            )
-            extraData = [:]
+        var extraData = [String: RawJSON]()
+        if !dto.user.extraData.isEmpty {
+            do {
+                extraData = try JSONDecoder.default.decode([String: RawJSON].self, from: dto.user.extraData)
+            } catch {
+                log.error(
+                    "Failed to decode extra data for user with id: <\(dto.user.id)>, using default value instead. "
+                        + "Error: \(error)"
+                )
+            }
         }
         
         let mutedUsers: [ChatUser] = try dto.mutedUsers.map { try $0.asModel() }
--- a/Sources/StreamChat/Database/DTOs/MessageDTO.swift
+++ b/Sources/StreamChat/Database/DTOs/MessageDTO.swift
@@ -1154,6 +1154,53 @@ extension NSManagedObjectContext: MessageDatabaseSession {
             $0.localState = .pendingUpload
         }
     }
+    
+    func loadMessage(
+        before id: MessageId,
+        cid: String
+    ) throws -> MessageDTO? {
+        try MessageDTO.loadMessage(
+            before: id,
+            cid: cid,
+            deletedMessagesVisibility: deletedMessagesVisibility ?? .alwaysVisible,
+            shouldShowShadowedMessages: shouldShowShadowedMessages ?? true,
+            context: self
+        )
+    }
+    
+    func loadMessages(
+        from fromIncludingDate: Date,
+        to toIncludingDate: Date,
+        in cid: ChannelId,
+        sortAscending: Bool
+    ) throws -> [MessageDTO] {
+        try MessageDTO.loadMessages(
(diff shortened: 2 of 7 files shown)


Sources/StreamChat/Database/DTOs/CurrentUserDTO.swift
```diff
@@ -203,15 +203,16 @@ extension CurrentChatUser {
...
-        do {
-            extraData = try JSONDecoder.default.decode([String: RawJSON].self, from: dto.user.extraData)
-        } catch {
-            log.error(
-                "Failed to decode extra data for user with id: <\(dto.user.id)>, using default value instead. "
-                    + "Error: \(error)"
-            )
-            extraData = [:]
+        var extraData = [String: RawJSON]()
+        if !dto.user.extraData.isEmpty {
```
laevandus, line 207: If data is empty, error is logged which can happen when trying to convert deleted DTO

Sources/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -1154,6 +1154,53 @@ extension NSManagedObjectContext: MessageDatabaseSession {
             $0.localState = .pendingUpload
         }
     }
+    
+    func loadMessage(
```
laevandus, line 1158: Moved these under DatabaseSession so that I can use DatabaseSession in async version of the read()

Sources/StreamChat/Database/DatabaseContainer.swift
```diff
@@ -253,13 +253,29 @@ class DatabaseContainer: NSPersistentContainer {
...
+        completion: @escaping (Result<T, Error>) -> Void
+    ) {
         context.perform {
             do {
+                let changeCounts = context.currentChangeCounts()
                 let results = try actions(context)
-                if context.hasChanges {
-                    assertionFailure("Background context is read only, but calling actions() created changes")
+                if changeCounts != context.currentChangeCounts() {
+                    assertionFailure("Context is read only, but actions created changes: (updated=\(context.updatedObjects), inserted=\(context.insertedObjects), deleted=\(context.deletedObjects)")
```
laevandus, line 278: Did not figure out why deleted message is kept tracked by the context which triggered this assert. Just changed it around so that we only trigger it when actually the actions closure changed something.

Sources/StreamChat/Database/DatabaseContainer.swift
```diff
@@ -332,11 +323,24 @@ class DatabaseContainer: NSPersistentContainer {
...
+                // Merging changes triggers DB observers to react to deletions
                 NSManagedObjectContext.mergeChanges(
                     fromRemoteContextSave: [NSDeletedObjectsKey: deletedObjectIds],
                     into: contexts
                 )
             }
+            // Finally reset states of all the contexts after batch delete and deletion propagation.
+            if let writableContext = self?.writableContext, let allContext = self?.allContext {
+                writableContext.invalidateCurrentUserCache()
+                writableContext.reset()
```
nuno-vieira, line 335: Did you also tried reseting these context when a delete is performed on `write` function? (Not sure if this will do anything tho)

nuno-vieira (approved):
LGTM! ✅

Sources/StreamChat/Database/DatabaseContainer.swift
laevandus: Did not test that. Slightly afraid to do it because we use these contexts for DB observers. Thinking about letting it be 😄

nuno-vieira: Yeah probably better 👍

nuno-vieira: Out of curiosity, I tested it and it works. Maybe if we still have problems with this in the future we can go back here and apply the following change before the `CatabaseContainer.write` :
```swift
if !self.backgroundReadOnlyContext.deletedObjects.isEmpty {
      self.backgroundReadOnlyContext.reset()
 }
```

testableapple (approved):
LGTM 👍

Issue #2021, opened 2022-05-24
Offline mode resets channels list


pelczark:
## What did you do?
In my app user has over 100 channels with a single user. I'm using the default UI from StreamChatUI.
Channels list is paginated. I scroll down to the bottom to see f.e. 80th channel.
I encountered two issues:
1. When the internet connection is lost and then is reconnected again, then I try to load more pages, the channel list scrolls up, and I see only the first channels page (20 channels).
~~2. When I open the 80th channel details screen, turn on and off flight mode, and try to send a new message the red indicator shows up. I long press the message and press "Try again" without success.~~

## What did you expect to happen?
1. The channels list shouldn't scroll up and user should be able to see the 80th channel.
~~2. User should be able to send a message when connection is regained.~~

## What happened instead?
1. Channels list scrolls up, only first 20 channels are visible.
~~2. The message cannot be sent.~~

## GetStream Environment
**GetStream Chat version:** 4.15.0
**GetStream Chat frameworks:** StreamChat, StreamChatUI
**iOS version:** 15.4.1
**Swift version:** 5
**Xcode version:** 13.3.1
**Device:** iPhone 11 Pro

polqf:
HI @pelczark,

Thank you for opening this ticket.

The first issue is known, and it is currently not a priority. We'll let you know when this changes.

The second one is unknown, and should not happen. Let me come back to you after I investigate it. Is there any particular thing you are doing to reproduce it?

hugobernalstream:
Hi @pelczark,

After a first investigation of 2. with the details you provided, I wasn't able to reproduce the issue you mention. I was able to properly resend the message after turning airplane mode off, regaining connection and long-pressing the failed message to resend it.

Is it possible for you to send a video reproducing the issue? 
Have you tried to reproduce the issue in our DemoApp?
Is there any customization you are adding around ChatChannel?

pelczark:
Hi. I retested `2` using the newest library and it is working. Apologies, It didn't work in the older version.

polqf:
That's great to hear! We'll keep you posted on `1` as it is not in our roadmap now

nuno-vieira:
Closing this issue since it is not reproducible any more in the latest versions of the SDK.

Let us know if you have any further issues.

Best,
Nuno

nuno-vieira:
Hello!

[...]

PR #3367, opened 2024-08-09
Run offline state sync in the background

laevandus:
### 🔗 Issue Links

Resolves [PBE-4293](https://stream-io.atlassian.net/browse/PBE-4293)

Resolves [stream-chat-swift/issues/2021](https://github.com/GetStream/stream-chat-swift/issues/2021)

### 🎯 Goal

Allow regular API calls to be made while offline sync is running which gives quicker UI updates when UI update requires API call to finish.

### 📝 Summary

- Add a sync v2 implementation controllable with a `StreamRuntimeCheck._isSyncV2Enabled`

### 🛠 Implementation

Adds a new version of the offline state sync which only pauses regular API requests when handling automatic retries. The sync is more lightweight and only syncs currently active channels (active = server state has been requested in this session aka calling synchronize). If sync is not possible (too many events), then data is fully refreshed for active channel lists (same as synchronize but not resetting to the first page). 

#### Improvements
- Does not block other API requests if sync takes a bit of time (in debug and up-to-date data it runs <0.5 seconds)
- Less data exchanges as it only syncs active channels (for other channels we use controller synchronize or get methods anyway)

#### Notes
- These changes do not touch the code (direct call to `syncExistingChannelsEvents`) which is triggered by the `ChatRemoteNotificationHandler`. This code path triggers the `request.fetchLimit = 100` case which the regular sync actually does not. We should look this in a separate task.
- I kept refreshing channel lists in V2 because UI-tests were relying on these (and probably some other hidden cases as well). Could be changed in next iterations.

#### Summary of the V2 mode
*Recovery mode (pauses regular API requests while it is running)*
1. Enter recovery
2. Runs offline API requests
3. Exit recovery

*Background mode (other regular API requests are allowed to run at the same time)*
1. Collect all the **active** channel ids (from instances of `Chat`, `ChannelList`, `ChatChannelController`, `ChatChannelListController`)
2. Ask updates from the /sync endpoint for these channels
2.a Apply changes
2.b When there are too many events, just refresh channel lists (channels for current pages in `ChannelList`, `ChatChannelListController`)

#### Comparison

| Description | V1 | V2 |
|--------|--------|--------|
| Channel list sync | All in the local storage | Only active channel lists |
| Resets channel list to the first page | Yes, always | No, refreshes currently loaded channels, UI does not just when data changes |
| Always starts watching channels | Yes | No |
| Deletes channels | Yes, when reseting channel list and channel not referenced by active instances | No |
| Blocks regular API requests | Yes | Only when retrying offline requests, otherwise no | 

#### Case 1: Channel list updates
1. Log in, go offline
2. Other user triggers channel list updates (sends new messages etc)
3. Come back online, channel list updates

#### Case 2: Sending messages while offline
1. Log in, go offline
2. Send some messages, edit some, add reactions etc
3. Come back online, messages are sent and edits are visible for other participants

#### Case 3: Not resetting channel list (https://github.com/GetStream/stream-chat-swift/issues/2021)
1. Log in
2. Scroll down in the channel list to load multiple pages
3. Go offline
4. Trigger updates using a different account (for verifying that channel list updates when coming back online)
5. Come back online. Channel list does not reset to the first page.

### ☑️ Contributor Checklist

[PBE-4293]: https://stream-io.atlassian.net/browse/PBE-4293?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/StreamChat/Config/StreamRuntimeCheck.swift
+++ b/Sources/StreamChat/Config/StreamRuntimeCheck.swift
@@ -31,4 +31,9 @@ public enum StreamRuntimeCheck {
     ///
     /// Enables reusing unchanged converted items in database observers.
     public static var _isDatabaseObserverItemReusingEnabled = true
+    
+    /// For *internal use* only
+    ///
+    /// Uses vresion 2 for offline state sync.
+    public static var _isSyncV2Enabled = true
 }
--- a/Sources/StreamChat/Config/StreamRuntimeCheck.swift
+++ b/Sources/StreamChat/Config/StreamRuntimeCheck.swift
@@ -34,6 +34,6 @@ public enum StreamRuntimeCheck {
     
     /// For *internal use* only
     ///
-    /// Uses vresion 2 for offline state sync.
+    /// Uses version 2 for offline state sync.
     public static var _isSyncV2Enabled = true
 }
--- a/Sources/StreamChat/Repositories/SyncRepository.swift
+++ b/Sources/StreamChat/Repositories/SyncRepository.swift
@@ -156,9 +156,8 @@ class SyncRepository {
     ///
     /// Background mode (other regular API requests are allowed to run at the same time)
     /// 1. Collect all the **active** channel ids (from instances of `Chat`, `ChannelList`, `ChatChannelController`, `ChatChannelListController`)
-    /// 2. Ask updates from the /sync endpoint for these channels
-    /// 2.a Apply changes
-    /// 2.b When there are too many events, just refresh channel lists (channels for current pages in `ChannelList`, `ChatChannelListController`)
+    /// 2. Apply updates from the /sync endpoint for these channels
+    /// 3. Refresh channel lists (channels for current pages in `ChannelList`, `ChatChannelListController`)
     private func syncLocalStateV2(lastSyncAt: Date, completion: @escaping () -> Void) {
         let context = SyncContext(lastSyncAt: lastSyncAt)
         var operations: [Operation] = []
@@ -186,7 +185,7 @@ class SyncRepository {
         // 2. /sync
         operations.append(SyncEventsOperation(syncRepository: self, context: context, recovery: false))
         
-        // 2.b.1 (these run only if sync was not possible in the previous step)
+        // 3. Refresh channel lists (required even after applying events)
         operations.append(contentsOf: activeChannelLists.allObjects.map { RefreshChannelListOperation(channelList: $0, context: context) })
         operations.append(contentsOf: activeChannelListControllers.allObjects.map { RefreshChannelListOperation(controller: $0, context: context) })
         
--- a/Sources/StreamChat/Repositories/SyncRepository.swift
+++ b/Sources/StreamChat/Repositories/SyncRepository.swift
@@ -36,25 +36,25 @@ class SyncRepository {
     private let database: DatabaseContainer
     private let apiClient: APIClient
     private let channelListUpdater: ChannelListUpdater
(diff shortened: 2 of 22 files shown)


nuno-vieira (approved):
LGTM! ✅ Nice that this is behind a flag so that we can minimize the risk on this one 👍

Sources/StreamChat/Config/StreamRuntimeCheck.swift
```diff
@@ -31,4 +31,9 @@ public enum StreamRuntimeCheck {
     ///
     /// Enables reusing unchanged converted items in database observers.
     public static var _isDatabaseObserverItemReusingEnabled = true
+    
+    /// For *internal use* only
+    ///
+    /// Uses version 2 for offline state sync.
+    public static var _isSyncV2Enabled = true
```
martinmitrevski, line 38: We should also have a ticket to remove this in 2-3 releases and only support V2.

Sources/StreamChat/Repositories/SyncRepository.swift
```diff
@@ -108,10 +137,76 @@ class SyncRepository {
...
+        var operations: [Operation] = []
+        let start = CFAbsoluteTimeGetCurrent()
+        log.info("Starting to refresh offline state", subsystems: .offlineSupport)
+        
+        //
+        // Recovery mode operations (other API requests are paused)
+        //
+        if config.isLocalStorageEnabled {
+            apiClient.enterRecoveryMode()
+            operations.append(ExecutePendingOfflineActions(offlineRequestsRepository: offlineRequestsRepository))
```
martinmitrevski, line 172: I assume there are no changes in this logic, right? We have some logic for sending new messages - after some interval they shouldn't be sent when you come online. Just making sure we still have that.

Sources/StreamChat/Repositories/SyncRepository.swift
```diff
@@ -108,10 +137,76 @@ class SyncRepository {
...
+    /// Recovery mode (pauses regular API requests while it is running)
+    /// 1. Enter recovery
+    /// 2. Runs offline API requests
+    /// 3. Exit recovery
+    ///
+    /// Background mode (other regular API requests are allowed to run at the same time)
+    /// 1. Collect all the **active** channel ids (from instances of `Chat`, `ChannelList`, `ChatChannelController`, `ChatChannelListController`)
+    /// 2. Apply updates from the /sync endpoint for these channels
+    /// 3. Refresh channel lists (channels for current pages in `ChannelList`, `ChatChannelListController`)
+    private func syncLocalStateV2(lastSyncAt: Date, completion: @escaping () -> Void) {
```
martinmitrevski, line 161: We also have some internal docs about the sync mechanism, that probably require revisiting.

martinmitrevski (approved):
Looks good to me ✅  We should do solid testing here @testableapple

testableapple (approved):

Sources/StreamChat/Config/StreamRuntimeCheck.swift
laevandus: Added a [ticket](https://stream-io.atlassian.net/browse/PBE-5533)

Sources/StreamChat/Repositories/SyncRepository.swift
laevandus: It is still there, that logic is in `OfflineRequestsRepository` what did not change.

PR #3372, opened 2024-08-12
High CPU usage while typing

laevandus:
### 🔗 Issue Links

Resolves [PBE-5497](https://stream-io.atlassian.net/browse/PBE-5497)

### 🎯 Goal

Reduce CPU usage caused by recreating `CurrentChatUserController` when sending typing events

### 📝 Summary

* Add `ChatClient.UserSession` for reusing objects from connect to logout
* User the shared `CurrentChatUserController` for reading privacy settings 

### 🛠 Implementation

We check if we should send typing events using an internal property in the `ChatChannelController` which created a new instance of the `CurrentChatUserController`. This is called often when typing to a channel. This led to high CPU usage because controller is created and the local database is read on every character change. If the current user has, for example, a lot of muted channels, then recreating `CurrentChatUser` gets really expensive.
```swift
internal var shouldSendTypingEvents: Bool {
  let currentUserPrivacySettings = client.currentUserController().currentUser?.privacySettings
  …
}      
```

### 🎨 Showcase

iPhone 12 Pro Max (low power mode was on)
An account with 10 muted channels (important). I used Chewbacca.
1. Open a channel
2. Type 10 characters 
3. Investigate main thread usage, especially `shouldSendTypingEvents`

| Before | After |
| ------ | ----- |
| 666 ms | 56 ms |
| ------ | ----- |
|  Before   |  After  |

`shouldSendTypingEvents` does not spend ~66 ms per character change just for checking this property. Instead, there is a single ~56 ms cost once (or almost none if the shared controller had been already created. Currently it is created on demand.)

### 🧪 Manual Testing Notes

- Try typing events and make sure these appear as expected.
- Log in and log out, trigger more typing events

### ☑️ Contributor Checklist

[PBE-5497]: https://stream-io.atlassian.net/browse/PBE-5497?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
+++ b/Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
@@ -147,7 +147,7 @@ public class ChatChannelController: DataController, DelegateCallable, DataStoreP
     /// It is `true` if the channel typing events are enabled as well as the user privacy settings.
     internal var shouldSendTypingEvents: Bool {
         /// Ignore if user typing indicators privacy settings are disabled. By default, they are enabled.
-        let currentUserPrivacySettings = client.currentUserController().currentUser?.privacySettings
+        let currentUserPrivacySettings = client.sharedCurrentUserController.currentUser?.privacySettings
         let isTypingIndicatorsForCurrentUserEnabled = currentUserPrivacySettings?.typingIndicators?.enabled ?? true
         let isChannelTypingEventsEnabled = channel?.canSendTypingEvents ?? true
         return isTypingIndicatorsForCurrentUserEnabled && isChannelTypingEventsEnabled
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## StreamChat
 ### ⚡ Performance
 - Use background database observers for `CurrentUserController.currentUser`, `ChatChannelMemberListController.members`, and `ChatMessageController.message` which reduces CPU usage on the main thread [#3357](https://github.com/GetStream/stream-chat-swift/pull/3357)
+- `CurrentChatUserController` was often recreated when sending typing events [#3372](https://github.com/GetStream/stream-chat-swift/pull/3372)
 ### 🐞 Fixed
 - Fix rare crashes when deleting local database content on logout [#3355](https://github.com/GetStream/stream-chat-swift/pull/3355)
 - Fix rare crashes in `MulticastDelegate` when accessing it concurrently [#3361](https://github.com/GetStream/stream-chat-swift/pull/3361)
--- a/Sources/StreamChat/ChatClient.swift
+++ b/Sources/StreamChat/ChatClient.swift
@@ -51,6 +51,9 @@ public class ChatClient {
 
     /// The notification center used to send and receive notifications about incoming events.
     private(set) var eventNotificationCenter: EventNotificationCenter
+    
+    private var _sharedCurrentUserController: CurrentChatUserController?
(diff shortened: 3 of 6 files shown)


Sources/StreamChat/ChatClient+UserSession.swift
```diff
@@ -0,0 +1,42 @@
...
+    ///
+    /// For example, creating a current user controller for reading a single
+    /// property is expensive if this code path is called often. This is because
+    /// creating a controller leads to database fetch of the model.
+    ///
+    /// ```swift
+    /// if client.currentUserController().currentUser?.privacySettings { …
+    /// ```
+    ///
+    final class UserSession {
```
laevandus, line 20: Option 2: No `UserSession` class and put everything into `ChatClient` directly. Maybe too over-engineered based on the single need.

WDYT? This or option 2?

martinmitrevski: I would prefer option 2 tbh, but up to you.

martinmitrevski (approved):
LGTM ✅

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
```diff
@@ -147,7 +147,7 @@ public class ChatChannelController: DataController, DelegateCallable, DataStoreP
     /// It is `true` if the channel typing events are enabled as well as the user privacy settings.
     internal var shouldSendTypingEvents: Bool {
         /// Ignore if user typing indicators privacy settings are disabled. By default, they are enabled.
-        let currentUserPrivacySettings = client.currentUserController().currentUser?.privacySettings
+        let currentUserPrivacySettings = client.sharedCurrentUserController.currentUser?.privacySettings
```
nuno-vieira, line 150: TBH, I don't think this is really needed 🤔  Having the `ChannelController` holding a `currentUserController` should be enough 🤔 WDYT?

nuno-vieira, line 150: When creating the `ChannelController,` we can probably also create a currentUserController and maintain a reference to it. 🤔 
```suggestion
        let currentUserPrivacySettings = self.currentUserController.currentUser?.privacySettings
```

Sources/StreamChat/ChatClient+UserSession.swift
nuno-vieira: I would actually go to option 3 😄  Which is just keep a ref of the current user controller in the `ChatChannelController`

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
laevandus: I considered this but then there is current user controller per channel controller (if user typed in that controller) which is not as efficient than limiting it to a single shared instance. Currently I went for more complex but more efficient solution. 
Slightly afraid that changes in muted channels triggers too many updates if multiple current user controllers are alive.

nuno-vieira (approved):
LGTM! ✅

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
nuno-vieira: I see, makes sense 👍 Lets do it like this then. Since the API is private, I'm good with both option 1 or 2 👍

Sources/StreamChat/ChatClient+UserSession.swift
laevandus: Removed this class and kept it more simple by moving it to the `ChatClient` instead.

PR #3373, opened 2024-08-13
Fix reading the local database state just after the initial write

laevandus:
### 🔗 Issue Links

Resolves [PBE-5519](https://stream-io.atlassian.net/browse/PBE-5519)

### 🎯 Goal

- Enable reading the local database state within `synchronize()`'s completion block when the initial DB state was empty

### 📝 Summary

- Remove the additional `OperationQueue` from the `BackgroundDatabaseObserver` (less latency)
- Process DB changes immediately on the managed context queue (less latency)
- Fix an issue where the initial observer state was set to empty although data was just written to the DB
- Fix a timing issue with will change callback what returned changed data (timing related)

### 🛠 Implementation

`BackgroundDatabaseObserver` tries to load the initial state as soon as observing starts using a background thread (less blocking of the main thread). Since it is an async operation, there is a data race between loading the initial state and the managed object context changing at the same time. For avoiding it, the observer keeps an internal state to figure out when it can use the cached items array. This internal state was incorrect if the DB was empty and there just was a DB write as well. The observer's implementation is similar to the `StateLayerDatabaseObserver` now where we use NSManagedObjectContext's thread for processing and accessing the state (better data consistency).

Account Chewbacca in the demo app:
```swift
// ComposerVC.swift:925
    quotedMessageId: content.quotingMessage?.id,
    skipEnrichUrl: content.skipEnrichUrl,
    extraData: content.extraData
  )
  test()
}

func test() {
  guard let chatClient = channelController?.client else { return }
  let identifier = ChannelId(type: .messaging, id: "D65B357E-8")
  let query = ChannelQuery(cid: identifier, pageSize: 20)
  let channelController = chatClient.channelController(for: query)
        
  channelController.synchronize { error in
    if let error {
        log.error("Failed to get messages: \(error)")
        return
     }
     // The last 20 messages should be in channelController.messages
     let messages = channelController.messages
     print(messages.count)
  }
}
``` 

### 🧪 Manual Testing Notes

Account: Chewbacca
1. Paste that snippet to somewhere which gets triggered after connecting (I used ComposerVC.swift line 925 to call this snippet, this is triggered when sending a message)
2. Logout and login with Chewbacca
3. Send a message which triggers the snippet
4. Observe the message count, it should return 20 (checking console log, best is to put a breakpoint after the print)

Exploratory testing (because DB observers affect all the controllers)

### ☑️ Contributor Checklist

[PBE-5519]: https://stream-io.atlassian.net/browse/PBE-5519?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### ⚡ Performance
 - Use background database observers for `CurrentUserController.currentUser`, `ChatChannelMemberListController.members`, and `ChatMessageController.message` which reduces CPU usage on the main thread [#3357](https://github.com/GetStream/stream-chat-swift/pull/3357)
 - `CurrentChatUserController` was often recreated when sending typing events [#3372](https://github.com/GetStream/stream-chat-swift/pull/3372)
+- Reduce latency of the `BackgroundDatabaseObserver` by reducing thread switching when handling changes [#3373](https://github.com/GetStream/stream-chat-swift/pull/3373)
 ### 🐞 Fixed
 - Fix rare crashes when deleting local database content on logout [#3355](https://github.com/GetStream/stream-chat-swift/pull/3355)
 - Fix rare crashes in `MulticastDelegate` when accessing it concurrently [#3361](https://github.com/GetStream/stream-chat-swift/pull/3361)
+- Fix reading the local database state just after the initial write [#3373](https://github.com/GetStream/stream-chat-swift/pull/3373)
 ### 🔄 Changed
 - Made loadBlockedUsers in ConnectedUser public [#3352](https://github.com/GetStream/stream-chat-swift/pull/3352)
 - Removed Agora and 100ms related code, the recommended way to support calls is to use StreamVideo [#3364](https://github.com/GetStream/stream-chat-swift/pull/3364)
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -6,9 +6,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## StreamChat
 ### ⚡ Performance
 - Use background database observers for `CurrentUserController.currentUser`, `ChatChannelMemberListController.members`, and `ChatMessageController.message` which reduces CPU usage on the main thread [#3357](https://github.com/GetStream/stream-chat-swift/pull/3357)
+- Reduce latency of the `BackgroundDatabaseObserver` by reducing thread switching when handling changes [#3373](https://github.com/GetStream/stream-chat-swift/pull/3373)
```
nuno-vieira, line 9: @laevandus There is a -120.24% decrease in performance of the Channel List, we should re-run the benchmark to make sure it was not false positive

nuno-vieira (approved):
Code Wise looks good. But we need to make sure the benchmark passes and also we should do some smoke testing CC @testableapple

martinmitrevski (approved):
Looks good ✅  Please check the failed performance benchmarks before merging

CHANGELOG.md
laevandus: I'll rerun it but locally when profiling with Instruments I got 0 hangs on my iPhone 12 Pro Max. No DB reads on the main thread either in `BackgroundDatabaseObserver`.

laevandus: Ran it manually [here](https://github.com/GetStream/stream-chat-swift/actions/runs/10384123700/job/28750433059)

laevandus: One red, others green. Seems like the perf test is a bit random.

laevandus:
Manually triggered performance job [here](https://github.com/GetStream/stream-chat-swift/actions/runs/10401397685)

PR #3382, opened 2024-08-15
Fix the thread name printed by the `Logger`

SagarSDagdu:
### 🎯 Goal

Make the logger print the correct thread name

### 📝 Summary

The thread name was being read inside `loggerQueue.async {}`, which made it always print `io.getstream.logconfig`. Fixed to read it before entering the dispatch queue, and now it prints the correct thread name.

### ☑️ Contributor Checklist

### 🎁 Meme

image


nuno-vieira:
Ahah I see you missed us @SagarSDagdu 😛

Thank you for the fix!

laevandus:
Hi @SagarSDagdu, this change has been released with 4.63.0. Thank you!

PR #3419, opened 2024-09-11
Sorting order in lists with equal sorting criteria

laevandus:
### 🔗 Issue Links

Resolves: [PBE-5993](https://stream-io.atlassian.net/browse/PBE-5993)

### 🎯 Goal

When sorting by key(s) with equal values the sort order in SDK models is kept unchanged

### 📝 Summary

> Recommendation: Always try to add sorting criteria what contains one sorting key with unique values.

- Fix an issue where microseconds in timestamps were ignored causing channel list sorting to have many channels with the same `lastMessageAt` timestamp (triggers active reordering)
- Fix an issue where just sorting by `unreadCount` or other Int or Bool keys led to unexpected reordering in the channel list (many channels with equal sorting value). Break this by adding `updatedAt` sorting key internally.
- Fix an issue of runtime sorting (post FRC sorting used when FRC does not support sorting key) returning incorrect order
- Add `ChannelMemberListSortingKey.userId` for sorting channel members by id

##### Microseconds parsing
DateFormatter supports milliseconds and ignores microseconds and nanoseconds. This led to date properties to lose precision and causing many equal values when sorting the list.

##### Equal values and sorting
Our SDK returns items sorted by Core Data sort descriptors (or by runtime sorting if custom sorting keys are used). When we sort over non-unique values and there are many items with the same value, the order returned by the SDK is inconsistent: `channel1`, `channel2`, `channel3` can change to `channel2`, `channel3`, `channel1` when DB changes happen. API responses, on the other hand, return consistent order even in cases like these. If we paginate, then we also do not want to see any reordering happening when the next page of channels is inserted into the DB. One way is to internally add a tie breaker.

##### The issue with runtime sorting
Runtime sorting is used when we can't sort on the Core Data model layer (DTO does not have a key for that field). I found that runtime sorting does not match with how NSSortDescriptor sorts items (nil values, multiple sort criteria). This PR fixes these issues.

##### Channel list sorted by unread count

| Before | After |
| ------ | ----- |
|  img  | img  |

##### Prerequisites
Open `DemoAppCoordinator+DemoApp.swift:L105` and change default sorting to sort by unread count (gives many equal values)
```swift
channelListQuery = .init(
  filter: .containMembers(userIds: [userCredentials.id]),
  sort: [.init(key: .unreadCount, isAscending: false)]
)
```
##### Steps
1. Open the demo app and log in with an user with many channels
2. Scroll down to load more channels
Result: Channels do not reorder

### ☑️ Contributor Checklist

### 🎁 Meme

[PBE-5993]: https://stream-io.atlassian.net/browse/PBE-5993?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/StreamChat/Database/DTOs/ChannelDTO.swift
+++ b/Sources/StreamChat/Database/DTOs/ChannelDTO.swift
@@ -21,6 +21,8 @@ class ChannelDTO: NSManagedObject {
     @NSManaged var updatedAt: DBDate
     @NSManaged var lastMessageAt: DBDate?
 
+    @NSManaged var localInsertionAt: DBDate
+    
     // The oldest message of the channel we have locally coming from a regular channel query.
     // This property only lives locally, and it is useful to filter out older pinned/quoted messages
     // that do not belong to the regular channel query.
@@ -155,6 +157,7 @@ class ChannelDTO: NSManagedObject {
 
         let new = NSEntityDescription.insertNewObject(into: context, for: request)
         new.cid = cid.rawValue
+        new.localInsertionAt = DBDate()
         return new
     }
 }
@@ -371,9 +374,13 @@ extension ChannelDTO {
         let request = NSFetchRequest<ChannelDTO>(entityName: ChannelDTO.entityName)
 
         // Fetch results controller requires at least one sorting descriptor.
-        let sortDescriptors = query.sort.compactMap { $0.key.sortDescriptor(isAscending: $0.isAscending) }
-        request.sortDescriptors = sortDescriptors.isEmpty ? [ChannelListSortingKey.defaultSortDescriptor] : sortDescriptors
-
+        var sortDescriptors = query.sort.compactMap { $0.key.sortDescriptor(isAscending: $0.isAscending) }
+        sortDescriptors = sortDescriptors.isEmpty ? [ChannelListSortingKey.defaultSortDescriptor] : sortDescriptors
+        
+        // For consistent order, we need to have a sort descriptor which breaks ties
+        sortDescriptors.append(NSSortDescriptor(keyPath: \ChannelDTO.localInsertionAt, ascending: true))
+        request.sortDescriptors = sortDescriptors
+        
         let matchingQuery = NSPredicate(format: "ANY queries.filterHash == %@", query.filter.filterHash)
         let notDeleted = NSPredicate(format: "deletedAt == nil")
 
--- a/Sources/StreamChat/Database/DTOs/ChannelDTO.swift
+++ b/Sources/StreamChat/Database/DTOs/ChannelDTO.swift
@@ -21,8 +21,6 @@ class ChannelDTO: NSManagedObject {
     @NSManaged var updatedAt: DBDate
     @NSManaged var lastMessageAt: DBDate?
 
-    @NSManaged var localInsertionAt: DBDate
-    
     // The oldest message of the channel we have locally coming from a regular channel query.
     // This property only lives locally, and it is useful to filter out older pinned/quoted messages
     // that do not belong to the regular channel query.
@@ -157,7 +155,6 @@ class ChannelDTO: NSManagedObject {
 
         let new = NSEntityDescription.insertNewObject(into: context, for: request)
         new.cid = cid.rawValue
-        new.localInsertionAt = DBDate()
         return new
     }
 }
@@ -375,12 +372,16 @@ extension ChannelDTO {
 
         // Fetch results controller requires at least one sorting descriptor.
         var sortDescriptors = query.sort.compactMap { $0.key.sortDescriptor(isAscending: $0.isAscending) }
-        sortDescriptors = sortDescriptors.isEmpty ? [ChannelListSortingKey.defaultSortDescriptor] : sortDescriptors
         
-        // For consistent order, we need to have a sort descriptor which breaks ties
-        sortDescriptors.append(NSSortDescriptor(keyPath: \ChannelDTO.localInsertionAt, ascending: true))
-        request.sortDescriptors = sortDescriptors
+        // For consistent order we need to have a sort descriptor which breaks ties
+        if !sortDescriptors.isEmpty, !sortDescriptors.contains(where: { $0.key == ChannelListSortingKey.updatedAt.localKey }) {
+            if let tieBreaker = ChannelListSortingKey.updatedAt.sortDescriptor(isAscending: false) {
+                sortDescriptors.append(tieBreaker)
+            }
+        }
         
+        request.sortDescriptors = sortDescriptors.isEmpty ? [ChannelListSortingKey.defaultSortDescriptor] : sortDescriptors
+
         let matchingQuery = NSPredicate(format: "ANY queries.filterHash == %@", query.filter.filterHash)
         let notDeleted = NSPredicate(format: "deletedAt == nil")
 
--- a/Sources/StreamChat/Query/Sorting/ChannelListSortingKey.swift
+++ b/Sources/StreamChat/Query/Sorting/ChannelListSortingKey.swift
@@ -62,7 +62,7 @@ public struct ChannelListSortingKey: SortingKey, Equatable {
 
     /// Sort channels by their unread count.
     public static let unreadCount = Self(
-        keyPath: \.unreadCount,
+        keyPath: \.unreadCount.messages,
         localKey: #keyPath(ChannelDTO.currentUserUnreadMessagesCount),
         remoteKey: "unread_count"
     )
--- a/Sources/StreamChat/Query/Sorting/ChannelListSortingKey.swift
+++ b/Sources/StreamChat/Query/Sorting/ChannelListSortingKey.swift
@@ -124,7 +124,7 @@ extension Array where Element == Sorting<ChannelListSortingKey> {
     }
 }
 
-private extension Sorting where Key == ChannelListSortingKey {
+extension Sorting where Key == ChannelListSortingKey {
     var sortValue: SortValue<ChatChannel>? {
         SortValue(keyPath: key.keyPath, isAscending: isAscending)
     }
--- a/Sources/StreamChat/Query/Sorting/SortValue.swift
+++ b/Sources/StreamChat/Query/Sorting/SortValue.swift
@@ -10,49 +10,54 @@ struct SortValue<T> {
 }
 
 extension Array {
-    func sort(using sorting: [SortValue<Element>]) -> [Element] {
-        var result = self
-
-        for sort in sorting.reversed() {
-            result = result.sorted { lhs, rhs in
-                let lhsValue = lhs[keyPath: sort.keyPath]
-                let rhsValue = rhs[keyPath: sort.keyPath]
-
-                if sort.isAscending, evaluate(lhs: lhsValue, rhs: rhsValue, isAscending: sort.isAscending) {
-                    return true
-                } else if !sort.isAscending, evaluate(lhs: lhsValue, rhs: rhsValue, isAscending: sort.isAscending) {
-                    return true
-                } else {
-                    return false
+    /// Returns the elements of the sequence, sorted using the given sort values as the comparison between elements.
+    ///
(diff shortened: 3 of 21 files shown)


Sources/StreamChat/Database/DatabaseSession.swift
```diff
@@ -303,6 +305,26 @@ protocol ChannelDatabaseSession {
     func removeChannels(cids: Set<ChannelId>)
 }
 
+extension ChannelDatabaseSession {
```
laevandus, line 308: Without it, there are so many errors because of missing the `offset` argument (most of it is in tests). If we want to avoid this extension, I can take the time to fill it in.

Sources/StreamChat/Query/Sorting/ChannelListSortingKey.swift
```diff
@@ -124,7 +124,7 @@ extension Array where Element == Sorting<ChannelListSortingKey> {
     }
 }
 
-private extension Sorting where Key == ChannelListSortingKey {
+extension Sorting where Key == ChannelListSortingKey {
```
laevandus, line 127: For unit-tests

Sources/StreamChat/Query/Sorting/SortValue.swift
```diff
@@ -10,49 +10,54 @@ struct SortValue<T> {
...
-            return true
-        } else if lhs != nil, rhs == nil, isAscending {
-            return true
+    
+    /// Dedicated nil handling for Any type which returns false if we do `lhs == nil`.
+    private func nilComparison(lhs: Any, rhs: Any, isAscending: Bool) -> Bool? {
+        func isAnyNil(_ value: Any) -> Bool {
+            let mirror = Mirror(reflecting: value)
+            return mirror.displayStyle == .optional && mirror.children.isEmpty
         }
```
laevandus, line 48: Is there a better way?

Sources/StreamChat/Utils/Sequence+CompactMapLoggingError.swift
```diff
@@ -4,7 +4,7 @@
 
 import Foundation
 
-extension Collection {
+extension Sequence {
```
laevandus, line 7: Needed it for sorting `.enumerated()`

Tests/StreamChatTests/Query/Sorting/SortingValue_Tests.swift
```diff
@@ -61,8 +61,141 @@ final class SortingValue_Tests: XCTestCase {
...
+        let nsArrayChannels = NSArray(array: channels.map { ChannelBoxed(channel: $0) })
+        let sortDescriptors = sortingKeys.compactMap { $0.key.sortDescriptor(isAscending: $0.isAscending) }
+        XCTAssertEqual(2, sortDescriptors.count)
+        
+        let sortDescriptorResult = (nsArrayChannels.sortedArray(using: sortDescriptors) as! [ChannelBoxed]).compactMap(\.id)
+        let expectedIds = ["1", "2", "3", "4", "5"]
+        XCTAssertEqual(expectedIds, sortDescriptorResult, "\(sortDescriptorResult)")
+        XCTAssertEqual(expectedIds, runtimeSortResult, "\(runtimeSortResult)")
+    }
+
```
laevandus, line 177: TODO: Add one test-case for ascending and descending sorting. E.g. sort by unread ascending AND sort by created descending

Sources/StreamChat/Database/DatabaseSession.swift
martinmitrevski: can't we provide a default value for the offset?

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
```diff
@@ -21,6 +21,13 @@ class ChannelDTO: NSManagedObject {
     @NSManaged var updatedAt: DBDate
     @NSManaged var lastMessageAt: DBDate?
 
+    // Index of the channel in the channel list query response.
+    // It is used for tie breaker when sorting by possibly equal values.
+    // Note: When the channel is part of multiple channel list controllers,
+    // it might still cause some movement because the property is shared.
+    // Recommendation: use at least one sorting parameter with unique values.
+    @NSManaged var sortIndex: Int
```
martinmitrevski, line 29: another option can be to keep a map of the queries and the indexes, but I guess it's not worth it

Sources/StreamChat/Database/DatabaseSession.swift
martinmitrevski: or maybe move these to the test target?

Sources/StreamChat/Query/Sorting/SortValue.swift
martinmitrevski: can't we do something like this instead:
```
func isAnyNil(_ value: Any) -> Bool {
     return (value as Any?) == nil
}
```

martinmitrevski (commented):
I tested it also on SwiftUI and I still get an occasional jump. Have you tried it there?

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
laevandus: I considered it but it felt like it gets too complex and also sort descriptor wants a key-path.

Sources/StreamChat/Query/Sorting/SortValue.swift
laevandus: This also does not work. 

```swift
func test() {
    struct Item {
        let index: Int
        let date: Date?
    }

    let item1 = Item(index: 1, date: nil)
    let item2 = Item(index: 2, date: Date())
    let keyPath: PartialKeyPath<Item> = \.date

    let lhsValue = item1[keyPath: keyPath]
    let rhsValue = item2[keyPath: keyPath]

    print(lhsValue, rhsValue)
    print(isAnyNil(lhsValue), isAnyNil(rhsValue))
    print((lhsValue as Any?) == nil, (rhsValue as Any?) == nil)
}

func isAnyNil(_ value: Any) -> Bool {
    let mirror = Mirror(reflecting: value)
    return mirror.displayStyle == .optional && mirror.children.isEmpty
}
```
gives
```swift
nil Optional(2024-09-17 10:18:17 +0000)
true false
false false
```

Sources/StreamChat/Database/DatabaseSession.swift
laevandus: Unfortunately can't set a default value in protocol but moving the extension to test target is a good idea.

laevandus: Moved it

Sources/StreamChat/Query/Sorting/SortValue.swift
nuno-vieira: @laevandus You can check `DebugObjectViewController`, we do nil comparison there.

martinmitrevski: I want to avoid reflection though, this will be run for every channel so it might impact performance.

nuno-vieira (commented):
This approach sounds much better 👍 Just added minor comments but overall LGTM!

Sources/StreamChat/Query/Sorting/ChannelListSortingKey.swift
```diff
@@ -62,7 +62,7 @@ public struct ChannelListSortingKey: SortingKey, Equatable {
 
     /// Sort channels by their unread count.
     public static let unreadCount = Self(
-        keyPath: \.unreadCount,
+        keyPath: \.unreadCount.messages,
```
laevandus, line 65: unreadCount is a struct, not Int, this is used for runtime sorting

Sources/StreamChat/Query/Sorting/SortValue.swift
```diff
@@ -10,49 +10,57 @@ struct SortValue<T> {
...
+                    return result
+                } else if let result = areInIncreasingOrder(lhs: lhsValue, rhs: rhsValue, type: String.self, isAscending: isAscending) {
+                    return result
+                } else if let result = areInIncreasingOrder(lhs: lhsValue, rhs: rhsValue, type: Int.self, isAscending: isAscending) {
+                    return result
+                } else if let result = areInIncreasingOrder(lhs: lhsValue, rhs: rhsValue, type: Double.self, isAscending: isAscending) {
+                    return result
+                } else if let lBool = lhsValue as? Bool, let rBool = rhsValue as? Bool, lBool != rBool {
+                    // Backend considers boolean sorting in reversed order.
+                    return isAscending ? lBool && !rBool : !lBool && rBool
```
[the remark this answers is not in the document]
laevandus: We still need, this code-path is triggered when someone uses a custom sorting key which can't be represented by the FRC's sort descriptors. Most of the time, this code never runs, only these specific cases when someone goes for custom sorting keys.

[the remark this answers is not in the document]
laevandus: Sorting keys what have localKey nil. Nil means we can't use sort descriptor and need to resort post-FRC sorting.
```
public static let hasUnread = Self(
        keyPath: \.hasUnread,
        localKey: nil,
        remoteKey: "has_unread"
    )
    public static func custom<T>(keyPath: KeyPath<ChatChannel, T>, key: String) -> Self {
        .init(keyPath: keyPath, localKey: nil, remoteKey: key)
    }
```

Sources/StreamChat/Utils/Codable+Extensions.swift
```diff
@@ -167,6 +167,24 @@ extension DateFormatter {
...
+        guard let index = string.lastIndex(of: ".") else { return date }
+        let range = string.suffix(from: index)
+            .dropFirst(4) // . and ms part
+            .dropLast() // Z
+        var fractionWithoutMilliseconds = String(range)
+        if fractionWithoutMilliseconds.count < 3 {
+            fractionWithoutMilliseconds = fractionWithoutMilliseconds.padding(toLength: 3, withPad: "0", startingAt: 0)
+        }
+        guard let microseconds = TimeInterval("0.000".appending(fractionWithoutMilliseconds)) else { return date }
+        return date.addingTimeInterval(microseconds)
```
laevandus, line 184: Maybe not the most beautiful but it needs to handle cases like
```
2024-09-18T13:49:11.3241Z
2024-09-18T13:49:11.32412Z
2024-09-18T13:49:11.324123Z
2024-09-18T13:49:11.3241234Z
2024-09-18T13:49:11.32412345Z
2024-09-18T13:49:11.324123456Z
```

Sources/StreamChat/Utils/Codable+Extensions.swift
```diff
@@ -167,6 +167,24 @@ extension DateFormatter {
     }
 }
 
+extension ISO8601DateFormatter {
+    func dateWithMicroseconds(from string: String) -> Date? {
```
martinmitrevski, line 171: hmm, could this logic affect the parsing time?

martinmitrevski (approved):
looks good to me and things are working well now. Also, we should update the SwiftUI demo app after merging this.

Sources/StreamChat/Utils/Codable+Extensions.swift
laevandus: I did profiling:
1. Demo app in logged out state
2. Log in with chewbacca
3. Open release 4.63.0 channel, scroll to the very first message
4. Check how much time the dateWithMicroseconds takes, compare it with original 'date(from:)'

In both cases the result was 48 ms which means the additional parsing is negligible compared to what the formatter takes.

nuno-vieira (approved):
LGTM! ✅ But better for @testableapple do to some smoke testing before merging

testableapple (approved):
There is a slight movement of the channels after loading the first page, but it's not a big deal and almost invisible. One possible workaround is to add another sorting key on top of the unread count.

PR #3430, opened 2024-09-24
Linking and unlinking channels based on the loaded list of channels

laevandus:
### 🔗 Issue Links

Resolves: [PBE-6097](https://stream-io.atlassian.net/browse/PBE-6097)

### 🎯 Goal

- Fix issues with linking a channel when dynamic filter is used

### 📝 Summary

- Fix issues where web-socket event updates the channel and it was not linked to the current query (extra data changes in a way that it matches with the current query now, think about the custom pinning feature in the demo app)
- Revisit the whole web-socket event handling for linking and make sure we link only when after sorting, the changed channel would be part of the currently loaded channel list (makes sure we do not insert it when it is from unloaded pages and therefore would disturb pagination)

##### Case 1: Linking a channel after being added to it
1. Launch two clients A and B
2. On the client B, A's user is added to a channel
Result: The channel appears in the A's channel list

##### Case 2: Receiving a message in channel not loaded in the channel list
1. Launch two clients A and B
2. On the client B, send a message to an old channel which includes A's user
Result: The channel appears in the A's channel list

##### Case 3: Hiding/showing channels
1. Launch two clients A and B with the **same** user
2. On the client A, use the hidden channels filter
3. On the client B, hide a channel at the top of the channel list (Channel's debug menu)
Result: The channel appears in the A's channel list
4. On the client B, navigate to hidden channels and "show" the channel (Channel's debug menu > Show Channel)
Result: The channel disappears in the A's channel list

##### Case 4: Extra data based filtering 
1. Launch two clients A and B with the **same** user
2. Enable channel pinning for both (configuration when logged out) (this makes pinned channels to appear at the top of the list)
3. On the client B, pin a channel
Result: A sees the channel at the top of the channel list
4. On the client B, unpin the same channel
Result: A sees the channel moving to a different location

### ☑️ Contributor Checklist

[PBE-6097]: https://stream-io.atlassian.net/browse/PBE-6097?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/StreamChat/Workers/ChannelListLinker.swift
+++ b/Sources/StreamChat/Workers/ChannelListLinker.swift
@@ -8,15 +8,16 @@ import Foundation
 ///
 /// Requires either `filter` or `isChannelAutomaticFilteringEnabled` to be set.
 /// - Channels are inserted (linked) only when they would end up on the currently loaded pages.
-/// - Channels are removed (unlinked) when not on the currently loaded pages.
+/// - Channels are removed (unlinked) when not on the currently loaded pages. For example, event changes
+/// extra data which makes it not to match with the current filter closure anymore.
 final class ChannelListLinker {
     private let clientConfig: ChatClientConfig
     private let databaseContainer: DatabaseContainer
     private var eventObservers = [EventObserver]()
     private let filter: ((ChatChannel) -> Bool)?
     private let loadedChannels: () -> StreamCollection<ChatChannel>
-    private let query: ChannelListQuery
     private let worker: ChannelListUpdater
+    let query: ChannelListQuery
     
     init(
         query: ChannelListQuery,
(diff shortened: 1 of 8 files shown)


Sources/StreamChat/Workers/ChannelListLinker.swift
```diff
@@ -53,7 +53,12 @@ final class ChannelListLinker {
             EventObserver(
                 notificationCenter: nc,
                 transform: { $0 as? ChannelUpdatedEvent },
-                callback: { [weak self] event in self?.unlinkChannelIfNeeded(event.channel) }
+                callback: { [weak self] event in
+                    guard let self else { return }
+                    let shouldUnlink = self.unlinkChannelIfNeeded(event.channel)
+                    guard !shouldUnlink else { return }
+                    self.linkChannelIfNeeded(event.channel)
+                }
```
nuno-vieira, line 61: According to the documentation of `ChannelListLinker` this can break some other customers:

> /// - Channel is updated: We only check if we should remove it from the current query.
> ///   We don't try to add it to the current query to not mess with pagination.

For now, the only solution I can think of is to delegate the linking of channels to the customer when a channel is updated. 

We can either expose a closure whenever a channel is updated, or maybe better and even more flexible is to expose the linking and unlinking functions so that customers can handle linking themselves based on any event.

nuno-vieira (changes requested):
Requested changes since this can break other customers. I've added a couple suggestions, but we will discuss this further internally 👍

nuno-vieira (approved):
LGTM! ✅

martinmitrevski (approved):

laevandus:
We did QA with @testableapple on this one. Marking as QAd.

PR #3431, opened 2024-09-24
[CI] Fix snapshots path

testableapple:


fastlane/Fastfile
```diff
@@ -604,8 +604,10 @@ lane :test_ui do |options|
     next if png_files.empty?
 
     # Discard all files apart from the snapshots
-    png_files.each { |png| sh("git add #{png}") || true }
-    sh('git restore .')
+    Dir.chdir('..') do
+      png_files.each { |png| sh("git add #{png}") || true }
+      sh('git restore .')
+    end
```
laevandus, line 610: Because of the fastlane's default path handling?

testableapple: Because `png_files` now do not have `fastlane/` in their path, after we moved the logic to the plugin.

laevandus: Thank you for clarifying

laevandus (approved):

PR #3432, opened 2024-09-25
Avoid rare optimistic locking failure by refreshing the object before saving

laevandus:
### 🔗 Issue Links

Resolves: [PBE-6035](https://stream-io.atlassian.net/browse/PBE-6035)

### 🎯 Goal

Try fixing rare crash on managed object context save

### 🛠 Implementation

I learned that the rare crash we have seen on context save is related to Core Data failing to resolve constraint conflict (although I can't see how this could happen). NSManagedObjectContext has [refresh(_:mergeChanges:)](https://developer.apple.com/documentation/coredata/nsmanagedobjectcontext/1506224-refresh) for forcing the context to fetch the persistent state and applying local changes. I would like to see if this resolves these rare cases.

> Merging the local values into object always succeeds, and never results in a merge conflict.

> You typically use this method to ensure data freshness if multiple managed object contexts share a single persistent store. You can use this method to resolve an optimistic locking failure when attempting to save.

### 🧪 Manual Testing Notes

- [ ] Extensive exploratory testing
- [x] Time profiler

### ☑️ Contributor Checklist

[PBE-6035]: https://stream-io.atlassian.net/browse/PBE-6035?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/StreamChat/Database/DatabaseContainer.swift
+++ b/Sources/StreamChat/Database/DatabaseContainer.swift
@@ -222,10 +222,9 @@ class DatabaseContainer: NSPersistentContainer {
                 try actions(self.writableContext)
                 FetchCache.clear()
 
+                // Refresh the state by merging persistent state and local state for avoiding optimistic locking failure
                 for object in self.writableContext.updatedObjects {
-                    if object.changedValues().isEmpty {
-                        self.writableContext.refresh(object, mergeChanges: true)
-                    }
+                    self.writableContext.refresh(object, mergeChanges: true)
                 }
 
                 if self.writableContext.hasChanges {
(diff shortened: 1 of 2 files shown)


Sources/StreamChat/Database/DatabaseContainer.swift
```diff
@@ -222,10 +222,9 @@ class DatabaseContainer: NSPersistentContainer {
                 try actions(self.writableContext)
                 FetchCache.clear()
 
+                // Refresh the state by merging persistent state and local state for avoiding optimistic locking failure
                 for object in self.writableContext.updatedObjects {
-                    if object.changedValues().isEmpty {
-                        self.writableContext.refresh(object, mergeChanges: true)
-                    }
```
nuno-vieira, line 228: We need extensive testing first, to make sure this does not make things worse. Since this impacts pretty much the whole SDK, we need to be careful with this change

laevandus: Ran it with time profiler and it adds around 1-2% of the whole write closure.
Screenshot 2024-09-26 at 11 00 51

martinmitrevski (approved):

PR #3436, opened 2024-09-30
[CI] Improvements

testableapple:
- Extract `merge_main_to_develop` to an indie lane to be able to run it locally
- Trigger `release_publish.yml` workflow via `gh` instead of `push` event that did not work


fastlane/Fastfile
```diff
@@ -88,6 +88,11 @@ end
 
 lane :merge_release do |options|
   merge_release_to_main(author: options[:author])
+  sh('gh workflow run release-publish.yml --ref main')
```
laevandus, line 91: Did you verify that it works on CI?

testableapple: https://github.com/GetStream/stream-chat-swift/actions/runs/11126371804/job/30916141810 👍

laevandus (approved):

PR #3446, opened 2024-10-03
Fix UI glitch in thread parent message when sending a message and scrolling

nuno-vieira:
### 🔗 Issue Links

Resolves https://stream-io.atlassian.net/browse/PBE-5999

### 🎯 Goal

Fix an issue where the thread parent message would shrink after sending a thread reply and quickly scroll to the parent message.

#### Problem
The reason for the issue is that the decorator's views rely on showing and hiding the top and bottom stack views in `ChatMessageCell`, but when quickly scrolling, the message cell, since it has the same identifier as the regular reply, is reused and the decorators are not shown instantly and it requires a cell reload to re-calculate its size. 

#### Solution
Using our custom `ContainerStackView` fixes the problem since it relies on manually disabling and enabling constraints, which will re-calculate the cell's size.

### 🧪 Manual Testing Notes

Follow to the steps in https://stream-io.atlassian.net/browse/PBE-5999

### ☑️ Contributor Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 # Upcoming
 
-### 🔄 Changed
+## StreamChatUI
+### 🐞 Fixed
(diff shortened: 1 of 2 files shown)


laevandus (approved):
Although it would be nice to see more UIStackView instead, most of the layout is built with ContainerStackView anyway so I think it is OK to change it here as well.

PR #3448, opened 2024-10-07
Improve performance of presenting ChatChannelVC

laevandus:
### 🔗 Issue Links

Resolves: [PBE-6192](https://stream-io.atlassian.net/browse/PBE-6192)

### 🎯 Goal

Present `ChatChannelVC` quicker

### 📝 Summary

- Reduce cell reloads after applying message list changes
- Skip duplicate `updateContent` call when creating the `ChatMessageContentCell`

### 🛠 Implementation

`ChatMessageContentView` parses markdown and links which can get expensive. Reloading a cell repeats the parsing. This PR aims to reduce cell reloads. 

### 🎨 Showcase

Measured with iPhone 12 Pro Max and 3 different channels using a separate sample app what just presents the `ChatChannelVC` in a sheet. Time spent by the main thread.

Channel with 25 simple messages with text. Negligible change here. **5-10%** faster.
| Criteria | Before | After |
| ----- | ------ | ----- |
| Main thread |  269 ms  |  251 ms  |
| Hang time | 297 ms  | 291 ms  |

Channel with complex markdown and some other simple messages with links. **~50%** faster
| Criteria | Before | After |
| ----- | ------ | ----- |
| Main thread |  969 ms  |  420 ms  |
| Hang time | 811 ms  | 480 ms |

All of this really depends on how much processing the message cell requires. If more, the impact is higher because it is less time recalculated.

##### Verify sending giphys
1. Open giphy preview
2. Send/cancel a giphy

##### Verify deleting messages and displaying user avatars.

1. Log in from 2 devices/simulators
2. Open the same channel with 2 different users
3. Send 2 messages: "A" and "B"
4. Hard delete the "B"
Result: Avatar is shown for message "A"

##### Verify sending messages and displaying user avatars.
1. Send message "A" > avatar is shown for it
2. Send message "B" > avatar is shown for "B" and not for "A" since messages are grouped

### ☑️ Contributor Checklist

[PBE-6192]: https://stream-io.atlassian.net/browse/PBE-6192?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
+++ b/Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
@@ -271,7 +271,6 @@ open class ChatMessageListVC: _ViewController,
 
     override open func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
         super.traitCollectionDidChange(previousTraitCollection)
-
         view.layoutIfNeeded()
         listView.adjustContentInsetToPositionMessagesAtTheTop()
     }
@@ -1304,10 +1303,14 @@ private extension ChatMessageListVC {
             }
 
             UIView.performWithoutAnimation {
-                self?.scrollToBottomIfNeeded(with: changes, newestChange: newestChange)
-                self?.reloadMovedMessage(newestChange: newestChange)
-                self?.reloadPreviousMessagesForVisibleRemoves(with: changes)
-                self?.reloadPreviousMessageWhenInsertingNewMessage()
+                guard let self else { return }
+                self.scrollToBottomIfNeeded(with: changes, newestChange: newestChange)
+                var additionalChanges = Set<IndexPath>()
+                additionalChanges.formUnion(self.reloadMovedMessage(newestChange: newestChange))
+                additionalChanges.formUnion(self.reloadPreviousMessagesForVisibleRemoves(with: changes))
+                additionalChanges.formUnion(self.reloadPreviousMessageWhenInsertingNewMessage(with: changes))
+                guard !additionalChanges.isEmpty else { return }
+                self.listView.reloadRows(at: additionalChanges.sorted(), with: .none)
             }
 
             self?.scrollPendingMessageIfNeeded()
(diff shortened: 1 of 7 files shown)


Sources/StreamChatUI/StreamSwiftyMarkdown/SwiftyMarkdown/PerfomanceLog.swift
```diff
@@ -16,7 +16,7 @@ class PerformanceLog {
 	
 	init( with environmentVariableName : String, identifier : String, log : OSLog  ) {
 		self.log = log
-		self.enablePerfomanceLog = (ProcessInfo.processInfo.environment[environmentVariableName] != nil)
```
martinmitrevski, line 19: why do we need to change this? It's in the markdown library. If really needed, probably we need to update something in the scripts, since the change would be overwritten if we re-embed it.

laevandus: `ProcessInfo.processInfo.environment` keeps showing up in the profile (~15ms). Since it is not used, I ended up disabling it.

martinmitrevski: ok, please check with Alexey if changing it only here is enough.

Tests/StreamChatUITests/SnapshotTests/ChatChannel/ChatChannelVC_Tests.swift
```diff
@@ -652,6 +652,10 @@ final class ChatChannelVC_Tests: XCTestCase {
 
         vc.components.isMessageEditedLabelEnabled = true
 
+        // Also update the default because in snapshot tests, message cells are created before they are in the responder chain
+        defer { Components.default.isMessageEditedLabelEnabled = false }
+        Components.default.isMessageEditedLabelEnabled = true
```
nuno-vieira, line 657: Why is this not a problem for the other tests tho? 🤔

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -1354,28 +1380,6 @@ private extension ChatMessageListVC {
         listView.contentOffset.y = newOffset
     }
 
-    // If we are inserting messages at the bottom, update the previous cell
-    // to hide the timestamp of the previous message if needed.
-    func reloadPreviousMessageWhenInsertingNewMessage() {
```
nuno-vieira, line 1359: nit: We could have remained with the functions but changed them to return the indexPaths instead of performing the reload. For example, `indexPathForPreviousMessageWhenInsertingNewMessage() -> IndexPath?`

But not a big deal

nuno-vieira (approved):
LGTM! ✅ Just added non-blocking questions.

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
laevandus: I'll consider it

Tests/StreamChatUITests/SnapshotTests/ChatChannel/ChatChannelVC_Tests.swift
laevandus: Depends on the timing: when view is added to the view hierarchy vs when the modified property is accessed. Can't figure out another way how to get this test working. It worked before because ChatMessageContentView.updateContent was called twice (at the second time, the view was part of the view hierarchy) but now it is only once. And really hard to change the timing of the first because it is triggered from the `_View` base view everyone is using.

Sources/StreamChatUI/StreamSwiftyMarkdown/SwiftyMarkdown/PerfomanceLog.swift
testableapple: We need to delete `SwiftyMarkdown` from the `Makefile`. We've already implemented a ton of breaking changes. We are not going to update it ever.

laevandus: Sure, I can add that although I am sure we are not going to update it (it looks abandoned)

laevandus: @testableapple done

Sources/StreamChatUI/ChatMessageList/ChatMessageListVC.swift
```diff
@@ -1304,10 +1303,14 @@ private extension ChatMessageListVC {
...
-                self?.scrollToBottomIfNeeded(with: changes, newestChange: newestChange)
-                self?.reloadMovedMessage(newestChange: newestChange)
-                self?.reloadPreviousMessagesForVisibleRemoves(with: changes)
-                self?.reloadPreviousMessageWhenInsertingNewMessage()
+                guard let self else { return }
+                self.scrollToBottomIfNeeded(with: changes, newestChange: newestChange)
+                var additionalChanges = Set<IndexPath>()
+                additionalChanges.formUnion(self.reloadMovedMessage(newestChange: newestChange))
+                additionalChanges.formUnion(self.reloadPreviousMessagesForVisibleRemoves(with: changes))
+                additionalChanges.formUnion(self.reloadPreviousMessageWhenInsertingNewMessage(with: changes))
```
nuno-vieira, line 1311: 👨‍🍳👌

PR #3450, opened 2024-10-08
Reduce channel list updates when syncing

laevandus:
### 🔗 Issue Links

Resolves: [PBE-6191](https://stream-io.atlassian.net/browse/PBE-6191)

### 🎯 Goal

Reduce channel list updates when opening a channel list while syncing is in progress

### 📝 Summary

- Run channel list refresh before syncing events (`/sync`)
- Do not (again) fetch events for channels in the channel list (because these were just refreshed)
- Replace `Array` with `Set` in the context

### 🛠 Implementation

Currently the sequence has been (historically), call the `/sync` endpoint first and then refresh channel lists. This can trigger an issue where events applied from the `/sync` endpoint trigger local database change (channels can reorder) and then a moment later the whole channel list is refreshed which might cause another reordering. This PR changes things around by first refreshing channel list and then calling /sync only for active channels not part of the channel list (aka channel controllers).
Result:
1. Less reordering
2. Less cases where /sync exceeds the max event count API error 
3. Less data is fetched when syncing (duplicate data is skipped)

### 🧪 Manual Testing Notes

N/A

### ☑️ Contributor Checklist

[PBE-6191]: https://stream-io.atlassian.net/browse/PBE-6191?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,6 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Fix a rare crash caused by missing uniqueness constraints in polls [#3454](https://github.com/GetStream/stream-chat-swift/pull/3454)
 - Fix rare crash in `WebSocketPingController.connectionStateDidChange` [#3451](https://github.com/GetStream/stream-chat-swift/pull/3451)
 - Improve reliability and performance of resetting ephemeral values [#3439](https://github.com/GetStream/stream-chat-swift/pull/3439)
-- Reduce channel list updates when syncing by reording syncing sequence and skipping channel based events sync (if already in sync) [#3450](https://github.com/GetStream/stream-chat-swift/pull/3450)
 - Reduce channel list updates when updating the local state [#3450](https://github.com/GetStream/stream-chat-swift/pull/3450)
 
 ## StreamChatUI
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Fix a rare crash caused by missing uniqueness constraints in polls [#3454](https://github.com/GetStream/stream-chat-swift/pull/3454)
 - Fix rare crash in `WebSocketPingController.connectionStateDidChange` [#3451](https://github.com/GetStream/stream-chat-swift/pull/3451)
 - Improve reliability and performance of resetting ephemeral values [#3439](https://github.com/GetStream/stream-chat-swift/pull/3439)
+- Reduce channel list updates when syncing by reording syncing sequence and skipping channel based events sync (if already in sync) [#3450](https://github.com/GetStream/stream-chat-swift/pull/3450)
 
 ## StreamChatUI
 ### 🐞 Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Fix rare crash in `WebSocketPingController.connectionStateDidChange` [#3451](https://github.com/GetStream/stream-chat-swift/pull/3451)
 - Improve reliability and performance of resetting ephemeral values [#3439](https://github.com/GetStream/stream-chat-swift/pull/3439)
 - Reduce channel list updates when syncing by reording syncing sequence and skipping channel based events sync (if already in sync) [#3450](https://github.com/GetStream/stream-chat-swift/pull/3450)
+- Reduce channel list updates when updating the local state [#3450](https://github.com/GetStream/stream-chat-swift/pull/3450)
 
 ## StreamChatUI
 ### 🐞 Fixed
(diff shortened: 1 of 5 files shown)


Sources/StreamChat/Repositories/SyncRepository.swift
```diff
@@ -183,12 +183,12 @@ class SyncRepository {
...
-        // 2. /sync
-        operations.append(SyncEventsOperation(syncRepository: self, context: context, recovery: false))
-        
-        // 3. Refresh channel lists (required even after applying events)
+        // 2. Refresh channel lists
         operations.append(contentsOf: activeChannelLists.allObjects.map { RefreshChannelListOperation(channelList: $0, context: context) })
         operations.append(contentsOf: activeChannelListControllers.allObjects.map { RefreshChannelListOperation(controller: $0, context: context) })
+
+        // 3. /sync (for channels what not part of active channel lists)
+        operations.append(SyncEventsOperation(syncRepository: self, context: context, recovery: false))
```
nuno-vieira, line 191: Why would we want to sync channels that are not part of active channel lists? 🤔 I think we probably could discard these events. These channels will have the most up-to-date data whenever we refetch them

laevandus: One case I could think of is channel controller targeting a channel id which has not been loaded by the channel list yet (but sure yes it sounds like a special case). Or some sort of custom integrations not having channel list alive (e.g. operating on channel ids). Hard to know how valid these cases are.

Sources/StreamChat/Repositories/SyncRepository.swift
```diff
@@ -156,8 +156,8 @@ class SyncRepository {
     ///
     /// Background mode (other regular API requests are allowed to run at the same time)
     /// 1. Collect all the **active** channel ids (from instances of `Chat`, `ChannelList`, `ChatChannelController`, `ChatChannelListController`)
-    /// 2. Apply updates from the /sync endpoint for these channels
-    /// 3. Refresh channel lists (channels for current pages in `ChannelList`, `ChatChannelListController`)
+    /// 2. Refresh channel lists (channels for current pages in `ChannelList`, `ChatChannelListController`)
+    /// 3. Apply updates from the /sync endpoint for channels (which were not already refreshed, note max 1000 events is supported)
```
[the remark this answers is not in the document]
laevandus: There has been at least one customer who reported hitting the 400 API error. I think it happened when opening an old account (therefore a lot more events). Also, a bit special case here, but now it happens even less.

CHANGELOG.md
```diff
@@ -3,7 +3,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 # Upcoming
 
-### 🔄 Changed
+## StreamChat
+### 🐞 Fixed
+- Reduce channel list updates when syncing by reording syncing sequence and skipping channel based events sync (if already in sync) [#3450](https://github.com/GetStream/stream-chat-swift/pull/3450)
```
martinmitrevski, line 8: - typo: reording
- the entry itself is a bit long and not clear enough, maybe we can simplify it?

Sources/StreamChat/Repositories/SyncRepository.swift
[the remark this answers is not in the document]
martinmitrevski: nice that we mention this - I've also seen this happening few times.

Sources/StreamChat/Repositories/SyncRepository.swift
martinmitrevski: There are tab/segmented control based channel lists, which are not active at all times, but can be invoked with a tap. So I agree with Toomas on this one to keep this logic.

CHANGELOG.md
laevandus: Thank you, sure

Sources/StreamChat/Repositories/SyncRepository.swift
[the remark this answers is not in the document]
nuno-vieira: The problem is that syncing 1000 events in the DB will be very expensive, and for sure we don't need those 1000 events, most of those events will be outdated.

Sources/StreamChat/Repositories/SyncRepository.swift
nuno-vieira: @martinmitrevski @laevandus If they are not active, meaning, they have not be loaded yet, when a user opens a tab, we will fetch the data from the server, the most up-to-date one, so the events are useless

laevandus: I was thinking about the case where one creates channel controller, then app goes to background for awhile, comes back, then sync repo fetches events what the user missed.

nuno-vieira: But in that case, the controller should be part of the active controllers I believe

laevandus: We do not call update each on each active channel controller (relies on /sync, otherwise too many API calls)

Sources/StreamChat/Repositories/SyncRepository.swift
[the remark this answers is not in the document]
laevandus: Looked it up, it is actually 2000 :)

martinmitrevski (approved):
LGTM ✅

PR #3457, opened 2024-10-10
[CI] Unstable data container test

laevandus:
### 🎯 Goal

Fix an unstable test what has race between another write call and reading the database state.

### ☑️ Contributor Checklist

### DIFF
--- a/Tests/StreamChatTests/Database/DatabaseContainer_Tests.swift
+++ b/Tests/StreamChatTests/Database/DatabaseContainer_Tests.swift
@@ -143,11 +143,6 @@ final class DatabaseContainer_Tests: XCTestCase {
             expectation.fulfill()
         }
         
-        // Save just after triggering remove all
-        container.write { session in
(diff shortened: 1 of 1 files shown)


Tests/StreamChatTests/Database/DatabaseContainer_Tests.swift
```diff
@@ -143,11 +143,6 @@ final class DatabaseContainer_Tests: XCTestCase {
             expectation.fulfill()
         }
         
-        // Save just after triggering remove all
-        container.write { session in
-            try session.saveChannel(payload: self.dummyPayload(with: .unique), query: nil, cache: nil)
-        }
-
```
laevandus, line 150: The flow of the test:

1. write
2. delete all
3. write
4. read and validate state

There is a race between 3 and 4. The idea of the test was that if write is triggered and then we delete all, then everything is deleted, but I accidentally added another write on top of the delete.

martinmitrevski (approved):

PR #3460, opened 2024-10-15
[POC] Reuse ViewModels across SwiftUI and UIKit

nuno-vieira:
### Summary
This is just a proof of concept to see how feasible it would be to reuse view models across the SwiftUI and UIKit SDK. 

Spoiler Alert: It looks pretty promising 👌 

### Further work
- We would need to make sure we do not import UIKit or SwiftUI in the ViewModels to make sure they do not depend on any UI framework
- We would need to expose a constructor to replace the "@Injected" dependencies from SwiftUI
- We would need to make the View Models, Formatters and all presentation logic part of LLC or create an additional framework
- Maybe even move SwiftUI to this repo?

###  Demo

| Loading Threads | New Threads |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/911e3316-200c-45f1-b834-55e15e58fd8b"/> | <video src="https://github.com/user-attachments/assets/ccba9b0b-dc4b-42a0-8551-13c3b56340c5"/> |

### DIFF
--- a/Sources/StreamChatUI/ChatThreadList/ChatThreadListViewModel.swift
+++ b/Sources/StreamChatUI/ChatThreadList/ChatThreadListViewModel.swift
@@ -1,0 +1,221 @@
+//
+// Copyright © 2024 Stream.io Inc. All rights reserved.
+//
+
+import Combine
+import StreamChat
+import UIKit
+
+/// The ViewModel for the `ChatThreadListView`.
+open class ChatThreadListViewModel: ObservableObject, ChatThreadListControllerDelegate, EventsControllerDelegate {
+    /// Context provided dependencies.
+    public var chatClient: ChatClient!
+
+    /// The controller that manages the thread list data.
+    private var threadListController: ChatThreadListController!
+
+    /// The controller that manages thread list events.
+    private var eventsController: EventsController!
+
+    /// A boolean value indicating if the initial threads have been loaded.
+    public private(set) var hasLoadedThreads = false
+
+    /// The current selected thread.
+    @Published public var selectedThread: ThreadSelectionInfo?
+
+    /// The list of threads.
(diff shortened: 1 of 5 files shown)


Sources/StreamChatUI/ChatThreadList/ChatThreadListViewModel.swift
```diff
@@ -0,0 +1,221 @@
...
+//
+
+import Combine
+import StreamChat
+import UIKit
+
+/// The ViewModel for the `ChatThreadListView`.
+open class ChatThreadListViewModel: ObservableObject, ChatThreadListControllerDelegate, EventsControllerDelegate {
+    /// Context provided dependencies.
+    public var chatClient: ChatClient!
```
nuno-vieira, line 12: This is the view model exactly from SwiftUI, the only difference is that it does not have the Injected property wrapper

Sources/StreamChatUI/Navigation/ChatThreadListRouter.swift
```diff
@@ -33,3 +33,32 @@ open class ChatThreadListRouter: NavigationRouter<ChatThreadListVC>, ComponentsP
         }
     }
 }
+
+/// A `NavigationRouter` subclass that handles navigation actions of `ChatChannelListVC`.
+@available(iOSApplicationExtension, unavailable)
+open class ChatThreadListStatefulRouter: NavigationRouter<ChatThreadListStatefulVC>, ComponentsProvider {
```
nuno-vieira, line 39: Had to create a new router, since the previous one is coupled with `ChatThreadListC`, but this was just to make things work

nuno-vieira:
Closing this one for now. But the PoC was a success and for sure validated that we can share view models between SwiftUI and UIKit in the future.

PR #3467, opened 2024-10-18
Lock vale version to 2.1.0

testableapple:


nuno-vieira (approved):

ipavlidakis (approved):
Can we also apply it on Video?

PR #3473, opened 2024-10-25
[POC] Creating live ephemeral messages

nuno-vieira:
### 🔗 Issue Links

None

### 🎯 Goal

PoC to test how creating live ephemeral messages would work.

### How to use

The `ComposerVC` example should be simple to understand on how to use it. There 3 new `ChannelController` methods:
- `createEphemeralMessage()`: This one creates the ephemeral message ready to be updated. You need to store the messageId somewhere.
- `updateEphemeralMessage()`: This one should be used to live update the content of the message.
- `publishEphemeralMessage()`: This one should be used to publish the message to the server. 

There is also an alternative approach [here](https://github.com/GetStream/stream-chat-swift/pull/3473/commits/e59021f042fe044ec9b2f16ebf6772440512d31b) that uses an `EphemeralMessageUpdater` object, but I think for now this one is simple enough. In the future we should refine and decide the final API. We could even opt for creating a new controller or object, that handles the state of the ephemeral message, but will discuss this internally. 

## Update 
Tried a new, simpler version, which contains only 2 methods, and is easier to use:
- `channelController.ephemeralMessageEditor.updateMessage(text:)`
- `channelController.ephemeralMessageEditor.publish()`

### DIFF
--- a/Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
+++ b/Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
@@ -1315,7 +1315,7 @@ public class ChatChannelController: DataController, DelegateCallable, DataStoreP
         skipPush: Bool = false,
         skipEnrichUrl: Bool = false,
         extraData: [String: RawJSON] = [:],
-        completion: ((Result<EphemeralMessageUpdater, Error>) -> Void)? = nil
+        completion: ((Result<MessageId, Error>) -> Void)? = nil
     ) {
         /// Perform action only if channel is already created on backend side and have a valid `cid`.
         guard let cid = cid, isChannelAlreadyCreated else {
@@ -1341,19 +1341,17 @@ public class ChatChannelController: DataController, DelegateCallable, DataStoreP
             extraData: extraData
         ) { result in
             self.callback {
-                switch result {
-                case .success(let message):
-                    let updater = EphemeralMessageUpdater(messageId: message.id) { [weak self] newText in
-                        self?.updater.updateEphemeralMessage(id: message.id, text: newText)
-                    } publish: { [weak self] in
-                        self?.updater.publishEphemeralMessage(id: message.id)
-                    }
-                    completion?(.success(updater))
-                case .failure(let error):
-                    completion?(.failure(error))
-                }
+                completion?(result.map(\.id))
             }
         }
+    }
+
+    public func updateEphemeralMessage(id: MessageId, text: String) {
+        updater.updateEphemeralMessage(id: id, text: text)
+    }
+    
+    public func publishEphemeralMessage(id: MessageId) {
+        updater.publishEphemeralMessage(id: id)
     }
 
     deinit {
--- a/Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
+++ b/Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
@@ -1304,6 +1304,58 @@ public class ChatChannelController: DataController, DelegateCallable, DataStoreP
         }
     }
 
+    public func createEphemeralMessage(
+        messageId: MessageId? = nil,
+        text: String,
+        pinning: MessagePinning? = nil,
+        isSilent: Bool = false,
+        attachments: [AnyAttachmentPayload] = [],
+        mentionedUserIds: [UserId] = [],
+        quotedMessageId: MessageId? = nil,
+        skipPush: Bool = false,
+        skipEnrichUrl: Bool = false,
+        extraData: [String: RawJSON] = [:],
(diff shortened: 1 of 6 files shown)


Sources/StreamChatUI/Composer/ComposerVC.swift
```diff
@@ -1445,6 +1442,15 @@ open class ComposerVC: _ViewController,
...
         content.text = textView.text
+
+        guard let ephemeralMessageId = self.ephemeralMessageId else {
+            channelController?.createEphemeralMessage(text: content.text) { [weak self] result in
+                self?.ephemeralMessageId = try? result.get()
+            }
+            return
+        }
+
+        channelController?.updateEphemeralMessage(id: ephemeralMessageId, text: content.text)
```
nuno-vieira, line 1453: When updating the text, we need to check if the ephemeral message is already created. If not, we create it, if yes, we update it.

Hopefully we can improve this, and handle this internally.

Sources/StreamChatUI/Composer/ComposerVC.swift
```diff
@@ -936,15 +938,10 @@ open class ComposerVC: _ViewController,
...
-            attachments: content.attachments,
-            mentionedUserIds: content.mentionedUsers.map(\.id),
-            quotedMessageId: content.quotingMessage?.id,
-            skipEnrichUrl: content.skipEnrichUrl,
-            extraData: content.extraData
-        )
+        if let ephemeralMessageId {
+            channelController?.publishEphemeralMessage(id: ephemeralMessageId)
+            self.ephemeralMessageId = nil
+        }
```
nuno-vieira, line 944: When pressing publish, we simply call publish on the ephemeral message

Sources/StreamChat/Workers/ChannelUpdater.swift
```diff
@@ -344,6 +344,74 @@ class ChannelUpdater: Worker {
...
+                skipPush: skipPush,
+                skipEnrichUrl: skipEnrichUrl,
+                poll: poll,
+                extraData: extraData
+            )
+            if quotedMessageId != nil {
+                newMessageDTO.showInsideThread = true
+            }
+            newMessageDTO.type = MessageType.regular.rawValue
+            newMessageDTO.localMessageState = .sending
```
nuno-vieira, line 389: For testing proposes, I've used the `.sending` type, this ofc will have the Sending UI, but this is easily changed if we add a new state or a new property. This should also be discussed in the offical API.

Sources/StreamChatUI/Composer/ComposerVC.swift
nuno-vieira: New WIP simpler version, does not require all this logic.

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
```diff
@@ -1313,6 +1365,34 @@ public class ChatChannelController: DataController, DelegateCallable, DataStoreP
     }
 }
 
+public class EphemeralMessageEditor {
+    private unowned let channelController: ChatChannelController
```
nuno-vieira, line 1369: Ideally, in the official API, this would use the channel updater and not the controller directly

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
```diff
@@ -1313,6 +1365,34 @@ public class ChatChannelController: DataController, DelegateCallable, DataStoreP
...
+        channelController.createEphemeralMessage(text: text) {
+            self.messageId = try? $0.get()
+            group.leave()
+        }
+    }
+
+    public func publish() {
+        messageId.map { channelController.publishEphemeralMessage(id: $0) }
+        messageId = nil
+    }
```
nuno-vieira, line 1393: It could be useful to know when it is published, for that we could the use our new `waitForAPIRequest`.

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
```diff
@@ -1313,6 +1365,34 @@ public class ChatChannelController: DataController, DelegateCallable, DataStoreP
     }
 }
 
+public class EphemeralMessageEditor {
+    private unowned let channelController: ChatChannelController
+    private var messageId: MessageId?
```
nuno-vieira, line 1370: This cold potentially be a dictionary, so that we could support having multiple live ephemeral messages

martinmitrevski (commented):
@nuno-vieira let's clean up this PR and prepare it for the release in 2 weeks. I think the API can be simplified, maybe we can hide the ephemeral message editor from customers and expose things directly from the controller?

martinmitrevski (commented):
Actually, we will park this for now, we will take a bit of different direction

Issue #3338, opened 2024-07-28
WebSocket Does Not Reconnect When App Resumes from Background (Guest User)


masarusanjp:
## What did you do?

I launched the app, waited for a few minutes, and then resumed it.

## What did you expect to happen?

I expected the WebSocket to reconnect when the app resumes from the background.

## What happened instead?

The WebSocket does not reconnect when the app resumes from the background.

## GetStream Environment
**GetStream Chat version:** 4.60.0
**GetStream Chat frameworks:** StreamChat, StreamChatUI
**iOS version:** iOS17.4
**Swift version:** 5.9
**Xcode version:** 15.3
**Device:** Simulator iPhone 15 Pro

### Reproduce Step

- launch DemoApp 
- Tap `Guest User`
- Suspend app
- Waiting for few minutes
- Resume app
- Tap `Start a chat`

### Screencast

https://drive.google.com/file/d/1-parw6tuS02VBZW7cG3__1Pb4e3ciogL/view?usp=sharing
(Here is a link to GoogleDrive due to its large size)

### Additional Information

I discovered that the connectGuestUser function calls logOut first. As a result, ConnectionRecoveryHandler.stop is invoked, causing the monitoring of appDidBecomeActive and appDidEnterBackground to stop. This leads to the issue where "the WebSocket does not reconnect when the app resumes from the background.

https://github.com/GetStream/stream-chat-swift/pull/3303/files#diff-5aa17a3514ab85c33231ca8983e4c32096742c79d97924fe31e71ebb0ac7d540R385

This issue reliably reproduces with a guest user, but it is likely to occur under any condition where logOutFirst is true.

nuno-vieira:
Hi @masarusanjp, thanks for spotting this issue!

The actual issue is that when reconnecting, logOut should not be happening, so the logic should be changed here. Stopping the recovery handler on logout is something that we want, but it caused this unwanted side effect due to the fact that logOutFirst in guest user is not properly implemented it seems.

98prabowo:
Hi @nuno-vieira, does this only affect guest user? or all user?

nuno-vieira:
> Hi @nuno-vieira, does this only affect guest user? or all user?

Hello! Only guest and anonymous user yes.

nuno-vieira:
This is now fixed on the latest release! 🚀 

Thank you for your patience!

PR #3483, opened 2024-11-07
Fix connection not resuming after guest user goes to background

nuno-vieira:
### 🔗 Issue Links

https://github.com/GetStream/stream-chat-swift/issues/3338
https://linear.app/stream/issue/IOS-75/[reconnection]-[github]-websocket-disconnectes-when-guest-user-comes

### 🎯 Goal

Fixes connection not resuming after guest user goes to background

### 📝 Summary

- Fixes connection not resuming after guest user goes to background
- Improves Demo App Token Refreshing UX

### 🛠 Implementation

The problem was that in a previous PR, we introduced stopping the recovery handler whenever a user disconnects... But this is an issue because whenever the user comes from the background and connects again, it will try to "logout" the user, and so it will stop the recovery handle before the connection.

This is at the moment the simplest solution. The con, is that the recovery handler will keep running, even after disconnection, but it s not a big deal (and it was working like this since the begining) 

### 🧪 Manual Testing Notes

- Open Demo App Configuration
- Disable `staysConnectedInBackground`
- Login with Guest User
- Go to background, wait 5s to make sure everything is disconnected
- Go back to foreground
- Connection should resume

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/ChatClient.swift
+++ b/Sources/StreamChat/ChatClient.swift
@@ -456,7 +456,6 @@ public class ChatClient {
     /// Disconnects the chat client from the chat servers. No further updates from the servers
     /// are received.
     public func disconnect(completion: @escaping () -> Void) {
-        connectionRecoveryHandler?.stop()
         connectionRepository.disconnect(source: .userInitiated) {
             log.info("The `ChatClient` has been disconnected.", subsystems: .webSocket)
             completion()
--- a/DemoApp/Screens/AppConfigViewController/AppConfigViewController.swift
+++ b/DemoApp/Screens/AppConfigViewController/AppConfigViewController.swift
@@ -28,11 +28,10 @@ struct DemoAppConfig {
         /// The app secret from the dashboard.
         let appSecret: String
         /// The duration in seconds until the token is expired.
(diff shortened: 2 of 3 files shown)


Sources/StreamChat/ChatClient.swift
```diff
@@ -456,7 +456,6 @@ public class ChatClient {
     /// Disconnects the chat client from the chat servers. No further updates from the servers
     /// are received.
     public func disconnect(completion: @escaping () -> Void) {
-        connectionRecoveryHandler?.stop()
```
laevandus, line 459: Seems to be safe because recovery handler checks for current connection state and does not attempt to reconnect  if the state is `disconnected` with `userInitiated`. 🤔 


```
var canReconnectFromOffline: Bool {
        …
        switch webSocketClient.connectionState {
        case .disconnected(let source) where source == .userInitiated:
            return false
```

nuno-vieira: Yes exactly the reconnection handler although alive, it won't do nothing

martinmitrevski (approved):

PR #3495, opened 2024-11-18
Use Core Data prefetching when StreamRuntimeCheck._isDatabasePrefetchingEnabled

laevandus:
### 🔗 Issue Links

Part of [IOS-23](https://linear.app/stream/issue/IOS-23)

### 🎯 Goal

- Add a feature flag for turning on data prefetching on the Core Data level

### 📝 Summary

`StreamRuntimeCheck._isDatabasePrefetchingEnabled` turns on setting:
- `returnsObjectsAsFaults` to false
- Populating `relationshipKeyPathsForPrefetching`

### 🛠 Implementation

There have been crash reports happening when Core Data is fulfilling a fault. Instruments shows that some of the fault related requests take very long time sometimes. If the runtime check is enabled, these slow requests are not reproducible anymore. The aim of PR is to add this feature flag and see if it remedies these issues. Moreover, it makes sense to use this configuration because most of our `asModel()` functions end up accessing all the data of the Core Data model (including many relationships). For such cases, Apple also recommends setting faulting to false and using prefetching ([1](https://developer.apple.com/documentation/coredata/nsfetchrequest/1506756-returnsobjectsasfaults), [2](https://developer.apple.com/documentation/coredata/nsfetchrequest/1506813-relationshipkeypathsforprefetchi)).

Tried our demo app and saw 25% reduction (4004 to 3034) of fault related requests (even when requests created by us have returnsObjectsAsFaults=false and relationship key paths set, it does remove all the fault related requests. Core Data internally makes many requests as well when merging contexts etc).  

Example of the slow fault related request:
slow_fault

### 🧪 Manual Testing Notes

Set `StreamRuntimeChecks._isDatabasePrefetchingEnabled = true` and play around with the demo app. Does anything feel off?

### ☑️ Contributor Checklist

### DIFF
--- a/StreamChat.xcodeproj/project.pbxproj
+++ b/StreamChat.xcodeproj/project.pbxproj
@@ -3174,7 +3174,6 @@
 		43F4750B26F4E4FF0009487D /* ChatMessageReactionItemView.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ChatMessageReactionItemView.swift; sourceTree = "<group>"; };
 		43F4750D26FB247C0009487D /* ChatReactionPickerReactionsView.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ChatReactionPickerReactionsView.swift; sourceTree = "<group>"; };
 		4A4E184528D06CA30062378D /* Documentation.docc */ = {isa = PBXFileReference; lastKnownFileType = folder.documentationcatalog; path = Documentation.docc; sourceTree = "<group>"; };
-		4A51230029D3170C005CEA9B /* docusaurus */ = {isa = PBXFileReference; lastKnownFileType = folder; path = docusaurus; sourceTree = "<group>"; };
 		4F05C0702C8832C40085B4B7 /* URLRequest+cURL.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = "URLRequest+cURL.swift"; sourceTree = "<group>"; };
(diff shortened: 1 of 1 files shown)


Sources/StreamChat/Utils/Database/NSManagedObject+Extensions.swift
```diff
@@ -33,20 +33,47 @@ extension NSFetchRequestGettable where Self: NSManagedObject {
...
+extension NSFetchRequestPrefetching where Self: NSManagedObject {
+    /// Turns off Core Data object faulting and sets prefetched relationship keypaths.
+    ///
+    /// Note: Reduces additional Core Data fetches when most of the data is accessed from fetched objects.
+    static func applyPrefetchingState(to request: NSFetchRequest<Self>) {
+        guard StreamRuntimeCheck._isDatabasePrefetchingEnabled else { return }
+        request.returnsObjectsAsFaults = false
+        request.relationshipKeyPathsForPrefetching = Self.prefetchedRelationshipKeyPaths()
+    }
+}
```
laevandus, line 57: This method is not in the `NSFetchRequest` extension because extending generic Obj-C classes is complicated.

It would be nicer to write `request.applyPrefetchingState()`, but did not find any good way for it which enables each subclass to define its relationship key paths. Moreover, load methods in this file make it even more complicated.

nuno-vieira (approved):
LGTM ✅  Lets test drive this and see if there are any improvements 👍

martinmitrevski (commented):
What about the increase of hitches in the checks, do you think there might be a performance impact? Although, by default this thing is off, so weird that we get this regression, maybe something with the check itself @testableapple?
In any case, this should be tested a bit with the flag on.

PR #3497, opened 2024-11-19
Fix converting deleted database objects

laevandus:
### 🔗 Issue Links

Part of [IOS-23](https://linear.app/stream/issue/IOS-23)
Resolves [IOS-20](https://linear.app/stream/issue/IOS-20)

### 🎯 Goal

Fix issues around deleting objects in the local database

### 📝 Summary

- Skip converting deleted database objects which lead to model objects returning empty properties
- Do not delete muted users when the current user object is deleted which leads to channel members having inconsistent state (member's user data is empty)
- Skip willSave if the DTO is in deleted state

### 🛠 Implementation

There have been crashes around accessing member data of the channel objects and one the possible culprits seems to be the automatic user deletion rule for the current user. User objects are shared between many objects and therefore should not be deleted like that. In addition, when an object is deleted, we should skip trying to convert these objects because the data is not present anymore (e.g. one can get a `ChatMessage` object with no properties being set, id is empty etc). That said, it is pretty rare since deleting data is not running often.

### 🧪 Manual Testing Notes

Running the demo app and doing some smoke testing which also includes logging out the current user.

### ☑️ Contributor Checklist


Sources/StreamChat/Repositories/PollsRepository.swift
```diff
@@ -144,12 +144,18 @@ class PollsRepository {
...
+        var userId: UserId?
         var filterHash: String?
         database.write { session in
             let voteDto = try session.removePollVote(with: voteId, pollId: pollId)
+            exists = voteDto != nil
             filterHash = voteDto?.queries?.first?.filterHash
-            pollVote = try voteDto?.asModel()
+            answerText = voteDto?.answerText
+            optionId = voteDto?.optionId
+            userId = voteDto?.user?.id
```
laevandus, line 158: Skipped `asModel()` here because now it requires the DTO to not be in the deleted state.

Tests/StreamChatTests/Database/DTOs/CurrentUserDTO_Tests.swift
```diff
@@ -290,4 +290,47 @@ final class CurrentUserModelDTO_Tests: XCTestCase {
         XCTAssertEqual(true, loadedCurrentUser.privacySettings.readReceipts?.enabled)
         XCTAssertEqual(true, loadedCurrentUser.privacySettings.typingIndicators?.enabled)
     }
+    
+    func test_deletingCurrentUser() throws {
```
laevandus, line 294: Covering the cascade deletion change. If the rule is cascade for mutedUsers, then this test fails and the channel's member ids are: "", "current_user_id"

Sources/StreamChat/Database/DTOs/ChannelReadDTO.swift
```diff
@@ -18,7 +18,10 @@ class ChannelReadDTO: NSManagedObject {
 
     override func willSave() {
         super.willSave()
-
+        
+        guard !isDeleted else {
+            return
+        }
```
laevandus, line 24: Most of the DTOs have this check, but found some which did not.

Sources/StreamChat/Repositories/PollsRepository.swift
nuno-vieira: Can you clarify this part a bit? I did not fully understood it

nuno-vieira: Ah nevermind got it 👍

nuno-vieira (approved):
LGTM! ✅

laevandus:
QAed with @testableapple

Sources/StreamChat/Database/DTOs/PollOptionDTO.swift
```diff
@@ -58,7 +58,7 @@ extension PollOptionDTO {
 
 extension PollOptionDTO {
     func asModel() throws -> PollOption {
-        guard !isDeleted else { throw ClientError.DeletedModel(self) }
+        guard !isDeleted else { throw DeletedModel(self) }
```
nuno-vieira, line 61: Couldn't add a method to all DTOs, like: `try isNotDeleted()` ?  So that don't repeat this guard everywhere 🤔

laevandus: Good idea

PR #3500, opened 2024-11-19
Cache NSManagedObjectID instead of the CurrentUserDTO

laevandus:
### 🔗 Issue Links

Resolves [IOS-21](https://linear.app/stream/issue/IOS-21)

### 🎯 Goal

Cache `NSManagedObjectID` instead of the full `CurrentUserDTO` for increased reliability. 

### 📝 Summary

- Cache `NSManagedObjectID` which is less susceptible to possible threading issues (although I haven't found out that his is actually the case, but something seems to have corrupted the DTO in rare cases)

### 🛠 Implementation

We have seen a crash when NSManagedObjectContext's currentUser property is accessed which leads to a crash when the `user.id` is called. This PR tries to fix this issue by not keeping the DTO alive all the time and instead asking context to retrieve it instead of (which is not slow).
I measured the time taken for accessing the `currentUser` property and when all the call sites are combined, it was 1 ms.

### 🧪 Manual Testing Notes

Quick testing round of logging in and out

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/Repositories/AuthenticationRepository.swift
+++ b/Sources/StreamChat/Repositories/AuthenticationRepository.swift
@@ -123,13 +123,8 @@ class AuthenticationRepository {
     func fetchCurrentUser() {
         var currentUserId: UserId?
 
-        let context = databaseContainer.viewContext
-        if Thread.isMainThread {
-            currentUserId = context.currentUser?.user.id
-        } else {
-            context.performAndWait {
-                currentUserId = context.currentUser?.user.id
-            }
+        databaseContainer.backgroundReadOnlyContext.performAndWait {
+            currentUserId = databaseContainer.backgroundReadOnlyContext.currentUser?.user.id
         }
         self.currentUserId = currentUserId
     }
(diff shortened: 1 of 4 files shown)


Sources/StreamChat/Repositories/AuthenticationRepository.swift
```diff
@@ -123,13 +123,8 @@ class AuthenticationRepository {
...
-        if Thread.isMainThread {
-            currentUserId = context.currentUser?.user.id
-        } else {
-            context.performAndWait {
-                currentUserId = context.currentUser?.user.id
-            }
+        databaseContainer.backgroundReadOnlyContext.performAndWait {
+            currentUserId = databaseContainer.backgroundReadOnlyContext.currentUser?.user.id
         }
         self.currentUserId = currentUserId
```
laevandus, line 129: This looked odd that we have such check here, therefore I changed it around.

nuno-vieira (approved):
LGTM! ✅ Much better approach 👌

martinmitrevski (approved):
LGTM ✅  There are some failing tests though that need checking.

laevandus:
QA was done with @testableapple

PR #3502, opened 2024-11-20
Replace shared current user observer used for reading privacy settings

laevandus:
### 🔗 Issue Links

Part of [IOS-558](https://linear.app/stream/issue/IOS-558)

### 🎯 Goal

Remove current user controller since it takes unnecessary CPU time when reacting to DB changes

### 📝 Summary

- Replace internal current user controller with just reading the data directly

### 🛠 Implementation

We had a perf issue some time ago where we used to create the `ChatCurrentUserController` every single time when checking for typing state (on each key press). Then we started to use a shared instance for that. Solved the issue, but there is one downside, reacting to current user changes can get expensive (e.g. current user has many muted channels which triggers recreating channel objects when updating the immutable `CurrentChatUser` type). Instead all of this, we should just read the data we need and skip controllers.

### 🧪 Manual Testing Notes

Try typing events, if disabled, should not appear, otherwise appear.

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
+++ b/Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
@@ -151,7 +151,11 @@ public class ChatChannelController: DataController, DelegateCallable, DataStoreP
             completion(false)
             return
         }
-        eventSender.currentUserTypingEventsState(completion: completion)
+        eventSender.database.read { session in
+            session.currentUser?.isTypingIndicatorsEnabled ?? true
+        } completion: { result in
+            completion(result.value ?? false)
+        }
     }
 
     /// Set the delegate of `ChannelController` to observe the changes in the system.
--- a/Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
+++ b/Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
@@ -146,12 +146,12 @@ public class ChatChannelController: DataController, DelegateCallable, DataStoreP
 
     /// A boolean value indicating if it should send typing events.
     /// It is `true` if the channel typing events are enabled as well as the user privacy settings.
-    internal var shouldSendTypingEvents: Bool {
-        /// Ignore if user typing indicators privacy settings are disabled. By default, they are enabled.
-        let currentUserPrivacySettings = client.sharedCurrentUserController.currentUser?.privacySettings
-        let isTypingIndicatorsForCurrentUserEnabled = currentUserPrivacySettings?.typingIndicators?.enabled ?? true
-        let isChannelTypingEventsEnabled = channel?.canSendTypingEvents ?? true
-        return isTypingIndicatorsForCurrentUserEnabled && isChannelTypingEventsEnabled
+    func shouldSendTypingEvents(completion: @escaping (Bool) -> Void) {
+        guard channel?.canSendTypingEvents ?? true else {
+            completion(false)
+            return
+        }
+        eventSender.currentUserTypingEventsState(completion: completion)
     }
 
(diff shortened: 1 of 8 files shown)


Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
```diff
@@ -146,12 +146,12 @@ public class ChatChannelController: DataController, DelegateCallable, DataStoreP
...
-        let currentUserPrivacySettings = client.sharedCurrentUserController.currentUser?.privacySettings
-        let isTypingIndicatorsForCurrentUserEnabled = currentUserPrivacySettings?.typingIndicators?.enabled ?? true
-        let isChannelTypingEventsEnabled = channel?.canSendTypingEvents ?? true
-        return isTypingIndicatorsForCurrentUserEnabled && isChannelTypingEventsEnabled
+    func shouldSendTypingEvents(completion: @escaping (Bool) -> Void) {
+        guard channel?.canSendTypingEvents ?? true else {
+            completion(false)
+            return
+        }
+        eventSender.currentUserTypingEventsState(completion: completion)
```
nuno-vieira, line 154: Could we do this instead? 🤔 The DataStore has access to the current user, so it is probably the best approach in this scenario? It will simplify the code as well
```suggestion
        let currentUserPrivacySettings = dataStore.currentUser().privacySettings
        let isTypingIndicatorsForCurrentUserEnabled = currentUserPrivacySettings?.typingIndicators?.enabled ?? true
        let isChannelTypingEventsEnabled = channel?.canSendTypingEvents ?? true
        return isTypingIndicatorsForCurrentUserEnabled && isChannelTypingEventsEnabled
```

nuno-vieira, line 154: This new logic also does not seem correct. The privacy settings should be evaluate first, like it was before 🤔

laevandus: It is going to be expensive because `dataStore.currentUser()` fetches the DTO, then converts it to the full immutable type.

Sources/StreamChat/Workers/TypingEventsSender.swift
```diff
@@ -39,6 +39,19 @@ class TypingEventsSender: Worker {
...
+            session.currentUser?.isTypingIndicatorsEnabled ?? true
+        } completion: { result in
+            switch result {
+            case .success(let isEnabled):
+                completion(isEnabled)
+            case .failure:
+                completion(false)
+            }
+        }
+    }
```
nuno-vieira, line 54: Like I said above, I think accessing the dataStore.currentUser() directly, should be enough. WDYT?

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
nuno-vieira: I see...

laevandus: Old implementation:
```
let currentUserPrivacySettings = client.sharedCurrentUserController.currentUser?.privacySettings
        let isTypingIndicatorsForCurrentUserEnabled = currentUserPrivacySettings?.typingIndicators?.enabled ?? true
        let isChannelTypingEventsEnabled = channel?.canSendTypingEvents ?? true
        return isTypingIndicatorsForCurrentUserEnabled && isChannelTypingEventsEnabled
```
It can be true only when both privacy settings and channel capabilities are available, therefore, the functionality should be the same. 🤔

nuno-vieira (approved):
LGTM! ✅ It complicates a bit the code, but for now there's no alternative 👍

laevandus:
QAed with @testableapple

PR #3508, opened 2024-11-26
Convert to models after NSFetchedResultsController finishes reporting changes

laevandus:
### 🔗 Issue Links

Resolves [IOS-569](https://linear.app/stream/issue/IOS-569)

### 🎯 Goal

Fix a crash caused by infinite loop in Core Data when converting models

### 📝 Summary

- Delay converting models to the point of time when `NSFetchedRequestController` has finished reporting changes

### 🛠 Implementation

We had an internal crash report which showed that Core Data change came in, triggered `NSFetchedRequestController` delegate, then we converted the DTO to a model type, but since the model type did another fetch request, this triggered the whole cycle again. Therefore, let's avoid it by delaying the model conversion to the point of time where `NSFetchedRequestController` has finished reporting changes.

### 🧪 Manual Testing Notes

Retrying the flow in IOS-569 which triggered it.

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/Controllers/DatabaseObserver/ListChange.swift
+++ b/Sources/StreamChat/Controllers/DatabaseObserver/ListChange.swift
@@ -150,9 +150,8 @@ class ListChangeAggregator<DTO: NSManagedObject, Item>: NSObject, NSFetchedResul
     /// Called with the aggregated changes after `FetchResultsController` calls controllerDidChangeContent` on its delegate.
     var onDidChange: (([ListChange<Item>]) -> Void)?
 
-    /// An array of changes in the current update. It gets reset every time `controllerWillChangeContent` is called, and
-    /// published to the observer when `controllerDidChangeContent` is called.
-    private var currentChanges: [ListChange<Item>] = []
+    /// An array of changes in the current update.
+    private var currentChanges: [ListChange<DTO>] = []
 
     /// Creates a new `ChangeAggregator`.
     ///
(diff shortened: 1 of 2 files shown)


Sources/StreamChat/Controllers/DatabaseObserver/ListChange.swift
```diff
@@ -189,35 +190,54 @@ class ListChangeAggregator<DTO: NSManagedObject, Item>: NSObject, NSFetchedResul
...
 
         default:
             break
         }
     }
 
     func controllerDidChangeContent(_ controller: NSFetchedResultsController<NSFetchRequestResult>) {
-        onDidChange?(currentChanges)
+        // Model conversion is safe when all the changes have been processed (Core Data's _processRecentChanges can be called if conversion triggers additional fetch requests).
+        let itemChanges = currentChanges.compactMap { dtoChange in
```
nuno-vieira, line 223: This makes sense. Does this potentially fix the current workaround we have the `depth` stuff?

nuno-vieira (approved):
Nice find 👌

nuno-vieira:
I've QA it myself, it seems to work fine 👍 At least I was not able to reproduce the crash anymore.

Sources/StreamChat/Controllers/DatabaseObserver/ListChange.swift
laevandus: Unfortunately no because the circular cycle can still occur with MessageDTO's replies contains a quoted message.

PR #3513, opened 2024-11-28
Fix reconnection timeout handler not working in the token provider phase

nuno-vieira:
### 🔗 Issue Links
Fixes https://linear.app/stream/issue/IOS-78

### 🎯 Goal

Fixes the reconnection timeout handler not working in the token provider phase.

### 🛠 Implementation

The main change is that the `reconnectionTimeoutHandler` was moved from the `ConnectionRecoveryHandler` to the `ChatClient` so that it can handle the token provider phase. For this to work, a couple of things required some changes. There are comments in the PR explaining each change.

Flow diagram to explain why we need to set the state to `.initialize` again whenever we call connect in `ChatClient`:

### Before
```mermaid
stateDiagram-v2
    [*] --> ChatClient.init
    ChatClient.init--> ChatClient.Connect
    ChatClient.Connect --> TokenProvider
    
    TokenProvider --> ReconnectionTimeout: Fails
    TokenProvider --> WebSocket.Connect: Success
    
     ReconnectionTimeout --> disconnected
     disconnected --> ChatClient.Connect
    
    note right of WebSocket.Connect
        State reported: .connecting
    end note
    
    note right of ChatClient.init
        State reported: .initialize
    end note
    
    note right of disconnected
        State reported: .disconnected
    end note
    
    note left of disconnected
        On the second cycle, the state is not reported
        Because it did not change from .disconnected
    end note
```

### After
```mermaid
stateDiagram-v2
    [*] --> ChatClient.init
    ChatClient.init--> ChatClient.Connect
    ChatClient.Connect --> TokenProvider
    
    TokenProvider --> ReconnectionTimeout: Fails
    TokenProvider --> WebSocket.Connect: Success
    
     ReconnectionTimeout --> disconnected
     disconnected --> ChatClient.Connect
    
    note right of WebSocket.Connect
        State reported: .connecting
    end note
    
    note right of ChatClient.Connect
        State reported: .initialize
    end note
    
    note right of disconnected
        State reported: .disconnected
    end note
```

### 🧪 Manual Testing Notes
1. Open the Demo App Configuration
2. Setup token refresh details and make it fail, for example, 20 times
3. Set a reconnection timeout of 15 seconds
4. Connect a user
5. Wait 15 seconds
6. Should automatically disconnect
7. Should not report anymore "Token refresh failed" in the console

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/ChatClient.swift
+++ b/Sources/StreamChat/ChatClient.swift
@@ -96,6 +96,9 @@ public class ChatClient {
     /// Used as a bridge to communicate between the host app and the notification extension. Holds the state for the app lifecycle.
     let extensionLifecycle: NotificationExtensionLifecycle
 
+    /// The component responsible to timeout the user connection if it takes more time than the `ChatClientConfig.reconnectionTimeout`.
+    var reconnectionTimeoutHandler: StreamTimer?
+
     /// The environment object containing all dependencies of this `Client` instance.
     private let environment: Environment
     
@@ -219,12 +222,18 @@ public class ChatClient {
         setupOfflineRequestQueue()
         setupConnectionRecoveryHandler(with: environment)
         validateIntegrity()
+
+        reconnectionTimeoutHandler = environment.reconnectionHandlerBuilder(config)
+        reconnectionTimeoutHandler?.onChange = { [weak self] in
+            self?.timeout()
+        }
     }
 
     deinit {
         Self._activeLocalStorageURLs.mutate { $0.subtract(databaseContainer.persistentStoreDescriptions.compactMap(\.url)) }
         completeConnectionIdWaiters(connectionId: nil)
         completeTokenWaiters(token: nil)
+        reconnectionTimeoutHandler?.stop()
     }
 
     func setupTokenRefresher() {
@@ -254,8 +263,7 @@ public class ChatClient {
             extensionLifecycle,
             environment.backgroundTaskSchedulerBuilder(),
             environment.internetConnection(eventNotificationCenter, environment.internetMonitor),
-            config.staysConnectedInBackground,
-            config.reconnectionTimeout.map { ScheduledStreamTimer(interval: $0, fireOnStart: false, repeats: false) }
+            config.staysConnectedInBackground
         )
     }
     
@@ -300,7 +308,9 @@ public class ChatClient {
         tokenProvider: @escaping TokenProvider,
         completion: ((Error?) -> Void)? = nil
     ) {
+        reconnectionTimeoutHandler?.start()
         connectionRecoveryHandler?.start()
+        connectionRepository.initialize()
 
         authenticationRepository.connectUser(
             userInfo: userInfo,
@@ -393,7 +403,9 @@ public class ChatClient {
         userInfo: UserInfo,
         completion: ((Error?) -> Void)? = nil
     ) {
+        connectionRepository.initialize()
         connectionRecoveryHandler?.start()
+        reconnectionTimeoutHandler?.start()
         authenticationRepository.connectGuestUser(userInfo: userInfo, completion: { completion?($0) })
     }
     
@@ -417,6 +429,8 @@ public class ChatClient {
     /// Connects an anonymous user
     /// - Parameter completion: The completion that will be called once the **first** user session for the given token is setup.
     public func connectAnonymousUser(completion: ((Error?) -> Void)? = nil) {
+        connectionRepository.initialize()
+        reconnectionTimeoutHandler?.start()
         connectionRecoveryHandler?.start()
         authenticationRepository.connectAnonymousUser(
             completion: { completion?($0) }
@@ -458,7 +472,7 @@ public class ChatClient {
             completion()
         }
         authenticationRepository.clearTokenProvider()
-        authenticationRepository.cancelTimers()
+        authenticationRepository.reset()
     }
 
     /// Disconnects the chat client from the chat servers. No further updates from the servers
@@ -617,6 +631,16 @@ public class ChatClient {
             completion?($0)
         }
     }
+
+    private func timeout() {
+        completeConnectionIdWaiters(connectionId: nil)
+        authenticationRepository.completeTokenCompletions(error: ClientError.ReconnectionTimeout())
+        completeTokenWaiters(token: nil)
+        authenticationRepository.reset()
+        connectionRecoveryHandler?.stop()
+        let webSocketConnectionState = webSocketClient?.connectionState ?? .initialized
+        connectionRepository.disconnect(source: .timeout(from: webSocketConnectionState)) {}
+    }
 }
 
 extension ChatClient: AuthenticationRepositoryDelegate {
@@ -646,6 +670,17 @@ extension ChatClient: ConnectionStateDelegate {
         )
         connectionRecoveryHandler?.webSocketClient(client, didUpdateConnectionState: state)
         try? backgroundWorker(of: MessageSender.self).didUpdateConnectionState(state)
+
+        switch state {
+        case .connecting:
+            if reconnectionTimeoutHandler?.isRunning == false {
+                reconnectionTimeoutHandler?.start()
+            }
+        case .connected:
+            reconnectionTimeoutHandler?.stop()
+        default:
+            break
+        }
     }
 }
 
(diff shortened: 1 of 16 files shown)


Sources/StreamChat/Repositories/ConnectionRepository.swift
```diff
@@ -42,6 +42,10 @@ class ConnectionRepository {
         self.timerType = timerType
     }
 
+    func initialize() {
+        webSocketClient?.initialize()
+    }
```
nuno-vieira, line 47: This needs to be introduced. Otherwise, we won't report the status of the new connection to the delegates. This is especially important when using ChatClient as a singleton. This is the problem:

1. `connectUser()` (`connectionStatus = .initialized`)
2. `timeout()` (connectinStatus = .disconnected) - Disconnected status is reported ✅ 
3. `connectUser()` again (`connectionStatus == .disconnected`) - Connecting status is not reported ❌ , because the status did not change

Since we don't reset the `connection status`, it keeps at the `.disconnected`, and so the `.connecting` or `.initialized` is not reported when manually reconnecting.

Sources/StreamChat/Repositories/ConnectionRepository.swift
```diff
@@ -95,14 +99,6 @@ class ConnectionRepository {
...
         }
 
-        if connectionId == nil {
-            if source == .userInitiated {
-                log.warning("The client is already disconnected. Skipping the `disconnect` call.")
-            }
-            completion()
-            return
-        }
-
```
nuno-vieira, line 105: This is an unnecessary optimization that just complicates things. For example, when we timeout, we don't have the connectionId, so the webSocketClient? .disconnect won't be called. Which means no connection status changes will be reported.

Sources/StreamChat/WebSocketClient/WebSocketClient.swift
```diff
@@ -137,23 +141,18 @@ class WebSocketClient {
...
-        connectionState = .disconnecting(source: source)
-        engineQueue.async { [engine, eventsBatcher] in
-            engine?.disconnect()
-
-            eventsBatcher.processImmediately(completion: completion)
+        switch connectionState {
+        case .initialized, .disconnected, .disconnecting:
+            connectionState = .disconnected(source: source)
+        case .connecting, .waitingForConnectionId, .connected:
+            connectionState = .disconnecting(source: source)
```
nuno-vieira, line 148: when disconnecting, if we are already disconnected, we should not put the state to disconnecting since the engine won't do anything, and so it won't report the disconnected afterwards. For this reason, we need to instantly report as disconnected for this scenario.

Sources/StreamChat/Repositories/ConnectionRepository.swift
martinmitrevski: when you call `connectUser` again, shouldn't it move to `initialized` / `connecting` already? Because there should be a status change when you start connecting again.

Sources/StreamChat/Repositories/ConnectionRepository.swift
martinmitrevski: yes, this makes sense.

Sources/StreamChat/WebSocketClient/WebSocketClient.swift
martinmitrevski: this seems tricky, previously it seems we were also processing events. Can you explain a bit more why we delete that part? 
Also, the switch itself is a bit strange to me, especially the second case. If you are disconnected, and you are connecting, why should it go disconnecting?

nuno-vieira: I did not change the processing events part. The github diffing is not very clear, but the only thing changed here is the report of `connectionState`

> Also, the switch itself is a bit strange to me, especially the second case. If you are disconnected, and you are connecting, why should it go disconnecting?
Where does this happen? We report disconnected if we were not connected previously, so theres nothing to disconnect, that is why we instantly report that we are disconnected. If we are connecting, or waiting for connection or connected, then we need to call disconnect to the engine, so the engine will eventually report the disconnected state.

nuno-vieira: It is quite clear actually:

When the state is `initialized`, `.disconnected`, and `disconnecting`, and the user disconnects, we instantly report `disconnected` since we don't need to disconnect the websocket engine (it is already disconnected).

When the state is `connected`, `connecting` or `waiting for connection`, we need to call the `engine.disconnect()` and the websocket delegate will report the `disconnected` state. This is why here we report `disconnecting` instead of `disconnected`.

Sources/StreamChat/Repositories/ConnectionRepository.swift
nuno-vieira: I've added more details to point `1.`. The problem here is that we only set `connectionState = .connecting` after the token phase finishes. So, if the token provider keeps failing, we never set the websocket to `connecting` and it will be kept `disconnected` on the second try. And so, the connection status is not reported ever

nuno-vieira:
```mermaid
stateDiagram-v2
    [*] --> initialize
    note right of initialize
        State reported: .initialize
    end note
    
    initialize --> Connect
    Connect --> TokenProvider
    TokenProvider --> ReconnectionTimeout

     note right of TokenProvider
        State not changed to .connecting
        Because the websocket engine is not called
    end note
    
    ReconnectionTimeout --> disconnected
    note right of disconnected
        State reported: .disconnected
    end note
    
    disconnected --> Connect2
    Connect2 --> TokenProvider2
    TokenProvider2 --> ReconnectionTimeout2
    ReconnectionTimeout2 --> disconnected2
    
    note right of disconnected2
        State NOT reported
        Because it did not change from .disconnected
    end note
```

laevandus: Ideally it would be nice to see `.connecting` as the state just  after calling `connect()`. As a SDK user, who does not know internals, that would make sense. That said, it might make sense to do this separately (because we have this distinction at the moment where `.initialized` is used during the token fetch phase).

Especially because documentation says (one typo there):
```
    /// The initial state meaning that  there was no atempt to connect yet.
    case initialized
```
That is confusing for the SDK user. 

My view is that I feel like it is OK in the PR to go for `.initialized`, because this is what how the status has been reported previously, but then in a follow-up PR actually change this to `.connecting` because that would make more sense for SDK users (thinking about splitting because no idea what side-effects it could have and that should be tested throughly).

martinmitrevski: Yes, exactly, I had the same thinking as Toomas. Basically, this part:
"The problem here is that we only set connectionState = .connecting after the token phase finishes.", doesn't feel correct. We are connecting (we are trying to get a token), but that's not reflected anywhere. To reduce risks, I'm fine to merge it like this as well, but yeah, we should make this more explicit.

Sources/StreamChat/WebSocketClient/WebSocketClient.swift
martinmitrevski: yeah, now after seeing the diagram and the explanations, it makes more sense, thanks.

martinmitrevski (approved):
Ok, let's merge it with the current state. But first, @testableapple should show this PR some love.

Sources/StreamChat/Repositories/ConnectionRepository.swift
nuno-vieira: Yes, changing this would be very risky at the moment, that is why I decided not to do, and I remember I had a couple of issues trying it.

Sources/StreamChat/ChatClient.swift
```diff
@@ -458,7 +472,7 @@ public class ChatClient {
             completion()
         }
         authenticationRepository.clearTokenProvider()
-        authenticationRepository.cancelTimers()
+        authenticationRepository.reset()
```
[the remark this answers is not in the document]
nuno-vieira: It does not really matter, because this is triggered by the `reconnectionTimeoutHandler` which has `repeats: false`, so it will be already stopped by the time it reaches here

laevandus:
✅ 
Tested the branch and develop. In develop it did not disconnect after 15 seconds, but in this PR it correctly disconnected and the channel list updated.

laevandus (approved):

Sources/StreamChat/Repositories/ConnectionRepository.swift
nuno-vieira: @laevandus I changed that comment since it is internal either way. I tried changing the state to `.connecting` but this breaks other stuff, and so for now I prefer to not risk it. At most, adding another internal state, like `.fetchingToken `or something like this, would be better and not cause any impact on the rest of the logic. But for now, I think this is more than enough 👍

laevandus: Agree that we should try to tackle in another issue & PR.

PR #3518, opened 2024-12-02
Pinning channels for the current user

laevandus:
### 🔗 Issue Links

Part of [IOS-553](https://linear.app/stream/issue/IOS-553)

### 🎯 Goal

Add a new feature enabling users to pin channels (pinned status is tied to the user, not a global action)

### 📝 Summary

* Public API changes:
  * Add `Chat.pin(scope:)` and `Chat.unpin(scope:)`
  * Add `ChatChannelController.pin(scope:completion:)` and `ChatChannelController.unpin(scope:completion:)`
  * Add `ChannelPinningScope` with only one supported option of `me` (pinning the channel for everyone is not supported)
  * Add `FilterKey.pinned` for filtering channel lists
  * Add `ChannelListSortingKey.pinnedAt`
  * Add `ChatChannel.membership.pinnedAt`
  * Add `ChatChannel.isPinned`
* Parse `pinned_at` in member payloads
* Add supporting code for calling member's partial update with `pinned: true` using the new `MemberUpdatePayload` type or unsetting `pinned`
* Updated the demo app and replaced the custom pinning support. Additionally added channel list filter for pinned channels. 

### 🛠 Implementation

Pinning and unpinning works by calling partial update on the chat channel member (current user) and setting either `pinned: true` or unsetting this. Therefore, one could also use `ChatChannelMemberController.partialUpdate(extraData:unsetProperties:completion:)` where the user if is the current user id and extra data is `["pinned": .boolean(true)]` or unsetProperties is set to `["pinned"]`. New high-level pin and unpinning methods end up calling the same member updater's code, just make it more natural and discoverable.

Note: Pinning is per user. Meaning that if user A pins a channel with user B, then user B does not get the pinned state.

#### Pin and unpin channel

1. Open channel list
2. Pin a channel (debug menu > Pin channel)
Result: Pinned channel gets a yellow background (demo app has a custom message cell)
3. Unpin the channel (debug menu > Unpin channel)
Result: yellow background is removed

#### Querying pinned channels
1. Open channel list
2. Pin multiple channels (debug menu > Pin channel)
3. Open "Filter channels" menu in the toolbar > Pinned channels
Result: Only pinned channels are displayed 
4. Unpin a channel
Result: channel list removes the unpinned channel

#### Sorting
Change the demo app: DemoAppCoordinator+DemoApp.swift:L98 to
```
channelListQuery = .init(
                filter: .containMembers(userIds: [userCredentials.id]),
                sort: [.init(key: .pinnedAt), .init(key: .default)]
            )
```
1. Launch the app
2. Pinning a channel moves it to the top of the list
3. Unpin a channel and the list is reordered again

### ☑️ Contributor Checklist


DemoApp/StreamRuntimeCheck+StreamInternal.swift
```diff
@@ -15,4 +15,10 @@ extension StreamRuntimeCheck {
         guard let intValue = Int(value) else { return nil }
         return LogLevel(rawValue: intValue)
     }
+    
+    static var subsystems: LogSubsystem? {
+        guard let value = ProcessInfo.processInfo.environment["STREAM_LOG_SUBSYSTEM"] else { return nil }
+        guard let intValue = Int(value) else { return nil }
+        return LogSubsystem(rawValue: intValue)
+    }
```
laevandus, line 23: For making it easy to only get API requests and responses to the console.
E.g. this gives me only curl styled logs for every request we make with `APIClient`
- STREAM_LOG_LEVEL = 0
- STREAM_LOG_SUBSYSTEM = 4

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
```diff
@@ -331,9 +331,11 @@ extension NSManagedObjectContext {
         }
 
         // Note: membership payload should be saved before all the members
+        var membershipData: (dto: MemberDTO, payload: MemberPayload)?
```
laevandus, line 334: I will remove it when backend change has been deployed, currently it just acts as a workround

Sources/StreamChat/Extensions/URLRequest+cURL.swift
```diff
@@ -32,7 +32,9 @@ extension URLRequest {
                 .replacingOccurrences(of: "\"", with: "\\\"")
             cURL.append("-d \"\(escapedBody)\"")
         }
-        cURL.append("\"\(url.absoluteString)\"")
+        let urlString = url.absoluteString
+            .replacingOccurrences(of: "$", with: "%24") // encoded JSON payload
+        cURL.append("\"\(urlString)\"")
```
laevandus, line 37: curl description did not work for channel list queries where we have JSON for the payload parameter in the URL. This is only for debugging and does not affect the actual request.

Sources/StreamChat/Query/ChannelListQuery.swift
```diff
@@ -112,6 +112,23 @@ public extension FilterKey where Scope: AnyChannelListFilterScope {
     /// A filter key for matching the `blocked` value.
     /// Supported operators: `equal`
     static var blocked: FilterKey<Scope, Bool> { .init(rawValue: "blocked", keyPathString: #keyPath(ChannelDTO.isBlocked)) }
+    
+    /// A filter key for matching the `pinned` value.
```
laevandus, line 116: More complicated because filtered value needs to be a boolean, but the locally stored value is Date

Sources/StreamChat/Workers/ChannelMemberUpdater.swift
```diff
@@ -38,6 +38,33 @@ class ChannelMemberUpdater: Worker {
...
+    func pinMemberChannel(
+        _ isPinned: Bool,
+        userId: UserId,
+        cid: ChannelId,
+        completion: @escaping (Error?) -> Void
+    ) {
+        partialUpdate(
+            userId: userId,
+            in: cid,
+            extraData: isPinned ? ["pinned": .bool(true)] : nil,
```
nuno-vieira, line 51: The overall implementation LGTM, only this part I have some doubts. The "pinned" should be a native field on our side right? So it should not be extra data?

laevandus: I'll make it more explicit.

laevandus: I reviewed other partial methods we have. I think the naming `extraData` is a bit unfortunate here. It is actually `set`.

```
static func partialMemberUpdate(
        userId: UserId,
        cid: ChannelId,
        extraData: [String: RawJSON]?,
        unset: [String]?
    ) -> Endpoint<PartialMemberUpdateResponse> {
        var body: [String: AnyEncodable] = [:]
        if let extraData {
            body["set"] = AnyEncodable(extraData)
        }
        if let unset {
            body["unset"] = AnyEncodable(unset)
        }
```
It actually does not mean all the data is going to be part of member's extra data. All in all, it is correct.

Channel updater has

```
func partialChannelUpdate(
        updates: ChannelEditDetailPayload, // <--- to `set`
        unsetProperties: [String], <-- `to unset`
    ) {
```

@nuno-vieira I am not sure if I should change anything here at the moment. Ideally we should fix this naming in the next major. Meanwhile, I could add a new partial update method with better naming, but does it really make sense at the moment.

martinmitrevski: The damage for this one is already done long time ago. It should be `set`, not `extraData`, but there's no point in changing it only here. Imo, we should check if we have it in the tech debt / v5 docs.

martinmitrevski (approved):
Looks good to me ✅ Can you also do this in the SwiftUI app? Most of the stuff is already in place.

Sources/StreamChat/Workers/ChannelMemberUpdater.swift
nuno-vieira: @martinmitrevski This is a new method, and it was actually introduced by me. Initially, the goal was just to update extra data, which is it is named like that. The backend structure, also make is kinda hard to map this properly. But yeah, the `CHannelEditPayload` is what we want in this case actually. So I would prefer we do something like that. So `ChannelMemberUpdatePayload`

CHANGELOG.md
```diff
@@ -3,7 +3,15 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
-### 🔄 Changed
+## StreamChat
+### ✅ Added
+- Pinning channels [#3518](https://github.com/GetStream/stream-chat-swift/pull/3518)
+  - Add `Chat.pin()` and `Chat.unpin()`
+  - Add `ChatChannelController.pin(completion:)` and `ChatChannelController.unpin(completion:)`
+  - Add `FilterKey.pinned` for filtering channel lists
+  - Add `ChannelListSortingKey.pinnedAt`
+  - Add `ChatChannel.membership.pinnedAt`
+  - Add `ChatChannel.isPinned`
```
nuno-vieira, line 14: I think we can improve the API, since this action only makes sense for the current user:

```suggestion
- Pinning channels [#3518](https://github.com/GetStream/stream-chat-swift/pull/3518)
  - Add `ConnectedUser.pin(channelId:)` and `ConnectedUser.unpin(channelId:)`
  - Add `CurrentChatUserController.pin(channelId:)` and `CurrentChatUserController.unpin(channelId:)`
  - Add `FilterKey.pinned` for filtering channel lists
  - Add `ChannelListSortingKey.pinnedAt`
  - Add `ChatChannel.membership.pinnedAt`
  - Add `ChatChannel.isPinned`
```

nuno-vieira (commented):
After some internal discussions, it looks to me we need to:
1. Change the Updater to have a `ChannelMemberUpdatePayload` so that we can support native properties and extra data
2. Move this new functions to the Current User / Connected User (State layer probably is fine, but controllers makes sense to move to CurrentUserController)

DemoApp/StreamChat/Components/DemoChatChannelListItemView.swift
```diff
@@ -11,7 +11,7 @@ final class DemoChatChannelListItemView: ChatChannelListItemView {
         if content?.searchResult?.message != nil {
             return super.contentBackgroundColor
         }
-        if AppConfig.shared.demoAppConfig.isChannelPinningEnabled && content?.channel.isPinned == true {
+        if content?.channel.isPinned == true {
```
nuno-vieira, line 14: Out of scope (but not intended 😄 ) but in the future if we have global pinning, we might need to have more than 1 property to distinguish `isPinned` from local only or global only 🤔

Sources/StreamChat/Models/ChannelPinningScope.swift
```diff
@@ -0,0 +1,11 @@
...
+// Copyright © 2024 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+/// The scope of the channel pinning action.
+public enum ChannelPinningScope: String {
+    /// Channel is pinned only for the currently connected user. Other channel members do not see channels as pinned.
+    case me
+}
```
nuno-vieira, line 11: Since this not meant to be used by the customer to perform switching it should be fine. But in theory, adding a new case would be a breaking change. But I guess it is fine in this case 👍

nuno-vieira (approved):
LGTM! ✅

testableapple (approved):

PR #3531, opened 2024-12-13
Add Static Location and Live Location Support

nuno-vieira:
### 🔗 Issue Links
Resolves https://linear.app/stream/issue/IOS-578/location-attachment

### 🎯 Goal

Adds support for static and live location attachments in the Low-Level Client SDK.

The UI has been implemented in the Demo App to demonstrate how to use the new location APIs.

#### New APIs:
- `ChatChannelController`
   - `sendStaticLocation()` - Sends a static location message to the channel.
   - `startLiveLocationSharing()` - Starts a live location-sharing message in the channel.
 - `ChatMessageController`
   - `partialUpdateMessage()` - Updates the message partially. (It was missing from the SDK)
   - `stopLiveLocationSharing()` - Stops sharing the live location attachment if it has one.
- `CurrentChatUserController`
   - `updateLiveLocation()` - Updates the location of **all** active live location messages for the current user. 
   - `loadActiveLiveLocationMessages()` - Loads all active locations of the current user. Should be called only once.
- `CurrentChatUserControllerDelegate`
  - `didStartSharingLiveLocation()` - Notifies whenever the current user is sharing any live location.
  - `didStopSharingLiveLocation()` - Notifies whenever the current user stopped/expired all live locations.
  - `didChangeActiveLiveLocationMessages()` - Notifies whenever the current user's live location messages change.
  - `didFailToUpdateLiveLocation()` - Called whenever a live location failed to update. Mostly should be used for debugging.
- `Throttler`
   - The Throttler was part of the UI SDK somehow, so it was moved to the LLC and made public like the Debouncer.
- `ChatMessage`
   - `sharedLocation` - Returns the location if it has one, either live or static.

### 🛠 Implementation

The SDK at the moment only handles updating the location attachments. The location tracking should be provided by the App. Something like the [`LocationProvider`](https://github.com/GetStream/stream-chat-swift/blob/7957eab0d582331cb28a98757ea8c255dea2e444/DemoApp/LocationProvider.swift) in the Demo App should be implemented by the customer.

#### Creating a location attachment

In order to create a new message with a location attachment, the developer can use the `ChannelController.sendStaticLocation()` or the `ChannelController.startSharingLiveLocation()`.

#### Sending location updates (Live Location)

The customer is responsible for sending new location updates to the SDK. This is done through the `CurrentChatUserController.updateLiveLocation()` method. This method will update all the current user's active location attachments. Internally, it uses a `activeLiveLocationMessagesObserver` that keeps track of the active location attachments of the current user. These changes are also available to the customer through the `CurrentChatUserControllerDelegate` to make it easier for the developer to know when it should track location updates and when it can turn them off.

#### Stopping live location attachment

- `ChatMessageController.stopLiveLocationSharing()`: Stops a live location attachment in the given message if it has an active location attachment.

#### Overall Data Flow

```mermaid
sequenceDiagram
    participant App
    participant LocationProvider
    participant SDK
    participant Backend

    Note over App,Backend: Start New Location Share
    App->>LocationProvider: getCurrentLocation()
    LocationProvider-->>App: Return current location
    App->>SDK: startLiveLocationSharing(location)
    SDK->>Backend: Create location message
    Backend-->>SDK: Confirm creation
    SDK->>App: onStartLiveLocationSharing()
    App->>LocationProvider: startMonitoring()

    Note over App,Backend: Location Updates
    LocationProvider->>LocationProvider: Monitor location changes
    LocationProvider->>SDK: didUpdateLocation(location)
    SDK->>Backend: updateLiveLocation
    Note over SDK: 3s Throttling

    Note over App,Backend: Stop Location Share
    App->>SDK: stopLiveLocationSharing()
    SDK->>Backend: Stop location sharing
    Backend-->>SDK: Confirm stop
    SDK->>App: onStopLiveLocationSharing()
    App->>LocationProvider: stopMonitoring()
```

### 🎨 Showcase

| Static | Live |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/8f257501-8d58-49b3-8672-9a0877165053"/> | <video src="https://github.com/user-attachments/assets/ed59c4d6-6945-43ef-bc12-718617175d5d"/> | 

### 🧪 Manual Testing Notes

**Note:** To simulate live location updates in the Simulator, with the Simulator selected, Go to `Features > Location > City Bicycle Ride` in the top bar of the Mac.

**Precondition:** Select the Frankfurt C2 environemnt to test the location feature.

<details>
<summary><strong>Send a static location message ✅</strong></summary>

1. Open a channel with location sharing enabled  
2. Tap on the Attachments Icon  
3. Tap "Send Current Location"  
4. It should render an image snapshot of a map with a Pin  
5. Tapping the attachment should open the map full screen with the Pin  

</details>

<details>
<summary><strong>Send a live location message ✅</strong></summary>

1. Open a channel with location sharing enabled  
2. Tap on the Attachments Icon  
3. Tap "Share Live Location"  
4. Set an end time  
5. It should render an image of the map with the avatar of the user in it and a button saying “Stop Sharing”  
   - The map is just a static image (Does not update live)  
6. Tapping the attachment should open the map with the user's avatar, and the user's avatar should move when the location updates.  
7. Tapping the "Stop Sharing" button should stop sharing the live location.  
8. It should update the UI, and a label with “Live location ended” appears.  
9. In the full-screen map view, the avatar should not move anymore.  

</details>

<details>
<summary><strong>Receiving a live location message ✅</strong></summary>

1. Share a live location attachment from another user  
2. The receiving user should see the location message with “Live until XXX” date.  
3. Once the location has reached the end, it should update the message with “Live location ended”  
4. The same behaviour should happen if the user is in the full-screen map view.  
5. In the full-screen map view, the avatar should not move anymore.  

</details>

<details>
<summary><strong>Send a live location message when one from the current user is already active ✅</strong></summary>

1. Open a channel with location sharing enabled  
2. Tap on the Attachments Icon  
3. Tap "Share Live Location"  
4. Set an end time  
5. It should render an image of the map with the avatar of the user in it and a button saying “Stop Sharing”  
6. Send another live location
7. The new live location message should be sent, and the previous one should stop.

</details>

<details>
<summary><strong>Active live location is ended when the last update is after expiration ✅</strong></summary>

1. Share a live location attachment with a 1-minute expiration  
2. The user keeps updating the location  
3. When the last update has already passed the expiration, it should update to “Live location ended”  
4. The same should happen in the full-screen map view.  

</details>

<details>
<summary><strong>Active live location is ended after the expiration is reached ✅ </strong></summary>

1. Share a live location attachment with a 1-minute expiration  
2. The user is still, so no location updates are being sent to the server  
3. When the 1-minute is reached, the UI should update to “Live Location Ended” 
4. The same should happen in the full-screen map view. 
5. The location monitoring should also be stopped 

</details>

<details>
<summary><strong>Active live location from another user ends after the expiration is reached ✅</strong></summary>

1. Share a live location attachment with a 1-minute expiration with User A  
2. User B receive the live location in the message list.  
3. User A is still, so no location updates are being sent to the server  
4. When the 1-minute is reached, the UI from User B should update to “Live Location Ended” ❌  
5. The same should happen in the full-screen map view. ❌  

</details>

### ☑️ Contributor Checklist

## Summary by CodeRabbit

- **New Features**
  - Introduced comprehensive location sharing support, including static and live location sharing in chat messages.
  - Added the ability to send, update, and stop live location sharing within chat channels.
  - Enhanced message and channel models to display and manage shared locations.
  - New UI components for viewing and controlling shared locations, including live status indicators, map previews, user annotations, and a control banner.
  - Added partial message update functionality for selective content and attachment updates.
  - Added delegate callbacks and controller methods for live location lifecycle events.
  - Added a location provider to manage device location services and permissions.
  - Added background worker to track active live location end times and update states accordingly.

- **Improvements**
  - Updated privacy settings and permissions to support location access and background location updates.
  - Expanded channel capabilities to indicate and control location sharing permissions.
  - Enhanced message actions to disable editing for location messages.
  - Improved UI handling for location attachments and live location sharing controls.
  - Added throttling to live location updates to optimize network usage.
  - Refined snapshot caching and UI updates for location previews.
  - Reorganized and extended app configuration and plist for location sharing.

- **Bug Fixes**
  - Improved handling and filtering of live location messages to ensure accurate state and error reporting.

- **Tests**
  - Added extensive test coverage for location sharing, partial message updates, database persistence, API interactions, and error handling.
  - Added tests for live location lifecycle, end time tracking, and message update scenarios.

- **Chores**
  - Updated documentation, changelogs, configuration files, and test data to reflect new location sharing features and permissions.


DemoApp/StreamChat/Components/CustomAttachments/LocationAttachment/LocationDetailViewController.swift
```diff
@@ -27,20 +39,125 @@ class LocationDetailViewController: UIViewController {
...
+        )
+    }
+
+    func updateUserLocation(
+        _ coordinate: CLLocationCoordinate2D,
+        userImage: UIImage?,
+        userName: String
+    ) {
+        if let existingAnnotation = userAnnotation {
+            UIView.animate(withDuration: 5) {
```
martinmitrevski, line 87: why do we need to do 5 second long animation?

DemoApp/StreamChat/StreamChatWrapper+DemoApp.swift
```diff
@@ -23,7 +19,18 @@ extension StreamChatWrapper {
         if client == nil {
             client = ChatClient(config: config)
         }
-        client?.registerAttachment(LocationAttachmentPayload.self)
+
+        // Custom Attachments
+        if AppConfig.shared.demoAppConfig.isLocationAttachmentsEnabled {
```
martinmitrevski, line 24: I think it should be always on, since it's the demo app. Our config screen becomes very complex at this point.

martinmitrevski (commented):
Couldn't test a lot, since a had constant crashes (shared details on Slack). Let's figure that out first, then I will do another round.

DemoApp/Screens/DemoAppTabBarController.swift
```diff
@@ -69,4 +80,39 @@ class DemoAppTabBarController: UITabBarController, CurrentChatUserControllerDele
...
+    func currentUserController(
+        _ controller: CurrentChatUserController,
+        didChangeActiveLiveLocationMessages messages: [ChatMessage]
+    ) {
+        guard !messages.isEmpty else {
+            return
+        }
+
+        messages.forEach {
+            LocationAttachmentSnapshotView.snapshotsCache.removeObject(forKey: $0.id as NSString)
```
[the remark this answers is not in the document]
nuno-vieira: This will completely change, I'm changing the implementation to how iMessage works. Not a good time to review this now. I'll let you know once everything is ready 👍

DemoApp/StreamChat/Components/CustomAttachments/LocationAttachment/LocationDetailViewController.swift
nuno-vieira: This is mostly to have a smooth animation. It simulates the pin constantly being moved since the animation is always running. It looks quite good this way. But it is also possible to have an hard animation like WhatsApp, but for now this is quite smooth

DemoApp/StreamChat/StreamChatWrapper+DemoApp.swift
nuno-vieira: It was disabled by default before, but since we now support locations out of the box, I can enable it by default

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
```diff
@@ -825,6 +832,181 @@ public class ChatChannelController: DataController, DelegateCallable, DataStoreP
...
+    /// it will fail with an error.
+    ///
+    /// In order to
+    ///
+    /// - Parameters:
+    ///  - location: The location information.
+    ///  - text: The text of the message.
+    ///  - extraData: Additional extra data of the message object.
+    ///  - completion: Called when saving the message to the local DB finishes,
+    ///  not when the message reaches the server.
```
laevandus, line 905: Seems like a formatting issue

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
```diff
@@ -825,6 +832,181 @@ public class ChatChannelController: DataController, DelegateCallable, DataStoreP
...
+    ///  - extraData:  Additional extra data of the message object.
+    ///  - completion: Called when saving the message to the local DB finishes, not when the message reaches the server.
+    public func sendStaticLocation(
+        _ location: LocationAttachmentInfo,
+        text: String? = nil,
+        messageId: MessageId? = nil,
+        quotedMessageId: MessageId? = nil,
+        extraData: [String: RawJSON] = [:],
+        completion: ((Result<MessageId, Error>) -> Void)? = nil
+    ) {
```
laevandus, line 851: Planning to add similar functions to `Chat`?

Sources/StreamChat/Controllers/MessageController/MessageController.swift
```diff
@@ -272,6 +285,84 @@ public class ChatMessageController: DataController, DelegateCallable, DataStoreP
...
+    /// Updates the message partially and submits the changes directly to the server.
+    ///
+    /// **Note:** The `message.localState` is not changed in this method call.
+    ///
+    /// - Parameters:
+    ///   - text: The text in case the message
+    ///   - attachments: The attachments to be updated.
+    ///   - extraData: The additional data to be updated.
+    ///   - completion: Called when the server updates the message.
+    public func updateMessage(
```
laevandus, line 297: `partialUpdateMessage` for matching to the common naming pattern?

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
```diff
@@ -1790,6 +1979,17 @@ extension ClientError {
             "You can't specify a value outside the range 1-120 for cooldown duration."
         }
     }
+
+    final class ActiveLiveLocationAlreadyExists: ClientError {
+        let messageId: MessageId
+
+        public init(messageId: MessageId) {
```
laevandus, line 1986: Internal?

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
```diff
@@ -1766,7 +1955,7 @@ private extension ChatChannelController {
 
 // MARK: - Errors
 
-extension ClientError {
+public extension ClientError {
```
laevandus, line 1958: Keeping it internal?

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
laevandus: Should we move most of the logic in these action methods to separate handler/sender for making them reusable for `Chat`?
`TypingEventsSender`, `ReadStateHandler`?

Sources/StreamChat/Controllers/MessageController/MessageController.swift
nuno-vieira: The reason I did not put this name is because editMesasge is kinda already an partial update as well. All the existing data will remain the same. At least the end result will be the same, even though internally we provide the non-changed data to the server. 

The only real diff between the editMessage, and updateMessage, is that one is scheduled, and the other one is not 🤔

laevandus: Asking because we need to fill in public docs for https://getstream.io/chat/docs/ios-swift/send_message/#partial-update and might be useful for it to look similar. Currently Kotlin snippet uses `updateMessage` for full update, might get confusing. 

We should also include a way to `unset` properties like partial updates do.

nuno-vieira: AH nice catch in that case it probably makes sense to change then 👍

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
nuno-vieira: This ones should actually be public, it can be useful for customers to understand which errors they can react to

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
nuno-vieira: Yes this one should be internal 👍

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
nuno-vieira: I'm thinking about creating a seperate ticket for this maybe. The PR is already way to big, so I think it is better we do it seperatly 🤔

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
nuno-vieira: Fixed!

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
laevandus: OK with me! It sure is long already

Sources/StreamChat/Controllers/MessageController/MessageController.swift
nuno-vieira: Done ✅

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
nuno-vieira: I will create the ticket then 👍

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
```diff
@@ -825,6 +832,179 @@ public class ChatChannelController: DataController, DelegateCallable, DataStoreP
...
+                completion?(.failure(error ?? ClientError.Unknown()))
+            }
+            return
+        }
+
+        client.messageRepository.getActiveLiveLocationMessages(for: cid) { result in
+            switch result {
+            case let .success(messages):
+                guard let message = messages.first,
+                      let liveLocation = message.liveLocationAttachments.first
```
laevandus, line 978: Could you please clarify why only the first message in the array is used for stopping. Because I can have only one active message with live sharing?

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
```diff
@@ -825,6 +832,179 @@ public class ChatChannelController: DataController, DelegateCallable, DataStoreP
...
+        guard let cid = cid, isChannelAlreadyCreated else {
+            channelModificationFailed { error in
+                self.callback {
+                    completion?(.failure(error ?? ClientError.Unknown()))
+                }
+            }
+            return
+        }
+
+        client.messageRepository.getActiveLiveLocationMessages(for: cid) { [weak self] result in
```
laevandus, line 919: Side comment: when public functions are added to the state layer then probably most of this gets be moved to channel updater for easier re-usage (message repo call and the later create new message).

Sources/StreamChat/Controllers/CurrentUserController/CurrentUserController.swift
```diff
@@ -230,6 +266,25 @@ public extension CurrentChatUserController {
...
+    /// The updates are throttled to avoid sending too many requests.
+    ///
+    /// - Parameter location: The new location to be updated.
+    func updateLiveLocation(_ location: LocationAttachmentInfo) {
+        guard let messages = activeLiveLocationMessagesObserver?.items, !messages.isEmpty else {
+            return
+        }
+
+        locationUpdatesThrottler.execute { [weak self] in
+            for message in messages {
```
laevandus, line 280: Since this is throttled, should we read the array of messages from the observer again to avoid any timing issues.

Sources/StreamChat/Controllers/CurrentUserController/CurrentUserController.swift
```diff
@@ -378,23 +448,79 @@ private extension CurrentChatUserController {
...
+            return nil
+        }
+        return environment.currentUserActiveLiveLocationMessagesObserverBuilder(
+            client.databaseContainer,
+            MessageDTO.activeLiveLocationMessagesFetchRequest(
+                currentUserId: currentUserId,
+                channelId: nil
+            ),
+            { try $0.asModel() },
+            NSFetchedResultsController<MessageDTO>.self
```
laevandus, line 463: We are not reusing models if only 1 message out of 5 changes. Probably fine since there are not many live location messages anyway.

Sources/StreamChat/Repositories/MessageRepository.swift
```diff
@@ -281,7 +281,32 @@ class MessageRepository {
...
         }
     }
-    
+
+    func getActiveLiveLocationMessages(
+        for channelId: ChannelId,
+        completion: @escaping (Result<[ChatMessage], Error>) -> Void
+    ) {
+        let context = database.backgroundReadOnlyContext
+        context.perform {
```
laevandus, line 290: Alternative: there is also the `database.read` method

laevandus: but I guess you used this one because I now saw that later NSManagedObjectContext is needed, but read gives the session one. Session would require new load method to be defined.

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
nuno-vieira: Yes, at the moment, only one live attachment per message is supported

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
nuno-vieira: Yeah, I'm planning on creating an addittional object to reuse some of the logic when I jump on the state layer. This will also make the CurrentUserController testable, which at the moment it is not

Sources/StreamChat/Controllers/CurrentUserController/CurrentUserController.swift
nuno-vieira: Did not understood, can you clarify? 🤔

laevandus: I was thinking about the case where this function is called, timer starts, observer changes so its items changes, timer fires, timer currently uses captured messages (not the ones observer has have changes). Probably not a big deal anyway.

laevandus (approved):

Sources/StreamChat/Controllers/CurrentUserController/CurrentUserController.swift
nuno-vieira: I see; in this case, it is not a big deal because the timer will execute immediately, or it won't execute at all if it is not within the 3s interval. But ofc depends on the frequence of the updates. But I would say it is a not big deal for now at least 👍 Maybe when doing the state layer I'll do this change 👍

DemoApp/StreamChat/Components/CustomAttachments/LocationAttachment/LocationAttachmentSnapshotView.swift
```diff
@@ -3,21 +3,45 @@
...
         }
     }
 
-    var snapshotter: MKMapSnapshotter?
-
     var didTapOnLocation: (() -> Void)?
+    var didTapOnStopSharingLocation: (() -> Void)?
+
+    let mapOptions: MKMapSnapshotter.Options = .init()
+    let mapHeight: CGFloat = 150
```
martinmitrevski, line 41: We should do probably do better here. With fixed height, it doesn't look good on larger screens, e.g. iPads. Maybe we do fixed ratio instead?
Simulator Screenshot - iPad Pro 13-inch (M4) - 2025-01-08 at 13 59 10

Sources/StreamChat/Controllers/MessageController/MessageController.swift
```diff
@@ -183,8 +184,14 @@ public class ChatMessageController: DataController, DelegateCallable, DataStoreP
 
     /// The worker used to fetch the remote data and communicate with servers.
     private let messageUpdater: MessageUpdater
+
+    /// The polls repository to fetch polls data.
     private let pollsRepository: PollsRepository
+
+    /// The replies pagination hdler.
```
martinmitrevski, line 191: *handler

DemoApp/StreamChat/StreamChatWrapper+DemoApp.swift
martinmitrevski: It would be great if it's on, we want more customers to see it. Sensitive data is not an issue, since our test users are hardcoded and you can't identify a person. About spamming, I can see that as a potential issue, but we can disable it later if we see it happening.

Sources/StreamChat/Controllers/CurrentUserController/CurrentUserController.swift
```diff
@@ -37,6 +37,9 @@ public class CurrentChatUserController: DataController, DelegateCallable, DataSt
         return _basePublishers as? BasePublishers ?? .init(controller: self)
     }
 
+    /// The observer for the active live location messages.
```
martinmitrevski, line 40: what about the state layer objects? Shouldn't we also expose the location feature?

DemoApp/StreamChat/Components/CustomAttachments/DemoComposerVC.swift
```diff
@@ -2,55 +2,86 @@
...
 
         return actions
     }
 
-    func sendLocation() {
-        guard let location = dummyLocations.randomElement() else { return }
-        let locationAttachmentPayload = LocationAttachmentPayload(
-            coordinate: .init(latitude: location.latitude, longitude: location.longitude)
-        )
+    func sendInstantStaticLocation() {
```
martinmitrevski, line 39: off topic for this PR: but the Composer doesn't look / feel good with the addition of polls and location buttons. We should try to find time to make it in app eventually.

DemoApp/StreamChat/Components/CustomAttachments/LocationAttachment/LocationAttachmentSnapshotView.swift
```diff
@@ -3,21 +3,45 @@
 //
 
 import MapKit
+import StreamChat
 import StreamChatUI
 import UIKit
 
-class LocationAttachmentSnapshotView: _View {
-    static var snapshotsCache: NSCache<NSString, UIImage> = .init()
+class LocationAttachmentSnapshotView: _View, ThemeProvider {
```
martinmitrevski, line 10: Do you think it's a lot of effort if we also do this in the SwiftUI SDK?

DemoApp/StreamChat/Components/CustomAttachments/LocationAttachment/LocationAttachmentSnapshotView.swift
```diff
@@ -71,79 +145,151 @@ class LocationAttachmentSnapshotView: _View {
...
             pinImage?.draw(at: point)
         }
-        return image
     }
-}
 
-private extension LocationCoordinate {
-    var cachingKey: NSString {
-        NSString(string: "\(latitude),\(longitude)")
+    private func updateAnnotationView() {
```
martinmitrevski, line 249: The snapshot view re-loads when I do a long press. Can we improve that part?

martinmitrevski (commented):
Looks good! Left few comments, let me know what you think.

DemoApp/StreamChat/StreamChatWrapper+DemoApp.swift
nuno-vieira: The main problem I see is that if multiple people are using the same user and that user is sharing the location, the map will be buggy since it is reporting updates in multiple locations. If we are testing the feature, it could be annoying. But we can also use a user that is not used often

DemoApp/StreamChat/Components/CustomAttachments/LocationAttachment/LocationAttachmentSnapshotView.swift
```diff
@@ -36,7 +60,36 @@ class LocationAttachmentSnapshotView: _View {
...
+        button.addTarget(self, action: #selector(handleStopButtonTap), for: .touchUpInside)
+        return button
+    }()
+
+    lazy var avatarView: ChatUserAvatarView = {
+        let view = ChatUserAvatarView()
+        view.translatesAutoresizingMaskIntoConstraints = false
+        view.shouldShowOnlineIndicator = false
+        view.layer.masksToBounds = true
+        view.layer.cornerRadius = 15
```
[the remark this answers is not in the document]
nuno-vieira: It is 15, because it is half of the size of the view (30). Probably should have extracted the 30 to a property and divided by 2, but since it is demo app code, I did not care much 😅 But I can change this to make it clear

DemoApp/StreamChat/Components/CustomAttachments/LocationAttachment/LocationAttachmentSnapshotView.swift
nuno-vieira: The problem here is actually not the location attachment but the SDK itself. We have a weird logic on the attachment container that makes the attachments on the iPad look way too big. iMessage on iPad, just renders the attachments with the same size. Example:

image

And it is the same for locations.

So the actual fix here should be to improve how we render attachments on iPad. Or I can add a fixed width for the map 🤔 And then it will work the same for all devices

nuno-vieira: Either way, this is Demo App code, so the customer can also improve this and change the size of the map as they desire

Sources/StreamChat/Controllers/CurrentUserController/CurrentUserController.swift
nuno-vieira: Yeah @laevandus also mentioned this. it is better we create a separate PR, since this is one is already way to big. I will create a ticket for this

DemoApp/StreamChat/Components/CustomAttachments/LocationAttachment/LocationAttachmentSnapshotView.swift
nuno-vieira: Most of the code is for the Demo App, so it might take some time. Although most of the stuff can be reused from UIKit I believe 🤔

DemoApp/StreamChat/Components/CustomAttachments/LocationAttachment/LocationAttachmentSnapshotView.swift
nuno-vieira: Let me check 👁️

DemoApp/StreamChat/Components/CustomAttachments/LocationAttachment/LocationAttachmentSnapshotView.swift
nuno-vieira: Fixed this ✅ But yeah, the customer is still limited by the `maxContentWidthMultiplier`. Which is not ideal.

DemoApp/StreamChat/Components/CustomAttachments/LocationAttachment/LocationAttachmentSnapshotView.swift
nuno-vieira: Fixed ✅

Sources/StreamChat/Controllers/MessageController/MessageController.swift
nuno-vieira: Done

DemoApp/StreamChat/Components/CustomAttachments/LocationAttachment/LocationAttachmentSnapshotView.swift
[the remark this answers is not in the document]
nuno-vieira: Extract this to a property 👍

DemoApp/StreamChat/StreamChatWrapper+DemoApp.swift
nuno-vieira: Updated ✅

DemoApp/StreamChat/Components/CustomAttachments/LocationAttachment/LocationAttachmentSnapshotView.swift
martinmitrevski: much better 👍

martinmitrevski (approved):
LGTM! ✅

DemoApp/StreamChat/Components/CustomAttachments/LocationAttachment/LocationAttachmentSnapshotView.swift
```diff
@@ -3,21 +3,53 @@
 //
 
 import MapKit
+import StreamChat
 import StreamChatUI
 import UIKit
 
-class LocationAttachmentSnapshotView: _View {
-    static var snapshotsCache: NSCache<NSString, UIImage> = .init()
+class LocationAttachmentSnapshotView: _View, ThemeProvider {
```
martinmitrevski, line 10: I think at some point we should consider putting this in the UI SDK - lot of boilerplate here

Sources/StreamChat/Workers/Background/ActiveLiveLocationsEndTimeTracker.swift
```diff
@@ -0,0 +1,92 @@
...
+    private func scheduleInactiveLocation(for messageId: String, at endAt: Date) {
+        // Cancel any existing work item for the same messageId
+        cancelWorkItem(for: messageId)
+
+        let workItem = DispatchWorkItem { [weak self] in
+            self?.setInactiveLocation(for: messageId)
+        }
+        workItems[messageId] = workItem
+
+        let endAtTime = endAt.timeIntervalSinceNow
```
martinmitrevski, line 72: should we also add handling for past dates? (e.g. the app was killed before this runs, and we never stop the location)

Sources/StreamChat/Workers/Background/ActiveLiveLocationsEndTimeTracker.swift
```diff
@@ -0,0 +1,92 @@
...
+
+import Foundation
+
+/// An observer type that observes all active live locations in the database.
+typealias ActiveLiveLocationsObserver = StateLayerDatabaseObserver<ListResult, MessageDTO, MessageDTO>
+
+/// A worker that is responsible for tracking when the end time of active locations is reached.
+class ActiveLiveLocationsEndTimeTracker: Worker {
+    private let activeLiveLocationsObserver: ActiveLiveLocationsObserver
+    private var workItems: [String: DispatchWorkItem] = [:]
```
martinmitrevski: should we make this thread safe?

martinmitrevski (commented):
Looks great ✅  Why is this still only on F2? Also, would be interesting to see the guide - I'm thinking it might be too complex to implement everything on customer side.

Sources/StreamChat/Workers/Background/ActiveLiveLocationsEndTimeTracker.swift
nuno-vieira: Yeah I guess I can use the existing queue, to make sure it is thread-safe 👍

Sources/StreamChat/Workers/Background/ActiveLiveLocationsEndTimeTracker.swift
nuno-vieira: That is handled by our `currentUserControllerDidStopSharingLiveLocation()` delegate event. When the user connects, and there are no more active locations, the customer will be notified to stop monitoring locations

DemoApp/StreamChat/Components/CustomAttachments/LocationAttachment/LocationAttachmentSnapshotView.swift
nuno-vieira: Yes, we can 👍 I think it makes sense to add some of the helpers. Although, Ideally we would implement the whole UI, without directly depending on any of the Apple's location SDK. Since some customers might use a different SDK.

martinmitrevski: Yes, I think you will notice this when writing the guide. If it becomes too complex, it might make sense to do it.

Sources/StreamChat/Workers/Background/ActiveLiveLocationsEndTimeTracker.swift
nuno-vieira: Actually, this not needed, because all code will be run from the CoreData's writableContext, so it is already thread-safe

PR #3534, opened 2024-12-17
Improve database model conversion performance by caching extra data and keeping row cache

laevandus:
### 🔗 Issue Links

Resolves: [IOS-610](https://linear.app/stream/issue/IOS-610)

### 🎯 Goal

- Improve performance of model conversion with more extreme use-cases (e.g. extra data is used a lot)

### 📝 Summary

- Cache extra data because it can be large and makes JSONDecoder way too busy when DB changes occur
- Remove reading the model state from the persistent storage in DatabaseContainer.write because it degrades performance a lot (other managed object contexts need to refetch everything, affects row cache)

#### Performance measurement setup

Demo user `Chewbacca` has 19 muted channels and has a channel named "Performance Testing" with 12 members where each of the member has extra data filled with 162 fields (nested data).
```swift
var myData = [String: RawJSON]()
for index in 0..<40 {
    myData["extra_data_key_\(index)"] = .dictionary([
        "timestamp": .number(Date().timeIntervalSinceReferenceDate),
        "value": .string("value_\((0..<100).randomElement()!)"),
        "number": .number(Double(index)),
        "flag": .bool(true)
    ])
}
let updates = MemberUpdatePayload(extraData: [
    "extra_data_testing": .dictionary(myData),
    "extra_data_testing_2": .string("just testing stuff")
])
```

I added a simple snippet to the demo app (locally) which sent 20 messages with 1 second delay. While the message sending was taking place, I used time profiler to measure how busy threads are.

Time spent for extra data decoding went down a lot as expected, but time profiler was also showing that Core Data is busy fetching data for DTOs when `asModel()` functions were called (for ChannelDTO.asModel it was around 40% of time). What triggered this behaviour is the snippet in the `DatabaseContainer.write` where updated objects are fully refetched from the persistent store. This seems to affect the Core Data's row cache and makes it to discard it which in turn means that `backgroundReadOnlyContext` needs to refetch everything (because each context has its own DTO objects). The current state is that all the writes go through the `writableContext`, therefore there is not any need to actually sync update objects in the write call because there is no-one else changing the persistent store.

### 🎨 Showcase

Baseline
Cache - only extra data caching
Cache & merge off - extra data caching and not force updating updated objects
Cache & merge off & data prefetching - extra data caching, not force updating updated objects, and relationship prefetching enabled (`StreamRuntimeCheck._isDatabasePrefetchingEnabled`)

| Measurement | Baseline | Cache | Cache & merge off | Cache & merge off & data prefetching |
| ------ | ------ | ----- | ----- | ----- |
|  Total time (all threads)  |  6.45   |  5.39  |  4.46  |  4.35  |
|  `ChannelDTO.asModel()` (from channel controller)  |  1.97   |  1.44  |  0.80  |  0.82  |
|  `ChannelDTO.asModel()` (total)  |  4.07   |  2.96  |  2.34  |  1.75  |
|  `CurrentUserDTO.asModel()` (total)  |  1.82   |  1.49  |  1.46  |  0.84  |
|  Extra data decode (total)  |  1.33   |  0.12  |  0.13  |  0.13  |

Caching extra data also reduces the memory footprint for RawJSON

| Baseline | Caching |
| ------ | ------ |
| NoCache | Cache |

#### Takeaways
- Extra data caching is needed when extra data is used a lot (here we had an extreme case)
- `refresh:mergeChanges:` affects a lot (**~45%** improvement in channel controller)
- In general, this use-case saw **~33%** of improvement in CPU usage
-`CurrentChatUser` should not reference full `ChatChannel` models (cids would be better)
- Channel updates trigger current user model changes because updating a UserDTO referenced by CurrentUserDTO leads to CurrentUserDTO to trigger an update  

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/Database/DTOs/MemberModelDTO.swift
+++ b/Sources/StreamChat/Database/DTOs/MemberModelDTO.swift
@@ -189,7 +189,7 @@ extension ChatChannelMember {
         
         let extraData: [String: RawJSON]
         do {
-            extraData = try JSONDecoder.default.decode([String: RawJSON].self, from: dto.user.extraData)
+            extraData = try JSONDecoder.stream.decodeCachedRawJSON(from: dto.user.extraData)
         } catch {
             log.error(
                 "Failed to decode extra data for user with id: <\(dto.user.id)>, using default value instead. "
@@ -198,13 +198,15 @@ extension ChatChannelMember {
             extraData = [:]
         }
 
-        var memberExtraData: [String: RawJSON] = [:]
-        if let dtoMemberExtraData = dto.extraData {
-            do {
-                memberExtraData = try JSONDecoder.default.decode([String: RawJSON].self, from: dtoMemberExtraData)
-            } catch {
-                memberExtraData = [:]
-            }
+        let memberExtraData: [String: RawJSON]
+        do {
+            memberExtraData = try JSONDecoder.stream.decodeCachedRawJSON(from: dto.extraData)
+        } catch {
+            log.error(
+                "Failed to decode extra data for channel member with id: <\(dto.user.id)>, using default value instead. "
+                    + "Error: \(error)"
+            )
+            memberExtraData = [:]
         }
 
         let role = dto.channelRoleRaw.flatMap { MemberRole(rawValue: $0) } ?? .member
--- a/Sources/StreamChat/Database/DatabaseContainer.swift
+++ b/Sources/StreamChat/Database/DatabaseContainer.swift
@@ -222,11 +222,6 @@ class DatabaseContainer: NSPersistentContainer, @unchecked Sendable {
                 try actions(self.writableContext)
                 FetchCache.clear()
 
-                // Refresh the state by merging persistent state and local state for avoiding optimistic locking failure
-                for object in self.writableContext.updatedObjects {
-                    self.writableContext.refresh(object, mergeChanges: true)
-                }
-
                 if self.writableContext.hasChanges {
                     log.debug("Context has changes. Saving.", subsystems: .database)
                     try self.writableContext.save()
--- a/Sources/StreamChat/Utils/Codable+Extensions.swift
+++ b/Sources/StreamChat/Utils/Codable+Extensions.swift
@@ -9,6 +9,7 @@ import Foundation
 final class StreamJSONDecoder: JSONDecoder, @unchecked Sendable {
     let iso8601formatter: ISO8601DateFormatter
     let dateCache: NSCache<NSString, NSDate>
+    let rawJSONCache: RawJSONCache
 
     override convenience init() {
         let iso8601formatter = ISO8601DateFormatter()
@@ -17,12 +18,21 @@ final class StreamJSONDecoder: JSONDecoder, @unchecked Sendable {
         let dateCache = NSCache<NSString, NSDate>()
         dateCache.countLimit = 5000 // We cache at most 5000 dates, which gives good enough performance
 
-        self.init(dateFormatter: iso8601formatter, dateCache: dateCache)
+        self.init(
+            dateFormatter: iso8601formatter,
+            dateCache: dateCache,
+            rawJSONCache: RawJSONCache(countLimit: 500)
+        )
     }
 
-    init(dateFormatter: ISO8601DateFormatter, dateCache: NSCache<NSString, NSDate>) {
+    init(
+        dateFormatter: ISO8601DateFormatter,
+        dateCache: NSCache<NSString, NSDate>,
+        rawJSONCache: RawJSONCache
+    ) {
         iso8601formatter = dateFormatter
         self.dateCache = dateCache
+        self.rawJSONCache = rawJSONCache
 
         super.init()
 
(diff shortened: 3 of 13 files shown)


Sources/StreamChat/Database/DatabaseContainer.swift
```diff
@@ -222,11 +222,6 @@ class DatabaseContainer: NSPersistentContainer, @unchecked Sendable {
                 try actions(self.writableContext)
                 FetchCache.clear()
 
-                // Refresh the state by merging persistent state and local state for avoiding optimistic locking failure
-                for object in self.writableContext.updatedObjects {
-                    self.writableContext.refresh(object, mergeChanges: true)
-                }
-
```
nuno-vieira, line 229: Can you explain why was this needed before, and now it is not? 🤔 What are the chances of this producing some regressions? 🤔

Sources/StreamChat/Utils/Codable+Extensions.swift
```diff
@@ -17,12 +18,21 @@ final class StreamJSONDecoder: JSONDecoder, @unchecked Sendable {
         let dateCache = NSCache<NSString, NSDate>()
         dateCache.countLimit = 5000 // We cache at most 5000 dates, which gives good enough performance
 
-        self.init(dateFormatter: iso8601formatter, dateCache: dateCache)
+        self.init(
+            dateFormatter: iso8601formatter,
+            dateCache: dateCache,
+            rawJSONCache: RawJSONCache(countLimit: 500)
```
nuno-vieira, line 24: In the performance benchmark table, we are not checking the memory usage. Is there any significant increase in memory by caching all the extra data?

Sources/StreamChat/Database/DTOs/MemberModelDTO.swift
```diff
@@ -198,13 +198,15 @@ extension ChatChannelMember {
...
         }
 
-        var memberExtraData: [String: RawJSON] = [:]
-        if let dtoMemberExtraData = dto.extraData {
-            do {
-                memberExtraData = try JSONDecoder.default.decode([String: RawJSON].self, from: dtoMemberExtraData)
-            } catch {
-                memberExtraData = [:]
-            }
+        let memberExtraData: [String: RawJSON]
```
nuno-vieira, line 201: Another interesting thing that we could try to improve is that `ChatMember` and `ChatUser` are both classes. And I think we should be able to convert them to structs without API changes. Could be interesting to see if there are any performance improvements by doing so 🤔

nuno-vieira (commented):
Added a couple of comments for now 👍

Sources/StreamChat/Database/DatabaseContainer.swift
nuno-vieira: Isn't there a chance that Data will be missing or the FRC won't trigger updates?

Sources/StreamChat/Database/DTOs/MemberModelDTO.swift
martinmitrevski: why do you think structs would improve performance? Probably only the SDK size would increase.

Sources/StreamChat/Database/DatabaseContainer.swift
martinmitrevski: Yes, we need more details on why was this removed.

Sources/StreamChat/Utils/Codable+Extensions.swift
martinmitrevski: I don't expect it, but maybe we clean it up on memory warning notifications, just in case?

martinmitrevski (commented):
Looks good! I'm only concerned about the refresh removal, let's test this a bit.

Sources/StreamChat/Database/DatabaseContainer.swift
laevandus: I looked into the history of this and it goes like this:

1. In [2021](https://github.com/GetStream/stream-chat-swift/commit/f22b62e3c73b2b88a652dcb0b404dcb5fa454730) it started as an optimisation change for reducing context changes
```swift
// If you touch ManagedObject and update one of it properties to same value
// Object will be marked as `updated` even it hasn't changed.
// By reseting such objects we remove updates that are not updates.
for object in self.writableContext.updatedObjects {
    if object.changedValues().isEmpty {
        self.writableContext.refresh(object, mergeChanges: false)
    }
}
```
refresh(_:mergeChanges:false) = "Updates the persistent properties of a managed object to use the latest values from the persistent store. If flag is false, the context discards pending changes and the managed object becomes a fault. Upon next access, the context reloads the object’s values from the persistent store or last cached state."

2. In early [2023](https://github.com/GetStream/stream-chat-swift/commit/14773b53decfa46f5682856811a63903786c754d) it gets changed to 
`self.writableContext.refresh(object, mergeChanges: true) `
"If flag is true, the context reloads the object’s property values from the store or the cache. Then the context applies local changes over the newly loaded values. Merging the local values into object always succeeds, and never results in a merge conflict."
I think this was done because change 1 was making FRC to lose track of changes. 
Reasoning:
We use DTOs' willSave to force FRC updates by accessing the "parent" DTO and assigning a property and setting back the same value to the DTO. For example, UserDTO is getting saved and if it happens to be CurrentUserDTO.user, then UserDTO.willSave forces the CurrentUserDTO to change as well. Then FRC observing the CurrentUserDTO picks up the change. Discarding changes in these cases breaks FRC observing which I think the change number 2 tackled. The change in number 2 made sure these updates go through.
From UserDTO:
```swift
if let currentUser = currentUser, !currentUser.hasChanges {
    let fakeNewUnread = currentUser.unreadChannelsCount
    currentUser.unreadChannelsCount = fakeNewUnread
}
```

3. I removed the if check when testing if it fixes these some crashes in CoreData. It didn't.
 
Since the change number 2 some things have happened like we do not use viewContext anymore (except  reading the state in DataStore, DB observers do not use it anymore). All the DB mutations go through the `writableContext`. Therefore, there should never be a case where `writableContext` itself would require to refresh the DTO from the persistent store (because it is the only one changing it). That is the main reason I believe we can remove it.
Interesting side of it is that refreshing an object seems to discard cached data in Core Data (cached data in other contexts) which leads to performance degradation.

Sources/StreamChat/Utils/Codable+Extensions.swift
laevandus: Internally it is NSCache which clears itself on low memory warning. I used a wrapping class because of unit-tests. Can't mock NSCache from the Swift side (it is objc generic class and brings restrictions).

Sources/StreamChat/Database/DTOs/MemberModelDTO.swift
laevandus: ChatUser.init itself is visible in the time profiler, but it is like 1 ms which is nothing compared to other things happening. At some point it would be nice to only use structs for models (consistent).

nuno-vieira: Usually  structs a lightweight to create and destroy, so it should improve things, but maybe not significantly. @laevandus We not only need to check the init, but also how much time is spend destroying existing objects 🤔

Sources/StreamChat/Database/DatabaseContainer.swift
martinmitrevski: Didn't we also introduce a state layer context? Will this have an impact on it, have you checked that?

Sources/StreamChat/Utils/Codable+Extensions.swift
laevandus: There is significant decrease in memory usage used for RawJSON.
Before:
NoCache
After
Cache

Sources/StreamChat/Database/DatabaseContainer.swift
laevandus: For state layer, with similar test, the change is 1.60 s to 0.77 s

Sources/StreamChat/Database/DTOs/MemberModelDTO.swift
laevandus: Based on time profiler I don't see this popping up really, a couple of ms for `ChatUser.__deallocating_deinit` and `ChatChannelMember.__deallocating_deinit`. We should do that for consistency, but in a separate ticket. No real performance improvement here.

nuno-vieira: Okay nice 👍

nuno-vieira (approved):
LGTM! ✅ ⚡

martinmitrevski (approved):
LGTM ✅

PR #3536, opened 2024-12-18
Add FilterKey.members and support for equal operator to match an array

laevandus:
### 🔗 Issue Links

Resolves [IOS-609](https://linear.app/stream/issue/IOS-609)

### 🎯 Goal

Enable filtering channel list with `$eq` and an array of channel member ids

### 📝 Summary

- Add `FilterKey.members`
- Add support for `$eq` to match an array of values

### 🛠 Implementation

Filtering by `members` supports `$in` and `$eq` operators. We did not expose the latter. For supporting it, some changes were required for enabling Core Data to match channels with `$eq` and an array of values.

### 🧪 Manual Testing Notes

1. Log in with `chewbacca`
2. Try different channel queries (R2-D2 Channels uses the new members equals query and returns all the channels where members equal to `current user` and `r2-d2`)

Example snippet
```swift
DispatchQueue.main.asyncAfter(deadline: .now() + .seconds(5)) {
    Task {
        let query = ChannelListQuery(filter: .equal(.members, values: ["chewbacca", "luke_skywalker"]))
        let channelList = self.makeChannelList(with: query)
        try await channelList.get()
        let channels = channelList.state.channels
        for channel in channels {
            print(channel.cid, channel.memberCount, channel.lastActiveMembers.map(\.id).sorted())
        }
    }
}
```

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/Query/ChannelListQuery.swift
+++ b/Sources/StreamChat/Query/ChannelListQuery.swift
@@ -46,12 +46,6 @@ extension Filter where Scope: AnyChannelListFilterScope {
     }
 }
 
-// We don't want to expose `members` publicly because it can't be used with any other operator
-// than `$in`. We expose it publicly via the `containMembers` filter helper.
-extension FilterKey where Scope: AnyChannelListFilterScope {
-    static var members: FilterKey<Scope, UserId> { .init(rawValue: "members", keyPathString: #keyPath(ChannelDTO.members.user.id)) }
-}
-
 /// Filter values to be used with `.invite` FilterKey.
 public enum InviteFilterValue: String, FilterValue {
     case pending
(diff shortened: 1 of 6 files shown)


martinmitrevski (approved):
LGTM ✅  Also, please update the docs in the LLC

testableapple (approved):

Sources/StreamChat/Query/ChannelListQuery.swift
```diff
@@ -46,12 +46,6 @@ extension Filter where Scope: AnyChannelListFilterScope {
     }
 }
 
-// We don't want to expose `members` publicly because it can't be used with any other operator
-// than `$in`. We expose it publicly via the `containMembers` filter helper.
```
nuno-vieira, line 50: Shouldn't we expose something like `equalMembers(userIds:)` since only $eq and $in operator are supported?

laevandus: It feels more closer to public docs if we allow using it directly. But I am lacking some historic knowledge on channel list filters so not sure what is best here.

laevandus: With other keys we can create incompatible combinations.

nuno-vieira: Yeah, my guess is that for some properties, we started limiting the operators, especially for performance reasons. But for other properties we did not follow the same approach.

IMO, it would be better that programmatically you could only use the operators that you are allowed to, and that it would not require you to look into docs to see which operators are supported. It would be better to just limit every field but now it is kinda too late, since we already expose most of them :/ 

So yeah, maybe it is better we open up this one for consistency

laevandus: Since we have that containMembers, maybe it is better then to have equalMembers. Hmm

nuno-vieira (approved):

PR #3541, opened 2024-12-23
Read data using a background context in the DataStore

laevandus:
### 🔗 Issue Links

Resolves [IOS-626](https://linear.app/stream/issue/IOS-626)

### 🎯 Goal

`DataStore` is public and therefore we should not have the main thread requirement which can be misused easily

### 📝 Summary

* Force all thread data reads to the background context's queue

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/Database/DatabaseContainer.swift
+++ b/Sources/StreamChat/Database/DatabaseContainer.swift
@@ -282,6 +282,24 @@ class DatabaseContainer: NSPersistentContainer, @unchecked Sendable {
             }
         }
     }
+    
+    func readAndWait<T>(_ actions: (DatabaseSession) throws -> T) throws -> T {
+        let context = viewContext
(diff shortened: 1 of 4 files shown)


Sources/StreamChat/Database/DatabaseContainer.swift
```diff
@@ -282,6 +282,24 @@ class DatabaseContainer: NSPersistentContainer, @unchecked Sendable {
...
+            } catch {
+                readError = error
+            }
+        }
+        if let result {
+            return result
+        } else {
+            throw readError ?? ClientError.Unknown()
+        }
+    }
```
laevandus, line 302: Throwing performAndWait is iOS 15 or later, therefore this is not so pretty at the moment.

martinmitrevski (approved):

PR #3548, opened 2025-01-03
Add file property to the ImageAttachmentPayload

laevandus:
### 🔗 Issue Links

Resolves [IOS-319](https://linear.app/stream/issue/IOS-319/)

### 🎯 Goal

Fix an issue where `ImageAttachmentPayload` does not set `file_size` and/or `mime_type`

### 📝 Summary

* Add `ImageAttachmentPayload.file` and new init method requiring setting `file`
* Deprecate `init(title:imageRemoteURL:originalWidth:originalHeight:extraData:)` in favour of `init(title:imageRemoteURL:file:originalWidth:originalHeight:extraData:)`
* Aligns `ImageAttachmentPayload` to other attachment payloads which require `file` to be set

### 🛠 Implementation

Other attachment payloads require `AttachmentFile` type when creating a payload. `ImageAttachmentPayload` is the only one which represents a file, but does not require it. This leads to all the image attachments sent from the iOS SDK to not set `file_size` and/or `mime_type`.

### ☑️ Contributor Checklist


Sources/StreamChat/Models/Attachments/ChatMessageImageAttachment.swift
```diff
@@ -24,6 +24,8 @@ public struct ImageAttachmentPayload: AttachmentPayload {
     public var originalWidth: Double?
     /// The original height of the image in pixels.
     public var originalHeight: Double?
+    /// The image file size information.
+    public var file: AttachmentFile
```
laevandus, line 28: It is not optional like in other payload types. This means I need to set some sort of values in the deprecated init methods.

martinmitrevski: how did it work so far if it's not optional? I think we should have it optional instead of passing default values.

nuno-vieira: Yeah, something is off here 🤔

nuno-vieira: Even though the property might not be optional from the backend, it is better to keep it optional to not cause breaking changes or needing to change the initializers

nuno-vieira: Besides that, there might be a chance that the backend is not able to retreive information, so we are not 100% sure it won't be nil at any point

Sources/StreamChat/Models/Attachments/ChatMessageImageAttachment.swift
```diff
@@ -35,10 +37,38 @@ public struct ImageAttachmentPayload: AttachmentPayload {
...
+    ///   - title: A title, usually the name of the image.
+    ///   - imageRemoteURL: A link to the image.
+    ///   - file: The image file size information.
+    ///   - originalWidth: The original width of the image in pixels.
+    ///   - originalHeight: The original height of the image in pixels.
+    ///   - extraData: Custom data associated with the attachment.
+    public init(
+        title: String?,
+        imageRemoteURL: URL,
+        file: AttachmentFile,
```
nuno-vieira, line 55: This needs to be optional since a customer can use a custom CDN and they might not have the file info
```suggestion
        file: AttachmentFile?,
```

Sources/StreamChat/Models/Attachments/ChatMessageImageAttachment.swift
laevandus: Only downside is that then it is not going to be similar to other payloads:
[VideoAttachmentPayload](https://github.com/GetStream/stream-chat-swift/blob/be21a9e2654d12f2c84d68a55c60f3985d757525/Sources/StreamChat/Models/Attachments/ChatMessageVideoAttachment.swift#L26), 
[VoiceRecordingAttachmentPayload](https://github.com/GetStream/stream-chat-swift/blob/be21a9e2654d12f2c84d68a55c60f3985d757525/Sources/StreamChat/Models/Attachments/ChatMessageVoiceRecordingAttachment.swift#L25), [FileAttachmentPayload](https://github.com/GetStream/stream-chat-swift/blob/be21a9e2654d12f2c84d68a55c60f3985d757525/Sources/StreamChat/Models/Attachments/ChatMessageFileAttachment.swift#L24), and [AudioAttachmentPayload](https://github.com/GetStream/stream-chat-swift/blob/be21a9e2654d12f2c84d68a55c60f3985d757525/Sources/StreamChat/Models/Attachments/ChatMessageAudioAttachment.swift#L24) have it as non-optional.

nuno-vieira: Ah... did not know it was non-optional in other payloads 🤔

nuno-vieira (approved):
LGTM! ✅

Sources/StreamChat/Models/Attachments/ChatMessageImageAttachment.swift
nuno-vieira: Either way, I would still provide optional in this case. This deprecation might be too alarming for customers, and they might not know what to do. So I would keep it optional, but I don't have a strong opinion on this one 👍

Sources/StreamChat/Models/Attachments/ChatMessageImageAttachment.swift
```diff
@@ -78,6 +110,8 @@ public struct ImageAttachmentPayload: AttachmentPayload {
         imageURL = imageRemoteURL
         self.originalWidth = originalWidth
         self.originalHeight = originalHeight
+        let fileType = AttachmentFileType(ext: imageRemoteURL.pathExtension)
+        file = AttachmentFile(type: fileType, size: 0, mimeType: nil)
```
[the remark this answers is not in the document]
laevandus: This is the existing deprecated init, therefore I just filled in with empty data.

Sources/StreamChat/Models/Attachments/ChatMessageImageAttachment.swift
laevandus: The whole `AttachmentFile` handling is a little bit odd. Other payload types require it (init takes a non-optional value), therefore I followed the same flow.

PR #3550, opened 2025-01-03
Do not use view context in StreamChat

laevandus:
### 🎯 Goal

Do not use `viewContext` in LLC

### 🛠 Implementation

We should use background context for reading and `writableContext` for writing. There is not need for `viewContext` anymore. Remove viewContext references from LLC (except DataContainer). Many of our tests still access `viewContext`.

### 🧪 Manual Testing Notes

N/A

### ☑️ Contributor Checklist


Sources/StreamChat/Database/DatabaseContainer.swift
```diff
@@ -284,7 +284,7 @@ class DatabaseContainer: NSPersistentContainer, @unchecked Sendable {
     }
     
     func readAndWait<T>(_ actions: (DatabaseSession) throws -> T) throws -> T {
-        let context = viewContext
+        let context = backgroundReadOnlyContext
```
laevandus, line 287: This was a mistake in [PR-3541](https://github.com/GetStream/stream-chat-swift/pull/3541)

Sources/StreamChat/StateLayer/ChatClient+Factory.swift
```diff
@@ -10,8 +10,11 @@ extension ChatClient {
...
     ///
     /// - Throws: An error if no user is currently logged-in.
-    @MainActor public func makeConnectedUser() throws -> ConnectedUser {
-        let user = try CurrentUserDTO.load(context: databaseContainer.viewContext)
+    public func makeConnectedUser() throws -> ConnectedUser {
+        let user = try databaseContainer.readAndWait { session in
+            guard let dto = session.currentUser else { throw ClientError.CurrentUserDoesNotExist() }
+            return try dto.asModel()
+        }
         return ConnectedUser(user: user, client: self)
```
laevandus, line 18: All our other make methods do not require `@MainActor`, therefore we should skip it here as well. Downside is that this change would generate a warning that no await is needed if it was called outside of MainActor. Can we make this change?

martinmitrevski: yes, that's fine

martinmitrevski (approved):

PR #3561, opened 2025-01-15
Fix thread reply action shown when inside a Thread

nuno-vieira:
### 🔗 Issue Links

Resolves IOS-187

### 🎯 Goal

Fix thread reply action shown when inside a Thread.

### 🛠 Implementation

We need context in the Message Actions where the view is being presented. This was previously requested by other customers since this can be useful for other things. So now, whenever we present the message actions, we have a new boolean property that tells if the actions are being presented inside a thread or not.

### 🧪 Manual Testing Notes

1. Open a Thread
2. Long press the Parent Message
3. Should not show "Thread Reply Action"

### ☑️ Contributor Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 # Upcoming
 
-### 🔄 Changed
+## StreamChatUI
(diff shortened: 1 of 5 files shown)


laevandus (approved):
Looks good ✅

martinmitrevski (approved):

PR #3564, opened 2025-01-16
Add support for transforming Messages, Channels and Members

nuno-vieira:
### 🔗 Issue Links

Resolves https://linear.app/stream/issue/IOS-638

### 🎯 Goal

Add support for transforming Messages, Channels and Members. This will allow changing some of the data of these models at run time.

The most common use case for this feature is to easily allow E2EE encryption.

> [!NOTE]
> Our Members and User models are different, and since Member is a subclass of User (Tech Debt), transforming Users would not reflect on Members, so at the moment we only provide transforming members. If in the future Users is needed, it needs to be a new function to transform Users. 

### 📝 Summary

A new `StreamModelsTransformer` protocol is added and can be implemented by customers. Currently, only Channels, Messages and Channel Members are transformable, with limited data that can be changed.

Example usage:
```swift
class CustomStreamModelsTransformer: StreamModelsTransformer {
    func transform(channel: ChatChannel) -> ChatChannel {
        channel.replacing(
            name: "Hey!",
            imageURL: channel.imageURL,
            extraData: channel.extraData
        )
    }

    func transform(message: ChatMessage) -> ChatMessage {
        message.replacing(
            text: "Yo",
            extraData: message.extraData,
            attachments: message.allAttachments
        )
    }

    func transform(newMessageInfo: NewMessageTransformableInfo) -> NewMessageTransformableInfo {
        newMessageInfo.replacing(
            text: "Changed",
            attachments: newMessageInfo.attachments,
            extraData: newMessageInfo.extraData
        )
    }
    
    func transform(member: ChatChannelMember) -> ChatChannelMember {
        member.replacing(
            name: "\(member.name ?? member.id)",
            imageURL: member.imageURL,
            userExtraData: member.extraData,
            memberExtraData: member.memberExtraData
        )
    }
}
```

With the `replacing` function, if customers want to only change the data of a specific controller, they can do it like so:

```swift
let transformedChannels = controller.channels.map {
    $0.replacing(name: "YO", imageURL: $0.imageURL, extraData: $0.extraData)
}

// with combine:
controller.observableObject
      .channels
      .publisher
      .map {
         $0.replacing(name: "YO", imageURL: $0.imageURL, extraData: $0.extraData)
      }
      .sink(receiveValue: { transformedChannel in
         print(transformedChannel)
      })
```

### 🛠 Implementation

The way it works is that when `asModel()` transformation happens, we provide a way to change it before returning the final model.

### 🧪 Manual Testing Notes

N/A

### ☑️ Contributor Checklist


Sources/StreamChat/Config/StreamModelsTransformer.swift
```diff
@@ -0,0 +1,77 @@
...
+        text: String,
+        attachments: [AnyAttachmentPayload],
+        extraData: [String: RawJSON]
+    ) {
+        self.text = text
+        self.attachments = attachments
+        self.extraData = extraData
+    }
+
+    public func mapped(
```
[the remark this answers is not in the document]
laevandus: Since we use the term `transformer` then it could be just `transform` as well.

Random example from Apple APIs:
mutating func transform(_ transform: ([Column](doc://com.apple.documentation/documentation/tabulardata/column?language=swift)<WrappedElement>.[Element](doc://com.apple.documentation/documentation/tabulardata/column/element?language=swift)) throws -> [Column](doc://com.apple.documentation/documentation/tabulardata/column?language=swift)<WrappedElement>.[Element](doc://com.apple.documentation/documentation/tabulardata/column/element?language=swift)) rethrows

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
```diff
@@ -1638,7 +1647,11 @@ private extension ChatChannelController {
             let observer = BackgroundEntityDatabaseObserver(
                 database: self.client.databaseContainer,
                 fetchRequest: ChannelDTO.fetchRequest(for: cid),
-                itemCreator: { try $0.asModel() as ChatChannel }
+                itemCreator: {
+                    try $0.asModel(
+                        transformer: self.client.config.modelsTransformer
+                    ) as ChatChannel
```
laevandus, line 1653: When thinking ahead to cleaning up `deletedMessagesVisibility` and others then that would require to have an access to `ChatConfig` within `asModel()`. Therefore, should we just pass in the config here and read the transformer within the `asModel` implementation?

laevandus:
StateLayer needs also this.

Sources/StreamChat/Config/StreamModelsTransformer.swift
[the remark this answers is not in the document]
nuno-vieira: Yeah I initially thought about transform, but didn't want it to be repetetive 🤔 But I guess that is okay... Either here are some other alternatives:
- `with()`
- `replacing()`
- `updating()`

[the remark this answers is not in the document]
nuno-vieira: I actually like `replacing` it is very clear what it will do, it will indeed replace what ever you provide it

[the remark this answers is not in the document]
laevandus: True, while looking at these examples then it is clear what it does. 👍

nuno-vieira:
@laevandus I've add it to the state layer already 🤔

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
nuno-vieira: Yeah I'm not sure how that will look like, but we can improve it once we work on that 👍 But I don't think we need to pass the whole config to the asModel though 🤔

Sources/StreamChat/Config/StreamModelsTransformer.swift
```diff
@@ -0,0 +1,77 @@
...
+//
+
+import Foundation
+
+/// An object that provides a way to transform Stream Chat models.
+///
+/// Only some data can be changed. The method `replacing()` is used to create a new object from existing data.
+/// All transform functions have default implementation, so you can override only the ones you need.
+///
+/// **Note:** When using `replacing()` method, you need to provide all the properties of the existing object.
```
laevandus, line 12:
```suggestion
/// - Note: When using `replacing()` method, you need to provide all the properties of the existing object.
```
For nicer formatting in Xcode.

Sources/StreamChat/Config/StreamModelsTransformer.swift
```diff
@@ -0,0 +1,77 @@
...
+/// Or you can pass `nil` and that will erase the existing value.
+///
```
laevandus, line 14: We should probably add something like:
```
/// - Important: Transform methods can be called often and therefore, it must be performant.
```

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
laevandus: Sure, can be changed later if needed. When I briefly checked it then `ChannelDTO.asModel` requires multiple values from the config.

nuno-vieira: I see, in that case I will probably change this yet 👍

DemoApp/Shared/StreamChatWrapper.swift
```diff
@@ -192,3 +193,34 @@ extension StreamChatWrapper {
...
+
+// MARK: - Stream Models Transformer
+
+// An object to test the Stream Models transformer.
+// By default it is not used. To use it, set it to the `modelsTransformer` property of the `ChatClientConfig`.
+
+class CustomStreamModelsTransformer: StreamModelsTransformer {
+    func transform(channel: ChatChannel) -> ChatChannel {
+        channel.replacing(
+            name: "Hey!",
```
martinmitrevski, line 205: Nit: maybe just uppercase all the letter of the channel name? Would feel weird otherwise to see only "Hey!" everywhere.
Same for messages, some small transformation of the original text would be nice.

Sources/StreamChat/Models/Channel.swift
```diff
@@ -228,6 +228,46 @@ public struct ChatChannel {
         self.muteDetails = muteDetails
         self.previewMessage = previewMessage
     }
+
+    public func replacing(
```
martinmitrevski, line 232: we need to document this one (and the one below)

martinmitrevski (commented):
Looks great, I like this approach more! Few more finishing touches (docs, tests) and we're good to 🚢

DemoApp/Shared/StreamChatWrapper.swift
nuno-vieira: Done ✅

Sources/StreamChat/Models/Channel.swift
nuno-vieira: Done ✅

martinmitrevski (approved):

laevandus (approved):
✅

PR #3566, opened 2025-01-17
Make accessing `ChatClientConfig` in the Database Layer safer

nuno-vieira:
### 🔗 Issue Links

Resolves https://linear.app/stream/issue/IOS-639/

### 🎯 Goal

Make accessing the `ChatClientConfig`internally more thread safer.

### 🛠 Implementation

Initially, the idea was to pass `ChatClientConfig` around instead of accessing directly in the `NSManagedObjectContext`. The problem is that this would cause a massive amount of breaking changes. So the plans changed, and now the whole `ChatClientConfig` is stored in the `NSManagedObjectContext.userInfo`, but we make it thread-safe.

### 🧪 Manual Testing Notes

N/A

### ☑️ Contributor Checklist


Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
```diff
@@ -1638,7 +1638,7 @@ private extension ChatChannelController {
             let observer = BackgroundEntityDatabaseObserver(
                 database: self.client.databaseContainer,
                 fetchRequest: ChannelDTO.fetchRequest(for: cid),
-                itemCreator: { try $0.asModel() as ChatChannel }
+                itemCreator: { try $0.asModel(config: self.client.config) as ChatChannel }
```
laevandus, line 1641: We need to avoid capturing self in itemCreators

nuno-vieira: This is changed now, so no need 👍

Sources/StreamChat/Database/DatabaseSession.swift
```diff
@@ -4,7 +4,23 @@
...
 
-extension NSManagedObjectContext: DatabaseSession {}
+extension NSManagedObjectContext: DatabaseSession {
+    private static let chatClientConfigKey = "io.getStream.StreamChat.config.key"
+
+    var chatClientConfig: ChatClientConfig {
+        var config: ChatClientConfig?
+        performAndWait {
+            config = userInfo[Self.chatClientConfigKey] as? ChatClientConfig
+        }
```
laevandus, line 14: This does bring slight overhead (it is not huge, just tried sending 20 messages). Thinking if we want to do this since we fixed actual misuses already and this accessor is only used from `asModel()`. But that said, this makes it future-proof.
Screenshot 2025-01-21 at 09 55 03

nuno-vieira: Hum I see, I'll do a couple more perf tests. But if there is a significant impact, maybe it is better we remove the performAndWait

nuno-vieira: After running the benchmarking script on a real device, it looks like there is no significant impact: 

image

For now, lets keep it this way to bulletproof, but if we find any issues with this in the future, we can remove the `performAndWait` 👍

nuno-vieira: Let's re-test this with this branch: https://github.com/GetStream/stream-chat-swift/pull/3564

For now we merge it as it

PR #3509, opened 2024-11-27
[CI] Bump Xcode version

testableapple:
### 🔗 Issue Links

https://linear.app/stream/issue/IOS-12/

### 🧪Testing Notes

✅ [Cron checks](https://github.com/GetStream/stream-chat-swift/actions/runs/12072961939)

### ☑️ Contributor Checklist

- [x] I have signed the [Stream CLA](https://docs.google.com/forms/d/e/1FAIpQLScFKsKkAJI7mhCr7K9rEIOpqIDThrWxuvxnwUq2XkHyG154vQ/viewform) (required)
- [ ] This change should be manually QAed
- [ ] Changelog is updated with client-facing changes
- [ ] Changelog is updated with new localization keys
- [ ] New code is covered by unit tests
- [ ] Comparison screenshots added for visual changes
- [ ] Affected documentation updated (docusaurus, tutorial, CMS)

PR #3568, opened 2025-01-22
Fix reaction author's view with shrinked reaction images in iOS 18

nuno-vieira:
### 🔗 Issue Links

Resolves https://linear.app/stream/issue/IOS-652/[uikit]-reaction-icons-have-invalid-size-in-reactions-overlay

### 🎯 Goal

Fixes reaction author's view with shrinked reaction images introduced on iOS 18. 

We did not spot the bug when re-recording iOS 18 snapshots here: https://github.com/GetStream/stream-chat-swift/pull/3509

### 🛠 Implementation

The root of the issue is that on iOS 18, the `Button.image.size` is `nil` in the `ChatMessageReactionItemView.intrinsicContentSize` when it is inside the reaction author cell view. The best way I found to fix the issue without breaking changes is to provide a default size instead of `super.intrinsicContentSize` in case the image does not have any size.

### 🧪 Manual Testing Notes

It is not necessary to do manual tests since it is covered by snapshots.

### ☑️ Contributor Checklist


testableapple (approved):
LGTM 👍

laevandus (approved):

PR #3569, opened 2025-01-23
Dismiss thread when hard deleting root message

nuno-vieira:
### 🔗 Issue Links
Resolves https://linear.app/stream/issue/IOS-282/hard-delete-on-a-root-message-from-a-thread-is-broken

### 🎯 Goal

Fix the Thread not dismissing when the root message is hard deleted.

### 🛠 Implementation

I've decided to implement this on the Demo App since hard deleting is an action that only exists on the Demo App. Usually, we do not advertise the feature much since it is not recommended.

Besides that, the logic will most likely differ depending on the app. So this is mostly app logic and not SDK logic.

Either way, while doing this, I spotted some bugs:
- The `ChatThreadVC.didReceiveEvent` delegate was always being called twice. This was due to the fact that we had `channelEventsController` and `eventsController` assigning both delegates to the `ChatThreadVC`
- The `MessageDeletedEvent` was not being surfaced to the `didReceiveEvent` delegate. This only happened for the hard deleted message, and the reason is that this event requires a `message: ChatMessage` property. But, since the mesasge is hard deleted, it also does not exist in CoreData, so we can't provide the whole message, only the ID. For this reason, I created a new event `MessageHardDeletedEvent` and deprecated the `MessageDeletedEvent.isHardDeleted` property. This was the only option to avoid breaking changes.

**UPDATE:** In order to avoid creating a new `MessageHardDeletedEvent`, and diverge from the other SDKs, we decided to map the message payload directly to the model so that the `MessageDeletedEvent.message` is present.

#### Dismissing Thread
1. Create / Open a Thread
2. Hard delete the root message inside a thread
3. Should dismiss the view

#### Hard Deleting Thread from Channel
1. Create a Thread
2. Send a reply also in channel
3. Go back to the Channel
4. Hard delete the root thread
5. Should delete the thread and replies also in channel

### ☑️ Contributor Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,6 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
    - Add `ChatMessage.replacing()`
    - Add `ChatChannel.replacing()`
    - Add `ChatChannelMember.replacing()`
-- Fix duplicated `didReceiveEvent` inside `ChatThreadVC` [#3569](https://github.com/GetStream/stream-chat-swift/pull/3569)
 - Fix hard deleted message events not being reported in `EventsController` [#3569](https://github.com/GetStream/stream-chat-swift/pull/3569)
 ### 🔄 Changed
 - Deprecates `MessageDeletedEvent.isHardDeleted` [#3569](https://github.com/GetStream/stream-chat-swift/pull/3569)
@@ -33,6 +32,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### 🐞 Fixed
 - Fix thread reply action shown when inside a Thread [#3561](https://github.com/GetStream/stream-chat-swift/pull/3561)
 - Fix reaction author's view with shrinked reaction images in iOS 18 [#3568](https://github.com/GetStream/stream-chat-swift/pull/3568)
+- Fix duplicated `didReceiveEvent` inside `ChatThreadVC` [#3569](https://github.com/GetStream/stream-chat-swift/pull/3569)
 ### 🔄 Changed
 - Deprecates `ChatThreadVC.channelEventsController` [#3569](https://github.com/GetStream/stream-chat-swift/pull/3569)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## StreamChat
 ### ✅ Added
 - Add `MessageHardDeletedEvent` [#3569](https://github.com/GetStream/stream-chat-swift/pull/3569)
+- Expose `Event.name` to easily check which event it is [#3569](https://github.com/GetStream/stream-chat-swift/pull/3569)
 ### 🐞 Fixed
 - Calling async `connectUser()` methods can sometimes throw `CurrentUserDoesNotExist()` unexpectedly [#3565](https://github.com/GetStream/stream-chat-swift/pull/3565)
 - Fix creating controllers from background threads leading to rare crashes [#3566](https://github.com/GetStream/stream-chat-swift/pull/3566)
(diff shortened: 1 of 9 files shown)


DemoApp/StreamChat/Components/DemoChatThreadVC.swift
```diff
@@ -70,6 +70,30 @@ class DemoChatThreadVC: ChatThreadVC, CurrentChatUserControllerDelegate {
...
     }
 
+    // MARK: - Dismissing the thread if the root message was hard deleted.
+
+    override func messageController(
+        _ controller: ChatMessageController,
+        didChangeReplies changes: [ListChange<ChatMessage>]
+    ) {
+        // If the message is hard deleted, do not update the UI
+        // (Otherwise it would should the root message disappearing)
```
laevandus, line 80: Tiny typo
```suggestion
        // (Otherwise it would show the root message disappearing)
```

CHANGELOG.md
```diff
@@ -4,6 +4,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 # Upcoming
 
 ## StreamChat
+### ✅ Added
+- Add `MessageHardDeletedEvent` [#3569](https://github.com/GetStream/stream-chat-swift/pull/3569)
```
martinmitrevski, line 8: We should remove this now

CHANGELOG.md
```diff
@@ -12,6 +15,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
    - Add `ChatMessage.replacing()`
    - Add `ChatChannel.replacing()`
    - Add `ChatChannelMember.replacing()`
+- Fix hard deleted message events not being reported in `EventsController` [#3569](https://github.com/GetStream/stream-chat-swift/pull/3569)
+### 🔄 Changed
+- Deprecates `MessageDeletedEvent.isHardDeleted` [#3569](https://github.com/GetStream/stream-chat-swift/pull/3569)
```
martinmitrevski, line 20: Also this as well

Sources/StreamChat/WebSocketClient/Events/MessageEvents.swift
```diff
@@ -241,3 +246,75 @@ public struct NewMessageErrorEvent: Event {
     public let messageId: MessageId
     public let error: Error
 }
+
+// MARK: - Workaround to map a deleted message to Model.
+
+// At the moment our SDK does not support mapping Payload -> Model
+// So this is just a workaround for `MessageDeletedEvent` to have the `message` non-optional.
+// So some of the data will be incorrect, but for this is use case is more than enough.
```
martinmitrevski, line 254: which data would be incorrect? Isn't this everything coming from the payload?

Sources/StreamChatUI/ChatThread/ChatThreadVC.swift
```diff
@@ -24,6 +24,7 @@ open class ChatThreadVC: _ViewController,
     public var initialReplyId: MessageId?
 
     /// Controller for observing typing events for this thread.
+    @available(*, deprecated, message: "Events are now handled by the `eventsController`.")
```
martinmitrevski, line 27: why do we deprecate it in this PR?

Sources/StreamChat/WebSocketClient/Events/MessageEvents.swift
nuno-vieira: No, we are not mapping reactions, the `isSentByCurrentUser` is also a computed property on ou DTOs, so that is also incorrect. `isFlaggedByCurrentUser` as well... etc.. But those are minors

Sources/StreamChatUI/ChatThread/ChatThreadVC.swift
nuno-vieira: > The ChatThreadVC.didReceiveEvent delegate was always being called twice. This was due to the fact that we had channelEventsController and eventsController assigning both delegates to the ChatThreadVC

Its in the PR description. Since it is not used anymore, it is deprecated. The delegate of this controller was conflicting with the `eventsController`

laevandus (approved):
LGTM ✅

martinmitrevski (approved):

PR #3578, opened 2025-01-30
[CI] Lock sonar scanner and mint versions and get rid of brew

testableapple:
### 🥷 Problem

- Sonar released a new version that broke our pipeline LLC testing. While investigating I figured out that we do not lock the `sonar-scanner` and `mint` versions (even though we did, they were not locked on CI)
- Test results were not published to Sonar on merges to `develop` and `main`

### 📝 Summary

- Lock `sonar-scanner` and `mint` versions
- Get rid of `brew`
    - `sonar-scanner` is now installed via downloading an image directly
    - `mint` is now installed via `mint` itself
- Move `LLC test` action to a separate workflow so the results can be published to SonarCloud on merges to `develop` and `main` as well as on PRs


nuno-vieira (approved):
LGTM! ✅

.github/workflows/smoke-checks-llc.yml
```diff
@@ -0,0 +1,66 @@
...
+        job_name: "Test LLC (Debug)"
+        fields: message,commit,author,action,workflow,job,took
+      env:
+        SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
+        MATRIX_CONTEXT: ${{ toJson(matrix) }}
+      if: ${{ github.event_name == 'push' && failure() }}
+    - name: Parse xcresult
+      if: failure()
+      run: |
+        brew install chargepoint/xcparse/xcparse
```
laevandus, line 58: This one is still OK to install with homebrew, right? Asking because there was a comment above about homebrew.

testableapple: Yeah, this is used only when tests fail, which should be a rare case. But can't install it another way.

PR #3588, opened 2025-02-07
Add Draft Messages Support

nuno-vieira:
### 🔗 Issue Links

https://linear.app/stream/issue/IOS-640/draft-messages-support

### 🎯 Goal

 Add draft messages support.

### 📝 Summary

Channel Drafts:
- `channel.draftMessage?`
- `channelController.updateDraftMessage()`
- `channelController.deleteDraftMessage()`
- `channelController.loadDraftMessage()`

Thread Drafts:
- `message.draftReply?`
- `messageController.updateDraftReply()`
- `messageController.deleteDraftReply()`
- `messageController.loadDraftReply()`

Query Drafts:
- `currentUserController.draftMessages`
- `currentUserController.deleteDraft(cid:threadId:)`
- `currentUserController.loadDraftMessages()`
- `currentUserController.loadMoreDraftMessages()`
- `currentUserControllerDelegate.didUpdateDraftMessages()`

Feature Flags:
- `Components.default.isDraftMessagesEnabled`

#### User Flow

```mermaid
flowchart LR
    A[User Action]
    A -->|Leave Chat| F[Update Channel Draft]
    A -->|Leave Thread| E[Update Thread Draft]
    A -->|Erase Input Message| D[Delete Draft]
    A -->|Publish Message| D[Delete Draft]
```

#### `CurrentUserController.loadDraftMessages()` vs `DraftMessagesListController`
There was the possibility of creating a `ChatClient.draftMessageListController()` similar to what we have with other controllers, but in this case, it is a bit overkill, and since draft messages are only for the current user, it makes sense to be part of the current user controller.

#### Reusing `ChatMessage` and `MessageDTO`
The backend opted to create a new model instead of reusing the existing Message model. But on the SDK, there are a lot of benefits in reusing the same model since the `ChatMessage` model is used everywhere, so it makes sense to reuse it. There are some properties that are not really useful for the draft message, so I also had the idea of creating a `DraftMessage` model that would wrap the `ChatMessage`, but this would require a function like `DraftMessage.toMessage()` in order to convert it to a regular massage so that it could be used in the composer etc. In the end, I don't think the extra complexity is needed.

**UPDATE:**
We decided to introduce a new `DraftMessage` model. This will not only allow more flexibility in the future by only exposing the properties that are need for draft messages, but most importantly it reduces the SDK size significantly. Since nested structs increase a lot the binary size. 

### 🎨 Showcase

https://github.com/user-attachments/assets/9f85c1bf-ef61-4c26-b5fc-7bc6f1c6bc87

### 🧪 Manual Testing Notes

**Pre-condition:**
1. Tap on Demo App Configuration
2. Change the  AppKey to `pd67s34fzpgw`
3. Change the baseURL to `https://chat-edge-frankfurt-ce1.stream-io-api.com/`

<details><summary>Update a Channel Draft Message</summary>
<p>

1. Open a Channel
4. Input some text in the composer (Optional: Include attachments, commands)
5. Leave the Channel
6. Comeback to the channel
7. The composer should have the same content as the composer
8. When going to the Channel List, the channel preview should have the draft message
9. Open the channel on another device 
10. The composer should have the same content (Besides the attachments)

</p>
</details>

<details><summary>Update a Thread Draft Reply</summary>
<p>

1. Open a Thread
2. Input some text in the composer (Optional: Include attachments, commands)
3. Leave the Thread
4. Comeback to the Thread
5. The composer should have the same content as the composer
6. When going to the Thread List, the thread preview should have the draft message
7. Open the thread on another device 
8. The composer should have the same content (Besides the attachments)

</p>
</details>

<details><summary>Fetching all Drafts</summary>
<p>

1. Open the Draft tab
2. All drafts should be loaded including ones in the thread
3. The list should be updated whenever a draft is added or removed
4. When tapping a draft from the channel, should open the channel
5. When tapping a draft from a thread, should open the thread

</p>
</details>

<details><summary>Deleting a Draft (Channel + Thread)</summary>
<p>

1. Open a Channel/Thread
2. Input some text in the composer (Optional: Include attachments, commands)
3. Leave the Channel/Thread
4. Comeback to the Channel/Thread
5. The composer should have the same content as the composer
7. Erase the content of the composer
8. Leave the Channel/Thread
11. When coming back to the Channel/Thread the composer should be empty
12. When coming back to the Channel List/Thread List the preview should not include the draft
13. The draft should disappear from the Draft List.
14. The draft should disappear on the other devices as well. (Previews and Thread List)

</p>
</details>

<details><summary>Updating a draft with composer open on another device</summary>
<p>

1. Open the same channel with the same user on two devices (A,B)
2. Write content in the composer on Device A
3. Leave the channel on Device A
4. Device B should have the composer updated with the draft from Device A

</p>
</details>

### ☑️ Contributor Checklist


martinmitrevski (commented):
Looks great so far! I think we should offer the draft list item UI as part of the components. Also, we should do this for SwiftUI.

DemoApp/StreamChat/Components/DemoChatChannelListItemView.swift
```diff
@@ -21,5 +28,23 @@ final class DemoChatChannelListItemView: ChatChannelListItemView {
         super.updateContent()
 
         backgroundColor = contentBackgroundColor
+
+        if content?.channel.draftMessage != nil {
```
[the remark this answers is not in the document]
nuno-vieira: The reason I did not added it is because it is not part of our design, but I can ask the design team if they approve

DemoApp/StreamChat/Components/DemoChatChannelListItemView.swift
```diff
@@ -21,5 +28,23 @@ final class DemoChatChannelListItemView: ChatChannelListItemView {
...
+            // Highlight Draft text in subtitleLabel using attributed string
+            let attributedString = NSMutableAttributedString(string: subtitleLabel.text ?? "")
+            if let range = (subtitleLabel.text as NSString?)?.range(of: "Draft:") {
```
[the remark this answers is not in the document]
nuno-vieira: Yes, this should be improved when it is part of the SDK

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
```diff
@@ -232,7 +234,11 @@ public class ChatChannelController: DataController, DelegateCallable, DataStoreP
             client.apiClient
         )
         pollsRepository = client.pollsRepository
-        
+        draftsRepository = self.environment.draftMessagesRepositoryBuilder(
```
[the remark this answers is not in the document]
nuno-vieira: What is the point of sharing the instance? 🤔

Sources/StreamChat/Controllers/MessageController/MessageController.swift
```diff
@@ -187,6 +187,12 @@ public class ChatMessageController: DataController, DelegateCallable, DataStoreP
     private let replyPaginationHandler: MessagesPaginationStateHandling
     private var replyPaginationState: MessagesPaginationState { replyPaginationHandler.state }
 
+    /// The drafts repository.
+    private lazy var draftsRepository: DraftMessagesRepository = environment.draftMessagesRepositoryBuilder(
```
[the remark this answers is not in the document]
nuno-vieira: Usually, repositories are not shared instances wherever they are used. Overall it does not really matter, but depending on the implementation it could introduce concurrency issues in the future. In general, it is safer if they are separate instances

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
```diff
@@ -834,6 +840,89 @@ public class ChatChannelController: DataController, DelegateCallable, DataStoreP
...
+    ///   - text: The text of the draft message.
+    ///   - attachments: The attachments of the draft message.
+    ///   - mentionedUserIds: The mentioned user ids of the draft message.
+    ///   - quotedMessageId: The message that the draft message is quoting.
+    ///   - command: The command of the draft message.
+    ///   - extraData: The extra data of the draft message.
+    ///   - completion: Called when the draft message is saved to the server.
+    public func updateDraftMessage(
+        text: String,
+        isSilent: Bool = false,
```
laevandus, line 856: `isSilent` is not documented.
Should it be just `silent: Bool` instead? Typically function arguments do not use `is`

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
```diff
@@ -1493,7 +1582,14 @@ extension ChatChannelController {
             _ database: DatabaseContainer,
             _ apiClient: APIClient
         ) -> ChannelUpdater = ChannelUpdater.init
-        
+
+        var draftMessagesRepositoryBuilder: (
+            _ database: DatabaseContainer,
+            _ apiClient: APIClient
+        ) -> DraftMessagesRepository = {
+            DraftMessagesRepository(database: $0, apiClient: $1)
```
laevandus, line 1590: Could also be shorter `= DraftMessagesRepository.init`, but does not really matter.

Sources/StreamChat/Controllers/CurrentUserController/CurrentUserController.swift
```diff
@@ -350,6 +449,21 @@ extension CurrentChatUserController {
             _ fetchedResultsControllerType: NSFetchedResultsController<CurrentUserDTO>.Type
         ) -> BackgroundEntityDatabaseObserver<CurrentChatUser, CurrentUserDTO> = BackgroundEntityDatabaseObserver.init
 
+        var draftMessagesObserverBuilder: (
+            _ database: DatabaseContainer,
+            _ fetchRequest: NSFetchRequest<MessageDTO>,
+            _ itemCreator: @escaping (MessageDTO) throws -> ChatMessage
+        ) -> BackgroundListDatabaseObserver<ChatMessage, MessageDTO> = {
+            .init(database: $0, fetchRequest: $1, itemCreator: $2)
```
laevandus, line 457: Let's add reuse keypaths as well. Otherwise each change triggers full DTO to model conversion (compared to just converting the ones which changed).

Sources/StreamChat/Controllers/MessageController/MessageController.swift
```diff
@@ -843,6 +849,76 @@ public class ChatMessageController: DataController, DelegateCallable, DataStoreP
...
+    ///   - attachments: The attachments of the draft message.
+    ///   - mentionedUserIds: The mentioned user ids of the draft message.
+    ///   - quotedMessageId: The message that the draft message is quoting.
+    ///   - showReplyInChannel: If the draft message should be shown in the channel.
+    ///   - command: The command of the draft message.
+    ///   - extraData: The extra data of the draft message.
+    ///   - completion: Called when the draft message is saved to the server.
+    public func updateDraftReply(
+        text: String,
+        isSilent: Bool = false,
```
laevandus, line 867: Same docs comment and question about if we should use `silent: Bool` instead as above.

Sources/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -974,12 +1072,123 @@ extension NSManagedObjectContext: MessageDatabaseSession {
...
+
+        do {
+            dto.extraData = try JSONEncoder.default.encode(draftDetailsPayload.extraData)
+        } catch {
+            log.error(
+                "Failed to decode extra payload for Message with id: <\(dto.id)>, using default value instead. "
+                    + "Error: \(error)"
+            )
+            dto.extraData = Data()
+        }
```
laevandus, line 1166: There is also the short `decodeRawJSON` method on the decoder for handling nil and empty data with less code.

Sources/StreamChat/Repositories/DraftMessagesRepository.swift
```diff
@@ -0,0 +1,152 @@
...
+                        }
+                        return try session
+                            .saveDraftMessage(payload: $0, for: channelId, cache: nil)
+                            .asModel()
+                    }
+                }, completion: { error in
+                    if let error {
+                        completion(.failure(error))
+                    }
+                    completion(.success(DraftListResponse(drafts: drafts, next: response.next)))
```
laevandus, line 42: Should be if else, otherwise there are 2 completion calls

Sources/StreamChat/Repositories/DraftMessagesRepository.swift
```diff
@@ -0,0 +1,152 @@
...
+                    self?.database.write { session in
+                        let draftPayload = response.draft
+                        let messageDTO = try session.saveDraftMessage(
+                            payload: draftPayload,
+                            for: cid,
+                            cache: nil
+                        )
+                        let message = try messageDTO.asModel()
+                        completion?(.success(message))
+                    }
```
laevandus, line 100: Write's completion is not used here although it probably works OK. Depends what the completion does, if it would try to read there is a race between write updating CoreData and read trying to read the same data.

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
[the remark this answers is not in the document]
martinmitrevski: the usage of drafts is not limited to the channel only, so it makes sense it has broader scope - we already do that with the `pollsRepository` above it.

Sources/StreamChat/Controllers/MessageController/MessageController.swift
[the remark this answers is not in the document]
martinmitrevski: what kind of concurrency issues? I don't see any mutable state here 🤔

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
martinmitrevski: I agree, plus I would add that we should follow the naming of backend API. I assume it's `silent` there.

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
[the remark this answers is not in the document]
nuno-vieira: YEs, that is why it is used here in multiple places as well. But sharing instancs it is recipe for concurrency issues

Sources/StreamChat/Controllers/MessageController/MessageController.swift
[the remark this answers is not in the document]
nuno-vieira: At the moment, there is none (At first sight), but in future, there will be. I don't see any advantage in sharing instances here. I would prefer to be safer

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
nuno-vieira: Although I agree, we already have `silent` everywhere in other places. So I think it is better we keep the consistency here. Otherwise it will look like it is something different

Sources/StreamChat/Controllers/CurrentUserController/CurrentUserController.swift
nuno-vieira: I did not use it because performance here is not really important, but I can add it if you think is necessary

Sources/StreamChat/Repositories/DraftMessagesRepository.swift
nuno-vieira: Ah, I forgot the return clause, nice catch. Tests would have caught this, but I did not added them yet :D

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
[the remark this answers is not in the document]
martinmitrevski: I don't find your arguments strong enough:
- we have all repositories in the client atm (except this one)
- there are no concurrency issues whatsoever
- the scope is bigger than the place of the creation
- and we do this 3 times unnecessarily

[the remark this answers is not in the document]
nuno-vieira: Threads also use one repository instance, but to keep consistency with the rest, I'll move it. Even tho this will complicate testing a bit

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
nuno-vieira: Fixed ✅

Sources/StreamChat/Controllers/CurrentUserController/CurrentUserController.swift
nuno-vieira: Done ✅

Sources/StreamChat/Repositories/DraftMessagesRepository.swift
nuno-vieira: Solves

Sources/StreamChat/Repositories/DraftMessagesRepository.swift
nuno-vieira: Solved

Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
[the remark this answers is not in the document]
nuno-vieira: Done

Sources/StreamChat/Repositories/DraftMessagesRepository.swift
```diff
@@ -0,0 +1,161 @@
...
+                            for: cid,
+                            cache: nil
+                        )
+                        let message = try messageDTO.asModel()
+                        completion?(.success(message))
+                    }, completion: { error in
+                        if let error {
+                            completion?(.failure(error))
+                        }
+                    })
```
laevandus, line 105: I think it would be nice to move the completion call of the success case to the write's completion. This ensures that Core Data store has saved the changes before the completion is called. It is important for state layer because then the DB observer would fire before the completion is called.
Otherwise something like this would not work:
```
try await connectedUser.updateDraft()
let drafts = connectedUser.state.drafts
```

Sources/StreamChat/Repositories/DraftMessagesRepository.swift
```diff
@@ -0,0 +1,161 @@
...
+    ) {
+        apiClient.request(
+            endpoint: .deleteDraftMessage(channelId: cid, threadId: threadId)
+        ) { [weak self] result in
+            switch result {
+            case .success:
+                self?.database.write { session in
+                    session.deleteDraftMessage(in: cid, threadId: threadId)
+                }
+                completion(nil)
```
laevandus, line 155: Would also be good to move it to write's completion block (see previous comment for details)

nuno-vieira: Fixed

Sources/StreamChat/Repositories/DraftMessagesRepository.swift
```diff
@@ -0,0 +1,161 @@
...
+            switch result {
+            case .success(let response):
+                self?.database.write({ session in
+                    let messageDTO = try session.saveDraftMessage(
+                        payload: response.draft,
+                        for: cid,
+                        cache: nil
+                    )
+                    let message = try messageDTO.asModel()
+                    completion?(.success(message))
```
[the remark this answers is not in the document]
nuno-vieira: Fixed

Sources/StreamChat/Repositories/DraftMessagesRepository.swift
nuno-vieira: Fixed

martinmitrevski (commented):
Looks great ✅ As discussed, few open points before merging, especially the size impact, but other than that, great work!

Sources/StreamChat/Repositories/DraftMessagesRepository.swift
```diff
@@ -0,0 +1,165 @@
...
+    ) {
+        apiClient.request(
+            endpoint: .deleteDraftMessage(channelId: cid, threadId: threadId)
+        ) { [weak self] result in
+            switch result {
+            case .success:
+                self?.database.write { session in
+                    session.deleteDraftMessage(in: cid, threadId: threadId)
+                }
+                completion(nil)
```
laevandus, line 159: Probably fine for now, but causes the same issue as mentioned before for state layer - db observer reacts after action has finished. Calling from write's completion is the safest, but can be changed later as well if needed

Sources/StreamChat/Repositories/DraftMessagesRepository.swift
```diff
@@ -0,0 +1,165 @@
...
+                    let messageDTO = try session.saveDraftMessage(
+                        payload: response.draft,
+                        for: cid,
+                        cache: nil
+                    )
+                    draft = try messageDTO.asModel()
+                }) { error in
+                    if let draft {
+                        completion?(.success(draft))
+                    } else if let error {
```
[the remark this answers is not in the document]
nuno-vieira: That is not possible because if the draft is not filled, it means there was an error inside the try/catch block

Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
```diff
@@ -149,38 +154,34 @@ open class ChatChannelListItemView: _View, ThemeProvider, SwiftUIRepresentable {
...
-            if let pollPreviewText = pollAttachmentPreviewText(for: previewMessage) {
-                return pollPreviewText
-            }
-
-            if isLastMessageVoiceRecording {
-                return previewMessageForAudioRecordingMessage(messageText: previewMessage.text)
-            }
+        if let dratMessage = content.channel.draftMessage {
+            let previewText = previewMessageContextText(previewMessage: dratMessage)
+            return previewMessageTextForDraft(messageText: previewText)
```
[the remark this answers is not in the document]
nuno-vieira: I was trying to avoid in using the flag everywhere, but yeah that is a good point.

Sources/StreamChat/Repositories/DraftMessagesRepository.swift
nuno-vieira: ✅

Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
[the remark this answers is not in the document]
nuno-vieira: Done ✅

Sources/StreamChat/Models/DraftMessage.swift
```diff
@@ -0,0 +1,170 @@
...
+    public let showReplyInChannel: Bool
+
+    /// Additional data associated with the message.
+    public let extraData: [String: RawJSON]
+
+    /// Quoted message.
+    ///
+    /// If message is inline reply this property will contain the message quoted by this reply.
+    public var quotedMessage: ChatMessage? { _quotedMessage() }
+    let _quotedMessage: () -> ChatMessage?
```
laevandus, line 46: This should be `public let quotedMessage: ChatMessage?` because we have different types now.

Sources/StreamChat/Models/DraftMessage.swift
```diff
@@ -0,0 +1,170 @@
...
+        extraData = message.extraData
+        _quotedMessage = { message.quotedMessage }
+        mentionedUsers = message.mentionedUsers
+        attachments = message.allAttachments
+        author = message.author
+    }
+}
+
+extension DraftMessage: Equatable {
+    public static func == (lhs: DraftMessage, rhs: DraftMessage) -> Bool {
```
laevandus, line 111: If _quotedMessage closure is removed then all of this is not needed

Sources/StreamChatUI/Utils/UILabel+highlightText.swift
```diff
@@ -0,0 +1,41 @@
...
+        if let currentFont = font {
+            attributedString.addAttribute(.font, value: currentFont, range: fullRange)
+        }
+
+        let highlightRange = currentTextString.range(of: text)
+        attributedString.addAttribute(
+            .foregroundColor,
+            value: options.color,
+            range: highlightRange
+        )
```
laevandus, line 31: Did not realise that this is safe here where highlightRange can be `NSRange(location: NSNotFound, length: 0)`

martinmitrevski (approved):
Things are looking great from my side ✅ We just need to add the docs and fix the comments from Toomas. Would be great if @testableapple does another run of QA.

Sources/StreamChat/Models/DraftMessage.swift
nuno-vieira: Actually we can't 😢 Because `ChatMessage` contains `draftReply` which is `DraftMessage`. So we have the same problem.

Sources/StreamChat/Controllers/CurrentUserController/CurrentUserController.swift
```diff
@@ -334,8 +359,79 @@ public extension CurrentChatUserController {
         }
     }
 
-    private func createMemberUpdater() -> ChannelMemberUpdater {
-        .init(database: client.databaseContainer, apiClient: client.apiClient)
+    /// Loads the draft messages for the current user.
```
laevandus, line 362: Minor: the first `the` can be removed?

laevandus (approved):
✅ Did some testing and worked as we would expect.

martinmitrevski (approved):
LGTM! ✅

PR #3589, opened 2025-02-11
[CI] Snapshots recording improvement

testableapple:
### 🎯 Goal

Avoid re-recording all the snapshots on CI

### ✅ Test

- [Soft-recording](https://github.com/GetStream/stream-chat-swift/actions/runs/13262103938/job/37020948565)
- [Hard-recording](https://github.com/GetStream/stream-chat-swift/actions/runs/13262560489/job/37022450912)

### DIFF
--- a/.github/workflows/smoke-checks.yml
+++ b/.github/workflows/smoke-checks.yml
@@ -9,12 +9,12 @@ on:
   workflow_dispatch:
     inputs:
       record_snapshots:
-        description: 'Soft-record snapshots on CI?'
+        description: 'Record missing snapshots on CI?'
         type: boolean
         required: false
         default: false
       remove_and_record_snapshots:
-        description: 'Hard-record snapshots on CI?'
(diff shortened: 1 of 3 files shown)


.github/workflows/smoke-checks.yml
```diff
@@ -8,8 +8,13 @@ on:
...
     inputs:
-      snapshots:
-        description: 'Should Snapshots be recorded on CI?'
+      record_snapshots:
+        description: 'Soft-record snapshots on CI?'
+        type: boolean
+        required: false
+        default: false
+      remove_and_record_snapshots:
+        description: 'Hard-record snapshots on CI?'
```
laevandus, line 17: Maybe it would be clearer to use "Record missing snapshots on CI?" and "Reset all the snapshots on CI?"

testableapple: Updated 👍

laevandus (approved):

PR #3595, opened 2025-02-13
Add sending messages to only specific members

martinmitrevski:
### 🔗 Issue Links

Resolves https://linear.app/stream/issue/IOS-688/private-messages-sdk-support.

### 🎯 Goal

Allow sending messages only to a specific member of a channel.

### 📝 Summary

Exposed new property that's available to support this feature.

Important note: there's a bug on the backend for our filter, so the message appears in the channel list for other members. This is being currently addressed by the backend team, and we should wait before merging. 

### 🛠 Implementation

Just exposes a new field.

### 🎨 Showcase

| Before | After |
| ------ | ----- |
|  img   |  img  |

### 🧪 Manual Testing Notes

There's a new action in the channel, "Say Hi to a specific member". You need to enter a valid member id and verify that only that member received the message in the channel.

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
+++ b/Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
@@ -747,6 +747,7 @@ public class ChatChannelController: DataController, DelegateCallable, DataStoreP
     ///   - quotedMessageId: An id of the message new message quotes. (inline reply)
     ///   - skipPush: If true, skips sending push notification to channel members.
     ///   - skipEnrichUrl: If true, the url preview won't be attached to the message.
+    ///   - restrictedVisibility: The list of user ids that should be able to see the message.
     ///   - extraData: Additional extra data of the message object.
     ///   - completion: Called when saving the message to the local DB finishes.
     ///
@@ -760,6 +761,7 @@ public class ChatChannelController: DataController, DelegateCallable, DataStoreP
         quotedMessageId: MessageId? = nil,
         skipPush: Bool = false,
         skipEnrichUrl: Bool = false,
+        restrictedVisibility: [UserId] = [],
         extraData: [String: RawJSON] = [:],
         completion: ((Result<MessageId, Error>) -> Void)? = nil
     ) {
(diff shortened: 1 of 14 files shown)


Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
```diff
@@ -760,6 +761,7 @@ public class ChatChannelController: DataController, DelegateCallable, DataStoreP
         quotedMessageId: MessageId? = nil,
         skipPush: Bool = false,
         skipEnrichUrl: Bool = false,
+        restrictedVisibility: [UserId] = [],
```
nuno-vieira, line 764: Should we also expose this on `createSystemMessage()` ? I'm not sure, but I think I've heard some customers wanted this for system messages?

martinmitrevski: that's a great point, we should add it yes

nuno-vieira (approved):
LGTM! ✅

Sources/StreamChat/StateLayer/Chat.swift
```diff
@@ -501,7 +501,8 @@ public class Chat {
         silent: Bool = false,
         skipPushNotification: Bool = false,
         skipEnrichURL: Bool = false,
-        messageId: MessageId? = nil
+        messageId: MessageId? = nil,
+        restrictedVisibility: [UserId] = []
```
laevandus, line 505: Could we please add a doc comment for this argument as well

Sources/StreamChat/StateLayer/Chat.swift
```diff
@@ -536,6 +538,7 @@ public class Chat {
     public func sendSystemMessage(
         with text: String,
         messageId: MessageId? = nil,
+        restrictedVisibility: [UserId] = [],
```
laevandus, line 541: Could we please add a doc comment for this argument as well

laevandus (approved):
I was thinking about if replies should be supported as well, but not sure what could be the use-case there. 
Added two minor comments about docs.
✅ Otherwise looks good

Sources/StreamChat/StateLayer/Chat.swift
martinmitrevski: done

Sources/StreamChat/StateLayer/Chat.swift
martinmitrevski: done

testableapple (changes requested):
There are a couple of issues with this PR. Discussing on Slack.

testableapple (approved):
LGTM, the existing issues are on the backend side.

PR #3601, opened 2025-02-20
Fetch only required data in ChatRemoteNotificationHandler (no writes to persistent store)

laevandus:
### 🔗 Issue Links

Resolves: [IOS-70](https://linear.app/stream/issue/IOS-70)

### 🎯 Goal

Fetch only required data and avoid updating persistent store from app extension (more reliable)

### 📝 Summary

- Do not sync channel events for 100 channels when push notification extension runs (uses ChatRemoteNotificationHandler). Instead fetch the channel data directly.  
- Do not write to persistent store from app extensions (reliability increase, concrete example is issues with previewMessage of a channel getting cleared when merge policy merges changes)
- Each notification now triggers 1 query channel call whereas before it was a call to the /sync endpoint
- Remove `NotificationExtensionLifecycle` which is not used anymore for sharing web-socket state of the app

### 🛠 Implementation

When push notification is received and app has push notification extension, then `ChatRemoteNotificationHandler` is used for fetching `ChatMessage` and `ChatChannel` for the push notification payload. Currently message is fetched using an API call, but channel is read from the persistent store after making a call to the /sync endpoint to get missing channel events. This writes new data to the persistent store from an app extension and has shown issues of data consistency when the app is brought back to foreground and it fetches new data. Since persistent store has changed, app's managed object context runs the merge policy logic to resolves conflicts in just fetched data and data updated by the app extension (recently it was discovered that the default merge policy ends up erasing previewMessage property in that case (really hard to reproduce)). The main goal is to leave updating the persistent store just for the app, and not doing it from app extensions. In the future we can cut out the persistent store access completely which makes it also faster.

### 🧪 Manual Testing Notes

1. Launch the app, login with A, put the app to the background
2. Send a message to a channel A is part of
Push notification with title containing *channel name* is presented (if not, push extension failed to get channel information)
3. Tap on the notification and channel is presented

### ☑️ Contributor Checklist


Sources/StreamChat/APIClient/ChatRemoteNotificationHandler.swift
```diff
@@ -70,18 +70,14 @@ public class ChatRemoteNotificationHandler {
     var client: ChatClient
     var content: UNNotificationContent
     let chatCategoryIdentifiers: Set<String> = ["stream.chat", "MESSAGE_NEW"]
-    let database: DatabaseContainer
-    let syncRepository: SyncRepository
+    let channelRepository: ChannelRepository
     let messageRepository: MessageRepository
-    let extensionLifecycle: NotificationExtensionLifecycle
```
laevandus, line 76: If we go for the proposed change, all the extension lifecycle code can be deleted as well.

Sources/StreamChat/APIClient/Endpoints/ChannelEndpoints.swift
```diff
@@ -28,7 +28,7 @@ extension Endpoint {
             path: path,
             method: .post,
             queryItems: nil,
-            requiresConnectionId: query.options.contains(oneOf: [.presence, .state, .watch]),
+            requiresConnectionId: query.options.contains(oneOf: [.presence, .watch]),
```
laevandus, line 31: Query channel works without connection id if I just want to get the data (e.g. no watching)

Sources/StreamChat/Database/DatabaseContainer.swift
```diff
@@ -222,6 +222,19 @@ class DatabaseContainer: NSPersistentContainer, @unchecked Sendable {
...
+        write { session in
+            result = try actions(session)
+        } completion: { error in
+            if let result {
+                completion(.success(result))
+            } else {
+                completion(.failure(error ?? ClientError.Unknown()))
+            }
+        }
+    }
```
laevandus, line 237: Convenience for cases where the actions should return the model.

Sources/StreamChat/Repositories/MessageRepository.swift
```diff
@@ -242,7 +242,7 @@ class MessageRepository {
                     if !store {
                         // Force load attachments before discarding changes
                         _ = message?.attachmentCounts
-                        self.database.writableContext.discardCurrentChanges()
+                        self.database.writableContext.rollback()
```
laevandus, line 245: Realised that `context.save()` was happening here as well when it should have not. `discardCurrentChanges()` is our own addition.

Sources/StreamChat/APIClient/ChatRemoteNotificationHandler.swift
```diff
@@ -118,23 +114,18 @@ public class ChatRemoteNotificationHandler {
...
-                completion(nil, nil)
-                return
-            }
-
-            self?.syncRepository.syncExistingChannelsEvents { _ in
-                database.backgroundReadOnlyContext.perform {
-                    let channel = try? ChannelDTO.load(cid: cid, context: database.backgroundReadOnlyContext)?.asModel()
-                    completion(message, channel)
+    
+    private func getContent(cid: ChannelId, messageId: MessageId, completion: @escaping (ChatMessage?, ChatChannel?) -> Void) {
```
laevandus, line 118: TODO: Test tapping on the notification

Sources/StreamChat/APIClient/Endpoints/ChannelEndpoints.swift
martinmitrevski: this might be risky to change at this point, since customers might stop receiving events if they had only `state`.

Sources/StreamChat/Repositories/SyncRepository.swift
```diff
@@ -209,45 +207,11 @@ class SyncRepository {
         operationQueue.addOperations(operations, waitUntilFinished: false)
     }
 
-    /// Syncs the events for the active chat channels using the last sync date.
-    /// - Parameter completion: A block that will get executed upon completion of the synchronization
-    func syncExistingChannelsEvents(completion: @escaping (Result<[ChannelId], SyncError>) -> Void) {
```
martinmitrevski, line 214: how do we get the channel ids for the sync endpoint now? (the one we trigger on app start)

Sources/StreamChat/APIClient/ChatRemoteNotificationHandler.swift
```diff
@@ -118,23 +114,18 @@ public class ChatRemoteNotificationHandler {
...
+        var query = ChannelQuery(cid: cid, pageSize: 10, membersLimit: 10)
+        query.options = .state
+        channelRepository.getChannel(for: query, store: false) { [messageRepository] channelResult in
+            let channel = channelResult.value
+            // When message is already available, skip fetching the message
+            if let message = channel?.latestMessages.first(where: { $0.id == messageId }) {
```
nuno-vieira, line 124: We might have the message already available, but is the content updated though?

Sources/StreamChat/Repositories/MessageRepository.swift
nuno-vieira: This might be risky, I remember I've tested rollback a few times and it did not work as I expected. I tested it when doing optimistic updates 🤔

Sources/StreamChat/Repositories/SyncRepository.swift
nuno-vieira: AFAIK this was also called when coming back from background 🤔 Or maybe I'm wrong?

nuno-vieira (commented):
Overall it is looking good, just added some comments to clarify things 👍

Sources/StreamChat/APIClient/Endpoints/ChannelEndpoints.swift
laevandus: True, I'll see if I can do something else instead.

Sources/StreamChat/APIClient/ChatRemoteNotificationHandler.swift
laevandus: Since `getChannel` does channel fetch using an API call, then it must be up to date. 🤔

Sources/StreamChat/Repositories/MessageRepository.swift
laevandus: Maybe save() was called before in that case? Then it would not do what we would expect. I'll review, maybe calling reset() would be better here.

Sources/StreamChat/APIClient/Endpoints/ChannelEndpoints.swift
laevandus: Discarded this change in favour of using separate endpoint instance.

Sources/StreamChat/Repositories/MessageRepository.swift
laevandus: Re-read docs and for a case like here where I add some changes to the context and then want to discard these changes without saving, then rollback seems to be exactly what I need. 🤔

Sources/StreamChat/Repositories/SyncRepository.swift
laevandus: There is a separate operation to get relevant channel ids for /sync. Relevant are only the ones we have channel controller of Chat instance. Channel lists are always refetched.

martinmitrevski (approved):
LGTM! ✅

nuno-vieira (approved):
LGTM! ✅

testableapple (approved):

PR #3613, opened 2025-03-05
Add channel reads pagination support

laevandus:
### 🔗 Issue Links

Resolves: [IOS-666](https://linear.app/stream/issue/IOS-666/)

### 🎯 Goal

Add support for paginating channel reads for accurate read counts in large channels

### 📝 Summary

> [!IMPORTANT]
> Channel reads pagination is only supported through channel query endpoint while paginating members (slightly hidden feature)

* Use channel query endpoint in `ChatChannelController.loadChannelReads(pagination:completion:)`, `ChatChannelController.loadMoreChannelReads(limit:completion:)`, `Chat.loadMembers(with:)` and `Chat.loadMoreMembers(limit:)` which also paginates channel reads for members
* Update `ChannelQuery` to support setting pagination parameter for members
* Do not reset `ChannelDTO.reads` when saving channel payloads, only reset it when loading the first page of messages using the same endpoint (aka `ChatChannelController.synchronize` or `Chat.get(watch:)`)
* Recalculate `MessageDTO.readBy` for existing messages when saving channel payloads
* Add demo app debug command for loading more members 

### 🛠 Implementation

We use `ChatMessage.readBy` for showing read counts for current user's messages. This property is updated using `ChannelDTO.reads` (manual filtering). Channel reads are fetched based on requested channel members and by default up to 100 channel members are fetches (and reads for these users). If there are many members and we would like to get accurate channel reads, we would need to paginate members using channel query endpoint (members endpoint does not give channel reads).
`Chat` already incorporates member pagination. Before it was using /members endpoint, but this PR changes it to use the channel query endpoint. 
End result is that creating an instance of `Chat` and calling `get(watch:)` will automatically give access to channel members fetched using the regular channel query call (default is 100 members and their channel reads). If we want more, we just need to call `loadMoreMembers(limit:)` and we get next page of members and their channel read information.

```swift
let chat = client.makeChat(for: cid)
try await chat.get(watch: true)
let members = chat.state.members // first page, 100
let reads = chat.state.channel?.reads // first 100, `ChatMessage.readBy` is updated based on 100 reads
try await chat.loadMoreMembers(limit: 100)
let members2 = chat.state.members // 200
let reads2 = chat.state.channel?.reads // 200, `ChatMessage.readBy` is updated based on 200 reads
```

Similar support also for controllers.
```swift
channelController.synchronize { error in
    if let error {
        // handle
        return
    }
    let reads = channelController.channel?.reads // 100
    channelController.loadMoreChannelReads(limit: 100) { paginationError in
        if let paginationError {
            // handle
            return
        }
        let reads = channelController.channel?.reads // 200
    }
}
```

### 🎨 Showcase

https://github.com/user-attachments/assets/68403bd1-4036-4559-96c8-068e14eef9bc

### 🧪 Manual Testing Notes

1. Log in with Chewbacca and open read receipts 2 channel which has a lot of reads
2. Observe how loading more users updates channel reads (Debug menu > Load more members)

Verify that the typical flow of sending a message and other participant reading it, triggers read receipt (double checkmark).

### ☑️ Contributor Checklist


Sources/StreamChat/Workers/ChannelUpdater.swift
```diff
@@ -51,9 +51,10 @@ class ChannelUpdater: Worker {
             paginationStateHandler.begin(pagination: pagination)
         }
         
-        let didLoadFirstPage = channelQuery.pagination?.parameter == nil
-        let didJumpToMessage: Bool = channelQuery.pagination?.parameter?.isJumpingToMessage == true
-        let resetMembers = didLoadFirstPage
+        let didRequestMessages = channelQuery.pagination?.pageSize != 0 // no pagination returns default set of messages
```
martinmitrevski, line 54: This feels so complicated 🤯 Can you explain a bit more? Also maybe add additional comments.

martinmitrevski (commented):
Looks good in general, but would be nice to provide more details about some parts.

Sources/StreamChat/Workers/ChannelUpdater.swift
laevandus: I'll see if I can make it simpler and possibly separate member pagination from the main update method.

Sources/StreamChat/Workers/ChannelUpdater.swift
```diff
@@ -69,12 +69,22 @@ class ChannelUpdater: Worker {
...
+                    
                     if let channelDTO = session.channel(cid: payload.channel.cid) {
                         if resetMessages {
                             channelDTO.cleanAllMessagesExcludingLocalOnly()
                         }
-                        if resetMembers {
+                        if resetMembersAndReads {
+                            memberListQueryDTO?.members.removeAll()
                             channelDTO.members.removeAll()
+                            channelDTO.reads.removeAll()
```
laevandus, line 87: Note that this used to happen every time saveChannel was called, but since we want to paginate reads, it needs to have similar logic as other properties we paginate.

Sources/StreamChat/Workers/ChannelUpdater.swift
laevandus: Changed it around and separate members and reads pagination calls from the main update call (aka synchronize).

martinmitrevski (approved):
LGTM ✅  Good if also @nuno-vieira has a look, since he knows this part of the codebase better.

DemoApp/StreamChat/Components/DemoChatChannelListRouter.swift
```diff
@@ -212,6 +213,19 @@ final class DemoChatChannelListRouter: ChatChannelListRouter {
...
+                    let chat = client.makeChat(for: cid)
+                    do {
+                        try await chat.loadMoreMembers(limit: 100)
+                    } catch {
+                        self.rootViewController.presentAlert(
+                            title: "Couldn't load more members to channel \(cid)",
+                            message: "\(error)"
+                        )
+                    }
+                }
```
nuno-vieira, line 227: Isn't it better to use the controllers here? All the examples use the controllers, and it is also what most customers use, so it makes sense we test often what is most used by our customers 🤔 Any reason why we used the state layer for this case?

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### ✅ Added
 - Add `MarkdownParser` for parsing and styling markdown strings [#3590](https://github.com/GetStream/stream-chat-swift/pull/3590)
 - Add `Fonts.title2` for supporting markdown headers [#3590](https://github.com/GetStream/stream-chat-swift/pull/3590)
+- Add channel reads pagination support through `Chat.loadMembers(with:)` and `Chat.loadMoreMembers(limit:)` [#3613](https://github.com/GetStream/stream-chat-swift/pull/3613)
```
nuno-vieira, line 10: Does this mean that using the regular controller it is not possible?

testableapple:
I did some QA and it worked as described in the PR description, but would be great if @nuno-vieira could also check this out.

CHANGELOG.md
laevandus: Added controller support

DemoApp/StreamChat/Components/DemoChatChannelListRouter.swift
laevandus: Added controller support

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### ✅ Added
 - Add `MarkdownParser` for parsing and styling markdown strings [#3590](https://github.com/GetStream/stream-chat-swift/pull/3590)
 - Add `Fonts.title2` for supporting markdown headers [#3590](https://github.com/GetStream/stream-chat-swift/pull/3590)
+- Add channel reads pagination support through `ChatChannelController.loadChannelReads(pagination:completion:)`, `ChatChannelController.loadMoreChannelReads(limit:completion:)`, `Chat.loadMembers(with:)`, and `Chat.loadMoreMembers(limit:)` [#3613](https://github.com/GetStream/stream-chat-swift/pull/3613)
```
nuno-vieira, line 10: Nit:
```suggestion
- Add channel reads pagination support [#3613](https://github.com/GetStream/stream-chat-swift/pull/3613)
    - Add `ChatChannelController.loadChannelReads()`,
    - Add `ChatChannelController.loadMoreChannelReads()`
    - Add `Chat.loadMembers()`
    - Add `Chat.loadMoreMembers()
```

nuno-vieira (approved):

PR #3618, opened 2025-03-13
Fix channel membership not updated when current user adds itself to a channel

nuno-vieira:
### 🔗 Issue Links

https://linear.app/stream/issue/IOS-745/issue-with-chatchannelcontroller-not-updating-membership-property

### 🎯 Goal

Fix channel membership not updated when the current user adds itself to a channel.

### 🛠 Implementation

The root of the problem is that the backend does not report a `notification.added_to_channel` event; instead, it reports a regular `member.added` event for when the current user adds itself to the channel.

But, we can provide a fallback to fix this, and update the membership in case the event is a regular `member.added` event.

### 🧪 Manual Testing Notes

N/A. Covered by tests.

### ☑️ Contributor Checklist


laevandus (approved):
✅ Looks good although not sure what is the state of the CI

PR #3623, opened 2025-03-21
Add Message Reminders

nuno-vieira:
### 🔗 Issue Links
IOS-696

### 🎯 Goal

Add message reminders feature. This will allow users to save messages for later and get notified about them.

### 📝 Summary

Public:
- Adds new `Filter.isNil` to make it easier to filter nil values
- Adds `ChatMessageController.createReminder()`
- Adds `ChatMessageController.updateReminder()`
- Adds `ChatMessageController.deleteReminder()`
- Adds `ChatCurrentUserController.loadReminders()` + `loadMoreReminders()`

Internal:
 Renames `Filter+ChatChannel.swift` to `Filter+predicate` so that we can start using it for other data and not only channels. For the Reminders feature it is quite useful, since most likely customers will have multiple reminders lists at the same time, like in our demo app.

### 🛠 Implementation

TODO

### 🎨 Showcase

TODO

### 🧪 Manual Testing Notes

TODO

### ☑️ Contributor Checklist


DemoApp/Screens/DemoAppTabBarController.swift
```diff
@@ -65,8 +83,53 @@ class DemoAppTabBarController: UITabBarController, CurrentChatUserControllerDele
...
+    }
+    
+    // MARK: - EventsControllerDelegate
+    
+    func eventsController(_ controller: EventsController, didReceiveEvent event: Event) {
+        if AppConfig.shared.demoAppConfig.isRemindersEnabled,
+           let reminderDueEvent = event as? ReminderDueEvent {
+            handleReminderDueEvent(reminderDueEvent)
+        }
+    }
```
nuno-vieira, line 118: I'm not sure if this is going to stay the same since the goal is for the backend to send a remote push notification.

DemoApp/Screens/DemoReminderListVC.swift
```diff
@@ -0,0 +1,691 @@
...
+
+import StreamChat
+import StreamChatUI
+import UIKit
+
+class DemoReminderListVC: UIViewController, ThemeProvider {
+    var onLogout: (() -> Void)?
+    var onDisconnect: (() -> Void)?
+
+    private let currentUserController: CurrentChatUserController
```
nuno-vieira, line 13: At the moment, one controller is managing the different types of reminders. Although this works fine, we could also use a controller per reminder type. This way we would avoid having to many loadReminders requests. Since right now, whenever we switch the filter, we re-refetch the reminders. So yeah I might optimize this. WDYT?

Sources/StreamChat/APIClient/Endpoints/MessageEndpoints.swift
```diff
@@ -94,6 +94,54 @@ extension Endpoint {
...
+        )
+    }
+    
+    // Queries reminders with the provided parameters
+    static func queryReminders(query: MessageReminderListQuery) -> Endpoint<RemindersQueryPayload> {
+        .init(
+            path: .reminders,
+            method: .post,
+            queryItems: nil,
+            requiresConnectionId: false,
```
martinmitrevski, line 139: how do we receive reminder events if we don't require connection id?

Sources/StreamChat/APIClient/Endpoints/Payloads/MessagePayloads.swift
```diff
@@ -382,3 +386,68 @@ public struct Command: Codable, Hashable {
         self.args = args
     }
 }
+
+/// An object describing a reminder JSON payload.
+struct ReminderPayload: Decodable {
```
martinmitrevski, line 391: As mentioned on Slack, let's try using class here.

Sources/StreamChat/Query/MessageReminderListQuery.swift
```diff
@@ -0,0 +1,130 @@
...
+    /// A filter key for matching the `remind_at` value.
+    /// Supported operators: `equal`, `greaterThan`, `lessThan`, `greaterOrEqual`, `lessOrEqual`
+    static var remindAt: FilterKey<Scope, Date> { .init(rawValue: "remind_at", keyPathString: #keyPath(MessageReminderDTO.remindAt)) }
+
+    /// A filter key for matching the `created_at` value.
+    /// Supported operators: `equal`, `greaterThan`, `lessThan`, `greaterOrEqual`, `lessOrEqual`
+    static var createdAt: FilterKey<Scope, Date> { .init(rawValue: "created_at", keyPathString: #keyPath(MessageReminderDTO.createdAt)) }
+}
+
+public extension Filter where Scope: AnyMessageReminderListFilterScope {
```
martinmitrevski, line 36: how did you come up with these filters?

martinmitrevski (commented):
Looks great so far 👍 I think the most important thing is to check how / when push notifications will be supported - the due feature doesn't make much sense without it.
Additionally, apart from docs, we need a notion doc for other SDKs.

DemoApp/Screens/DemoReminderListVC.swift
nuno-vieira: The thing is that we should show them how to optimize their integration code. This means less calls, and less stress to our servers 🤔

DemoApp/Screens/AppConfigViewController/AppConfigViewController.swift
```diff
@@ -50,7 +52,8 @@ class AppConfig {
             isLocationAttachmentsEnabled: false,
             tokenRefreshDetails: nil,
             shouldShowConnectionBanner: false,
-            isPremiumMemberFeatureEnabled: false
+            isPremiumMemberFeatureEnabled: false,
+            isRemindersEnabled: false
```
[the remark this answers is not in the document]
nuno-vieira: They are disabled, and enabled by default using our Developers Scheme. Since the design is not approved by our Designers, it is probably not a good idea to make it enabled by default 🤔 Although that applies to drafts as well

Sources/StreamChat/APIClient/Endpoints/MessageEndpoints.swift
nuno-vieira: Yeah this might need to be true, will need to recheck

nuno-vieira: Although, I think reminders will be probably receive events even if you don't query them. Since you might not have a reminders list. But still want to get reminded of your reminders. So this is not needed AFAIK

Sources/StreamChat/Query/MessageReminderListQuery.swift
nuno-vieira: It is in the Notion's Backend Specs

nuno-vieira: Ah, or do you mean the helpers? If you mean the helpers, it is because it will be the most common filters, so I think it should be useful for customers

DemoApp/Screens/DemoReminderListVC.swift
laevandus: I wrote a longer reply to our Slack thread with a summary of supporting the idea of MessageReminderListController (mainly because reminders query has a filter)

DemoApp/Screens/AppConfigViewController/AppConfigViewController.swift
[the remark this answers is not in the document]
martinmitrevski: I would enable it for the default scheme as well. It helps with discoverability, and, let's be honest, you did a great job in the design part!

DemoApp/Screens/DemoReminderListVC.swift
martinmitrevski: ok, that's also fine with me 👍

DemoApp/Screens/DemoAppTabBarController.swift
martinmitrevski: do you know when that would be ready?

Sources/StreamChat/APIClient/Endpoints/MessageEndpoints.swift
martinmitrevski: ok, just checking - if it works without the connection id we can resolve this one.

DemoApp/Screens/AppConfigViewController/AppConfigViewController.swift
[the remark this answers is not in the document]
nuno-vieira: Ahah thanks, okay I will enable it by default 👍

DemoApp/Screens/DemoAppTabBarController.swift
nuno-vieira: I've deleted this part of the codebase. Push should be working eventually, and we should only merge once that is working, but at the moment the backend is still analysing why we don't receive a remote push.

martinmitrevski (approved):
Looks great and works well, don't have any remarks about the code 👍 
Before merging we should:
- make sure push works (also good if we retest again)
- have docs PR ready 
- make sure things are deployed on every shard 
Btw, I don't think we need to implement this in SwiftUI as well. A lot of work, while the feature is mostly LLC - good docs and this sample should be enough for SwiftUI.

Sources/StreamChat/Repositories/RemindersRepository.swift
```diff
@@ -0,0 +1,250 @@
...
+    ///   - query: The query containing filtering and sorting parameters.
+    ///   - completion: Called when the operation completes.
+    func loadReminders(
+        query: MessageReminderListQuery,
+        completion: @escaping (Result<ReminderListResponse, Error>) -> Void
+    ) {
+        apiClient.request(endpoint: .queryReminders(query: query)) { [weak self] result in
+            switch result {
+            case .success(let response):
+                var reminders: [MessageReminder] = []
```
laevandus, line 41: There is also convenience write(converting: for returning the model type in the actions closure and having result type for the completion handler. Meaning, separate variable is not needed for capturing models.

Sources/StreamChat/Controllers/MessageReminderListController/MessageReminderListController+Combine.swift
```diff
@@ -0,0 +1,49 @@
+//
+// Copyright © 2025 Stream.io Inc. All rights reserved.
+//
+
+import Combine
+import Foundation
+
+@available(iOS 13, *)
```
laevandus, line 8: These should not be needed because our deployment target is 13 already

laevandus (approved):
✅ Looks nicer with the dedicated controller

Sources/StreamChat/Controllers/MessageReminderListController/MessageReminderListController+Combine.swift
nuno-vieira: Good catch 👍

Sources/StreamChat/Repositories/RemindersRepository.swift
nuno-vieira: DOne ✅

nuno-vieira:
Everything is Ready, but it is blocked by Push v3. So converting it to draft.

PR #3625, opened 2025-03-24
Fix local draft attachments erased when draft updated event is triggered

nuno-vieira:
### 🔗 Issue Links
Resolves IOS-757

### 🎯 Goal

Fix local draft attachments being deleted when navigating back to the channel list.

### 🛠 Implementation

The issue was that when navigating to the Channel List, it would save the draft message, but the draft.update event would be triggered as well, and we were wrongly deleting the existing local draft, which would erase the local attachments. 

### 🧪 Manual Testing Notes

1- Open a channel
2- Save a draft with attachment
3- Go to the channel list
4- The attachemnt should appear and not disappear from the draft preview

### ☑️ Contributor Checklist


laevandus (approved):
✅ Looks great. Verified it manually and the draft message with attachments did not get removed. Also tried quitting the app and relaunching it - all good.

martinmitrevski (approved):

PR #3640, opened 2025-04-08
Remove current user device automatically when logout

nuno-vieira:
### 🔗 Issue Links

https://linear.app/stream/issue/IOS-708/make-removal-of-the-device-as-part-of-logout-function

### 🎯 Goal

Remove the current user's device automatically when logging out. Multiple customers forget to do this, so we are adding it automatically by default. However, if a customer has already deleted the user device before calling logout, it won't call remove device again.

This can be disabled by setting `removeDevice: false` when calling logout.

### 📝 Summary

- Add `logout(removeDevice: Bool)`
- Remove current user device automatically when logout is called.

### 🧪 Manual Testing Notes

Smoke testing around logout/login flow.

### ☑️ Contributor Checklist

- [x] Changelog is updated with new localization keys


CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Add `ChatChannelController.deletePoll()` for deleting polls [#3632](https://github.com/GetStream/stream-chat-swift/pull/3632)
 - Add `ChatChannel.canSendPoll` capability [#3635](https://github.com/GetStream/stream-chat-swift/pull/3635)
 - Add `ChatChannel.canCastPollVote` capability [#3635](https://github.com/GetStream/stream-chat-swift/pull/3635)
+- Add `removeDevice: Bool` parameter to `ChatClient.logout` [#3640](https://github.com/GetStream/stream-chat-swift/pull/3640)
```
nuno-vieira, line 11:
```suggestion
- Add `removeDevice: Bool` parameter to `ChatClient.logout()` [#3640](https://github.com/GetStream/stream-chat-swift/pull/3640)
```

Sources/StreamChat/ChatClient.swift
```diff
@@ -483,15 +483,33 @@ public class ChatClient {
...
     }
 
     /// Disconnects the chat client from the chat servers and removes all the local data related.
-    public func logout(completion: @escaping () -> Void) {
-        authenticationRepository.logOutUser()
+    /// - Parameters:
+    ///  - removeDevice: If `true`, it removes the current device from the user's registered devices automatically.
+    ///  By default it is enabled.
+    public func logout(
+        removeDevice: Bool = true,
```
martinmitrevski, line 497: Should we go with `false` initially? Since it's a breaking change, and many customers would receive "device not exists errors" when they do their request.

Sources/StreamChat/ChatClient.swift
```diff
@@ -646,7 +664,7 @@ public class ChatClient {
 
 extension ChatClient: AuthenticationRepositoryDelegate {
     func logOutUser(completion: @escaping () -> Void) {
-        logout(completion: completion)
+        logout(removeDevice: false, completion: completion)
```
martinmitrevski, line 667: Why do we need it here?

nuno-vieira: When logging in with a different user, we previously did not remove the device. And doing this while there is a user transition, I observed some problems, so better to remain the same behaviour as before. Either way, if this is called it means there is a bad integration, since it means the customer did not logout before login with a different user.

Sources/StreamChat/ChatClient.swift
nuno-vieira: The removeDevice is called only if it was not done before. I've tested this, and with the old demo app code, this additional call is not performed because currentDevice has been deleted already. It is stated in the PR description:

> However, if a customer has already deleted the user device before calling logout, it won't call logout again.

martinmitrevski: ah cool, missed that

nuno-vieira: But if we want to minimize the risk as much as possible we could set this flag to `false`. But the problem is that this will keep increasing the support requests, as well as require updating the documentation that this flag needs to be set in order for the remove device to be executed. On Android, this is also done by default, so it would be weird if had this different from them. That is why I tried as much as possible to make this enabled by default, without breaking existing integrations

martinmitrevski (approved):

PR #3660, opened 2025-05-05
Swift 6: complete concurrency checking for StreamChatUI

laevandus:
> [!IMPORTANT]
> Merges to the main `swift-6` branch, not `develop`

### 🔗 Issue Links

Resolves [IOS-735](https://linear.app/stream/issue/IOS-735)

### 🎯 Goal

* Enable complete concurrency checking for StreamChatUI

### 📝 Summary

* Sets `SWIFT_STRICT_CONCURRENCY` to `complete`, but keeps the Swift language version to 5 due to the Xcode 15 requirement (language version will be changed in a separate PR when Xcode 15 support is dropped).
  * Completion blocks need to be `@Sendable` in most of the cases
  * Controller completions and delegates must ensure main thread (through custom `MainActor.ensureIsolated`) 
  * Controller delegates get `nonisolated` because any thread thread could call it and often MainActor guarded type implements these delegates (e.g. UIViewController subclass)
* Upgrade Nuke to version 12.8 (otherwise we can't compile with complete concurrency checking)
* Patch SwiftyGif for supporting complete concurrency checking

### 🛠 Implementation

LLC controllers support any queue for delegate callbacks and completion handlers. Therefore we need to ensure that these run on the main thread when updating UI code. This is where the custom `MainActor.ensureIsolated(_)` comes into the play. It just ensures that code triggered from these code paths are always on the main thread.

Public API which requires MainActor are annotated with `@preconcurrency @MainActor` which is the [backwards compatibility support](https://www.swift.org/migration/documentation/swift-6-concurrency-migration-guide/incrementaladoption#Backwards-Compatibility). In the next major version we would clean it up.

- [ ] How to handle Appearance (MainActor vs guarding it with queue/lock)

### 🎨 Showcase

| Before | After |
| ------ | ----- |
|  img   |  img  |

### 🧪 Manual Testing Notes

### ☑️ Contributor Checklist


Sources/StreamChatUI/Appearance+Formatters/MarkdownFormatter.swift
```diff
@@ -187,7 +187,7 @@ private extension UIFont {
 @available(iOS 15.0, *)
 private extension InlinePresentationIntent {
     /// An intent that represents bold with italic presentation.
-    static var extremelyStronglyEmphasized = InlinePresentationIntent(rawValue: 3)
+    static var extremelyStronglyEmphasized: InlinePresentationIntent { InlinePresentationIntent(rawValue: 3) }
```
laevandus, line 190: Xcode 15 failed to compile it without explicit type

Sources/StreamChatUI/Appearance+Formatters/ChannelNameFormatter.swift
```diff
@@ -15,7 +15,7 @@ open class DefaultChannelNameFormatter: ChannelNameFormatter {
     public init() {}
 
     /// Internal static property to add backwards compatibility to `Components.channelNamer`
-    internal static var channelNamer: (
+    nonisolated(unsafe) internal static var channelNamer: (
```
laevandus, line 18: TODO: Review if MainActor is more approriate here

Sources/StreamChatUI/Appearance.swift
```diff
@@ -39,5 +39,11 @@ public struct Appearance {
...
 
 public extension Appearance {
-    static var `default`: Appearance = .init()
+    static var `default`: Appearance {
+        get { queue.sync { _default } }
+        set { queue.sync { _default = newValue } }
+    }
+    
+    private static let queue = DispatchQueue(label: "io.getstream.appearance", target: .global())
+    nonisolated(unsafe) private static var _default: Appearance = .init()
```
laevandus, line 48: On the second look it actually does not feel like a good idea. Since this is used for UI code then it sounds like making the static property MainActor makes more sense here. Otherwise we do unnecessary context switches. Making it main actor will have consequences:
* app level setup code must be done on the main thread
* other code interacting with default must require main thread

laevandus: In general, options are:
a) opt-out with nonisolated(unsafe) which kind of loses the point of strict concurrency checking. SDK users can trigger crashes.
b) Internally uses main thread when accessing the `default` (which should be the 99% of cases when calling it)
```swift
    static var `default`: Appearance {
        get {
            MainActor.ensureIsolated { _default }
        }
        set {
            MainActor.ensureIsolated { _default = newValue }
        }
    }
```
and making `Appearance` `@unchecked Sendable` (lot of work to make it fully Sendable since this type captures so many other types, e.g formatters)
c) Make the `default` `@MainActor` and add MainActor requirement to all the other callsites

c) is the cleanest, b) is the hackiest, a) loses the point of strict concurrency checking since we opt-out

laevandus: c) actually does not require many changes. Only side-effect is that setup code for SDK users need to handle the main actor requirement.

Sources/StreamChatUI/ChatMessageList/ChatMessage/Decorators/ChatMessageHeaderDecoratorView.swift
```diff
@@ -51,7 +51,7 @@ public final class ChatChannelMessageHeaderDecoratorView: ChatMessageDecorationV
         container.axis = .vertical
         container.spacing = 4
 
-        [dateView, unreadCountView].forEach(container.addArrangedSubview)
+        [dateView, unreadCountView].forEach { container.addArrangedSubview($0) }
```
laevandus, line 54: Xcode 15 fails to compile such code when complete concurrency checking is enabled (stack was not useful) OK with Xcode 16.

Sources/StreamChatUI/ChatMessageList/SwipeToReplyGestureHandler.swift
```diff
@@ -60,10 +60,10 @@ open class SwipeToReplyGestureHandler {
 
     public init(
         listView: ChatMessageListView,
-        impactFeedbackGenerator: UIImpactFeedbackGenerator = .init(style: .medium)
+        impactFeedbackGenerator: UIImpactFeedbackGenerator? = nil
     ) {
         self.listView = listView
-        self.impactFeedbackGenerator = impactFeedbackGenerator
+        self.impactFeedbackGenerator = impactFeedbackGenerator ?? UIImpactFeedbackGenerator(style: .medium)
```
laevandus, line 66: Xcode 15 had compiler crashes and this was the only way I could get it compiling

Sources/StreamChatUI/Appearance.swift
laevandus: a) loses the point of concurrency checking
b) makes it easier to use default from places without main actor annotation (easier adoption), on the other hand, it does not look the nicest, but at least does not cause performance issues due to context switches
c) cleanest but harder to use for users since it forces MainActor
e) performance wise not good because this is used by UI code and should be almost always run on the main thread (that said, SDK users could have setup running on a background thread)

Sources/StreamChatUI/Appearance.swift
```diff
@@ -39,5 +39,16 @@ public struct Appearance {
...
+            MainActor.ensureIsolated { _default }
+        }
+        set {
+            MainActor.ensureIsolated { _default = newValue }
+        }
+    }
+    
+    // Shared instance is mutated only on the main thread without explicit
+    // main actor annotation for easier SDK setup.
+    nonisolated(unsafe) private static var _default: Appearance = .init()
```
laevandus, line 53: At first I was thinking about making it `@MainActor`, but this is used in multiple init methods meaning all of these must be `@MainActor` as well. Moreover, it might make the SDK initialization more complex as well since SDK users need to make sure `Appearance` is changed only from the `MainActor`.
a) we just make `default` nonisolated(unsafe)
b) we do what we have above which internally forcing main actor (for making sure SDK users do not corrupt the default by changing it from background threads)

nuno-vieira (commented):
Most of the changes are related to `@MainActor`. So lets resume the review after the decision of the tech discussion.

laevandus:
I am going to make breaking changes what we discussed separately, take all the feedback from this PR, and later create a brand new PR.

PR #3661, opened 2025-05-05
Swift 6: complete concurrency checking (LLC and UIKit)

laevandus:
> [!IMPORTANT]
> StreamChatUI changes are in #3660 and will be merged to here. This PR can be reviewed already.

### 🔗 Issue Links

Resolves [IOS-735](https://linear.app/stream/issue/IOS-735)

### 🎯 Goal

* Set Swift version to 6 and implement complete concurrency checking

### 📝 Summary

* `Sendable` conformance to types
* `@Sendable` to completion handlers
* Many thread-safety related changes to `@unchecked Sendable` types (internal classes have manual handling and found many which did not guard that state properly)
* Fixes warnings in the demo app (adding `@Sendable` to completion handlers creates warnings at callsites)
* Backported AllocatedUnfairLock which is needed for static concurrency safe properties (our Atomic property wrapper can't be used)

### 🧪 Manual Testing Notes

Manual regression testing round when UIKit changes are merged into this branch.

### ☑️ Contributor Checklist

### DIFF
--- a/Tests/StreamChatTests/APIClient/ChatRemoteNotificationHandler_Tests.swift
+++ b/Tests/StreamChatTests/APIClient/ChatRemoteNotificationHandler_Tests.swift
@@ -262,7 +262,7 @@ final class ChatRemoteNotificationHandler_Tests: XCTestCase {
             "message.updated": .messageUpdated
         ]
 
-        var assertions: [Bool] = []
+        nonisolated(unsafe) var assertions: [Bool] = []
         expectation.expectedFulfillmentCount = notificationTypes.count
 
         for notificationType in notificationTypes {
--- a/Tests/StreamChatTests/Controllers/MessageReminderListController/MessageReminderListController+Combine_Tests.swift
+++ b/Tests/StreamChatTests/Controllers/MessageReminderListController/MessageReminderListController+Combine_Tests.swift
@@ -42,7 +42,7 @@ final class MessageReminderListController_Combine_Tests: iOS13TestCase {
             .store(in: &cancellables)
 
         // Keep only the weak reference to the controller. The existing publisher should keep it alive.
-        weak var controller: MessageReminderListController? = reminderListController
+        nonisolated(unsafe) weak var controller: MessageReminderListController? = reminderListController
         reminderListController = nil
 
         controller?.delegateCallback { $0.controller(controller!, didChangeState: .remoteDataFetched) }


.github/workflows/smoke-checks.yml
```diff
@@ -68,7 +68,8 @@ jobs:
   build-old-xcode:
     name: Build LLC + UI (Xcode 15)
     runs-on: macos-15
-    if: ${{ github.event.inputs.record_snapshots != 'true' }}
+    # if: ${{ github.event.inputs.record_snapshots != 'true' }}
+    if: false # disable Xcode 15 builds
```
laevandus, line 72: Disabled Xcode 15 builds

Sources/StreamChat/Utils/BoxedAny.swift
```diff
@@ -0,0 +1,29 @@
+//
+// Copyright © 2025 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+/// Erase type for structs which recursively contain themselves.
```
laevandus, line 7: It is easier to use a type instead of a closure when dealing with Sendable.

Sources/StreamChat/APIClient/Endpoints/Payloads/RawJSON.swift
```diff
@@ -7,15 +7,15 @@ import Foundation
...
-public indirect enum RawJSON: Codable, Hashable {
+public indirect enum RawJSON: Codable, Hashable, Sendable {
     case number(Double)
     case string(String)
     case bool(Bool)
     case dictionary([String: RawJSON])
     case array([RawJSON])
     case `nil`
 
-    static let double = number
```
laevandus, line 18: Review if setting Swift language to 6 allows removing the nonisolated(unsafe) change.

Sources/StreamChat/Config/StreamRuntimeCheck.swift
```diff
@@ -30,5 +40,14 @@ public enum StreamRuntimeCheck {
...
+    public static var _isDatabasePrefetchingEnabled: Bool {
+        get { queue.sync { __isDatabasePrefetchingEnabled } }
+        set { queue.async { __isDatabasePrefetchingEnabled = newValue } }
+    }
+
+    nonisolated(unsafe) private static var __isDatabasePrefetchingEnabled = false
+    
+    // MARK: -
+    
+    private static let queue = DispatchQueue(label: "io.getstream.stream-runtime-check", target: .global())
```
laevandus, line 52: Alternative is to skip using queue here and making everything nonisolated(unsafe), especially because these are internal runtime checkes.

Sources/StreamChat/Repositories/PollsRepository.swift
```diff
@@ -135,60 +135,64 @@ class PollsRepository {
...
-        completion: ((Error?) -> Void)? = nil
+        completion: (@Sendable(Error?) -> Void)? = nil
     ) {
-        var exists = false
-        var answerText: String?
-        var optionId: String?
-        var userId: UserId?
-        var filterHash: String?
-        database.write { session in
+        struct WriteResponse {
```
martinmitrevski, line 147: why do we need a struct here?

Sources/StreamChat/Utils/AllocatedUnfairLock.swift
```diff
@@ -0,0 +1,39 @@
+//
+// Copyright © 2025 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+import os
+
+@available(iOS, introduced: 13.0, deprecated: 16.0, message: "Use OSAllocatedUnfairLock instead")
+final class AllocatedUnfairLock<State>: @unchecked Sendable {
```
martinmitrevski, line 9: where do we use an unfair lock?

Sources/StreamChat/Utils/BoxedAny.swift
martinmitrevski: yes, this is a nice approach

Sources/StreamChat/Utils/MainQueue+Synchronous.swift
```diff
@@ -4,19 +4,23 @@
...
-    static func performSynchronouslyOnMainQueue(_ action: () throws -> Void) rethrows {
+    /// Used for ensuring we are on the main thread when compiler can't know it. For example,
+    /// controller completion handlers by default are called from main thread, but one can
+    /// configure controller to use background thread for completions instead.
+    static func ensureIsolated<T>(_ action: @MainActor @Sendable() throws -> T) rethrows -> T where T: Sendable {
         if Thread.current.isMainThread {
-            try action()
-        } else {
-            try DispatchQueue.main.sync {
+            return try MainActor.assumeIsolated {
```
martinmitrevski, line 15: why is this needed?

Sources/StreamChat/Utils/StreamCollection.swift
```diff
@@ -9,30 +9,27 @@ import Foundation
...
-    private let _startIndex: () -> Index
-
+    private let baseCollection: [Element]
+    
     /// Creates an instance of the collection using the base collection as the data source.
     public init<BaseCollection>(_ baseCollection: BaseCollection) where BaseCollection: RandomAccessCollection, BaseCollection.Element == Element, BaseCollection.Index == Index {
-        _endIndex = { baseCollection.endIndex }
-        _position = { baseCollection[$0] }
-        _startIndex = { baseCollection.startIndex }
+        self.baseCollection = Array(baseCollection)
```
martinmitrevski, line 16: why do we change this?

Sources/StreamChat/WebSocketClient/WebSocketPingController.swift
```diff
@@ -78,23 +86,30 @@ class WebSocketPingController {
...
         }
     }
 
     private func schedulePongTimeoutTimer() {
         cancelPongTimeoutTimer()
         // Start pong timeout timer.
-        pongTimeoutTimer = timerType.schedule(timeInterval: Self.pongTimeoutTimeInterval, queue: timerQueue) { [weak self] in
-            log.info("WebSocket Pong timeout. Reconnect")
-            self?.delegate?.disconnectOnNoPongReceived()
+        queue.sync {
```
martinmitrevski, line 100: why is this needed?

Sources/StreamChat/APIClient/CDNClient/CDNClient.swift
```diff
@@ -60,14 +60,15 @@ public extension CDNClient {
...
 /// Default implementation of CDNClient that uses Stream CDN
-class StreamCDNClient: CDNClient {
+final class StreamCDNClient: CDNClient, Sendable {
     static var maxAttachmentSize: Int64 { 100 * 1024 * 1024 }
 
     private let decoder: RequestDecoder
     private let encoder: RequestEncoder
     private let session: URLSession
     /// Keeps track of uploading tasks progress
-    @Atomic private var taskProgressObservers: [Int: NSKeyValueObservation] = [:]
```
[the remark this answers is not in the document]
laevandus: `Stored property '_taskProgressObservers' of 'Sendable'-conforming class 'StreamCDNClient' is mutable`

Alternative is to use backwards compatible `AllocatedUnfairLock` (I added it for mutable static properties).
`@Atomic` is OK in many places because the class/type is `@unchecked Sendable`. In this particular case it is directly `Sendable`.

Property wrappers can't ensure concurrency safeness (there is a long thread about it on Swift forums)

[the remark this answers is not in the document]
laevandus: It does look cleaner:
```swift
     /// Keeps track of uploading tasks progress
-    private nonisolated(unsafe) var _taskProgressObservers: [Int: NSKeyValueObservation] = [:]
-    private let queue = DispatchQueue(label: "io.getstream.stream-cdn-client", target: .global())
+    private let taskProgressObservers = AllocatedUnfairLock([Int: NSKeyValueObservation]())

     init(
         encoder: RequestEncoder,
@@ -149,13 +148,13 @@ final class StreamCDNClient: CDNClient, Sendable {

             if let progressListener = progress {
                 let taskID = task.taskIdentifier
-                queue.async {
-                    self._taskProgressObservers[taskID] = task.progress.observe(\.fractionCompleted) { [weak self] progress, _ in
+                taskProgressObservers.withLock { observers in
+                    observers[taskID] = task.progress.observe(\.fractionCompleted) { [weak self] progress, _ in
                         progressListener(progress.fractionCompleted)
                         if progress.isFinished || progress.isCancelled {
-                            self?.queue.async { [weak self] in
-                                self?._taskProgressObservers[taskID]?.invalidate()
-                                self?._taskProgressObservers[taskID] = nil
+                            self?.taskProgressObservers.withLock { observers in
+                                observers[taskID]?.invalidate()
+                                observers[taskID] = nil
                             }
                         }
                     }
```

WDYT, let's use the snippet above?

Sources/StreamChat/Config/StreamRuntimeCheck.swift
laevandus: I'll revert, makes sense

Sources/StreamChat/Controllers/EventsController/EventsController.swift
```diff
@@ -50,6 +50,8 @@ public class EventsController: Controller, DelegateCallable {
         get { multicastDelegate.mainDelegate }
         set { multicastDelegate.set(mainDelegate: newValue) }
     }
+    
+    private let queue = DispatchQueue(label: "io.getstream.events-controller", target: .global())
```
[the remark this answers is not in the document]
laevandus: Unnused! Removing, thank you

Sources/StreamChat/Database/DatabaseContainer.swift
```diff
@@ -223,23 +226,34 @@ class DatabaseContainer: NSPersistentContainer, @unchecked Sendable {
...
+            } else if let result {
                 completion(.success(result))
             } else {
-                completion(.failure(error ?? ClientError.Unknown()))
+                completion(.failure(ClientError.Unknown()))
+            }
+        }
+    }
+    
+    func writeConverting<T>(_ actions: @escaping @Sendable(DatabaseSession) throws -> T) async throws -> T where T: Sendable {
```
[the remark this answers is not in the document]
laevandus: Unused, I found a different way and forgot it here

Sources/StreamChat/Models/Attachments/AttachmentTypes.swift
```diff
@@ -153,7 +153,7 @@ public struct AttachmentFile: Codable, Hashable {
     /// A mime type.
     public let mimeType: String?
     /// A file size formatter.
-    public static let sizeFormatter = ByteCountFormatter()
+    public static var sizeFormatter: ByteCountFormatter { ByteCountFormatter() }
```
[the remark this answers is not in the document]
laevandus: `Static property 'sizeFormatter' is not concurrency-safe because non-'Sendable' type 'ByteCountFormatter' may have shared mutable state`
I'll see if there is something I can do here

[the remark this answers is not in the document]
laevandus: I'll just make it unsafe

Sources/StreamChat/Repositories/PollsRepository.swift
laevandus: I created a struct which is passed into the completion handler with the write converting method. Then these variables are not shared between different closures. But that said, we can see it is actually safe anyway and makes things to complex. I'll add some nonisolated(unsafe)

Sources/StreamChat/Utils/StreamCollection.swift
laevandus: `Stored property '_endIndex' of 'Sendable'-conforming generic struct 'StreamCollection' has non-sendable type '() -> StreamCollection<Element>.Index' (aka '() -> Int')`

At first I tried to make closures `@Sendable`, which requires `BaseCollection` to be `Sendable`. All good until I saw that StreamCollection is in some cases initialized with a list of `MessageDTO`. `MessageDTO` can't be `Sendable`, therefore I can't do this. 

Fortunately we want to remove `StreamCollection` in v5 anyway.

laevandus: Don't have a good solution here.

Sources/StreamChat/WebSocketClient/WebSocketPingController.swift
laevandus: Problem here is that connectionDidChange is called from thread X and timer calls `sendPing` from thread Y which in turn will also access the `_pingTimerControl` (possible threading issue).

AllocatedUnfairLock would be more readable here.

Since we haven't heard about crashes with ping controller, I am open to keep it as is as well and not do these changes.

Sources/StreamChat/Utils/AllocatedUnfairLock.swift
laevandus: Only a couple of places in TestTools. If we prefer not to add this type, we can use some `nonisolated(unsafe)` in test tools (I am OK to do that).

Screenshot 2025-05-23 at 14 46 49

.github/workflows/smoke-checks.yml
```diff
@@ -67,8 +67,9 @@ jobs:
 
   build-old-xcode:
     name: Build LLC + UI (Xcode 15)
-    runs-on: macos-14
-    if: ${{ github.event.inputs.record_snapshots != 'true' }}
+    runs-on: macos-15
+    # if: ${{ github.event.inputs.record_snapshots != 'true' }}
```
martinmitrevski, line 71: what should we do here?

DemoApp/Screens/DemoDraftMessageListVC.swift
```diff
@@ -105,7 +107,9 @@ class DemoDraftMessageListVC: UIViewController, ThemeProvider {
 
         isPaginatingDrafts = true
         currentUserController.loadMoreDraftMessages { [weak self] _ in
-            self?.isPaginatingDrafts = false
+            Task { @MainActor in
```
martinmitrevski, line 110: we should probably have a Swift 6 migration guide - customers would probably need to do the same things in their apps.

Package.swift
```diff
@@ -51,7 +51,10 @@ let package = Package(
             dependencies: ["StreamChat"],
             path: "TestTools/StreamChatTestMockServer",
             exclude: ["Info.plist"],
-            resources: [.process("Fixtures")]
+            resources: [.process("Fixtures")],
+            swiftSettings: [
+                .swiftLanguageMode(.v5)
```
martinmitrevski, line 56: why do we set this?

Sources/StreamChat/APIClient/CDNClient/CDNClient.swift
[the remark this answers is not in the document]
martinmitrevski: yes, I think so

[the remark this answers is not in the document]
martinmitrevski: btw, we have this now in Stream Core, so you should align it, because that one will replace this.

Sources/StreamChat/Audio/AudioPlayer/AudioPlaying.swift
```diff
@@ -131,15 +131,19 @@ open class StreamAudioPlayer: AudioPlaying, AppStateObserverDelegate {
     open func play() {
         do {
             try audioSessionConfigurator.activatePlaybackSession()
-            player.play()
+            MainActor.ensureIsolated {
+                player.play()
```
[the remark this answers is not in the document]
martinmitrevski: why do we need this `StreamConcurrency.onMain`?

Sources/StreamChat/Database/DatabaseContainer.swift
```diff
@@ -2,7 +2,7 @@
 // Copyright © 2025 Stream.io Inc. All rights reserved.
 //
 
-import CoreData
+@preconcurrency import CoreData
```
martinmitrevski, line 5: why `@preconcurrency`?

Sources/StreamChat/Database/DatabaseContainer.swift
```diff
@@ -70,7 +70,8 @@ class DatabaseContainer: NSPersistentContainer, @unchecked Sendable {
 
     let chatClientConfig: ChatClientConfig
 
-    static var cachedModels = [String: NSManagedObjectModel]()
+    private static let cachedModelsQueue = DispatchQueue(label: "io.getstream.database-container", target: .global())
```
martinmitrevski, line 73: why static?

Sources/StreamChat/Utils/AllocatedUnfairLock.swift
martinmitrevski: given that we use it until iOS 16, that might not be a bad idea (to remove it)

Sources/StreamChat/Utils/StateBuilder.swift
```diff
@@ -5,14 +5,20 @@
 import Foundation
 
 /// A builder for objects requiring @MainActor.
-struct StateBuilder<State> {
-    private let builder: (@MainActor() -> State)
+struct StateBuilder<State: Sendable>: Sendable {
+    private var builder: ((@Sendable @MainActor() -> State))?
```
martinmitrevski, line 9: why is it optional?

Sources/StreamChat/Utils/StateBuilder.swift
```diff
@@ -5,14 +5,20 @@
...
     }
     
-    @MainActor func build() -> State {
-        builder()
+    @MainActor mutating func reuseOrBuild() -> State {
+        if let _state { return _state }
+        let state = builder!()
+        _state = state
+        // Release captured values in the closure
+        builder = nil
```
martinmitrevski, line 21: related to above - why do we need to nil it out?

Sources/StreamChat/WebSocketClient/BackgroundTaskScheduler.swift
```diff
@@ -36,10 +36,12 @@ class IOSBackgroundTaskScheduler: BackgroundTaskScheduler {
 
     var isAppActive: Bool {
         if Thread.isMainThread {
-            return app?.applicationState == .active
+            return MainActor.assumeIsolated {
```
martinmitrevski, line 39: why do we still need this here?

Sources/StreamChat/WebSocketClient/WebSocketPingController.swift
martinmitrevski: is this some warning/error in Swift 6?

.github/workflows/smoke-checks.yml
nuno-vieira: Probably better to disable Xcode 15 for now, and once we add Xcode 26, we then put here Xcode 16 support. WDYT? @testableapple

DemoApp/Screens/DemoDraftMessageListVC.swift
nuno-vieira: Isn't there a way to mark the controllers as MainActor by default? 

This is cleary a breaking change. IMO for supporting Swift 6, we should do a major bump of the SDK. Most libraries do this. It is an opportunity to do a light major version bump. We can remove deprecations and change minor things. For example, here we can remove the callback stuff, and just always dispatch the controller completion blocks to the main thread. I don't think there was every any customer changing the callback thread.

DemoApp/Screens/MembersViewController.swift
```diff
@@ -37,7 +37,9 @@ class MembersViewController: UITableViewController, ChatChannelMemberListControl
 
     private func synchronizeAndUpdateData() {
         membersController.synchronize { [weak self] _ in
-            self?.updateData()
+            Task { @MainActor in
+                self?.updateData()
+            }
```
nuno-vieira, line 42: Yeah, we can't put this everywhere... if customers require doing this, this a breaking change. Will this produce a warning or an error?

Package.swift
nuno-vieira: Shouldn't this be v6?

Sources/StreamChat/Audio/AssetPropertyLoading.swift
```diff
@@ -105,16 +105,19 @@ public struct StreamAssetPropertyLoader: AssetPropertyLoading {
...
         of asset: Asset,
-        completion: @escaping (Result<Asset, AssetPropertyLoadingCompositeError>) -> Void
+        completion: @escaping @Sendable(Result<Asset, AssetPropertyLoadingCompositeError>) -> Void
     ) {
         // it's worth noting here that according to the documentation, the completion
         // handler will be invoked only once, regardless of the number of
         // properties we are loading.
         // https://developer.apple.com/documentation/avfoundation/avasynchronouskeyvalueloading/1387321-loadvaluesasynchronously
+        
+        // On iOS 15 we should switch to async load methods
```
nuno-vieira, line 115: Can we add a TODO here?

Sources/StreamChat/ChatClient.swift
```diff
@@ -25,7 +25,7 @@ public class ChatClient {
 
     /// The app configuration settings. It is automatically fetched when `connectUser` is called.
     /// Can be manually refetched by calling `loadAppSettings()`.
-    public private(set) var appSettings: AppSettings?
+    @Atomic public private(set) var appSettings: AppSettings?
```
nuno-vieira, line 28: Why are we adding `@Atomic` in properties that were not before, and removing existing `@Atomic` properties? Can you clarify this? 🙏

Sources/StreamChat/ChatClient.swift
```diff
@@ -46,17 +46,22 @@ public class ChatClient {
...
     /// The notification center used to send and receive notifications about incoming events.
     private(set) var eventNotificationCenter: EventNotificationCenter
 
     /// The registry that contains all the attachment payloads associated with their attachment types.
     /// For the meantime this is a static property to avoid breaking changes. On v5, this can be changed.
-    private(set) static var attachmentTypesRegistry: [AttachmentType: AttachmentPayload.Type] = [
+    private(set) static var attachmentTypesRegistry: [AttachmentType: AttachmentPayload.Type] {
+        get { Self.queue.sync { _attachmentTypesRegistry } }
+        set { Self.queue.sync { _attachmentTypesRegistry = newValue } }
+    }
```
nuno-vieira, line 62: Why don't we follow this strategy for the other properties as well to keep conistency?

Sources/StreamChat/ChatClient.swift
```diff
@@ -46,17 +46,22 @@ public class ChatClient {
...
+
+    nonisolated(unsafe) private(set) static var _attachmentTypesRegistry: [AttachmentType: AttachmentPayload.Type] = [
```
nuno-vieira, line 64: Why didn't we add thread-safety to this one?

Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift
```diff
@@ -105,7 +106,7 @@ public class ChatChannelListController: DataController, DelegateCallable, DataSt
         return observer
     }()
 
-    var _basePublishers: Any?
+    @Atomic private var _basePublishers: Any?
```
nuno-vieira, line 109: Any reason why we are adding Atomic here? And not using a queue?

Sources/StreamChat/Controllers/ConnectionController/ConnectionController.swift
```diff
@@ -17,10 +17,15 @@ public extension ChatClient {
...
-    var _basePublishers: Any?
+public class ChatConnectionController: Controller, DelegateCallable, DataStoreProvider, @unchecked Sendable {
+    public var callbackQueue: DispatchQueue {
+        get { queue.sync { _callbackQueue } }
+        set { queue.sync { _callbackQueue = newValue } }
+    }
+    
+    private let queue = DispatchQueue(label: "io.getstream.chat-connection-controller", target: .global(qos: .userInteractive))
+    private var _callbackQueue: DispatchQueue = .main
+    @Atomic private var _basePublishers: Any?
```
nuno-vieira, line 28: If we opt for a major version, we can make the Controllers as MainActors. This will simplify the code on our side, as well as the customer's side:

```suggestion
@MainActor
public class ChatConnectionController: Controller, DelegateCallable, DataStoreProvider, @unchecked Sendable {
    private var basePublishers: Any?
```

Sources/StreamChat/Controllers/CurrentUserController/CurrentUserController.swift
```diff
@@ -104,50 +104,63 @@ public class CurrentChatUserController: DataController, DelegateCallable, DataSt
...
     /// The worker used to update messages. Used for updating live locations of messages.
     private lazy var messageUpdater = createMessageUpdater()
 
     // MARK: - Drafts Properties
 
     /// The query used for fetching the draft messages.
-    private var draftListQuery = DraftListQuery()
+    private var draftListQuery: DraftListQuery {
+        get { queue.sync { _draftListQuery } }
+        set { queue.sync { _draftListQuery = newValue } }
```
nuno-vieira, line 133: Can this be async? I'm always a bit concerned of using sync when it is not needed, to avoid deadlocks. 
```suggestion
        set { queue.async { _draftListQuery = newValue } }
```

Sources/StreamChat/Controllers/ThreadListController/ThreadListController.swift
```diff
@@ -40,14 +45,15 @@ public class ChatThreadListController: DataController, DelegateCallable, DataSto
...
-            client.databaseContainer,
-            client.apiClient
-        )
+    private let threadsRepository: ThreadsRepository
 
     /// A Boolean value that returns whether pagination is finished.
-    public private(set) var hasLoadedAllThreads: Bool = false
+    public private(set) var hasLoadedAllThreads: Bool {
+        get { queue.sync { _hasLoadedAllThreads } }
+        set { queue.sync { _hasLoadedAllThreads = newValue } }
```
nuno-vieira, line 53: Can we put this async?
```suggestion
        set { queue.async { _hasLoadedAllThreads = newValue } }
```

Sources/StreamChat/Controllers/UserController/UserController.swift
```diff
@@ -46,23 +46,45 @@ public class ChatUserController: DataController, DelegateCallable, DataStoreProv
...
+    private var userUpdater: UserUpdater {
+        queue.sync {
+            if let _userUpdater {
+                return _userUpdater
+            }
+            let updater = createUserUpdater()
+            _userUpdater = updater
+            return updater
+        }
+    }
```
nuno-vieira, line 58: Can we move this to the initialiser?

Sources/StreamChat/Repositories/SyncOperations.swift
```diff
@@ -5,12 +5,12 @@
...
+final class SyncContext: @unchecked Sendable {
     let lastSyncAt: Date
-    var localChannelIds: Set<ChannelId> = Set()
-    var synchedChannelIds: Set<ChannelId> = Set()
-    var watchedAndSynchedChannelIds: Set<ChannelId> = Set()
-    var unwantedChannelIds: Set<ChannelId> = Set()
+    @Atomic var localChannelIds: Set<ChannelId> = Set()
+    @Atomic var synchedChannelIds: Set<ChannelId> = Set()
+    @Atomic var watchedAndSynchedChannelIds: Set<ChannelId> = Set()
+    @Atomic var unwantedChannelIds: Set<ChannelId> = Set()
```
nuno-vieira, line 13: Why is there no `mutate` additional changes for this? When using @Atomic with collections, we should be using `mutate` to change the collection

Sources/StreamChat/StateLayer/MessageSearch.swift
```diff
@@ -5,10 +5,10 @@
 import Foundation
 
 /// An object which represents a list of ``ChatMessage`` for the specified search query.
-public class MessageSearch {
+public class MessageSearch: @unchecked Sendable {
```
nuno-vieira, line 8: Why is this one unchecked? Where it seems the other ones are all compile-safe Sendable?

Sources/StreamChat/Utils/Atomic.swift
```diff
@@ -23,7 +23,7 @@ import Foundation
 /// itself from multiple threads can cause a crash.
 
 @propertyWrapper
-public class Atomic<T> {
+public class Atomic<T>: @unchecked Sendable {
```
nuno-vieira, line 26: According to our v5 goals, we should be using less and less this Atomic wrapper. It is easy to create deadlocks, and it is also easy to forget to use `mutate`. 

It also creates inconsistency in our code base. Sometimes we use Atomic, and other times we use regular queues. Ideally, new code should not be using this

Sources/StreamChat/Utils/Logger/Logger.swift
```diff
@@ -41,170 +41,253 @@ public struct LogSubsystem: OptionSet {
...
     /// Log formatters to be applied in order before logs are outputted. Defaults to empty (no formatters).
     /// Please see `LogFormatter` for more info.
-    public static var formatters = [LogFormatter]() {
-        didSet {
+    public static var formatters: [LogFormatter] {
+        get {
+            queue.sync { _storage.formatters }
+        }
+        set {
+            queue.sync { _storage.formatters = newValue }
```
nuno-vieira, line 83: Why do we sync on this one instead of async?

Sources/StreamChat/Utils/StateBuilder.swift
nuno-vieira: Yes why it is reuseOrBuild now? Can we add a comment explaining this?

Sources/StreamChat/WebSocketClient/WebSocketPingController.swift
nuno-vieira: I would prefer to not do any changes. We already have to do a lot of Swift 6 changes. The ones which are not 100% needed, we should not do them. To make reviewing easier, and also to have less risks when making the final release.

Sources/StreamChat/Workers/EventNotificationCenter.swift
```diff
@@ -7,30 +7,45 @@ import Foundation
 
 /// The type is designed to pre-process some incoming `Event` via middlewares before being published
 class EventNotificationCenter: NotificationCenter, @unchecked Sendable {
-    private(set) var middlewares: [EventMiddleware] = []
+    private let queue = DispatchQueue(label: "io.getstream.event-notification-center-sync", target: .global())
+    private(set) var middlewares: [EventMiddleware] {
+        get { queue.sync { _middlewares } }
+        set { queue.sync { _middlewares = newValue } }
```
nuno-vieira, line 13: why not async? We should try to use async whenever possible

Sources/StreamChat/Workers/EventNotificationCenter.swift
```diff
@@ -7,30 +7,45 @@ import Foundation
...
     /// The database used when evaluating middlewares.
     let database: DatabaseContainer
 
     var eventPostingQueue = DispatchQueue(label: "io.getstream.event-notification-center")
 
     // Contains the ids of the new messages that are going to be added during the ongoing process
-    private(set) var newMessageIds: Set<MessageId> = Set()
+    private(set) var newMessageIds: Set<MessageId> {
+        get { queue.sync { _newMessageIds } }
+        set { queue.sync { _newMessageIds = newValue } }
```
nuno-vieira, line 26: same here

Sources/StreamChat/Workers/EventNotificationCenter.swift
```diff
@@ -7,30 +7,45 @@ import Foundation
...
     init(database: DatabaseContainer) {
         self.database = database
         super.init()
     }
 
     func add(middlewares: [EventMiddleware]) {
-        self.middlewares.append(contentsOf: middlewares)
+        queue.sync {
+            _middlewares.append(contentsOf: middlewares)
+        }
```
nuno-vieira, line 39: same here

.github/workflows/smoke-checks.yml
testableapple: isn't it `macos-15`?

DemoApp/Screens/MembersViewController.swift
laevandus: Warning if they use swift 5, error when swift 6

DemoApp/Screens/DemoDraftMessageListVC.swift
laevandus: Yeah, controller completions could be main actor if we would not have that callback queue property. At the moment there is no good way.

Sources/StreamChat/StateLayer/MessageSearch.swift
laevandus: Class must be final for being Sendable, but we subclass these in tests so it can't be final. Which is sad.

Sources/StreamChat/Utils/Logger/Logger.swift
laevandus: Need to look it up, but asyn required sendability so more types need to be annotated.

Sources/StreamChat/StateLayer/MessageSearch.swift
nuno-vieira: I see... 😞

Sources/StreamChat/Utils/Logger/Logger.swift
nuno-vieira: I see..

laevandus:
I am going to make breaking changes what we discussed separately, take all the feedback from this PR, and later create a brand new PR.

PR #3671, opened 2025-05-19
Add `SendMessageInterceptor` to intercept send message request

nuno-vieira:
### 🔗 Issue Links

IOS-857

### 🎯 Goal

Add a way to intercept our send message request

### 📝 Summary

- Add `SendMessageInterceptor`
- Add `SendMessageInterceptorFactory`
- Add `ChatMessage.changing()`
- Improve documentation of `ChatMessage.replacing()`

#### `SendMessageInterceptor`
A new protocol has been introduced, `SendMessageInterceptor`. This will only be used if the customer provides an interceptor. If no interceptor is provided, nothing is changed.

The `SendMessageInterceptor` is not exposed directly in the `ChatClientConfig`. Instead,  it can be injected through a `SendMessageInterceptorFactory`. The factory gives us more flexibility, especially to make it possible to inject the `ChatClient`, which can be useful in case a customer wants to use the interceptor but still wants to use our internal APIClient to make the request.  I created an example on how, in the future, we can provide this to customers [here](https://github.com/GetStream/stream-chat-swift/pull/3671/commits/3985f12566b27b3a70ed563ba76fb34d5540a670). For now, I reverted it, but if any customer needs to use our APIClient, it will be easy to provide it. It can be useful for customers who might want to intercept our message requests, but only for a certain type of messages, or simply want to override some data.

#### `ChatMessage.changing()` vs `ChatMessage.replacing()`
I added documentation to `ChatMessage.replacing()` since it was lacking, and it could be misleading on how to use it. The `replacing()` function is useful when overriding data, especially if the customer wants to delete some of it. For example, providing the `state` with `nil` will make the message state as published. Since this behaviour might not be what some customers need, a `changing()` function has been introduced for the case where customers just want to change some data, but the rest should be left unchanged (It is different from `replacing()` because in this case it is not possible to erase data, since `nil` will fallback to the original data). 

### 🧪 Manual Testing Notes

N/A. Covered by unit tests.

### ☑️ Contributor Checklist


martinmitrevski (approved):
This is great, love how you made it as future proof as it can be!

Sources/StreamChat/Repositories/MessageRepository.swift
```diff
@@ -49,43 +57,46 @@ class MessageRepository {
...
             // Change the message state to `.sending` and the proceed with the actual sending
-            self?.database.write({
-                let messageDTO = $0.message(id: messageId)
-                messageDTO?.localMessageState = .sending
-            }, completion: { error in
-                if let error = error {
-                    log.error("Error changing localMessageState message with id \(messageId) to `sending`: \(error)")
-                    self?.markMessageAsFailedToSend(id: messageId) {
-                        completion(.failure(.failedToSendMessage(error)))
+            self?.database.write(
```
laevandus, line 63: Alternative
There is also `write(converting:` which allows returning a type within the action closure and uses Result type in the completion handler.

Sources/StreamChat/Repositories/MessageRepository.swift
```diff
@@ -158,6 +169,57 @@ class MessageRepository {
...
+        completion: @escaping (Result<ChatMessage, MessageRepositoryError>) -> Void
+    ) {
+        switch result {
+        case let .success(response):
+            let message = response.message
+            database.write { session in
+                guard let messageDTO = session.message(id: message.id) else { return }
+                if message.localState == nil {
+                    messageDTO.markMessageAsSent()
+                }
```
laevandus, line 211: Is this right that it is being compared to nil (since the state is set to `.sending` before)?

nuno-vieira: Yes, maybe I can add a comment here, the goal is that if a customer returns `localState == nil` it will mark it has Sent. Otherwise, it will wait for the `message.new` event to mark it as read. For most customers they will not change this. But in case a customer wants to treat it has published after the sendMessage response, they can do it

Sources/StreamChat/Repositories/MessageRepository.swift
nuno-vieira: Updated 👍

Sources/StreamChat/Repositories/MessageRepository.swift
nuno-vieira: Added the comment 👍

laevandus (approved):

PR #3674, opened 2025-05-22
Show message with restricted visiblity in the message list on the `message.updated` web-socket event

laevandus:
### 🔗 Issue Links

Resolves: [IOS-874](https://linear.app/stream/issue/IOS-874)

### 🎯 Goal

Fix publishing `MessageUpdatedEvent` if restricted visibility changes and automatically adding the message to the message list

### 📝 Summary

* Fix publishing `MessageUpdatedEvent` for messages with restricted visibility
* Show message with restricted visibility in the message list on the `message.updated` web-socket event

### 🛠 Implementation

SDK users would like to support a flow where:
1. Message is first sent to the current user only
2. The current user interacts with the message and this triggers `message.updated` with some changes including adding more users to the `restricted_visibility` field
Current issue is that although we receive the `message.updated` WS event, it does not get published by the SDK because the message payload is not saved to the local DB.

This is because:

```swift
func toDomainEvent(session: DatabaseSession) -> Event? {
        guard
            let userDTO = session.user(id: user.id),
            let messageDTO = session.message(id: message.id), <-- otherwise this fails, and event is not dispatched
            let channelDTO = session.channel(cid: cid)
        else { return nil }

        return try? MessageUpdatedEvent(
            user: userDTO.asModel(),
            channel: channelDTO.asModel(),
            message: messageDTO.asModel(),
            createdAt: createdAt
        )
    }
```

This PR will enable publishing the event and also adding the message to the message list.

### 🎨 Showcase

img

### ☑️ Contributor Checklist

## Summary by CodeRabbit

- **New Features**
  - Added support for restricting message visibility to specific users when sending or editing messages.
  - Updated message update functionality to allow specifying which users can view a message.
  - Added a new action in the demo app to send and update messages with restricted visibility.

- **Bug Fixes**
  - Fixed issues where message update events for messages with restricted visibility were not properly published or displayed in the message list.

- **Tests**
  - Added and updated tests to verify correct handling of messages with restricted visibility.


DemoApp/StreamChat/Components/DemoChatChannelVC.swift
```diff
@@ -89,4 +89,14 @@ final class DemoChatChannelVC: ChatChannelVC, UIGestureRecognizerDelegate {
...
+    
+    // MARK: - EventsControllerDelegate
+
+    override func eventsController(_ controller: EventsController, didReceiveEvent event: Event) {
+        super.eventsController(controller, didReceiveEvent: event)
+        
+        if let updatedEvent = event as? MessageUpdatedEvent {
+            channelController.synchronize()
+        }
+    }
```
laevandus, line 101: Remove this later. Just to demonstrate how we can now observe the event and refresh the list of loaded messages which will then reveal the message with restricted visibility.

DemoApp/StreamChat/Components/DemoChatChannelVC.swift
```diff
@@ -89,4 +89,14 @@ final class DemoChatChannelVC: ChatChannelVC, UIGestureRecognizerDelegate {
...
+    
+    // MARK: - EventsControllerDelegate
+
+    override func eventsController(_ controller: EventsController, didReceiveEvent event: Event) {
+        super.eventsController(controller, didReceiveEvent: event)
+        
+        if event is MessageUpdatedEvent {
+            channelController.synchronize()
+        }
+    }
```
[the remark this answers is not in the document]
nuno-vieira: The problem here is that if the users has loaded previous pages, this will reset the channel right? And if the message belongs to an old page, it will not even show? 🤔

Sources/StreamChat/Database/DatabaseSession.swift
```diff
@@ -918,8 +918,16 @@ extension DatabaseSession {
...
-private extension EventType {
+private extension EventPayload {
     var shouldCreateMessageInDatabase: Bool {
-        [.channelUpdated, .messageNew, .notificationMessageNew, .channelTruncated].contains(self)
+        switch eventType {
+        case .channelUpdated, .messageNew, .notificationMessageNew, .channelTruncated:
+            return true
+        case .messageUpdated:
+            guard let message else { return false }
+            return !message.restrictedVisibility.isEmpty
```
martinmitrevski, line 928: shouldn't we check whether the current user is in this array?

laevandus: Done

PR #3683, opened 2025-05-29
Read and update token request completions within a single closure for avoiding data race issues

laevandus:
### 🔗 Issue Links

Resolves: [IOS-879](https://linear.app/stream/issue/IOS-879)

### 🎯 Goal

Avoid a possibility of calling token request completion multiple times due to data race issue with read and updating a shared dictionary

### 📝 Summary

* Read and mutate within the same barrier block

### 🛠 Implementation

We had similar issue in ConnectionRepository a while back which got fixed after doing a similar change.

### 🧪 Manual Testing Notes

Try logging in and out with and without network for simulating retries.

### ☑️ Contributor Checklist


Sources/StreamChat/Repositories/AuthenticationRepository.swift
```diff
@@ -285,15 +285,13 @@ class AuthenticationRepository {
...
-        }
-        completionBlocks?.forEach { $0(error) }
-        tokenQueue.async(flags: .barrier) {
+            self._isGettingToken = false
             self._tokenRequestCompletions = []
             self._consecutiveRefreshFailures = 0
+            return completions
         }
+        completionBlocks?.forEach { $0(error) }
     }
```
laevandus, line 295: I was not able to reproduce manually, but we sure have a data race issue with reading and mutating `_tokenRequestCompletions`.

laevandus: Flow can be that first we read the list of completion handlers, then call these which in turn could theoretically reenter this function before existing completions are cleared.

nuno-vieira (approved):
LGTM! ✅

Issue #3703, opened 2025-06-26
[Bug]: PushNotificationType doesn't support the reaction.new event type


alemar11:
### Actual behaviour

When a push notification of type "reaction.new" is received, the PushNotificationType is initialized as nil because .reactionNew is [not handled](https://github.com/GetStream/stream-chat-swift/blob/ba80a94585c8f3085beb70c1fcddbde7fbddcd23/Sources/StreamChat/APIClient/ChatRemoteNotificationHandler.swift#L34).

This behavior causes ChatRemoteNotificationHandler.handleNotification(completion:) to call the completion handler with an UnknownNotificationContent.

### Expected behaviour

ChatRemoteNotificationHandler.handleNotification(completion:) should return a valid MessageNotificationContent in the completion when handling a "reaction.new" push notification.

### Steps to reproduce

```bash
Just add a reaction to a chat message
```

### Which packages are you using?

StreamChat

### Which SDK version are you using?

4.80.0

### Which iOS version is the issue reproducible?

iOS 18

### Which iPhone Device is the issue reproducible?

-

### Which Xcode version are you using?

-

laevandus:
Hi, thank you for the report. With new notification types it is an issue indeed. Will update the issue when it is fixed.

nuno-vieira:
Hello @alemar11,

This now available on the latest release (4.81.0)!

There were some minor changes in the naming of the push notification types. This due to the fact, that only after this release the new push notification types are officially supported.

Best,
Nuno

PR #3708, opened 2025-07-02
Add new Push v3 notification types

nuno-vieira:
### 🔗 Issue Links
- https://linear.app/stream/issue/IOS-957/bug-pushnotificationtype-doesnt-support-the-reactionnew-event-type
- https://github.com/GetStream/stream-chat-swift/issues/3703

### 🎯 Goal

Add support for the new Push v3 notification types

### 📝 Summary

All supported push notification types:
- `PushNotificationType.messageNew`
- `PushNotificationType.messageUpdated`
- `PushNotificationType.messageReminderDue`
- `PushNotificationType.reactionNew`

### 🧪 Manual Testing Notes

- Open Demo App Configuration
- Tap "App Key"
- Use "Frankfurt C2 Staging"
- Login with Han Solo
- Add reactions
- Update messages
- Other users should receive push notifications

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **New Features**
  * Added support for handling new push notification types, including message updates and new reactions.
  * Expanded immediate handling for certain notification types, allowing new reaction notifications to bypass further processing.

* **Tests**
  * Added tests to verify correct handling of both recognized and unrecognized push notification types.

### DIFF
--- a/Sources/StreamChat/APIClient/ChatRemoteNotificationHandler.swift
+++ b/Sources/StreamChat/APIClient/ChatRemoteNotificationHandler.swift
@@ -38,14 +38,6 @@ public struct PushNotificationType: Equatable {
     public static var messageUpdated: PushNotificationType = .init(eventType: .messageUpdated)
     /// When the push notification is for a new reaction.
     public static var reactionNew: PushNotificationType = .init(eventType: .reactionNew)
-
-    /// Checks if the push notification type is known and supported by the SDK.
-    var isKnown: Bool {
-        name == EventType.messageNew.rawValue ||
-            name == EventType.messageReminderDue.rawValue ||
-            name == EventType.messageUpdated.rawValue ||
-            name == EventType.reactionNew.rawValue
-    }
 }
 
 public class UnknownNotificationContent {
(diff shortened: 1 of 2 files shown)


Sources/StreamChat/APIClient/ChatRemoteNotificationHandler.swift
```diff
@@ -31,15 +32,21 @@ public struct PushNotificationType: Equatable {
 
     init?(eventType: EventType) {
         switch eventType {
-        case .messageNew, .messageReminderDue:
+        case .messageNew, .messageReminderDue, .reactionNew, .messageUpdated:
```
laevandus: Looking at this and hinking if we could just use EventType and not having the manual mapping which needs to be updated if there are more event types in the future. Since it represents EventType, could we just do typealias here (typealias PushNotificationType = EventType) or just always init it using event type's raw value. 🤔

nuno-vieira: Yes, it needs to be manually updated. My concern here is that, if we just do this:
```swift
 init(eventType: EventType) {
        self.name = eventType.rawValue
    }
```

Then, all events will generate a valid push notification type. Since now push notifications can be changed and updated, it means that if the backend decides to create a new push notification based on an event, it means the parsing will pass and will be a valid push notification. Which could be problematic if the customer is not switching over this new event type 🤔 I don't know... I the idea here was to make this more contrallable, and whenever a new event is added, we need to act on it and add a new `PushNotificationType` case.

nuno-vieira: I could also do something like this: 🤔 

```swift
var isKnown: Bool {
        return name == EventType.messageNew.rawValue ||
               name == EventType.messageReminderDue.rawValue ||
               name == EventType.messageUpdated.rawValue ||
               name == EventType.reactionNew.rawValue
    }
```

nuno-vieira: I've updated to the latter, let me know what you think @laevandus. We can always remove the `isKnown` in the future, if we believe it will not be needed or is problematic.

martinmitrevski: what's the benefit of this being controlled by the SDK? If new push notifications are introduced, customers should be able to directly use them imo. Therefore, if there are no real benefits, I would even remove the `isKnown` check.

nuno-vieira: Removed it then 👍

martinmitrevski (approved):

PR #3713, opened 2025-07-08
[CI] Improve linters and git hooks

testableapple:
### 🎯 Goal

We bumped SwiftFormat, SwiftLint and SwiftGen versions and introduced a new approach to git hooks on Feeds. The goal of this PR is to align this across all iOS repos.


Sources/StreamChat/Models/ChatMessage.swift
```diff
@@ -578,6 +579,8 @@ extension ChatMessage: Hashable {
         return true
     }
 
+    // swiftlint:enable cyclomatic_complexity
```
laevandus, line 582: Not part of this PR, but feels like we should disable rules where we need to manually opt-out.

testableapple: I'm not sure we have to manually opt out here, but I added it just in case. Let me know if it's not required.

nuno-vieira (approved):

PR #3738, opened 2025-07-22
Handle Push v3 Reaction Notifications + Skip edit message push notification

nuno-vieira:
### 🔗 Issue Links

Related to https://linear.app/stream/issue/IOS-855/ios-support-push-preferences

### 🎯 Goal

Handle Push v3 changes related to skip push and emoji code in reactions.

### 📝 Summary

- Add `skipPush` parameter to `MessageController.editMessage()`
- Add `skipPush` and `emojiCode` to `MessageController.addReaction()`
- Add `Components.availableReactionsPushEmojis` for rendering emojis in reaction push notifications
- Add `Components.isReactionPushEmojisEnabled` to control whether an emoji should be set when adding a reaction
- Add `MessageNotificationContent.reaction` that can be used by Notification Service Extension

### 🛠 Implementation

By default, when adding a reaction, an emoji will be assigned to a reaction. If it is not assigned, the customer can always use the NSE to customise the content of the push notification and render the emoji or whatever they need in the NSE.

### 🧪 Manual Testing Notes

1. Add a reaction with User A
2. User B (recipient) should receive a push notification with an emoji
3. Go to the Demo App configuration (On both devices)
4. Change the isReactionPushEmojisEnabled to false (On both devices)
5. Repeat the process
6. It should still render the emoji (This time it will use the NSE)

### ☑️ Contributor Checklist

### DIFF
--- a/DemoApp/Screens/AppConfigViewController/AppConfigViewController.swift
+++ b/DemoApp/Screens/AppConfigViewController/AppConfigViewController.swift
@@ -180,6 +180,7 @@ class AppConfigViewController: UITableViewController {
 
     enum ComponentsConfigOption: String, CaseIterable {
         case isUniqueReactionsEnabled
+        case isReactionPushEmojisEnabled
         case shouldMessagesStartAtTheTop
         case shouldAnimateJumpToMessageWhenOpeningChannel
         case shouldJumpToUnreadWhenOpeningChannel
@@ -445,6 +446,10 @@ class AppConfigViewController: UITableViewController {
         case .isUniqueReactionsEnabled:
             cell.accessoryView = makeSwitchButton(Components.default.isUniqueReactionsEnabled) { newValue in
                 Components.default.isUniqueReactionsEnabled = newValue
+            }
+        case .isReactionPushEmojisEnabled:
+            cell.accessoryView = makeSwitchButton(Components.default.isReactionPushEmojisEnabled) { newValue in
+                Components.default.isReactionPushEmojisEnabled = newValue
             }
         case .shouldMessagesStartAtTheTop:
             cell.accessoryView = makeSwitchButton(Components.default.shouldMessagesStartAtTheTop) { newValue in
--- a/Sources/StreamChat/APIClient/ChatRemoteNotificationHandler.swift
+++ b/Sources/StreamChat/APIClient/ChatRemoteNotificationHandler.swift
@@ -9,15 +9,18 @@ import UserNotifications
 public class MessageNotificationContent {
     public let message: ChatMessage
     public let channel: ChatChannel?
+    public let reaction: PushReactionInfo?
     public let type: PushNotificationType
 
     init(
         message: ChatMessage,
         channel: ChatChannel?,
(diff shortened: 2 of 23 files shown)


Sources/StreamChat/APIClient/ChatRemoteNotificationHandler.swift
```diff
@@ -90,6 +93,35 @@ public class ChatPushNotificationInfo {
     }
 }
 
+/// The information about a reaction in a push notification.
+public struct PushReactionInfo {
+    /// The type of the reaction.
+    public let rawType: String
```
martinmitrevski: why not just `type` or `reactionType` to be more consistent with the API?

TestTools/StreamChatTestTools/Mocks/StreamChat/Workers/MessageUpdater_Mock.swift
```diff
@@ -309,13 +316,14 @@ final class MessageUpdater_Mock: MessageUpdater {
             break
         }
         
-        //downloadAttachment_completion_result?  .invoke(with: completion)
+        // downloadAttachment_completion_result?  .invoke(with: completion)
```
martinmitrevski: should we remove this?

DemoApp/Screens/AppConfigViewController/AppConfigViewController.swift
```diff
@@ -180,6 +180,7 @@ class AppConfigViewController: UITableViewController {
 
     enum ComponentsConfigOption: String, CaseIterable {
         case isUniqueReactionsEnabled
+        case isReactionPushEmojisEnabled
```
martinmitrevski, line 183: Do we need config for this? Better to just have it on always.

Sources/StreamChat/APIClient/ChatRemoteNotificationHandler.swift
nuno-vieira: Yeah initially I wanted to use raw types, but does not make sense, I will change this 👍

TestTools/StreamChatTestTools/Mocks/StreamChat/Workers/MessageUpdater_Mock.swift
nuno-vieira: No idea why this is here. @laevandus Do you know? It seems you are last in the git blame

DemoApp/Screens/AppConfigViewController/AppConfigViewController.swift
nuno-vieira: Yes we need to disable it and QA the NSE logic

TestTools/StreamChatTestTools/Mocks/StreamChat/Workers/MessageUpdater_Mock.swift
laevandus: Seems like something what I forgot to delete since it is handled above

PR #3750, opened 2025-07-25
Add `LivestreamChannelController`

nuno-vieira:
### 🔗 Issue Links
https://linear.app/stream/issue/IOS-1025/create-a-livestream-controller

### 🎯 Goal

Create a new `LivestreamChannelController` which is a controller suitable for Livestreams due to better performance. It does not rely on DB persistence. But has fewer features than a regular channel controller.

### 📝 Summary

- `LivestreamChannelController` handles events manually instead of relying on Event Middlewares and DB updates
- Adds Payload -> Model mapping
- Adds a `DemoLivestreamChatChannelVC` in the Demo App to test drive the controller
- Adds a way for discarding older messages automatically:
    - Configurable through `LivestreamChannelController.maxMessageLimitOptions`
    - Exposes functions to handle pagination while discarding older messages:
         - `pause()` that will stop collecting messages in memory on `message.new` event
         - `resume()` that will load the first page and start collecting messages again on `message.new`
         - When `createNewMessage()` is called, it resumes automatically if it was paused.

#### Performance Benefits

The `LivestreamChannelController` is optimised for livestream scenarios:
- **More performant and more efficient** - No database overhead, which reduces CPU usage
- **Lower memory usage** - Messages are not saved into the DB, and older in-memory messages are automatically discarded once they reach a defined limit

#### Comparison with Regular Chat

| Feature | Regular Chat | Live Chat |
|---------|--------------|-----------|
| Performance | Good | Excellent |
| Memory Usage | Higher | Lower |
| Messages Capping | ❌ | ✅ |
| Reactions | ✅ | ✅  |
| Attachments Rendering | ✅ | ✅  |
| Attachment Upload Progress | ✅ | ❌ |
| Commands | ✅ | ❌ |
| Read Indicators | ✅ | ❌ |
| Typing Indicators | ✅ | ❌ |
| Message Editing | ✅ | ❌ |
| Threads | ✅ | ❌ |
| Offline Support | ✅ | ❌ |

#### Current Issues/Limitations
- The channel list won't re-order for these live channels, since the last message is not saved into the DB. (Workaround is to perform a light-weight fetch of the channel, to trigger re-order on viewWillDisappear)
- Ephemeral commands are not correctly rendered
- Attachment upload progress not working (Since it relies on DB updates)
- When channel is updated, ownCapabilities is removed, which makes the Reactions UI not working. The channel.updated event does not come with `ownCapabilities`.

### 🧪 Manual Testing Notes

How to test:
1. Swipe a channel in the channel list
2. Tap on "..."
3. Tap on "Show as Livestream Controller"

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **New Features**
  * Introduced a Livestream Chat Channel experience, including a new in-memory LivestreamChannelController for real-time chat without local persistence.
  * Added a dedicated Livestream Chat UI with message list, composer, reactions, and pinning support.
  * Provided custom message actions and a SwiftUI-based reactions list for livestream messages.
  * Exposed Combine publishers for livestream channel state updates.
  * Enabled manual event handling for channels opting out of middleware processing.
  * Added a new alert action to show channels as Livestream Controllers.

* **Improvements**
  * Enhanced message and channel model conversion from payloads.
  * Added new methods for updating channel and message models selectively.
  * Improved slow mode management with explicit enable/disable and input validation removal.
  * Made various UI and controller methods and properties publicly accessible for greater flexibility.
  * Added support for memory warning notifications to app state observers.
  * Included channel context explicitly in message event processing for better event handling.

* **Bug Fixes**
  * Corrected internal logic for disabling slow mode and event processing with channel context.

* **Chores**
  * Updated project structure to include new files and organize demo and core components.


Sources/StreamChat/Controllers/ChannelController/LivestreamChannelController.swift
```diff
@@ -0,0 +1,502 @@
...
+        )
+        
+        // Clear current messages when loading first page
+        messages = []
+        
+        updateChannelData(channelQuery: query, completion: completion)
+    }
+
+    /// Marks the channel as read.
+    public func markRead(completion: ((Error?) -> Void)? = nil) {
```
martinmitrevski: Not sure if we should support this - maybe better to just log a warning that marking read is not supported?

Sources/StreamChat/Controllers/ChannelController/LivestreamChannelController.swift
```diff
@@ -0,0 +1,502 @@
...
+
+// MARK: - Delegate Protocol
+
+/// Delegate protocol for `LivestreamChannelController`
+public protocol LivestreamChannelControllerDelegate: AnyObject {
+    /// Called when the channel data is updated
+    /// - Parameters:
+    ///   - controller: The controller that updated
+    ///   - channel: The updated channel the controller manages.
+    func livestreamChannelController(
```
martinmitrevski, line 1025: maybe we keep the same delegate method signatures as the regular controller, to make the integration easier?

Sources/StreamChat/Controllers/ChannelController/LivestreamChannelController.swift
```diff
@@ -0,0 +1,502 @@
...
+    ) {}
+    
+    func livestreamChannelController(
+        _ controller: LivestreamChannelController,
+        didUpdateMessages messages: [ChatMessage]
+    ) {}
+}
+
+private extension ChatMessage {
+    func updateReadBy(
```
martinmitrevski: also not sure if we're going to need this

Sources/StreamChat/Models/Channel.swift
```diff
@@ -602,3 +646,164 @@ public extension ChatChannel {
         ownCapabilities.contains(.shareLocation)
     }
 }
+
+// MARK: - Payload -> Model Mapping
+
+extension ChannelPayload {
```
martinmitrevski: nit: let's remove these in a separate file, this file becomes too big.

Sources/StreamChat/Models/Channel.swift
```diff
@@ -602,3 +646,164 @@ public extension ChatChannel {
...
+        unreadCount: ChannelUnreadCount?
+    ) -> ChatChannel {
+        let channelPayload = channel
+
+        // Map members
+        let mappedMembers = members.compactMap { $0.asModel(channelId: channelPayload.cid) }
+
+        // Map latest messages
+        let reads = channelReads.map { $0.asModel() }
+        let latestMessages = messages.prefix(5).compactMap {
```
martinmitrevski: why 5?

Sources/StreamChat/Models/Channel.swift
```diff
@@ -602,3 +646,164 @@ public extension ChatChannel {
...
+            notificationsMuted: notificationsMuted,
+            memberExtraData: [:]
+        )
+    }
+}
+
+extension ChannelReadPayload {
+    /// Converts the ChannelReadPayload to a ChatChannelRead model
+    /// - Returns: A ChatChannelRead instance
+    func asModel() -> ChatChannelRead {
```
martinmitrevski: maybe instead we always return an empty array for reads?

Sources/StreamChat/Models/ChatMessage.swift
```diff
@@ -692,3 +693,129 @@ public struct MessageDeliveryStatus: RawRepresentable, Hashable {
     /// The message delivery state for message failed to be sent/edited/deleted.
     public static let failed = Self(rawValue: "failed")
 }
+
+// MARK: - Payload -> Model Mapping
+
+extension MessagePayload {
```
martinmitrevski: same here, we should move the direct mappings to separate files.

Sources/StreamChat/Models/ChatMessage.swift
```diff
@@ -692,3 +693,129 @@ public struct MessageDeliveryStatus: RawRepresentable, Hashable {
...
+            isShadowed: isShadowed,
+            reactionScores: reactionScores,
+            reactionCounts: reactionCounts,
+            reactionGroups: [:],
+            author: author,
+            mentionedUsers: mentionedUsers,
+            threadParticipants: threadParticipants,
+            attachments: attachments,
+            latestReplies: [],
+            localState: nil,
```
martinmitrevski: will this have an impact in some cases, e.g. if you upload attachments?

Sources/StreamChat/WebSocketClient/Events/MessageEvents.swift
```diff
@@ -197,6 +197,9 @@ public struct MessageReadEvent: ChannelSpecificEvent {
 
     /// The unread counts of the current user.
     public let unreadCount: UnreadCount?
+
+    /// The last read message id.
+    public let lastReadMessageId: MessageId?
```
martinmitrevski: why do we need this?

Sources/StreamChat/Workers/EventNotificationCenter.swift
```diff
@@ -64,6 +103,171 @@ class EventNotificationCenter: NotificationCenter, @unchecked Sendable {
...
+    }
+    
+    // MARK: - Event Creation Helpers
+    
+    private func createMessageNewEvent(from payload: EventPayload, cid: ChannelId) -> MessageNewEvent? {
+        guard
+            let userPayload = payload.user,
+            let messagePayload = payload.message,
+            let createdAt = payload.createdAt,
+            let channel = try? database.writableContext.channel(cid: cid)?.asModel(),
```
martinmitrevski: how does this exactly work? We won't have the channel in the db, right?

Sources/StreamChatUI/ChatChannel/ChatLivestreamChannelVC.swift
```diff
@@ -0,0 +1,498 @@
...
+        isLastMessageFullyVisible
+    }
+
+    /// A component responsible to handle when to load new or old messages.
+    private lazy var viewPaginationHandler: StatefulViewPaginationHandling = {
+        InvertedScrollViewPaginationHandler.make(scrollView: messageListVC.listView)
+    }()
+
+    /// The throttler to make sure that the marking read is not spammed.
+    var markReadThrottler: Throttler = Throttler(interval: 3, queue: .main)
```
martinmitrevski: we should also clean up this part related to marking read

Sources/StreamChatUI/ChatChannel/ChatLivestreamChannelVC.swift
```diff
@@ -0,0 +1,498 @@
+//
+// Copyright © 2025 Stream.io Inc. All rights reserved.
+//
+
+import StreamChat
+import UIKit
+
+open class ChatLivestreamChannelVC: _ViewController,
```
martinmitrevski: thinking if we even need this one - maybe we just add support to the regular one for livestream channel controller? But only if it's not too much work.

martinmitrevski (commented):
Looks good so far! From the testing, I think we should be more aggressive with fetching the next page (do it sooner), there's a bit of delay while loading more messages. Also, we should think about the initial blank state when opening the chat - since there's no local data available now - do we expose some loading state?

Sources/StreamChat/Controllers/ChannelController/LivestreamChannelController.swift
```diff
@@ -0,0 +1,502 @@
...
+        let endpoint: Endpoint<ChannelPayload> =
+            .updateChannel(query: channelQuery)
+        
+        let requestCompletion: (Result<ChannelPayload, Error>) -> Void = { [weak self] result in
+            DispatchQueue.main.async { [weak self] in
+                guard let self = self else { return }
+
+                switch result {
+                case .success(let payload):
+                    self.handleChannelPayload(payload, channelQuery: channelQuery)
```
laevandus, line 776: Looks like it could run on the api client's queue instead of main?

Sources/StreamChat/Controllers/ChannelController/LivestreamChannelController.swift
```diff
@@ -0,0 +1,502 @@
...
+    private func updateMessagesArray(with newMessages: [ChatMessage], pagination: MessagesPagination?) {
+        let newMessages = Array(newMessages.reversed())
+        switch pagination?.parameter {
+        case .lessThan, .lessThanOrEqual:
+            // Loading older messages - append to end
+            messages.append(contentsOf: newMessages)
+            
+        case .greaterThan, .greaterThanOrEqual:
+            // Loading newer messages - insert at beginning
+            messages.insert(contentsOf: newMessages, at: 0)
```
laevandus: Could cause invalid order of messages when web-sockets events are coming in while paginating. Maybe it should not support jump to since that is the easiest way to break it. Probably you have TODO about sorting order questions.

Sources/StreamChat/Controllers/ChannelController/LivestreamChannelController.swift
```diff
@@ -0,0 +1,502 @@
...
+        currentMessages.insert(message, at: 0)
+        messages = currentMessages
+
+        notifyDelegateOfChanges()
+    }
+    
+    private func handleUpdatedMessage(_ updatedMessage: ChatMessage) {
+        var currentMessages = messages
+
+        if let index = currentMessages.firstIndex(where: { $0.id == updatedMessage.id }) {
```
laevandus: If Instruments show that id lookups are a bit too slow, we can do binary search here since the the array of messages should be sorted by createdAt.

Sources/StreamChat/Controllers/ChannelController/LivestreamChannelController.swift
nuno-vieira: Does livestream channels don't have a read count? Otherwise I can remove this

Sources/StreamChat/Controllers/ChannelController/LivestreamChannelController.swift
nuno-vieira: It should be the same. What is different?

Sources/StreamChat/Controllers/ChannelController/LivestreamChannelController.swift
nuno-vieira: Yes this is code I forgot to delete when I was working on reads

Sources/StreamChat/Models/Channel.swift
nuno-vieira: Did not notice this actually, cursor things 😂 I will update htis

Sources/StreamChat/Models/Channel.swift
nuno-vieira: We can keep this for now. It might be useful when needing to map channels to payloads for other things. Since we are doing this, better to try mapping all the data we can

Sources/StreamChat/Models/ChatMessage.swift
nuno-vieira: For now it is working fine, but I will revisit this

Sources/StreamChat/Workers/EventNotificationCenter.swift
nuno-vieira: The problem here is that our domain events require a channel model, and most of the events don't include the channel data. So I'm basically fetching the channel, just to be able to create the event, but we don't really need it. I will probably cache this data as well.

Sources/StreamChatUI/ChatChannel/ChatLivestreamChannelVC.swift
nuno-vieira: What do you mean? Having the livestream controller inside ChatChannelVC?

nuno-vieira: I think it is going to make it very hard to maintain

Sources/StreamChat/Controllers/ChannelController/LivestreamChannelController.swift
nuno-vieira: What do you mean? 🤔

Sources/StreamChat/Controllers/ChannelController/LivestreamChannelController.swift
nuno-vieira: We have skipping messages logic on the view side, so in that case, it should be fine. But yeah, good point, it could be worth doing knda the same logic inside the controller 🤔

Sources/StreamChat/Controllers/ChannelController/LivestreamChannelController.swift
nuno-vieira: I think we should handle all completions on the main thread, so that it is easier to handle concurrency, as well as calling the delegates on the main thread

Sources/StreamChat/Models/Channel.swift
nuno-vieira: Done

Sources/StreamChat/Controllers/ChannelController/LivestreamChannelController.swift
nuno-vieira: Ok, so when the first page is not loaded, we do not add new messages to the data, so for now this should be more than enough 👍 Let me know what you think!

Sources/StreamChatUI/ChatChannel/ChatLivestreamChannelVC.swift
nuno-vieira: Done

Sources/StreamChat/Controllers/ChannelController/LivestreamChannelController.swift
nuno-vieira: Done

Sources/StreamChat/Controllers/ChannelController/LivestreamChannelController.swift
martinmitrevski: the livestream prefix

Sources/StreamChatUI/ChatChannel/ChatLivestreamChannelVC.swift
martinmitrevski: probably having both of them would be harder to maintain, but whatever you prefer. My initial thinking was that they share the same interface - we just replace the implementation in ChatChannelVC.

Sources/StreamChat/Controllers/ChannelController/LivestreamChannelController.swift
martinmitrevski: it depends - the completion handlers can be called on the main thread, but here I guess we're handling / parsing the channel payload, so might be good to use the api client's queue.

Sources/StreamChat/Controllers/ChannelController/LivestreamChannelController.swift
martinmitrevski: one note here - some customers (especially the ones we built this for), won't use our standard chat UI components.

Sources/StreamChat/Controllers/ChannelController/LivestreamChannelController.swift
nuno-vieira: Ah, yes, that is a good point. I didn't see any performance penalty by doing it on the main thread, but I will probably improve this then

Sources/StreamChat/Controllers/ChannelController/LivestreamChannelController.swift
nuno-vieira: Yes, this logic is not relying on our UI components 👍 It is built-in inside the controller

Sources/StreamChat/Controllers/ChannelController/LivestreamChannelController.swift
nuno-vieira: All looks good for now, so resolving this 👍

Sources/StreamChatUI/ChatChannel/ChatLivestreamChannelVC.swift
nuno-vieira: It is going to have a lot of if-statements everywhere, unless we create a sharable interface between them, which could be a good option. For now, I think we should not include any UI out of the box, since the customer is using the controller directly. At most, we should create a dummy UI for the Demo App only.

Sources/StreamChat/Workers/EventNotificationCenter.swift
nuno-vieira: Added the caching 👍

Sources/StreamChat/Controllers/ChannelController/LivestreamChannelController.swift
nuno-vieira: I'm going to leave it like this, since this makes sure that the setting of the channels and messages are done on the main thread. The performace of doing the mapping out side of the main thread is also not a big deal

Sources/StreamChat/Controllers/ChannelController/LivestreamChannelController.swift
```diff
@@ -0,0 +1,894 @@
...
+        }
+    }
+
+    // MARK: - Public Methods
+
+    /// Synchronizes the controller with the backend data.
+    /// - Parameter completion: Called when the synchronization is finished.
+    public func synchronize(_ completion: ((_ error: Error?) -> Void)? = nil) {
+        // Populate the initial data with existing cache.
+        if loadInitialMessagesFromCache, let cid = self.cid, let channel = dataStore.channel(cid: cid) {
```
laevandus, line 215: In the existing controller we do DB fetch in the channel of messages accessor, should we follow the same pattern or fine to have it in a different way? I guess it is fine since synchornize should always be called and it is better if such logic is in a single place.

Sources/StreamChat/Controllers/ChannelController/LivestreamChannelController.swift
```diff
@@ -0,0 +1,894 @@
...
+    /// - Parameter completion: Callback when the API call is completed.
+    public func loadFirstPage(_ completion: ((_ error: Error?) -> Void)? = nil) {
+        var query = channelQuery
+        query.pagination = .init(
+            pageSize: channelQuery.pagination?.pageSize ?? .messagesPageSize,
+            parameter: nil
+        )
+
+        // Clear current messages when loading first page
+        messages = []
```
laevandus: Might cause a flash where messages becomes empty, api request is still loading for, then messages gets populated and UI refreshes.

Sources/StreamChat/Controllers/ChannelController/LivestreamChannelController.swift
```diff
@@ -0,0 +1,894 @@
...
+        updateChannelData(channelQuery: query, completion: completion)
+    }
+
+    /// Loads next messages from backend.
+    /// - Parameters:
+    ///   - messageId: ID of the current first message. You will get messages `newer` than the provided ID.
+    ///   - limit: Limit for page size. By default it is 25.
+    ///   - completion: Called when the network request is finished.
+    public func loadNextMessages(
+        after messageId: MessageId? = nil,
```
laevandus, line 274: Theoretically I could cause duplicate messages in the messages array by calling this twice with the same or overlapping ids, but I guess it is OK to skip this in the first iteration since it is kind of special case/misusing the API as intended.

Sources/StreamChat/Controllers/ChannelController/LivestreamChannelController.swift
```diff
@@ -0,0 +1,894 @@
...
+    }
+
+    // MARK: - Helpers
+
+    /// Helper method to execute the callbacks on the main thread.
+    func callback(_ action: @escaping () -> Void) {
+        DispatchQueue.main.async {
+            action()
+        }
+    }
```
laevandus: How do we feel about adding @MainActor to completion handlers?

Sources/StreamChat/Workers/ManualEventHandler.swift
```diff
@@ -0,0 +1,254 @@
...
+    // The channels for which events will not be processed by the middlewares.
+    private var channelIds: Set<ChannelId> = []
+
+    // Some events require the chat channel data, so we need to fetch it from local DB.
+    // We try to only do this once, to avoid unnecessary DB fetches.
+    private var cachedChannels: [ChannelId: ChatChannel] = [:]
+
+    init(
+        database: DatabaseContainer,
+        queue: DispatchQueue = DispatchQueue(label: "io.getstream.chat.manualEventHandler", qos: .background)
```
laevandus: I think we want to use .utility here instead.

> On iPhones, discretionary and background operations, including networking, are paused when Low Power Mode is enabled.

Sources/StreamChat/Controllers/ChannelController/LivestreamChannelController.swift
```diff
@@ -0,0 +1,894 @@
...
+        }
+
+        // If we don't have the first page loaded, do not insert new messages
+        // they will be inserted once we load the first page again.
+        if !hasLoadedAllNextMessages {
+            return
+        }
+
+        currentMessages.insert(message, at: 0)
+        messages = currentMessages
```
martinmitrevski: this is not really thread safe

Sources/StreamChat/Workers/ManualEventHandler.swift
nuno-vieira: Makes sense 👍

Sources/StreamChat/Controllers/ChannelController/LivestreamChannelController.swift
nuno-vieira: It is, since the events controller always calls the delegate on the main thread. So we only manipulate the state on the main thread

martinmitrevski: hmm, do we want everything to come on the main thread internally? Imagine 30+ assignments of the whole messages array per second, might put some pressure.

martinmitrevski: in some cases also the contains operation

Sources/StreamChat/Controllers/ChannelController/LivestreamChannelController.swift
```diff
@@ -0,0 +1,894 @@
...
+    }
+
+    private func handleUpdatedMessage(_ updatedMessage: ChatMessage) {
+        var currentMessages = messages
```
martinmitrevski: if everything runs on the main thread, do we even need to copy this to a temp var? We can update the array directly, right?

Sources/StreamChat/Workers/ChannelUpdater.swift
```diff
@@ -593,14 +593,25 @@ class ChannelUpdater: Worker {
     /// - Parameters:
     ///   - cid: Channel id of the channel to be marked as read
     ///   - cooldownDuration: Duration of the time interval users have to wait between messages.
-    ///   Specified in seconds. Should be between 0-120. Pass 0 to disable slow mode.
     ///   - completion: Called when the API call is finished. Called with `Error` if the remote update fails.
     func enableSlowMode(cid: ChannelId, cooldownDuration: Int, completion: ((Error?) -> Void)? = nil) {
+        guard cooldownDuration >= 1, cooldownDuration <= 120 else {
```
martinmitrevski: this feels unnecessary here - especially if the supported interval changes.

Sources/StreamChat/Workers/ManualEventHandler.swift
```diff
@@ -0,0 +1,254 @@
...
+
+    // MARK: - Event Creation Helpers
+
+    private func createMessageNewEvent(from payload: EventPayload, cid: ChannelId) -> MessageNewEvent? {
+        guard
+            let userPayload = payload.user,
+            let messagePayload = payload.message,
+            let createdAt = payload.createdAt,
+            let channel = getLocalChannel(id: cid),
+            let currentUserId = database.writableContext.currentUser?.user.id
```
martinmitrevski, line 101: can't we just fetch the current user id once from the DB, instead of on every event?

Sources/StreamChat/Workers/ManualEventHandler.swift
```diff
@@ -0,0 +1,254 @@
...
+    }
+
+    // This is only needed because some events wrongly require the channel to create them.
+    private func getLocalChannel(id: ChannelId) -> ChatChannel? {
+        if let cachedChannel = cachedChannels[id] {
+            return cachedChannel
+        }
+
+        let channel = try? database.writableContext.channel(cid: id)?.asModel()
+        cachedChannels[id] = channel
```
martinmitrevski: is everything handled on the main thread here as well? If not, we should handle this.

Sources/StreamChat/Workers/ManualEventHandler.swift
```diff
@@ -0,0 +1,254 @@
...
+        )
+    }
+
+    // This is only needed because some events wrongly require the channel to create them.
+    private func getLocalChannel(id: ChannelId) -> ChatChannel? {
+        if let cachedChannel = cachedChannels[id] {
+            return cachedChannel
+        }
+
+        let channel = try? database.writableContext.channel(cid: id)?.asModel()
```
martinmitrevski: is it possible we end up without a channel in the DB? Since that would make us not handle any of these events.

Sources/StreamChat/Controllers/ChannelController/LivestreamChannelController.swift
nuno-vieira: It is better to do it here. Having it on accessing a property does not make sense. It is something we need to improve on v5. When we are accessing a property, we are not expecting to perform any fetch or any action, unless it is a function. So the behaviour of the other controllers is misleading, since you think you are just accessing a normal property, but no, it can perform a DB fetch.

Sources/StreamChat/Controllers/ChannelController/LivestreamChannelController.swift
nuno-vieira: Yup, makes sense 👍 If the request fails, it will also clear the channel for no reason

Sources/StreamChat/Controllers/ChannelController/LivestreamChannelController.swift
nuno-vieira: Yup. In fact, we probably should be even just using the cursor API, but to make this consisten with the regular channel controller, I didn't change it

Sources/StreamChat/Controllers/ChannelController/LivestreamChannelController.swift
nuno-vieira: Yes I think we can do that already or just wait for v5. I'll see if we can do this without any setbacks 👍

Sources/StreamChat/Controllers/ChannelController/LivestreamChannelController.swift
nuno-vieira: Yes this one we can do directly 👍

Sources/StreamChat/Workers/ChannelUpdater.swift
nuno-vieira: Yeah, for me it also does not make much sense. But I wanted to keep the logic like it is on the other controller. But we can remove this. For me I think it is better

Sources/StreamChat/Workers/ManualEventHandler.swift
nuno-vieira: This one should be fine, because the currentUser is already cached, we are actually not going to the DB in this case. So it is a bit redudant adding one more cache on top of a cache 😄

Sources/StreamChat/Workers/ManualEventHandler.swift
nuno-vieira: Hum, yeah this one seems like a potential thread-safe issue, I will investigate 👍

Sources/StreamChat/Workers/ManualEventHandler.swift
nuno-vieira: Hum, if the customer is not using the Channel List Controller, I think yeah, it could be problematic here 🤔 Not a big deal for now, since this specific customer is using it, but maybe what we can do is that on `synchronize()` we can make sure we save the channel into the DB.

Sources/StreamChat/Controllers/ChannelController/LivestreamChannelController.swift
nuno-vieira: Done 👍

Sources/StreamChat/Controllers/ChannelController/LivestreamChannelController.swift
```diff
@@ -0,0 +1,1034 @@
...
+    /// Disabled by default. If enabled, older messages will be automatically discarded
+    /// once the limit is reached. The `MaxMessageLimitOptions.recommended` is the recommended
+    /// configuration which uses 200 max messages with 50 discard amount.
+    /// This can be used to further improve the memory usage of the controller.
+    ///
+    /// - Note: In order to use this, if you want to support loading previous messages,
+    /// you will need to use `pause()` method before loading older messages. Otherwise the
+    /// pagination will also be capped. Once the user scrolls back to the newest messages, you
+    /// can call `resume()`. Whenever the user creates a new message, the controller will
+    /// automatically resume the controller.
```
laevandus:
```suggestion
    /// automatically resume.
```

Sources/StreamChat/Controllers/ChannelController/LivestreamChannelController.swift
```diff
@@ -0,0 +1,1034 @@
...
+        let limit = limit ?? channelQuery.pagination?.pageSize ?? .messagesPageSize
+        var query = channelQuery
+        query.pagination = MessagesPagination(pageSize: limit, parameter: .around(messageId))
+
+        updateChannelData(channelQuery: query, completion: completion)
+    }
+
+    /// Cleans the current state and loads the first page again.
+    /// - Parameter completion: Callback when the API call is completed.
+    public func loadFirstPage(_ completion: ((_ error: Error?) -> Void)? = nil) {
```
laevandus: I wonder if we could miss new messages when there are like 50 coming in per second and we start the api request. Do we need to internally collect ws events, then merge messages from these after api request finishes?

Sources/StreamChat/Workers/ManualEventHandler.swift
martinmitrevski: actually, they are not using the channel list here

Sources/StreamChat/Workers/ManualEventHandler.swift
nuno-vieira: Done

Sources/StreamChat/Controllers/ChannelController/LivestreamChannelController.swift
nuno-vieira: From my experience I didn't see this problem, because usually they will come with the request. But we can optimize this later on. For now I think it is not very crucial 🤔

Sources/StreamChat/Controllers/ChannelController/LivestreamChannelController.swift
nuno-vieira: Done ✅

Sources/StreamChat/Workers/ChannelUpdater.swift
nuno-vieira: Done

Sources/StreamChat/Workers/ManualEventHandler.swift
nuno-vieira: Fixed

Sources/StreamChat/Workers/ManualEventHandler.swift
nuno-vieira: Done ✅

Sources/StreamChat/Controllers/ChannelController/LivestreamChannelController.swift
nuno-vieira: Having the message processing in another queue, will make things more complex. For now, doing this on the main thread should give us enough performance. Later if we think that we need even better performance than we can change this

nuno-vieira: After profiling it, the messages processing like appending, deleting etc, does not even appear at the heaviest stack traces, it has very low impact, so for now it is not worth optimising this:

image

image

DemoApp/Screens/Livestream/DemoLivestreamMessageActionsVC.swift
```diff
@@ -0,0 +1,338 @@
...
+        let alert = UIAlertController(title: "Error", message: error.localizedDescription, preferredStyle: .alert)
+        alert.addAction(UIAlertAction(title: "OK", style: .default))
+        present(alert, animated: true)
+    }
+}
+
+// MARK: - Reactions Action Item
+
+/// Action item for adding reactions to a message
+public struct ReactionsActionItem: ChatMessageActionItem {
```
laevandus, line 312: Does not need to be public, but not a big deal.

Sources/StreamChat/Controllers/ChannelController/LivestreamChannelController.swift
```diff
@@ -0,0 +1,1100 @@
...
+    public private(set) var channelQuery: ChannelQuery
+
+    /// The identifier of a channel this controller observes.
+    public var cid: ChannelId? { channelQuery.cid }
+
+    /// The `ChatClient` instance this controller belongs to.
+    public let client: ChatClient
+
+    /// The channel the controller represents.
+    public private(set) var channel: ChatChannel? {
```
laevandus, line 38: This (and `messages`) is mutated from the main thread, but I could read it from any thread. Might become a problem if this is being done by users, but at the same time we could apply proper guards later in V5. That said, multiple other properties in existing controllers are not properly thread-safe. So maybe we should just have a ticket to make things properly thread-safe separately in a new ticket.

Sources/StreamChat/Workers/ManualEventHandler.swift
```diff
@@ -0,0 +1,256 @@
...
+        case .channelUpdated:
+            return createChannelUpdatedEvent(from: eventPayload, cid: cid)
+
+        default:
+            return nil
+        }
+    }
+
+    private func isRegistered(channelId: ChannelId) -> Bool {
+        queue.sync { channelIds.contains(channelId) }
```
laevandus, line 90: This should always run on writableContext's queue, therefore queue.sync could be skipped.

laevandus (approved):
I am good with these changes. ✅

PR #3753, opened 2025-07-28
[CI] Update swiftformat config

testableapple:


lefthook.yml
```diff
@@ -1,23 +1,38 @@
...
+        - "**/Generated/**"
+        - Sources/StreamChatSwiftUI/StreamNuke/**
+        - Sources/StreamChatSwiftUI/StreamSwiftyGif/**
+        - Sources/StreamChatSwiftUI/StreamDifferenceKit/**
       skip:
         - merge
         - rebase
 
-    - run: mint run swiftlint lint --config .swiftlint.yml --strict --progress --quiet --reporter json {staged_files}
+pre-push:
```
laevandus, line 28: So the summary is that if I commit, then we autocorrect and include autocorrected files with the commit. If I push, then we just lint again for making sure no linting related issues end up in develop?

testableapple: yeap

laevandus (approved):

PR #3759, opened 2025-08-01
Use Swift 6 with complete concurrency checking

laevandus:
### 🔗 Issue Links

Resolves: [IOS-734](https://linear.app/stream/issue/IOS-734/swift-6-llc-strict-concurrency-checking)
Resolves: [IOS-735](https://linear.app/stream/issue/IOS-735/swift-6-uikit-strict-concurrency-checking)

### 🎯 Goal

Use Swift 6 with strict concurrency checking set to complete

### 📝 Summary

* Set Swift version to 6 with strict concurrency checking complete (including test tools, demo app, mock server)
* Drop Xcode 15 support
* Add Sendable conformances to public types
* Add @Sendable to many closure callbacks
* LLC: Controller delegates and completion handlers are @MainActor (removed callbackDelegate)
* UIKit: Nuke was upgraded from 10 to 12 (brings Swift 6 support)
* UIKit: Use @MainActor in delegates

### 🛠 Implementation

Tried to make it as concise as possible, therefore in some cases there is `nonisolated(unsafe)`, because fixing them properly leads to huge refactoring (too large scope). In many cases this is actually all right.

### 🧪 Manual Testing Notes

Needs full regression testing round.

### ☑️ Contributor Checklist


Sources/StreamChatUI/StreamNuke/Nuke/Caching/Cache.swift
```diff
@@ -0,0 +1,213 @@
+//
```
laevandus, line 1: Do not review anything in the StreamNuke folder since this is not ours.

Sources/StreamChat/APIClient/Endpoints/Payloads/UserPayloads.swift
```diff
@@ -35,7 +35,7 @@ enum UserPayloadsCodingKeys: String, CodingKey, CaseIterable {
 // MARK: - GET users
 
 /// An object describing the incoming user JSON payload.
-class UserPayload: Decodable {
+class UserPayload: Decodable, @unchecked Sendable {
```
martinmitrevski, line 38: Why not final, but unchecked Sendable like the others?

Sources/StreamChat/Controllers/Controller.swift
```diff
@@ -8,19 +8,17 @@ import Foundation
...
-        if Thread.current.isMainThread && callbackQueue == .main {
-            action()
+    func callback(_ action: @escaping @MainActor @Sendable() -> Void) {
+        if Thread.current.isMainThread {
+            MainActor.assumeIsolated {
+                action()
+            }
         } else {
-            callbackQueue.async {
+            DispatchQueue.main.async {
```
martinmitrevski, line 21: Shouldn't we use the StreamConcurrency.onMain here?

Sources/StreamChat/Controllers/PollController/PollController.swift
```diff
@@ -135,10 +134,16 @@ public class PollController: DataController, DelegateCallable, DataStoreProvider
         )
         pollsRepository = client.pollsRepository
         super.init()
-        eventsController.delegate = self
+        eventsObserver = client.subscribe(toEvent: PollVoteChangedEvent.self, handler: { [weak self] event in
```
martinmitrevski, line 137: Why do we change this?

Sources/StreamChat/Database/DatabaseContainer.swift
```diff
@@ -2,7 +2,7 @@
 // Copyright © 2025 Stream.io Inc. All rights reserved.
 //
 
-import CoreData
+@preconcurrency import CoreData
```
martinmitrevski, line 5: Interesting, why do we need it?

Sources/StreamChat/Utils/Data+Gzip.swift
```diff
@@ -150,7 +150,7 @@ struct GzipError: Swift.Error {
 
     init(code: Int32, msg: UnsafePointer<CChar>?) {
         message = {
-            guard let msg = msg, let message = String(validatingUTF8: msg) else {
+            guard let msg = msg, let message = String(validatingCString: msg) else {
```
martinmitrevski, line 153: Why do we change this?

martinmitrevski (commented):
Looks good ✅ Left few small comments, and let's do a round of tests. Did you test the changes already?

Sources/StreamChat/Utils/Data+Gzip.swift
laevandus: It is a deprecated in Swift 6: [reference](https://developer.apple.com/documentation/swift/string/init(validatingutf8:)-208fn) and Xcode shows a warning here since we bumped Swift version.

martinmitrevski:
@laevandus actually it doesn't build for me on Xcode 16.2.

Sources/StreamChat/Database/DatabaseContainer.swift
laevandus: Does not like setting merge policy.

`Sources/StreamChat/Database/DatabaseContainer.swift:25:31 Reference to var 'NSMergeByPropertyObjectTrumpMergePolicy' is not concurrency-safe because it involves shared mutable state`

Sources/StreamChat/APIClient/Endpoints/Payloads/UserPayloads.swift
laevandus: It is because it is subclassed here:
`final class CurrentUserPayload: UserPayload, @unchecked Sendable {`

Sources/StreamChat/Controllers/Controller.swift
laevandus: There is a tiny difference and I did not want to change it (being extra careful). `StreamConcurrency.onMain` has `DispatchQueue.main.sync` here because it supports returning a value (needed in some cases).

Sources/StreamChat/Controllers/PollController/PollController.swift
laevandus: Since we changed controller delegates to be @MainActor (delegates ran on the callbackQueue which is main), then this has a side-effect where controller can't conform to `EventsControllerDelegate` easily since delegate conformance would then require main actor, but controller types are not. 

`@MainActor public protocol EventsControllerDelegate: AnyObject {`

Option 1 (current): Add @MainActor to`EventsControllerDelegate` and change the code where we use it in controllers
Option 2: Do not add @MainActor to `EventsControllerDelegate`, but if @MainActor view model or etc implements it, it as to use nonisolated for the delegate method and dispatch the call to main thread.

laevandus: This was discussed internally and we decided to use the subscribe method instead for controllers.

Sources/StreamChat/Audio/AudioPlayer/AudioPlaying.swift
```diff
@@ -126,7 +126,10 @@ open class StreamAudioPlayer: AudioPlaying, AppStateObserverDelegate {
 
     open func subscribe(_ subscriber: AudioPlayingDelegate) {
         multicastDelegate.add(additionalDelegate: subscriber)
-        subscriber.audioPlayer(self, didUpdateContext: context)
+        nonisolated(unsafe) let unsafeSubscriber = subscriber
+        StreamConcurrency.onMain {
+            unsafeSubscriber.audioPlayer(self, didUpdateContext: context)
+        }
```
nuno-vieira, line 132: Why do we need this? Can't we make StreamAudioPlayer MainActor?

nuno-vieira (approved):
Great Job! 👏

martinmitrevski (approved):
Looks good, let's merge it in the base branch ✅

PR #3788, opened 2025-08-21
Added a new upload endpoint

martinmitrevski:
### 🔗 Issue Links

Resolves https://linear.app/stream/issue/IOS-927/ios-support-for-the-new-upload-endpoint.

### 🎯 Goal

Allow possibility to upload attachments without tying them to a channel/message.

### 📝 Summary

### 🛠 Implementation

Exposed the new endpoint.

### 🎨 Showcase

| Before | After |
| ------ | ----- |
|  img   |  img  |

### 🧪 Manual Testing Notes

Run this somewhere:

```swift
if let image = UIImage(named: "Stream_logo")?.pngData() {
    let imageURL = URL(fileURLWithPath: NSTemporaryDirectory()).appendingPathComponent("image.png")
    try? image.write(to: imageURL)
    let uploadingState = AttachmentUploadingState(
        localFileURL: imageURL,
        state: .pendingUpload,
        file: .init(type: .png, size: Int64(image.count), mimeType: "image/png")
    )
    let attachment = StreamAttachment(
        type: .image,
        payload: image,
        downloadingState: nil,
        uploadingState: uploadingState
    )
    self.client?.upload(attachment, progress: { progress in
        print("=========== Progress: \(progress)")
    }, completion: { result in
        switch result {
        case .success(let file):
            print("======== \(file.fileURL)")
        case .failure(let error):
            print("======== Error: \(error)")
        }
    })
}
```

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **New Features**
  * Upload standalone attachments (not tied to a channel/message) via ChatClient with optional progress updates, returning an UploadedFile (URL and metadata).
  * New StreamAttachment type to build uploads with payload and local uploading/downloading state.
* **Refactor**
  * Renamed channel-scoped upload path to uploadChannelAttachment and added a separate path for standalone uploads.
  * Standalone uploads are not queued for offline processing.
  * Added a public initializer to simplify creating AttachmentUploadingState.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 # Upcoming
 
-### 🔄 Changed
+## StreamChat
+### ✅ Added
+- Add support for `user.messages.deleted` event [#3792](https://github.com/GetStream/stream-chat-swift/pull/3792)
 
 # [4.86.0](https://github.com/GetStream/stream-chat-swift/releases/tag/4.86.0)
 _August 21, 2025_
--- a/Sources/StreamChat/Controllers/ChannelController/LivestreamChannelController.swift
+++ b/Sources/StreamChat/Controllers/ChannelController/LivestreamChannelController.swift
@@ -790,11 +790,20 @@ public class LivestreamChannelController: DataStoreProvider, EventsControllerDel
     // MARK: - EventsControllerDelegate
 
     public func eventsController(_ controller: EventsController, didReceiveEvent event: Event) {
-        guard let channelEvent = event as? ChannelSpecificEvent, channelEvent.cid == cid else {
-            return
+        if let channelEvent = event as? ChannelSpecificEvent, channelEvent.cid == cid {
+            handleChannelEvent(event)
         }
 
-        handleChannelEvent(event)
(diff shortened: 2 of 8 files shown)


Sources/StreamChat/Models/Attachments/ChatMessageAttachment.swift
```diff
@@ -4,6 +4,39 @@
 
 import Foundation
 
+public struct StreamAttachment<Payload> {
```
nuno-vieira, line 7: The Stream prefix could be a bit generic, and it does not give enough context on how this is used. What about this:
```suggestion
/// An attachment that is not assigned to any message or channel.
public struct StandaloneAttachment<Payload> {
```

nuno-vieira: Another option is to try to just reuse the API from `ChannelUpdater.uploadFile()`, and not expose any new type 🤔

martinmitrevski: About the naming, I would actually prefer to keep `StreamAttachment`. The reason is that on other products, all attachments would be standalone - not tied to channels, posts, etc. And it matches the API one, which is just `Attachment`.

laevandus (approved):
✅ Looks good

PR #3794, opened 2025-08-26
Do not reset blocked state when channel specific WS event without blocked state is handled

laevandus:
### 🔗 Issue Links

Resolves: [IOS-1094](https://linear.app/stream/issue/IOS-1094)

### 🎯 Goal

Channel list removes blocked channel on it is presented

### 📝 Summary

* Any WS event which contains channel information can reset blocked state of a channel in channel lists what include blocked channels

### 🛠 Implementation

If there is a need for showing both blocked and unblocked channels (blocked are automatically hidden) then one needs to have a query looking like this:
```swift
filter: .and([
            .containMembers(userIds: [currentUserId]),
            .or([
                .and([.equal(.blocked, to: false), .equal(.hidden, to: false)]),
                .and([.equal(.blocked, to: true), .equal(.hidden, to: true)])
            ])
        ])
```
This works fine until blocked channel is opened which triggers notification.mark_read WS event which has channel data, but blocked field is nil. This triggered SDK to incorrectly setting isBlocked of the channel to false which in turn triggered removing it from the channel list.

### 🧪 Manual Testing Notes

1. Log in with user which has someone blocked (I used r2-d2 who has blocked anakin)
2. Open channel list query picker and choose "Blocked Unblocked with Matching Hidden Channels"
3. Open a 1:1 channel with blocked user
4. Navigate back to channel list
Result: that channel is still there.

### ☑️ Contributor Checklist

## Summary by CodeRabbit

- New Features
  - Added a demo filter to view channels where blocked and hidden states match (both on or both off).

- Bug Fixes
  - Prevented channels from disappearing from the list when blocked channels are included.
  - Ensured blocked/hidden states remain consistent when update events omit those fields.

- Tests
  - Added tests validating channel list behavior for blocked/hidden scenarios.

- Chores
  - Updated data model defaults and test helpers to support optional blocked/hidden values.

- Documentation
  - Added upcoming CHANGELOG entry describing the bug fix.


DemoApp/StreamChat/Components/DemoChatChannelListVC.swift
```diff
@@ -48,6 +48,16 @@ final class DemoChatChannelListVC: ChatChannelListVC {
             .equal(.blocked, to: true)
         ])
     ]))
+    
+    lazy var blockedUnblockedWithHiddenChannelsQuery: ChannelListQuery = .init(
```
martinmitrevski, line 52: why this one? We already have:
```
lazy var allBlockedChannelsQuery: ChannelListQuery = .init(filter: .and([
        .containMembers(userIds: [currentUserId]),
        .or([
            .equal(.blocked, to: false),
            .equal(.blocked, to: true)
        ])
    ]))
```
This is the one customers would use.

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
```diff
@@ -285,7 +285,7 @@ extension NSManagedObjectContext {
         // for blocked 1:1 channels on channel list query
         if let isBlocked = payload.isBlocked {
             dto.isBlocked = isBlocked
-        } else {
+        } else if query != nil {
```
martinmitrevski: is this condition enough? I think it's better that this is only solved on the backend.

DemoApp/StreamChat/Components/DemoChatChannelListVC.swift
laevandus: What I learned was that they are going to use the one with blocked and hidden fields.

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
laevandus: To be honest I am not sure why the else part was added anyway. But there is a comment just above which seems to indicate that this is required. Ideally we wouldn't change the local state if payload does not have it. Currently I only change it in a way that it matches to the comment above. Maybe I should try to remove the else part completely. 🤔

martinmitrevski: that could be dangerous too - if the backend just removes the field if you unblock someone.

tcamin: Looking at the git history I found this [comment](https://github.com/GetStream/stream-chat-swift/pull/3223/files#r1636081727), which still doesn't clarify the need for this fallback. I believe the existing `isBlocked` state should remain unchanged if a websocket event payload doesn’t include it, wdyt?

laevandus: I removed the else part and tested our manual regression testing test cases, all was good! Just in case added default value to the Core Data model as well since it did not have it before (looks mistake, but it was fine since ChannelDTO always updated this property).

Sources/StreamChat/Database/StreamChatModel.xcdatamodeld/StreamChatModel.xcdatamodel/contents
```diff
@@ -49,7 +49,7 @@
         <attribute name="hasUnreadSorting" attributeType="Integer 16" defaultValueString="0" usesScalarValueType="YES"/>
         <attribute name="id" optional="YES" attributeType="String"/>
         <attribute name="imageURL" optional="YES" attributeType="URI"/>
-        <attribute name="isBlocked" optional="YES" attributeType="Boolean" usesScalarValueType="YES"/>
+        <attribute name="isBlocked" optional="YES" attributeType="Boolean" defaultValueString="NO" usesScalarValueType="YES"/>
```
laevandus, line 52: Safe if default value is set, but it is optional (since now ChannelDTO might not update the property anymore).

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
```diff
@@ -285,8 +285,6 @@ extension NSManagedObjectContext {
         // for blocked 1:1 channels on channel list query
         if let isBlocked = payload.isBlocked {
             dto.isBlocked = isBlocked
-        } else {
-            dto.isBlocked = false
         }
```
laevandus, line 288: Went through manual regression testing steps for blocked, and all was good.

martinmitrevski (approved):

martinmitrevski:
@laevandus check the LLC test failures before merging

PR #3802, opened 2025-09-03
Add channelRole filter key for ChannelListQuery

laevandus:
### 🔗 Issue Links

Resolves: [IOS-1112](https://linear.app/stream/issue/IOS-1112)

### 🎯 Goal

New `channel_role` filter key was added which needs to be exposed with local DB filtering support

### 📝 Summary

* Add `FilterKey.channelRole` for `ChannelListQuery`
* Store original unmapped role value in the local DB for local filtering (note that MemberRole struct maps `channel_member` to `member`)

### 🛠 Implementation

It is unfortunate that `MemberRole` maps `member` and `channel_member` to `member` which means that having a filter like:
```
filter: .and([
            .containMembers(userIds: [currentUserId]),
            .equal(.channelRole, to: "channel_moderator")
        ])
```
would fail because CoreData DB stores "member". This PR changes it so that now we store whatever backend returns.

### 🧪 Manual Testing Notes

1. Log in with r2-d2
2. Change channel list filter to Moderator Channels
Result: one channel appears

### ☑️ Contributor Checklist

## Summary by CodeRabbit

- New Features
  - Added “Moderator Channels” filter in the demo app to quickly view channels where you are a moderator.
  - Introduced a new channel list filter option that lets you filter channels by your role.
- Documentation
  - CHANGELOG updated with details about the new role-based channel filtering capability.

### DIFF
--- a/Sources/StreamChat/Models/Member.swift
+++ b/Sources/StreamChat/Models/Member.swift
@@ -172,11 +172,26 @@ public struct MemberRole: RawRepresentable, Codable, Hashable, ExpressibleByStri
 
     public init(rawValue: String) {
         self.rawValue = rawValue
+        rawChannelValue = rawValue
     }
 
     public init(stringLiteral value: String) {
         self.init(rawValue: value)
     }
+    
+    init(rawChannelValue: String) {
+        // Historically these have been mapped when decoding JSON
+        switch rawChannelValue {
+        case "channel_member":
+            self.init(rawValue: Self.member.rawValue)
+        case "channel_moderator":
+            self.init(rawValue: Self.moderator.rawValue)
+        default:
+            self.init(rawValue: rawChannelValue)
+        }
+    }
+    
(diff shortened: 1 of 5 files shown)


Sources/StreamChat/Models/Member.swift
```diff
@@ -207,6 +222,8 @@ public extension MemberRole {
         default:
             self = MemberRole(rawValue: value)
         }
+        // Store the original raw value from the backend for local filtering
+        rawChannelValue = value
```
laevandus, line 226: This has the value DB gives us and now we save backend value to the CoreData store. We used to save `member` instead of `channel_member` which messes up local filtering. I added new init for MemberRole structure so that now when we have `channel_member` in the DB, initing `MemberRole` still gives me `MemberRole.member`.
Don't want to risk breaking integrations where someone locally compares channel member roles.

PR #3826, opened 2025-10-02
[V5] Remove controllerWillChangeChannels and cleanup willChange handling

laevandus:
### 🔗 Issue Links

Resolves: [IOS-1004](https://linear.app/stream/issue/IOS-1004)

### 🎯 Goal

Remove `controllerWillChangeChannels` because it requires complicated handling on the DB observer level

### ☑️ Contributor Checklist


nuno-vieira:
> does this have any impact on the UI layer? Also, @nuno-vieira what do you think?

Should be fine. No UI is using this at all, so we should remove it.

PR #3836, opened 2025-10-07
Add support for deleting messages only for the current user

nuno-vieira:
### 🔗 Issue Links

https://linear.app/stream/issue/IOS-1151/delete-only-for-me-feature

### 🎯 Goal

Add support for deleting messages only for the current user.

### 📝 Summary

- Added `ChatMessage.deletedForMe` property
- Added `MessageDeletedEvent.deletedForMe` property
- Added `ChatMessageController.deleteMessageForMe()`
- Added demo app message action to test the deleted only for me feature

### 🧪 Manual Testing Notes

⚠️ Do not use `luke_skywalker` or `han_solo`.  At the moment of writing, there is a bug where delete for me events are leaked to admins.

1. Open the app with Developers Scheme
2. Open a channel
3. Long tap one of the current users message
4. Delete the message only for you
5. It should delete the message and show the "Only for you" label
6. It should not delete the message for the other user.

### ☑️ Contributor Checklist

## Summary by CodeRabbit

- **New Features**
  - You can now delete a message only for yourself; it will appear deleted in your view while remaining visible to others.
  - A “Delete only for me” action is available in message actions for messages you sent.

- **UI**
  - Message timestamps/labels append an indicator when a message is deleted only for you; message state reflects that.

- **Documentation**
  - Changelog updated with a new StreamChat entry describing the feature.

- **Tests**
  - Added tests covering delete-for-me behavior, payloads, and event handling.

### DIFF
--- a/Sources/StreamChat/Database/DTOs/MessageDTO.swift
+++ b/Sources/StreamChat/Database/DTOs/MessageDTO.swift
@@ -991,7 +991,10 @@ extension NSManagedObjectContext: MessageDatabaseSession {
 
         dto.isSilent = payload.isSilent
         dto.isShadowed = payload.isShadowed
-        dto.deletedForMe = payload.deletedForMe ?? false
+        if let deletedForMe = payload.deletedForMe {
+            dto.deletedForMe = deletedForMe
+        }
+
         // Due to backend not working as advertised
         // (sending `shadowed: true` flag to the shadow banned user)
         // we have to implement this workaround to get the advertised behavior


Sources/StreamChat/Database/DTOs/MessageDTO.swift
```diff
@@ -990,6 +991,7 @@ extension NSManagedObjectContext: MessageDatabaseSession {
 
         dto.isSilent = payload.isSilent
         dto.isShadowed = payload.isShadowed
+        dto.deletedForMe = payload.deletedForMe ?? false
```
laevandus: We have had issues before where using default value has led to the state being out of sync. I think it is better to avoid it here and update the state only when the payload contains it.

nuno-vieira: In this case, it should be fine. If there is no deletedForMe, it actually means it is false. For the message deleted event, though I was careful with this and only updated if it actually exists

nuno-vieira: Anyway, just be extra safe, I applied the same logic just in case ✅

martinmitrevski (approved):
LGTM ✅

PR #3846, opened 2025-10-16
Add support for message delivered info

nuno-vieira:
### 🔗 Issue Links

https://linear.app/stream/issue/IOS-1152/message-delivery-status

### 🎯 Goal

Add information to the message whenever it has been delivered to the recipient.

### 📝 Summary

**New Public API:**
- Adds `ChatRemoteNotificationHandler.markMessageAsDelivered(deliveries:)`
- Adds `ChatChannel.reads(message:)` and `ChatChannel.deliveredReads(message:)`
- Adds `ChatChannelRead.lastDeliveredAt`
- Adds `ChatChannelRead.lastDeliveredMessageId`

**New in UI SDK:**
- Adds support for double grey checkmark when message is delivered but not read

**Demo App:**
- Updates the Demo App to show off the delivery feature:
   - Adds gray double checkmark to the message view when message is delivered but not read
   - Adds gray double checkmark to the channel item view when preview message is delivered but not read
   - Adds "Show messages reads info" action to the message to show all delivered and reads info

**TODO:**
- [x] Backend config flag to feature flag the feature
- [x] Update the NotificationService to not use the Tracker internally since it is useless
- [x] User Privacy Settings update

### 🛠 Implementation

This PR introduces a message delivery tracking system that automatically marks messages as delivered when they are received through WebSocket events, channel list synchronisation, or push notifications. The system includes throttling to prevent server spam and ensures efficient delivery status updates.

The message delivery tracking consists of four components that work together:

1. **ChannelDeliveryTracker (New)** - Core component that manages pending deliveries and throttling.
2. **ChannelDeliveredMiddleware (New)** - WebSocket event middleware that uses the `ChannelDeliveryTracker` to respond to the following events: `message.new`, `message.delivered`, `message.read`
3. **ChannelListController** - Marks channels as delivered without throttling when fetching channels.
4. **ChatRemoteNotificationHandler** - Handles delivery marking for push notifications, ~~which internally uses the `ChannelDeliveryTracker` to make sure requests are throttled as well~~.
5. **MessageDeliveryCriteriaValidator**: Responsible for validating if a message can be marked as delivered.

> [!TIP]
>**Debouncer vs Throttling vs Throttling-Trailing**
>If you are confused about these, here is a visual demo:
>https://claude.ai/public/artifacts/8e0e8f6a-674e-4154-9b97-85a99f4956d0

#### ChannelDeliveryTracker

The `ChannelDeliveryTracker` is the central component that manages pending channel deliveries and implements throttling to prevent server spam.

**Key Features:**
- Tracks channels that need delivery marking with their latest message IDs (`[ChannelId: MessageId]`)
- Uses a throttler to limit delivery requests to at most one per second. It uses `broadcastLatest: true` to make sure the last call is always executed.
-  Whenever a delivery call is executed, it only removes the channels from the pending delivery dictionary that still match the same message ID, so that new message IDs from the same channel are not discarded.
- **Thread Safety**: All operations are performed on a concurrent queue with barrier flags

**Rules to mark a message as delivered:**
- Must be from another user
- Message hasn't been marked as delivered yet (createdAt > lastDeliveryAt)
- Message hasn't been marked as read yet (createdAt > lastReadAt)

#### ChannelDeliveredMiddleware

The `ChannelDeliveredMiddleware` processes WebSocket events and automatically triggers delivery marking for new messages.

**Event Handling:**
- **MessageNewEvent**: Adds channel to pending deliveries and executes the request if not throttled.
- **NotificationMarkReadEvent**: Cancels pending deliveries for the given channel.
- **MessageDeliveredEvent**: Updates local channel read data with delivery data. Useful to update the users who received the message.

#### ChannelListController

The `ChannelListController` automatically marks channels as delivered during synchronisation operations.

**Integration Points:**
- **Initial Sync**: When `synchronize()` is called, channels are marked as delivered instantly after a successful fetch.
- **Load Next Channels**: When `loadNextChannels()` is called, new channels are marked as delivered.

**Note:** Only the channels that respect the rules mentioned above are marked as delivered.

#### ChatRemoteNotificationHandler

The `ChatRemoteNotificationHandler` exposes an easy way to mark messages as delivered when they arrive via push notifications. It internally checks if the message was already delivered to avoid redundant API calls. If the App is active, once the push notification is handled, the message will already be marked as delivered.

Customers just need to call the function `ChatRemoteNotificationHandler.markMessageAsDelivered()` and everything will be handled automatically for them.

> [!NOTE]
> The `ChatRemoteNotificationHandler` does not use the `ChannelDeliveryTracker` internally since the `NotificationService` is spawned for every push, which means we can't have a throttling mechanism here since everything is created from scratch.

#### Full flow with edge cases

```
1. New message from Channel A is added to pending delivery
2. Request is fired instantly since no throttling is happening
3. New message from Channel B is added before 1s (Request not fired)
4. New message from Channel C is added before 1s (Request not fired)
6. After 1s, Request from 4. is executed with Channel B & Channel C
7. While it is being executed, a new message from the same channel B is added
8. Request from 5. finishes and removes Channel C from pending delivery, not Channel B, since it changed the message meanwhile
9. After 1s again, the request from 6. is executed with Channel B, since it was not cleared by 7.
```

### 🎨 Showcase

https://github.com/user-attachments/assets/0112ec6f-2eda-4cd4-a195-568b90e76c39

### 🧪 Manual Testing Notes

**Message is marked delivered when the user receives a new message:**
- Open Channel List with User A
- Send a message with User B
- User B should see a double grey checkmark on the new message
- Long-tapping the new message should show the delivered user
- Going to the Channel List with User B should show a double grey checkmark

**Message is marked delivered when the user receives a push notification:**
- Close the app with User A
- Send a message with User B
- User A should receive a push notification
- User B should see a double grey checkmark on the new message
- Long-tapping the new message should show the delivered user
- Going to the Channel List with User B should show a double grey checkmark

**Message is marked delivered when the user loads the channel list:**
- Turn off push notifications with User A
- Close the app with User A
- Send a message with User B
- User B should NOT see a double grey checkmark on the new message
- Open the Channel List with User A
- User B should see a double grey checkmark on the new message
- Going to the Channel List with User B should show a double grey checkmark

**Message is NOT marked delivered when the user is:** (Follow mark read rules)
- The channel has delivery feature disabled.
- The channel is muted or hidden
- The current user has delivery receipts disabled (Demo App User Profile View)
- The message is not from a shadow-banned or muted user
- The message is a thread reply
- The message is from the current user

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **New Features**
  * Message delivery tracking per channel and a delivered state.
  * Delivery receipts privacy setting to enable/disable delivery notifications.
  * "Show Message Reads" UI: per-user Delivered/Read view for messages.
  * Option to opt channels into delivery event tracking.

* **Improvements**
  * UI shows a double grey checkmark for delivered messages.
  * Message status now distinguishes sent, delivered, and read.
* **Chores**
  * Demo app settings and notification flow updated to surface delivery behavior.

### DIFF
--- a/DemoApp/Shared/DemoUsers.swift
+++ b/DemoApp/Shared/DemoUsers.swift
@@ -5,7 +5,7 @@
 import Foundation
 import StreamChat
 
-var apiKeyString = ProcessInfo.processInfo.environment["CUSTOM_API_KEY"] ?? DemoApiKeys.usEastC6.rawValue
+var apiKeyString = ProcessInfo.processInfo.environment["CUSTOM_API_KEY"] ?? DemoApiKeys.frankfurtC1.rawValue
 let applicationGroupIdentifier = "group.io.getstream.iOS.ChatDemoApp"
 
 enum DemoUserType {


Sources/StreamChat/APIClient/Endpoints/ChannelEndpoints.swift
```diff
@@ -292,6 +292,16 @@ extension Endpoint {
         )
     }
 
+    static func markChannelsDelivered(payload: ChannelDeliveredRequestPayload) -> Endpoint<EmptyResponse> {
```
martinmitrevski, line 295: a bit strange name for the endpoint

Sources/StreamChat/APIClient/ChatRemoteNotificationHandler.swift
```diff
@@ -128,12 +128,14 @@ public class ChatRemoteNotificationHandler {
     let chatCategoryIdentifiers: Set<String> = ["stream.chat", "MESSAGE_NEW"]
     let channelRepository: ChannelRepository
     let messageRepository: MessageRepository
+    let channelDeliveryTracker: ChannelDeliveryTracker
```
nuno-vieira: Actually... it is kind of pointless to use the throttler here. The notification service is spawned fresh whenever a new push is delivered, so we can't keep track of a dictionary

Sources/StreamChat/WebSocketClient/Events/MessageEvents.swift
```diff
@@ -258,3 +258,57 @@ public struct NewMessageErrorEvent: ChannelSpecificEvent {
     public let cid: ChannelId
     public let error: Error
 }
+
+/// Triggered when a message is delivered to a user.
+public struct MessageDeliveredEvent: ChannelSpecificEvent {
```
nuno-vieira: Probably worth converting to class to save SDK size. (We can probably do this too all events actually, there are a lot of them that are still structs)

Sources/StreamChat/APIClient/ChatRemoteNotificationHandler.swift
martinmitrevski: yes, we don't need throttling here.

Sources/StreamChat/APIClient/Endpoints/EndpointPath.swift
```diff
@@ -125,6 +126,7 @@ enum EndpointPath: Codable {
         case let .markChannelRead(channelId): return "channels/\(channelId)/read"
         case let .markChannelUnread(channelId): return "channels/\(channelId)/unread"
         case .markAllChannelsRead: return "channels/read"
+        case .markChannelsDelivered: return "channels/delivered"
```
martinmitrevski, line 129: Better naming would be markMessageDelivered. Although the endpoint is directly under channels, the payload has message info (and it's for a message), so it's more appropriate imo.

Sources/StreamChat/WebSocketClient/Events/MessageEvents.swift
martinmitrevski: yes, let's use class here from this PR, we can convert the other ones in another PR.

martinmitrevski (commented):
Looks great! Added few small comments, mostly around naming the endpoint.

DemoApp/StreamChat/Components/DeliveredMessages/DemoMessageReadsInfoView.swift
```diff
@@ -0,0 +1,214 @@
...
+            }
+        }
+    }
+    
+    // MARK: - Computed Properties
+    
+    private var deliveredUsers: [ChatUser] {
+        channel?.deliveredReads(for: message)
+            .sorted { $0.lastDeliveredAt ?? Date.distantPast < $1.lastDeliveredAt ?? Date.distantPast }
+            .map(\.user) ?? []
```
laevandus, line 92: It keeps resorting often when I have 2 other simulators open and scroll the channel. Sounds like we update delivery status even when the user has marked it as delivered?

Sources/StreamChat/WebSocketClient/ChannelDeliveryTracker.swift
```diff
@@ -0,0 +1,104 @@
...
+
+import Foundation
+
+/// A component that tracks pending channel deliveries and manages throttled requests to mark messages as delivered.
+///
+/// This component is used in the event middleware and the push notifications handler to
+/// avoid spamming the server whenever there is multiple messages coming in.
+class ChannelDeliveryTracker {
+    /// Dictionary to track channels and their latest message IDs for delivery marking.
+    private(set) var pendingDeliveredChannels: [ChannelId: MessageId] = [:]
```
laevandus, line 13: Should it also track channels and messages which were marked as delivered. Otherwise we can have duplicate API requests for the same message and channel

Sources/StreamChat/APIClient/Endpoints/ChannelEndpoints.swift
nuno-vieira: Yeah I agree.. but it is consistent with the backend. Initially I had everything named channel delivered. But it was just too odd. So I kept the internals as channel delivered, but the rest as message delivered 🤔

Sources/StreamChat/APIClient/Endpoints/EndpointPath.swift
nuno-vieira: Yeah I followed backend's approach. Since mark read is also related to channel. And this one is too, but feels weird.

Sources/StreamChat/WebSocketClient/ChannelDeliveryTracker.swift
nuno-vieira: Yes, this is on my TODO list as well, I will add this soon

Sources/StreamChat/APIClient/Endpoints/ChannelEndpoints.swift
martinmitrevski: probably the most accurate representation of the backend would be `markChannelDelivered(payload: MessageDeliveredRequestPayload)`?

nuno-vieira: It is in plural, it is accepts multiple channels, that is why it is "channelsDelivered", not "channelDelivered"

Sources/StreamChat/APIClient/Endpoints/EndpointPath.swift
nuno-vieira: I'm going to stick with the current endpoint path. This is internal, so we can always change later, but at the moment, it follows the backend pattern. IMO this should live in the current user's endpoint path, since you change the messages delivered from multiple channels from the current user. So you are affecting the current user

DemoApp/StreamChat/Components/DeliveredMessages/DemoMessageReadsInfoView.swift
nuno-vieira: This view is refreshed whenever the channel is updated. When marking the channel read, it updates the channel, so the view is updated. Since the sorting does not have a tie breaker, it could lead to re-ordering. I've fixed it by adding a tie breaker 👍

Sources/StreamChat/WebSocketClient/ChannelDeliveryTracker.swift
nuno-vieira: Done ✅

CHANGELOG.md
```diff
@@ -20,7 +22,6 @@ _October 22, 2025_
 - Add support for deleting messages only for the current user [#3836](https://github.com/GetStream/stream-chat-swift/pull/3836)
   - Add `ChatMessageController.deleteMessageForMe()`
   - Add `ChatMessage.deletedForMe`
-- Allow observing poll changes in `PollVoteListController` [#3849](https://github.com/GetStream/stream-chat-swift/pull/3849)
```
martinmitrevski, line 26: why's this deleted?

Sources/StreamChat/APIClient/Endpoints/Payloads/ChannelDeliveredPayload.swift
```diff
@@ -0,0 +1,45 @@
+//
+// Copyright © 2025 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+/// A payload representing a delivered message with its channel and message identifiers.
+struct DeliveredMessagePayload: Encodable, Equatable {
```
martinmitrevski, line 8: shouldn't we use classes here?

Sources/StreamChat/Models/Channel.swift
```diff
@@ -383,6 +383,46 @@ extension ChatChannel {
...
+    /// A user is considered to have delivered a message when:
+    /// - The user has a read state in the channel
+    /// - The user's last delivered timestamp is after the message creation time
+    /// - The user is not the author of the message
+    ///
+    /// - Parameter message: The message to check delivery status for.
+    /// - Returns: Array of read states for users who have delivered the message.
+    public func deliveredReads(for message: ChatMessage) -> [ChatChannelRead] {
+        reads.filter { read in
+            // We add a 1 second error interval in case the delivery is instant.
```
martinmitrevski, line 420: don't quite understand, why do we need this?

martinmitrevski (approved):
Looks good to me ✅  Left few small comments.
"Should we provide the double grey checkmark as part of the UI SDK if deliveries are enabled?"
Yes, we already have it, right? It just needs to consider the delivery events from now on.

CHANGELOG.md
nuno-vieira: Ups, probably bad conflict resolution, will fix it 👍

Sources/StreamChat/APIClient/Endpoints/Payloads/ChannelDeliveredPayload.swift
nuno-vieira: We are using structs for all payload responses. The events is where it is problematic because it has a lot of nested classes inside, those should be classes. I've changed all message events to be class, and now we have a drop of 400KB :)

Sources/StreamChat/Models/Channel.swift
nuno-vieira: Comparing dates is very tricky. What was happening was that if I sent a message, and the other user was online in the channel list, the delivery date was the same as the `message.created`. In this case, the comparison below sometimes would pass or fail, so the double grey checkmark sometimes would not show at all. So the option would be to either always perform the delivery request 1s after, or just add here an additional 1 second, to make sure that we don't compare equal dates. This option was the least effort one, so I opted for this one

Sources/StreamChatUI/ChatChannel/ChatChannelVC.swift
```diff
@@ -586,6 +586,13 @@ open class ChatChannelVC: _ViewController,
            draftUpdatedEvent.cid == channelController.cid, draftUpdatedEvent.draftMessage.threadId == nil {
             messageComposerVC.content.draftMessage(draft)
         }
+
+        // The `ChatMessage` object has no data related to delivery info.
+        // So the diffing won't notify of UI changes, so we need to manually refresh
+        // the latest cell to update the double gray checkmark.
+        if event is MessageDeliveredEvent {
+            messageListVC.listView.reloadRows(at: [.init(item: 0, section: 0)], with: .none)
+        }
```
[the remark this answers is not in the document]
nuno-vieira: Ups! Thanks!

Sources/StreamChat/Models/Channel.swift
xsahil03x: @nuno-vieira We are using `leading: false` for the throttler to always post the initial delivery after 1 second.

xsahil03x: You can test it using the updated claude artifact

https://claude.ai/public/artifacts/a252f6fb-a950-4a16-89df-3136e0ac4a08

Sources/StreamChat/WebSocketClient/ChannelDeliveryTracker.swift
```diff
@@ -0,0 +1,104 @@
...
+        throttler.execute { [weak self] in
+            let deliveredMessages: [MessageDeliveryInfo] = self?.queue.sync {
+                return self?.pendingDeliveredChannels.map { channelId, messageId in
+                    MessageDeliveryInfo(channelId: channelId, messageId: messageId)
+                }
+            } ?? []
+
+            guard !deliveredMessages.isEmpty else { return }
+
+            self?.currentUserUpdater.markMessagesAsDelivered(deliveredMessages) { [weak self] error in
```
xsahil03x, line 83: The current API only supports upto 100 messages in payload. So you might also want to do it in a batch of 100 messages.

testableapple (approved):

PR #3857, opened 2025-10-28
Fix duplicated watch channel requests when a channel is created and it belongs to multiple queries

nuno-vieira:
### 🔗 Issue Links

https://linear.app/stream/issue/IOS-1221

### 🎯 Goal

Fix duplicated watch channel requests when a channel is created and it belongs to multiple queries.

#### Problem
When a channel was created, our `ChannelListLinker` would start watching it if it belonged to the query. The problem is that if the customer had multiple channel lists at the same time, and if this channel matched all queries, it would perform a watch request for all channel lists in parallel. 

#### Solution
A `ChannelWatcherHandler` object was introduced to track if there is any ongoing watch channel request already if yes, then do not perform any additional request.

#### Bonus Solution
We also had a problem when a channel was updated. Now, we also check if this channel already belongs to another query; if yes, then we do not watch it, since it is already being watched.

### 🧪 Manual Testing Notes

N/A, covered by unit tests.

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate "watch channel" requests when a channel is created and belongs to multiple channel lists.
  * Stopped unnecessary "watch channel" requests when a channel is updated but already belongs to another channel list, reducing redundant network activity and improving real-time update stability.

### DIFF
--- a/StreamChat.xcodeproj/project.pbxproj
+++ b/StreamChat.xcodeproj/project.pbxproj
@@ -1783,6 +1783,7 @@
 		ADB97E8C2EB129FB0065824C /* ChannelWatcherHandler.swift in Sources */ = {isa = PBXBuildFile; fileRef = ADB97E8B2EB129FB0065824C /* ChannelWatcherHandler.swift */; };
 		ADB97E8D2EB129FB0065824C /* ChannelWatcherHandler.swift in Sources */ = {isa = PBXBuildFile; fileRef = ADB97E8B2EB129FB0065824C /* ChannelWatcherHandler.swift */; };
 		ADB97E8F2EB1368D0065824C /* ChannelWatcherHandler_Tests.swift in Sources */ = {isa = PBXBuildFile; fileRef = ADB97E8E2EB1368D0065824C /* ChannelWatcherHandler_Tests.swift */; };
+		ADB97E912EB165890065824C /* ChannelWatcherHandler_Mock.swift in Sources */ = {isa = PBXBuildFile; fileRef = ADB97E902EB165890065824C /* ChannelWatcherHandler_Mock.swift */; };
 		ADBBDA1F279F0CEA00E47B1C /* VideoDurationFormatter.swift in Sources */ = {isa = PBXBuildFile; fileRef = ADBBDA1E279F0CEA00E47B1C /* VideoDurationFormatter.swift */; };
 		ADBBDA20279F0CEA00E47B1C /* VideoDurationFormatter.swift in Sources */ = {isa = PBXBuildFile; fileRef = ADBBDA1E279F0CEA00E47B1C /* VideoDurationFormatter.swift */; };
 		ADBBDA22279F0CFA00E47B1C /* UploadingProgressFormatter.swift in Sources */ = {isa = PBXBuildFile; fileRef = ADBBDA21279F0CFA00E47B1C /* UploadingProgressFormatter.swift */; };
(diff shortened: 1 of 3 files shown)


Sources/StreamChat/Workers/ChannelListLinker.swift
```diff
@@ -69,30 +72,46 @@ final class ChannelListLinker {
...
+                    return
+                }
+
+                guard !ongoingWatchingChannels.isExecutingRequest(for: channel.cid) else { return }
+
+                ongoingWatchingChannels.add(channelIds: [channel.cid])
+
                 worker.startWatchingChannels(withIds: [channel.cid]) { error in
+                    ongoingWatchingChannels.remove(channelIds: [channel.cid])
+
```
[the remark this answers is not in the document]
nuno-vieira: Thank you, Jedi 🙏 

Your Padawan,
Nuno

Sources/StreamChat/Workers/WatchingChannelsOngoingRequests.swift
```diff
@@ -0,0 +1,48 @@
+//
+// Copyright © 2025 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+/// Keeps track of ongoing channel watching requests to prevent duplicate calls.
+class WatchingChannelsOngoingRequests {
```
laevandus: Alternative idea:
Did you also consider creating something like `WatchChannelHandler` which would internally do this tracking (naming wise we have some other XYZHandler types as well). Just for keeping the coalescing code in a single type. Moreover, that one could also be reused for SyncRepository improvements which wants to start watching many channels and it is not tied to linker at all.

Naming wise it was a bit hard to read at first and to understand what exactly it does. We do use the term Active in some other places. ActiveChannelWatchingRequests. But up to you.

Sources/StreamChat/Workers/WatchingChannelsOngoingRequests.swift
```diff
@@ -0,0 +1,48 @@
...
+    private var ongoingWatchRequests: Set<ChannelId> = []
+    private let queue = DispatchQueue(
+        label: "io.getstream.WatchingChannelsOngoingRequests",
+        attributes: .concurrent
```
laevandus: Could be just more efficient serial queue as well but OK like this as well.

Sources/StreamChat/Workers/WatchingChannelsOngoingRequests.swift
nuno-vieira: > Did you also consider creating something like WatchChannelHandler which would internally do this tracking (naming wise we have some other XYZHandler types as well). Just for keeping the coalescing code in a single type. Moreover, that one could also be reused for SyncRepository improvements which wants to start watching many channels and it is not tied to linker at all.

I'm not sure I fully understand the goal of this handler as well 🤔 But the goal here was to do as minimal changes as possible. Also, the Linker exists per channel list, and it is not a single instance, if I refactored the code to have a `WatchChannelHandler` to do the synchronisation and also firing the requests, it would involve quite a bit of changes which I wanted to avoid for now.

> Naming wise it was a bit hard to read at first and to understand what exactly it does. We do use the term Active in some other places. ActiveChannelWatchingRequests. But up to you.

Yeah, the naming here was quite difficult. I'm open to suggestions 👍  The `ActiveChannelWatchingRequests` name can be misleading, because it looks like it keeps the state of which channels are being watched, but it is not true. This component only keeps the state of requests that are currently being executed.

nuno-vieira: Some other alternatives:
1. `ChannelWatchingExecutingRequests`
2. `ChannelWatchingActiveRequests`

Sources/StreamChat/Workers/WatchingChannelsOngoingRequests.swift
nuno-vieira: Yeah I actually prefer to have serial queues by default, and only use barrier if it needs better performance. AI added the concurrent queue and I left it :D But I can change it to serial to keep it is simpler

nuno-vieira: I changed it to serial in the new refactoring 👍

Sources/StreamChat/Workers/ChannelListLinker.swift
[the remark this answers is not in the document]
nuno-vieira: I solved it in a different way but thank you anyways :)

Sources/StreamChat/Workers/WatchingChannelsOngoingRequests.swift
nuno-vieira: I've actually gone with the Handler solution. 👍 Let me know what you think

PR #3876, opened 2025-11-12
[CI] Introduce Backend Integration Tests

testableapple:
## Test on CI

- ✅ https://github.com/GetStream/stream-chat-swift/actions/runs/19328524731

## Trigger on CI

- ✅ https://github.com/GetStream/stream-chat-swift/actions/workflows/backend-checks.yml


.github/workflows/backend-checks.yml
```diff
@@ -0,0 +1,41 @@
+name: Backend Checks
```
martinmitrevski, line 1: It will be possible for the backend team to run this workflow independently on demand, right?

testableapple: yes, https://github.com/GetStream/stream-chat-swift/actions/workflows/backend-checks.yml

PR #3878, opened 2025-11-14
Reduce SDK size by converting Events from structs to classes

nuno-vieira:
### 🔗 Issue Links

https://linear.app/stream/issue/IOS-1257/reduce-sdk-size-by-converting-events-from-struct-to-classes

### 🎯 Goal

Reduce SDK size by converting Events from structs to classes.

### 📝 Summary

For now, I converted all events to classes to check how much size we can reduce, but there are a couple of candidates that might be better to stick with structs:
- Typing Events
- Reaction Events

The above events can be constantly being created and destroyed, which could reduce SDK performance. Before merging, we need to benchmark the SDK with our performance scripts and see if there is any performance degradation.

### 🧪 Manual Testing Notes

Requires regression testing and retest with our benchmark scripts.

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **Refactor**
  * Converted many event types to reference-style classes and added explicit initializers for consistent construction across AI typing, channel, connection, member, notification, poll, reaction, thread, typing, and user events.
* **Performance**
  * Reduced SDK size through the event-model consolidation.
* **Changed**
  * Public event APIs changed; consumers may need to rebuild against the updated SDK.

### DIFF
--- a/Sources/StreamChat/WebSocketClient/Events/DraftEvents.swift
+++ b/Sources/StreamChat/WebSocketClient/Events/DraftEvents.swift
@@ -5,7 +5,7 @@
 import Foundation
 
 /// Triggered when a draft message is updated or created.
-public class DraftUpdatedEvent: Event {
+public final class DraftUpdatedEvent: Event {
     /// The channel identifier of the draft.
     public let cid: ChannelId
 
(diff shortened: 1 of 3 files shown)


martinmitrevski (approved):
Awesome!

PR #2734, opened 2023-08-16
Make sure loud speaker is always used for playback

polqf:
### 🔗 Issue Links

Resolves https://github.com/GetStream/ios-issues-tracking/issues/441

### 🎯 Goal

Make sure all voice messages are played using the loud speaker

### 📝 Summary

AudioSession does not allow overriding the speaker output unless the category is set to `.playAndRecord`. When we were tying to do so, an error was triggered, leaving the whole Audio Player in an inconsistent state which would result in all the voice messages being played from the earpiece.

### 🛠 Implementation

- Changed the category to `.playAndRecord` for playback.
- Removed unnecessary guards 

### 🧪 Manual Testing Notes

- Open message list
- Play audio
- Audio plays from Loud Speaker
- Record Audio Message
- Play any audio

Expected result:
- All audio is coming from the loud speaker

### ☑️ Contributor Checklist

- [x] I have signed the [Stream CLA](https://docs.google.com/forms/d/e/1FAIpQLScFKsKkAJI7mhCr7K9rEIOpqIDThrWxuvxnwUq2XkHyG154vQ/viewform) (required)
- [x] This change follows zero ⚠️ policy (required)
- [x] This change should be manually QAed
- [ ] Changelog is updated with client-facing changes
- [ ] New code is covered by unit tests
- [ ] Comparison screenshots added for visual changes
- [ ] Affected documentation updated (docusaurus, tutorial, CMS)

PR #3884, opened 2025-11-27
Fix audio recordings not using AirPods automatically

nuno-vieira:
### 🔗 Issue Links

https://linear.app/stream/issue/IOS-87/async-messages-audio-always-plays-from-speaker-even-if-airpods-are

### 🎯 Goal

- Fix audio recordings not playing from AirPods automatically.
- Fix audio recordings not using the mic from AirPods automatically.

#### AirPods automatic usage

This issue was introduced by the following PR: https://github.com/GetStream/stream-chat-swift/pull/2734, which forced the audio to always play from the speaker.

This PR changes it so that by default it uses the Speaker, but if there is any Bluetooth device connected, it uses that device.

**UPDATE:** I've also fixed the audio recordings not using the AirPods mic automatically.

#### Audio Session Configuration

We also need to update the Website docs on how to provide a custom audio session configuration:

**UIKit:**

```swift
final class CustomAudioPlayer: StreamAudioQueuePlayer {
    required convenience init() {
        self.init(
            assetPropertyLoader: StreamAssetPropertyLoader(),
            audioSessionConfigurator: CustomAudioSessionConfigurator()
        )
    }
}
Components.default.audioPlayer = CustomAudioPlayer.self
``` 

**SwiftUI:**

```swift
utils.audioPlayerBuilder = {
      let player = StreamAudioPlayer()
      player.configure(CustomAudioSessionConfigurator())
      return player
 }
 // or
 utils.audioPlayerBuilder = {
      StreamAudioPlayer(
           assetPropertyLoader: StreamAssetPropertyLoader(),
           audioSessionConfigurator: CustomAudioSessionConfigurator()
       )
 }
```

### 🧪 Manual Testing Notes

**Play a voice recording:**
AirPods connected? Plays from AirPods, otherwise plays from iPhone Speaker.

**Record audio:**
AirPods connected? Records from AirPods, otherwise records from iPhone mic.

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed audio recordings not playing from AirPods automatically and not using AirPods mic automatically.
* **Behavior**
  * Playback now defaults to speaker while still allowing Bluetooth headsets; recording/playback routing simplified.
* **Documentation**
  * Updated notes to reflect the revised routing behavior.
* **Tests**
  * Updated and reduced audio routing tests to match the new routing and activation behavior.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/Sources/StreamChat/Audio/AudioSessionConfiguring.swift
+++ b/Sources/StreamChat/Audio/AudioSessionConfiguring.swift
@@ -74,6 +74,8 @@ open class StreamAudioSessionConfigurator: AudioSessionConfiguring {
             mode: .spokenAudio,
             policy: .default,
             options: [
+                // It is deprecated, but for now we need to use it,
+                // since the newer ones are not available in Xcode 15.
                 .allowBluetooth
             ]
         )


Sources/StreamChat/Audio/AudioSessionConfiguring.swift
```diff
@@ -68,15 +68,17 @@ open class StreamAudioSessionConfigurator: AudioSessionConfiguring {
...
     open func activateRecordingSession() throws {
         try audioSession.setCategory(
             .playAndRecord,
             mode: .spokenAudio,
             policy: .default,
-            options: []
+            options: [
+                // It is deprecated, but for now we need to use it,
+                // since the newer ones are not available in Xcode 15.
+                .allowBluetooth
```
ipavlidakis, line 79: We have a workaround for that on Video that allows you to use allowBluetootHFP on Xcode 15.

```swift
#if !canImport(AVFoundation, _version: 2360.61.4.11)
    /// Older SDKs only expose ``allowBluetooth`` so we map the HFP alias to it
    /// to avoid peppering the codebase with availability checks.
    public static let allowBluetoothHFP = Self.allowBluetooth
#endif
```

Sources/StreamChat/Audio/AudioSessionConfiguring.swift
```diff
@@ -68,15 +68,17 @@ open class StreamAudioSessionConfigurator: AudioSessionConfiguring {
...
+            ]
         )
-        try setUpPreferredInput(.builtInMic)
```
ipavlidakis, line 79: That's just a preference and not really setting the mic to the builtIn (at least not always). You are right though, better not to have any preference here.

Sources/StreamChat/Audio/AudioSessionConfiguring.swift
```diff
@@ -90,16 +92,18 @@ open class StreamAudioSessionConfigurator: AudioSessionConfiguring {
...
-    /// ensure that the output port can be set to `.speaker`.
+    /// - Note: This method uses the `.playAndRecord` category with `.default` mode and policy.
     open func activatePlaybackSession() throws {
         try audioSession.setCategory(
             .playAndRecord,
             mode: .default,
             policy: .default,
-            options: []
+            options: [
+                .defaultToSpeaker,
```
ipavlidakis, line 102: The problem with the `.defaultToSpeaker` is that you cannot play audio from the earpiece which can be problematic(for reference, when setting the `.defaultToSpeaker` the earpiece isn't available at all, even from the control center output selection list).

Sources/StreamChat/Audio/AudioSessionConfiguring.swift
```diff
@@ -130,12 +134,10 @@ open class StreamAudioSessionConfigurator: AudioSessionConfiguring {
     // MARK: - Helpers
 
     private func activateSession() throws {
-        try audioSession.overrideOutputAudioPort(.speaker)
```
ipavlidakis, line 133: Rather than removing the override and setting categoryOption (which completely disables the phone's earpiece) we can consider, checking the audioSession.currentRoute and if it's an external one do not apply any override, otherwise we apply it. But if we are alright always playing audio from the speaker (if no external device) then this approach works.

Sources/StreamChat/Audio/AudioSessionConfiguring.swift
martinmitrevski: haha, not again this ugly code 😅 yeah, we can keep it, Ilias already convinced me for video.

martinmitrevski (approved):
LGTM ✅

Sources/StreamChat/Audio/AudioSessionConfiguring.swift
nuno-vieira: That is weird, since this should set only the "default", not force it to be the speaker always 🤔

Sources/StreamChat/Audio/AudioSessionConfiguring.swift
nuno-vieira: We are going to remove Xcode 15 support pretty soon, not sure if it is worth adding this 🤔

PR #3916, opened 2026-01-09
Fix crash when querying Channel List by `.filterTags`

nuno-vieira:
### 🔗 Issue Links

https://linear.app/stream/issue/IOS-1346

### 🎯 Goal

Fix crash when querying by filter tags.

### 🛠 Implementation

Using the `IN` operator against Transformable properties in CoreData is not very safe, and it seems to cause crashes. So this PR changes the filter tags to be a CoreData relationship instead of an array.

It is weird that the unit test does not catch this crash, and it is only reproducible after using the query only the first time. But with CoreData, I guess we never know what to expect 😄 

Either way, we should only use "Transformable" if the data is not meant to be queryable. If it is, then it is better to use a CoreData relationship.

### 🧪 Manual Testing Notes

Steps to repro on our Demo App:
1. Tap on the Filter icon top right corner
2. Select "Premium Channels"
3. Tap on the Filter icon top right corner again
4. Select "Initial Channels"
5. Tap on the Filter icon top right corner again
6. Select "Premium Channels"
7. 💥 

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a crash when querying the Channel List by filterTags.

* **Stability**
  * Improved handling and storage of channel tags for more reliable tag-based filtering.

* **Tests**
  * Updated tests to reflect the new tag handling and key-path mapping.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## StreamChat
 ### 🐞 Fixed
 - Possible fix for rare `thereIsNoSadnessLikeTheDeathOfOptimism` crash in CoreData [#3908](https://github.com/GetStream/stream-chat-swift/pull/3908)
+- Fix rare deadlock when reconnecting to the web-socket [#3913](https://github.com/GetStream/stream-chat-swift/pull/3913)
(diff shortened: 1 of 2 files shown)


nuno-vieira:
@martinmitrevski No, we don't have migrations for anything set up in CoreData at the moment. The DB data is always destroyed and re-created if the schema changes.

PR #3924, opened 2026-01-14
[CI][v5] Remove xcframeworks

testableapple:
### 🔗 Issue Links

Resolve https://linear.app/stream/issue/IOS-1337

### 🎯 Goal

Speed up release process and make it simpler by removing unused package managers

### 📝 Summary

- All scripts, lanes, artifacts related to xcframeworks were removed


Scripts/buildStaticLibraries.sh
```diff
@@ -1,142 +0,0 @@
-#!/usr/bin/env bash
-#
-# Usage: ./Scripts/buildStaticLibraries.sh
-#
-# This script creates static libraries for our libraries bundled in xcframeworks. 
-# It also creates .bundle files that need to be copied for the libraries to work
```
nuno-vieira, line 6: I guess, this means this ticket is no longer needed: https://linear.app/stream/issue/IOS-1329/v5-streamchatcommonui-add-static-target

Which means we can also remove the current static targets we have on the project?

testableapple: Removed static targets 👍

martinmitrevski (approved):

PR #3925, opened 2026-01-15
V5: Add token based color definitions

laevandus:
### 🔗 Issue Links

Related: [IOS-1328](https://linear.app/stream/issue/IOS-1328)

### 🎯 Goal

Copy over autogenerated colors

### 📝 Summary

- Copy over autogenerated colors
- There are many references to self in the generated code, therefore switch over to class type for colors for avoiding any mutability issues with the struct

### 🛠 Implementation

Just manually copying it over at the moment. Later when other colors are migrated to token colors, the whole file will be autogenerated. 

### ☑️ Contributor Checklist


DemoApp/Screens/Livestream/DemoLivestreamMessageActionsVC.swift
```diff
@@ -145,7 +145,7 @@ class DemoLivestreamMessageActionsVC: UIViewController {
         button.contentEdgeInsets = UIEdgeInsets(top: 8, left: 8, bottom: 8, right: 8)
 
         let isSelected = message?.currentUserReactions.contains { $0.type == reactionType } ?? false
-        let colorPalette = Appearance.default.colorPalette
+        var colorPalette = Appearance.default.colorPalette
```
laevandus: Because ColorPalette itself is struct at the moment and accentPrimary is lazy var now.

laevandus: OK, this makes it difficult to use even with `@Injected`

martinmitrevski (commented):
The SDK size increases a lot

laevandus:
> The SDK size increases a lot

Ah it is because it is comparing against develop where we don't have StreamChatCommonUI. Should compare against v5 ideally.

Sources/StreamChatCommonUI/Appearance+ColorPalette.swift
```diff
@@ -7,7 +7,7 @@ import SwiftUI
 import UIKit
 
 public extension Appearance {
-    struct ColorPalette: Sendable {
+    final class ColorPalette: @unchecked Sendable {
```
laevandus, line 10: Since I would like to use lazy var, struct makes it hard since then accessing lazy var requires colors to be mutable. 
```
@Injected(\.colors) var colors

…

colors.accentPrimary <- won't compile because colors is not mutable in this context
```

Ideally Appearance as a whole is @MainActor (it is currently `@unchecked Sendable` as well).

I am proposing to make the whole Appearance @MainActor in a separate ticket. WDYT?

nuno-vieira: Yes, for me it makes sense for Appearance to be main actor 👍

Sources/StreamChatCommonUI/Appearance+ColorPalette.swift
```diff
@@ -132,24 +130,206 @@ public extension Appearance {
...
+        public lazy var colorBrand200: UIColor = UIColor(light: .blue200, dark: .blue700)
+        public lazy var colorBrand300: UIColor = UIColor(light: .blue300, dark: .blue600)
+        public lazy var colorBrand400: UIColor = UIColor(light: .blue400, dark: .blue500)
+        public lazy var colorBrand50: UIColor = UIColor(light: .blue50, dark: .blue900)
+        public lazy var colorBrand500: UIColor = UIColor(light: .blue500, dark: .blue400)
+        public lazy var colorBrand600: UIColor = UIColor(light: .blue600, dark: .blue300)
+        public lazy var colorBrand700: UIColor = UIColor(light: .blue700, dark: .blue200)
+        public lazy var colorBrand800: UIColor = UIColor(light: .blue800, dark: .blue100)
+        public lazy var colorBrand900: UIColor = UIColor(light: .blue900, dark: .blue50)
+        public lazy var colorBrand950: UIColor = UIColor(light: .blue950, dark: .baseWhite)
```
nuno-vieira, line 275: All these variants feel like too much. Can't we simplify this on the design system? Not sure if this was discussed already?

laevandus: It is being discussed to reduce the amount of variables, but not sure when it lands. For now, we'll have a bit too many but it should get better over time.

Sources/StreamChatCommonUI/Generated/DesignSystemTokens.swift
```diff
@@ -0,0 +1,164 @@
...
+import Foundation
+
+public enum DesignSystemTokens {
+    // MARK: Foundations
+    static let fontFamilyGeist: String = "Geist"
+    static let fontFamilyGeistMono: String = "Geist Mono"
+    static let fontFamilyRoboto: String = "Roboto"
+    static let fontFamilyRobotoMono: String = "Roboto Mono"
+    static let fontFamilySfMono: String = "SF Mono"
+    static let fontFamilySfPro: String = "SF Pro"
```
nuno-vieira: On iOS, we are probably going to use only the SF font, right? Maybe for the tokens, we can just export the font sizes, since we are probably not going to use the font names directly, but just the system font

Sources/StreamChatCommonUI/Appearance+ColorPalette.swift
laevandus: Captured it [here](https://linear.app/stream/issue/IOS-1356/v5-make-appearance-mainactor)

martinmitrevski: yes, makes sense

Sources/StreamChatCommonUI/Generated/DesignSystemTokens.swift
martinmitrevski: yes, font families should be deleted. But also, why do we need font sizes? Shouldn't we rely on semantic values, like body, headline, title, etc?

laevandus: There is ongoing work to define [text styles](https://github.com/GetStream/design-system-tokens/blob/main/tokens/chat/primitives/ios.json#L308) in tokens. In this PR, I think will delete most of these and then in later PRs bring in text styles etc. Naming is pretty ugly here at the moment.

Sources/StreamChatCommonUI/Appearance+ColorPalette.swift
```diff
@@ -67,8 +67,6 @@ public extension Appearance {
         public var alternativeActiveTint: UIColor = .streamAccentGreen
         public var inactiveTint: UIColor = .streamGray
         public var alternativeInactiveTint: UIColor = .streamGrayGainsboro
-
-        public var accentPrimary: UIColor = .streamAccentPrimary
```
laevandus, line 71: Note that accent slightly changes so snapshots needed to be updated.

PR #3926, opened 2026-01-15
Fix scaling of Fonts in SwiftUI

martinmitrevski:
### 🔗 Issue Links

https://linear.app/stream/issue/IOS-1339/v5-streamchatcommonui-fix-dynamic-font-size-in-swiftui

### 🎯 Goal

### 📝 Summary

### 🛠 Implementation

### 🎨 Showcase

| Before | After |
| ------ | ----- |
|  img   |  img  |

### 🧪 Manual Testing Notes

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChatCommonUI/Appearance+Fonts.swift
+++ b/Sources/StreamChatCommonUI/Appearance+Fonts.swift
@@ -39,9 +39,8 @@ public extension Appearance {
         public var headlineBold: Font = .headline.bold()
         public var title: Font = .title
 
-        // Backing storage (optional overrides)
-        private var _title2: Font? = nil
-        private var _title3: Font? = nil
+        private var _title2: Font?
+        private var _title3: Font?
 
         // Publicly mutable properties
         public var title2: Font {


Sources/StreamChatCommonUI/Appearance+Fonts.swift
```diff
@@ -25,42 +25,42 @@ public extension Appearance {
...
+            set { _title2 = newValue }
+        }
+
+        public var title3: Font {
+            get {
+                if let v = _title3 { return v }
+                if #available(iOS 14.0, *) { return .title3 }
+                return .title
+            }
+            set { _title3 = newValue }
```
nuno-vieira, line 61: Why do we need this? 🤔 Why they can't be directly mutable?

martinmitrevski: title2 and title3 are iOS 14 and above, the common module is iOS 13+ 😭

Sources/StreamChatCommonUI/Appearance.swift
```diff
@@ -18,7 +18,7 @@ public struct Appearance: @unchecked Sendable {
     public var fonts = Fonts()
     
     /// SwiftUI representation of the fonts.
-    public var fontsSwiftUI: FontsSwiftUI
+    public var fontsSwiftUI = FontsSwiftUI()
```
nuno-vieira, line 21: This means SwiftUI will use `fontsSwiftUI` directly right, not `fonts`? The bad thing is that we will need documentation to say that if you use SwiftUI, you will need to this one instead of `fonts`. 🤔 

What if we make `Fonts` an enum? Maybe it is less likely to make mistakes since you can only have one font config. Example:

```swift
enum Fonts {
  case uiFont(FontsUIKit)
  case font(FontsSwiftUI)
}
```

martinmitrevski: I tried that, but a lot of build errors to fix and some casting to do. Not sure if it's worth it

martinmitrevski: btw, SwiftUI handles that internally, so customers will use fonts like before (no changes are needed in SwiftUI SDK)

PR #3934, opened 2026-01-26
V5: Update tokens and add them to Appearance

laevandus:
### 🔗 Issue Links

Related: [IOS-1328](https://linear.app/stream/issue/IOS-1328)

### 🎯 Goal

Sync tokens and add them to Appearance

### 📝 Summary

- Appearance gets designTokens property for easy access
- `Bg` to `Background`
- Drop `color` from accent colors

### ☑️ Contributor Checklist


Sources/StreamChatCommonUI/Appearance.swift
```diff
@@ -31,6 +31,9 @@ public struct Appearance: @unchecked Sendable {
     /// you can change how data is formatted to textual representation.
     public var formatters = Formatters()
     
+    /// A set of tokens defining the rules for layout.
+    public let designTokens = DesignSystemTokens()
```
laevandus: In SwiftUI we can then expose it through Injected making it nicer. In the future we can consider making them mutable (kept them immutable for now)

Sources/StreamChatCommonUI/Appearance+ColorPalette.swift
```diff
@@ -179,131 +182,125 @@ public extension Appearance {
...
+        public lazy var brand150: UIColor = UIColor(light: .blue150, dark: .blue700)
+        public lazy var brand200: UIColor = UIColor(light: .blue200, dark: .blue700)
+        public lazy var brand300: UIColor = UIColor(light: .blue300, dark: .blue600)
+        public lazy var brand400: UIColor = UIColor(light: .blue400, dark: .blue500)
+        public lazy var brand50: UIColor = UIColor(light: .blue50, dark: .blue900)
+        public lazy var brand500: UIColor = UIColor(light: .blue500, dark: .blue400)
+        public lazy var brand600: UIColor = UIColor(light: .blue600, dark: .blue300)
+        public lazy var brand700: UIColor = UIColor(light: .blue700, dark: .blue200)
+        public lazy var brand800: UIColor = UIColor(light: .blue800, dark: .blue150)
+        public lazy var brand900: UIColor = UIColor(light: .blue900, dark: .blue100)
```
nuno-vieira: We should probably not expose this directly, but instead be private, but out of scope of this PR 👍

Sources/StreamChatCommonUI/Appearance.swift
```diff
@@ -31,6 +31,9 @@ public struct Appearance: @unchecked Sendable {
     /// you can change how data is formatted to textual representation.
     public var formatters = Formatters()
     
+    /// A set of tokens defining the rules for layout.
+    public var designTokens = DesignSystemTokens()
```
nuno-vieira:
```suggestion
    public var tokens = DesignSystemTokens()
```

Maybe tokens is enough 🤔 Since it is already inside "Appearance" object, so we already know "tokens" are related to Appearance / Design.

This way it is also more consistent with the other Injected properties (colors, images, fonts, etc..)

Sources/StreamChatCommonUI/Appearance+ColorPalette.swift
laevandus: I will delete these manually for now.

PR #3945, opened 2026-01-30
[v5] Convert Xcode groups to buildable folders

testableapple:
Follow up https://github.com/GetStream/stream-chat-swift/pull/3940


nuno-vieira (approved):
🚀

PR #4029, opened 2026-03-30
Fix pin/unpin updates not propagated in LivestreamChannelController for old messages

nuno-vieira:
### 🔗 Issue Links

[IOS-1553](https://linear.app/stream/issue/IOS-1553/patreonllc-no-channel-update-received-for-old-pinned-messages-in)

### 🎯 Goal

Fix pin/unpin updates not being propagated through `channelChangePublisher` in `LivestreamChannelController` when the pinned message is not in the local messages list (e.g., old messages that haven't been paginated to).

### 📝 Summary

- Update `handleUpdatedMessage` in `LivestreamChannelController` to also update `channel.pinnedMessages` when the message is not in the local messages array
- Extract shared pin/unpin logic into a reusable `updatePinnedMessages(for:)` method

### 🛠 Implementation

Previously, `handleUpdatedMessage` only updated `channel.pinnedMessages` if the message existed in the local `messages` array. For old messages that haven't been paginated into the list, pin/unpin events were silently ignored, so the channel's `pinnedMessages` was never updated and `channelChangePublisher` never fired.

The fix adds an `else` branch that checks whether the updated message is newly pinned or was previously in `pinnedMessages` (now unpinned), and updates accordingly. The pin/unpin logic is extracted into `updatePinnedMessages(for:)` to avoid duplication.

### 🧪 Manual Testing Notes

1. Open a livestream channel with `LivestreamChannelController`
2. Pin an old message that is not currently in the paginated messages list
3. Verify that `channel.pinnedMessages` is updated in the Channel Info debugger
4. Unpin the same message and verify the update propagates

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/Controllers/ChannelController/LivestreamChannelController.swift
+++ b/Sources/StreamChat/Controllers/ChannelController/LivestreamChannelController.swift
@@ -1116,15 +1116,23 @@ public class LivestreamChannelController: DataStoreProvider, EventsControllerDel
             messages[index] = updatedMessage
 
             if existingMessage.isPinned != updatedMessage.isPinned {
-                var pinnedMessages = channel?.pinnedMessages ?? []
-                if updatedMessage.isPinned {
-                    pinnedMessages.append(updatedMessage)
-                } else {
-                    pinnedMessages.removeAll(where: { $0.id == existingMessage.id })
-                }
-                channel = channel?.changing(pinnedMessages: pinnedMessages)
+                updatePinnedMessages(for: updatedMessage)
             }
+        } else if updatedMessage.isPinned || channel?.pinnedMessages.contains(where: { $0.id == updatedMessage.id }) == true {
+            updatePinnedMessages(for: updatedMessage)
         }
+    }
+
+    private func updatePinnedMessages(for message: ChatMessage) {
+        var pinnedMessages = channel?.pinnedMessages ?? []
+        if message.isPinned {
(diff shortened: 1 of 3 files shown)


martinmitrevski (approved):
LGTM ✅ We should also cherry pick it in develop

Sources/StreamChat/Controllers/ChannelController/LivestreamChannelController.swift
```diff
@@ -1116,15 +1116,23 @@ public class LivestreamChannelController: DataStoreProvider, EventsControllerDel
...
+        } else if updatedMessage.isPinned || channel?.pinnedMessages.contains(where: { $0.id == updatedMessage.id }) == true {
+            updatePinnedMessages(for: updatedMessage)
+        }
+    }
+
+    private func updatePinnedMessages(for message: ChatMessage) {
+        var pinnedMessages = channel?.pinnedMessages ?? []
+        if message.isPinned {
+            if !pinnedMessages.contains(where: { $0.id == message.id }) {
+                pinnedMessages.append(message)
```
laevandus, line 1130: OK to not keep it sorted since loading pinned messages has sorting parameter?

nuno-vieira: This was not sorted before right? 🤔 Sorting the pinned messages is important when using it as a channel controller not channel list, or am I missing something?

PR #4056, opened 2026-04-08
Unified CDN & Media Loading Architecture (v5)

nuno-vieira:
### 🎯 Goal

Consolidate and simplify all CDN and media loading logic into unified protocols shared between UIKit and SwiftUI SDKs, with async support for pre-signed URLs.

### 📝 Summary

- Add `CDNRequester` protocol for read-only URL transformation (signing, headers, resize) with async completion handler support
- Add `CDNStorage` protocol replacing both `CDNClient` and `AttachmentUploader` for upload/delete operations
- Add `CDNRequest` value type carrying URL, headers, and caching key
- Add `CDNImageResize` model for server-side image resize parameters
- Add `StreamCDNRequester` default implementation with Stream CDN resize/cache logic
- Add `StreamCDNStorage` default implementation using Stream's REST API for multipart uploads
- Move `ImageResize` from StreamChatUI to StreamChatCommonUI
- Add unified `MediaLoader` protocol in StreamChatCommonUI replacing separate `ImageLoader` and `VideoLoader`
- Add `StreamMediaLoader` default implementation handling both image loading and video preview generation
- Add `ImageDownloading` protocol so CommonUI never depends on Nuke directly
- Add `StreamImageDownloader` as the UIKit `ImageDownloading` backend
- Add Options pattern types (`ImageLoadOptions`, `ImageBatchLoadOptions`, `VideoLoadOptions`) for extensible method signatures
- Add `cdnRequester` and `cdnStorage` as public mutable properties on `ChatClient`
- Update `ChatMessageController.downloadAttachment()` to use `ChatClient.cdnRequester` internally for URL transformation
- Remove old `ImageCDN`, `StreamImageCDN`, `ImageLoading`, `NukeImageLoader`, `VideoLoading`, `CDNClient`, `AttachmentUploader`
- Update `Components` to use `mediaLoader: MediaLoader` (replacing separate `imageLoader` + `videoLoader`)
- Update `ImageLoaderOptions` and `ImageDownloadOptions` to include `cdnRequester` in the options struct
- Update `ChatClientConfig` with `cdnRequester` and `cdnStorage` replacing `customCDNClient` and `customAttachmentUploader`

### 🛠 Implementation

The architecture splits CDN concerns into **read** and **write** paths with unified protocols replacing scattered types across both repos:

| Protocol | Module | Responsibility |
|----------|--------|----------------|
| `CDNRequester` | StreamChat | Read-side URL transformation (signing, headers, resize) before loading images/files |
| `CDNStorage` | StreamChat | Write-side attachment upload and delete operations |
| `MediaLoader` | StreamChatCommonUI | Unified image downloading, caching, batch loading, video preview generation |
| `ImageDownloading` | StreamChatCommonUI | Pluggable HTTP download + cache backend (e.g. Nuke) |

**CDNRequester vs CDNStorage:**

- `CDNRequester` handles the **read path** — transforming URLs before GET requests (adding resize query params, signing, injecting headers, computing cache keys). The default `StreamCDNRequester` adds width/height/resize query params for `stream-io-cdn.com` URLs and builds stable caching keys.
- `CDNStorage` handles the **write path** — uploading attachments (multipart, with progress) and deleting them by remote URL. The default `StreamCDNStorage` uses Stream's REST API. Configured via `ChatClientConfig.cdnStorage` (optional; nil uses the default).
- Both `cdnRequester` and `cdnStorage` are public mutable properties on `ChatClient`, initialized from `ChatClientConfig` but changeable at runtime.

**Unified MediaLoader:**

`MediaLoader` merges the previous `ImageLoader` and `VideoLoader` into a single protocol. This eliminates a stale-reference problem: previously, `StreamVideoLoader` held a reference to `ImageLoader` at init time, so replacing `components.imageLoader` wouldn't affect video thumbnail loading. Now there's one `mediaLoader` property in `Components` (UIKit) and `Utils` (SwiftUI).

The protocol uses an **Options pattern** — `ImageLoadOptions`, `ImageBatchLoadOptions`, and `VideoLoadOptions` struct parameters — so the `CDNRequester` is passed on every call rather than captured at init. This means changing `chatClient.cdnRequester` at runtime takes effect immediately. The Options pattern also makes the protocol extensible: adding future parameters requires only adding a field with a default value, not changing the method signature.

`StreamMediaLoader` handles both image loading (delegates to `ImageDownloading` backend) and video preview generation (AVFoundation). Video thumbnails are cached in an `NSCache`.

**Attachment downloads:**

`ChatMessageController.downloadAttachment()` and `Chat.downloadAttachment()` no longer expose a `remoteURL` parameter. Instead, they internally use `ChatClient.cdnRequester.fileRequest()` to transform the attachment URL before passing it to the download worker.

### 🧪 Manual Testing Notes

1. Open the Demo App
2. Verify images load correctly in channel list (avatars) and message list (image/video attachments)
3. Verify video previews load correctly
4. Test file attachment downloads
5. Test file attachment previews

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **New Features**
  * CDN image resize parameters and signed-URL support for downloads.
  * Unified MediaLoader handling images, video previews and assets (async APIs).

* **Improvements**
  * CDN-backed storage for attachment uploads/deletes with progress and better error handling.
  * More reliable thumbnail/video preview generation, caching, and deallocation safety.
  * Simplified avatar rendering and consistent composer attachment size fallback.
  * downloadAttachment now accepts an optional remoteURL for CDN workflows.


Sources/StreamChat/APIClient/CDNClient/CDNClient.swift
```diff
@@ -134,20 +54,31 @@ final class StreamCDNClient: CDNClient, @unchecked Sendable {
...
+                file: try .init(url: localUrl)
+            )
+        } catch {
+            completion(.failure(error))
+            return
+        }
+
+        guard let fileData = try? Data(contentsOf: localUrl) else {
             return completion(.failure(ClientError.Unknown()))
         }
```
laevandus: Not sure if this runs on main or background thread already. Thinking about possible slowness when the file is huge.
Or `Data(contentsOf:options: .mappedIfSafe)` for reducing the memory usage.

Sources/StreamChatCommonUI/ImageLoading/StreamVideoLoader.swift
```diff
@@ -0,0 +1,132 @@
...
+
+    open func loadPreview(
+        with attachment: ChatMessageVideoAttachment,
+        completion: @escaping @MainActor (Result<UIImage, Error>) -> Void
+    ) {
+        let videoURL = attachment.videoURL
+        if let cached = cache.object(forKey: videoURL as NSURL) {
+            Task { @MainActor in
+                completion(.success(cached))
+            }
```
laevandus: Thinking if `StreamConcurrency.onMain` could be a better fit since `Task` will run on the next runloop cycle if we are on the main, while onMain is synchronous if already on main.

Sources/StreamChatUI/Utils/ImageLoading/NukeImageDownloader.swift
```diff
@@ -0,0 +1,48 @@
+//
+// Copyright © 2026 Stream.io Inc. All rights reserved.
+//
+
+import StreamChatCommonUI
+import UIKit
+
+/// Nuke-backed implementation of ``ImageDownloading`` for the UIKit SDK.
+open class NukeImageDownloader: ImageDownloading, @unchecked Sendable {
```
laevandus: If we would drop Nuke for image loading, then I guess using `StreamImageDownloader` would be better.

Sources/StreamChat/APIClient/CDNClient/CDNRequest.swift
```diff
@@ -0,0 +1,22 @@
+//
+// Copyright © 2026 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+/// The result of a CDN URL transformation, containing the final URL,
+/// optional HTTP headers, and an optional caching key.
+public struct CDNRequest: Sendable {
```
martinmitrevski: maybe final class instead?

Sources/StreamChatUI/Utils/ImageLoading/ImageLoader+UIKit.swift
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    /// Downloads multiple images and returns all results.
+    public func downloadMultipleImages(
+        with requests: [ImageDownloadRequest],
+        completion: @escaping @MainActor ([Result<UIImage, Error>]) -> Void
+    ) {
+        let group = DispatchGroup()
+
+        final class BatchResult: @unchecked Sendable {
```
martinmitrevski: maybe we extract this internal to the file, instead of the method?

Sources/StreamChatUI/Utils/ImageLoading/NukeImageDownloader.swift
martinmitrevski: My suggestion would be to rename this one and hide the nuke reference, so we can just replace the implementation when we want.

martinmitrevski: Also, why's it only used on UIKit? What do we have on SwiftUI?
Alternatively, if it's used only on UIKit, we move this code in the UIKit SDK?

martinmitrevski (commented):
Looks good! We should make sure we test this extensively, since it's an important part of the SDK.

Sources/StreamChatUI/Utils/ImageLoading/NukeImageDownloader.swift
nuno-vieira: Since we need Nuke for image loading, initially, I moved Nuke to the common module, but we would still need nuke for other stuff in the UIKit and SwiftUI SDKs, so I just created a "ImageLoading" interface, and now both UIKit and SwiftUI inject it to the common module. They both should have this

Sources/StreamChat/APIClient/CDNClient/CDNRequester.swift
```diff
@@ -0,0 +1,98 @@
...
+        try await withCheckedThrowingContinuation { continuation in
+            fileRequest(for: url, options: options) { result in
+                continuation.resume(with: result)
+            }
+        }
+    }
+}
+
+/// Options for an image request through the CDN.
+public struct ImageRequestOptions: Sendable {
```
martinmitrevski, line 71: maybe we should use final class instead? Here and multiple other places.

Sources/StreamChatCommonUI/ImageLoading/StreamMediaLoader.swift
```diff
@@ -0,0 +1,230 @@
...
+
+            let resize: ImageResize? = options.loadThumbnails ? ImageResize(options.thumbnailSize) : nil
+            let imageOptions = ImageLoadOptions(resize: resize, cdnRequester: options.cdnRequester)
+            loadImage(url: avatarUrl, options: imageOptions) { result in
+                switch result {
+                case let .success(loaded):
+                    batchResult.images.append(loaded)
+                case .failure:
+                    if !options.placeholders.isEmpty {
+                        let placeholderIndex = index % options.placeholders.count
```
martinmitrevski: why %? Maybe a simple bounds check instead? Anyway, not a big deal.

Sources/StreamChatUI/Components.swift
```diff
@@ -679,7 +680,10 @@ import UIKit
     /// The router responsible for presenting alerts.
     public var alertsRouter: AlertsRouter.Type = AlertsRouter.self
 
-    public init() {}
-
+    public init() {
+        self.cdnRequester = StreamCDNRequester()
```
martinmitrevski: why not initialize these in the declaration, like the rest of the properties in Components?

martinmitrevski (approved):
LGTM! ✅  Left few small things, only bigger one I would consider is using final classes instead of structs in the options / requests. Size has grown a bit, which I think we can reduce.

Sources/StreamChatUI/Components.swift
nuno-vieira: This is left from old code, will clean up some stuff

Sources/StreamChatCommonUI/ImageLoading/StreamMediaLoader.swift
nuno-vieira: I'll check this, not sure if this old or new code

Sources/StreamChat/APIClient/CDNClient/CDNRequester.swift
nuno-vieira: Yeah I will try to make them class to see if there is any improvement. But I don't think so because the properties are pretty much raw types, but lets see

Sources/StreamChatCommonUI/ImageLoading/StreamMediaLoader.swift
nuno-vieira: Yes this is to avoid the bound check not a big deal, actually looks clean

Sources/StreamChatCommonUI/ImageLoading/MediaLoader.swift
```diff
@@ -0,0 +1,180 @@
...
+/// A unified protocol for loading images and video previews.
+///
+/// Merges the responsibilities of image loading and video preview generation
+/// into a single protocol, eliminating stale-reference problems when customers
+/// replace just one of the two loaders.
+///
+/// Configuration is passed via options structs on every call, so concrete
+/// implementations remain stateless with respect to CDN configuration.
+/// Changing the requester on `ChatClientConfig` takes effect immediately without
+/// recreating the loader.
```
nuno-vieira, line 14:
```suggestion
/// A unified protocol for loading images and video previews.
///
/// Configuration is passed via options structs on every call, so concrete
/// implementations remain stateless with respect to CDN configuration.
/// Changing the requester on `ChatClientConfig` takes effect immediately without
/// recreating the loader.
```

Sources/StreamChatCommonUI/ImageLoading/MediaLoader.swift
```diff
@@ -0,0 +1,180 @@
...
+    ///
+    /// - Parameters:
+    ///   - url: The video URL.
+    ///   - options: Options controlling CDN behavior.
+    ///   - completion: A completion handler called on the main actor with the preview image.
+    func loadVideoPreview(
+        at url: URL,
+        options: VideoLoadOptions,
+        completion: @escaping @MainActor (Result<MediaLoaderVideoPreview, Error>) -> Void
+    )
```
nuno-vieira: This is not needed

Sources/StreamChatCommonUI/ImageLoading/MediaLoader.swift
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    public func loadVideoPreview(
+        with attachment: ChatMessageVideoAttachment,
+        options: VideoLoadOptions,
+        completion: @escaping @MainActor (Result<MediaLoaderVideoPreview, Error>) -> Void
+    ) {
+        loadVideoPreview(at: attachment.videoURL, options: options, completion: completion)
+    }
+}
```
nuno-vieira: Remove this

Sources/StreamChatCommonUI/ImageLoading/StreamMediaLoader.swift
```diff
@@ -0,0 +1,211 @@
...
+/// The default ``MediaLoader`` implementation.
+///
+/// Delegates URL transformation to the ``CDNRequester`` provided via options,
+/// image downloading to an ``ImageDownloading`` backend (typically Nuke),
+/// and video preview generation to AVFoundation.
+open class StreamMediaLoader: MediaLoader, @unchecked Sendable {
+    /// The backend that performs the actual image download and caching.
+    public let downloader: ImageDownloading
+
+    private let videoPreviewCache: NSCache<NSURL, UIImage>
```
nuno-vieira, line 20: Do we need this still?

Sources/StreamChatUI/Utils/ImageProcessor/StreamImageProcessor.swift
```diff
@@ -21,7 +21,7 @@ public protocol ImageProcessor: Sendable {
 }
 
 /// This class provides resizing operations for `UIImage`. It internally uses `Nuke` porcessors to implement operations on images.
-open class NukeImageProcessor: ImageProcessor, @unchecked Sendable {
+open class StreamImageProcessor: ImageProcessor, @unchecked Sendable {
```
nuno-vieira, line 24: Is this used?

Sources/StreamChatUI/Components.swift
```diff
@@ -90,17 +90,18 @@ import UIKit
     /// The view that shows a loading indicator.
     public var loadingIndicator: ChatLoadingIndicator.Type = ChatLoadingIndicator.self
 
-    /// Object with set of function for handling images from CDN
-    public var imageCDN: ImageCDN = StreamImageCDN()
+    /// The CDN requester used for URL transformation (signing, headers, resizing).
+    public var cdnRequester: CDNRequester = StreamCDNRequester()
```
nuno-vieira, line 94: This should only live in ChatClientConfig

Sources/StreamChatUI/Components.swift
```diff
@@ -90,17 +90,18 @@ import UIKit
...
 
-    /// Object which is responsible for loading images
-    public var imageLoader: ImageLoading = NukeImageLoader()
+    /// The object responsible for loading images and video previews.
+    public var mediaLoader: MediaLoader = StreamMediaLoader(downloader: StreamImageDownloader())
 
     /// Object responsible for providing resizing operations for `UIImage`
-    public var imageProcessor: ImageProcessor = NukeImageProcessor()
+    public var imageProcessor: ImageProcessor = StreamImageProcessor()
```
nuno-vieira, line 100: Do we need this?

PR #4065, opened 2026-04-13
Move UIKit localizations to respective module

laevandus:
### 🔗 Issue Links

 [IOS-1619: v5 separate localization files for UIKit and SwiftUI](https://linear.app/stream/issue/IOS-1619/v5-separate-localization-files-for-uikit-and-swiftui)

  ### 🎯 Goal

UIKit only loc keys should not be part of common

  ### 📝 Summary

  - Split `Localizable.strings` / `Localizable.stringsdict`: CommonUI keeps the 16 keys it actually consumes (`channel.name.*`, `dates.time-ago-*`,
  `polls.date.{days,weeks}-ago`, `recording.presentation.name`); everything else moves to a new `Sources/StreamChatUI/Resources/en.lproj/`.
  - `L10n` in both modules is now **`internal`** (breaking change — was `public` in CommonUI). Each module has its own SwiftGen pipeline (`.swiftgen.yml` +
  `L10n_template.stencil` + generated `L10n.swift`).
  - New `Bundle.streamChatUI` accessor in `StreamChatUI/Utils/Bundle+Extensions.swift`, mirroring the existing `Bundle.streamChatCommonUI` pattern. `Bundle.streamChatCommonUI`
  is retained for back-compat.
  - `Components.init()` now sets `Appearance.bundle = .streamChatUI`, mirroring `stream-chat-swiftui`'s `StreamChat.init` pattern.
  - `Appearance.localizationProvider` default now cascades: **CommonUI bundle → `Appearance.bundle` → `Bundle.main`**.
  - `Package.swift`: `StreamChatUI` target gains `resources: [.process("Resources")]` and excludes its `Generated/L10n_template.stencil`.
  - `StreamChat.xcodeproj`: `PBXFileSystemSynchronizedBuildFileExceptionSet`s updated so the new UI strings/stringsdict and generated files belong to the correct targets. The
  `SwiftGen` build phase now also runs against `Sources/StreamChatUI/.swiftgen.yml`.

  ### 🛠 Implementation

  **Key partition.** I grepped every `L10n.*` reference under `Sources/StreamChatCommonUI/` and `Sources/StreamChatUI/` (excluding the generated `L10n.swift` itself) and
  resolved each to its underlying string key. Only 16 keys are referenced from inside CommonUI — all in `DateUtils`, `ChatChannelNamer`, `PollTimestampFormatter`,
  `AudioRecordingNameFormatter`. Everything else (~125 keys) was moved to StreamChatUI. There is no accidental overlap: UI never references a CommonUI-only key and vice versa.

  **Lookup cascade.** `Appearance.localizationProvider` now reads:

  ```swift
  public var localizationProvider: @Sendable (_ key: String, _ table: String) -> String = { key, table in
      let commonBundle = Bundle.streamChatCommonUI
      let value = commonBundle.localizedString(forKey: key, value: nil, table: table)
      if value != key { return value } // Foundation returns the key itself on miss (value: nil)
      let fallback = Appearance.bundle ?? .main
      return fallback.localizedString(forKey: key, value: nil, table: table)
  }
  ```

  Both generated `L10n` enums route through this provider via `Appearance.default.localizationProvider(key, table)`, so the same code path serves both catalogs. CommonUI's
  `Appearance` cannot directly reference `Bundle.streamChatUI` (would be a layer violation), so `Components.init()` registers it as the injected bundle on first use — same trick
   `stream-chat-swiftui` already uses at `Sources/StreamChatSwiftUI/StreamChat.swift:20`.

  **SwiftGen.** Each module now owns its own `.swiftgen.yml`, `L10n_template.stencil`, and `Generated/L10n.swift`. CommonUI's template had `{% set accessModifier %}public{%
  endset %}` flipped to `internal`; UI's template is a copy with `BundleToken.bundle = .streamChatUI`. The existing `SwiftGen` run-script build phase on the CommonUI target now
  invokes both configs sequentially so a single build regenerates both catalogs.

  **Xcode project.** `Sources/` is a `PBXFileSystemSynchronizedRootGroup` referenced by three exception sets (one per target: `StreamChat`, `StreamChatUI`,
  `StreamChatCommonUI`). Added `/Localized: StreamChatUI/Resources/Localizable.strings`, `Localizable.stringsdict`, plus `StreamChatUI/Generated/L10n.swift`,
  `Generated/L10n_template.stencil`, `Utils/Bundle+Extensions.swift`, and `.swiftgen.yml` to the appropriate exception sets so the new files attach to the right targets.

  **`stream-chat-swiftui`.** Verified it only depends on `StreamChatCommonUI`, has its own independent `L10n`/`Localizable.strings`, never references
  `Bundle.streamChatCommonUI`, and already sets `Appearance.bundle = .streamChatUI` (its own private accessor) — nothing in that repo needs to change.

### 🎨 Showcase

| Before | After |
| ------ | ----- |
|  img   |  img  |

### 🧪 Manual Testing Notes

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **New Features**
  * Added a complete set of StreamChatUI English localizations, pluralization rules, and a bundle resolver to surface UI strings reliably.
  * Components now initialize the UI bundle for consistent resource resolution.

* **Bug Fixes**
  * Improved localization fallback: cascades through injected, UI, then main bundles to reduce missing translations.

* **Chores**
  * Updated resource processing and codegen configuration; removed duplicate/obsolete localization entries from the common UI.


martinmitrevski (commented):
Basically, there are strings that common needs, right? Not sure if this is the best way to go, maybe we keep things as they are?
also @nuno-vieira

PR #4071, opened 2026-04-20
Grouped channels endpoint

martinmitrevski:
### 🔗 Issue Links

Resolves https://linear.app/stream/issue/IOS-1635/support-for-grouped-channels-endpoint.

### 🎯 Goal

- Add `ChatChannelListController.prefill(channels:completion:)` for priming controller-local channel data before the first synchronize call while preserving normal pagination, observation, and offline refresh behavior
- Add `ChatClient.groupedQueryChannels(limit:watch:presence:)` to fetch grouped channel groups as `GroupedChannels`, preserving backend group keys and exposing normalized per-group channels and unread counts for integrators
- Add optional `groupedUnreadChannels` data to grouped unread websocket events and persist it on `CurrentChatUser` for integrators

### 📝 Summary

Test on https://github.com/GetStream/GroupedChannelsSample.

### 🛠 Implementation

### 🎨 Showcase

| Before | After |
| ------ | ----- |
|  img   |  img  |

### 🧪 Manual Testing Notes

### ☑️ Contributor Checklist


Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift
```diff
@@ -69,6 +69,8 @@ public class ChatChannelListController: DataController, DelegateCallable, DataSt
 
     /// A Boolean value that returns whether pagination is finished
     public private(set) var hasLoadedAllPreviousChannels: Bool = false
+    private var loadedChannelsCount = 0
```
laevandus: Thinking about why we need extra state here because it will get out of sync when channel linker inserts or removes channels for this query. Why can't we use `channels.count`. Prefill saves channels for query so DB observer should reflect that.

Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift
```diff
@@ -194,6 +209,38 @@ public class ChatChannelListController: DataController, DelegateCallable, DataSt
...
+    ///
+    /// The prefetched channels are persisted in the local storage and linked only to this
+    /// controller query, so pagination, local observation and offline refresh keep working.
+    public func prefill(
+        channels: [ChatChannel],
+        completion: ((Error?) -> Void)? = nil
+    ) {
+        let prefilledChannels = filter.map { runtimeFilter in
+            channels.filter(runtimeFilter)
+        } ?? channels
```
laevandus: If this is the first snapshot returned by backend, do we need to do this?

Sources/StreamChat/ChatClient.swift
```diff
@@ -487,6 +487,49 @@ public class ChatClient {
...
+                    try session.saveCurrentUserGroupedUnreadChannels(groupedUnreadChannels)
+                }
+                databaseContainer.write(converting: { session in
+                    try Self.groupedChannels(from: payload, session: session)
+                }, completion: completion)
+            case let .failure(error):
+                completion(.failure(error))
+            }
+        }
+    }
```
laevandus: File organisation:
Would be nice to move these new functions before // MARK: - Upload attachments and have a separate // MARK for these two functions. At the moment it is in the middle of connect and disconnect.

I would move function implementations to ChannelListUpdater as well to keep the file more compact.

Sources/StreamChat/ChatClient.swift
```diff
@@ -487,6 +487,49 @@ public class ChatClient {
...
+        apiClient.request(endpoint: endpoint) { [databaseContainer] result in
+            switch result {
+            case let .success(payload):
+                databaseContainer.write { session in
+                    let groupedUnreadChannels = payload.groups.mapValues(\.unreadChannels)
+                    try session.saveCurrentUserGroupedUnreadChannels(groupedUnreadChannels)
+                }
+                databaseContainer.write(converting: { session in
+                    try Self.groupedChannels(from: payload, session: session)
+                }, completion: completion)
```
laevandus: Would be better to have a single write there because currently errors from the first one are ignored,

Sources/StreamChat/ChatClient.swift
```diff
@@ -833,6 +876,27 @@ extension ChatClient: ConnectionDetailsProviderDelegate {
...
+            return GroupedChannelsGroup(
+                channels: channels,
+                unreadChannels: groupPayload.unreadChannels
+            )
+        }
+
+        return GroupedChannels(
+            groups: groups
+        )
+    }
```
laevandus: This should be part of the `DatabaseSession`/`ChannelListUpdater` to keep things organised.

Sources/StreamChat/ChatClient.swift
```diff
@@ -847,6 +911,41 @@ extension ChatClient {
...
+        let derivedUnreadChannels = channels.reduce(into: 0) { partialResult, channel in
+            if channel.unreadCount.messages > 0 {
+                partialResult += 1
+            }
+        }
+
+        self.unreadChannels = max(unreadChannels, derivedUnreadChannels)
+    }
+}
+
```
laevandus: These should also be moved to `Models` folder for keeping `ChatClient.swift` shorter,

Sources/StreamChat/ChatClient.swift
```diff
@@ -487,6 +487,49 @@ public class ChatClient {
         authenticationRepository.setToken(token: token, completeTokenWaiters: true)
     }
 
+    /// Loads grouped channel groups for the app.
+    public func groupedQueryChannels(
```
laevandus: Wouldn't `queryGroupedChannels` read better?

Sources/StreamChat/ChatClient.swift
```diff
@@ -487,6 +487,49 @@ public class ChatClient {
...
+        limit: Int? = nil,
+        watch: Bool = false,
+        presence: Bool = false,
+        completion: @escaping (Result<GroupedChannels, Error>) -> Void
```
laevandus: Most of our public layer completion handlers run on the MainActor, we should do the same here.

Sources/StreamChat/ChatClient.swift
```diff
@@ -487,6 +487,49 @@ public class ChatClient {
         authenticationRepository.setToken(token: token, completeTokenWaiters: true)
     }
 
+    /// Loads grouped channel groups for the app.
+    public func groupedQueryChannels(
+        limit: Int? = nil,
+        watch: Bool = false,
```
laevandus: Don't we want `watch` to be true by default?

CHANGELOG.md
```diff
@@ -3,7 +3,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 # Upcoming
 
+### ✅ Added
+- Add `ChatChannelListController.prefill(channels:completion:)` for priming controller-local channel data before the first synchronize call while preserving normal pagination, observation, and offline refresh behavior
+- Add `ChatClient.groupedQueryChannels(limit:watch:presence:)` to fetch grouped channel groups as `GroupedChannels`, preserving backend group keys and exposing normalized per-group channels and unread counts for integrators
+- Add optional `groupedUnreadChannels` data to grouped unread websocket events and persist it on `CurrentChatUser` for integrators
```
laevandus: We should probably drop "for integrators". Most of the time our release notes are more concise

```
- Add `ChatChannelListController.prefill(channels:completion:)` for priming controller-local channel data
- Add `ChatClient.groupedQueryChannels(limit:watch:presence:)` to fetch grouped channels with per group unread counts
- Add optional `groupedUnreadChannels` data to relevant web-socket events and to `CurrentChatUser`
```

Sources/StreamChat/ChatClient.swift
martinmitrevski: I think we will keep it as false by default, and just rely on events coming to members regardless of watching

CHANGELOG.md
martinmitrevski: yes, this was AI generated, will be revisited

Sources/StreamChat/ChatClient.swift
martinmitrevski: sure, that's also fine

Sources/StreamChat/ChatClient.swift
martinmitrevski: we should also change them to final class

Sources/StreamChat/Config/BaseURL.swift
```diff
@@ -7,7 +7,7 @@ import Foundation
 /// A struct representing base URL for `ChatClient`.
 public struct BaseURL: CustomStringConvertible {
     /// The default base URL  for StreamChat service.
-    public static let `default` = BaseURL(urlString: "https://chat.stream-io-api.com/")!
+    public static let `default` = BaseURL(urlString: "https://chat-edge-dublin-ce2.stream-io-api.com/")!
```
laevandus, line 10: Reminder for this change

Sources/StreamChat/ChatClient.swift
```diff
@@ -645,6 +645,36 @@ public class ChatClient {
         }
     }
     
+    // MARK: - Grouped Channels
+    
+    /// Queries grouped channel groups for the app.
+    public func queryGroupedChannels(
```
laevandus, line 651: Reminder that sync repository needs to use this endpoint as well for not spamming query channels endpoint when app reconnects.

laevandus:
Closing the PR because we are targeting develop for this feature.

PR #4075, opened 2026-04-27
Stabilise StreamCDNRequester caching key across query-item orders

nuno-vieira:
### 🔗 Issue Links

https://linear.app/stream/issue/IOS-1652

Companion to the SwiftUI fix in [stream-chat-swiftui#1439](https://github.com/GetStream/stream-chat-swiftui/pull/1439). This PR addresses the underlying root cause that affected both the UIKit and SwiftUI SDKs.

### 🎯 Goal

`StreamCDNRequester` was producing different caching keys for the same image at the same size on different calls. Any UI surface that re-mounts an attachment in a fresh view tree (most visibly the SwiftUI reactions overlay, but the bug applied equally to the UIKit SDK and any other caller of `StreamCDNRequester`) would write the bitmap under one key and look it up under another, turning every re-render into a Nuke cache miss and showing a loading spinner over an already-cached image.

### 📝 Summary

- Sort the persisted resize parameters (`crop`, `h`, `resize`, `w`) inside `StreamCDNRequester.buildCachingKey` so the caching key collapses onto a single canonical value per `(image, size)` pair regardless of upstream insertion order.
- Add test coverage that locks in the stable ordering across query-item orders, repeated invocations, and explicit alphabetical sort.

### 🛠 Implementation

`StreamCDNRequester.buildImageURL` assembles the resize query items from a `[String: String]` dictionary literal:

```swift
var queryItems: [String: String] = [
    "w": ..., "h": ..., "resize": ..., "ro": "0"
]
for (key, value) in queryItems { items.append(...) }
```

Swift dictionaries do not preserve insertion order — the iteration order is hash-based and can vary between calls. The resulting `URLComponents.queryItems` therefore comes out in different orders on different calls, and `buildCachingKey` (which filters but preserves order) emits caching keys like `?h=576&resize=clip&w=768` for one call and `?h=576&w=768&resize=clip` for another. Same image, same size — different `String` values, so Nuke treats them as two different cache entries.

The minimal fix is to sort the persisted parameters alphabetically inside `buildCachingKey` before re-serialising the URL. That guarantees a single canonical caching key per `(image, size)` pair regardless of how the upstream URL was assembled, which also makes the function defensive against custom `CDNRequester` implementations and manually-constructed CDN URLs.

### 🧪 Manual Testing Notes

1. Open a channel that contains a message with an image attachment.
2. Long-press the message to open the reactions overlay.
3. The attachment should appear instantly with no loading spinner — previously, the spinner would flash on top of an already-cached image.
4. Repeat the open/close several times: the lookup should hit the memory cache every time.

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed image caching so cache keys are generated consistently, reducing redundant downloads and improving image load performance.

* **Tests**
  * Added tests to ensure cache key consistency across equivalent image URLs and prevent regressions.

### DIFF
--- a/Sources/StreamChat/APIClient/CDNClient/StreamCDNRequester.swift
+++ b/Sources/StreamChat/APIClient/CDNClient/StreamCDNRequester.swift
@@ -89,12 +89,6 @@ public final class StreamCDNRequester: CDNRequester, Sendable {
         }
 
         let persistedParameters = ["w", "h", "resize", "crop"]
-        // Sort the persisted parameters so the caching key is independent
-        // of the upstream query-item insertion order. Without sorting, two
-        // URLs that describe the same image at the same size but were
-        // assembled in different orders would hash to different cache
-        // entries — `buildImageURL` builds them from a `[String: String]`
-        // dictionary whose iteration order is not stable across calls.
         let newParameters = (components.queryItems ?? [])
             .filter { persistedParameters.contains($0.name) }
             .sorted { $0.name < $1.name }
--- a/Sources/StreamChat/APIClient/CDNClient/StreamCDNRequester.swift
+++ b/Sources/StreamChat/APIClient/CDNClient/StreamCDNRequester.swift
@@ -89,7 +89,15 @@ public final class StreamCDNRequester: CDNRequester, Sendable {
         }
 
         let persistedParameters = ["w", "h", "resize", "crop"]
-        let newParameters = components.queryItems?.filter { persistedParameters.contains($0.name) } ?? []
+        // Sort the persisted parameters so the caching key is independent
+        // of the upstream query-item insertion order. Without sorting, two
+        // URLs that describe the same image at the same size but were
+        // assembled in different orders would hash to different cache
+        // entries — `buildImageURL` builds them from a `[String: String]`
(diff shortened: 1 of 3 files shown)


Sources/StreamChat/APIClient/CDNClient/StreamCDNRequester.swift
```diff
@@ -89,7 +89,15 @@ public final class StreamCDNRequester: CDNRequester, Sendable {
         }
 
         let persistedParameters = ["w", "h", "resize", "crop"]
-        let newParameters = components.queryItems?.filter { persistedParameters.contains($0.name) } ?? []
+        // Sort the persisted parameters so the caching key is independent
+        // of the upstream query-item insertion order. Without sorting, two
+        // URLs that describe the same image at the same size but were
+        // assembled in different orders would hash to different cache
+        // entries — `buildImageURL` builds them from a `[String: String]`
+        // dictionary whose iteration order is not stable across calls.
```
nuno-vieira:
```suggestion
```

laevandus (approved):

martinmitrevski (approved):

PR #4076, opened 2026-04-28
Grouped channels endpoint (v5)

laevandus:
### 🔗 Issue Links

Resolves https://linear.app/stream/issue/IOS-1635/support-for-grouped-channels-endpoint.

### 🎯 Goal

- Add `ChatClient.queryGroupedChannels(limit:presence:watch:)` (callback + `async throws` overloads) that fetches grouped channel groups from `POST /channels/grouped` and returns them as a public `ChannelGroup` model (group key, channel ids, per-group unread channel count).
- Add `ChatClient.makeChannelList(with: groupKey)` — state-layer factory that creates a `ChannelList` for a single group, with pagination, observation, and offline-sync wired in.
- Add `CurrentChatUser.groupedUnreadCount: [String: Int]?` — per-group unread-channel counts, decoded from `grouped_unread_channels` on relevant WS events (`NotificationMessageNew`, `NotificationMarkRead`, `NotificationMarkUnread`, `MessageNew`, `ChannelTruncated`, `NotificationChannelDeleted`) via the new `HasGroupedUnreadCount` protocol and persisted on `CurrentUserDTO`.

### 📝 Summary

Test on https://github.com/GetStream/GroupedChannelsSample.

This is the v5 port of the `grouped-channels-endpoint` work that originally landed on the v4 line. All public-API changes are additive; the legacy `ChatChannelListController` is intentionally **not** extended for grouped channels — group-based observation lives only in the state-layer `ChannelList`.

### 🛠 Implementation

- **New endpoint:** `POST /channels/grouped` wired through `ChannelEndpoints`/`EndpointPath` with `GroupedQueryChannelsRequestBody` / `GroupedQueryChannelsPayload` / `GroupedQueryChannelsGroupPayload`.
- **State layer:** `ChannelList` branches on `query.groupKey` — initial fetch and pagination go through `queryGroupedChannels` (the regular `/channels` path is skipped). Pagination cursors are persisted on `ChannelListQueryDTO.next`.
- **Sync:** `SyncGroupedChannelsOperation` runs once during the offline-sync chain and issues a single shared `queryGroupedChannels` for all active group-based `ChannelList`s, populating their query DTOs so each list's subsequent reads hit the cache.
- **Linking:** `ChannelListLinker` detects a channel's group via `channel.extraData["group"]` (`GroupedChannelKey.extraData`) and routes link/unlink decisions accordingly. The reserved `"all"` group key (`GroupedChannelKey.all`) is a union — every channel that has a non-empty group is linked to it.
- **Grouped unread count maintenance:** `ChannelListUpdater.link` increments and `unlink` decrements `CurrentUserDTO.groupedUnreadCount[groupKey]` (floored at 0) when the channel actually has unread messages and the membership set actually changes. The `"all"` group is skipped — its count is driven by server-side WS events.
- **Page size sentinel:** The grouped endpoint rejects `limit > 10`. Instead of hard-coding `10`, `ChannelListQuery(groupKey:)` uses `Pagination(pageSize: .unsetPageSize)` (= `-1`), and `Pagination.encode` skips writing `limit` when negative — letting the backend pick its default.
- **Core Data:** Two new optional attributes; lightweight migration, no model version bump.
  - `ChannelListQueryDTO.next: String?` — per-group pagination cursor.
  - `CurrentUserDTO.unreadGroupedChannelsCounts: Data?` — JSON-encoded `[String: Int]`, exposed via the computed `groupedUnreadCount` accessor.
- **Concurrency:** Adapted to Swift 6 strict concurrency on `develop`: `Sendable` conformance on the new payloads/models, `@Sendable` on completion handlers, `nonisolated(unsafe)` for mutated locals captured in `database.write` blocks, and explicit `@Sendable` on the inner closures inside `SyncGroupedChannelsOperation` to escape `@MainActor` inheritance from the `queryGroupedChannels` callback (otherwise crashes at `_swift_task_checkIsolatedSwift` on `.utility-qos`).

### 🎨 Showcase

| Before | After |
| ------ | ----- |
|  img   |  img  |

### 🧪 Manual Testing Notes

Use the [GroupedChannelsSample](https://github.com/GetStream/GroupedChannelsSample) app. Verify:
- Listing groups via `queryGroupedChannels` returns groups with correct per-group unread channel counts.
- Creating a `ChannelList` for a `groupKey` shows that group's channels and paginates through them.
- Sending/reading messages updates `groupedUnreadCount` on `CurrentChatUser` via WS events.
- Linking/unlinking a channel into/out of a non-`"all"` group adjusts the local `groupedUnreadCount` immediately; the `"all"` group reflects the server's value.

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **New Features**
  * Fetch channels grouped by backend "group" with cursor pagination; new client API to query grouped channels and factory to create/observe a grouped ChannelList.
  * Current user exposes per-group unread channel counts; relevant sync, websocket events, and notifications now carry grouped unread counts.
  * Added ChannelGroup model and group-key aware query/pagination behavior.

* **Tests**
  * Added comprehensive tests for grouped queries, payload decoding, pagination, persistence, sync, and unread-count handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/GetStream/stream-chat-swift/pull/4076?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


Sources/StreamChat/Config/BaseURL.swift
```diff
@@ -7,7 +7,7 @@ import Foundation
 /// A struct representing base URL for `ChatClient`.
 public struct BaseURL: CustomStringConvertible, Sendable {
     /// The default base URL  for StreamChat service.
-    public static let `default` = BaseURL(urlString: "https://chat.stream-io-api.com/")!
+    public static let `default` = BaseURL(urlString: "https://chat-edge-dublin-ce2.stream-io-api.com/")!
```
laevandus: Note for not committing

Sources/StreamChat/Database/DTOs/CurrentUserDTO.swift
```diff
@@ -7,6 +7,7 @@ import Foundation
 
 @objc(CurrentUserDTO)
 class CurrentUserDTO: NSManagedObject {
+    @NSManaged var groupedUnreadChannelsData: Data?
```
martinmitrevski: hmm, is this storing all the channels from the grouped endpoint?

Sources/StreamChat/Models/GroupedChannels.swift
```diff
@@ -0,0 +1,56 @@
+//
+// Copyright © 2026 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+/// A grouped channels response returned by `ChatClient.queryGroupedChannels`.
+public struct GroupedChannels: Sendable {
```
martinmitrevski: shouldn't we do final classes instead?

Sources/StreamChat/Models/GroupedChannels.swift
```diff
@@ -0,0 +1,56 @@
...
+    }
+}
+
+/// A grouped channels group returned by `ChatClient.queryGroupedChannels`.
+public struct GroupedChannelsGroup: Sendable {
+    /// The group key as returned by the backend (e.g. `"all"`, `"new"`, `"current"`).
+    public let groupKey: String
+
+    /// The channels that belong to this group.
+    public let channels: [ChatChannel]
```
martinmitrevski: feels strange to return this as part of the current user, wdyt?

Sources/StreamChat/Models/GroupedChannels.swift
```diff
@@ -0,0 +1,56 @@
...
+    init(
+        groupKey: String,
+        channels: [ChatChannel],
+        unreadChannels: Int,
+        next: String? = nil,
+        prev: String? = nil
+    ) {
+        self.groupKey = groupKey
+        self.channels = channels
+        let derivedUnreadChannels = channels.reduce(into: 0) { partialResult, channel in
```
martinmitrevski: why do we do this?

Sources/StreamChat/Models/UnreadCount.swift
```diff
@@ -4,6 +4,9 @@
 
 import Foundation
 
+/// Grouped unread channel counts keyed by the backend-provided group identifier.
+public typealias GroupedUnreadChannels = [String: Int]
```
martinmitrevski: do we still use this?

Sources/StreamChat/Query/ChannelListQuery.swift
```diff
@@ -53,6 +53,12 @@ public struct ChannelListQuery: Encodable, Sendable, LocalConvertibleSortingQuer
         self.membersLimit = membersLimit
     }
 
+    init(groupKey: String) {
+        self.init(filter: .and([]))
```
martinmitrevski: why do we do this?

Sources/StreamChat/Query/ChannelListQuery.swift
```diff
@@ -53,6 +53,12 @@ public struct ChannelListQuery: Encodable, Sendable, LocalConvertibleSortingQuer
...
+        self.groupKey = groupKey
+        self.pagination = Pagination(pageSize: 10)
```
martinmitrevski: isn't this server limited? Not sure if it's a good idea to hardcode it client side.

Sources/StreamChat/Repositories/SyncOperations.swift
```diff
@@ -94,14 +94,48 @@ final class RefreshChannelListOperation: AsyncOperation, @unchecked Sendable {
...
-                    log.error("Failed refreshing channel list (\(channelList.query.filter) with error \(error)", subsystems: .offlineSupport)
+                    log.error("Failed refreshing channel list (\(query.filter) with error \(error)", subsystems: .offlineSupport)
+                    done(.retry)
+                }
+            }
+        }
+    }
+}
+
+final class SyncGroupedChannelsOperation: AsyncOperation, @unchecked Sendable {
```
martinmitrevski, line 123: probably good to document what this does

Sources/StreamChat/Repositories/SyncRepository.swift
```diff
@@ -160,7 +160,8 @@ class SyncRepository: @unchecked Sendable {
     ///
     /// Background mode (other regular API requests are allowed to run at the same time)
     /// 1. Collect all the **active** channel ids (from instances of `Chat`, `ChannelList`, `ChatChannelController`, `ChatChannelListController`)
-    /// 2. Refresh channel lists (channels for current pages in `ChannelList`, `ChatChannelListController`)
+    /// 2. Refresh channel lists (channels for current pages in `ChannelList`, non-grouped `ChatChannelListController`)
+    /// 2.5 Refresh the shared grouped channels response when any grouped `ChatChannelListController` or `ChannelList` is active
```
martinmitrevski: 2.5? 😄

Sources/StreamChat/StateLayer/ChannelList.swift
```diff
@@ -64,31 +66,54 @@ public class ChannelList: @unchecked Sendable {
     /// - Throws: An error while communicating with the Stream API.
     /// - Returns: An array of channels for the pagination.
     @discardableResult public func loadChannels(with pagination: Pagination) async throws -> [ChatChannel] {
-        try await channelListUpdater.loadChannels(query: query, pagination: pagination)
+        if let groupKey = query.groupKey {
+            let groupedChannels = try await channelListUpdater.queryGroupedChannels(
+                groupPagination: .init(groupKey: groupKey, next: pagination.cursor),
+                limit: pagination.pageSize,
+                watch: true,
```
martinmitrevski: Why's this hardcoded to true? Afaik, it should be false, and we rely on membership events. Or this changed in the meantime?

Sources/StreamChat/StateLayer/ChannelListState+Observer.swift
```diff
@@ -5,18 +5,17 @@
 import Foundation
 
 extension ChannelListState {
-    final class Observer {
-        private let channelListObserver: StateLayerDatabaseObserver<ListResult, ChatChannel, ChannelDTO>
+    @MainActor final class Observer {
```
martinmitrevski: should this be on the main actor?

Sources/StreamChat/Workers/ChannelListLinker.swift
```diff
@@ -119,35 +133,42 @@ final class ChannelListLinker: Sendable {
...
+    private func unlinkChannel(_ channel: ChatChannel) {
         isInChannelList(channel) { [worker, query] exists, _ in
-            guard exists else {
-                completion?()
-                return
-            }
-            worker.unlink(channel: channel, with: query) { _ in
-                completion?()
-            }
+            guard exists else { return }
```
martinmitrevski: we're not calling completion here, while we did that before. Is that expected?

Sources/StreamChat/Workers/ChannelListLinker.swift
```diff
@@ -119,35 +133,42 @@ final class ChannelListLinker: Sendable {
...
-            // query matches the DB Predicate. So here we default to saying it always belong to the current query.
-            return true
+    private func linkingAction(for channel: ChatChannel) -> LinkingAction {
+        if let groupKey = query.groupKey {
+            // "all" group key is special, all the other groups are always linked to it
+            let currentGroupKey = channel.extraData["group"]?.stringValue?
+                .trimmingCharacters(in: .whitespacesAndNewlines)
+                .lowercased()
+            if let currentGroupKey, !currentGroupKey.isEmpty {
+                return groupKey == currentGroupKey || groupKey == "all" ? .link : .unlink
```
martinmitrevski: this check is not ideal, are we sure it's always going to be called "all"? But not sure if there's better way, we don't have any clue about the filter.

Sources/StreamChat/Workers/ChannelListLinker.swift
```diff
@@ -119,35 +133,42 @@ final class ChannelListLinker: Sendable {
...
+            }
+            return .none
+        } else {
+            if let filter = filter {
+                return filter(channel) ? .link : .unlink
+            }
+            
+            if clientConfig.isChannelAutomaticFilteringEnabled {
+                // When auto-filtering is enabled the channel will appear or not automatically if the
+                // query matches the DB Predicate. So here we default to saying it always belong to the current query.
```
martinmitrevski: not sure I fully understand - does that mean that in this case, we delegate the decision to the db predicate?

martinmitrevski (commented):
Looks good in general! Need to test a bit, also left some comments, let me know what you think.

Sources/StreamChat/Controllers/DatabaseObserver/BackgroundDatabaseObserver.swift
```diff
@@ -16,7 +16,7 @@ class BackgroundDatabaseObserver<Item: Sendable, DTO: NSManagedObject>: @uncheck
     private let sorting: [SortValue<Item>]
 
     /// Used to observe the changes in the DB.
-    private(set) var frc: NSFetchedResultsController<DTO>
+    let frc: NSFetchedResultsController<DTO>
```
laevandus: revert

Sources/StreamChat/Models/GroupedChannels.swift
laevandus: It was part of the initial work what you did. I don't think this should be here either. I am gonna clean it up.

Sources/StreamChat/Database/DTOs/CurrentUserDTO.swift
laevandus: Let me rename it. It is actually String Int dictionary (group <-> unread count)

Sources/StreamChat/Models/GroupedChannels.swift
martinmitrevski: yeah, the initial implementation didn't have unread count in the response, we should remove it now.

Sources/StreamChat/Models/GroupedChannels.swift
laevandus: Yeah, now we should drop these and the way for reading channels is through ChannelList

Sources/StreamChat/Models/GroupedChannels.swift
laevandus: For public data types we still do struct, so for consistency reasons I think it is better to follow the same pattern

Sources/StreamChat/Query/ChannelListQuery.swift
laevandus: It is a problem with existing public interface where pageSize is non-optional. API throws an error when passing too large page size (10 is max). Probably makes sense to workaround it and use negative value here which skips encoding the page size when making requests.

Sources/StreamChat/Query/ChannelListQuery.swift
laevandus: Let me make it nicer. Filter itself is not used with grouped channels, and the filter property is non-optional.

Sources/StreamChat/StateLayer/ChannelList.swift
laevandus: Changed it everywhere back to false

Sources/StreamChat/Workers/ChannelListLinker.swift
laevandus: This is expected, it was a bit off before where same event could do both unlink and link (only one action would actually do something)

Sources/StreamChat/Workers/ChannelListLinker.swift
laevandus: I added some extensive documentation above

Sources/StreamChat/StateLayer/ChannelList.swift
martinmitrevski: I assume backend confirmed this, right?

Sources/StreamChat/Query/Pagination.swift
```diff
@@ -17,6 +17,12 @@ public extension Int {
     static let channelWatchersPageSize = 30
 }
 
+extension Int {
+    /// Sentinel signalling "no client-side page size" — the request omits `limit` and the backend uses its default.
+    /// `0` matches `NSFetchRequest.fetchLimit`'s "no limit" semantics, so it safely passes through to Core Data unchanged.
+    static let unsetPageSize = 0
```
martinmitrevski: did you check also with the backend that 0 will return their default?

martinmitrevski (approved):
lgtm, let's test this extensively

Sources/StreamChat/Query/Pagination.swift
laevandus: It is more future proof to use -1 here since some APIs can interpret it as return nothing (although the pagination encoding below was filtering it out).

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
```diff
@@ -281,7 +281,9 @@ extension NSManagedObjectContext {
         dto.deletedAt = payload.deletedAt?.bridgeDate
         dto.updatedAt = payload.updatedAt.bridgeDate
         dto.defaultSortingAt = (payload.lastMessageAt ?? payload.createdAt).bridgeDate
-        dto.lastMessageAt = payload.lastMessageAt?.bridgeDate
+        if let lastMessageAt = payload.lastMessageAt {
+            dto.lastMessageAt = lastMessageAt.bridgeDate
+        }
```
nuno-vieira: @laevandus This can be dangerous, AFAIK, in some scenarios, we do want lastMessageAt to be reset to nil. This most likely will break some scenarios. Why was this unwrapping added?

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
```diff
@@ -497,8 +499,13 @@ extension ChannelDTO {
...
         request.predicate = NSCompoundPredicate(type: .and, subpredicates: subpredicates)
-        request.fetchLimit = query.pagination.pageSize
-        request.fetchBatchSize = query.pagination.pageSize
+        request.fetchLimit = query.pagination.pageSize == .unsetPageSize ? 0 : query.pagination.pageSize
+        // For grouped queries `pageSize` is `.unsetPageSize`, which would disable batching.
+        // Fall back to the standard channels page size to keep memory bounded as the linked set
+        // grows across pagination.
+        request.fetchBatchSize = query.pagination.pageSize == .unsetPageSize
+            ? .channelsPageSize
+            : query.pagination.pageSize
```
nuno-vieira: Why is this needed? Not very clear from the docs 🤔

Sources/StreamChat/Models/ChannelGroup.swift
```diff
@@ -0,0 +1,49 @@
...
+        self.next = next
+    }
+}
+
+/// Constants used by the grouped channels feature.
+enum GroupedChannelKey {
+    /// Special group key whose list contains channels from every other group.
+    static let all = "all"
+    /// `ChatChannel.extraData` field that carries the channel's group membership.
+    static let extraData = "group"
```
nuno-vieira: This is a bit weird, shouldn't it be `static let group = "group"`?

Sources/StreamChat/Query/ChannelListQuery.swift
```diff
@@ -71,6 +77,15 @@ public struct ChannelListQuery: Encodable, Sendable, LocalConvertibleSortingQuer
         try options.encode(to: encoder)
         try pagination.encode(to: encoder)
     }
+    
+    var groupKey: String?
+
+    /// The stable identity used for locating / linking the corresponding `ChannelListQueryDTO`.
+    /// Uses `groupKey` when set (stable across date-bearing filters from the grouped endpoint);
+    /// otherwise falls back to `filter.filterHash`.
```
nuno-vieira: What are date-bearing filters? Maybe we can select a bit easier wording here

Sources/StreamChat/StateLayer/ChannelList.swift
```diff
@@ -42,53 +42,99 @@ public class ChannelList: @unchecked Sendable {
...
+    ///   `watch` / `presence` flags persisted by the most recent
+    ///   ``ChatClient/queryGroupedChannels(groups:limit:presence:watch:)`` call for the group.
+    ///
+    /// Subsequent pages are loaded via ``loadMoreChannels(limit:)``.
     ///
-    /// - Important: Loaded channels in ``ChannelListState/channels`` are reset.
+    /// - Important: For group-based lists, prefer `get()` only when fetching the first page for a *single* group
+    /// in isolation. When the app needs first pages for multiple groups, call
+    /// ``ChatClient/queryGroupedChannels(groups:limit:presence:watch:)`` instead — it returns every group
+    /// in one request, which is significantly more efficient than calling `get()` per `ChannelList`.
```
nuno-vieira: I'm a bit concerned we are adding a lot of docs to one of the most used functions for something that will be disabled for 99% of customers. This will just add too much noise. Maybe we could introduce a separate function called `getGrouped()` or something like this

Sources/StreamChat/StateLayer/ChannelList.swift
```diff
@@ -42,53 +42,99 @@ public class ChannelList: @unchecked Sendable {
...
+
     // MARK: - Channel List Pagination
-    
+
     /// Loads channels for the specified pagination parameters and updates ``ChannelListState/channels``.
     ///
-    /// - Important: If the pagination offset is 0 and cursor is nil, then loaded channels are reset.
+    /// - Important: For filter-based lists, loaded channels are reset when the pagination offset is 0 and
+    /// the cursor is nil. For group-based lists, only ``Pagination/cursor`` is used — the offset is ignored —
+    /// and the grouped endpoint controls the page contents.
```
nuno-vieira: Same here. Feels like we should have seperate functions for this special niche feature.

Sources/StreamChat/StateLayer/ChatClient+Factory.swift
```diff
@@ -51,6 +51,21 @@ extension ChatClient {
...
+    public func makeChannelList(with groupKey: String) -> ChannelList {
+        let channelList = ChannelList(
+            query: .init(groupKey: groupKey),
+            dynamicFilter: nil,
+            client: self
+        )
+        // Start tracking immediately, because the first page is meant to be fetches with queryGroupedChannels
+        syncRepository.startTrackingChannelList(channelList)
+        return channelList
+    }
```
nuno-vieira, line 68: Just like we have a seperate function for grouped channels, it probably makes sense to have seperate functions to interact with it. Or even return a `ChannelGroupedList` object.

Sources/StreamChat/Workers/ChannelListUpdater.swift
```diff
@@ -214,7 +322,7 @@ extension ChannelListUpdater {
             }
         }
     }
-    
+
     // MARK: -
```
nuno-vieira: Can we remove this empty mark?

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
laevandus: This is from the initial batch of changes when Martin started with it. I think I should remove it because the initial implementation had logic around it.

Sources/StreamChat/Query/ChannelListQuery.swift
laevandus: I should remove it, this is part of the second iteration of this feature and it is not relevant with what we have now. At some point there were queries with Date() in it.

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
laevandus: Currently it is hard to use backend driven page sizes because out `Pagination` type requires page size to be set. Therefore I created a new -1 value for representing it. Query grouped channels throws an error if page size is more than 10 - I don't want to hard code it. That is OK, but for keeping CoreData to fetch less data, both of these properties should be set with >0 values.

Sources/StreamChat/Database/DTOs/ChannelDTO.swift
martinmitrevski: yes, I think there was a bug with it missing in some payload. But should be removed - please test though.

Sources/StreamChat/APIClient/Endpoints/Payloads/ChannelListPayload.swift
```diff
@@ -25,6 +25,99 @@ extension ChannelListPayload: Decodable {
...
+    enum CodingKeys: String, CodingKey {
+        case limit
+        case next
+    }
+
+    func encode(to encoder: Encoder) throws {
+        var container = encoder.container(keyedBy: CodingKeys.self)
+        if let limit, limit > 0 {
+            try container.encode(limit, forKey: .limit)
+        }
```
laevandus: -1 for using backend limits case. Just in case not sending although backend does check for negative values.

Sources/StreamChat/StateLayer/ChatClient+Factory.swift
laevandus: After exploring this the end decision was to keep it in ChannelList.

Sources/StreamChat/APIClient/Endpoints/Payloads/ChannelListPayload.swift
martinmitrevski: can't we use `nil` everywhere instead?

Sources/StreamChat/StateLayer/ChannelList.swift
```diff
@@ -64,7 +64,20 @@ public class ChannelList: @unchecked Sendable {
     /// - Throws: An error while communicating with the Stream API.
     /// - Returns: An array of channels for the pagination.
     @discardableResult public func loadChannels(with pagination: Pagination) async throws -> [ChatChannel] {
-        try await channelListUpdater.loadChannels(query: query, pagination: pagination)
+        if let groupKey = query.groupKey {
+            let state = try await channelListUpdater.paginationState(for: groupKey)
+            let channelGroups = try await channelListUpdater.queryGroupedChannels(
+                groups: [groupKey: .init(limit: pagination.pageSize, next: pagination.cursor)],
+                limit: nil,
+                watch: state.watch ?? true,
```
martinmitrevski, line 72: do we keep true as a fallback?

martinmitrevski (approved):
I think it turned out nicer with the new linker

Sources/StreamChat/StateLayer/ChannelList.swift
laevandus: Currently We have to bc otherwise it won't work, channel.updated not coming through. Needs backend changes to send more events when not watching, but those are not planned at the moment.

martinmitrevski (approved):
lgtm ✅ good if @testableapple runs another QA after these changes. Or maybe we merge both PRs (since they're related), and do one QA?

PR #4077, opened 2026-04-29
Fix mid-page jump breaking channel list preview and reopen flicker

nuno-vieira:
### 🔗 Issue Links

- [IOS-1658](https://linear.app/stream/issue/IOS-1658/v5-qa-fix-jumping-to-mid-page-breaking-channel-preview)

### 🎯 Goal

Fix the v5 regression where, after tapping a quoted message (or any flow that paginates to a mid-page) and popping back to the channel list, the channel list item rendered "No messages" instead of the actual latest message. This regression surfaced because v5 moved channel-list previews onto `ChatChannel.latestMessages`.

The most natural fix — stop wiping the cached channel messages on mid-page jumps — exposed two pre-existing latent issues that this PR also addresses, since they would otherwise be visible to users once the cache is preserved:

1. Reopening a channel that was last left mid-page briefly shows the persisted mid-page slice for ~1 frame before the first page reloads.
2. After a mid-page jump, previously-cached newer messages leak into the active message-list page, because the page-bounds (`oldestMessageAt` / `newestMessageAt`) live on `ChannelDTO` and `NSFetchedResultsController` does not re-evaluate cached `MessageDTO` rows when only the parent changes.

### 📝 Summary

- `ChannelUpdater.update`: stop wiping locally cached channel messages on every fetch — only reset `channel.messages` on first-page (re-)loads. Keeps `ChatChannel.latestMessages` populated after a mid-page jump.
- New `ChannelUpdater.cleanStaleMidPageStateIfNeeded(for:isInRecoveryMode:)` helper (callback + async overloads): when about to issue a fresh first-page request and the channel still holds mid-page state on disk, drop the cached slice and reset `oldestMessageAt` / `newestMessageAt` so the message observers don't briefly surface stale data.
- `ChatChannelController.synchronize` and `Chat.get(watch:)` both call the new helper before invoking the updater, so the fix reaches both the controller layer and the async-await state layer.
- `ChannelDTO.willSave`: extend the existing `truncatedAt` precedent to also nudge linked `MessageDTO`s when `oldestMessageAt` / `newestMessageAt` change, forcing any active FRC to re-evaluate `MessageDTO.channelMessagesPredicate` against rows it had already cached.

### 🛠 Implementation

**Why the channel list preview broke (v5 regression)**

`ChannelDTO.cleanAllMessagesExcludingLocalOnly()` was being invoked on every successful `update` completion, including mid-page pagination fetches. Since v5 derives the channel-list preview from `ChatChannel.latestMessages`, wiping `channel.messages` for a mid-page jump emptied the preview. We now gate the wipe behind `didLoadFirstPage`, so the locally cached slice survives mid-page jumps.

**Why the active page leaked older messages (latent issue exposed by the fix above)**

`MessageDTO.channelMessagesPredicate` filters by `channel.oldestMessageAt` / `channel.newestMessageAt`. Once `ChannelUpdater.update` stopped wiping `channel.messages`, the previously-cached newest slice stayed linked to the channel after a mid-page jump. `NSFetchedResultsController` does not re-run its predicate against cached rows when only a parent property changes, so the active page kept showing those out-of-bounds messages alongside the newly fetched mid-page slice. Following the existing `truncatedAt` pattern in `ChannelDTO.willSave`, we now nudge each linked message on any `oldestMessageAt` / `newestMessageAt` change to force the FRC to re-evaluate its predicate. This covers every writer of the bounds (the main update path *and* `cleanStaleMidPageStateIfNeeded`) without per-call bookkeeping.

#### ✅ Pros

- Surgical, layered change with no Core Data migration; fully backwards-compatible.
- The FRC nudge lives next to the existing `truncatedAt` precedent in `ChannelDTO.willSave`, so any future writer of the page bounds gets the correct observer behavior for free.
- Common `synchronize` / `get` path stays synchronous; the async cleanup branch only runs when stale mid-page state is detected.
- Both layers (UIKit `ChatChannelController` and async-await `Chat`) get the same fix from the same helper, so no behavior drift between them.

#### 🔭 Future Improvement

Architecturally, the root cause is that `ChannelDTO.messages` is overloaded to mean both "the channel's recent history" (used by `ChatChannel.latestMessages` for previews) and "the active controller's paginated page" (used by the message-list observers). 

A cleaner long-term direction would be to introduce a dedicated `ChannelMessagesQueryDTO` that owns the active page and links to `MessageDTO`s through its own relationship. This way, the `channel.latestMessages` would be decoupled from the actual current active channel messages.

### 🧪 Manual Testing Notes

1. Open a channel and load several pages of older messages.
2. Tap a quoted-message reply (or use any flow that calls `loadPageAroundMessageId(...)`) to jump to a mid-page.
3. Pop back to the channel list.
   - Before: the channel list item shows "No messages".
   - After: the channel list item shows the actual latest message.
4. Tap the same channel again.
   - Before: the mid-page slice flickers for ~1 frame, then jumps to the first page.
   - After: the channel opens directly to the first page; no flicker.
5. While in step 2 (mid-page view), confirm the message list shows only the mid-page slice — no previously-cached newer messages should appear above the jumped-to message.
6. Repeat steps 2–5 in a SwiftUI demo wired through `Chat` to verify the state-layer path.

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented unintended clearing of cached messages during mid-page jumps and pagination; channel previews now retain their latest messages.
  * Stale mid-page cleanup now runs only when appropriate (first-page fetches with stale bounds) and is skipped during recovery or when pagination params apply.
  * Deallocation no longer triggers cache purge that removed messages unexpectedly.
  * Page-boundary timestamp updates now preserve linked message associations for accurate previews.

* **Tests**
  * Updated and expanded tests to assert retained cache behavior and cover stale-cleanup and failure scenarios.

* **Documentation**
  * Updated changelog entries for pagination and message-caching fixes.

### DIFF
--- a/Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
+++ b/Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
@@ -1882,14 +1882,6 @@ public class ChatChannelController: DataController, DelegateCallable, DataStoreP
             }
         }
     }
-
-    deinit {
-        guard self.isJumpingToMessage, let cid = self.cid else { return }
-        dataStore.database.write { session in
-            let channelDTO = session.channel(cid: cid)
-            channelDTO?.cleanAllMessagesExcludingLocalOnly()
-        }
-    }
 }
 
 // MARK: - Environment
--- a/Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
+++ b/Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
@@ -1930,34 +1930,31 @@ private extension ChatChannelController {
         isInRecoveryMode: Bool,
         _ completion: (@MainActor (_ error: Error?) -> Void)? = nil
     ) {
-        updater.cleanStaleMidPageStateIfNeeded(
-            for: channelQuery,
-            isInRecoveryMode: isInRecoveryMode
-        ) { [weak self] in
-            guard let self else { return }
-            let channelCreatedCallback = self.isChannelAlreadyCreated ? nil : self.channelCreated(forwardErrorTo: self.setLocalStateBasedOnError)
-            self.updater.update(
-                channelQuery: self.channelQuery,
-                isInRecoveryMode: isInRecoveryMode,
-                onChannelCreated: channelCreatedCallback,
-                completion: { result in
-                    switch result {
-                    case .success:
-                        self.state = .remoteDataFetched
-                        self.callback { completion?(nil) }
-                    case let .failure(error):
-                        self.state = .remoteDataFetchFailed(ClientError(with: error))
-                        self.callback { completion?(error) }
-                    }
+        let channelCreatedCallback = isChannelAlreadyCreated ? nil : channelCreated(forwardErrorTo: setLocalStateBasedOnError)
+        // `update` performs any required stale mid-page cleanup synchronously before
+        // dispatching the network request, so observers we start below in parallel
+        // with the API call see a clean cache.
+        updater.update(
+            channelQuery: channelQuery,
+            isInRecoveryMode: isInRecoveryMode,
+            onChannelCreated: channelCreatedCallback,
+            completion: { result in
+                switch result {
+                case .success:
+                    self.state = .remoteDataFetched
+                    self.callback { completion?(nil) }
+                case let .failure(error):
+                    self.state = .remoteDataFetchFailed(ClientError(with: error))
+                    self.callback { completion?(error) }
                 }
-            )
-
-            /// Setup observers if we know the channel `cid` (if it's missing, it'll be set in `set(cid:)`
-            /// Otherwise they will be set up after channel creation, in `set(cid:)`.
-            if let cid = self.cid {
-                self.setupEventObservers(for: cid)
-                self.setLocalStateBasedOnError(self.startDatabaseObservers())
             }
+        )
+
+        /// Setup observers if we know the channel `cid` (if it's missing, it'll be set in `set(cid:)`
+        /// Otherwise they will be set up after channel creation, in `set(cid:)`.
+        if let cid = cid {
+            setupEventObservers(for: cid)
+            setLocalStateBasedOnError(startDatabaseObservers())
         }
     }
 
--- a/Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
+++ b/Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
@@ -1930,7 +1930,10 @@ private extension ChatChannelController {
         isInRecoveryMode: Bool,
         _ completion: (@MainActor (_ error: Error?) -> Void)? = nil
     ) {
-        clearStaleMidPageStateIfNeeded(isInRecoveryMode: isInRecoveryMode) { [weak self] in
+        updater.cleanStaleMidPageStateIfNeeded(
+            for: channelQuery,
+            isInRecoveryMode: isInRecoveryMode
+        ) { [weak self] in
             guard let self else { return }
             let channelCreatedCallback = self.isChannelAlreadyCreated ? nil : self.channelCreated(forwardErrorTo: self.setLocalStateBasedOnError)
             self.updater.update(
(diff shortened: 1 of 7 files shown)


Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
```diff
@@ -1882,14 +1882,6 @@ public class ChatChannelController: DataController, DelegateCallable, DataStoreP
...
         }
     }
-
-    deinit {
-        guard self.isJumpingToMessage, let cid = self.cid else { return }
-        dataStore.database.write { session in
-            let channelDTO = session.channel(cid: cid)
-            channelDTO?.cleanAllMessagesExcludingLocalOnly()
-        }
-    }
```
laevandus, line 1892: Wow, did not realise we have something like that

Sources/StreamChat/Workers/ChannelUpdater.swift
```diff
@@ -122,6 +121,76 @@ class ChannelUpdater: Worker, @unchecked Sendable {
         }
     }
 
+    /// When the channel was last left in a mid-page state (the user jumped to a message and
```
laevandus: I am looking at this and wondering if we could hide all of this in the `channelUpdater.update(` implementation for keeping ChatChannelController and Chat cleaner. We know based on the query when initial page is getting loaded, can't we use that (+ another argument if needed) and put all of this into the update.

nuno-vieira: I wanted to do this, but it was tricky because of the observers. We need to perform this before starting the observers. Starting the observers is in parallel with the update call, and we can't wait for the HTTP call to start the observers and load the data from cache.

So the flow is: Check mid-page state, then update call + start observers at the same time.

At least on the UIKit controllers, it was tricky.

Another disadvantage was testing, it would make testing much more complicated if everything is inside the update function

Sources/StreamChat/Workers/ChannelUpdater.swift
```diff
@@ -122,6 +121,76 @@ class ChannelUpdater: Worker, @unchecked Sendable {
...
+    func cleanStaleMidPageStateIfNeeded(
+        for channelQuery: ChannelQuery,
+        isInRecoveryMode: Bool,
+        completion: @escaping @Sendable () -> Void
+    ) {
+        let isFirstPageFetch = channelQuery.pagination?.parameter == nil
+        guard !isInRecoveryMode,
+              isFirstPageFetch,
+              let cid = channelQuery.cid,
+              channelHasStaleMidPageState(cid: cid) else {
```
laevandus: Because this is blocking it takes away the flickering? Wondering why not run it inside the database write and return early there.

Sources/StreamChat/Workers/ChannelUpdater.swift
```diff
@@ -122,6 +121,76 @@ class ChannelUpdater: Worker, @unchecked Sendable {
...
+        }
+
+        database.write({ session in
+            guard let channelDTO = session.channel(cid: cid),
+                  channelDTO.newestMessageAt != nil else { return }
+            channelDTO.cleanAllMessagesExcludingLocalOnly()
+            channelDTO.oldestMessageAt = nil
+            channelDTO.newestMessageAt = nil
+        }, completion: { _ in
+            DispatchQueue.main.async { completion() }
```
laevandus: Can we remove main.async here, typically updater returns on the DB queue

Sources/StreamChat/Workers/ChannelUpdater.swift
nuno-vieira: How do we return early inside the block?

laevandus: At first I was looking at why
```
database.write({ session in
            guard let channelDTO = session.channel(cid: cid),
                  channelDTO.newestMessageAt != nil else { return }
            channelDTO.cleanAllMessagesExcludingLocalOnly()
            channelDTO.oldestMessageAt = nil
            channelDTO.newestMessageAt = nil
        }, completion: { _ in
```
this is not enough. Does the same check as `channelHasStaleMidPageState`. But I think I understand that it is crucial for `channelHasStaleMidPageState` to be blocking for really taking the flicker away?

nuno-vieira: Yes it needs to be blocking because of the flicker, otherwise the previous messages would show up for a split second

laevandus: OK, last one. The flicker was not caused by `DispatchQueue.main.async { completion() }`?

Sources/StreamChat/Workers/ChannelUpdater.swift
```diff
@@ -832,6 +837,65 @@ class ChannelUpdater: Worker, @unchecked Sendable {
...
+    /// without driving a full `update` round-trip.
+    private func cleanStaleMidPageStateIfNeeded(
+        for channelQuery: ChannelQuery,
+        isInRecoveryMode: Bool
+    ) {
+        let isFirstPageFetch = channelQuery.pagination?.parameter == nil
+        guard !isInRecoveryMode,
+              isFirstPageFetch,
+              let cid = channelQuery.cid,
+              channelHasStaleMidPageState(cid: cid) else {
```
laevandus: Wait a sec, we can delete this now because the writableContext does the same thing internally. Then we have just one blocking call.

laevandus (approved):

PR #4085, opened 2026-05-05
Fix CurrentUserController.setPushPreference not updating local state

laevandus:
### 🔗 Issue Links

[IOS-1570](https://linear.app/stream/issue/IOS-1570)

### 🎯 Goal

`CurrentUserController.setPushPreference(level:)` succeeded against the API but the local `currentUser.pushPreference` never reflected the change, so UIs bound to it stayed stale until the app reconnected and re-fetched the current user. This PR makes the local round-trip work the way callers expect.

### 📝 Summary

- Save the `PushPreferenceDTO` under the actual current user id and wire `CurrentUserDTO.pushPreference = savedDTO`.
- Strongly capture `self` in `CurrentUserController.setPushPreference` and `snoozePushNotifications` callbacks, matching the rest of the controller.
- Cover the round-trip in `CurrentUserUpdater_Tests` (saved-id assertion + `currentUser.pushPreference` wired up + regression check that no DTO is saved under the literal `"currentUserId"`).

### 🛠 Implementation

`CurrentUserUpdater.setPushPreference` was previously writing the DTO under the literal id `"currentUserId"`. The remote-fetch path (`CurrentUserDTO.saveCurrentUser`) saves the same kind of DTO under the real user id and sets the inverse relationship to `CurrentUserDTO`. The local-save therefore created an orphan `PushPreferenceDTO` that was never linked to the current user, which is why the model never refreshed.

The fix reads `session.currentUser` inside the existing `database.write` block, saves with `currentUserDTO.user.id`, and explicitly sets `currentUserDTO.pushPreference = savedDTO`. Core Data's inverse handling sets `pushPreferenceDTO.currentUser = currentUserDTO` automatically. If there is no current user in the DB the save is logged and skipped — the API success still propagates to the completion to preserve the existing public contract.

The two `CurrentUserController` callsites for `currentUserUpdater.setPushPreference(_:)` were the only methods on that controller that captured `self` weakly while routing through `self.callback { … }`. They are now consistent with `updateUserData`, `addDevice`, `removeDevice`, etc.

### 🎨 Showcase

N/A — no UI surface in this SDK PR.

### 🧪 Manual Testing Notes

In the demo app or an integration test:

1. Sign in.
2. Call `chatClient.currentUserController().setPushPreference(level: .none)`; assert success.
3. Read `chatClient.currentUserController().currentUser?.pushPreference?.level` — should be `.none`.
4. Call again with `.all`; the observed value should flip to `.all` without restarting the client.

Before this PR, step 3/4 reflected the value only after a fresh fetch of the current user; after this PR, it updates immediately.

### ☑️ Contributor Checklist

## Release Notes

* **Bug Fixes**
  * Fixed push notification preferences not updating locally when changed through preference settings or notification snooze actions.
  * Improved push preference storage to properly associate settings with the current user account.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### 🐞 Fixed
 - Fix image cache misses caused by non-deterministic caching keys in `StreamCDNRequester` [#4075](https://github.com/GetStream/stream-chat-swift/pull/4075)
 - Fix `ChatChannel.latestMessages` being wiped on mid-page pagination [#4077](https://github.com/GetStream/stream-chat-swift/pull/4077)
-- Fix `CurrentUserController.setPushPreference` and `snoozePushNotifications` not updating the local push preference state
+- Fix `CurrentUserController.setPushPreference` and `snoozePushNotifications` not updating the local push preference state [#4085](https://github.com/GetStream/stream-chat-swift/pull/4085)
 
 ## StreamChatUI
 ### 🐞 Fixed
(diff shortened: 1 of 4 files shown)


nuno-vieira (approved):
LGTM! ✅

PR #4091, opened 2026-05-12
Allow custom CDNStorage to mutate attachment payload via UploadedFile

nuno-vieira:
### 🔗 Issue Links

https://linear.app/stream/issue/IOS-1700/cdnstorage-mutate-attachment-payload

### 🎯 Goal

Custom `CDNStorage` implementations can already replace the file URL and the thumbnail URL of an uploaded attachment, but they have no way to write upload-time metadata (CDN-assigned IDs, codecs, dimensions, signed thumbnail URLs, custom titles, …) into the typed payload.

The existing `UploadedAttachmentPostProcessor` runs *after* the upload, on an `UploadedAttachment` that only carries `remoteURL` and `thumbnailURL`, so it has no access to the custom response data the storage saw during upload. This forces customers to re-fetch or re-derive that data outside of the upload flow.

This PR closes that gap without breaking the existing v5 `CDNStorage` contract.

### 📝 Summary

- Add an optional `attachment: AnyChatMessageAttachment?` to `UploadedFile`, so custom `CDNStorage` implementations can hand back a mutated attachment together with the file URLs.
- `APIClient.uploadAttachment` now uses `file.attachment ?? attachment` when constructing the final `UploadedAttachment`, so the storage's mutated attachment flows through the regular upload pipeline.
- Document the workflow on `CDNStorage.uploadAttachment(_:options:completion:)` and on `UploadedFile.attachment`, calling out that this is only meaningful for the message-attached overload (standalone uploads have no attachment context).

### 🛠 Implementation

`UploadedFile` gains a new `public let attachment: AnyChatMessageAttachment?` field with a defaulted `nil` initializer parameter, so the existing call sites (and the default `StreamCDNStorage`) keep compiling unchanged. The constructor signature stays source-compatible (the new parameter has a default value).

To preserve the existing wire format, `UploadedFile` now defines an explicit `CodingKeys` enum that omits `attachment` and a custom `init(from:)` that decodes only `fileURL` and `thumbnailURL`. The `attachment` field is therefore a runtime-only carrier from the CDN storage to the API client and is never serialized into a network payload.

`APIClient.uploadAttachment` was the only place that wrapped `UploadedFile` into an `UploadedAttachment`. It now prefers `file.attachment` over the original attachment when present, so any payload mutation a custom storage performs propagates through `AttachmentQueueUploader` and `UploadedAttachmentPostProcessor` as if the SDK had produced it itself.

The default `StreamCDNStorage` does not opt into this — `UploadedFile.attachment` is left at `nil` and the API client falls back to the original attachment, matching today's behavior.

### 🧪 Manual Testing Notes

This change is exercised through unit tests:

- `UploadedFile` initializes with and without an attachment, and `init(from:)` always decodes `attachment` as `nil` (network payloads stay forward/backward compatible).
- `APIClient.uploadAttachment` uses the storage-returned attachment when provided, and falls back to the original attachment otherwise.
- Default `StreamCDNStorage` upload result has `attachment == nil`.

To validate end-to-end, drop in a custom `CDNStorage` that mutates the attachment payload using `AnyAttachmentUpdater` (see the docs-content PR linked below), upload an image/video/audio attachment, and confirm the mutated `title` / `extraData` survive into `ChatMessage.attachments`.

### ☑️ Contributor Checklist

- [x] Documentation has been updated in the `docs-content` repo (see [docs-content#1280](https://github.com/GetStream/docs-content/pull/1280))

## Summary by CodeRabbit

* **New Features**
  * File uploads can now include attachment metadata returned by custom CDN responses; the SDK will use CDN-provided attachment metadata when present and fall back to the original attachment otherwise.

* **Tests**
  * Added tests covering CDN-provided attachment handling and fallback behavior.

* **Documentation**
  * Clarified how custom CDN implementations can mutate and return attachment payloads.

[Review Change Stack](https://app.coderabbit.ai/change-stack/GetStream/stream-chat-swift/pull/4091)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 # Upcoming
 
+## StreamChat
+### ✅ Added
+`UploadedFile` now carries an optional `attachment` so custom `CDNStorage` implementations can mutate the attachment payload [#4091](https://github.com/GetStream/stream-chat-swift/pull/4091)
+
 ## StreamChatCommonUI
 ### 🔄 Changed
 - `Appearance.localizationProvider` now falls back from `Bundle.main` to the SDK bundle, so apps can override keys without installing a custom provider [#4088](https://github.com/GetStream/stream-chat-swift/pull/4088)
--- a/Sources/StreamChat/APIClient/APIClient.swift
+++ b/Sources/StreamChat/APIClient/APIClient.swift
@@ -321,7 +321,7 @@ class APIClient: @unchecked Sendable {
             self?.cdnStorage.uploadAttachment(attachment, options: .init(progress: progress)) { [weak self] result in
                 let mappedResult = result.map { file in
                     UploadedAttachment(
-                        attachment: attachment,
+                        attachment: file.attachment ?? attachment,
                         remoteURL: file.fileURL,
                         thumbnailURL: file.thumbnailURL
                     )
--- a/Sources/StreamChat/APIClient/CDNClient/CDNStorage.swift
+++ b/Sources/StreamChat/APIClient/CDNClient/CDNStorage.swift
@@ -11,6 +11,10 @@ import Foundation
 public protocol CDNStorage: Sendable {
     /// Uploads an attachment tied to a message and returns the uploaded file information.
     ///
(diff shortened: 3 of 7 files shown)


martinmitrevski (approved):

laevandus (approved):
Looks good

PR #4103, opened 2026-05-19
Add LivestreamChat state-layer component

nuno-vieira:
### 🔗 Issue Links

- [IOS-1707](https://linear.app/stream/issue/IOS-1707)
- Docs PR: [GetStream/docs-content#1299](https://github.com/GetStream/docs-content/pull/1299)

### 🎯 Goal

Expose the livestream-optimized channel API to the `StateLayer`, so SwiftUI / Combine / async-await consumers can adopt the same in-memory, performance-focused engine that `LivestreamChannelController` already provides to UIKit / delegate consumers.

### 📝 Summary

- Extract the shared in-memory state and event handling out of `LivestreamChannelController` into a new `LivestreamChatHandler`.
- `LivestreamChannelController` becomes a thin delegate-driven adapter over the handler. Public API is preserved.
- Add `LivestreamChat` (async/await) and `LivestreamChatState` (`@MainActor` `ObservableObject` with `@Published` channel/messages/typingUsers/isPaused/skippedMessagesAmount) backed by the same handler.
- Add `ChatClient.makeLivestreamChat(for:)` / `makeLivestreamChat(with:)` factory methods.
- Track active `LivestreamChat` instances in `SyncRepository` and re-watch them on reconnection via `WatchChannelOperation`.
- Refactor the YouTube Clone example to drive its chat through `LivestreamChat` instead of `LivestreamChannelController`, mirroring the `DemoApp` livestream screen's structure.
- Add a `LivestreamChat_Tests` suite covering init, get/watch/stopWatching, pagination, pause/resume, and event handling.
- Remove the unused `pinning` parameter from `LivestreamChannelController.pin(messageId:completion:)` (It is dead code, and for now it is coupled with the local DB, the expiration property)

### 🛠 Implementation

The `LivestreamChannelController` already had all the livestream-specific behavior (no DB, capped messages, pause/resume, typing). Rather than duplicate that into `LivestreamChat`, the in-memory state container (`channel`, `messages`, `isPaused`, `skippedMessagesAmount`, `typingUsers`) and the event-processing pipeline were lifted into `LivestreamChatHandler`. Both the controller and the chat now compose the same handler and just translate its callbacks into their respective surface: a `MulticastDelegate` for the controller, `@Published` properties on `LivestreamChatState` for the chat.

`LivestreamChat` mirrors `Chat` (factory on `ChatClient`, `StateBuilder<LivestreamChatState>`, an `Environment` for DI, `async throws` methods for every operation). `NotificationAddedToChannelEvent` is intercepted in both `LivestreamChannelController.didReceiveEvent` and `LivestreamChat.didReceiveEvent` to trigger a re-watch — synchronously for the controller (preserving the existing test contract) and via a `Task` for the chat.

`SyncRepository` gained an `activeLivestreamChats: ThreadSafeWeakCollection<LivestreamChat>` plus `start/stopTrackingLivestreamChat(_:)`, and `ActiveChannelIdsOperation` / `WatchChannelOperation` were extended to cover livestream chats during sync.

The YouTube Clone refactor is the integration test: `YTLiveChatViewController` now subclasses `_ViewController`, observes `livestreamChat.state` via Combine, and routes every action (send, react, keystroke, stop-typing, pagination, pause/resume) through `async/await`. A small `YTLiveChatComposerVC` subclass and an inlined pause banner replace the controller-driven message creation and overlap fix for the scroll-to-bottom button.

### 🧪 Manual Testing Notes

In the YouTube Clone example app:

1. Run the `YouTube` scheme.
2. Open the livestream screen.
3. Send messages, react, type to verify typing indicators, scroll up to trigger pause, scroll back / send to resume.
4. Backgrounding the app and reconnecting should re-watch the channel via `SyncRepository`.

In the main `DemoApp`, the existing livestream screen continues to use `LivestreamChannelController` and should be unaffected.

### ☑️ Contributor Checklist

- [x] Documentation has been updated in the `docs-content` repo

## Summary by CodeRabbit

* **New Features**
  * LivestreamChat state layer with observable state, async/await APIs, and ChatClient factory methods.
* **Refactor**
  * Livestream chat moved to a handler-backed, in-memory architecture with updated watch/pagination flows.
* **Examples**
  * YouTube Clone UI/composer updated to use the new livestream chat flow.
* **Bug Fixes**
  * Public pin API simplified (removed unused pinning parameter).
* **Tests**
  * Extensive new/updated tests covering lifecycle, pagination, pause/resume, events.
* **Documentation**
  * CHANGELOG updated to mention LivestreamChat.

[Review Change Stack](https://app.coderabbit.ai/change-stack/GetStream/stream-chat-swift/pull/4103?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


Examples/YouTubeClone/YTLiveChatViewController.swift
```diff
@@ -2,29 +2,410 @@
...
+        keyboardHandler.stop()
+
+        resignFirstResponder()
+    }
+
+    // MARK: - State Observation
+
+    private func observeStateChanges() {
+        livestreamChat.state.$channel
+            .receive(on: DispatchQueue.main)
```
laevandus: State is main actor and this class is main actor so there should not be need for `receive(on:`

Examples/YouTubeClone/YTLiveChatViewController.swift
```diff
@@ -2,29 +2,410 @@
...
+    ) -> ChatMessageDecorationView? {
+        nil
+    }
+
+    // MARK: - EventsControllerDelegate
+
+    func eventsController(_ controller: EventsController, didReceiveEvent event: any Event) {
+        if event is NewMessagePendingEvent {
+            if livestreamChat.state.isPaused {
+                pauseBannerView.setState(.resuming)
```
laevandus, line 380: Should this call `livestreamChat.resume()` as well?

Sources/StreamChat/Controllers/ChannelController/LivestreamChannelHandler.swift
```diff
@@ -0,0 +1,603 @@
...
+/// - Pause / skipped messages bookkeeping.
+/// - The typing cleanup timers that compensate for the lack of
+///   `TypingStartCleanupMiddleware` on livestream channels.
+/// - The handling of channel-specific events (messages, reactions, members,
+///   watchers, typing, etc.).
+///
+/// State changes are surfaced through a ``Handlers`` struct whose closures are
+/// always invoked on the main thread, matching the existing controller's
+/// delegate dispatching behavior.
+final class LivestreamChannelHandler: DataStoreProvider, @unchecked Sendable {
```
laevandus: Should this be MainActor class because internally it does not guard any shared state?
Or maybe it needs to be guarded from outside with a lock instead.

Sources/StreamChat/Controllers/ChannelController/LivestreamChatHandler.swift
```diff
@@ -0,0 +1,603 @@
...
+    private var currentUserId: UserId? { client.currentUserId }
+
+    /// Per-user typing cleanup timers. Required because livestream channels
+    /// bypass `TypingStartCleanupMiddleware`.
+    private var typingCleanupTimers: [UserId: TimerControl] = [:]
+
+    // MARK: - Handlers
+
+    /// Closures invoked on the main thread when state changes.
+    struct Handlers {
```
laevandus, line 132: Probably we should skip default values here, because otherwise we might forgot to set these.

Sources/StreamChat/StateLayer/LivestreamChat.swift
```diff
@@ -0,0 +1,694 @@
...
+public class LivestreamChat: AppStateObserverDelegate, @unchecked Sendable {
+    private let client: ChatClient
+    private let channelUpdater: ChannelUpdater
+    private let apiClient: APIClient
+    private let typingEventsSender: TypingEventsSender
+    private let appStateObserver: AppStateObserving
+    private let handler: LivestreamChannelHandler
+    private var eventObserver: AnyCancellable?
+    @MainActor private var stateBuilder: StateBuilder<LivestreamChatState>
+    @MainActor private var isResuming: Bool = false
```
laevandus: Should this be internal in LivestreamChatState?

Sources/StreamChat/StateLayer/LivestreamChat.swift
```diff
@@ -0,0 +1,694 @@
...
+
+/// An object which represents a livestream ``ChatChannel`` and its state.
+///
+/// Unlike ``Chat``, ``LivestreamChat`` operates without local database persistence.
+/// It manages channel state in memory and communicates directly with the API. It
+/// is more performant for high-throughput livestream channels but has fewer
+/// features than ``Chat``, like for example:
+/// - Read updates are not tracked.
+/// - Replies / threads are not exposed.
+public class LivestreamChat: AppStateObserverDelegate, @unchecked Sendable {
```
laevandus, line 16: Looks like @unchcked can be dropped here and it could be Sendable with final class

Sources/StreamChat/StateLayer/LivestreamChat.swift
```diff
@@ -0,0 +1,694 @@
...
+                },
+                skippedMessagesAmountDidChange: { [weak self] skipped in
+                    self?.stateBuilder.state.skippedMessagesAmount = skipped
+                },
+                typingUsersDidChange: { [weak self] typingUsers in
+                    self?.stateBuilder.state.typingUsers = typingUsers
+                }
+            )
+        )
+    }
```
laevandus: Should we hide it to LivestreamChatState instead?

Sources/StreamChat/StateLayer/LivestreamChat.swift
```diff
@@ -0,0 +1,694 @@
...
+    ///   - limit: The limit for the page size. The default limit is 25.
+    ///
+    /// - Throws: An error while communicating with the Stream API.
+    public func loadOlderMessages(before messageId: MessageId? = nil, limit: Int? = nil) async throws {
+        let messageId = messageId
+            ?? handler.paginationStateHandler.state.oldestFetchedMessage?.id
+            ?? handler.messages.last?.id
+
+        guard let messageId else { throw ClientError.ChannelEmptyMessages() }
+        guard !handler.hasLoadedAllPreviousMessages, !handler.isLoadingPreviousMessages else { return }
```
laevandus: Should this be the first check here so that messageId related error won't fire if pagination is complete or there is nothing to load?

Sources/StreamChat/StateLayer/LivestreamChat.swift
```diff
@@ -0,0 +1,694 @@
...
+    }
+
+    /// Resumes the collecting of new messages.
+    ///
+    /// This will load the first page, resetting ``LivestreamChatState/messages`` to the
+    /// latest messages. After resuming, new messages will be added to
+    /// ``LivestreamChatState/messages`` again.
+    ///
+    /// - Throws: An error while communicating with the Stream API.
+    @MainActor public func resume() async throws {
```
laevandus, line 269: Should we be consistent and not have this function `@MainActor`?

Sources/StreamChat/StateLayer/LivestreamChat.swift
```diff
@@ -0,0 +1,694 @@
...
+        to messageId: MessageId,
+        with type: MessageReactionType,
+        score: Int = 1,
+        enforceUnique: Bool = false,
+        skipPush: Bool = false,
+        pushEmojiCode: String? = nil,
+        extraData: [String: RawJSON] = [:]
+    ) async throws {
+        try await withCheckedThrowingContinuation { (continuation: CheckedContinuation<Void, Error>) in
+            apiClient.request(
```
laevandus: I feel like we need async version of the request so that all these callsites can be without continuations

Sources/StreamChat/StateLayer/LivestreamChatState.swift
```diff
@@ -0,0 +1,105 @@
...
+/// Represents the observable state of a ``LivestreamChat`` instance.
+///
+/// ``LivestreamChatState`` mirrors the in-memory state managed by
+/// ``LivestreamChannelHandler`` and exposes it via `@Published` properties so it
+/// can be observed from SwiftUI views or other Combine consumers.
+@MainActor public final class LivestreamChatState: ObservableObject {
+    private let handler: LivestreamChannelHandler
+
+    /// The client instance the ``LivestreamChat`` was created with.
+    public let client: ChatClient
```
laevandus, line 17: Should we skip such public property?

Examples/YouTubeClone/YTLiveChatViewController.swift
nuno-vieira: No, the resume will be called once the message is actually published, so that it waits for the last page to load again

Sources/StreamChat/Controllers/ChannelController/LivestreamChannelHandler.swift
```diff
@@ -0,0 +1,603 @@
...
+/// - Pause / skipped messages bookkeeping.
+/// - The typing cleanup timers that compensate for the lack of
+///   `TypingStartCleanupMiddleware` on livestream channels.
+/// - The handling of channel-specific events (messages, reactions, members,
+///   watchers, typing, etc.).
+///
+/// State changes are surfaced through a ``Handlers`` struct whose closures are
+/// always invoked on the main thread, matching the existing controller's
+/// delegate dispatching behavior.
+class LivestreamChannelHandler: DataStoreProvider, @unchecked Sendable {
```
nuno-vieira: Maybe we can add a `LivestreamChannelHandling` protocol, so that only public properties are exposed without making tests more difficult

Sources/StreamChat/Controllers/ChannelController/LivestreamChannelHandler.swift
```diff
@@ -0,0 +1,603 @@
...
+        let typingUsers = currentTypingUsers.filter { $0.id != userId }
+        updateCurrentlyTypingUsers(typingUsers)
+    }
+
+    private func updateCurrentlyTypingUsers(_ typingUsers: Set<ChatUser>) {
+        // Compare by user id only. `Set<ChatUser>.==` falls back to `ChatUser.Equatable`,
+        // which checks ~13 fields including `lastActiveAt` and `extraData`. The server
+        // refreshes those on each re-emitted `typing.start`, so a strict comparison would
+        // report a change for every keystroke from an already-typing user and fire
+        // `didUpdateChannel` (plus rebuild the `ChatChannel` struct) for free.
```
nuno-vieira:
```suggestion
```

Sources/StreamChat/StateLayer/LivestreamChatState.swift
nuno-vieira: This one follows the same pattern as Chat

Sources/StreamChat/Controllers/ChannelController/LivestreamChannelHandler.swift
nuno-vieira: Considered all three options:

- **`@MainActor` on the handler** — forces `@MainActor` up onto `LivestreamChannelController` and `LivestreamChat`, which would diverge from `Chat` / `ChatChannelController` (both currently non-isolated public types). Tried it locally; the propagation is intrusive.
- **`actor` for the handler** — requires `await` at every call site, which is incompatible with the sync property getters on both wrappers (e.g. `controller.channel`, `controller.messages`, `controller.pause()`). No in-language workaround without breaking the public sync API.
- **Internal lock** — per-property locks don't fix the read-modify-write composites in the event handlers; we'd need method-level lock scopes around `handleChannelEvent`, `applyMessageLimit`, typing handling, `paginationStateHandler` calls, etc. Hot-path overhead on the exact workload this type is built for.

Sticking with `@unchecked Sendable` and keeping the access invariant upstream: `EventNotificationCenter.subscribe` does `.receive(on: DispatchQueue.main)`, the controller wraps every API completion in `DispatchQueue.main.async`, and the `Handlers` struct closures are typed `@MainActor` so the publication boundary is compile-time enforced. Happy to revisit if a non-main-thread call site shows up in practice.

Sources/StreamChat/StateLayer/LivestreamChat.swift
nuno-vieira: Done

Sources/StreamChat/Controllers/ChannelController/LivestreamChannelController.swift
```diff
@@ -626,11 +635,9 @@ public class LivestreamChannelController: DataStoreProvider, AppStateObserverDel
     /// Pins a message.
     /// - Parameters:
     ///   - messageId: The message identifier to pin.
-    ///   - pinning: The pinning expiration information. It supports setting an infinite expiration, setting a date, or the amount of time a message is pinned.
     ///   - completion: Called when the network request is finished. If request fails, the completion will be called with an error.
     public func pin(
         messageId: MessageId,
-        pinning: MessagePinning = .noExpiration,
```
martinmitrevski, line 633: This is a breaking change, let's keep it - we might need expiration time at some point.

Sources/StreamChat/Controllers/ChannelController/LivestreamChatHandler.swift
```diff
@@ -0,0 +1,617 @@
...
+            MainActor.assumeIsolated { callback(handlers) }
+        } else {
+            DispatchQueue.main.async {
+                callback(handlers)
+            }
+        }
+    }
+
+    private func updateMessagesArray(with newMessages: [ChatMessage], pagination: MessagesPagination?) {
+        let newMessages = Array(newMessages.reversed())
```
martinmitrevski, line 267: why are we reversing the array here?

Sources/StreamChat/Controllers/ChannelController/MainQueueLivestreamChatHandler.swift
```diff
@@ -0,0 +1,152 @@
...
+/// thread always observe a consistent snapshot, and callers from cooperative
+/// pool tasks get the same eventual-consistency guarantee they had before,
+/// without the risk of reading a partially-mutated multi-word value.
+///
+/// When the caller is already on the main thread, the dispatch is inlined so
+/// existing main-thread call sites (init, configuration during setup, UIKit
+/// actions) keep their synchronous semantics. This preserves wrapper code that
+/// relies on a mutation being visible before the next line runs as long as the
+/// caller is on main, which is true for the wrappers' public surface today.
+final class MainQueueLivestreamChatHandler: LivestreamChatHandling, @unchecked Sendable {
```
laevandus: Should we revert this addition and just do Atomic in the handler?

Sources/StreamChat/StateLayer/LivestreamChat.swift
```diff
@@ -0,0 +1,679 @@
...
+            quotedMessageId: quotedMessageId,
+            skipPush: skipPushNotification,
+            skipEnrichUrl: skipEnrichURL,
+            restrictedVisibility: restrictedVisibility,
+            extraData: transformableInfo.extraData
+        )
+        client.eventNotificationCenter.process(
+            NewMessagePendingEvent(message: localMessage, cid: cid)
+        )
+        return localMessage.id
```
laevandus, line 351: There is logic in `Chat` for waiting the API request to happen in MessageSender. Probably makes sense to have it here as well.

Sources/StreamChat/StateLayer/LivestreamChat.swift
```diff
@@ -0,0 +1,679 @@
...
+    ///   - limit: The number of reactions to load. Default is 25.
+    ///   - offset: The starting position of the desired range to be fetched. Default is 0.
+    ///
+    /// - Throws: An error while communicating with the Stream API.
+    /// - Returns: An array of reactions for the specified message.
+    public func loadReactions(
+        for messageId: MessageId,
+        limit: Int = 25,
+        offset: Int = 0
+    ) async throws -> [ChatMessageReaction] {
```
laevandus, line 456: We probably want @discardableResult for all the functions which return something.

Sources/StreamChat/Controllers/ChannelController/LivestreamChannelController.swift
nuno-vieira: It is dead code at the moment, does not make sense to keep it. And making it work will require more changes because it is dependent on DB

Sources/StreamChat/Controllers/ChannelController/LivestreamChatHandler.swift
nuno-vieira: This is the same implementation as before. Now I don't remember, but most likely because messages do not come in the correct order: https://github.com/GetStream/stream-chat-swift/blob/develop/Sources/StreamChat/Controllers/ChannelController/LivestreamChannelController.swift#L1027

Sources/StreamChat/StateLayer/LivestreamChat.swift
nuno-vieira: At the moment Chat.swift does not have this

Sources/StreamChat/StateLayer/LivestreamChat.swift
nuno-vieira: Since we are not using the DB here, better to keep it like this for now

martinmitrevski (approved):

laevandus (approved):

PR #4105, opened 2026-05-20
Forward CDN headers when generating local video previews

nuno-vieira:
### 🔗 Issue Links

N/A

### 🎯 Goal

Fix a regression in `StreamMediaLoader` where local video preview thumbnails fail to load for videos served from authenticated/signed CDNs.

### 📝 Summary

- `StreamMediaLoader.generateVideoPreview` now forwards the CDN headers to `AVURLAsset`, matching `loadVideoAsset`.
- Extracts the asset-construction logic into a shared private helper so both call sites can't drift again.
- Adds unit tests for the headers-forwarding contract (`assetOptions(for:)`).

### 🛠 Implementation

`generateVideoPreview` was calling `cdnRequester.fileRequest(for:)` to obtain the signed URL and any authentication headers, but it constructed the asset as `AVURLAsset(url: adjustedUrl)` — silently dropping `cdnRequest.headers`. For self-hosted/authenticated CDNs, the resulting `AVAssetImageGenerator` request reached the server without `Authorization` / custom headers and failed, so local preview thumbnails for authenticated videos never rendered. `loadVideoAsset` had the correct behaviour all along (passing headers via `AVURLAssetHTTPHeaderFieldsKey`), so the two call sites were inconsistent.

The fix extracts the asset-construction logic into two helpers on `StreamMediaLoader`:

- `assetOptions(for:)` — pure helper that returns the options dictionary (or `nil` when there are no headers). Kept `internal` (not `private`) so the headers-forwarding contract is directly unit-testable via `@testable import`.
- `makeAVURLAsset(from:)` — private convenience that composes `AVURLAsset(url:options:)` with `assetOptions(for:)`.

Both `loadVideoAsset` and `generateVideoPreview` now route through `makeAVURLAsset(from:)`, so they can't drift again. `loadVideoAsset` now uses `[weak self]` + a "MediaLoader was deallocated" failure fallback for symmetry with `generateVideoPreview`.

I considered the reporter's alternative suggestion to have the preview path delegate to `loadVideoAsset`, but it would force constructing/discarding a `MediaLoaderVideoAsset` wrapper and change behaviour more than necessary. The shared helper gives the same correctness with a smaller blast radius.

### 🧪 Manual Testing Notes

- Configure a `CDNRequester` (or use the default `StreamCDNRequester` against a private/signed CDN) that returns non-empty `headers` from `fileRequest(for:options:)`.
- Send/render a video attachment without a remote thumbnail (so `generateVideoPreview` runs).
- Before the fix: the preview is blank because `AVAssetImageGenerator` fails authentication; after the fix: the thumbnail renders.

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * CDN authentication headers are now correctly forwarded when generating local video preview thumbnails.

* **Tests**
  * Added tests verifying header forwarding and correct behavior when CDN headers are nil or empty.

* **Documentation**
  * Added a changelog entry documenting the fix.

[Review Change Stack](https://app.coderabbit.ai/change-stack/GetStream/stream-chat-swift/pull/4105?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/Sources/StreamChatCommonUI/ImageLoading/StreamMediaLoader.swift
+++ b/Sources/StreamChatCommonUI/ImageLoading/StreamMediaLoader.swift
@@ -177,9 +177,6 @@ open class StreamMediaLoader: MediaLoader, @unchecked Sendable {
     /// Returns `nil` when there are no headers to forward. Otherwise returns a
     /// dictionary containing `AVURLAssetHTTPHeaderFieldsKey` so the asset's
     /// underlying network requests carry the CDN-provided authentication headers.
-    ///
-    /// Internal (not `private`) so the headers-forwarding contract can be
-    /// directly unit-tested via `@testable import`.
     func assetOptions(for cdnRequest: CDNRequest) -> [String: Any]? {
         guard let headers = cdnRequest.headers, !headers.isEmpty else { return nil }
         return ["AVURLAssetHTTPHeaderFieldsKey": headers]
--- a/Sources/StreamChatCommonUI/ImageLoading/StreamMediaLoader.swift
+++ b/Sources/StreamChatCommonUI/ImageLoading/StreamMediaLoader.swift
@@ -94,14 +94,16 @@ open class StreamMediaLoader: MediaLoader, @unchecked Sendable {
         options: VideoLoadOptions,
         completion: @escaping @MainActor (Result<MediaLoaderVideoAsset, Error>) -> Void
     ) {
-        cdnRequester.fileRequest(for: url, options: .init()) { result in
+        cdnRequester.fileRequest(for: url, options: .init()) { [weak self] result in
+            guard let self else {
+                StreamConcurrency.onMain {
+                    completion(.failure(ClientError.Unknown("MediaLoader was deallocated")))
+                }
+                return
+            }
             switch result {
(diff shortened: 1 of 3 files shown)


martinmitrevski (approved):

Sources/StreamChatCommonUI/ImageLoading/StreamMediaLoader.swift
```diff
@@ -169,6 +171,26 @@ open class StreamMediaLoader: MediaLoader, @unchecked Sendable {
...
 
+    /// Builds the options dictionary used when constructing the `AVURLAsset`
+    /// for a video URL returned by the CDN.
+    ///
+    /// Returns `nil` when there are no headers to forward. Otherwise returns a
+    /// dictionary containing `AVURLAssetHTTPHeaderFieldsKey` so the asset's
+    /// underlying network requests carry the CDN-provided authentication headers.
+    ///
+    /// Internal (not `private`) so the headers-forwarding contract can be
+    /// directly unit-tested via `@testable import`.
```
nuno-vieira:
```suggestion
```

PR #4109, opened 2026-05-22
Fix reconnection data race leaving WebSocket stuck at .disconnecting

laevandus:
### 🔗 Issue Links

[IOS-1715](https://linear.app/stream/issue/IOS-1715)

Companion v5+ fix: [GetStream/stream-core-swift#50](https://github.com/GetStream/stream-core-swift/pull/50)

### 🎯 Goal

When the device loses and regains Wi-Fi, the WebSocket would sometimes stay stuck in `.disconnecting` and never reconnect. The root cause is a data race between `DefaultConnectionRecoveryHandler.disconnectIfNeeded()` (which runs on `io.getStream.chat.internet-monitor`) and `WebSocketClient`'s engine-driven state writes (on `engineQueue`). The handler performs a check-then-act on `connectionState`: it reads `.connected`, the engine queue independently writes `.disconnected(.serverInitiated)`, then the handler unconditionally writes `.disconnecting(.systemInitiated)` — overwriting the legitimate `.disconnected`. The engine has already closed the socket and never re-fires `webSocketDidDisconnect`, so the state stays stuck. Automatic reconnection then refuses to fire because `isAutomaticReconnectionEnabled` only matches `.disconnected`.

### 📝 Summary

- Wrap `disconnectIfNeeded`, the two `appDidBecomeActive`/internet-ON reconnect call sites (now extracted into a single `reconnectIfNeededFromOffline()` helper), and the reconnection-timer `onFire` closure in `webSocketClient.engineQueue.async { ... }` so the handler's check-and-act runs on the same serial queue that owns engine state mutations.
- Widen `WebSocketClient.engineQueue` from `private` to internal access so the handler in the same module can dispatch onto it.
- Add three regression tests covering all three new dispatch points (`disconnectIfNeeded`, internet-ON reconnect, timer-driven reconnect).
- Update four existing tests that asserted synchronously on `connect_called` / `disconnect_called` / `disconnect_source` immediately after a trigger to use `AssertAsync.willBe...`. The triggers now go through `engineQueue.async`, so the call is observed shortly after, not synchronously.

### 🛠 Implementation

Both writers (the handler and the engine's `WebSocketEngineDelegate` callbacks) now run on `WebSocketClient.engineQueue` — a serial dispatch queue. That serialises the previously racing accesses without introducing new locks or actor boundaries. Two possible orderings remain, and both end correctly at `.disconnected`:

- Engine's `webSocketDidDisconnect` runs first → handler block runs second, sees `.disconnected`, `canBeDisconnected` returns `false`, no `disconnect()` call.
- Handler block runs first → transitions to `.disconnecting`, queued engine callback then matches the `.disconnecting` case in `WebSocketClient.webSocketDidDisconnect` and transitions to `.disconnected(source: source)`.

Deadlock-safety was audited on the equivalent v5+ change in `stream-core-swift`: every `engineQueue` usage is `.async` (no `.sync`), `@Atomic`'s lock and the engine/connect-request locks are never nested, and the `connectionState` `didSet` (and therefore the delegate notification) fires outside the `@Atomic` setter's lock. The v4 layout is structurally identical, so the same conclusions apply.

### 🧪 Manual Testing Notes

Reproduce the original bug scenario on a physical iPhone:

1. Launch a chat demo, sign in, confirm `Web socket connection state changed: connected(...)` in logs.
2. Turn Wi-Fi **off**. State should reach `.disconnected(.serverInitiated)`. It should not flip to `.disconnecting(.systemInitiated)`; if it does, it must immediately return to `.disconnected` (engine's `.disconnecting → .disconnected` recovery path).
3. Wait ~10 seconds.
4. Turn Wi-Fi **on**. Reconnection should fire automatically: `connecting → waitingForConnectionId → connected(...)`.
5. Sanity check the lifecycle path: foreground → background → foreground while connected. Disconnect on background, reconnect on foreground, no stuck state.

The new and updated unit tests run as part of `StreamChatTests`.

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChat/Workers/Background/ConnectionRecoveryHandler.swift
+++ b/Sources/StreamChat/Workers/Background/ConnectionRecoveryHandler.swift
@@ -108,9 +108,7 @@ extension DefaultConnectionRecoveryHandler {
 
         backgroundTaskScheduler?.endTask()
 
-        if canReconnectFromOffline {
-            webSocketClient.connect()
-        }
+        reconnectIfNeededFromOffline()
     }
 
     private func appDidEnterBackground() {
@@ -151,9 +149,7 @@ extension DefaultConnectionRecoveryHandler {
         log.debug("Internet -> \(isAvailable ? "✅" : "❌")", subsystems: .webSocket)
 
         if isAvailable {
-            if canReconnectFromOffline {
-                webSocketClient.connect()
-            }
+            reconnectIfNeededFromOffline()
         } else {
             disconnectIfNeeded()
         }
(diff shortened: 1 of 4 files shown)


martinmitrevski (approved):
looks good codewise, but @testableapple should run a QA

Sources/StreamChat/Workers/Background/ConnectionRecoveryHandler.swift
```diff
@@ -201,11 +197,52 @@ extension DefaultConnectionRecoveryHandler {
...
+    ///      `.disconnected(.serverInitiated)`.
+    ///   3. The internet-monitor queue resumes and unconditionally calls `disconnect(source: .systemInitiated)`,
+    ///      which writes `.disconnecting(.systemInitiated)`, overwriting the legitimate `.disconnected`.
+    ///   4. The engine has already closed the socket, so no further `webSocketDidDisconnect` fires to
+    ///      transition `.disconnecting → .disconnected`. State stays stuck at `.disconnecting`.
+    ///   5. When Wi-Fi returns, automatic reconnection is rejected because the state's
+    ///      `isAutomaticReconnectionEnabled` only matches `.disconnected`. The client never recovers.
+    ///
+    /// With this dispatch, the two writers are serialized on the same queue, so either ordering
+    /// ends at `.disconnected` — no stuck `.disconnecting` state.
```
nuno-vieira: This is a bit too extensive, can we update the docs here just to mention what is the current state, we don't need the previous state.

nuno-vieira (approved):
LGTM! Just added on small nit comment to improve code docs.

Sources/StreamChat/Workers/Background/ConnectionRecoveryHandler.swift
laevandus: I will slim it down.

PR #4113, opened 2026-05-26
Support predefined filters in ChannelListQuery

laevandus:
### 🔗 Issue Links

https://linear.app/stream/issue/IOS-1706

### 🎯 Goal

Let apps reference a server-side predefined channel-list filter by name and have the server resolve its filter/sort templates, instead of constructing the filter client-side.

### 📝 Summary

- New `ChannelListQuery(predefinedFilter:filterValues:sortValues:)` initializer.
- Server-resolved filter/sort returned on the `predefined_filter` response are decoded and re-applied locally for Core Data caching.
- `ChatChannelListController.query` and `ChannelListState.query` become `internal(set)` so the SDK can swap to the resolved query after the first response.
- `Filter` decoder now supports multi-key objects as implicit `$and` and the `{ key: value }` short form as implicit `$eq`, matching the Stream filter JSON shape.

### 🛠 Implementation

- `ChannelListQuery+PredefinedFilter.swift` decodes persisted filter/sort JSON and re-attaches Core Data wiring (`keyPathString`, `valueMapper`, `predicateMapper`) via `ChannelListFilterScope.predefinedFilterKeyMapping`. Unknown keys are logged and dropped.
- The `predefinedFilterKeyMapping` / `predefinedSortingKeyMapping` registries are generated by **Sourcery** from the channel-list `FilterKey` / `ChannelListSortingKey` declarations (`Generated/PredefinedFilter+Generated.swift`). Adding a key no longer needs a manual map edit — `make generate` regenerates them and CI fails if the committed output is stale.
- `ChannelListPayload` decodes the new `predefined_filter` response key. `ChannelListUpdater` persists the resolved filter+sort onto `ChannelListQueryDTO` and returns an updated `ChannelListQuery` to the controller/state layer.
- On controller/state init, `loadPredefinedFilter(for:)` rehydrates the locally cached resolved query so Core Data fetch predicates work before the first server response lands.

### 🧪 Manual Testing Notes

- The demo app's channel list picker exposes a "Predefined filter" entry — pick one and verify that the channel list loads, paginates, and updates on websocket events.
- For a plain (non-predefined) query, behaviour should be unchanged.

New pre-defined filters can be added through the [dashboard](https://beta.dashboard.getstream.io/organization/59390/102399/chat/predefined-filters/user_per_channel_type_channels/).

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **New Features**
  * Added support for predefined channel list filters via `ChannelListQuery(predefinedFilter:filterValues:sortValues:)`, enabling server-side filtering with substitutable parameter values.
  * Channel lists can now be filtered using predefined messaging and livestream filter templates.

* **Documentation**
  * Updated changelog with predefined filter query support reference.


Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift
```diff
@@ -146,14 +153,11 @@ public class ChatChannelListController: DataController, DelegateCallable, DataSt
         startChannelListObserverIfNeeded()
         channelListLinker.start(with: client.eventNotificationCenter)
         client.syncRepository.startTrackingChannelListController(self)
-        updateChannelList { [weak self] error in
-            guard let completion else { return }
-            self?.callback {
-                completion(error)
-            }
+        updateChannelList { result in
+            completion?(result.error)
```
martinmitrevski: I think I've mentioned this in the other PR as well, why not `self?.callback`?

Sources/StreamChat/Query/Sorting/ChannelListSortingKey.swift
```diff
@@ -78,6 +78,23 @@ extension ChannelListSortingKey {
 }
 
 extension ChannelListSortingKey {
+    /// Registry of every hardcoded channel-list sorting key, keyed by the server-side `remoteKey`.
+    ///
+    /// - Important: Always add new sorting keys to the map.
+    static let predefinedSortingKeyMapping: [String: Self] = Dictionary(
+        uniqueKeysWithValues: [
+            .cid,
```
martinmitrevski: are these documented/checked by the backend?

Sources/StreamChat/Query/ChannelListQuery+PredefinedFilter.swift
```diff
@@ -0,0 +1,86 @@
...
+            predicateMapper: coreDataMetadata.predicateMapper
+        )
+    }
+}
+
+/// Type-erased Core Data metadata extracted from a typed `FilterKey<ChannelListFilterScope, _>`.
+/// The metadata is what `Filter+predicate.swift` consumes when building `NSPredicate`s.
+struct ChannelListFilterKeyCoreDataMetadata: Sendable {
+    let keyPathString: String?
+    let valueMapper: (@Sendable (Any) -> FilterValue?)?
```
martinmitrevski: why Any? Shouldn't we be more specific here?

Sources/StreamChat/Query/ChannelListQuery.swift
```diff
@@ -325,3 +409,48 @@ internal extension FilterKey where Scope == ChannelListFilterScope {
...
+    /// Used to re-attach Core Data wiring after decoding predefined filter JSON.
+    ///
+    /// - Important: Always add new filter keys to the map.
+    static let predefinedFilterKeyMapping: [String: ChannelListFilterKeyCoreDataMetadata] = {
+        func map<Value: FilterValue>(_ key: FilterKey<ChannelListFilterScope, Value>) -> (String, ChannelListFilterKeyCoreDataMetadata) {
+            (key.rawValue, ChannelListFilterKeyCoreDataMetadata(key))
+        }
+        return Dictionary(
+            uniqueKeysWithValues: [
+                map(.archived),
```
martinmitrevski: not pretty - wonder how we can make sure to update this list when introducing new keys

Sources/StreamChat/Query/Filter.swift
```diff
@@ -507,35 +507,62 @@ extension Filter: Codable {
...
-                self.init(
-                    operator: rightSide.operator,
-                    key: key.stringValue,
-                    value: rightSide.value,
-                    isCollectionFilter: false
-                )
-                return
-            }
+
+        let keys = container.allKeys
```
martinmitrevski, line 515: why are all these changes needed?

Sources/StreamChat/StateLayer/ChannelListState+Observer.swift
```diff
@@ -6,15 +6,15 @@ import Foundation
 
 extension ChannelListState {
     final class Observer {
-        private let channelListObserver: StateLayerDatabaseObserver<ListResult, ChatChannel, ChannelDTO>
+        private var channelListObserver: StateLayerDatabaseObserver<ListResult, ChatChannel, ChannelDTO>
         private let clientConfig: ChatClientConfig
         private let channelListLinker: ChannelListLinker
         private let channelListUpdater: ChannelListUpdater
         private let database: DatabaseContainer
-        private let dynamicFilter: ((ChatChannel) -> Bool)?
```
martinmitrevski, line 14: no more dynamic filter?

Sources/StreamChat/APIClient/Endpoints/Payloads/ChannelListPayload.swift
```diff
@@ -7,24 +7,48 @@ import Foundation
...
-            channels: channels
+            channels: channels,
+            predefinedFilter: predefinedFilter
         )
     }
 }
 
+final class PredefinedFilterPayload: Decodable, Sendable {
+    let name: String
+    let filter: [String: RawJSON]
```
martinmitrevski, line 42: we won't use the typesafe Filter we use for the other cases? (Same for sorting)

martinmitrevski (commented):
looks good, but some changes at tricky places - we should do extensive testing here

Sources/StreamChat/APIClient/Endpoints/Payloads/ChannelListPayload.swift
laevandus: This gets saved in CoreData first so it is better to keep the form what backend gives us. The typed filter transformation can change the encoded JSON format (e.g. decoding adds implicit $eq).

Other case is that OpenAPI decoding will give me the same form so it is easier to migrate that later on.

Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift
laevandus: Adding back, got dropped with refactoring.

Sources/StreamChat/Query/ChannelListQuery.swift
laevandus: These are coming from our own set of filter keys.
```
public extension FilterKey where Scope == ChannelListFilterScope {
    static var cid: FilterKey<Scope, ChannelId> { .init(rawValue: "cid", keyPathString: #keyPath(ChannelDTO.cid), valueMapper: { $0.rawValue }) }
```

The most fragile and weakest point of the  predefined filter feature is that converting the JSON representation of the filter and sort to our type-safe types, we must also set CoreData keypaths. There is not a simple way for this because filter key definitions are static variables and no way to just enumerate over this. Somewhere needs to be a map/array which returns them all so we can use them during the conversion.

Options:
a) custom swiftlint rule which does parsing and throws and error when the mapping function does not contain just added key
b) …

Sources/StreamChat/Query/Filter.swift
laevandus: Backend implementation can return JSON which looks like:
```
{ "type": "messaging", "members": { "$in": ["amy"]
  } }
```
which should be treated as $and.

Current parsing would only pick the first node. These changes fixes it by turning that case into explicit $and. $eq has a similar case where the operator can be omitted from the JSON. 

I will see how to make this code more understandable.

Sources/StreamChat/StateLayer/ChannelListState+Observer.swift
laevandus: It is just unused variable here, it is passed into ChannelListLinker in the init method. Probably should revert this cleanup noise.

Sources/StreamChat/Controllers/ChannelListController/ChannelListController.swift
```diff
@@ -82,8 +82,15 @@ public class ChatChannelListController: DataController, DelegateCallable, DataSt
     }
 
     private(set) lazy var channelListObserver: BackgroundListDatabaseObserver<ChatChannel, ChannelDTO> = {
+        if let updated = worker.loadPredefinedFilter(for: query) {
+            query = updated
+        }
```
laevandus, line 87: For supporting app relaunches where API call has not finished and we load the filter from CoreData and pass it to FRC.

laevandus: I had 2 options here:
a) load it async, but then FRC needs to be recreated/reloaded anyway (2 times, one without correct local filter, second with loaded filter) (this is because only CoreData knows the fetches filter JSON)
b) block, load filter, create FRC once

Sources/StreamChat/Query/ChannelListQuery.swift
laevandus: I will explore this a bit more.

Sources/StreamChat/Query/Filter.swift
laevandus: OK, not really, but I did add more tests since there were some small issues with the implicit $and handling.

Sources/StreamChat/StateLayer/ChannelList.swift
```diff
@@ -9,7 +9,6 @@ public class ChannelList: @unchecked Sendable {
     private let channelListUpdater: ChannelListUpdater
     private let client: ChatClient
     @MainActor private var stateBuilder: StateBuilder<ChannelListState>
-    let query: ChannelListQuery
```
laevandus, line 12: I was only for sync repo and now it needs to be mutable. Easiest was to just get rid of it.

Sources/StreamChat/Generated/PredefinedFilter+Generated.swift
```diff
@@ -0,0 +1,76 @@
+// Generated using Sourcery 2.3.0 — https://github.com/krzysztofzablocki/Sourcery
+// DO NOT EDIT
+
+// Run `make generate` after changing channel-list FilterKey or ChannelListSortingKey members.
```
martinmitrevski, line 4: this is a good solution

martinmitrevski (approved):
lgtm ✅  another one for @testableapple to keep an eye on.

.github/workflows/smoke-checks.yml
```diff
@@ -71,6 +71,7 @@ jobs:
       - run: bundle exec fastlane rubocop
       - run: bundle exec fastlane run_swift_format strict:true
       - run: bundle exec fastlane validate_public_interface
+      - run: bundle exec fastlane validate_generated_code
```
testableapple, line 74: 🔥

testableapple (approved):

PR #4118, opened 2026-06-01
Respect `shouldShowShadowedMessages` in livestream channels

laevandus:
### 🔗 Issue Links

- [IOS-1738](https://linear.app/stream/issue/IOS-1738)

### 🎯 Goal

Make livestream channels respect the `shouldShowShadowedMessages` configuration so shadowed messages are hidden by default, matching the database-backed controllers.

### 📝 Summary

- `LivestreamChatHandler` now filters shadowed messages out of its in-memory list when `ChatClientConfig.shouldShowShadowedMessages` is `false`.
- Filtering is applied at every point a message enters the list: cache population, channel payload (initial/paginated fetch), and `MessageNewEvent`
- A message that is updated into a shadowed state is now removed from the list.
- Fixes both `LivestreamChannelController` and `LivestreamChat`, which share the handler.

### 🛠 Implementation

Added a private `isMessageVisible(_:)` helper that mirrors the shadowed branch of `MessageDTO.channelMessagesPredicate` and reads `client.config.shouldShowShadowedMessages`. It is applied in `populateFromCacheIfEnabled`, `handleChannelPayload` (without altering the pagination cursor, which must track the full server page), `handleNewMessage`, and `handleUpdatedMessage`.

`shouldShowShadowedMessages` is the only `ChatClientConfig` flag the channel message-list predicate consults; the remaining predicate conditions (reply-in-channel, message types, drafts, truncation, deleted handling) are structural and already filtered server-side in payloads, so they are intentionally out of scope.

### 🎨 Showcase

N/A — non-UI behavior change.

### 🧪 Manual Testing Notes

On a livestream channel with a shadow-banned user, send messages from that user. With `shouldShowShadowedMessages = false` (default) they should not appear in the message list; with `true` they should appear. 

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Livestream channels now consistently respect the setting to show or hide shadowed/hidden messages during loading, incoming messages, and message updates.

* **Tests**
  * Added tests verifying shadowed message visibility behavior across payload handling, cache population, incoming messages, and updates.

* **Documentation**
  * Added an “Upcoming” changelog entry noting the fix.

### DIFF
--- a/Sources/StreamChat/Controllers/ChannelController/LivestreamChatHandler.swift
+++ b/Sources/StreamChat/Controllers/ChannelController/LivestreamChatHandler.swift
@@ -173,7 +173,7 @@ final class LivestreamChatHandler: LivestreamChatHandling, DataStoreProvider, @u
             return
         }
         channel = cachedChannel
-        messages = cachedChannel.latestMessages
+        messages = cachedChannel.latestMessages.filter(isMessageVisible)
     }
 
     /// Applies the freshly-fetched channel payload to the in-memory state.
@@ -190,7 +190,7 @@ final class LivestreamChatHandler: LivestreamChatHandling, DataStoreProvider, @u
 
         let newMessages = payload.messages.compactMap {
             $0.asModel(cid: payload.channel.cid, currentUserId: currentUserId, channelReads: newChannel.reads)
-        }
+        }.filter(isMessageVisible)
 
         updateMessagesArray(with: newMessages, pagination: channelQuery.pagination)
     }
@@ -261,6 +261,13 @@ final class LivestreamChatHandler: LivestreamChatHandling, DataStoreProvider, @u
         DispatchQueue.main.async {
             callback(handlers)
         }
+    }
+
+    private func isMessageVisible(_ message: ChatMessage) -> Bool {
+        if message.isShadowed, !client.config.shouldShowShadowedMessages {
+            return false
+        }
+        return true
     }
 
     private func updateMessagesArray(with newMessages: [ChatMessage], pagination: MessagesPagination?) {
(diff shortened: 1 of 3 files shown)


Sources/StreamChat/Controllers/ChannelController/LivestreamChatHandler.swift
```diff
@@ -507,6 +518,13 @@ final class LivestreamChatHandler: LivestreamChatHandling, DataStoreProvider, @u
 
     private func handleUpdatedMessage(_ updatedMessage: ChatMessage) {
         if let index = messages.firstIndex(where: { $0.id == updatedMessage.id }) {
+            // If the message became hidden (e.g. shadowed while
+            // `shouldShowShadowedMessages` is disabled), remove it
+            guard isMessageVisible(updatedMessage) else {
```
martinmitrevski, line 521: shouldn't this be added in the `handleNewMessage` method as well?

Sources/StreamChat/Controllers/ChannelController/LivestreamChatHandler.swift
```diff
@@ -263,6 +263,13 @@ final class LivestreamChatHandler: LivestreamChatHandling, DataStoreProvider, @u
         }
     }
 
+    private func isMessageVisible(_ message: ChatMessage) -> Bool {
+        if message.isShadowed, !client.config.shouldShowShadowedMessages {
+            return false
+        }
+        return true
+    }
```
nuno-vieira, line 271: The description mentions that both MessageNewEvent / NewMessagePendingEvent are handled.

Only MessageNewEvent should be handled; the NewMessagePendingEvent is the locally created message, so it will never be shadowed

laevandus: OK, I'll move the check

Sources/StreamChat/Controllers/ChannelController/LivestreamChatHandler.swift
laevandus: It is there. I'll move things a bit for making it clearer

nuno-vieira (approved):
LGTM! ✅

martinmitrevski (approved):

PR #4127, opened 2026-06-09
Fix search-result icon following channel's latest message instead of matched message

### LINKED ISSUE #3843: [Bug]: Search results render wrong message type icon after sending a voice message
### Actual behaviour

User A sends message: "Hello 1" to User B.
→ User B searches "Hello" → result shows correctly

Image

Image

User A sends a voice message to User B.
→ User B searches "Hello" again → result shows wrongly: rendered with mic icon, even though it’s text ⚠️

Image

Image

User A sends another text message "Hello 2".
→ User B searches "Hello" → result shows correctly again

Image

Image

### Expected behaviour

Search results should always reflect the message type of the matched message, not depend on the latest message in the channel.

### Steps to reproduce

```bash
See description
```

### Which packages are you using?

StreamChatUI

### Which SDK version are you using?

4.90.0

### Which iOS version is the issue reproducible?

iOS 26

### Which iPhone Device is the issue reproducible?

-

### Which Xcode version are you using?

-

tsushanth:
## Why

Fixes #3843.

\`ChatChannelListItemView.subtitleIcon\` was reading \`isLastMessageVoiceRecording\`, which only inspects the channel's \`previewMessage\` (latest message). When a search result rendered a plain-text hit while the channel's last message happened to be a voice recording, the search-result cell still showed the mic icon next to the text — exactly the screenshots in the issue.

## What

In \`subtitleIcon\`, when \`content.searchedMessage\` is non-nil (we're rendering a search hit, not the normal channel preview), derive the icon from the matched message's attachments instead of the channel's last message. Non-search behavior is unchanged.

10-line diff in [\`ChatChannelListItemView.swift\`](https://github.com/GetStream/stream-chat-swift/blob/develop/Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift#L204).

## Test

I don't have access to a Stream backend to run the E2E repro from the issue, but the behavior is fully determined by \`content.searchedMessage\` and \`searchedMessage.voiceRecordingAttachments\` so the logic is straightforward to step through against the screenshots. Happy to add a snapshot test under \`Tests/StreamChatUITests/\` if you'd like — point me at a comparable existing search-result snapshot and I'll mirror it.

## Compat

No public API change; \`subtitleIcon\` is already \`open var\`, the override surface is identical.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed mic icon display in search results so it appears only when the searched message includes a voice recording attachment.
  * Preserved existing behavior in regular channel lists: the mic icon continues to show when the last message is a voice recording, keeping visuals consistent outside search results.

### DIFF
--- a/Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
+++ b/Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
@@ -202,10 +202,6 @@ open class ChatChannelListItemView: _View, ThemeProvider {
     }
 
     open var subtitleIcon: UIImage? {
-        // For search results, the cell renders the matched message rather than the
-        // channel's latest message — so the icon must follow the matched message,
-        // not the channel's last activity (which would, e.g., show a mic icon next
-        // to a plain text search hit when the channel's last message is voice).
         if let searchedMessage = content?.searchedMessage {
             return searchedMessage.voiceRecordingAttachments.isEmpty
                 ? nil
--- a/Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
+++ b/Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
@@ -202,7 +202,16 @@ open class ChatChannelListItemView: _View, ThemeProvider {
     }
 
     open var subtitleIcon: UIImage? {
-        isLastMessageVoiceRecording ? appearance.images.mic : nil
+        // For search results, the cell renders the matched message rather than the
+        // channel's latest message — so the icon must follow the matched message,
+        // not the channel's last activity (which would, e.g., show a mic icon next
+        // to a plain text search hit when the channel's last message is voice).
+        if let searchedMessage = content?.searchedMessage {
+            return searchedMessage.voiceRecordingAttachments.isEmpty
+                ? nil
+                : appearance.images.mic
+        }
+        return isLastMessageVoiceRecording ? appearance.images.mic : nil
     }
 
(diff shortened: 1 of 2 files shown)


Sources/StreamChatUI/ChatChannelList/ChatChannelListItemView.swift
```diff
@@ -202,7 +202,16 @@ open class ChatChannelListItemView: _View, ThemeProvider {
     }
 
     open var subtitleIcon: UIImage? {
-        isLastMessageVoiceRecording ? appearance.images.mic : nil
+        // For search results, the cell renders the matched message rather than the
+        // channel's latest message — so the icon must follow the matched message,
+        // not the channel's last activity (which would, e.g., show a mic icon next
+        // to a plain text search hit when the channel's last message is voice).
```
laevandus: Could you please drop the comment and add an entry to CHANGELOG.md (feel free to reword it)

```
## StreamChat
### 🐞 Fixed
- Fix showing voice recording icon in search results [#4127](https://github.com/GetStream/stream-chat-swift/pull/4127)
```

Thank you for your contribution

tsushanth: Done — dropped the inline comment in 939fd4… and added the changelog entry under `# Upcoming > StreamChat > 🐞 Fixed` with the suggested wording. Thanks for the review

laevandus:
Thank you for your contribution. I ran [CI](https://github.com/GetStream/stream-chat-swift/actions/runs/27607386605) checks separately and all of them passed.

PR #4132, opened 2026-06-16
Respect channel capabilities before sending typing events

laevandus:
### 🔗 Issue Links

Resolves: [IOS-1766](https://linear.app/stream/issue/IOS-1766/)

### 🎯 Goal

Stop sending `typing.start` / `typing.stop` events to channels whose `own_capabilities` don't include `send-typing-events` (e.g. livestream channels).

### 📝 Summary

- `ChatChannelController` no longer sends typing events when the channel model isn't loaded yet, instead of optimistically assuming typing is allowed.
- The async `Chat` state layer now also respects the channel's typing capability before sending typing events.

### 🛠 Implementation

The capability gate in `ChannelController.shouldSendTypingEvents` fell back to `true` when the channel had not yet been loaded from the local store (`channel == nil`). In that window, keystroke/start/stop events were sent regardless of `own_capabilities`, so a livestream channel (which lacks `send-typing-events`) still received them and the backend rejected them. The fallback is now `false`, matching `LivestreamChannelController`, `LivestreamChat`, and the typing-indicator display path.

The async `Chat.keystroke` / `Chat.stopTyping` had no capability check at all; they now guard on `canSendTypingEvents`, mirroring `LivestreamChat`.

**Backend verification:** Confirmed against the backend that this is config-driven, not role-driven. The `livestream` channel type defaults to `TypingEvents: false`, and the `send-typing-events` own-capability is derived directly from that flag. The server rejects `typing.start` / `typing.stop` for such channels before any permission check (event-support validation), which is the error this fixes client-side. The backend also folds the user's typing-indicator privacy setting into the same capability, so the SDK's existing `canSendTypingEvents` check already covers both the channel-config and privacy cases.

### 🧪 Manual Testing Notes

Connect to a channel whose type has typing events disabled (or a livestream channel) and type in the composer. No `typing.start` / `typing.stop` request should be issued.

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Stopped typing indicators from being sent when the channel doesn’t support typing events.
  * Updated typing behavior so typing actions no longer proceed when the channel is missing or typing events are disabled.
* **Tests**
  * Added/expanded tests covering keystroke/start/stop typing behavior for nil (not loaded) channels and for enabled vs. disabled typing capabilities.
* **Documentation**
  * Updated the changelog with the new “Upcoming → StreamChat → 🐞 Fixed” entry.

### DIFF
--- a/Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
+++ b/Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
@@ -151,7 +151,7 @@ public class ChatChannelController: DataController, DelegateCallable, DataStoreP
     /// A boolean value indicating if it should send typing events.
     /// It is `true` if the channel typing events are enabled as well as the user privacy settings.
     func shouldSendTypingEvents(completion: @escaping @Sendable (Bool) -> Void) {
-        guard channel?.canSendTypingEvents ?? true else {
+        guard channel?.canSendTypingEvents ?? false else {
             completion(false)
             return
         }
--- a/Sources/StreamChat/StateLayer/Chat.swift
+++ b/Sources/StreamChat/StateLayer/Chat.swift
@@ -1356,6 +1356,7 @@ public class Chat: @unchecked Sendable {
     ///
     /// - Throws: An error while communicating with the Stream API.
     public func keystroke(parentMessageId: MessageId? = nil) async throws {
+        guard await state.channel?.canSendTypingEvents ?? false else { return }
         try await typingEventsSender.keystroke(in: cid, parentMessageId: parentMessageId)
     }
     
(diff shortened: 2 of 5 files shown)


Sources/StreamChat/Controllers/ChannelController/ChannelController.swift
```diff
@@ -151,7 +151,7 @@ public class ChatChannelController: DataController, DelegateCallable, DataStoreP
     /// A boolean value indicating if it should send typing events.
     /// It is `true` if the channel typing events are enabled as well as the user privacy settings.
     func shouldSendTypingEvents(completion: @escaping @Sendable (Bool) -> Void) {
-        guard channel?.canSendTypingEvents ?? true else {
+        guard channel?.canSendTypingEvents ?? false else {
```
laevandus, line 154: This can be called before channel data is fetched

Sources/StreamChat/StateLayer/Chat.swift
```diff
@@ -1356,6 +1356,7 @@ public class Chat: @unchecked Sendable {
     ///
     /// - Throws: An error while communicating with the Stream API.
     public func keystroke(parentMessageId: MessageId? = nil) async throws {
+        guard await state.channel?.canSendTypingEvents ?? false else { return }
```
laevandus, line 1359: Another issue is that Chat lacked any capability checks

martinmitrevski (approved):
lgtm

PR #4133, opened 2026-06-16
OpenAPI: Endpoint and /app migration

laevandus:
### 🔗 Issue Links

[IOS-1620](https://linear.app/stream/issue/IOS-1620/openapi-generation-in-llc)

### 🎯 Goal

Start incremental OpenAPI adoption in StreamChat by migrating only the `/app` endpoint while keeping existing endpoint factories on `EndpointPath`.

### 📝 Summary

- Generate the OpenAPI Swift endpoint/model files for the `/app` migration and prune generated output to `getApp` plus its model closure.
- Make generated `Endpoint` use `EndpointPath` again, with `Scripts/generate.sh` injecting the existing v1 paths into the generated enum until each endpoint is migrated.
- Migrate app settings to `Endpoint.getApp()` / `GetApplicationResponse` while preserving the public `AppSettings` model.

### 🛠 Implementation

`Scripts/generate.sh` now generates the Swift OpenAPI client, drops the unused async API client, prunes endpoints/models to the current `/app` scope, strips public access modifiers, formats output, and injects the v1 `EndpointPath` cases into the generated enum.

`DefaultEndpoints.swift` owns `Endpoint`, `EndpointMethod`, and the combined `EndpointPath`. Existing hand-written endpoints continue using v1 enum cases, while the migrated app endpoint uses generated `.getApp` with `/api/v2/app`.

### 🧪 Manual Testing Notes

Run the demo app and verify the app settings request succeeds.

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **Changed**
  * Updated app settings retrieval to use the v2 “getApp” API endpoint.
  * Refreshed the “Upcoming” StreamChat note to reference the v2 app endpoint for app settings.
* **Bug Fixes**
  * Adjusted offline/queued handling so app settings requests are classified correctly.
* **Tests**
  * Updated endpoint tests to validate `getApp`, including offline-queue and path expectations.
  * Removed outdated app settings payload test coverage.


Sources/StreamChat/APIClient/Endpoints/EndpointPath+OfflineRequest.swift
```diff
@@ -13,16 +13,10 @@ extension EndpointPath {
              .deleteChannel, .channelUpdate, .muteChannel, .showChannel, .truncateChannel, .markChannelRead, .markChannelUnread,
              .markAllChannelsRead, .markChannelsDelivered, .channelEvent, .stopWatchingChannel, .pinnedMessages, .uploadChannelAttachment, .message,
              .replies, .reactions, .messageAction, .banMember, .flagUser, .flagMessage, .muteUser, .translateMessage,
-             .callToken, .createCall, .deleteFile, .deleteImage, .og, .appSettings, .threads, .thread, .markThreadRead, .markThreadUnread,
```
martinmitrevski, line 16: how are we going to handle this going forward?

Sources/StreamChat/Generated/OpenAPI/APIs/DefaultEndpoints.swift
```diff
@@ -4,60 +4,64 @@
...
 
     init(from decoder: Decoder) throws {
         let container = try decoder.container(keyedBy: CodingKeys.self)
-        path = try container.decode(EndpointPath.self, forKey: .path)
+        path = try container.decode(String.self, forKey: .path)
         method = try container.decode(EndpointMethod.self, forKey: .method)
         queryItems = try container.decodeIfPresent(Data.self, forKey: .queryItems)
         requiresConnectionId = try container.decode(Bool.self, forKey: .requiresConnectionId)
         requiresToken = try container.decode(Bool.self, forKey: .requiresToken)
+        shouldBeQueuedOffline = try container.decode(Bool.self, forKey: .shouldBeQueuedOffline)
```
martinmitrevski: oh, I see, it's coming from here. I guess this would still need to be done manually for all, right?

Sources/StreamChat/Repositories/OfflineRequestsRepository.swift
```diff
@@ -245,27 +243,71 @@ class OfflineRequestsRepository: @unchecked Sendable {
...
-                    guard let json = try? JSONSerialization.jsonObject(with: bodyData) as? [String: Any] else { return nil }
-                    guard let message = json["message"] as? [String: Any] else { return nil }
-                    return message["id"] as? String
-                default:
-                    return nil
+            recoveryAction = Self.recoveryAction(for: endpoint)
+        }
+
+        // Match the meaningful pattern at the end of the path so any prefix (e.g. `/api/v2`) is ignored.
+        private static let sendMessagePattern = "channels/([^/]+)/([^/]+)/message$"
```
martinmitrevski: hmm, is this safe enough to do?

martinmitrevski (commented):
looks good in general, although I am not sure about that regex matching

Sources/StreamChat/Repositories/OfflineRequestsRepository.swift
laevandus: That's the difficult part of it. 
With EndpointPath enum it is slightly easier, but hard to have 2 sets of enums: the current and the other one coming from the OpenAPI generation.

I will think about it. Maybe it is still better if EndpointPath comes from OpenAPI generated code. I think I will try that again and then try to have current v1 paths in it as well.

Sources/StreamChat/Generated/OpenAPI/APIs/DefaultEndpoints.swift
laevandus: Reverted this back to old way as well.

Sources/StreamChat/APIClient/Endpoints/EndpointPath+OfflineRequest.swift
laevandus: I brought the EndpointPath back so it will continue the same way now.

Sources/StreamChat/Generated/OpenAPI/models/AppResponseFields.swift
```diff
@@ -0,0 +1,55 @@
+//
+// Copyright © 2026 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+final class AppResponseFields: @unchecked Sendable, Codable, JSONEncodable, Hashable {
+    var asyncUrlEnrichEnabled: Bool
```
martinmitrevski, line 8: why aren't these let?

martinmitrevski (commented):
lgtm, just one small general comment about the vars

Sources/StreamChat/Generated/OpenAPI/models/AppResponseFields.swift
laevandus: I looked into generator. There is a special handling for `type` arguments (related to WS events) what get dropped from init. For cleanest fix here is to change generator and then this code gets updated. Opened [CHA-3547](https://linear.app/stream/issue/CHA-3547). I am proposing to keep this var for now in this PR and when generator is updated, then another PR cleans this up.

laevandus: Opened generator a PR for the generator.

martinmitrevski (approved):
ok, let's do the change from var to let in a follow up

PR #4149, opened 2026-06-29
Call attachment download completion even when the controller is not retained

laevandus:
### 🔗 Issue Links

Fixes: [IOS-1802](https://linear.app/stream/issue/IOS-1802/)

### 🎯 Goal

Ensure `ChatMessageController.downloadAttachment` always calls its completion handler, even when the controller is not externally retained until the download finishes.

### 📝 Summary

- `ChatMessageController.downloadAttachment(_:request:completion:)` now captures `self` strongly in its one-shot completion closure, so the completion handler is always called even when the controller is not retained by the caller.
- Scope is intentionally limited to this single function for now; the remaining controller methods will be addressed in follow-up changes.

### 🛠 Implementation

A root cause behind IOS-1802 is `[weak self]` callback handling in the controllers. `downloadAttachment` captured `self` weakly when forwarding the download response to the user's completion handler (`self?.callback { completion(...) }`). The `callback` helper does not retain the controller, so if the controller was deallocated before the response arrived, the completion was silently never called — forcing callers to manually retain the controller and breaking the "create controller → call method → throw it away" usage.

Capturing `self` strongly only extends the controller's lifetime until the download completes (the updater releases the closure after the response) — it is not a permanent retain cycle. A unit test asserts the completion still fires after the only external reference to the controller is dropped, and that the controller is then deallocated.

### 🧪 Manual Testing Notes

Try the download button on the attachment preview view in the SwiftUI SDK.

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed attachment download completion so it’s still called even if the message controller is released before the download finishes.
  * Added a changelog entry documenting the attachment download completion issue.

* **Tests**
  * Added a new test covering the completion callback behavior when the controller is not retained.
  * Updated/extended mocks and tests to record and validate the download completion callback behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### 🐞 Fixed
 - Fix sending typing events to channels without typing events capability [#4132](https://github.com/GetStream/stream-chat-swift/pull/4132)
 - Fix `typing.stop` being sent unconditionally on every message send [#4146](https://github.com/GetStream/stream-chat-swift/pull/4146)
+- Fix attachment download completion handler not being called when the message controller is not retained until the download finishes [#4149](https://github.com/GetStream/stream-chat-swift/pull/4149)
 
 ## StreamChatUI
 ### 🐞 Fixed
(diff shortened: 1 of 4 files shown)


nuno-vieira (commented):
I'm a bit concerned of changing so many completion blocks, it might add unnecessary risk, can we just do this for the download file? We have no other reports related to the other completion blocks

laevandus:
> I'm a bit concerned of changing so many completion blocks, it might add unnecessary risk, can we just do this for the download file? We have no other reports related to the other completion blocks

Noted. I will trim it down to the one I need tomorrow.

nuno-vieira (approved):

PR #4154, opened 2026-07-02
Fix unread messages banner disappearing too early after mark-as-unread

nuno-vieira:
### 🔗 Issue Links

[IOS-1817](https://linear.app/stream/issue/IOS-1817/regression-uikit-unread-seperator-disappears-on-channel-open)

### 🎯 Goal

Fix a regression where, after marking a message as unread, leaving the channel, and reopening it, the unread messages banner disappeared instantly instead of staying visible until the channel is opened again.

Regression caused by this PR: https://github.com/GetStream/stream-chat-swiftui/pull/1501

### 📝 Summary

- `ChatChannelVC.didUpdateChannel(_:)` recomputed the unread messages banner anchor (`firstUnreadMessageId`) on every channel update, including the local optimistic DB write triggered by the screen's own `markRead()` call. As soon as `markRead()` succeeded (which can happen almost immediately on open, since the newest message is typically already visible), the channel's unread count dropped to `0` and the banner was force-hidden within the same session.
- `updateUnreadMessagesBannerRelatedComponents(channel:forceUpdate:)` now keeps the banner anchored once set for the current channel screen session, and only recalculates it on a fresh channel open, or when `forceUpdate` is passed because the unread boundary genuinely moved (a new mark-as-unread action, locally or from a remote `NotificationMarkUnreadEvent`).
- Added a regression test that reproduces the exact scenario (verified it fails on the previous implementation).

### 🛠 Implementation

This mirrors the equivalent, intentional behavior already present in the SwiftUI SDK's `ChatChannelViewModel`, which explicitly keeps `firstUnreadMessageId` set after calling `channelController.markRead()`:

```swift
throttler.execute { [weak self] in
    self?.channelController.markRead()
    // We keep `firstUnreadMessageId` value set which keeps showing the new messages header in the channel view
}
```

`updateAllUnreadMessagesRelatedComponents(channel:forceUpdateBanner:)` threads a `forceUpdateBanner` flag through to `updateUnreadMessagesBannerRelatedComponents`, used only by the two call sites where the unread boundary actually changes (mark-as-unread action, remote mark-unread event).

### 🧪 Manual Testing Notes

1. Mark a message as unread in a channel.
2. Leave the channel (go back to the channel list).
3. Reopen the channel.
4. The unread messages banner should stay visible for this session, and only disappear the next time the channel is opened.

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where the unread messages banner could disappear instantly after marking a message as unread and reopening a channel.
  * Ensured the unread banner stays visible when a channel’s unread state is updated in the background.

* **Tests**
  * Added a regression snapshot test to verify unread banner visibility persists after background channel updates.

* **Documentation**
  * Updated the upcoming changelog entry with the unread banner fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### 🐞 Fixed
 - Fix `Spacer()` conflicting with SwiftUI's `Spacer` when building with Xcode 27 [#4131](https://github.com/GetStream/stream-chat-swift/pull/4131)
 - Fix showing voice recording icon in search results [#4127](https://github.com/GetStream/stream-chat-swift/pull/4127)
+- Fix unread messages banner disappearing instantly after marking a message as unread and reopening the channel [#4154](https://github.com/GetStream/stream-chat-swift/pull/4154)
 
 # [5.5.1](https://github.com/GetStream/stream-chat-swift/releases/tag/5.5.1)
 _June 11, 2026_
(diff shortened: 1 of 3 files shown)


laevandus (approved):
Verified that this scenario works. ✅

PR #4155, opened 2026-07-03
Strip unused properties from generated OpenAPI models

laevandus:
### 🔗 Issue Links

Related: [IOS-1620](https://linear.app/stream/issue/IOS-1620/openapi-generation-in-llc)

### 🎯 Goal

Keep the generated OpenAPI models limited to the properties the SDK actually reads.

### 📝 Summary

- `openapi_generate.sh` gets a post-processing step that strips model properties listed in `Scripts/openapi_models_ignored_fields.txt` (one `Model prop prop ...` line per model) and fails the run when an entry no longer matches a generated model/property
- The ignore list is prefilled from a usage sweep of the current models; models orphaned by the stripping (`Action`, `Field`, `ImageData`, `Images`, `UserResponse`) are removed from the allow list
- Endpoints whose response body the SDK ignores entirely (`blockUsers`, `unblockUsers`) are re-pointed at StreamCore's `EmptyResponse` instead of keeping empty generated models

### 🛠 Implementation

For each ignored property the step removes the declaration (with attached doc comments), the init parameter, the assignment, and the `CodingKeys` case; a `CodingKeys` enum left without cases is dropped. It runs after the model renames and before swiftformat. The whole script now uses bash + sed/awk only.

But this brings up a question if it would be just simpler to let AI tools to generate models instead with required fields.

### 🧪 Manual Testing Notes

Run `Scripts/openapi_generate.sh` twice — the output is idempotent (`git diff` stays empty). Adding a stale entry to the ignore file aborts the run with an out-of-sync error.

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved OpenAPI generation so only the required endpoint paths/models are kept, and ignored model fields are stripped more reliably.
  * Updated generated responses so certain moderation/user actions now use `EmptyResponse` where no payload is expected.

* **Tests**
  * Updated moderation and updater test expectations, fixtures, and dummy data factories to align with the streamlined generated models and response types.


nuno-vieira:
If the PR is Draft, it won't run SDK size metrics

Scripts/openapi_models_ignored_fields.txt
```diff
@@ -0,0 +1,10 @@
+# Properties stripped from the generated OpenAPI models (step 4c in openapi_generate.sh).
+# Format: <ModelTypeName> <propertyName> ... — one model per line, everything alphabetical.
+AppResponseFields id placement
+BlockedUserResponse blockedUser user userId
+CreateDeviceRequest hardwareId voipToken
+DeviceResponse disabled disabledReason hardwareId pushProvider pushProviderName userId voip
+GetApplicationResponse duration
+GetBlockedUsersResponse duration
+GetOGResponse actions authorIcon authorLink color custom duration fallback fields footer footerIcon giphy originalHeight originalWidth pretext type
+ListDevicesResponse duration
```
laevandus, line 10: These are fields we currently don't store in CoreData and therefore don't get exposed with public API models. Keeps SDK size a bit smaller.
Some of this data should get exposed, but requires breaking changes or deprecation of existing functions over new ones which expose other type. E.g. enrichUrl public func returns LinkAttachment, but should return GetOGResponse.

Options:
a) strip as much as possible in this PR, next PRs will work exposing some of this data in public models through CoreData
b) Don't strip anything, instead work on exposing this data in public models (-76 KB is not much)

laevandus: After thinking a bit . I should probably drop this PR and instead start working on exposing this data in public models instead. 😄

PR #4161, opened 2026-07-07
Replace Device struct with the OpenAPI generated model

laevandus:
### 🔗 Issue Links

Resolves: [IOS-1871](https://linear.app/stream/issue/IOS-1871/openapi-expose-deviceresponse-data-to-public-api-layer)

### 🎯 Goal

Expose the full set of device data the backend returns through the public API instead of only `id` and `createdAt`.

### 📝 Summary

- The OpenAPI-generated `DeviceResponse` model replaces the hand-written `Device` type as the public device model in the public API layer (renamed to `Device`).
- The public `Device` now exposes additional data in the public interface: `disabled`, `disabledReason`, `hardwareId`, `pushProvider`, `pushProviderName`, `userId`, and `voip` — in addition to the existing `id` and `createdAt`.
- The new fields are persisted locally and surfaced through `CurrentChatUser.devices` / `currentDevice` and `ConnectedUser.loadDevices()`.

### 🛠 Implementation

- `Scripts/openapi_generate.sh` renames the generated `DeviceResponse` to `Device` and makes it public, following the same mechanism used for `AppSettings`.
- `Device.createdAt` is now a non-optional `Date` (the backend always returns it); the remaining new fields follow the backend's optionality.
- `DeviceDTO` and the Core Data model gained the new attributes (additive, optional/defaulted → lightweight migration).
- The generated `Device` now flows through both the connect payload and the `listDevices` endpoint, so the legacy internal `DevicePayload`/`DeviceListPayload` were removed.

### 🎨 Showcase

N/A — no UI changes.

### 🧪 Manual Testing Notes

Launch the demo app and login (device gets added). No JSON parsing issues in the console. 

Test function, can be copy-pasted into demo app and called at DemoAppCoordinator+DemoApp.swift line 81

```swift
func testDeviceEndpoints() {
        // IOS-1871 verification: exercises the device write + read paths and logs the new public fields.
        let client = StreamChatWrapper.shared.client!
        let controller = client.currentUserController()

        func logDevices(_ context: String) {
            let devices = controller.currentUser?.devices ?? []
            print("[IOS-1871] \(context): \(devices.count) device(s)")
            for device in devices {
                print(
                    """
                    [IOS-1871] Device \
                    id=\(device.id), \
                    createdAt=\(device.createdAt), \
                    userId=\(device.userId), \
                    pushProvider=\(device.pushProvider), \
                    pushProviderName=\(device.pushProviderName ?? "nil"), \
                    hardwareId=\(device.hardwareId ?? "nil"), \
                    disabled=\(device.disabled.map { "\($0)" } ?? "nil"), \
                    disabledReason=\(device.disabledReason ?? "nil"), \
                    voip=\(device.voip.map { "\($0)" } ?? "nil")
                    """
                )
            }
        }

        // Write path: addDevice → CurrentUserUpdater.addDevice → saveCurrentDevice + createDevice endpoint.
        controller.addDevice(.apn(token: Data(repeating: 1, count: 32))) { error in
            if let error = error {
                print("[IOS-1871] addDevice failed: \(error)")
            }
            // Read path: synchronizeDevices → fetchDevices → listDevices → saveCurrentUserDevices([Device]) → asModel.
            controller.synchronizeDevices { error in
                if let error = error {
                    print("[IOS-1871] synchronizeDevices failed: \(error)")
                    return
                }
                logDevices("After sync")
            }
        }
    }
```

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **New Features**
  * Device records now include additional fields for push provider details, hardware ID, user ID, VoIP status, and disablement (with disablement reason).
  * Device information is now handled end-to-end using the unified device model (instead of a separate payload type).

* **Bug Fixes**
  * Improved saving/loading for current user devices so all supported device attributes persist correctly.
  * Updated the generated device representation so the device creation timestamp is handled more defensively.


Sources/StreamChat/Database/DTOs/DeviceDTO.swift
```diff
@@ -9,6 +9,13 @@ import Foundation
 class DeviceDTO: NSManagedObject {
     @NSManaged var id: String
     @NSManaged var createdAt: DBDate?
+    @NSManaged var disabled: NSNumber?
```
laevandus, line 12: Represents `Bool?` in the generated model. Not much to do here if we wanna switch to OpenAPI model directly without any additional wrapper.

Sources/StreamChat/Models/Device+Extensions.swift
```diff
@@ -0,0 +1,29 @@
...
+            && lhs.createdAt == rhs.createdAt
+            && lhs.disabled == rhs.disabled
+            && lhs.disabledReason == rhs.disabledReason
+            && lhs.hardwareId == rhs.hardwareId
+            && lhs.pushProvider == rhs.pushProvider
+            && lhs.pushProviderName == rhs.pushProviderName
+            && lhs.userId == rhs.userId
+            && lhs.voip == rhs.voip
+    }
+}
```
laevandus: OpenAPI generator change is coming which allows me to generate this. Next PR can clean it up.

Sources/StreamChat/Generated/OpenAPI/models/Device.swift
```diff
@@ -4,25 +4,25 @@
 
 import Foundation
 
-final class DeviceResponse: Sendable, Codable, JSONEncodable {
+public final class Device: Sendable, Codable, JSONEncodable {
     /// Date/time of creation
-    let createdAt: Date
+    public let createdAt: Date
```
laevandus: Note! `createdAt` used be an optional value in the public API layer. It is actually always returned. Can we accept making it non-optional (breaking API change because it was optional before)? If not, `DeviceResponse` (this PR currently renames it directly to `Device`) can't be directly exposed and `Device` type needs to wrap it.

```swift
/// An object representing a device which can receive push notifications.
public struct Device: Codable, Equatable, Identifiable, Sendable {
    private enum CodingKeys: String, CodingKey {
        case id
        case createdAt = "created_at"
    }

    /// The device identifier.
    public let id: DeviceId
    /// The date when the device for created.
    public let createdAt: Date?
```

Sources/StreamChat/Generated/OpenAPI/models/Device.swift
```diff
@@ -4,25 +4,25 @@
...
     /// Whether device is disabled or not
-    let disabled: Bool?
+    public let disabled: Bool?
```
laevandus, line 11: Optional booleans are not nice. Only way to avoid it is to not rename DeviceResponse to Device and keep Device to wrap DeviceResponse and adding default values. WDYT? Basically related to the comment above.

Sources/StreamChat/Generated/OpenAPI/models/Device.swift
laevandus: I think I add a script to make it optional and v6 can make this small breaking change.

Scripts/openapi_generate.sh
```diff
@@ -178,6 +179,7 @@ publicize_model() {
     "$file"
 }
 publicize_model AppSettings
+publicize_model Device
```
laevandus, line 205: OpenAPI models is not used as public model for Device type

Scripts/openapi_generate.sh
```diff
@@ -156,11 +156,23 @@ prune_models() {
...
+# Relax selected generated stored properties back to optional. Some models are
+#     exposed as public API where a property was historically optional (e.g.
+#     Device.createdAt was Date? before the OpenAPI migration).
+optionalize_property() {
+  local file="$OUTPUT_DIR_CHAT/models/$1.swift"
+  sed -i '' -E \
+    -e "s/^(    let $2: [^?]+)$/\1?/" \
+    "$file"
+}
+optionalize_property DeviceResponse createdAt
```
laevandus, line 177: In v6 this can be removed. Because Device is now public model, but our current model has createdAt as optional value, I need to change the generated model to match the current public model for avoiding API breaking changes in v5. Of course, in v6 we delete this and make a small breaking change.

Sources/StreamChat/Generated/OpenAPI/models/Device.swift
laevandus: Added post-processing step to make it back to optional in v5. In v6 we can properly fix it.

martinmitrevski (approved):
lgtm ✅

laevandus:
Manually verified it (sample script and running demo app and inspecting device queries). Will merge it.

